PR #1, opened 2011-11-22
Accessing self._en_wordlist raises AttributeError on first run

timClicks:
The `if not self._en_wordlist` clause does not protect from the AttributeError
that is raised when the user attempts to access the non-existent attribute.


stevenbird:
Thanks Tim.

PR #9, opened 2012-01-10
small ReST improvements

kmike:
...building docs (instead of globally installed nltk); some extra spaces are removed, some tabs are converted to spaces


stevenbird:
Thanks, especially for the conf.py patch.

PR #10, opened 2012-01-10
sem.skolemize and inference.resolution modules are fixed, tests for nltk.tree are corrected

kmike:
Please excuse me for an unpleasant diff (there are a lot of extra whitespaces in the source code and editor auto-strips them).

sem.skolemize and inference.resolution mudules desperately needed extra imports and I've added them; tests for nltk.tree didn't work because the Tree was not imported.


stevenbird:
Thanks.  NB please wrap lines longer than 80 chars.

PR #16, opened 2012-01-11
Test improvements

kmike:
In order to run tests across multiple python interpreters it is very convenient to use the http://tox.testrun.org/ tool. I've added a tox config and a TESTING.rst file with the draft info about how to run tests. 

Some doctests are also fixed (there are still a lot of them failing though).


kmike:
github doesn't like me :)

PR #17, opened 2012-01-11
Test improvements

kmike:
In order to run tests across multiple python interpreters it is very convenient to use the http://tox.testrun.org/ tool. I've added a tox config and a TESTING.rst file with the draft info about how to run tests.

Some doctests are also fixed (there are still a lot of them failing though).

### DIFF
--- a/TESTING.rst
+++ b/TESTING.rst
@@ -1,0 +1,16 @@
+NLTK testing
+============
+
+1. obtain nltk source code;
+2. install virtualenv and tox::
+
+       pip install virtualenv
+       pip install tox
+
+3. make sure python2.5, python2.6, python2.7 and pypy executables are in system PATH.
+   It is OK not to have all the executables, tests will be executed for available interpreters.
+
+4. run 'tox' command from the root nltk folder.
+
+It may take a long time at first run, but the subsequent runs will be much faster.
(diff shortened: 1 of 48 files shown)


kmike:
just discovered a better way to fix the doctests, please delay merging

stevenbird:
OK.  Thanks for the tox config.  Please note that some doctests are user documentation and not really regression tests, and will eventually need to live in a docstring.

> > > chunkscore.recall()
> > > 0.33333333333333331

In the docstring we permit ellipses, and can write the above result instead as 0.33333...

However, its a higher priority to get all our tests working where they are, so using float_equals() is fine.

kmike:
Yeah, but this may be 0.49999999999 vs 0.5 where ellipses won't help

stevenbird:
Ready to merge?

NB I intend to tweak docstring imports of the form "from nltk.x.y import z" -> "from nltk.x import z" wherever possible.  In general, x/**init**.py explicitly imports all names from x/y.py that are to be made "public".

kmike:
Yes, I think this is ready and can be merged. There are still a lot of tests failing but the number is reduced for a good amount.

The missing numpy causes 8 test failures but it is not available for pypy (that's why python 2.7 has 13 failed doctests and pypy has 21). We can either live with this or somehow tweak the tox.ini.

PR #18, opened 2012-01-13
Wrapper in nltk.classify for scikit-learn classifiers

larsmans:
I hacked up a wrapper for the classifiers in [scikit-learn](http://scikit-learn.org), of which I happen to be one of the maintainers. It's very simple at present and might not follow NLTK coding guidelines completely, but I though I'd open this PR before completely finishing it.

The module does not try to import scikit-learn; instead, it wants the user to create a sklearn classifier and pass that to a `SklearnClassifier` wrapper object. The reason for this setup is that the number of (meta-)classifiers in scikit-learn is too huge to enumerate them.

Included is a demo on the names example, using scikit-learn's Bernoulli Naive Bayes and linear SVM algorithms.

Please let me know if there is interest in this at all.

### DIFF
--- a/nltk/classify/scikitlearn.py
+++ b/nltk/classify/scikitlearn.py
@@ -1,0 +1,86 @@
+from nltk.classify.api import ClassifierI
+
+import numpy as np
+
+
+class SklearnClassifier(ClassifierI):
+    """Wrapper for scikit-learn classifiers."""
+
+    def __init__(self, estimator, dtype=float, sparse=True):
+        """
+        :param estimator: scikit-learn classifier object.
+
+        :dtype: data type used when building feature array. Use e.g. bool
+            when all features are binary.
+
+        :sparse: Whether to use sparse matrices. Note that the estimator
+            must support these; not all scikit-learn classifiers do.
+        """
+        self._clf = estimator
+        self._dtype = dtype
+        self._sparse = sparse
+
+    def batch_classify(self, featuresets):
+        X = self._featuresets_to_array(featuresets)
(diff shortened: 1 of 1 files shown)


kmike:
@stevenbird, how do you think, would it be better to deprecate the pysvmlight classifier in favor of scikit-learn? Scikit-learn also have a binding to svmlight. I was about to propose it and then @larsmans magically opened a pull request.

larsmans:
Actually, scikit-learn binds to libsvm and liblinear, not svmlight (which is not open source).

kmike:
Yes, you're right, I confused svmlight loader with svmlight classifier. 

Actually svmlight souce code is open (e.g. it is bundled here https://bitbucket.org/wcauchois/pysvmlight/src/tip/lib/ ) but it is available only for non-commercial use.

larsmans:
Yes, the source code is available, but the license was deemed too restrictive for scikit-learn (which is entirely BSD/MIT-licensed). Well, just drop me a line here if this is interesting to you folks. I'm not on `nltk-dev` because I get enough mail already, but if necessary I can subscribe to it in digest mode.

stevenbird:
@larsmans -- thanks Lars, this is great!  I'll merge this, and we'll take care of getting it to conform with our coding standards.
@kmike -- so long as a pysvmlight interface is not costing us to maintain, I'm happy to keep it -- but feel free to discuss that on nltk-dev.

larsmans:
That's quick! I'll try to hack on a `prob_classify` member as well; except another PR soon.

PR #19, opened 2012-01-13
change "Resource not found" msg for easier copy-pasting

larsmans:
Here's a very small change to `nltk/data.py` that allows easier copy-pasting of `nltk.download()` to the Python prompt; something I do on a regular basis.

(Also, novice users might think the full stop is part of the Python expression being displayed.)


stevenbird:
Good idea, thanks.

PR #21, opened 2012-01-13
implement SklearnClassifier.batch_prob_classify

larsmans:
Alright, here's scikit-learn classification with probabilistic outputs. Is there anything else I need to do for this module? Documentation?

### DIFF
--- a/nltk/classify/scikitlearn.py
+++ b/nltk/classify/scikitlearn.py
@@ -1,4 +1,5 @@
 from nltk.classify.api import ClassifierI
+from nltk.probability import DictionaryProbDist
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks!  Some docstrings with doctests would be a great way to show users how to use the code, cf:
https://github.com/nltk/nltk/blob/master/nltk/tokenize/regexp.py

If you want to include any formal regression tests, they would go here:
https://github.com/nltk/nltk/blob/master/nltk/test/classify.doctest

larsmans:
Alright, but tests would require importing scikit-learn. Is that a problem? Is there an established way to do a conditional import?

stevenbird:
We assume tests are run on machines which have all dependencies installed.  You can see how we do conditional imports of submodules in nltk/tag/**init**.py (tag.hmm is only imported if numpy is installed -- could you do the same for numpy in classify/**init**.py please?).  This means users get some functionality on a bare-bones install.

Also, I propose that you try to import the sklearn package and give a useful error message if its not there, cf. classify/megam.py:134.  This means users get a hint about how to get the code working, without needing to read it.

PR #22, opened 2012-01-14
More features for scikit-learn wrapper

larsmans:
Ok, here's some more work on the scikit-learn wrapper. The doctests are primitive since the string representation of a scikit-learn estimator depends on the scikit-learn version (new versions often introduce extra parameters). The main thing left to implement is `most_informative_features`, which should not be too hard to implement at least for linear estimators.

The module still does not try to import scikit-learn. That would require

```
try:
    import sklearn
except ImportError:  # scikit-learn < 0.8
    import scikit.learn
```

### DIFF
--- a/nltk/classify/__init__.py
+++ b/nltk/classify/__init__.py
@@ -97,6 +97,11 @@ from nltk.classify.util import accuracy, log_likelihood
 # Conditional imports
 
 try:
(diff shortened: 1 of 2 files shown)


stevenbird:
Looks very good, thanks.

PR #187, opened 2012-01-19
Make more string constants module docstrings

kmike:
There are several string constants that looks like module docstrings but are not because of improper placement.


stevenbird:
Excellent, thanks!

PR #189, opened 2012-01-25
Correct class name in docstring.

PiDelport:
`RegexpParser.__str__` refers to itself as `RegexpChunkParser`.


alexrudnick:
Good catch, thanks! :)

PR #192, opened 2012-01-31
Making ConditionalFreqDist a subclass of defaultdict

heatherleaf:
Here are some simplifications to `probability.py`. Most important is that I made `ConditionalFreqdist` a subclass of `defaultdict`. Apart from making it possible to remove some code, it fixes a bug:

Currently, the following gives an infinite loop:

```
for key in cfd: 
    print key, cfd[key]
```

The reason is that since the cfd has `__len__` and `__getitem__` you can loop over it. But Python tries to loop over indexes calls cfd[0], cfd[1], etc. Then `__getitem__` creates a new `FreqDist` for each of these values, and then `__len__` increases, and so on. 

The hard solution is to add a new method, `__iter__`. The simple solution is to make the class a subclass of `defaultdict`. Then we can also remove the definitions of `__getitem__` and `__len__`.

More changes:
- I removed `count`, `sorted` and `sorted_samples` from `FreqDist`, since they have anyway been obsolete for a long time.
- I removed `__eq__` and `__ne__` from `FreqDist` and `ConditionalFreqDist`, since they don't do anything else than what `dict` does. But I kept `__le__`, `__gt__` etc. since they have another semantics than for `dict`.
- I change `__str__` of `FreqDist` so that it only display the 10 most frequent items - otherwise the string can be veeery long.

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -47,6 +47,7 @@ import random
 import warnings
 from operator import itemgetter
 from itertools import imap, islice
+from collections import defaultdict
 
 ##//////////////////////////////////////////////////////
 ##  Frequency Distributions
@@ -160,11 +161,10 @@ class FreqDist(dict):
         """
         return len(self)
 
-    # deprecate this -- use keys() instead?
     def samples(self):
         """
         Return a list of all samples that have been recorded as
-        outcomes by this frequency distribution.  Use ``count()``
+        outcomes by this frequency distribution.  Use ``fd[sample]``
         to determine the count for each sample.
 
         :rtype: list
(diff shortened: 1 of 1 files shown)


heatherleaf:
Note that the bug still remains for `ConditionalProbDistI` (and its subclasses), since it defines `__len__` and `__getitem__`.

Is it ok to make the CPDs subclasses of `defaultdict` too? I can make that change if noone disagrees.

stevenbird:
Thanks.  Please go ahead.  Also, please run the tests in test/probability.doctest.

heatherleaf:
Oh, I forgot to mention:  Since CFD `.conditions()` just returns the dict keys, the resulting list is now unordered. Personally I think this is fine -- one can always sort the list afterwards. But it's a difference from before.

stevenbird:
This effects the book: chapter 2 says cfdist.conditions() produces an alphabetically sorted list of conditions, and has examples of its output.  These would need to become sorted(cfdist.conditions()) everywhere.  I would prefer not to make this change now, but rather in the context of revising the book.  Would you mind adding back the sorting, and possibly creating an issue in our issue tracker to propose that we change the behaviour of cfdist.conditions()?  Thanks.

PR #199, opened 2012-02-01
SENNA Interface for NLTK

aboSamoor:
This module is tested. However, It needs more documentation and polishing. Let me know your comments at this stage.


stevenbird:
Thanks.  Yes, please add sphinx docstrings including some doctests, and the test runner to the bottom of the module (see e.g. hunpos.py).  Note that the NLTK tagger package should load without errors even if the senna tagger is not installed, and if someone tries to use it, they should get a helpful error message.

Issue #6, opened 2011-12-21
Treebank Tokenizer buggy


desilinguist:
The NLTK TreebankWordTokenizer appears to have at least two bugs in it: 

(a) It doesn't properly handle sentences with quotes at the end. 

```
 Input:
``If we don't have any further suggestions as to how we should present our case, we should be able to finish this.''

Output:
['`', '`', 'If', 'we', 'do', "n't", 'have', 'any', 'further', 'suggestions', 'as', 'to', 'how', 'we', 'should', 'present', 'our', 'case', ',', 'we', 'should', 'be', 'able', 'to', 'finish', "this.''"]
```

(b) According to the official Penn Treebank Tokenization page at http://www.cis.upenn.edu/~treebank/tokenization.html, double quotes (") should be changed to doubled single forward- and backward- quotes (`` and ''). However, nltk.word_tokenize() (TreebankTokenizer) does not follow this rule. 

```
Input: 
He went to the "Concert for Hope" yesterday.

Output: 
['He',  'went',  'to',  'the',  '"',  'Concert',  'for',  'Hope',  '"',  'yesterday', '.']
```

One of my colleagues, Michael Heilman, took the official Penn Treebank sed script (http://www.cis.upenn.edu/~treebank/tokenizer.sed) and ported it to NLTK using the Tokenizer API. It can be found here: https://gist.github.com/1506443

This script should replicate the sed script and, therefore, should not have either of these issues.

stevenbird:
In this version, CONTRACTIONS2 requires tokens to be space separated, when \b should be used instead (as in the existing implementation).  Thus, it breaks in cases like this:

"This cannot cannot be right!" --> "This can not cannot be right !"
"This   cannot  work." --> "This    cannot  work ."

stevenbird:
Resolved in
https://github.com/nltk/nltk/commit/e660a6827d44e242b4114abba7fc30fef6b0125c

desilinguist:
Thanks steven!
- Nitin

aboSamoor:
In case of numbers that are written in the financial format like 2,700 the current tokenizer split the number to three tokens. That is actually does not comply with the treebank corpus. Moreover, it does not seem reasonable as it changes the meaning of the text.
Here is the output of my personal benchmark for tokenizers.
https://gist.github.com/1748164

PanderMusubi:
Master at the moment results in

```
>>> nltk.word_tokenize('He went to the "Concert for Hope" yesterday.')
[...]

PR #201, opened 2012-02-01
Complying with the Treebank Tokenization

aboSamoor:
I compared the results of tokenizing the first 1000 sentence in Penn Treebank using this module with the standard tokenization that I could retrieve from the Parse Trees.

I found that the module actually does not comply with the treebank tokenization for around 150 sentence. With these modification the number of differences drop to around 25 sentences.

I suggest that before merging those changes, we should take advantage of the portion of PTB that NLTK distributes. I do not think the raw text is part of that till now. However, I see it is under fair use to distribute the raw text that is corresponding to the first two sections.

To get the first 1000 sentence from the raw text of PTB, I split the paragraphs manually. I can send you the paragraphs splitted into sentences that cover the first two sections. This is in general will be a good dataset for unit testing for both tasks, sentence splitting and work tokenization. Not to mention possible classifiers that could be trained using such dataset.

### DIFF
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -49,6 +49,7 @@ class TreebankWordTokenizer(TokenizerI):
     at the end of a line.
     """
     # List of contractions adapted from Robert MacIntyre's tokenizer.
+
     _CONTRACTIONS2 = [re.compile(r"(?i)(.)('ll|'re|'ve|n't|'s|'m|'d)\b"),
                      re.compile(r"(?i)\b(can)(not)\b"),
                      re.compile(r"(?i)\b(D)('ye)\b"),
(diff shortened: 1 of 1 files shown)


stevenbird:
Regression testing happens in many places, so it is not effective to circulate test data some other way.  Can we use the raw section of the PTB directly for regression testing?

stevenbird:
Your fix cuts across other work that has recently been done on this tokenizer -- https://github.com/nltk/nltk/issues/6
Please re-run any tests and feel free to propose a new patch and test cases, thanks.

aboSamoor:
I looked at the new patch you mentioned above.
The patch I am offering will reduce the errors to 13
The mentioned above patch has 50 errors

I do not consider replacement of characters or symbols as errors. For example, ( is equivalent to LRB.

Issue #198, opened 2012-02-01
Infinite loop when modifying Tree children 


heatherleaf:
Tree:s (and ParentedTree:s) go into an infinite loop if you try to change or delete a child by using a string as key:

```
>>> t = nltk.Tree("(a b c (d e) g f r)")
>>> t["2"]
Tree('d', ['e'])
>>> t["2"] = 42
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/peter/Projekt/NLTK/nltk/nltk/tree.py", line 177, in __setitem__
    self[index[0]] = value
  (...)
  File "/Users/peter/Projekt/NLTK/nltk/nltk/tree.py", line 170, in __setitem__
    if isinstance(index, (int, slice)):
RuntimeError: maximum recursion depth exceeded in __instancecheck__
```

The reason is that neither `__setitem__` nor `__delitem__` tries to convert the key to an integer: `self[index[0]] = value`. `__getitem__` does that on the other hand: `return self[int(index[0])]`

The solution is simple, but there are two possibilities:
1. Convert string keys to integers as is done in `__getitem__`: replace `index[0]` with `int(index[0])` in `__setitem__` and `__delitem__`
2. Disallow string keys: replace `int(index[0])` with `index[0]` in `__getitem__`, and add a check `isinstance(index, (list, tuple))` to all three methods

Personally I prefer the second, since it is what Python does for lists.  I don't see the reason why `tree["123"]`should be allowed, and have the same meaning as `tree[1,2,3]`.  However, although I don't think the current behaviour is documented, the change might break some existing code.

What do the benevolent dictators think?

alexrudnick:
Is it typical to index into a tree using strings? That seems pretty bizarre. I like the second option as well.

stevenbird:
I agree that the existing API is inconsistent and that the type conversion is not useful functionality.  I also prefer the second option.

Issue #196, opened 2012-02-01
FeatStruct.reentrances() returns a list of id's


heatherleaf:
According to the documentation, FeatStruct.reentrances() should return a list of FeatStruct:

```
    Return a list of all feature structures that can be reached
    from ``self`` by multiple feature paths.
    :rtype: list(FeatStruct)
```

However, it returns a list of object id's, which makes the method practically useless. To make it return FeatStruct's, I think ._find_reentrances() need to store the object itself in the reentrances dict. This affects the following FeatStruct methods: .cyclic(), .reentrances(), ._repr() and ._str()

Alternatively, we can change the method doc, but as I said, the method becomes quite useless.

A third possibility is to return the feature paths themselves. But that's more difficult to implement I guess.

stevenbird:
Thanks for spotting this problem.  I think it would be fine to delete this method.

PR #202, opened 2012-02-02
Fixes for issues #196 and #198

heatherleaf:


heatherleaf:
More things:
- fixed a bug in `internals.slice_bounds()`, where the read-only slice object was assigned to
- added the option IGNORE_EXCEPTION_DETAILS (which alread exists in the doctest module) to doctest_driver.py

PR #203, opened 2012-02-02
Allow steps in slices for all kinds of Trees

heatherleaf:
(Multi)ParentedTree does not allow slices with steps:

```
>>> print ptree
(A (U v) (X ) (X ) (X ) (X ) (X ))
>>> ptree[1:-2:2] = ['a', 'b']
  . . .
ValueError: slices with steps are not supported by ParentedTree
```

But Tree does, and I think there is no good reason why this is not allowed for parented trees. So with this change we get the following:

```
>>> ptree[1:-2:2] = ['a', 'b']
>>> print ptree
(A (U v) a (X ) b (X ) (X ))
```

The tree doctest is changed accordingly


stevenbird:
Good catch!

Issue #130, opened 2012-01-17
ProbabilisticTree breaks when you copy.deepcopy it


alexrudnick:
<b>What steps will reproduce the problem?</b>

&gt;&gt;&gt; from nltk.tree import *
&gt;&gt;&gt; tr = ProbabilisticTree(&quot;TOP&quot;, [&quot;N&quot;], prob=0.3)
&gt;&gt;&gt; import copy
&gt;&gt;&gt; copy.deepcopy(tr)
Traceback (most recent call last):
  File &quot;&lt;stdin&gt;&quot;, line 1, in &lt;module&gt;
  File &quot;/usr/lib/python2.6/copy.py&quot;, line 189, in deepcopy
    y = _reconstruct(x, rv, 1, memo)
  File &quot;/usr/lib/python2.6/copy.py&quot;, line 323, in _reconstruct
    y = callable(*args)
  File &quot;/usr/lib/python2.6/copy_reg.py&quot;, line 93, in __newobj__
    return cls.**new**(cls, *args)
TypeError: **new**() takes at least 2 arguments (1 given)
&gt;&gt;&gt; 

<b>What is the expected output? What do you see instead?</b>

Should return a new ProbabilisticTree object, but it throws a TypeError instead!

<b>Please use labels and text to provide additional information.</b>

Will investigate further -- but the **new** class method in ProbabilisticTree takes one more argument than the one in Tree -- could that have something to do with it? Regular Tree objects can be copy.deepcopy'd.<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=613</p>

alexrudnick:
Fixed by commit 6cc57f05d91100f3dfdf57cb9ac221763423db76.

PR #204, opened 2012-02-02
Remove `__new__` from Tree classes

heatherleaf:
I have removed `__new__` from the Tree class and its subclasses.  This makes the code much more readable, and slightly shorter, than before.

I also added some regression tests for immutable trees, and all tests (new and old) pass

### DIFF
--- a/nltk/test/tree.doctest
+++ b/nltk/test/tree.doctest
@@ -96,6 +96,14 @@ type:
     (VP (V enjoyed) (NP my cookie))
     >>> print type(tree)
     <class 'nltk.tree.ImmutableTree'>
+    >>> tree[1] = 'x'
+    Traceback (most recent call last):
+      . . .
+    ValueError: ImmutableTree may not be modified
+    >>> del tree[0]
+    Traceback (most recent call last):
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks.  I'm not sure why **new** was considered necessary.  It would be helpful to check that the parser and chunker tests (that depend on trees) also still pass.

stevenbird:
You should be identified as a contributor to tree.py

alexrudnick:
Ah, great! This fixes issue 130 as well. Thank you, Peter!

heatherleaf:
Alex: I didn't know that, cool! Btw, I think you get an automatic link to the issue if you write #130. And it is closed automatically if you write fixes #130 or closes #130. 

(But perhaps that only works in commits, see the section "Commits+Issues" here:
https://github.com/blog/831-issues-2-0-the-next-generation
)

heatherleaf:
(At least) the following doctests pass: tree, treetransforms, parse, chunk, featstruct, featgram

stevenbird:
Excellent, thanks.

PR #206, opened 2012-02-03
Cleanup, new regression tests and properties in tree.py

heatherleaf:
Here are some mostly minor modifications to tree.py and tree.doctest
- some new regression tests in tree.doctest
- properties and classmethods in @decorator style
- general cleanup in the code to make it more readable


stevenbird:
Thanks for all this cleanup work!

PR #207, opened 2012-02-05
Properties in @decorator style, etc.

heatherleaf:
1. All occurrences of `property()`, `classmethod()` and `staticmethod()` are converted into `@property`, `@classmethod` and `@staticmethod`
2. Most tree properties in `tree.py` are changed to methods. Note that this might break some existing code.
3. The new file tag/senna.py had strange indentation and used new Python 3 syntax


stevenbird:
Thanks, this is excellent.

PR #209, opened 2012-02-07
Test running improvements

kmike:
In this pull request:
- nltk/test/runtests.py script (which can run all nltk tests) is introduced;
- a patched version of nose doctest plugin is bundled;
- old doctest testrunner is removed as well as old bundled coverage.py script; excludes from this script are migrated to runtests.py script;
- IBModel1 tests are skipped by changing `>>>` to `>>` (I had no luck neither with exclude nor with # doctest + SKIP);
- pypy test running is fixed.


stevenbird:
Thanks Mikhail.

PR #210, opened 2012-02-08
Mostly doctest issues

heatherleaf:
I restored the old `doctest_driver.py`, and removed the dependency on `epydoc`. This was easy -- I just copied the `TerminalController` class from `epydoc.util`.

I also made some minor optimizations to `align.IBMModel1`, and updated its doctests. The model is still awfully slow, so the doctests only train on 10 sentences.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
 *.pyo
+.tox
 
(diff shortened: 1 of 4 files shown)


stevenbird:
Thanks.

PR #213, opened 2012-02-09
Backtracking some python 2.5 incompatible changes

heatherleaf:
In this branch I have withdrawn two changes I did which were incompatible with python 2.5:
- 2.5 doesn't support the decorator @xxx.setter on properties
- 2.5 doesn't support collections.MutableSequence

Perhaps it's time to create a new branch for the 2.5 compatible version. I have called this branch py25-compatible, but if you want another name, that's ok with me.

### DIFF
--- a/nltk/align.py
+++ b/nltk/align.py
@@ -58,16 +58,14 @@ class AlignedSent(object):
     def mots(self):
         return self._mots
 
-    @property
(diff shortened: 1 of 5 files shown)


stevenbird:
Thanks Peter!

PR #225, opened 2012-02-19
Positive Naive Bayes Classifier

apresta:
An implementation of a "Positive Naive Bayes Classifier", introduced in

"Text Classification from Positive and Unlabeled Examples"
François Denis, Rémi Gilleron, Marc Tommasi (2002)
http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.8.5291

The source contains a simple explanation. 

An example application is the following: we want to highlight interesting articles for a user by looking at his click history. Clearly, a clicked article represents a positive example. However, an article that hasn't been clicked yet isn't necessarily a negative example. Still, the information that unlabeled examples give, together with positive-labeled ones, can help us build a naive bayes classifier for the task.
The only additional requirement is a prior distribution on the two labels (e.g., an estimate of how likely it is for an article to be interesting).

### DIFF
--- a/nltk/classify/__init__.py
+++ b/nltk/classify/__init__.py
@@ -90,6 +90,7 @@ from nltk.classify.mallet import config_mallet, call_mallet
 from nltk.classify.megam import config_megam, call_megam
 from nltk.classify.weka import WekaClassifier, config_weka
 from nltk.classify.naivebayes import NaiveBayesClassifier
+from nltk.classify.positivenaivebayes import PositiveNaiveBayesClassifier
 from nltk.classify.decisiontree import DecisionTreeClassifier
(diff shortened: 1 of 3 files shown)


xim:
Thank you for the contribution :)

I'll leave this one for when Steven returns - not sure what our policy is on new features before the release, or if we should have running tests before inclusion.

stevenbird:
Thanks.  Would you please add illustrative doctests to the train() docstring, following e.g. nltk/tokenize/*.py.  Optionally put (regression) test code in nltk/test/classify.doctest.

Also, please change the main program (currently demo code), to run the doctests, also following nltk/tokenize/*.py.

apresta:
Here's a doctest. What do you think?

stevenbird:
Thanks.  This is excellent.

PR #230, opened 2012-02-21
nltk.Tree -> Tree

japerk:
fix nltk.Tree calls


alexrudnick:
Thanks for the patch! :) Looks good.

PR #232, opened 2012-02-22
adding generalized hamming distance, and beeferman's Pk as metrics for evaluating text segmentation

DavidDoukhan:
2 metrics have been implemented for evaluating text segmentation.
Both of them can be used using the same arguments than the windowdiff metric
- generalized hamming distance
  python translation of V. Kulyukin C++ code (co-author of the paper on generalized hamming distance)
  the examples in the demo function are the same that were used in
  Kulyukin implementation, and allowed to compare the good behavior
  of this metric to Kulyukin implementation, considered as a reference
- Beeferman's Pk

we had a discussion  with Joel Nothman on nltk-dev
we talked about merging windowdiff, generalized hamming distance, and Pk into a single module, called nltk.metrics.segmentation
I'm new in this project, and it's also the first time i'm using github.
I saw most of the modules components into single py file. I did not found natural to put windowdiff, generalized hamming distance, and other metrics to come into a single file. I would find more natural to create a subdirectory called segmentation, than a file called segmentation.py. Since most of modules are not structured that way, let me know what do you prefer.

The way the contribution is done right now has at least the advantage of backwards-compatibility


DavidDoukhan:
All Joel Nothman's remarks have been taken into account.
coding style has been improved
separated unit tests have been put in nltk.test
a separate sudirectory called segmentation has been done to store the metrics

some suggestions:
- put nltk.metric.windowdiff in this directory as well
- call the directory msegmentation istead of segmentation, in case some one want to latter create modules of segmentation algorithms.

xim:
Thank you both for the work and cleanup =)

Steven should be available for comments (if he has any) very soon. My comments are on the PEP-8 and test side of things:
- Wildcard imports are a no-no
- Replace the <pre>
  >>> float_equal(function_call(), .5)
  True
  </pre> with <pre>
  >>> function_call()
  0.5
  </pre>

DavidDoukhan:
Hi xim,
I got some problems and questions related to your remarks:
- concerning the removal of the float_equal function in tests, it works fine for ghd tests, put for pk, it leads to small problems:
  
  <pre>
  Failed example:
  pk('100100', '000000', 2)
  Expected:
  0.04
  Got:
  0.039999999999999994
  </pre>
  
  I this case, I think it would be better to keep the float_equal statement, unless you got other ideas?
- It's ok for removing wildcard imports in the doctest, and in metrics/segmentation/`__`init`__`.py
  concerning those in metrics/`__`init`__`.py, it depends, there are many possibilities:
- remove the willcard, just letting import segmentation
- remove the willcard, and import manually pk and ghd  from there
- let the willcard, in a similar way than the nltk/`__`init`__`.py

xim:
At 03:45, Mon 2012-02-27, David Doukhan wrote:

> Hi xim,
> I got some problems and questions related to your remarks:

Super! =)

> - concerning the removal of the float_equal function in tests, it works fine for ghd tests, put for pk, it leads to small problems:
>   <pre>
>   Failed example:
>   pk('100100', '000000', 2)
>   Expected:
>   0.04
>   Got:
>   0.039999999999999994
>   </pre>
>   I this case, I think it would be better to keep the float_equal statement, unless you got other ideas?

Try print!

> > > 1./7
> > >    0.14285714285714285
> > > print 1./7
> > >    0.142857142857

If that isn't enough to round it to .04, use e.g round(pk(…), 3)

> - It's ok for removing wildcard imports in the doctest, and in metrics/segmentation/**init**.py
>   concerning those in metrics/**init**.py, it depends, there are many possibilities:
> - remove the willcard, just letting import segmentation
> - remove the willcard, and import manually pk and ghd  from there
> - let the willcard, in a similar way than the nltk/**init**.py

I guess it breaks the standard elsewhere used, but I vote we stop
introducing \* imports. To stick with the style from the others might be
the best – importing pk/ghd and the module.

Thank you!

DavidDoukhan:
ok,
thanks for the print trick and it works well!
concerning import statements, I'm not sure to have clearly understood what you mean, I've donne small changes based on what you told me.

There are still few last points to debate:
- rename the directory to msegmentation, to allow future use of dir name segmentation for segmentation algorithms, and not keeping it for segmentation metrics
- putting windowdiff in this directory
- dividing the result returnes windowdiff by len(seg)-k+1 to match windowdiff initial definition

xim:
Sorry I disappeared from the face of the earth there. On the import stuff, I meant we should keep to the style from metrics/**init**.py, i,e:

```
from nltk.metrics.segmentation import ghd, pk
```

The amount of import statements there are a bit excessive now =)

msegmentation, isn't the fact that this is segmentation metrics covered by the fact that it resides in metrics.segmentation?

DavidDoukhan:
i corrected the import statements

concerning the renaming issue, avoid future clash in another directory is called segmentation.
The way imports are done may lead to the following example: 

<pre>
import nltk
nltk.segmentation (ambiguity if another directory get called segmentation)
</pre>

However, I agree that's not a big issue for now.

What about windowdiff issues?

xim:
Apart from the points you raised on windowdiff, all looks good. I'm not the right person to comment on an algorithm I haven't looked at the details of. Matching the initial definition of the algorithm sounds like a good idea, though.

Issue #229, opened 2012-02-20
ValueError when using ParentedTree in 2.0.1rc4


aecay:
When using NLTK version 2.0.1rc4 (as packaged by Arch Linux), the following code results in an error:

```
>>> nltk.tree.ParentedTree("(FOO (BAR baz) (QUUX blorfle))")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/nltk/tree.py", line 1035, in __init__
    super(ParentedTree, self).__init__(node_or_str, children)
  File "/usr/lib/python2.7/site-packages/nltk/tree.py", line 821, in __init__
    self._setparent(child, i, dry_run=True)
  File "/usr/lib/python2.7/site-packages/nltk/tree.py", line 1115, in _setparent
    raise ValueError('Can not insert a subtree that already '
ValueError: Can not insert a subtree that already has a parent.
```

The same code does not give an error with version 2.0.1-rc2 (the next-most-recent version available as a binary package in Arch:

```
>>> nltk.tree.ParentedTree("(FOO (BAR baz) (QUUX blorfle))")
ParentedTree('FOO', [ParentedTree('BAR', ['baz']), ParentedTree('QUUX', ['blorfle'])])
```

xim:
Introduced in 8284903e2a80925fb681fc2176914193c32cbce6 – having a look now!

xim:
Disregard attempted fix. I'll leave this for someone that understands what the hell happens under the hood in tree.py.

aecay:
As of 2.0.2, this seems to have been resolved -- the example code above no longer raises an error.  I don't really know what change is responsible for the fix, though...only that it now works again.

stevenbird:
Thanks, closing this now.

aecay:
On further investigation, 200eea8 in combination with 1309882 is actually an anti-fix, since `_setparent` is now never called.  So now in 2.0.2, the following code returns `None`, bogusly.

```
t = nltk.tree.ParentedTree("(FOO (BAR baz))")
t[0]._parent
```

heatherleaf:
I think I finally solved it. Please have a look at pull request 292:

 https://github.com/nltk/nltk/pull/292

aecay:
[...]

PR #233, opened 2012-02-26
Fix for issue #229

xim:
This should really be reviewed by someone that actually understands what happens in tree.py. :)


xim:
/ignore. I broke the world.

PR #234, opened 2012-02-28
cosmetics: get rid of CVS-style $Id$ lines

slattarini:
The CVS-style $Id$ lines are pretty much useless with git; and
anyway, using them would require adding proper "ident" directives
to the .gitattributes file -- and we don't have such directives.


stevenbird:
Thanks for this.

stevenbird:
I've added your name to our list of contributors, which will be migrated to github sometime:
http://code.google.com/p/nltk/wiki/People

PR #235, opened 2012-02-29
Enhancements for ngram model.

dan-blanchard:
- Added padding to end of sentence as well as beginning so that we take into account the probability of words ending a sentence.
- Modified constructor to take lists of string lists for training in addition to just a list of strings so that we can handle training text that is both sentence and word tokenized.
- When give a list of string lists for training, the model will no longer just generate sentences that start with the same word.

### DIFF
--- a/nltk/model/ngram.py
+++ b/nltk/model/ngram.py
@@ -30,7 +30,8 @@ class NgramModel(ModelI):
     """
 
     # add cutoff
-    def __init__(self, n, train, estimator=None, *estimator_args, **estimator_kw_args):
+    def __init__(self, n, train, estimator=None, pad_initial=True, pad_final=False, 
+                 *estimator_args, **estimator_kw_args):
         """
         Creates an ngram language model to capture patterns in n consecutive
         words of training text.  An estimator smooths the probabilities derived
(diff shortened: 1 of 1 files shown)


desilinguist:
Can we merge this pull request? It seems helpful.

desilinguist:
I would like to merge it myself but I would prefer a second set of eyes since I work with Dan and want to avoid any conflict of interest.

stevenbird:
I like the modification to the training argument, so that it can be either a list or a list of lists.  However, I don't think the addition of both left and right padding is correct.  Is there a reference for this?  Perhaps we want left_padding and right_padding independently, where the defaults are left_padding=True and right_padding=False.

dan-blanchard:
I don't have a reference for adding padding to the right hand side too, but most language modeling toolkits (e.g., SRILM, and IRSTLM) add beginning and end of sentences markers during model training and when getting the probability of a sentence. This ensures that sentences that end with common words like determiners that can't end sentences don't receive high scores.  For example, we wouldn't want "I saw the" be very probable.

stevenbird:
I see.  But the proposed change is not backwards compatible.  It would be the appropriate change to make if the existing behavior is erroneous.  In the SRILM implementation, initial and final padding can be controlled independently (see the -no-sos and -no-eos flags http://www.speech.sri.com/projects/srilm/manpages/ngram-count.1.html).  Would you mind proposing a patch that permits independent control of initial and final padding please?

dan-blanchard:
In my opinion the previous behavior is erroneous, but it would also be nice to have options similar to the -no-sos and -no-eos flags from SRILM to prevent the padding if people don't want it for some reason. Would adding those options be sufficient, or would you prefer a setup where by default initial padding was used, but not final padding?

I would just implement them as default parameters called initial_padding and final_padding, so the only difference would be whether the default value of final_padding would be true or false.

dan-blanchard:
I've added those options, and set the defaults as you requested.

xim:
This change changes the NgramModel in a way that would make the existing call NgramModel(n, train, estimator=None, *estimator_args) pass the two first arguments (if any) meant for *estimator_args be passed into pad_initial=True and pad_final=False ...

Agree that the change is something we want, though. Do we care about the minor API breakage?

dan-blanchard:
Seeing as `*estimator_args` was only added at the end of November in my previous set of changes, I can't imagine the API change would be that big of a deal.

stevenbird:
Sorry, now that 2.0.1 is out, changes need to be backwards compatible if possible.  Is there some way to make this new functionality backwards compatible, or to detect incorrect usage and raise an error?

dan-blanchard:
I'm not aware of anyway to rearrange the arguments to prevent the issue @xim was talking about from happening. If being backwards compatible is paramount, I could always create a subclass of `NgramModel` that allowed you to pass in `pad_initial` and `pad_final` to its `__init__`, but that seems kind of clunky to me.

xim:
We can alternately process the kwargs variable and see if it contains those specific arguments. I'm (marginally) in favour of breaking the API rather than doing that, seeing as `*estimator_args` and `**estimator_kwargs` are relatively recent additions.

stevenbird:
Resolved in https://github.com/nltk/nltk/commit/258449703baf796fa973f0911c54b913103eb4d8

PR #237, opened 2012-02-29
prefer NotImplementedError over AssertionError in few places

slattarini:
The python documentation on "Built-in Exceptions":

  http://docs.python.org/library/exceptions.html

reads:

  ``In user defined base classes, abstract methods should raise
    this exception when they require derived classes to override
    the method.''

But existing NLTK code didn't consistently follow that best practice,
sometimes using 'AssertionError' or 'assert' in similar situations.
Fix that.  Also, make error messages less verbose, for consistency
with other existing code.


stevenbird:
Thanks.  What is the basis for reporting: "Abstract classes can't be instantiated" in some cases?

slattarini:
On 02/29/2012 11:39 PM, Steven Bird wrote:

> Thanks.
> 
> Thanks to you for your quick feeback.
> 
> What is the basis for reporting: "Abstract classes can't be
> instantiated" in some cases?
> 
> I'm not sure what you are exactly asking here, sorry ...  But I see
> two (and a half) possible interpretations for you question:

 1) Why I have converted exceptions about erroneous instantiation of
    abstract classes from AssertionError to NotImplementedError:

```
Basically for consistency with diagnostic about similar issues;
also, the concrete reason one shouldn't instantiate or use an
abstract class is because it only offers an interface or a partial
implementation, so raising a NotImplementedError seems appropriate.
```

 2) Why pass the explicit error message "Abstract classes can't be
    instantiated" to NotImplementedError invocation, instead of no
    specific message as done in other cases:

```
First, that error message was already present; moreover, I find it
quite helpful, since it makes clear that we are not failing because
the __init__ method is expected to always be overridden by subclasses
(which is not the cause, but which a bare 'NotImplementedError' would
let one assume), but only because the base class is not supposed to
be instantiated directly.
```

 3) Why some abstract classes need to explicitly punt and complain
    when instantiated:

```
I guess that's done to easily catch improper uses of them (but I
surmise you knew that better than me, so it's almost certain this
wasn't the sense of your question);
```

Hope this answers your question in a satisfactory way.

Regards,
  Stefano

PR #241, opened 2012-03-16
Added flag to ignore case when building a dispersion_plot.

nickgarvey:
This involved adding the flag to nltk.text.Text.dispersion_plot and nltk.draw.dispersion_plot.


stevenbird:
Good idea, thanks.

PR #244, opened 2012-03-27
Fixed "global name 'fileid' is not defined" error when calling CorpusReader.abspath with include_fileid=True 

dan-blanchard:
Someone wrote "fileid" instead of "fileids", so I fixed that.


alexrudnick:
Thanks, Dan! Good catch! :)

PR #245, opened 2012-03-28
Added LinThesaurusCorpusReader for handling Lin's Thesaurus

dan-blanchard:
I think this won't quite work until the thesaurus data is added to nltk_data, because I used `nltk.data.find` as part of the demo function.

I only modified `nltk/corpus/reader/__init__.py`, because I wasn't quite sure how to setup the `LazyCorpusReader` in `nltk/corpus/__init__.py`.


stevenbird:
Thanks for this!  I'll test it out over the weekend.

stevenbird:
Sorry for the delay...

stevenbird:
The data file needed to have a different structure (pushed down into a subdirectory), and the naming made consistent (lin_thesaurus vs lin_dependency both used).  This also forced me to modify nltk to use the github location instead of the googlecode one.  Hence the delays...

PR #246, opened 2012-04-02
optimize scikit-learn wrapper; example runs at 4x original speed

larsmans:
I've switched to `scipy.sparse.coo_matrix`, which gave similar speedups when implementing this functionality directly in scikit-learn.


stevenbird:
nice work!

PR #247, opened 2012-04-08
WordNetLemmatizer docs + code style

larsmans:
I updated the WordNetLemmatizer docstring, which I found confusing, and took the liberty of rewriting the code in a more Pythonic style while I was at it.


alexrudnick:
Very nice. Thanks, Lars! :)

Issue #255, opened 2012-05-07
util.breadth_first returns incorrect result with depth limit set


larsmans:
```
>>> from nltk.util import breadth_first
>>> class Node(object):
...     def __init__(self, label, children=()):
...         self.label = label
...         self.children = children
...         
>>> tree = Node(1, [Node(2, [Node(3)]),
...                 Node(4, [Node(5)])])
>>> [x.label for x in breadth_first(tree, lambda x: x.children, depth=2)]
[1, 2, 4, 3]
```

while it should be `[1, 2, 4, 3, 5]`, i.e. the complete tree. The problem is that the depth limit is handled incorrectly.

PR #256, opened 2012-05-07
Rewrote BFS for correctness and speed

larsmans:
Solves issue #255. In addition:
- renamed the `depth` argument `maxdepth`; it wasn't used as a keyword argument, so this is harmless
- removed the `queue` argument, which wasn't being used in client code
- replaced the `list`-based queue (O(_n_) pop) with a `deque`-based one (O(1) pop)
- replaced a catch-all `except` clause for safety.


larsmans:
Alright, I think this is really for pulling.

stevenbird:
Looks good, thanks.

PR #257, opened 2012-05-15
Score in LinThesaurusCorpusReader is now stored as float instead of string.

dan-blanchard:
I've fixed a bug where the score in the LinThesaurusCorpusReader was being stored as a string instead of a float, and updated my email address to my work address, since I developed this on ETS time (and contributed it with their permission).


xim:
Sorry for the delay =)

Issue #147, opened 2012-01-17
MaltEval in nltk.parse.malt?


alexrudnick:
Adding an interface to MaltEval (http://w3.msi.vxu.se/~jni/malteval/) to nltk.parse.malt would be absolutely wonderful.  If need be, I could provide some Polish data to help test such an interface.

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=656</p>

alvations:
Closing this because it seems like `MaltEval` is an orphaned project and it [cannot be found even on wayback machine](http://web.archive.org/web/*/http://w3.msi.vxu.se/users/jni/malteval). I would love to visualize it dependencies and see how `MaltEval` works too but now I think there're [`brat`](http://brat.nlplab.org/) that can do the same job #668  =(

PR #260, opened 2012-05-26
Issue #147 (MaltEval in nltk.parse.malt?) addressed

kacperChwialkowski:
Hi, 

I have created basic MaltEval interface implementation. I would appreciate very much if you could review my code and give me some feedback.There are also some issues that need to be discussed

1) While .docstest are really nice feature, they do not allow me to test all my code. Consider situation in which we want test suite to run in environment where MaltEval.jar might be missing. In unit test we can write

```
 @unittest.skipIf(Malteval.can_configure_properly()==False,"MaltEval.jar could not be found automatically ")
def test_run_default_cmd(self):
     .....
```

where static method Malteval.can_configure_properly() checks if we can find MaltEval.jar automatically. If we can't, test will be skipped and reason will be logged. This can not be achieved in .doctest so we will get error when in fact nothing bad has happened.

2) I have prepared class with unit tests  for this interface, but i am not sure if you use unit tests. Should I put it to tests directory or somewhere else?

Finally I would like to emphasize that code that i have written is basic and easy. I will push further changes if we all will agree that this basic code fits well nltk structure. I am looking forward to hearing from you. 

Kind regards,

Kacper  Chwialkowski


kmike:
Hi Kacper,

In the '2and3' branch (https://github.com/kmike/nltk) I gave up rewriting some doctests to be compatible with python 2 and python 3 and introduced a bit of unit tests (https://github.com/kmike/nltk/tree/2and3/nltk/test/unit). How do you think, is it a good convention?

kacperChwialkowski:
Hi Kmike,

 I am not a master of TDD in python ( btw it sounds strange), but it looks good for me. Should 
I clone '2and3' branch (https://github.com/kmike/nltk) add push my unit tests, .doctest, and MaltEval class there ?

Kind regards

Kacper

kmike:
I don't know :) 

I think all changes from '2and3' branch are backwards-compatible (modulo python 2.5 support), there are some bugfixes in this branch that are not related to python 3, and thus this branch can be merged into master as-is and it is good to base further development on it but I may be missing something.

There are also some gotchas in writing python 2+3 compatible code that nltk contributors should take care of so it is a bit harder to contribute to '2and3' branch (but I assure that's not a lot harder). For example, it seems that in order to make your pull request python3 compatible you'll have to add `from functools import reduce` to the imports block and maybe (I'm not sure) use codecs.open instead of open in _to_file (this is good even for python 2 in order to explicitly support unicode) - and that's all.

By the way, code can be improved a bit: try-finally should be better rewritten as with statement and there are some pep8 issues (spaces after commas, indentation of config_mateval). 

Alternatively, you could still base you pull request on a master branch and just follow the convention.

E.g. if you add 'nltk/tests/unit/test_malteval.py' file with your unittests then you should be able to run only this test under python 2.7 using

```
tox -e py27 -- unit.test_malteval
```

Let's summon @stevenbird and @xim to decide :)

kmike:
[offtop] well, some changes are now backwards-incompatible: corpus readers returns unicode, not bytestrings.

kacperChwialkowski:
Hi,

Firstly, thank you for your comments on my code. I have just read pep8 and I will make my best to write  code according to described conventions. I will also reactor already created code to meet those standards. 
Secondly I think I'll clone (fork) your repo '2and3' and pull my changes and unit test there. Its a truism that its better 
to use unit tests then not too .
Last but not least, I don't know how to summon @stevenbird nor @xim. 

I appreciate your help and feedback very much (seriously),

Kacper

xim:
I'll have to get around to this somewhere in June, as I'm currently doing exams, packing and moving!

Warm smiles,
xim

xim:
Aside from the issue of running tests, which I'll get around to later, I'd be thrilled to see some more PEP-8 compatible variable/method names. You've mixed names such as test_create_gold_file(), toprettyxml(), addValue() and Feats(). We try to standardize on names that conform with http://www.python.org/dev/peps/pep-0008/#naming-conventions

Summer smiles,
xim

stevenbird:
Sorry for the delay in giving feedback.  Thanks for the contribution. It would be really great to include an interface to MaltEval.  However, I don't think this solution is quite ready because it doesn't fit well enough with NLTK's organisation and conventions.

1) the nltk.metrics package is for generic metrics; evaluation code that is specific to a particular task (like tagging, parsing) belongs in the corresponding package (NB: "grep eval nltk/_/_.py" to see what I mean).

2) the modules maltevalconstans and maltevalxmlcommadns are not pep8 compliant and contain spelling mistakes in their names; breaking the contribution into these three modules clutters the package and is not consistent with practice elsewhere in NLTK; the use of the _MaltevalParameters class to define a set of string constants is similarly anomalous.

3) the process for finding the malteval binary should be more consistent with the process used by other modules that interface to external binaries (e.g. this uses "lookForJar", and btw we don't use mixedCase names in NLTK).

I hope you are able to address these issues.  The nltk-dev mailing list would be the best place to discuss them further, thanks.

Issue #262, opened 2012-06-03
nltk.grammar.demo() produces AttributeError


oubiwann:
NLTK version: 2.0.1
NLTK corpora installed: all
Python version: 2.7.3
OS: Mac 10.6.8

When running the demo, an AttributeError is raised. Before the error is raised, some output is produced:
- nonterminals
- a grammar with 14 productions
- a PCFG production and grammar

When the demo for inducing a PCFG grammar is attempted, the following traceback is produced:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/nltk/grammar.py", line 1470, in demo
    pcfg_demo()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/nltk/grammar.py", line 1401, in pcfg_demo
    for item in treebank.items[:2]:
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/nltk/corpus/util.py", line 71, in **getattr**
    return getattr(self, attr)
AttributeError: 'BracketParseCorpusReader' object has no attribute 'items'

oubiwann:
Looks like this didn't get closed when the pull request was merged.

Closing now -- thanks Peter Ljunglöf!

PR #264, opened 2012-06-03
Reworked the demo to not depend upon the items attr.

oubiwann:
By referencing 'items' on the treebank object, an AttributeError was getting
raised.

Fixes issue #262.

Caveat: I don't know if this is expected output of the demo or not, since I don't have something to compare this to.


stevenbird:
Thanks.  NB these demonstrations will ultimately be reworked as docstrings, at which time the expected output will be documented properly.

heatherleaf:
shouldn't issue #262 be closed too?

PR #269, opened 2012-06-08
Define TclError in try block

hlongmore:
TclError was not defined in the try block when importing from Tkinter, so that when a TclError occurred in the interactive download, it did not run the DownloaderShell; instead, the exception went uncaught.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -167,6 +167,7 @@ except:
 try:
     TKINTER = True
     from Tkinter import Tk, Frame, Label, Entry, Button, Canvas, Menu, IntVar
+    from Tkinter import TclError
     from tkMessageBox import showerror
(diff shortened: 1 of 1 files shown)


xim:
Any reason why you didn't just append it to the long list of `from Tkinter import Tk, [...]` ?

hlongmore:
I get a 404 when accessing http://nltk.org/contribute, so not knowing your 
preferred code style, I stuck with PEP 8, keeping the line to less than 80 
chars.  Appending it to the long list is fine, too.

Today at 04:50, Morten M. Neergaard said:

> Any reason why you didn't just append it to the long list of `from Tkinter import Tk, [...]` ?

stevenbird:
Resolved by adding TclError to existing list of Tkinter imports (wrapping for PEP-8 compliance)

PR #271, opened 2012-06-11
make code compatible to NLTK book example.

rsennrich:
A code example in the NLTK book no longer works due to the removal of wildcard imports.
can be fixed either here or in the book (https://github.com/nltk/nltk_book/pull/1).

Of course, only one pull request needs to be accepted (although nothing will break if you accept both).


stevenbird:
Thanks; we'll update the toolkit to support the examples that are published in the book.

PR #277, opened 2012-06-18
fixing lemmas() method in wordnet

desilinguist:
I had a use case for enumerating all the wordnet lemmas and found that proper noun lemmas like 'english', 'aramaic', 'russia' were enumerated via the all_lemma_names() method but did not return any output when used as the input to the 'lemmas()' method. The reason is that in the lemmas() method, the name matching is not done in a case-insensitive fashion. This doesn't make sense given that the 'synsets()' method lowercases everything.


desilinguist:
I could commit this myself but I wanted to make sure that this is actually a bug and not a "feature".

stevenbird:
Good idea, thanks.

PR #278, opened 2012-06-18
Support symlinking to a full installation of the Penn Treebank via the ptb module (and companion modules propbank_ptb and nombank_ptb).

nschneid:
Note that this change:
- Allows for retrieval of corpora symlinked from the `nltk_data` directory
- Introduces the class `CategorizedBracketParseCorpusReader`, which is used for the Penn Treebank to allow selection of categories such as `news` (WSJ) or `fiction`, `humor`, `romance`, etc. (Brown).


stevenbird:
Thanks.  It would be good to illustrate / test this in test/corpus.doctest.

PR #279, opened 2012-06-20
Documenting the ptb module (and cousins, propbank_ptb and nombank_ptb)

nschneid:
Here is an attempt to document them in `test/corpus.doctest`. I realized that in order to share the necessary `allcats.txt` file there should be a dedicated `ptb3` package (always stored as a directory rather than a zip file so users can add/symlink the data directories). How should this be accomplished? I'm not familiar with the inner workings of the downloader tool.

### DIFF
--- a/nltk/test/corpus.doctest
+++ b/nltk/test/corpus.doctest
@@ -293,12 +293,10 @@ Reading the Penn Treebank (Wall Street Journal sample):
       (. .))
 
 If you have access to a full installation of the Penn Treebank, NLTK 
-can be configured to load it as well. In your ``nltk_data`` directory, 
-create a directory called ``ptb3`` in which the Treebank files will 
-be stored (alternatively, ``ptb3`` can be a symlink to an existing 
-installation). The ``ptb3`` directory should contain the ``BROWN`` and 
-``WSJ`` subdirectories, as well as ``allcats.txt``. Then use the ``ptb`` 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks.  I'll create the ptb3 package as suggested.  Where is the allcats.txt file?

nschneid:
`allcats.txt`: https://gist.github.com/2989391

stevenbird:
I've created the ptb3 package in the gh-pages branch of the nltk_data repository.
You can get it using: sudo python -m nltk.downloader ptb3
Please check that it looks ok, then update the documentation, thanks.

nschneid:
Works for me. I think the documentation now matches the expected behavior.

stevenbird:
I see.  But now there's an inconsistency: corpora/ptb3 is loaded by nltk.corpus.ptb.  I'd prefer to call both just ptb (or both ptb3), following our practice with most other corpus readers.  If there's no objections I'll make this change.

nschneid:
OK, I guess I inferred from the `nombank.1.0` package that version numbers were often included. Either way is fine with me.

stevenbird:
True, though in this case I think it will cause less confusion if nltk_data and nltk both use the "ptb" name.  I've made this change now, and updated corpus/**init**.py and test/corpus.doctest accordingly.

PR #284, opened 2012-07-10
Bugfix for agreement.py removing dependence on ordered coder names.

cfedermann:
Fixed bug in `agr(self, cA, cB, i, data=None)` where a wrong order of coder/item combinations in data could cause a result of `0.0`, which affected results from `kappa()`, `pi()`, and `S()`.

Updated the `doctest` accordingly:

```
This would have returned a wrong value (0.0) in @785fb79 as coders are in
the wrong order. Subsequently, all values for pi(), S(), and kappa() would
have been wrong as they are computed with avg_Ao().
>>> t2 = AnnotationTask(data=[('b','1','stat'),('a','1','stat')])
>>> t2.avg_Ao()
1.0

The following, of course, also works.
>>> t3 = AnnotationTask(data=[('a','1','othr'),('b','1','othr')])
>>> t3.avg_Ao()
1.0
```

Please merge; feel free to adapt/change my code as you seem fit...

### DIFF
--- a/nltk/metrics/agreement.py
+++ b/nltk/metrics/agreement.py
@@ -27,10 +27,13 @@ functions that take two arguments, and return a value between 0.0 and 1.0
 indicating the distance between them.  If not supplied, the default is binary
 comparison between the arguments.
 
-The simplest way to initialize an AnnotationTask is with a list of equal-length
-lists, each containing a coder's assignments for all objects in the task:
+The simplest way to initialize an AnnotationTask is with a list of triples,
+each containing a coder's assignment for one object in the task:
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks Christian!  I'm asking Tom Lippincott for feedback before
merging.  Also, would you mind using 4-space indentation (following
PEP-8) please?

On 10 July 2012 17:55, Christian Federmann
reply@reply.github.com
wrote:

cfedermann:
Damn it -- it's too early: sorry about the 2-space indentation, will fix it asap ;)

PR #290, opened 2012-09-05
Make MaltParser safe to use in parallel

pflaquerre:
These commits make MaltParser safe(r) to use in parallel. The paths to temporary files used to be hard-coded, which would be problematic if you're running multiple instances at the same time. There is still a potential issue if the user leaves mco to the default value, but I wasn't sure whether the user would expect the model to still be there after nltk quits or not in this case.

Changes:
- Actually use self.working_dir
- Replace hard-coded temporary file names with tempfile.NamedTemporaryFile
- Use subprocess instead of os.system to avoid quoting issues when calling java
- Error out if subprocess fails instead of blindly going forward

### DIFF
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -6,9 +6,6 @@
 # URL: <http://www.nltk.org/>
 # For license information, see LICENSE.TXT
 
-# python2.5 compatibility
-from __future__ import with_statement
-
 import os
 import tempfile
 import glob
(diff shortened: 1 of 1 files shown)


heatherleaf:
Good things, but I don't like this:

```
def __init__(self, tagger=None, mco=None, working_dir=tempfile.gettempdir()):
    self.working_dir = working_dir
```

I think the old code is better:

```
def __init__(self, tagger=None, mco=None, working_dir=None):
    self.working_dir = tempfile.gettempdir() if working_dir is None else working_dir
```

The reason is that with your changes, the working_dir will be resolved when the module is imported. With the old code, working_dir is resolved when the MaltParser object is created. 

A minor issue, I agree.

pflaquerre:
Thanks for the comments. I've reverted that part of the code back to what it used to be. I've also updated the mco parameter documentation to better reflect what MaltParser actually does with it.

stevenbird:
Thanks for this.

PR #291, opened 2012-09-05
Plug DependencyGraph file descriptor leak

pflaquerre:
This is not a major leak, since the garbage collector will destroy the file object relatively soon, but it might still be problematic if called in a tight loop.

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -15,6 +15,9 @@ The input is assumed to be in Malt-TAB format
 Currently only reads the first tree in a file.
 """
 
+# python2.5 compatibility
+from __future__ import with_statement
+
 from nltk.tree import Tree
 from pprint import pformat
 import re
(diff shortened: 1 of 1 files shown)


heatherleaf:
IIRC, we dropped 2.5 compatibility when we released v2.0, so you don't need the `from __future__` statement

pflaquerre:
Ah. I assumed you still supported 2.5 given that http://nltk.org/install.html mentions "NLTK requires Python versions 2.5-2.7", and a few files import it (nltk/**init**.py, nltk/tag/hunpos.py, nltk/corpus/reader/lin.py and setup.py). Should I still remove it?

alexrudnick:
I think it's fine to keep it -- we can remove all the `from __future__` imports later, in one go...

alexrudnick:
Peter, would mind terribly if I go ahead and merge this?

heatherleaf:
Not at all, go ahead.

alexrudnick:
Merged! Thanks, pflaquerre!

stevenbird:
Note that we're still supporting Python 2.5. The port to Python 3 will drop support for 2.5.

PR #292, opened 2012-09-07
Fix to #229

### LINKED ISSUE #229: ValueError when using ParentedTree in 2.0.1rc4

heatherleaf:
I hope this finally fixes #229. 

I had to do some minor hacks in `ParentedTree.__init__` and `MultiParentedTree.__init__`, for fixing broken parent pointers. 

Maybe it was a bad idea to remove `__new__` in 8284903. But on the other hand, that code was not very nice either...

### DIFF
--- a/nltk/test/tree.doctest
+++ b/nltk/test/tree.doctest
@@ -697,13 +697,13 @@ operations:
     >>> ptree[0,0].remove(make_ptree('(Q p)'))
     Traceback (most recent call last):
       . . .
-    ValueError: list.index(x): x not in list
+    ValueError: ParentedTree('Q', ['p']) is not in list
     >>> ptree.remove('h'); pcheck(ptree)
     ok! (A (B (C (D )) g))
     >>> ptree.remove('h');
     Traceback (most recent call last):
       . . .
-    ValueError: list.index(x): x not in list
+    ValueError: 'h' is not in list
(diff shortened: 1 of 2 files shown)


kmike:
Python3 porting was way easier without `__new__`; working Tree without metaclasses is a plus.

PR #300, opened 2012-10-04
typo in tableau.py

badri:
The if condition in 163 needs to be removed. It prevents NLTK from being installed and used.


stevenbird:
Sorry, my fault. Thanks for the fix.

PR #305, opened 2012-11-07
Improvement in the _substitute method in chat/util.py

jkarni:
Hi,

  I'm hoping to contribute a bit to the chat bots, but as a quick first patch, I've changed the _substitute function so that now it can use multiword reflections. It is now a "greedy" (longest non-overlapping matches between str and reflections.keys()) regex-based substitution.
  I've also change a couple of the keys and values in the reflections dict, now that multi-word matches are allowed.
  Let me know if I'm missing something in the procedure.

Cheers,
  Julian
jkarni - at - gmail - dotcom

### DIFF
--- a/nltk/chat/util.py
+++ b/nltk/chat/util.py
@@ -14,20 +14,22 @@ import re
 import random
 
 reflections = {
-  "am"     : "are",
-  "was"    : "were",
-  "i"      : "you",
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for the contribution. We'll acknowledge you in the next release.

PR #308, opened 2012-11-14
Patch to fix same errors in unit tests

klangner:
from **future** import print_function
Should be the first line of the code. Even before comments.
Otherwise there are errors when running tests


klangner:
This version passes all tests.
Samo of the problematic tests were disabled. They should be fixed in the next step

kmike:
I'm not merging this pull request because it seems that there is a better way to (conditionally) skip doctests (using nose fixtures); I'm also going to rewrite some doctests to unittests soon (so skipping them using #doctest: +SKIP won't be required). But your code would be helpful to identify places where the fixes are necessary, so thanks for this (and sorry for the long feedback delay).

kmike:
We've fixed print_function issue in master branch; noseids issue is fixed in 2and3 branch; tests with unmet requirements are skipped in 2and3 branch (failing tests are not skipped because they indicate failures); extract_test_sentences issue is also fixed in a 2and3 branch (`extract_test_sentences.__test__ = False` attribute is added so that nose doesn't run this function as test, and nltk.parse exports are unchanged). 

So I'm closing this pull request now. Unfortunately no code was merged, and I understand that this may be a bad contributing experience, but I hope this won't discourage you from contributing to NLTK in future. Testing suite is in flux now, and some work is going in a 2and3 branch without backporting to master; I hope in future (after merging) the contributing experience would be better.

PR #315, opened 2012-12-04
made parse.chart deterministic

kmike:
nltk.parse.chart.Chart.parses is currently nondeterministic.

This causes random test failures in ccg.doctest (nltk.ccg.chart.CCGChartParser.nbest_parse may produce different output with the same input).

If I understand the issue correctly, Chart.child_pointer_lists returns dict keys:

```
return self._edge_to_cpls.get(edge, {}).keys()
```

The result is then used in Chart._trees method (which is called by Chart.trees method which is called by Chart.parses method). But dictionaries are unordered in Python and Chart.child_pointer_lists may return different results; this causes nondeterminism for other methods.

So in this pull request the result of Chart.child_pointer_lists is made deterministic (by using nltk.util.OrderedDict instead of dict). This affects parsing results (so tests are updated); ccg.doctest is always passes for me now. 

parse.doctest becomes broken and it is not completely fixed with my commit:

```
FAIL: Doctest: parse.doctest
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/doctest.py", line 2166, in runTest
AssertionError: Failed doctest test for parse.doctest
  File "/Users/kmike/svn/nltk/nltk/test/parse.doctest", line 0

File "/Users/kmike/svn/nltk/nltk/test/parse.doctest", line 834, in parse.doctest
Failed example:
    unittest(isawjohn, "I saw John with a dog with my cookie", 5)
Expected nothing
Got:
    Trees differ for parser: FeatureTopDownChartParser

```

Trees differ because the order of results become different for different parsers after this patch. I think they were the same by accident before, and the test should be fixed. 

But I'm not expert in nltk.parse, I haven't used it myself, etc.; it'll be great if somebody with more knowledge will review this pull request because it affects parsing results.

### DIFF
--- a/nltk/parse/chart.py
+++ b/nltk/parse/chart.py
@@ -43,6 +43,7 @@ import warnings
 
 from nltk.tree import Tree
 from nltk.grammar import WeightedGrammar, is_nonterminal, is_terminal
+from nltk.util import OrderedDict
 
 from nltk.parse.api import ParserI
 
@@ -589,7 +590,7 @@ class Chart(object):
             self._register_with_indexes(edge)
 
         # Get the set of child pointer lists for this edge.
-        cpls = self._edge_to_cpls.setdefault(edge,{})
+        cpls = self._edge_to_cpls.setdefault(edge, OrderedDict())
         chart_was_modified = False
         for child_pointer_list in child_pointer_lists:
             child_pointer_list = tuple(child_pointer_list)
--- a/nltk/test/ccg.doctest
+++ b/nltk/test/ccg.doctest
@@ -17,87 +17,84 @@ Construct a lexicon:
 
     >>> lex = lexicon.parseLexicon('''
     ...     :- S, NP, N, VP
-    ... 
+    ...
     ...     Det :: NP/N
     ...     Pro :: NP
     ...     Modal :: S\\NP/VP
-    ... 
+    ...
     ...     TV :: VP/NP
     ...     DTV :: TV/NP
-    ...     
+    ...
     ...     the => Det
-    ...     
+    ...
     ...     that => Det
     ...     that => NP
-    ... 
+    ...
     ...     I => Pro
     ...     you => Pro
     ...     we => Pro
-    ... 
+    ...
     ...     chef => N
     ...     cake => N
     ...     children => N
     ...     dough => N
-    ... 
+    ...
     ...     will => Modal
     ...     should => Modal
     ...     might => Modal
     ...     must => Modal
-    ... 
+    ...
     ...     and => var\\.,var/.,var
-    ...     
+    ...
     ...     to => VP[to]/VP
-    ...     
+    ...
     ...     without => (VP\\VP)/VP[ing]
-    ... 
+    ...
     ...     be => TV
     ...     cook => TV
     ...     eat => TV
-    ... 
+    ...
     ...     cooking => VP[ing]/NP
-    ... 
+    ...
     ...     give => DTV
-    ... 
+    ...
     ...     is => (S\\NP)/NP
     ...     prefer => (S\\NP)/NP
-    ... 
+    ...
     ...     which => (N\\N)/(S/NP)
-    ... 
+    ...
     ...     persuade => (VP/VP[to])/NP
     ...     ''')
     >>> parser = chart.CCGChartParser(lex, chart.DefaultRuleSet)
     >>> for parse in parser.nbest_parse("you prefer that cake".split(),1):
     ...    chart.printCCGDerivation(parse)
     ...    for parse in parser.nbest_parse("that is the cake which you prefer".split(), 1):
-    ...        chart.printCCGDerivation(parse) # doctest: +NORMALIZE_WHITESPACE
+    ...        chart.printCCGDerivation(parse)
     ...
-     you    prefer      that   cake 
-     NP   ((S\NP)/NP)  (NP/N)   N   
-         --------------------->B
-              ((S\NP)/N)
+     you    prefer      that   cake
+     NP   ((S\NP)/NP)  (NP/N)   N
+                      -------------->
+                            NP
          --------------------------->
                    (S\NP)
     --------------------------------<
                    S
-     that      is        the    cake      which       you    prefer    
-      NP   ((S\NP)/NP)  (NP/N)   N    ((N\N)/(S/NP))  NP   ((S\NP)/NP) 
-          --------------------->B
-               ((S\NP)/N)
-                               ------>T
-                             (N/(N\N))
-          --------------------------->B
-                ((S\NP)/(N\N))
-          ------------------------------------------->B
-                        ((S\NP)/(S/NP))
+     that      is        the    cake      which       you    prefer
+      NP   ((S\NP)/NP)  (NP/N)   N    ((N\N)/(S/NP))  NP   ((S\NP)/NP)
                                                      ----->T
                                                   (S/(S\NP))
                                                      ------------------>B
                                                            (S/NP)
+                                     ---------------------------------->
+                                                   (N\N)
+                               ----------------------------------------<
+                                                  N
+                       ------------------------------------------------>
+                                              NP
           ------------------------------------------------------------->
                                      (S\NP)
     -------------------------------------------------------------------<
                                      S
-
 
 
 Some other sentences to try:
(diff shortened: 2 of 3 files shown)


heatherleaf:
I'm not sure I think this is the best solution. It might affect efficiency, and I think it also affects the clarity of the code: People reading the code might start to believe that an OrderedDict is crucial for the correctness of the parser.

Also, I think we can get into the same problems in other places -- whenever there are sets and/or dicts around. And changing the code just to make doctests work is not ideal.

How about sorting the result trees in the doctests instead?

kmike:
OrderedDict is crucial for deterministic parsing, so if parser is intended to be deterministic then it is crucial for correctness, and if not then it is not needed and tests should be fixed as you proposed. 

I thought that parser was intended to be deterministic because of its interface. It doesn't return a set of parses, it returns a list of parses. Chart.child_pointer_list doesn't return a set, it also returns a list.

For example, consider this test failure (it sometimes passes and sometimes fails):

```
File "/Users/kmike/svn/nltk/nltk/test/ccg.doctest", line 118, in ccg.doctest
Failed example:
    for parse in parser.nbest_parse(sent,1):
        chart.printCCGDerivation(parse)  # doctest: +NORMALIZE_WHITESPACE
Expected:
     that      is        the    dough      which       you     will        eat          without           cooking
      NP   ((S\NP)/NP)  (NP/N)    N    ((N\N)/(S/NP))  NP   ((S\NP)/VP)  (VP/NP)  ((VP\VP)/VP['ing'])  (VP['ing']/NP)
          --------------------->B
               ((S\NP)/N)
                               ------->T
                              (N/(N\N))
          ---------------------------->B
                 ((S\NP)/(N\N))
          -------------------------------------------->B
                        ((S\NP)/(S/NP))
                                                      ----->T
                                                   (S/(S\NP))
                                                      ------------------>B
                                                            (S/VP)
                                                                                 ------------------------------------->B
                                                                                             ((VP\VP)/NP)
                                                                        ----------------------------------------------<Sx
                                                                                           (VP/NP)
                                                      ---------------------------------------------------------------->B
                                                                                   (S/NP)
          ------------------------------------------------------------------------------------------------------------>
                                                             (S\NP)
    ------------------------------------------------------------------------------------------------------------------<
                                                            S
Got:
     that      is        the    dough      which       you     will        eat          without           cooking     
      NP   ((S\NP)/NP)  (NP/N)    N    ((N\N)/(S/NP))  NP   ((S\NP)/VP)  (VP/NP)  ((VP\VP)/VP['ing'])  (VP['ing']/NP) 
          --------------------->B
               ((S\NP)/N)
                               ------->T
                              (N/(N\N))
                                                      ----->T
                                                   (S/(S\NP))
                                                      ------------------>B
                                                            (S/VP)
                                                                                 ------------------------------------->B
                                                                                             ((VP\VP)/NP)
                                                                        ----------------------------------------------<Sx
                                                                                           (VP/NP)
                                                      ---------------------------------------------------------------->B
                                                                                   (S/NP)
                                      -------------------------------------------------------------------------------->
                                                                           (N\N)
                               --------------------------------------------------------------------------------------->
                                                                          N
          ------------------------------------------------------------------------------------------------------------>
                                                             (S\NP)
    ------------------------------------------------------------------------------------------------------------------<
                                                            S

```

The parsing results are completely different because nbest_parse takes the first item from a list of parses. The existence of 'nbest_parse' method suggests that the order of results matter.

You're right that OrderedDict may have performance implications. The test script:

```
from nltk.ccg import chart, lexicon

lex = lexicon.parseLexicon('''
    :- S, NP, N, VP

    Det :: NP/N
    Pro :: NP
    Modal :: S\\NP/VP

    TV :: VP/NP
    DTV :: TV/NP

    the => Det

    that => Det
    that => NP

    I => Pro
    you => Pro
    we => Pro

    chef => N
    cake => N
    children => N
    dough => N

    will => Modal
    should => Modal
    might => Modal
    must => Modal

    and => var\\.,var/.,var

    to => VP[to]/VP

    without => (VP\\VP)/VP[ing]

    be => TV
    cook => TV
    eat => TV

    cooking => VP[ing]/NP

    give => DTV

    is => (S\\NP)/NP
    prefer => (S\\NP)/NP

    which => (N\\N)/(S/NP)

    persuade => (VP/VP[to])/NP
    '''
)
parser = chart.CCGChartParser(lex, chart.DefaultRuleSet)
sent = "that is the dough which you will eat without cooking".split()

```

Original version:

```
In [7]: %timeit parser.parse(sent)
10 loops, best of 3: 20.3 ms per loop
```

Version with nltk.util.OrderedDict:

```
In [3] %timeit parser.parse(sent)
10 loops, best of 3: 21.4 ms per loop
```

I may be wrong, but it seems that in practice the parser.nbest_parse method should be used more often than parser.parse. If so, that what matters is the performance of parser.nbest_parse, and maybe the way to improve it is to switch to iterators (and not build parses that are not requested)?

Also, if the parser is nondetermenistic by design, then _edge_to_cpls may be worth changed from a dict of dicts to a dict of sets. _edge_to_cpls is currently a dict of dicts which values are always True; switching to set may improve memory usage and code clarity. child_pointer_lists may also return set, not list in this case. 

What do you think about all these things?

heatherleaf:
The reason `nbest_parse` returns a list is because of the interface, and because _some_ parsers (read probabilistic ones) order their results. Nonprobabilistic chartparsing is nondeterminstic by definition -- a pure CFG cannot rank some parses higher than others.

I agree that the current interface is not the best. Since CFGs (and FCFGs) are unordered, the standard way to do things is to collect _all_ parse trees and do some post-processing to find out which one is best. That's why `nbest_parse` is more common. But the name (`nbest_parse`) suggests something completely different.

Also, a list is never a good thing when there might be thousands of trees. So I agree we should move to iterators instead. (This will make `iter_parse` obsolete).

But then we don't want to break compatibility, so here's a suggestion:
- `parse(s)` remains as before: It returns a single tree (or None); which one will be nondeterminstic for nonprobabilistic parsers
- `nbest_parse(s, n=None)` remains as before: It returns a list of trees, but it should be obsoleted for nonprobabilistic parsers
  - ideally I it should return an iterator, but that might break old code
- `all_parses(s)` or `parse_all(s)` is returns an iterator or a set, I don't know which is best here?

heatherleaf:
I like your suggestion to change `_edge_to_cpls` to a dict of sets. I'll add it to my Christmas TODO list :)

heatherleaf:
Oh, of course `all_parses(s)` should return an iterator! First, then it doesn't have to build the whole result set. Second, the trees might be duplicated, depending on the grammar.

kmike:
Oh, you're right about CFGs and their order; I was following my Python istincts and made such an oversight. 

It even wouldn't be possible to return a set of parse results because Tree is mutable (it is a subclass of list) and is not hashable because of that. 

+1 for not breaking the compatibility and not converting existing methods to return iterators.

But I'm lost now :) 

How is `iter_parse` different from the proposed `all_parses`? It seems nbest_parse returning `list(itertools.islice(self.iter_parse(sent), n))` is a right thing (and this is how ParserI.nbest_parse behave if there are no batch_nbest_parse, parse or batch_parse methods overriden).

heatherleaf:
You're right, `iter_parse` and my proposed `all_parses` are exactly the same thing. So we don't have to change the interface at all. 

But perhaps we should deprecate `nbest_parse` for nonprobabilistic parsers. But I guess that means making changes in the book too, so I guess Steven has to be able to say something too. :)

Anyway, is it ok with you Mikhail that we instead change the doctests to sort the results?

kmike:
Yes, changing tests is the way to go. I think tests shouldn't use `nbest_parse` (and `parse` by the way) for nonprobabilistic parsers and this should fix the issue.

It seems that for the better iteration support `nbest_parse` methods should be rewritten as `iter_parse` methods, so that `nbest_parse` would call `iter_parse` and not the other way round, but that's a different issue.

kmike:
Or maybe use (confusing) nbest_parse and sort the results, I don't know :) I'll take an another look tomorrow.

stevenbird:
Just a quick response for now: we can redesign interfaces in 2and3, even if it breaks the code samples and discussion in the book. A second edition of the book will be needed for Python 3 in any case.

kmike:
Surprisingly, the sorting of the results of nbest_parse and taking 0th item also produces unstable results. 

I think that's because of #319. 

So it seems it is not possible to fix ccg.doctest by sorting the results (neither of `nbest_parse` nor of `iter_parse`).

heatherleaf:
I guess the sorting should work after the fix to #319 is incorporated.

stevenbird:
I agree that nbest_parse() is anomalous. So is the way parse() returns the first result of (unsorted) nbest_parse(). Another option is for parse() and parse_iter() to both return sequences. An optional argument could limit the size of the result set of parse(). But perhaps it is time to switch to using iterators by default (i.e. NLTK methods that return sequences should do this in the form of iterators) and I've added a note about this to https://github.com/nltk/nltk/wiki/API-changes-for-Python-3

stevenbird:
By the way, I don't think the 5% performance loss is worth worrying about.

kmike:
Peter's commit improved the situation but it seems there are other things that prevent CCG results from being sorted. I think this has something to do with the ordering of ccg.api.FunctionalCategory and ccg.api.PrimitiveCategory instances (from which the resulting trees are constructed) - their `__cmp__` methods both return -1 if PrimitiveCategory and FunctionalCategory instances are compared with each other.

As for the performance, OrderedDict uses more memory than dict and I tested the speed only with a toy grammar. But IMHO the determinism is an useful property of the implementation, and for the real speed gains it may be a better idea to use a Cython-based implementation (or pray for PyPy improvements), so +1 for merging this even if it makes performance worse. Of course, it should be reverted if the loss will be proved to be dramatic, but it seems this is not the case.

heatherleaf:
Probably the problem is with `__cmp__` returning -1: This has the effect that x<y and y<x are both true. I'll add that as a new bug.

stevenbird:
Let's hope that fixes things. But if not, I don't think this issue with the ccg module should stand in the way. We could leave out the failing doctest for now.

kmike:
This doctest passes now (with OrderedDict fix), but fixing comparisons would be useful anyway. Fixed comparisons would be generally cleaner and (I suppose) they would enable working ccg.doctest even if we decide to revert OrderedDict fix.

heatherleaf:
fixing comparisons is now issue #324

PR #318, opened 2012-12-10
Fix for bug in internals.is_writable(path)

daghoidahl:
The check for whether the path is group writable did not consider all the user's groups.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -854,7 +854,7 @@ def is_writable(path):
         elif statdata.st_uid == os.getuid() and (perm & 0200):
             return True
         # are we in a group that can write to it?
-        elif statdata.st_gid == os.getgid() and (perm & 0020):
+        elif statdata.st_gid in os.getgroups() and (perm & 0020):
             return True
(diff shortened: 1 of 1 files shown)


kmike:
Is there an evidence that `os.getgid() in os.getgroups()` is always True? 
os.getgid() returns primary group and os.getgroups returns supplimentary groups. I tried this and os.getgid() was indeed the first item in os.getgroups() on my machine (OS X), but the man page for the Linux getgroups() C implementation says: "It is implementation-defined whether getgroups() also returns the effective group ID in the grouplist array." (http://linux.die.net/man/3/getgroups)

(a side note: I think that internals.is_writable is a bad style because it violates http://docs.python.org/2/glossary.html#term-eafp; internals.is_writable is used only once in nltk: in nltk.download; but this is an another question).

daghoidahl:
Good question. It's true for the Ubunto distro that I'm able to test with at the moment, but that does not prove anything. I've updated to check both `os.getgid()` and `os.getgroups()`.

PR #326, opened 2012-12-18
Adding Kneser-Ney Probability Distribution

iliakur:
Hello!

I noticed that while the NLTK probability module does contain most common probability distributions, it has none representing the Kneser-Ney technique. 
As a final project for a class I took a stab at creating one and was wondering if there would be any interest in incorporating it into the main branch of the NLTK project.  I currently coded up only one of the versions of this algorithm, but if there is demand and interest for more, I'm willing to write up some other versions as well.

Please feel free to let me know where you think my code or comments need improvements, I will gladly use this opportunity to learn something new about the toolkit and/or python.

Cheers!
Ilia.


nltk/probability.py
```diff
@@ -1221,7 +1221,7 @@ def __repr__(self):
 # species. (In drawing balls from an urn, the 'objects' would be balls
 # and the 'species' would be the distinct colors of the balls (finite
 # but unknown in number).
-#
+#
```
kmike: could you please remove the extra whitespaces (most editors have an option for this)?

nltk/probability.py
```diff
@@ -1621,6 +1621,155 @@ def update(self, sample, prob, log=True):
...
+        # set bins, discount parameter
+        self._bins = trigrams.B()
+        self._D = discount
+        # initialize cache used later to speed up probability calculation
+        self._cache = {}
+        # initialize internal bigram and trigram frequency distributions
+        self._bigrams = defaultdict(int)
+        self._trigrams = trigrams
+        # initialize helper dictionaries used to calculate probabilities
+        self._wordtypes_after = defaultdict(float)
```
kmike, line 1710: collections.Counter may be better here (for float division cast one of operands to float).

nltk/probability.py
```diff
@@ -1621,6 +1621,155 @@ def update(self, sample, prob, log=True):
...
+        self._trigrams_contain = defaultdict(float)
+        self._wordtypes_before = defaultdict(float)
+        # looping over trigrams set all the helper dictionaries
+        for w0, w1, w2 in trigrams.iterkeys():
```
kmike: `for w0, w1, w2 in trigrams:` seems more pythonic

nltk/probability.py
```diff
@@ -1621,6 +1621,155 @@ def update(self, sample, prob, log=True):
...
+            self._wordtypes_after[(w0,w1)] += 1
+            # increment trigram counter for trigrams containing w1
+            self._trigrams_contain[w1] += 1
+            # increment wordtype counter for words preceding w1 and w2
+            self._wordtypes_before[(w1,w2)] += 1
+
+    def prob(self, sample):
+        # allow sample to be a nested tuple
+        # this code deals with the possibilities
+        if len(sample) == 2 and type(sample[0]) == tuple:
```
kmike: Can you please provide an insight why nested tuple support is needed? 

If it is needed then isinstance() should be generally preferred; also, isn't tuple check too strict (what about lists)?

nltk/probability.py
```diff
@@ -1621,6 +1621,155 @@ def update(self, sample, prob, log=True):
...
+            raise ValueError('Expected a triple')
+        
+        if trigram in self._cache:
+            # check if the answer has been cached
+            return self._cache[trigram]
+        else:
+            # if not, perform computations
+            if trigram in self._trigrams:
+                # if the sample trigram was seen during training
+                prob = (self._trigrams[trigram] - self._D)/self._bigrams[(w0, w1)]
```
kmike: `self._D` is used here but there is a `discount()` method. `discount()` should be either remover or used systematically.

kmike:
Hello Ilia,

Thanks for the contribution!

Could you please also add some tests for KneserNeyProbDist to nltk/test/probability.doctest?

kmike:
I didn't dig into the implementation; the comments are from a programmer's point of view.

iliakur:
Perfect! That's exactly the part I'm not sure about :)

nltk/probability.py
iliakur, line 1710: Counter, unlike defaultdict, doesn't seem to be imported. Should I import that specially for this? What is the advantage of using that over defaultdict?

nltk/probability.py
iliakur: I forgot that dictionaries implement the iterator protocol.... will change to what you suggest.

nltk/probability.py
iliakur: I thought of discount() as a way for the user to retrieve the discount value for viewing/manipulating outside of the class. Should I rename it to 'get_discount()' or perhaps 'show_discount()'
I also thought that since we are referencing all sorts of count dictionaries internally, we may as well avoid the overhead of calling a method. It's not much, but I'm assuming people will want to run the prob() method on large chunks of data, so every little bit may count.

Let me know what you think!

nltk/probability.py
iliakur: This is something I actually wasn't entirely whether to add or not, so all/any input/ideas are most welcome.

My original thinking when adding support for nested tuples was that some people may prefer to store trigrams that way. Lists are really also something I should have considered.

I guess my question is: should I expand support for various input datatypes or should I completely deprecate it and instead strictly specify what is acceptable?

nltk/probability.py
kmike, line 1710: Please disregard my comment about collections.Counter because it requires Python 2.7. 

Advantage over the defaultdict is that the intent is more clear; Counter also has some methods (like "most_common") that defaultdict doesn't have, but they are not needed here anyway.

nltk/probability.py
kmike: I suggest to remove this code, document that you're expecting 3-tuple and unpack it using `w0, w1, w2 = sample`. This would also work with lists, namedtuples and other data types.

I don't even think that raising the ValueError is necessary (because the exception would be raised anyway if sample can't be converted to w0, w1, w2), but this is just a personal preference and some people prefer manual data validation.

nltk/probability.py
kmike: Then rename "_D" to "discount" and remove `discount()` and `set_discount()` methods. Getters and setters are generally considered a bad style in Python, and discount may be converted to property later if necessary (see e.g. http://blaag.haard.se/What-s-the-point-of-properties-in-Python/ ).

iliakur: Sounds good, thanks for the link!
I'll implement this and the tuple change as soon as I get home from work!

kmike: I've changed my mind :) You inherit from ProbDistI and it has discount() method, so having 'discount' attribute in KneserNeyProbDist would be very confusing. So IMHO it is better to 

a) make sure that KneserNeyProbDist.discount() means the same as ProbDistI.discount();
b) if they mean the same, use a function call to get the discount (so that overridden discount() method would behave as expected). I don't think it will have a performance impact worth worriyng about.

nltk/probability.py
```diff
@@ -1621,6 +1621,155 @@ def update(self, sample, prob, log=True):
...
+class KneserNeyProbDist(ProbDistI):
+    """
+    Kneser-Ney estimate of a probability distribution. This is a version of
+    back-off that counts how likely an ngram is provided the n-1gram had
+    been seen in training. Extends the ProbDistI interface, requires a trigram
+    FreqDist instance to train on. Optionally, a different from default discount
+    value can be specified. The default discount is set to 0.75.
+
+    """
+    def __init__(self, trigrams, discount=0.75):
```
kmike: nitpick: for most other ProbDistI subclasses the first argument is named 'freqdist'

nltk/probability.py
```diff
@@ -1621,6 +1621,155 @@ def update(self, sample, prob, log=True):
...
+        :type discount: float (preferred, but int possible)        
+        :rtype: None
+        """
+        self._D = discount
+
+    def samples(self):
+        return self._trigrams.keys()
+
+    def max(self):
+        return self._trigrams.keys()[0]
```
kmike: is this really correct?

iliakur: _trigrams is and instance of FreqDist. That class has a custom .keys() method that sorts the keys by values before returning a list of them.
This means the first member of this list will be the one with the highest frequency and (as a consequence) highest probability.

nltk/probability.py
iliakur: I think I'll just keep the unpacking line as you suggested. As for the error, I kind of want to keep the manual one so that it's more explicit about what exactly is expected as input.

nltk/probability.py
iliakur: I forgot that was the reason I bothered with discount() in the first place, haha!!

iliakur:
OK, the only thing that remains is to figure out how to make the new probdist play nice in the probability.doctest file. I suspect we may have to finagle around a bit to get it to process words with tags correctly. I will only have time to do this on the weekend, unfortunately. For now I've only gotten rid of some superfluous brackets in list comprehensions in the doctest file.

iliakur:
So theoretically the test should work. And it does on my machine up to the point when I try to evaluate the probdist. then it crashes. My computer is pretty powerful (lenovo x220 with i5 processor), so I suspect the problem is in the immensity of the data structures generated by both hmm.evaluate() and the Kneser-Ney distribution. 
I tried running it on a sample of 100 sentences from the brown corpus and got accuracy of 86%. For some reason when I tried running 300 sentences, the accuracy dropped significantly to 36%...

nltk/test/probability.doctest
```diff
@@ -59,10 +59,10 @@ We extract a small part (500 sentences) of the Brown corpus
 We create a HMM trainer - note that we need the tags and symbols
 from the whole corpus, not just the training corpus
 
-    >>> tag_set = list(set([tag for sent in corpus for (word,tag) in sent]))
+    >>> tag_set = list(set(tag for sent in corpus for (word,tag) in sent))
     >>> print len(tag_set)
     92
-    >>> symbols = list(set([word for sent in corpus for (word,tag) in sent]))
+    >>> symbols = list(set(word for sent in corpus for (word,tag) in sent))
```
iliakur, line 65: In this line and line 62 above, I didn't see the point of creating a list that would then be turned into a set. Passing a simple iterator to the set() works just fine as well.

nltk/test/probability.doctest
```diff
@@ -133,6 +135,38 @@ Good Turing Estimation
...
+our testing accordingly.
+
+    >>> corpus = [[((x[0],y[0],z[0]),(x[1],y[1],z[1])) 
+            for x, y, z in nltk.trigrams(sent)]
+                for sent in corpus]
+
+We will then need to redefine the rest of the training/testing variables
+    >>> tag_set = list(set(tag for sent in corpus for (word,tag) in sent))
+    >>> print len(tag_set)
+
```
kmike, line 156: Did you run this doctest? It doesn't look like doctest :) corpus = ... statement is incorrect (take a look at standard Python console output to get an idea how multiline statements look like in doctests). It is also better to write 

```
>>> len(tag_set)
1234
```

instead of `print len(tag_set)` (because print is unnecessary here and the syntax changes in Python 3.x)  - and don't forget to include the output values to check.

nltk/test/probability.doctest
kmike, line 65: Thanks for the notice! I've converted some list comprehensions to generators a couple of days ago in 2and3 branch ( https://github.com/nltk/nltk/commit/e92945018df7774b860643348b3740004fb92803 ), but this one (and several others) was missing; this is fixed in https://github.com/nltk/nltk/commit/89575bda2419f78746f864e6605a2e8538c017d4 (2and3 branch again).

nltk/probability.py
kmike: This builds a new list of all keys from self._trigrams and returns 0th item (and would fail under Python 3.x by the way, because keys() is not list under Python 3.x); self._trigrams.max() may be IMHO a better idea because it may reuse FreqDist._max_cache, and the intent of the code would be more clear.

iliakur:
Hey!
I was wondering what the next step in the process is and whether you need anything more from me in order to accept the pull request. 
If that is the case, just let me know, I'm glad to be of service :)

nltk/probability.py
```diff
@@ -1621,6 +1621,153 @@ def update(self, sample, prob, log=True):
...
+        """
+        self._D = discount
+
+    def samples(self):
+        return self._trigrams.keys()
+
+    def max(self):
+        return self._trigrams.max()
+
+    def __repr__(self):
```
kmike, line 1778: For the sake of correctness, please wrap KneserNeyProbDist with `@compat.python_2_unicode_compatible`. unicode_literal future import is in effect in this file, so `__repr__` would return unicode string and this may have subtle consequences if `@compat.python_2_unicode_compatible` decorator is not applied.

nltk/test/probability.doctest
```diff
@@ -133,6 +135,43 @@ Good Turing Estimation
...
+our testing accordingly.
+
+    >>> corpus = [[((x[0],y[0],z[0]),(x[1],y[1],z[1])) 
+    ...     for x, y, z in nltk.trigrams(sent)]
+    ...         for sent in corpus[:100]]
+
+We will then need to redefine the rest of the training/testing variables
+    >>> tag_set = list(set(tag for sent in corpus for (word,tag) in sent))
+    >>> len(tag_set)
+    >>> 906
```
kmike: I don't think this is a correct doctest (output should be without ">>>"). I think it may be a good idea to run the tests :) If you update code to the latest master, then probability.doctest may be run with the following command (it requires nose to be installed):

```
nltk/test/runtests.py nltk/test/probability.doctest
```

Take a look at this document for more info: https://github.com/nltk/nltk/blob/master/web/dev/local_testing.rst

iliakur:
Wow, doctests are completely new to me and, to be honest, a little confusing.... 
OK, so I tried running:
<code> nltk/test/runtests.py nltk/test/probability.doctest </code>
but got the following error:

  File "nltk/test/runtests.py", line 40, in <module>
    from nltk.test.doctest_nose_plugin import DoctestFix
  File "/home/ilia/ilia_dev/python/nltk/nltk/**init**.py", line 116, in <module>
    import ccg
  File "/home/ilia/ilia_dev/python/nltk/nltk/ccg/**init**.py", line 14, in <module>
    from nltk.ccg.combinator import (UndirectedBinaryCombinator, DirectedBinaryCombinator,
  File "/home/ilia/ilia_dev/python/nltk/nltk/ccg/combinator.py", line 8, in <module>
    from nltk.parse import ParserI
  File "/home/ilia/ilia_dev/python/nltk/nltk/parse/**init**.py", line 76, in <module>
    from malt import MaltParser
  File "/home/ilia/ilia_dev/python/nltk/nltk/parse/malt.py", line 17, in <module>
    from nltk.tag import RegexpTagger
  File "/home/ilia/ilia_dev/python/nltk/nltk/tag/**init**.py", line 76, in <module>
    from nltk.tag.hmm import HiddenMarkovModelTagger, HiddenMarkovModelTrainer
  File "/home/ilia/ilia_dev/python/nltk/nltk/tag/hmm.py", line 1041
    print starting
                 ^
SyntaxError: invalid syntax

Now, I've inspected that file and found nothing wrong with the syntax. Is this test trying to run things in "python 3.0" mode?

As for tox, I tried installing that, but it seems to require scipy and scikit-learn, which in turn require some C/Fortran libraries. 
I don't want to sound lazy, but must I install all this jazz to run the doctest? Am I missing something?

kmike:
hmm.py uses `from __future__ import print_function` to make code 2x-3x compatible (take a look at the top of the file) so "print 123" syntax is invalid in hmm.py (and generally in NLTK codebase), and this is not just for testing. 

See http://docs.python.org/2.6/whatsnew/2.6.html#pep-3105-print-as-a-function

Also, NLTK tests use `from __future__ import print_function` implicitly and tests should assume this import is in effect. 

This test doesn't fail in master; it looks like it fails because your introduced this print statement in https://github.com/Copper-Head/nltk/commit/5ef35031fc1fbc8f4d5a616322105a9d016b7023.

numpy and scikit-learn are indeed hard to install to virtualenv, and unfortunately probability.doctest requires numpy (because it tests HMM which uses numpy) so you can't just use "light" ..-nodeps environments. You're right that installing all these things just to run a simple doctest is a horrible contributing experience! By using `nltk/test/runtests.py nltk/test/probability.doctest` command you may avoid some troubles and run tests against current environment (which may have numpy installed from a binary package or system packages repository).

Another option is to create an "hmm.doctest" file and move HMM-related stuff from probability.doctest to this file; this way probability.doctest wouldn't require numpy and one would be able to run e.g.

```
tox -e py32-nodeps,py26-nodeps -- probability.doctest
```

iliakur:
Thanks for all the valuable info!

I've updated to the latest version of the current nltk source. I've also made a minor change in .gitignore, please let me know if you'd rather I revert it.

As for the doctest, I tried running the command:
<code> nltk/test/runtests.py nltk/test/probability.doctest </code>

And got the following output:

---

  File "/usr/local/lib/python2.7/dist-packages/nose/plugins/manager.py", line 154, in generate
    for r in result:
  File "/home/ilia/ilia_dev/python/nltk/nltk/test/doctest_nose_plugin.py", line 43, in loadTestsFromFile
    for case in cases:
  File "/usr/local/lib/python2.7/dist-packages/nose/plugins/doctests.py", line 287, in loadTestsFromFile
    filename=filename, lineno=0)
  File "/usr/lib/python2.7/doctest.py", line 645, in get_doctest
    return DocTest(self.get_examples(string, name), globs,
  File "/usr/lib/python2.7/doctest.py", line 659, in get_examples
    return [x for x in self.parse(string, name)
  File "/usr/lib/python2.7/doctest.py", line 621, in parse
    self._parse_example(m, name, lineno)
  File "/usr/lib/python2.7/doctest.py", line 691, in _parse_example
    lineno + len(source_lines))
  File "/usr/lib/python2.7/doctest.py", line 777, in _check_prefix
    (lineno+i+1, name, line))
ValueError: line 63 of the docstring for probability.doctest has inconsistent leading whitespace: '<<<<<<< HEAD'

---

Ran 1 test in 0.001s

FAILED (errors=1)

Now, when I try to indent this line, the same error pops up for line 68, then 73 if I indent 68.
I've updated to the latest master, should I indent all such lines or are they supposed to be unindented?

kmike:
That's because you has an unresolved merge conflict in this line:

```
<<<<<<< HEAD
... 
=======
... 
>>>>>>> e48192bf984571ed262532d68e46260178dd3da4
```

nltk/probability.py
```diff
@@ -1638,6 +1638,154 @@ def update(self, sample, prob, log=True):
...
+        '''
+        Return a string representation of this ProbDist
+
+        :rtype: str
+        '''
+        return '<KneserNeyProbDist based on {} trigrams'.format(self._trigrams.N())
```
kmike: `"{}".format(foo)` is only available in 2.7+; maybe use `"{0}".format(foo)`?

stevenbird:
Thanks for the contribution. Its looking good. I tested it on Python 2.6 and 3.2 and had no problems. Can I suggest fixing some typos in the KneserNeyProbDist docstring, and removing some redundant in-line comments about adding and incrementing. Also, please add your name to the top of probability.py as one of the authors.

iliakur:
Fixed.

stevenbird:
Thanks for the further fixes. I've tested them out, and done some minor reformatting (line length, comment cleanups), and merged.

Issue #321, opened 2012-12-11
Deprecate all algorithms from maxent.MaxentClassifier._SCIPY_ALGS


kmike:
Recent Scipy no longer has this stuff.

kmike:
This is done in a 2and3 branch.

PR #327, opened 2012-12-18
Fixed classify.doctest

hleeldc:
Fixed classify.doctest as well as related modules tested by it.

The controversial scipy.maxentropy module was added as nltk/classify/_scipy_maxentropy because scipy no longer maintains the module.


hleeldc:
I just saw issue #321 which requests to deprecate code that uses scipy.maxentropy.

hleeldc:
nltk.classify.svm depends on pysvmlight, which doesn't support python 3 currently. For this fix, I created a patch that works with python 3 (but not with python 2). Here's the patched version.

https://bitbucket.org/hleeldc/pysvmlight

hleeldc:
nltk.classify.maxent includes an algorithm that depends on TADM. TADM is out of date but I managed to compile it (https://github.com/hleeldc/nltk/wiki/Installing-tadm). The algorithm produced results that look suspicious. I excluded TADM from the doctest, until the results are verified...

stevenbird:
We've decided to drop support for scipy.maxentropy.

Issue #324, opened 2012-12-11
__cmp__ returns -1


heatherleaf:
In some classes, the `__cmp__` method returns -1 if `type(self) != type(other)`. This has the effect that `x<y` and `y<x` both become true.

The following classes are affected, as far as I can see:

```
featstruct.Feature
parse.chart.TreeEdge
parse.chart.LeafEdge
parse.featurechart.FeatureTreeEdge
parse.pchart.ProbabilisticTreeEdge
parse.projectivedependencyparser.DependencySpan
```

There are two solutions, which one is best?
- Fix `__cmp__(self,other)`. Convert each function like
  
  def **cmp**(self, other):
      if type(self) != type(other): return -1
      return Tree.**cmp**(self, other) or cmp(self.prob(), other.prob())

into something like this:

```
def __cmp__(self, other):
    return cmp(type(self), type(self)) or Tree.__cmp__(self, other) or cmp(self.prob(), other.prob())
```
- Replace all occurrences of `__cmp__` with rich comparisen operators, `__eq__`, `__le__`, etc.

Which one is best? The latter is tedious -- another reason for backporting compat to master.

kmike:
Thanks for the research!

Another option: fix it in a 2and3 branch :) This may be easier than fixing it in master because test running is improved in 2and3 (and there is compat.py). 

+1 for rewriting `__cmp__` with `__eq__`, `__lt__`, etc. We will have to do this anyway for Py3 compatibility. If the fixes are for master branch, it is fine to backport compat.py.

It may be better to raise an exception if the comparison functions don't know how to compare objects. This is one of the things that was systematically fixed in Python 3.x:

Python 2.x:

```
>>> 1 > "5"
False
```

Python 3.x:

```
>>> 1 > "5"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unorderable types: int() > str()
```

Why would anybody compare featstruct.Feature with parse.projectivedependencyparser.DependencySpan?

heatherleaf:
I didn't know about that change in Py3. I agree we fix it in 2and3 branch. 

I can volunteer, but it will take a few days.

kmike:
This would be great. 

My next TODO item is finishing the unicode support, it won't interfere with comparison fixes.

PR #328, opened 2012-12-21
Fixing `__cmp__` for 2and3

heatherleaf:
Here's a load of changes to the 2and3 branch. 

I've changed all `__cmp__` to `__eq__` and `__lt__` together with a `@total_ordering`. 

I have also made som changes to string handling, which might interfere with kmike's current work. Please look through them!

I've added a "text" format to `nltk.data.load` and added an optional `encoding` argument to several file reading and string parsing functions.

The following doctests now pass for py26, py27 and py32 (I don't have py33 so I can't test that):
- align ccg chat80 chunk collocations corpus data dependency drt featgram featstruct grammar grammartestsuites internals japanese logic metrics parse probability relextract resolution segmentation semantics simple stem tag tokenize tree treetransforms util wordnet

I can't test the following doctests, since I don't have the necessary dependencies:
- childes classify discourse gluesemantics inference nonmonotonic
  - (but all gluesemantics tests pass until it tries to use the MaltParser)

The following four tests fail:
- misc portugese_en sourcedstring toolbox


heatherleaf:
Oh, it fixes issue #324

kmike:
Wow, that's awesome!

stevenbird:
Thanks Peter! Mikhail: I'll leave you to merge this into 2and3 when you're ready.

heatherleaf:
Some more information:

There's an API change in `sem.glue.GlueDict`, so that `GlueDict('glue.semtype')` now must be written like this:
- `GlueDict('nltk:grammars/sample_grammars/glue.semtype')`

`nltk.data.load` has a new format for text:
- a new argument `encoding` for specifying the unicode encoding -- it works for all non-binary formats (i.e., all except pickle, yaml and raw)
- if `encoding` is not specified, `load` first tries with UTF-8, and if that fails it falls back to Latin-1
- this means that `load` reads a unicode file for all non-binary formats
- "text" format returns a unicode string, "raw" format returns a byte string

PR #331, opened 2012-12-29
First commit to track use of __str__

dzen:
I had some issue working on french (and unicode) text.

Here is a first commit.


kmike:
Hello, could you try using 2and3 branch? This issue should be fixed there.

PR #333, opened 2013-01-02
fix snowball stemmer bug in handling word `y's`

cllu:
STEP 1c in line 828 checks `len(word) > 2` as last condition. However, this will cause `string index out of range` error when stem the word `y's` since `word` will be `y` when we come to line 828. Move the condition `len(word)>2` to beginning will solve this issue.


kmike:
Thanks!

Issue #265, opened 2012-06-04
babelfish translator service is gone


kmike:
I've recently ported nltk.misc.babelfish to py3 and fixed #180 only to realize that today tests begin to fail again :) 

There is no longer http://babelfish.yahoo.com/translate_txt -  it redirects to http://www.microsofttranslator.com/ which is an another translator service with different API. See http://blogs.msdn.com/b/translation/archive/2012/05/30/welcoming-yahoo-babel-fish-users.aspx

hlongmore:
I noticed this while going through the [nltk book](http://nltk.googlecode.com/svn/trunk/doc/book/ch01.html), and started working on a version of babelize that would be able to use the Microsoft version of babelfish, or translate.google.com.  Both options have issues.

The MS version requires an access token, and thus an [Azure account](http://msdn.microsoft.com/en-us/library/hh454950), or the addition of another dependency such as selenium, unless I'm missing some other available option.

The Google version probably also requires something such as selenium, though it may be possible to get it to work by sending a different User-Agent through subclassing [UrlOpener or FancyUrlOpener](http://docs.python.org/library/urllib.html#url-opener-objects), or by setting the headers of a [urllib2.Request object](http://docs.python.org/library/urllib2.html#urllib2.Request).  I haven't gotten around to trying either of these options yet, but I'm guessing there is enough Javascript in the returned result to require selenium.

kmike:
I believe Google provide Translate API that is not free; scraping translate.google.com is possibly a ToS voilation.

hlongmore:
Yes, there is a for-fee API, and after checking https://developers.google.com/translate/v2/terms and https://developers.google.com/terms/, I agree, scraping translate.google.com is likely a ToS violation.  Though the terms apply to those using the Google APIs, and it could be argued that if one is not a Google API consumer, they never agreed to the terms, my guess is that goes against the intent of providing the API in the first place.

lusy:
It's really a pity that babelfish is broken and that both Google and Microsoft APIs don't seem to be ok to use in this case... 
But aren't there other options? For example I've just stumbled upon this API here: http://frengly.com/translationAPI, however, it doesn't seem to be that good either, because it requires username and password :/

kmike:
There is http://api.yandex.ru/translate/ that doesn't require authorization according to API docs and ToS.

[...]

PR #335, opened 2013-01-03
Remove nltk.misc.babelfish

kmike:
See also: https://github.com/nltk/nltk/issues/265

I've removed babelfish in master (which mean NLTK 2.0) to make an error message better for users of the published NLTK Book.


alexrudnick:
Thanks for doing this! Looks good to me.

PR #336, opened 2013-01-05
fixed doctype regex to handle doctype tag w/o square brackets

cadaeib:
If there's no '[' in the DOCTYPE declaration, [^[]\* will match as much of the block as possible, "eating" several start tags that we don't actually want to skip. This breaks [the sense-tagged WordNet corpus](http://wordnet.princeton.edu/glosstag.shtml). 

Since the XML spec doesn't allow ">" in names, this fix shouldn't break anything. Running runtests.py fails, but it fails with the same errors as without this fix, so the errors are (hopefully) not mine. (Should I write tests for this? I've tested it with the WordNet corpus, but I'm not sure if it's worth adding a unit test for this, especially since I'm not adding anything new.)

### DIFF
--- a/nltk/corpus/reader/xmldocs.py
+++ b/nltk/corpus/reader/xmldocs.py
@@ -220,7 +220,7 @@ class XMLCorpusView(StreamBackedCorpusView):
         (?P<COMMENT>        <!--.*?-->                          )|
         (?P<CDATA>          <![CDATA[.*?]]>                     )|
         (?P<PI>             <\?.*?\?>                           )|
-        (?P<DOCTYPE>        <!DOCTYPE\s+[^\[]*(\[[^\]]*])?\s*>  )|
+        (?P<DOCTYPE>        <!DOCTYPE\s+[^\[^>]*(\[[^\]]*])?\s*>)|
         # These are the ones we actually care about:
         (?P<EMPTY_ELT_TAG>  <\s*[^>/\?!\s][^>]*/\s*>            )|
         (?P<START_TAG>      <\s*[^>/\?!\s][^>]*>                )|


alexrudnick:
Sorry for the delay! I'll look into this next and get back to you soon!

alexrudnick:
Hey @cadaeib , would you paste some examples of doctype declarations that this regex handles that couldn't be handled previously? I took a minute to look at the corpus you mentioned, but it wasn't immediately obvious to me where to look.

Thanks!

cadaeib:
Hi @alexrudnick, thanks for taking a look at this! 

The line that the current version breaks on is " <!DOCTYPE wordnet SYSTEM "glosstag.dtd">" at the start of, for instance, nouns.xml. The part of the regex that is supposed to just match "wordnet SYSTEM "glosstag.dtd"" (or, in general, the "S Name (S ExternalID)? S?" part of [the XML spec](http://www.w3.org/TR/2008/REC-xml-20081126/#sec-prolog-dtd)) in fact greedily matches the closing angle bracket, as well as most of the rest of the block (since the regex eats newlines). This includes starting tags, which makes it sad later on when it finds end tags that don't have start tags.

Should I also commit the Python file I was using to test the changes?

alexrudnick:
Hey @cadaeib, sorry for so much discussion on such a little change, but as far as I can tell, the existing regex does match that doctype? Can you give a short code snippet that shows how the reader breaks in your use case? I've got noun.xml handy, and I can read it with the corpus reader like this:

```
>>> import nltk
>>> reader = nltk.corpus.reader.xmldocs.XMLCorpusReader('/home/alex/Desktop', ['noun.xml'])
>>> len(reader.words())
4523765
```

(this is with the latest code from the repository ...)

cadaeib:
Oops, forgot to mention that it's actually XMLCorpusView... I suppose that is an important detail.

This is what fails, for me (on master):

> > > import nltk
> > > view = nltk.corpus.reader.xmldocs.XMLCorpusView('/home/amy/pythonishstuffs/WordNet-3.0/glosstag/merged/noun.xml', '.*/terms')
> > > file = open("/home/amy/pythonishstuffs/WordNet-3.0/glosstag/merged/noun.xml")
> > > view.read_block(file)
> > >   File "<stdin>", line 1, in <module>
> > >   File "/usr/local/lib/python2.7/dist-packages/nltk-2.0.4-py2.7.egg/nltk/corpus/reader/xmldocs.py", line 326, in read_block
> > >     raise ValueError('Unmatched tag </%s>' % name)
> > > ValueError: Unmatched tag </def>

(And thank you for taking the time to do all these tests and code reviews! I am still learning a lot about nltk, python, and contributing to open source, so please excuse my cluelessness :))

alexrudnick:
OK, thanks for the patch and the explanation! Sorry it took me so long to understand the problem!

PR #338, opened 2013-01-10
Create interface to Stanford Parser

rohithb:
An interface for dependancy parsing using Stanford parser

### DIFF
--- a/nltk/parse/stanford-parser.py
+++ b/nltk/parse/stanford-parser.py
@@ -1,0 +1,55 @@
+#
+# Author: Rohith Bhaskaran <vrarohith@gmail.com>
+#
+
+import os
+import tempfile
+
+class StanfordParser:
+    def __init__(self,working_dir=None):
+        """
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for the contribution. Would you be happy to do a bit more work on it before we pull it into the NLTK repository? Here's the main things that need to happen:

a. review our developer's guide and make the required changes, e.g. assigning copyright, docstrings, PEP-8 spacing
http://code.google.com/p/nltk/wiki/DevelopersGuide

b. fix the naming: we don't use hyphens in module names (just call it stanford.py); there are many stanford parsers, so this could be StanfordDependencyParser

c. it should inherit from the ParserI interface class

d. use our existing approach for locating external binaries (see e.g. tag/stanford.py)

stevenbird:
Hi -- I hope you can take a look at this, so we can add your code to NLTK.

stevenbird:
Please re-open this issue once the above issues have been addressed.

PR #339, opened 2013-01-12
Bugfix for snowball stemmer typo introduced in #333

myleott:
The fix for #333 introduced a new bug. In the conditional:

```
self.__vowe
```

should be:

```
self.__vowels
```


alexrudnick:
Great! Thanks!

PR #341, opened 2013-01-16
Fixed by sorting in descending order of similarity score

aboveaboutbelow:


stevenbird:
Thanks for the fix. Resolved in https://github.com/nltk/nltk/commit/dc3ecdb1b58fea86e9057eb84152f8184c1e524d (using reverse=True flag).

PR #344, opened 2013-01-16
Two fixes for nltk.downloader

### LINKED ISSUE #342: downloader breaking on unicode issues
At the tip of master right now, on Python 2.7.3, trying to run nltk.download() and then doing 'u' for update, we get a break due to a unicode issue...

```
>>> import nltk
>>> nltk.download()
NLTK Downloader
    d) Download   l) List    u) Update   c) Config   h) Help   q) Quit
Downloader> u

Will update following packages (o=ok; x=cancel)
  File "/home/alex/nltk/nltk/downloader.py", line 645, in download
    self._interactive_download()
  File "/home/alex/nltk/nltk/downloader.py", line 961, in _interactive_download
    DownloaderShell(self).run()
  File "/home/alex/nltk/nltk/downloader.py", line 1000, in run
    self._simple_interactive_update()
  File "/home/alex/nltk/nltk/downloader.py", line 1047, in _simple_interactive_update
    print('  [ ] %s %s' % (pid.ljust(20, '.'), name))
TypeError: must be char, not unicode
```

It also breaks, in a different way, on Python 3.2.3:
    TypeError: unorderable types: Package() < Package()

Working on a fix for both!

alexrudnick:
This should fix #342 and make nltk.download() work again!
- Make the strings in the XML ElementTree attributes be unicode strings (even under Python 2)
- Make nltk.downloader.Collection and nltk.downloader.Package sortable, so the downloader doesn't barf in Python 3.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -259,6 +259,9 @@ class Package(object):
             xml.attrib[key] = compat.text_type(xml.attrib[key])
         return Package(**xml.attrib)
 
+    def __lt__(self, other):
+        return self.id < other.id
+
     def __repr__(self):
         return '<Package %s>' % self.id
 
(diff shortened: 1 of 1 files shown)


kmike:
Looks good!

PR #345, opened 2013-01-17
corrections on the windowdiff metric

DavidDoukhan:
Several corrections are proposed:
1 - distinction between the weighted and unweighted variants of WindowDiff
This metric was initially proposed as being unweighted, and implemeted weighted in NLTK.
This may be confusing for those who do not know about the weighted variant
2- bug in the window size: last implementation considered window sizes of length k+1, it should be k
3- final division, proposed in all variants, but not present in the current implementation

### DIFF
--- a/nltk/metrics/segmentation.py
+++ b/nltk/metrics/segmentation.py
@@ -48,22 +48,12 @@ except ImportError:
 
 from nltk.compat import xrange
 
-def windowdiff(seg1, seg2, k, boundary="1"):
+def windowdiff(seg1, seg2, k, boundary="1", weighted=True):
     """
     Compute the windowdiff score for a pair of segmentations.  A
     segmentation is any sequence over a vocabulary of two items
     (e.g. "0", "1"), where the specified boundary value is used to
     mark the edge of a segmentation.
-
-    >>> s1 = "00000010000000001000000"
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for the contribution. We need more information before making these changes. Are you trying to bring the implementation into line with the formula used by Pevzner and Hearst (2002, p26)? http://www.aclweb.org/anthology/J02-1002.pdf

DavidDoukhan:
yes.
this is done together with the weighted variant p27

2013/1/18 Steven Bird notifications@github.com

> Thanks for the contribution. We need more information before making these
> changes. Are you trying to bring the implementation into line with the
> formula used by Pevzner and Hearst (2002, p26)?
> http://www.aclweb.org/anthology/J02-1002.pdf
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/nltk/nltk/pull/345#issuecomment-12399521.

## 

David Doukhan

stevenbird:
Thanks. I'm doing some testing, adding tests, and fixing up the documentation. By the way, you used "rtype" (return type) instead of "type" to specify the type of a parameter.

stevenbird:
Based on Pevzner and Hearst's comment on p27, we should default to the unweighted version.

PR #351, opened 2013-01-27
fixes for tkinter in nltk.app and nltk.draw packages

hleeldc:
In the nltk.app and nltk.draw packages, updated modules that use tkinter by running 2to3.py. Manually fixed some spots that 2to3.py missed.

To ensure backward compatibility with python2.6, updated
- nltk/compat.py to add an import hook (PEP 302) for tkinter
- nltk/util.py to update in_idle function which was broken for py3

### DIFF
--- a/nltk/app/chartparser_app.py
+++ b/nltk/app/chartparser_app.py
@@ -36,14 +36,15 @@ edge you wish to apply a rule to.
 # At some point, we should rewrite this tool to use the new canvas
 # widget system.
 
-from __future__ import print_function
 
+
+import nltk.compat
 import pickle
-from tkFileDialog import asksaveasfilename, askopenfilename
-import Tkinter
+from tkinter.filedialog import asksaveasfilename, askopenfilename
+import tkinter
 import math
 import os.path
-import tkFont, tkMessageBox
+import tkinter.font, tkinter.messagebox
 
 from nltk.parse.chart import (BottomUpPredictCombineRule, BottomUpPredictRule,
                               Chart, LeafEdge, LeafInitRule, SingleEdgeFundamentalRule,
@@ -106,12 +107,12 @@ class ChartMatrixView(object):
         self._selected_cell = None
 
         if toplevel:
-            self._root = Tkinter.Toplevel(parent)
+            self._root = tkinter.Toplevel(parent)
             self._root.title(title)
             self._root.bind('<Control-q>', self.destroy)
             self._init_quit(self._root)
         else:
-            self._root = Tkinter.Frame(parent)
+            self._root = tkinter.Frame(parent)
 
         self._init_matrix(self._root)
         self._init_list(self._root)
@@ -127,18 +128,18 @@ class ChartMatrixView(object):
         self.draw()
 
     def _init_quit(self, root):
-        quit = Tkinter.Button(root, text='Quit', command=self.destroy)
+        quit = tkinter.Button(root, text='Quit', command=self.destroy)
         quit.pack(side='bottom', expand=0, fill='none')
 
     def _init_matrix(self, root):
-        cframe = Tkinter.Frame(root, border=2, relief='sunken')
+        cframe = tkinter.Frame(root, border=2, relief='sunken')
         cframe.pack(expand=0, fill='none', padx=1, pady=3, side='top')
-        self._canvas = Tkinter.Canvas(cframe, width=200, height=200,
+        self._canvas = tkinter.Canvas(cframe, width=200, height=200,
                                       background='white')
         self._canvas.pack(expand=0, fill='none')
 
     def _init_numedges(self, root):
-        self._numedges_label = Tkinter.Label(root, text='0 edges')
+        self._numedges_label = tkinter.Label(root, text='0 edges')
         self._numedges_label.pack(expand=0, fill='none', side='top')
 
     def _init_list(self, root):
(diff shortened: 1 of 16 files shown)


kmike:
Clever trick with import hook! But honestly, I think the less clever approach would be better. Something like this:

```
# compat.py
# ...
if PY3:
    import tkinter.font as tkinter_font
else:
    import tkFont as tkinter_font

# app code
from nltk import compat

compat.tkinter_font
```

Pros of a simpler approach:
- import nltk.compat doesn't have global side effects;
- less code;
- code is simpler.

Cons:
- instead of `tkinter.font` nltk.app code would have `compat.tkinter_font` so it wouldn't be possible to just remove "nltk.compat" and have Python3 code.

nltk/draw/dispersion.py
```diff
@@ -41,7 +41,7 @@ def dispersion_plot(text, words, ignore_case=False):
                     for y in range(len(words_to_comp))
                     if text_to_comp[x] == words_to_comp[y]]
     if points:
-        x, y = zip(*points)
+        x, y = list(zip(*points))
```
kmike, line 44: list() is not necessary here

nltk/app/chartparser_app.py
```diff
@@ -216,7 +217,7 @@ def remove_callback(self, event, func=None):
 
     def _fire_callbacks(self, event, *args):
         if event not in self._callbacks: return
-        for cb_func in self._callbacks[event].keys(): cb_func(*args)
+        for cb_func in list(self._callbacks[event].keys()): cb_func(*args)
```
kmike, line 220: I didn't check the code deeply, but why is `for cb_func in self.callbacks[event]:` not enough?

nltk/app/chunkparser_app.py
```diff
@@ -854,7 +855,7 @@ def show_help(self, tab):
         for (name, tabstops, text) in self.HELP:
             if name == tab:
                 text = text.replace('<<TAGSET>>', '\n'.join(
-                    ('\t%s\t%s' % item for item in sorted(self.tagset.items(),
+                    ('\t%s\t%s' % item for item in sorted(list(self.tagset.items()),
```
kmike, line 858: list() is not necessary here

nltk/app/collocations_app.py
```diff
@@ -284,7 +285,7 @@ def load_corpus(self, name):
 
     def non_default_corpora(self):
         copy = []
-        copy.extend(self.CORPORA.keys())
+        copy.extend(list(self.CORPORA.keys()))
```
kmike, line 288: list() is not necessary here

nltk/draw/table.py
```diff
@@ -365,7 +366,7 @@ def insert(self, index, *rows):
             if len(elt) != len(self._column_names):
                 raise ValueError('rows should be tuples whose length '
                                  'is equal to the number of columns')
-        for (lb,elts) in zip(self._listboxes, zip(*rows)):
+        for (lb,elts) in zip(self._listboxes, list(zip(*rows))):
```
kmike, line 369: list() is not necessary here

nltk/draw/table.py
```diff
@@ -298,8 +299,8 @@ def configure(self, cnf={}, **kw):
                 >>> mlb.configure(label_foreground='red')
                 >>> mlb.configure(listbox_foreground='red')
         """
-        cnf = dict(cnf.items() + kw.items())
-        for (key, val) in cnf.items():
+        cnf = dict(list(cnf.items()) + list(kw.items()))
+        for (key, val) in list(cnf.items()):
```
kmike, line 303: list() is not necessary

nltk/draw/tree.py
```diff
@@ -430,7 +431,7 @@ def tree_to_treesegment(canvas, t, make_node=TextWidget,
     leaf_attribs = {}
     loc_attribs = {}
 
-    for (key, value) in attribs.items():
+    for (key, value) in list(attribs.items()):
```
kmike, line 434: list() is not necessary

nltk/draw/tree.py
```diff
@@ -544,18 +545,18 @@ def bind_click_trees(self, callback, button=1):
...
     def bind_drag_trees(self, callback, button=1):
         """
         Add a binding to all tree segments.
         """
-        for tseg in self._expanded_trees.values():
+        for tseg in list(self._expanded_trees.values()):
             tseg.bind_drag(callback, button)
-        for tseg in self._collapsed_trees.values():
+        for tseg in list(self._collapsed_trees.values()):
             tseg.bind_drag(callback, button)
```
kmike, line 561: list() may be removed

kmike:
Generally, in Python 3.x list() should be applied to dct.keys(), dct.items() or dct.values() only if the "dct" is changed during iteration. Otherwise it is not necessary and may be harmful - it makes code more verbose, less efficient and harder to read (one can expect `for k in list(dct.keys()):` body to modify the "dct").

hleeldc:
The extra list() added done by 2to3.py. I wasn't paying too much attention to what it did.

stevenbird:
Thanks for the feedback @kmike. I'm not so keen on creating our own local names for tkinter methods. The solution that @hleeldc contributed will not require compat and its extra code once we drop support for Python 2. So the problems you list will all go away.

stevenbird:
Perhaps we should leave in the extra list() calls until we understand why 2to3.py added them.

PR #354, opened 2013-02-20
added files and fixed tox.ini for jenkins setup

hleeldc:
Added files needed for jenkins setup:
- `jenkins-job-config.xml` -- configuration file for a tox builder jeknins job. This can be read into jeknins to create a new job.
- `jenkins.sh` -- wrapper script that installs requirements, runs test, and generate coverage and pylint output.
- `pip-install.py` -- utility program for installing python packages.
- `pip-req.txt` -- list of python dependency packages.

Added two additional testenv sections for jenkins setup in the tox.ini file: `py26-jenkins` and `py32-jenkins`.


kmike:
@xim, what do you think?

xim:
Cool! Looks okay at a glance, with a few comments/questions off the top of my head:
- The return code of jenkins.sh.

Some failures should IMHO cause the build to fail. My solution to this was _not_ very elegant in the current Jenkins setup, but it works well enough; All commands are considered critical (i.e. bash -e) and some are "marked" optional (e.g. python -c "import nltk; nltk.download('all')" || echo "NLTK data download failed: $?")

I just like to see big, red failures if something is horribly wrong, but this isn't very important I guess.
- Multi-Python setup

I haven't seen or considered how this affects the representation – will the coverage/pylint reports be shown for only one python version, or both somehow? Is anything Shining Panda specific, on would it work on any Jenkins instance?
- What's changed/removed?

The existing setup has some more steps, exactly what is removed, and why? E.g. the magic sed, is the output from xunit now read correctly by cobertura without it?
- Dependencies

The existing installation does tons of crack for getting the Stanford tagger and Malt and whatnot running, is this handled anywhere here? I also had to custom compile a wxPython and put it in the LD_LIBRARY_PATH for some modules, is this not required anymore?

Will inspect the Shining Panda instance a bit later and look for answers to these and other questions in any case! Going away for a week or so from tomorrow, so may not be very available. But will take a look see and merge.

kmike:
re: multi-python setup 

If combined coverage reports are needed (are they?), then we can make use of coverage.py's ability of combining coverage reports ( http://nedbatchelder.com/code/coverage/cmd.html#cmd-combining ).

hleeldc:
Sorry for my slow response.

> The return code of jenkins.sh.
> ...
> I just like to see big, red failures if something is horribly wrong, but this isn't very important I guess.

I agree. Also, I agree that it's not very important at this stage, and probably it will have to evolve to fit our purpose as time goes.

> Multi-Python setup
> 
> I haven't seen or considered how this affects the representation – will the coverage/pylint reports be shown for only one python version, or both somehow? Is anything Shining Panda specific, on would it work on any Jenkins instance?

It's not Shining Panda specific, and should work with any Jenkins instance as long as it has required plugins.

I think it's best for you to look at the example. I set up a new builder job called "NLTK-TOX" with current Shining Panda account. Under the builder there sub tasks for each python version. Each version has its own reports. The top builder has a summary report too.

> What's changed/removed?
> 
> The existing setup has some more steps, exactly what is removed, and why? E.g. the magic sed, is the output from xunit now read correctly by cobertura without it?

I started from scratch although I learned a lot from existing setup. I didn't add anything if I didn't have to. The output seems to work without any modifications. But again, it will have to change over the course.

> Dependencies

I didn't account for any of non-python dependencies. There are more than a dozen tests that are skipped right now. The stuff like LD_LIBRARY_PATH can be added to jenkins.sh.

Ideally, the example setup (NLTK-TOX) and the original setup (NLTK-py2.6) should co-exist until NLTK-TOX stabilizes. Probably it will have to be revised many times. The good thing is that much of the configuration is in the repository and that makes it easier to get help from people.

stevenbird:
@kmike @xim -- what do you think about the NLTK-TOX entry at https://jenkins.shiningpanda.com/nltk/
Can we go ahead with this?

xim:
The pull request now has some needless merge commits. Did a cherry-pick to avoid them, added in 9dd4425.

stevenbird:
Thanks Morten.

stevenbird:
So, Haejoong now plans to: remove the "NLTK job", disable or remove the "NLTK-py26" job, and make the new "NLTK-TOX" job point at the master branch of NLTK. ok?

chaowu2009:
Error while after installed the nltk3 with python3 in ubuntu 12-04, 32 bit. 

## Then I was trying to download the other packages (in python3, import nltk; nltk.download(). The error message is 

Download which package (l=list; x=cancel)?
  Identifier> co
    Error loading co: Package 'co' not found in index

---

```
d) Download   l) List    u) Update   c) Config   h) Help   q) Quit
```

---

Downloader> l

Packages:
  File "nltk/downloader.py", line 645, in download
    self._interactive_download()
  File "nltk/downloader.py", line 961, in _interactive_download
    DownloaderShell(self).run()
  File "nltk/downloader.py", line 990, in run
    more_prompt=True)
  File "nltk/downloader.py", line 453, in list
    for info in sorted(getattr(self, category)()):
TypeError: unorderable types: Package() < Package()


Anybody can help? thanks. I am not sure where I should post the error message.

PR #356, opened 2013-02-26
Update nltk/tag/sequential.py

yanlend:
Remove need to pass train argument to BigramTagger.**init**() as in other NGramTagger-Classes


stevenbird:
Thanks for the fix.

PR #357, opened 2013-02-27
fix interface to megam maxent classifier!

alexrudnick:
Small change, but it makes megam work with Pythons 2 and 3.
- Instead of writing a gzipped training file, just write a regular text file. (megam is about to decompress it anyway) -- this avoids the hassle of writing binary files in the gzip library, which doesn't like unicode strings.
- Make sure, when we get the output of megam, that we turn it back into a standard string for the current Python.

### DIFF
--- a/nltk/classify/maxent.py
+++ b/nltk/classify/maxent.py
@@ -1361,8 +1361,8 @@ def train_maxent_classifier_with_megam(train_toks, trace=3, encoding=None,
 
     # Write a training file for megam.
     try:
-        fd, trainfile_name = tempfile.mkstemp(prefix='nltk-', suffix='.gz')
-        trainfile = gzip.open(trainfile_name, 'wb')
+        fd, trainfile_name = tempfile.mkstemp(prefix='nltk-')
+        trainfile = open(trainfile_name, 'w')
         write_megam_file(train_toks, encoding, trainfile, \
                             explicit=explicit, bernoulli=bernoulli)
         trainfile.close()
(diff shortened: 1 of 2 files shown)


alexrudnick:
As it stands, if you try to use the MaxentClassifier with algorithm='megam' on Python 3, you get an error like this:
TypeError: 'str' does not support the buffer interface. This fails without the patch, but works with it.

``` python
import nltk
from nltk.classify.maxent import MaxentClassifier
def document_features(document):
    return dict([('contains-word(%s)' % w, True) for w in document])

documents = [
    ("food jam apples bananas bagels peanut butter", "food"),
    ("butter steak beef", "food"),
    ("butter steak beef", "food"),
    ("hat trick space jam basketball play", "sports"),
    ("hat trick basketball hockey", "sports"),
    ("play sports espn", "sports")
]
nltk.classify.megam.config_megam(bin='/usr/local/bin/megam')
instances = []
for (text, label) in documents:
    feats = document_features(text.split())
    instances.append( (feats, label) )
feats = document_features("ham butter beef".split())
classifier = MaxentClassifier.train(instances, algorithm='megam')
```

stevenbird:
Thanks Alex. I guess the only benefit of compression would have been for very large training files, where its faster to compress it and write less to disk. But this might not be a high-frequency use case, and our first priority is functioning code.

alexrudnick:
Thanks! if this causes a problem, I'll turn the compression back on.

stevenbird:
I suppose we could add compression as a flag, if we figure out some other solution to the Python 2+3 problem, or once Python 2 goes away.

PR #358, opened 2013-02-28
Bugfix for downloader sorting

BrucePHill:
The downloader was crashing due to a call to sorted() whose arguments were not comparable. Using key=str fixed the problem for me and made sense as a way to sort things in the download tree.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -461,7 +461,7 @@ class Downloader(object):
         for category in categories:
             print('%s:' % category.capitalize())
             lines += 1 # for more_prompt
-            for info in sorted(getattr(self, category)()):
+            for info in sorted(getattr(self, category)(), key=str):
                 status = self.status(info, download_dir)
                 if status == self.INSTALLED and skip_installed: continue
                 if status == self.STALE: stale = True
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks for this fix. Did you mean to modify internals.py?

BrucePHill:
The change to internals.py was deliberate and part of the process by which I got the downloader to work. I think it's in two separate commits, but it's a simple one-line change to prevent object.**new**() from being passed extra parameters.

stevenbird:
Thanks. But I'm trying to understand why extra parameters in object.**new**() has not been a problem in the past. Is there some way that the changes in downloader.py broke internals.py?

BrucePHill:
In the latest version of nltk from the website, the downloader crashes due to an extra parameter being passed to object.**new**() in internals.py. I found two places where object.**new**() was being passed extra parameters. I believe one of these two locations was already fixed by someone else on the github repo. The change I've made here is to the other location, in order to prevent the same bug from occurring there. Perhaps it should have been part of a different pull request, but I made this change as part of the process to fix the downloader and I'm new to GitHub, so I wasn't sure of the proper procedure.

stevenbird:
OK, thanks for the further explanation.

PR #359, opened 2013-02-28
Bugfix for resource loading

BrucePHill:
Added encoding specifier to get resource loading to work.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -627,7 +627,7 @@ def load(resource_url, format='auto', cache=True, verbose=False,
     if format == 'raw':
         resource_val = opened_resource.read()
     elif format == 'pickle':
-        resource_val = pickle.load(opened_resource)
+        resource_val = pickle.load(opened_resource, encoding='iso-8859-1')
     elif format == 'yaml':
         import yaml
         resource_val = yaml.load(opened_resource)


stevenbird:
Unfortunately we do not have the code that was used for generating each of the trained models. Therefore it is more difficult to determine the character encodings in the files.

One way forward would be to default to whatever encoding would have been assumed under Python 2 NLTK, and allow this to be overridden manually.

kmike:
Another point: unpickling with proper encodings would fix the issue for Python 3.x, but not for 2.x. 

Under 2.x the issue is the following: if model is trained on data encoded to **X**, and then model is applied to data encoded to **Y**, then the model would use incorrect features and can produce invalid result. This may be hard to notice because often most of the text is ascii, and ascii text is usually encoded the same way in different encodings. So model may work for ascii text, but yield incorrect results for rare non-ascii symbols/tokens (like foreign words, umlauts or unicode punctuation); this results in lower quality, but doesn't raise an exception or somehow indicate the problem otherwise.

If model is trained on data encoded to **X** and applied to unicode text then _sometimes_ it would raise an exception, sometimes it would use sys.getdefaultencoding() or ascii as **Y** (and produce suboptimal results) - this depends on model implementation.

If we don't know **X** then we can't tell how should user encode data to get the correct results; if we know **X** and models are trained on byte strings then they would produce incorrect results for unicode data (obtained e.g. from corpora readers updated in NLTK 3) - users must pass data encoded to **X** in order to get optimal (correct) results.

That's why I think it is better to re-train models on unicode (maybe without actual retraining - we may just decode existing features using **X** if we can figure out what is **X**), even if this is more involved and, strictly speaking, backwards-incompatible.

stevenbird:
This is solid advice. The challenge is to recreate the trained models, requiring the right data and code. The first step, I think, is to audit the trained models try to estimate the scale of the required effort. What about posting this task as a challenge to the nltk-dev community, since we don't have the resources and its in the critical path for Python 3 NLTK.

stevenbird:
Unfortunately we can't accept this pull request, for the reasons discussed above. We do need help from the nltk-dev community to provide code to recreate the trained models, using the right character encodings...

stevenbird:
@kmike I think we should push ahead with a solution that works for Python 3.x. So why don't we try to determine the encoding for each pickled model, and specify this explicitly for correct unpickling? This can be done just for use in Python 3.x, so behaviour in 2.x is unchanged.

kmike:
I'm fine with that solution. It requires 2 things: a) determine the encoding used during training and b) implement support for setting unpickling encoding per model. As you said, it could be backwards-compatible and won't require changes to the models data.

The 2.x experience will be worse though. First, NLTK will still hide some bugs in user code under Python 2.x. Second, models trained on binary data don't work on unicode data correctly, and this could become a bigger problem in NLTK 3.0 because we switched corpus readers to return unicode.

I still think that the best solution would be to create an another index page for NLTK 3.0 downloads (the current page is out of sync, by the way), and to change models to use unicode (by re-training them or by just decoding their data and saving updated models).

After your proposed changes we'll know all the correct encodings - this is a big step forward, and it'll greatly simplify future changes. So if it easier for some reason, then let's do it, make NLTK 3 release and change models to use unicode later. 

For me it looks like determining correct encodings is the hard part, and all other steps are not that hard (pickling encoding suport, extra index page, updating models).

stevenbird:
In trying to convert the tagger pickle file "hmm_treebank_pos_tagger", I get the load error:
AttributeError: 'module' object has no attribute 'IdentityTransform'
We need to reinstate some code in hmm.py before this file can be converted.

stevenbird:
I have a Unicode version of the maxent_treebank_pos_tagger data, pickled using the Python 3 protocol. We need to decide if this sits inside the same nltk_data package (but with a different filename), or in a new package (alongside the existing one), or in a new nltk_data repository.

kmike:
Now I think that providing maxent_treebank_pos_tagger inside the same nltk_data package but with a different filename (and updated defaults for nltk.pos_tag) should be fine.

stevenbird:
It turns out that we have several files to modify. What do you think about creating a new repository (copying nltk_data's gh-pages branch), and having it downloaded to a location alongside nltk_data. E.g. on unix we would have the default of /usr/share/nltk_data and /usr/share/nltk_data3 (or some nicer name). The user code would not need to be modified (to use different files within the existing packages). Instead, their code would stay the same, and the NLTK data loader (data.py, downloader.py) can just be sensitive to Python 2 vs 3.

kmike:
Hmm, interesting. I like this idea, but I'm not sure the data path should be sensitive to Python 2 vs 3 - I thought it should be sensitive to NLTK 2 vs NLTK 3. 

The reason is that models should be updated for Python 2.x to be compatible with NLTK3, and this renders them incompatible with NLTK2, so with a single data path for NLTK2 and NLTK3 we could break NLTK < 3 code. 

There are cases when pickled models could be incompatible between Python 2.x and 3.x:

1) when some of the pickled functions/classes differ between Python 2.x and 3.x (I think this doesn't apply to functions/classes from stdlib because pickle handles this);
2) when Python 2.x's 'str' must be unpickled as 'str' for Python 2.x and as 'unicode' for Python 3.x (it will be unpickled as unicode in Python 2.x).
3) when 'bytes' are pickled using Python 2.x and unpickled using Python 3.x (they will be unpickled as unicode under Python 3.x);
4) when default 3 protocol is used for pickling under Python 3.x (the result won't be accessible in 2.x).

I think that (1) doesn't apply to NLTK because there is single codebase, to handle (3) we could pickle models using Python 3, and to handle (4) we could pickle them using protocol=2.

(2) should be rare, but it could become an issue. We could fix it by casting data to str() later, when it is necessary (usually when some function from Python 2.x stdlib is called that doesn't support unicode). 

I just tried to pickle/unpickle {b'bytes', u'unicode'} between Python 2.x. and 3.x. What worked properly is to pickle it in Python 3.x using protocol 2. I might have missing something, but now it looks like the way to go, at least for bytes/unicode issue. So I think it is posible to have pickled models that will work in both Python 2.x and 3.x and so there is no need to have Python 2.x / Python 3.x data paths.

stevenbird:
So we need to have separate data collections for NLTK 2 vs 3. My proposal to duplicate the whole nltk_data repository would work for this.

However, we have the opportunity to address one more important use case: suppose we want to upgrade any corpus or trained model after NLTK 3 is finalised. For instance, suppose WordNet changes its structure between versions, so we need to keep the old and new versions around (say wordnet_v1 and wordnet_v2), and NLTK 3.1 might require wordnet_v1 and NLTK 3.2 might require wordnet_v2.

We can solve this simply by adding version numbers to the corpus and model filenames, within the existing nltk_data repository.

There's a separate question about whether this versioning is hidden. E.g. would the corpus reader be called wordnet in both cases (and just read different data behind the scenes), or would there be a wordnet1 and wordnet2 corpus reader within a single version. I can see arguments both ways here.

stevenxxiu:
During unit testing of `nltk.test.unit.test_tag.test_basic` I've found another incompatibility due to pickling/unpickling of numpy's ndarray, which is used in the `english.pickle` file.

Here's an example demonstrating the issue:

In python 3:

```
import pickle
import numpy as np
arr=np.ndarray(shape=(2,2), dtype=float, order='F')
pickle.dump(arr,open('numpy.pickle','wb'),protocol=2)
```

In python 2:

```
import pickle
pickle.load(open('numpy.pickle','rb'))
```

I've dug deeper, and found out that the underlying cause is that `numpy.dtype.__setstate__` only accepts the second parameter when it's of type `str` in python 3, and `str` in python 2, but the pickle module pickles the latter as `unicode` when v2 is used. The pickle module appears to be correct in a compatibility sense but not numpy.

This appears to be unfixable, as the only way I can think of fixing this is to have a custom `Unpickler` class, which will slow down pickling as this means the c implementation of pickle cannot be used anymore. Monkey patching `numpy.dtype.__setstate__` is not possible as it's implemented in c.

So it appears that either that there is still a need for different sets of data for python 2 and 3, or we will need to use the slower python implementation of pickle.

The issue occurs in numpy 1.7.1.

kmike:
@jskda thanks for a good summary! I think this worths an issue in numpy tracker (is this is the code you're talking about? https://github.com/numpy/numpy/blob/704b456e49b13150aa693f4fd7d66dca0d541f0b/numpy/core/src/multiarray/methods.c#L1533). But in NLTK we need to support numpy < 1.7.1, so fixing it in numpy won't help us. 

Your Pickler class idea is interesting. Pickling speed doesn't matter because pickling will be done once; also, it seems that in Python 3 Pickler class could be subclassed without loosing C-based implementation. Anyways, what matters is the unpickling speed. 

Based on the code (I haven't checked it in console), it seems that dtype should accept byte strings in Python 3.x. So if all that custom Pickler does would be changing np.dtype arguments to bytes then it seems that unpickling should work without custom Unpickler.

That said, I'm not opposed to creating separate Python 2/3 models if Steven finds a way to support model versioning. 

@stevenbird - I haven't thought about what you've written about model versions yet. If adding version numbers is easier than changing repository path then there is nothing wrong with this approach.

kmike:
A side note: @jskda, could you please consider using feature branches for your pull requests? It could make merging and reviewing easier.

stevenxxiu:
> Your Pickler class idea is interesting. Pickling speed doesn't matter because pickling will be done once; also, it seems that in Python 3 Pickler class could be subclassed without loosing C-based implementation. Anyways, what matters is the unpickling speed.

Sorry, I meant `Unpickler`, the pickling speed doesn't matter, like you said as training isn't done frequently.

> Based on the code (I haven't checked it in console), it seems that dtype should accept byte strings in Python 3.x. So if all that custom Pickler does would be changing np.dtype arguments to bytes then it seems that unpickling should work without custom Unpickler.

The buggy method is `numpy.dtype.__setstate__`, which maps to `arraydescr_setstate`
https://github.com/numpy/numpy/blob/704b456e49b13150aa693f4fd7d66dca0d541f0b/numpy/core/src/multiarray/descriptor.c#LC2289, case 8
The code doesn't look like it's compatible between python 2 and 3
This is the error raised when `np.dtype.__setstate__` is called: `TypeError: must be a unicode character, not bytes`.

> A side note: @jskda, could you please consider using feature branches for your pull requests? It could make merging and reviewing easier.

No problems, I'll do that.

stevenbird:
I discussed the following general solution with @jskda, which he will implement for us to consider:

1) nltk.data.load() should know about the "version" of a dataset expected by the current version of NLTK. This can be stored as a dictionary inside data.py, e.g.:

VERSIONS = {'tokenizers/punkt/english.pickle": "tokenizers/punkt_2/english.pickle", ...}

2) nltk.data.load() should have an optional versioned=True flag, which when set to False does not do this name mapping.

3) the existing nltk_data collection is augmented with extra datasets over time, e.g. in this case, a new punkt_2.zip dataset; the nltk_data/collections/*.xml are updated to list all versions of a dataset (and the downloader code does not change)

4) old versions of NLTK without this new functionality are oblivious to the new versioned datasets, so there's nothing to do for them.

stevenbird:
There is a corresponding approach to corpus versioning that I would like to consider, cf the wordnet version example above. Its about as simple as it could be. At the moment, we do something like:

> > > from nltk.corpus import foo

This loads a corpus called foo, defined in nltk/corpus/**init**.py like this: foo = LazyCorpusLoader('foo', ...)

Suppose that the contents of foo is corrected but we don't want to change the behaviour of existing code. Or suppose that foo has a new interface, and we don't want to break existing code. So we can add:

foo = LazyCorpusLoader('foo_v2', ...)

If a third version of the data came out, we could update this, and provide access to the old one as well (and users would change their import: "from nltk.corpus import foo_v2 as foo")

foo = LazyCorpusLoader('foo_v3', ...)
foo_v2 = LazyCorpusLoader('foo_v2', ...)

Unfortunately, if you wanted to require the very first version, you would have to manually put the following at the top of your code: foo = LazyCorpusLoader('foo', ...)

We would also update the nltk_data collection, adding foo_v2.zip and foo_v3.zip and updating nltk_data/collections/*.xml

This approach (or something like it) permits us to roll out new versions of corpora at any time. What do people think?

stevenxxiu:
@stevenbird I have a question about your VERSIONS example, 

```
VERSIONS = {'tokenizers/punkt/english.pickle": "tokenizers/punkt_2/english.pickle", ...}
```

Should the punkt_2 directory contain all the necessary files for the punkt data set, including those unchanged from punkt?
If so, we could just use  `VERSIONS = {'tokenizers/punkt/','tokenizers/punkt_2/',...}` to map the directories instead.
Here's the code I currently have which implements this idea: https://github.com/jskda/nltk/commit/9ec325935c5dc55e28844d9a70dac21c846fb987

stevenbird:
@jskda yes, each new version would be complete, so mapping directories would work fine.

japerk:
I think corpus & model versioning is very needed, especially if it allows eventually standardizing the text encoding of all the corpora, and training new/better replacement models. It seems like the above proposal will do this with minimal breakage of old code.

kmike:
+1 for VERSIONS mapping for models. It looks simple enough. Maybe call it CURRENT_VERSIONS? 

What is the plan for models that are sensitive for Python 2/3? A very minor observation: having two CURRENT_VERSIONS dicts is not DRY.

```
if compat.PY3:
    CURRENT_VERSIONS = {
        # ...
        'some/other/model.pickle": "some/other/updated-model.pickle",
        'tokenizers/punkt/english.pickle": "tokenizers/punkt2014/english.pickle"
        # ...
    }
else:
    CURRENT_VERSIONS = {
        # ...
        'some/other/model.pickle": "some/other/updated-model.pickle",
        'tokenizers/punkt/english.pickle": "tokenizers/punkt2014_2x/english.pickle"
        # ...
    }
```

It is easy to forget to update one of the dicts, and info about "some/other/model" is duplicated.

What about this pattern (or something similar, like calling `if compat.PY3: CURRENT_VERSIONS.update(...)`)?

```
_PUNKT = "tokenizers/punkt2013_2x/english.pickle" if compat.PY3 else "tokenizers/punkt2013_2x/english.pickle"
CURRENT_VERSIONS = {
    'some/other/model.pickle": "some/other/updated-model.pickle",
    'tokenizers/punkt/english.pickle": _PUNKT,
    # ...
}
```

---

As for corpora, there are 2 use cases: "I want latest and greatest" and "I want reproducible results". Without "foo_v1" we can't have both. What do you think about the following?

```
foo_v3 = LazyCorpusLoader('foo_v3', ...)
foo_v2 = LazyCorpusLoader('foo_v2', ...)
foo_v1 = LazyCorpusLoader('foo', ...)
foo = foo_v3
```

This leads to some code duplication: for existing corpora there'll be redundant 'foo=foo_v1' assignments:

```
abc_v1 = LazyCorpusLoader(
    'abc', PlaintextCorpusReader, r'(?!\.).*\.txt', encoding=[
            ('science', 'latin_1'),
            ('rural', 'utf8')])
abc = abc_v1
```

But this way users can use "latest and greatest" foo_v1 and have "reproducible results". 

We can implement some magic to remove the need for `abc = abc_v1` lines, but I think it is unnecessary: `abc = abc_v1` lines are readable and explicit.

One more question: what is the plan for downloader GUI? Will it just list all models or you want to do some tricks (e.g. "show only current versions" flag)? This could be a separate issue of course.

@japerk It seems there is no need to standardize the encoding of corpora. Corpora readers in NLTK3 are aware of corpus encodings, and they all return unicode now (fire a bug if this is not the case). So there is no need for user to decode text or to even to know how is the original corpus data encoded. We already broke all user code here :)

stevenxxiu:
@kmike:

> What is the plan for models that are sensitive for Python 2/3

I propose to append string `_PY3`

> It is easy to forget to update one of the dicts, and info about "some/other/model" is duplicated.

Thanks for the suggestion, I've put this into my branch. In addition I'm using `sorteddict`, to map directories so that the mapped files can be loaded more efficiently.

Here's some magic I wrote up to handle the corpus case:

```
def add_corpus(name,versions):
    for key,val in versions.items():
        globals()['{}_v{}'.format(name,key)]=val
    globals()[name]=versions[max(versions.keys())]

add_corpus('abc',{
    1: LazyCorpusLoader(
        'abc', PlaintextCorpusReader, r'(?!\.).*\.txt', encoding=[
                ('science', 'latin_1'),
                ('rural', 'utf8')])
})
```

I think this looks nice and short too, do you think this would be appropriate?

kmike:
Good idea about _PY3. Maybe use _PY2 instead of _PY3?

As for the magic, as a developer of library I'd prefer some magic because it is less code. But as an user of library I'd prefer explicit abc_v1 and abc variables without any magic utilities, because it is easier to read code written this way, it enables IDE support (autocompletion, type inference) and it is just less concepts to worry about. So -0.5 from me for being too smart here.

stevenxxiu:
Thanks, I agree with both points.

Another issue is that since we have new package files, the downloader needs to unzip both the python 2 and python 3 versions of the packages if they require unzipping.

Here are some solutions that I have thought of
- create a new xml file for the python 3 package alongside the xml file for the python 2 package to specify that unzipping is required
- the downloader handles unzipping transparently by just looking at the xml file for python 2, and unzips both at once

I propose method 2 so we don't have to copy paste the old xml file and remember doing that every time.

heatherleaf:
Since Python 3 is supposed to be the default, it's better to treat py2-models as the exception. So, I agree with @kmike that old models should have _PY2 appended, but py3 models would have no suffix.

kmike:
.. but it seems that we can't have _PY2 appended to old models because it'll break backward compatibility with older NLTK versions (existing names should stay as-is). What we can have is _PY2 appended to new version of old models - I wonder if it'll be a common scenario.

heatherleaf:
So old models that are py2-specific must retain their names, which means that the corresponding py3-models must have _PY3 added to their names? 

I'm not sure I understand the suggestions completely. Could anyone give a description? What kind of model/corpus names will there be, and how should the shorcuts be handled?

heatherleaf:
We might want a better naming convention than _v1, _v2, _v3. If the original corpus already has a version, I guess we want to reuse it, such as wordnet 2.1, 3.0, 3.1. So, how should we name that?
- wordnet_v2.1
- wordnet_2.1
- wordnet_v_2_1

kmike:
Reusing original corpus versions is a good idea. 

I don't know which updated models and corpora are to be presented, so this naming discussion can be preliminary and abstract. Maybe @stevenbird and @jskda knows that better because they're updating the data and know all the details.

stevenxxiu:
> .. but it seems that we can't have _PY2 appended to old models because it'll break backward compatibility with older NLTK versions (existing names should stay as-is). What we can have is _PY2 appended to new version of old models - I wonder if it'll be a common scenario.

There are only 5 packages which require this extension, so I don't believe this is a common scenario. Your suggestion sounds good to me.
Here are the models which require updating:

```
tokenizers/punkt
taggers/maxent_treebank_pos_tagger
help/tagsets
corpora/brown
chunkers/maxent_ne_chunker
```

>    wordnet_v2.1
>    wordnet_2.1
>    wordnet_v_2_1

Here `.` is reserved in imports and we want to import the corpuses explicitly, so I think using that is a bad idea. I also think that reusing original corpus versions is a good idea, so either `wordnet_v_2_1` or `wordnet_v2_1` look good to me.

stevenbird:
I agree we should use original corpus version numbers, and prefer the format wordnet_v31.

+1 to @kmike 's suggestions for abc = abc_v1 and for the safer approach to defining CURRENT_VERSIONS.

I'd prefer to avoid _PY3 extensions to corpus names. A "3" suffix would be slightly less ugly.

An alternative to this last suggestion would keep user code clean at the expense of NLTK's internal transparency. We could update the contents of an existing corpus, by adding a new PY3 subdirectory with the required content. Then we just make the corpus reader check the Python version, and silently load the correct data. People would just have to know to re-run the downloader, and this automatically checks for updates. Once Python 2 support is dropped, we can again silently modify the corpus contents. This sidesteps the problem that @jskda mentioned for unzipping above. What do people think of this solution?

heatherleaf:
I think that's a good suggestion. Yes, it's slightly less transparent, but on the other hand it could happen in the future that we would want a PY36 or PYWIN subdirectory, and then we have the infrastructure for that.

stevenbird:
Thanks @heatherleaf 

A further implementation idea: if a corpus reader fails to find the expected PY3 subdirectory, it should produce an error message to tell the user: either use Python 2, or else use Python 3 and update that corpus by re-running the downloader.

stevenbird:
For each of the following data packages, there is now a PY3 subdirectory:
chunkers/maxent_ne_chunker.zip, help/tagsets.zip, taggers/maxent_treebank_pos_tagger.zip, tokenizers/punkt.zip

Data.py needs to add PY3/ to the path if we are using Python 3, and give an error if this path isn't found (tell the user to re-download the corpus).

stevenbird:
@kmike, @jskda: how about this approach? We can just add @py3_data to the initializers in data.py

stevenbird:
I've gone ahead and added the decorator -- please test.

stevenxxiu:
I think patching through the nltk protocol instead of using paths is better, since the user can possibly load a data file from from a path outside the nltk data packages, say for example his own version of "chunkers/maxent_ne_chunker", then the PY3 directory might not be present. If we do this through the nltk protocol then we can be sure the path points to a nltk data package.

stevenbird:
@jskda you are right, but I think the risk is small, and we will deal with this issue if it comes up in practice.

PR #360, opened 2013-02-28
Fix interface to the Stanford tagger so it works on Python 2 and 3

alexrudnick:
Previously, we would fail with file encoding issues on Python 3.

But now this little program works right for both, assuming you unzipped the Stanford Tagger into the same place that I did.

``` python
from __future__ import print_function
from nltk.tag.stanford import POSTagger

def get_tagger(enc=None):
    taggerhome = '/home/alex/software/stanford-postagger-2012-11-11'
    tagger = taggerhome + '/models/wsj-0-18-bidirectional-distsim.tagger'
    jar = taggerhome + '/stanford-postagger.jar'
    if enc:
        stanford_tagger = POSTagger(tagger, jar, encoding=enc)
    else:
        stanford_tagger = POSTagger(tagger, jar)
    return stanford_tagger

print("DEFAULT")
tagger = get_tagger()
print(tagger.tag("The man saw the dog with the telescope .".split()))
print("*" * 80)
print("UNICODE")
tagger = get_tagger(enc='utf-8')
print(tagger.tag("The man saw the dog with the telescope .".split()))
```

### DIFF
--- a/nltk/tag/stanford.py
+++ b/nltk/tag/stanford.py
@@ -36,7 +36,7 @@ class StanfordTagger(TaggerI):
     _SEPARATOR = ''
     _JAR = ''
 
-    def __init__(self, path_to_model, path_to_jar=None, encoding=None, verbose=False, java_options='-mx1000m'):
+    def __init__(self, path_to_model, path_to_jar=None, encoding='ascii', verbose=False, java_options='-mx1000m'):
 
         if not self._JAR:
             warnings.warn('The StanfordTagger class is not meant to be '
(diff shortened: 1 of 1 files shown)


heatherleaf:
I think your solution is over-complicated. Here's another suggestion:

In `__init__()`, line 51:

```
if compat.PY3 and not encoding:
    encoding = "ascii"
self._encoding = encoding
```

Then you can leave the rest of the file (i.e., `batch_tag()`) intact. 

I haven't tested this, but I have a good feeling about this :)

alexrudnick:
@heatherleaf Hey Peter!

That almost works, except that when we try to write to the non-binary output file, we get a type error on Python 3 when we try to write out bytes (it needs to be str).

Your proposal can be made to work by additionally changing the line with os.fdopen to:

``` python
_input_fh = os.fdopen(_input_fh, 'wb')
```

I want to say that it's better to be explicit about what the encoding for the training data is -- it has to be _something_, and it's OK to default to ASCII, but the current code pretends that the encoding might be "None", which is misleading.

I'd be OK with the smaller code change if you (or somebody else) would really prefer it, though!

alexrudnick:
That tiny patch does look nice, though. Maybe we can be more explicit about the encodings in another patch.

heatherleaf:
I agree with your point that the encoding should be explicit. So how about this:

```
def __init__(self, ..., encoding='ascii', ...):
    ...
    if compat.PY3:
        if not encoding: encoding = "ascii"
    else:
        if encoding == "ascii": encoding = None
```

I don't know which is best, and I don't have any strong preferences. So do whatever you think is best.

alexrudnick:
OK! I think this patch is both pretty clean and makes the encoding explicit.

@heatherleaf I'll go ahead and merge this. Let me know if there's any concerns! Thanks for the reviewing!

stevenbird:
Thanks @alexrudnick @heatherleaf

PR #369, opened 2013-03-15
Small tkinter-related fixes

kmike:
- Use lower-cased tkinter module name (uppercased may fail in Python 3.x on case-sensitive filesystems, and lower-cased tkinter it is fixed by nltk.compat for Python 2.x);
- imports in nltk.sem.drt_glue_demo are fixed (I didn't try the demo, but the code looks OK - I think it should work in 3.x as-is)

### DIFF
--- a/nltk/__init__.py
+++ b/nltk/__init__.py
@@ -155,7 +155,7 @@ else:
 
 from .downloader import download, download_shell
 try:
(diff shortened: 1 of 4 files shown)


alexrudnick:
Looks good, thanks!

PR #371, opened 2013-03-18
HMM improvements

kmike:
1. Faster HiddenMarkovModelTrainer.train_supervised (30-40x faster in demo_pos() example): HiddenMarkovModelTrainer._states and HiddenMarkovModelTrainer._symbols are lists so `if state not in self._states` and `if symbol not in self._symbols` are O(N). This is fixed by maintaining `known_symbols` and `known_states` sets.
2. more tests for HiddenMarkovModelTagger are added;
3. I think that probability calculations were incorrect for multiple observation sequences (they were not normalized);
4. when an initial model was passed to train_unsupervised the calculations were totally off because states and symbols of trainer and model was not the same - this is fixed; also, train_unsupervised didn't affect .tag if a model was used before (because of caching issues) - this is also fixed.

### DIFF
--- a/nltk/tag/hmm.py
+++ b/nltk/tag/hmm.py
@@ -280,7 +280,7 @@ class HiddenMarkovModelTagger(TaggerI):
 
     def _tag(self, unlabeled_sequence):
         path = self._best_path(unlabeled_sequence)
-        return list(zip(unlabeled_sequence, path))
+        return list(izip(unlabeled_sequence, path))
 
     def _output_logprob(self, state, symbol):
         """
@@ -315,9 +315,9 @@ class HiddenMarkovModelTagger(TaggerI):
         if not self._cache:
             N = len(self._states)
             M = len(self._symbols)
-            P = zeros(N, float32)
-            X = zeros((N, N), float32)
-            O = zeros((N, M), float32)
+            P = np.zeros(N, np.float32)
+            X = np.zeros((N, N), np.float32)
+            O = np.zeros((N, M), np.float32)
             for i in range(N):
                 si = self._states[i]
                 P[i] = self._priors.logprob(si)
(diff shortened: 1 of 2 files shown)


kmike:
It seems there is still an error somewhere because logprob still sometimes decreases during Baum-Welch iterations.

kmike:
I think this PR is ready for a review. 

Baum-Welch is still not very unusable (because it doesn't use model's estimator and because the implementation is impractically slow), but I think these changes should make further improvements easier.

nltk/tag/hmm.py
```diff
@@ -879,11 +878,18 @@ def train_unsupervised(self, unlabeled_sequences, **kwargs):
...
             model = HiddenMarkovModelTagger(self._symbols, self._states,
-                            transitions, output, priors)
+                            transitions, outputs, priors)
+
+        self._states = model._states
+        self._symbols = model._symbols
+
+        N = len(self._states)
+        M = len(self._symbols)
+        symbol_dict = dict((self._symbols[i], i) for i in range(M))
```
larsmans: Minor: `dict((sym, i) for i, sym in enumerate(self._symbols))` is clearer, IMHO.

kmike: Agree.

nltk/tag/hmm.py
```diff
@@ -83,13 +85,14 @@
...
-from nltk.util import LazyMap, LazyConcatenation, LazyZip
-from nltk.compat import python_2_unicode_compatible
-
+from nltk.util import LazyMap
+from nltk.compat import python_2_unicode_compatible, izip, imap
 from nltk.tag.api import TaggerI, HiddenMarkovModelTaggerTransformI
 
-# _NINF = float('-inf')  # won't work on Windows
-_NINF = float('-1e300')
+# won't work on Windows under Python 2.5, but we require Python 2.6
```
larsmans: Why not `-np.inf`?

kmike: To make less changes to hmm.py :) 

You're right, `-np.inf` looks better, I'll change _NINF to -np.inf

kmike: There is also nltk.probability._NINF, but nltk.probability doesn't require numpy so we should use float('-inf') there.

stevenbird:
Thanks @kmike for the work, and @larsmans for the helpful feedback. All looking good to me.

Issue #373, opened 2013-03-26
Maybe use non-uniform distribution for Baum-Welch by default?


kmike:
See https://github.com/nltk/nltk/issues/100. 

I encountered a suggestion for slightly non-uniform initial probability distributions (uniform + a bit of noize) in some paper (maybe about unsupervised POS tagging), but can't find the reference now. Most papers that employ Baum-Welch use either uniform distributions or distributions learned from  corpora using carefully crafted heuristics.

tyleha:
So as I understand it, and this is based on my advisor's [thesis](http://brl.ee.washington.edu/BRL_Pubs/Pdfs/Th037.pdf) (p.99 especially), a totally uniform distribution where each probable transition begins equivalent to every other results in a local minimum from which the BW algo will not escape. There is no probability gradient on which to update the matrices. Thus, some random jitter in the initialization is necessary. 

Ideally, yes - you would initialize the transition matrix with initial guesses based on learned distributions. However, most algorithms that require unlabeled training algorithms like BW do not have such heuristics in mind, but instead are willing to train a corpus of HMMs using randomly initiated probabilities and then select the HMM that shows best discrimination against some known dataset (this is how I use and evaluate HMMs). 

The most general solution to this problem is to assign the A and B (transition and output) matrices with a random set of probabilities where any value between 0 and 1 is equally probable (a uniform random distribution). Of course you could select a gaussian or a weighted or band-limited, but uniform is the most general. 

Seeing as I can't find much literature discussing the pros/cons of different types of random distributions, I'm happy to submit a pull request with a uniform random distribution function initializing the `HiddenMarkovModelTagger`'s `_transitions` and `_outputs`.

PR #375, opened 2013-03-26
train_unsupervised modification in initial probability matrix distributions

tyleha:
In response to issue #373, have created a new probability class in probability.py called RandomProbDist, which initializes a uniform random distribution instead of a flat, equivalent distribution (i.e. all elements in prob matrix are equal). 

With a flat distribution, Baum-Welch algo will not update as it finds itself in a local minimum. With the default distributions now randomized, the algo will perform as expected, optimizing the `_transitions` and `_outputs` matrices and gradually decreasing in total logprob until tolerance is reached.

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -650,6 +650,7 @@ class RandomProbDist(ProbDistI):
         self._probs = self.unirand(samples)
         self._samples = list(self._probs.keys())
 
+    @classmethod
     def unirand(self, samples):
         """
         The key function that creates a randomized initial distribution 
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -651,7 +651,7 @@ class RandomProbDist(ProbDistI):
         self._samples = list(self._probs.keys())
 
     @classmethod
-    def unirand(self, samples):
+    def unirand(cls, samples):
         """
         The key function that creates a randomized initial distribution 
         that still sums to 1. Set as a dictionary of prob values so that 
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -648,7 +648,7 @@ class RandomProbDist(ProbDistI):
             raise ValueError('A probability distribution must '+
                              'have at least one sample.')
         self._probs = self.unirand(samples)
-        self._samples = list(self._sampleset)
+        self._samples = list(self._probs.keys())
 
     def unirand(self, samples):
         """
@@ -662,10 +662,11 @@ class RandomProbDist(ProbDistI):
         for i, x in enumerate(randrow):
             randrow[i] = x/total
 
-        if sum(randrow) != 1:
+        total = sum(randrow)
+        if total != 1:
             #this difference, if present, is so small (near NINF) that it 
             #can be subtracted from any element without risking probs not (0 1)
-            randrow[-1] -= sum(randrow) - 1
+            randrow[-1] -= total - 1
 
         return dict((s, randrow[i]) for i, s in enumerate(samples))
 
@@ -676,7 +677,7 @@ class RandomProbDist(ProbDistI):
         return self._samples
 
     def __repr__(self):
-        return '<RandomUniformProbDist with %d samples>' %len(self._sampleset)
+        return '<RandomUniformProbDist with %d samples>' %len(self._probs)
 
 
 @compat.python_2_unicode_compatible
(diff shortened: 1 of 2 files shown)


tyleha:
Hmm.. Excellent point. Also, do we need UniformProbDist in probability.py? A quick search shows it isn't called anywhere but in train_unsupervised in all of NLTK...

kmike:
I think we should keep UniformProbDist because it could be used in user code, and because it is a well-known probability distribution.

nltk/probability.py
```diff
@@ -636,6 +636,48 @@ def samples(self):
...
+    Generates a random probability distribution whereby each sample
+    will be between 0 and 1 with equal probability (uniform random distribution.
+    Also called a continuous uniform distribution).
+    """
+    def __init__(self, samples):
+        if len(samples) == 0: 
+            raise ValueError('A probability distribution must '+
+                             'have at least one sample.')
+        self._probs = self.unirand(samples)
+        self._samples = list(self._sampleset)
```
kmike: This could be list(s for s in self._probs) or list(self._probs.keys()); "list" is for Python 3.x

nltk/probability.py
```diff
@@ -636,6 +636,48 @@ def samples(self):
...
+        return dict((s, randrow[i]) for i, s in enumerate(samples))
+
+    def prob(self, sample):
+        return self._probs.get(sample, 0)
+
+    def samples(self):
+        return self._samples
+
+    def __repr__(self):
+        return '<RandomUniformProbDist with %d samples>' %len(self._sampleset)
```
kmike: len(self._probs) or len(self._samples)

nltk/probability.py
```diff
@@ -636,6 +636,48 @@ def samples(self):
...
+        that still sums to 1. Set as a dictionary of prob values so that 
+        it can still be passed to MutableProbDist and called with identical
+        syntax to UniformProbDist
+        """
+        randrow = [random.random() for i in range(len(samples))]
+        total = sum(randrow)
+        for i, x in enumerate(randrow):
+            randrow[i] = x/total
+
+        if sum(randrow) != 1:
```
kmike: It seems that it is quite likely that sum(randrow) != 1; maybe avoid computing it twice?

kmike:
@tylerhartley Thanks for your work! All looks good to me. @stevenbird ?

nltk/probability.py
```diff
@@ -636,6 +636,49 @@ def samples(self):
...
+    Also called a continuous uniform distribution).
+    """
+    def __init__(self, samples):
+        if len(samples) == 0: 
+            raise ValueError('A probability distribution must '+
+                             'have at least one sample.')
+        self._probs = self.unirand(samples)
+        self._samples = list(self._probs.keys())
+
+    def unirand(self, samples):
```
kmike: It may be better to make unirand() classmethod: this way it will be possible to use RandomProbDist.unirand() without creating a dummy RandomProbDist instance.

tyleha: Seems reasonable to me. Could be useful in other applications

kmike:
There is a subtle issue with RandomProbDist: if there are duplicate items in samples (passed either to constructor or to unirand classmethod), probabilities won't sum up to 1.

UniformProbDist deals with this by deduplicating samples (casting them to set); we should probably do the same (both in constructor and in unirand).

Another way is to sum probabilities for duplicate samples so that "each sample will be between 0 and 1 with equal probability" would hold (assuming "sample" is an "object from samples", not "object + its position in samples") - but I think we should just deduplicate samples.

stevenbird:
Thanks @tylerhartley and @kmike. This looks good and I'm happy for it to be merged.

Note that the non-deterministic behaviour is a problem for any documentation. What do you think about adding an optional "seed" argument to the constructor? Then docs could use, say, seed=0 and get the same result each time it is used.

kmike:
+1.

This doesn't matter much here, but I think we should prefer passing "random" instances instead of "seed" because in stdlib "random" module some high-level functions (e.g. random.shuffle) return different results in Python >= 3.3 and Python < 3.3 even given the same seed.

"random" instances would allow users to use numpy.random or numpy.random.RandomState or https://pypi.python.org/pypi/random2  which don't have this issue.

stevenbird:
Thanks for suggesting a better approach. I'll merge this request now, but invite @tylerhartley to submit a patch which follows this suggestion.

PR #377, opened 2013-03-30
Faster Baum-Welch

kmike:
In this pull request most calculations required for HiddenMarkovModelTrainer.train_unsupervised are vectorized; this makes train_unsupervised about 30x faster (number is for demo_pos_bw example).

### DIFF
--- a/nltk/tag/hmm.py
+++ b/nltk/tag/hmm.py
@@ -793,10 +793,10 @@ class HiddenMarkovModelTagger(TaggerI):
                     ' '.join('%s/%s' % (token, tag)
                               for (token, tag) in predicted_sent))
                 print()
-                # print('Entropy:',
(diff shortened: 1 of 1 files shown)


kmike:
Initially I was going to introduce a Cython extension (with optimized logaddexp2 and _log_add functions); this extension could make Baum-Welch about 1.5x faster than in this pull request. I decided not to do this now (for simplicity) - 1.5x speed difference is not that critical.

kmike:
There are ways to improve this further in future (e.g. some matrices are re-calculated many times in this pull request).

kmike:
Cython extension code (unused in the final result): https://gist.github.com/kmike/5278404

stevenbird:
Thanks for a great contribution.

PR #378, opened 2013-03-30
porter.py changed check for double consonant from dictionary to list

knkumar:
unable to import due to syntax error - changed check for doubledec from dictionary to list


knkumar:
changed porter.py to doubledec a word correctly. I am hoping other things are not broken here, request a check and pull

kmike:
The constant is not dictionary, it is a set literal that is available since Python 2.7, but NLTK supports Python 2.6.

We should probably use `word.endswith(('l', 's', 'z'))` (and similar statements in the code before & after) or make these lists sets and precompute them. But this is a bigger task; this change looks good as-it. Thanks!

Issue #381, opened 2013-04-08
Remove nltk.probability.GoodTuringProbDist


kmike:
What do you think about removing nltk.probability.GoodTuringProbDist? I'm asking because: 

a) GoodTuring is not useful in practice (especially for high-frequency symbols where Nr(count+1) is often zero and thus probability is zero);
b) if it is for educational purposes, than arguably implementing it by yourself provides more educational value;
c) nltk.probability.GoodTuringProbDist looks broken: e.g. GoodTuringProbDist.prob() method contain a branch for unseen samples that does nothing (it just computes p0 variable which is not used later).

Another way is to fix (a) by mentioning Good-Turing issues in docstring and to fix (c) by fixing GoodTuringProbDist.prob() and adding some tests, but I think it is better to just remove GoodTuringProbDist - it is much easier and it would fix (a), (b) and (c).

kmike:
GoodTuringProbDist could be useful for FreqDist subclasses that implement smoothed Nr() method, but this is not obvious and there is no such FreqDist subclass in NLTK.

heatherleaf:
+1

Since we do have SimpleGoodTuringProbDist, which works, I don't think there's a need to keep this one.  There's already a good explanation in the comments, and we could simply move the class docstring for GoodTuring into the comments (e.g., efter the fist paragraph, on line 1294.

stevenbird:
+1

afourney:
+1

kmike:
Thanks for your support :)

PR #382, opened 2013-04-08
Remove GoodTuringProbDist (see #381)

kmike:
As @heatherleaf suggested, GoodTuringProbDist docstring is moved to a comment. I changed it a bit, please verify. 

Also, the default value for SimpleGoodTuringProbDist.bins is quite unuseful IMHO: freqdist.B() + 1 means that probability for unseen samples could be very large (up to several orders of magnitude larger than probability of the most common sample). Maybe there is a better estimate (e.g. could we somehow take Zipf law into account)?

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -2306,7 +2306,6 @@ def demo(numsamples=6, numoutcomes=500):
         HeldoutProbDist(fdist1, fdist2, numsamples),
         HeldoutProbDist(fdist2, fdist1, numsamples),
         CrossValidationProbDist([fdist1, fdist2, fdist3], numsamples),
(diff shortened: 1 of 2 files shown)


heatherleaf:
looks good to me

re. the default value for SGT, shouldn't that be another issue?

kmike:
Yes, this shold be an another issue. I mentioned it because it is related to a commit from this pull request (https://github.com/kmike/nltk/commit/8d20dc86d8180e61a83e5dc2deeb5b5f30e6071f).

kmike:
Opened https://github.com/nltk/nltk/issues/383 for 'bins' default value.

Issue #383, opened 2013-04-08
Default value for 'bins' argument in SimpleGoodTuringProbDist and WittenBellProbDist


kmike:
I think the default value for SimpleGoodTuringProbDist.bins is quite unuseful: freqdist.B() + 1 means that probability for unseen samples could be very large (up to several orders of magnitude larger than probability of the most common sample). This caused low performance of SimpleGoodTuringProbDist in NLTK tests: see  https://github.com/kmike/nltk/commit/8d20dc86d8180e61a83e5dc2deeb5b5f30e6071f

WittenBellProbDist also has a non-useful default 'bins' value (see https://github.com/nltk/nltk/issues/110).

Maybe there is a better estimate (e.g. could we somehow take Zipf law into account)?

kmike:
Let bins == fd.B()+1.

Then prob() method would return p = _prob_measure(0) / (1.0 \* self._bins - self._freqdist.B()) == _prob_measure(0), where _prob_measure(0) is 1.0 \* self._freqdist.Nr(1) / self._freqdist.N() - a ratio of samples with frequency 1.

A reasonable guess for self._freqdist.Nr(1) / self._freqdist.N() is 0.05. 

What does it mean: each unseen token gets 0.05 probability; this number is huge - frequent words usually have much lower probability.

desilinguist:
The value of the number of bins was from the original SGT C code reference
implementations. See the issue on google code for more detail.

It may give too high a probability mass to unseen tokens - that's why it's
called SimpleGoodTuring.

On Monday, April 8, 2013, Mikhail Korobov wrote:

> Let bins == fd.B()+1.
> 
> Then prob() method would return p = _prob_measure(0) / (1.0 \* self._bins -
> self._freqdist.B()) == _prob_measure(0), where _prob_measure(0) is 1.0 *
> self._freqdist.Nr(1) / self._freqdist.N() - a ratio of samples with
> frequency 1.
> 
> A reasonable guess for self._freqdist.Nr(1) / self._freqdist.N() is 0.05.
> 
> What does it mean: each unseen token gets 0.05 probability; this number is
> huge - frequent words usually have much lower probability.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/nltk/nltk/issues/383#issuecomment-16073707
> .

## 

Linguist, Desi Linguist
http://www.desilinguist.org

afourney:
Right. 

[...]

PR #384, opened 2013-04-08
More efficient FreqDist._Nr_cache

kmike:
FreqDist._Nr_cache is very sparse in real word. E.g. for Brown corpus there is only 553 non-zero values and 62160 zero values; for bigger corpora issue is worse, e.g. Russian Wikipedia has < 10K non-zero values and > 11M zero values.

Storing numbers in Python list is also quite inefficient: list of integers is an array of pointers to integers (and each integer is a Python object with a pointer to type and a refcount number). 

This change saves me about 100M of memory for FreqDist with unigrams from Russian Wikipedia (and a full _Nr_cache - SimpleGoodTuringProbDist fills it).

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -211,14 +211,14 @@ class FreqDist(dict):
         if self._Nr_cache is None:
             self._cache_Nr_values()
 
-        return (self._Nr_cache[r] if r < len(self._Nr_cache) else 0)
+        return self._Nr_cache.get(r, 0)
 
     def _cache_Nr_values(self):
-        Nr = [0]
+        Nr = defaultdict(int)
         for sample in self:
             c = self.get(sample, 0)
-            if c >= len(Nr):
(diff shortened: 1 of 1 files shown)


kmike:
New _Nr_cache implementation also enables much faster SimpleGoodTuringProbDist._r_Nr: 6s vs 18s for creating SimpleGoodTuringProbDist from FreqDist, and 50ms vs 12s for creating SimpleGoodTuringProbDist from FreqDist with primed cache.

stevenbird:
This is a much better implementation, thanks!

stevenbird:
It would be ideal to have a couple of regression tests as well, if you can manage it, but I'm happy to merge anyway.

kmike:
This is tested indirectly in probability.doctest, but you're right, some unit test would help. I was going to write some tests as a part of #383.

PR #389, opened 2013-04-16
add a __reduce__ method to ConditionalFreqDist so that it's pickle-able

alexrudnick:
At the moment, ConditionalFreqDist isn't pickle-able, but this patch makes it work with pickle.

This little script breaks when trying to unpickle on the current master, but works with the patch.

``` python
import nltk
import pickle

cfd = nltk.probability.ConditionalFreqDist()
cfd['foo'].inc('hello')
with open("cfd.pickle", "wb") as outfile:
    pickle.dump(cfd, outfile)
with open("cfd.pickle", "rb") as infile:
    cfd2 = pickle.load(infile)
```

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -1854,6 +1854,13 @@ class ConditionalFreqDist(defaultdict):
             for (cond, sample) in cond_samples:
                 self[cond].inc(sample)
 
+    def __reduce__(self):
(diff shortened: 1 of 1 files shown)


kmike:
Thanks!

alexrudnick:
Yay, thanks!

PR #392, opened 2013-04-18
ConditionalFreqDist should remember how many times each sample happened, not just that it happened

alexrudnick:
Sorry, #389 I only stored each condition and each sample that happened with it. But we should have been storing the FreqDist objects themselves, because they store the _count_ of the individual samples.

(the `__init__` method on ConditionalFreqDist isn't all that useful...)

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -1855,11 +1855,8 @@ class ConditionalFreqDist(defaultdict):
                 self[cond].inc(sample)
 
     def __reduce__(self):
-        cond_samples = []
-        for condition in self.conditions():
-            cond_samples.extend([(condition, sample)
-                                 for sample in self[condition].samples()])
-        return (self.__class__, (cond_samples,))
+        kv_pairs = [(cond, self[cond]) for cond in self.conditions()]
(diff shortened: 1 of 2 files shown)


nltk/probability.py
```diff
@@ -1855,11 +1855,8 @@ def __init__(self, cond_samples=None):
                 self[cond].inc(sample)
 
     def __reduce__(self):
-        cond_samples = []
-        for condition in self.conditions():
-            cond_samples.extend([(condition, sample)
-                                 for sample in self[condition].samples()])
-        return (self.__class__, (cond_samples,))
+        kv_pairs = [(cond, self[cond]) for cond in self.conditions()]
+        return (self.__class__, (), None, None, iter(kv_pairs))
```
kmike: I didn't check it, but what about using generator expression?

```
kv_pairs = ((cond, self[cond]) for cond in self.conditions())
return (self.__class__, (), None, None, kv_pairs)
```

This way we could avoid building an intermediate list, and iter() won't be necessary.

alexrudnick:
Good idea, yes! Thanks for taking a look.

kmike:
@alexrudnick, thanks for the fix!

PR #394, opened 2013-04-22
Make ConditionalProbDist pickle-able by adding a __missing__ method

### LINKED ISSUE #393: can't pickle NgramModels because we can't pickle ConditionalFreqDists because Python can't pickle lambdas ...
Right now, we can't pickle an NgramModel, like so:

``` python
lm = nltk.model.ngram.NgramModel(3, ['a', 'b', 'c'])
pickle.dumps(lm)
```

... but I know why this is! We can't pickle any ConditionalFreqDist, and the NgramModel has one of those as a member.

``` python
from nltk.probability import ConditionalProbDist
from nltk.probability import ConditionalFreqDist
from nltk.probability import SimpleGoodTuringProbDist
cfd = ConditionalFreqDist()
cfd['hello'].inc('everybody')
cfd['hello'].inc('everybody')
model = ConditionalProbDist(cfd, SimpleGoodTuringProbDist)
pickle.dumps(model)
```

And this can't be done because ConditionalProbDist carries a lambda around.
https://github.com/nltk/nltk/blob/master/nltk/probability.py#L2091

Any thoughts on the best way to fix this, so we can pickle NgramModels again? I'll take a stab at it...

alexrudnick:
This should fix #393. Now the ConditionalProbDist objects don't have to hang on to a lambda.

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -2087,15 +2087,11 @@ class ConditionalProbDist(ConditionalProbDistI):
         self._factory_args = factory_args
         self._factory_kw_args = factory_kw_args
 
-        ## factory = lambda: probdist_factory(FreqDist(),
-        ##                                   *factory_args, **factory_kw_args)
-        ## defaultdict.__init__(self, factory)
         for condition in cfdist:
             self[condition] = probdist_factory(cfdist[condition],
                                                *factory_args, **factory_kw_args)
 
     def __missing__(self, key):
-        print("YEEEAAAHHH!")
         self[key] = self._probdist_factory(FreqDist(),
                                            *self._factory_args,
                                            **self._factory_kw_args)
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -2083,18 +2083,23 @@ class ConditionalProbDist(ConditionalProbDistI):
         :type factory_kw_args: (any)
         :param factory_kw_args: Extra keyword arguments for ``probdist_factory``.
         """
-        # self._probdist_factory = probdist_factory
-        # self._cfdist = cfdist
-        # self._factory_args = factory_args
-        # self._factory_kw_args = factory_kw_args
+        self._probdist_factory = probdist_factory
+        self._factory_args = factory_args
+        self._factory_kw_args = factory_kw_args
 
-        factory = lambda: probdist_factory(FreqDist(),
-                                           *factory_args, **factory_kw_args)
-        defaultdict.__init__(self, factory)
+        ## factory = lambda: probdist_factory(FreqDist(),
+        ##                                   *factory_args, **factory_kw_args)
+        ## defaultdict.__init__(self, factory)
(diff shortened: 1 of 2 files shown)


kmike:
After this change it doesn't make sense to inherit from defaultdict because
- useful behavior (`__missing__` method) is overriden, and 
- public defaultdict API (default_factory attribute) is no longer supported because constructor is not called.

But if we set default_factory then pickling will be broken again. 

Creating `__reduce__` method (instead of overriding `__missing__`) will enable default_factory attribute (and thus defaultdict API). But I don't understand right now how to deal with **factory_kw_args in this case.

It was @heatherleaf who made CPD a defaultdict subclass - Peter, maybe you have some ideas about this ticket?

alexrudnick:
Thanks for the thoughts, @kmike! That sounds sensible.

Trying it out with `__reduce__` ...

alexrudnick:
Tried it with `__reduce__`, and it's really not clear what to do with the kwargs, you're right. You can't easily fit the named arguments into the tuple that `__reduce__` has to return.

What do we get by having ConditionalProbDist be a subclass of defaultdict? It does make the code shorter and probably faster...

And how bad would it be if we just document that the `default_factory` attribute, for this particular subclass, isn't how new entries get created?

heatherleaf:
Oh, I didn't remember that, so I had to search through the history and found these two: 7ab305d12f379cf4e3cb7720ffb214022f1b620a, 0b740f45bb4f600205a1dd46c76505e937a45f45

It seems that by subclassing it from defauldict, I could remove 20-30 lines of code. Not a big deal -- I'm perfectly ok with removing the defaultdict dependency.

heatherleaf:
I found some uninteresting statistics for the two commits: they added 43 lines and removed 147 lines of code.

kmike:
@alexrudnick , @heatherleaf - what about making ConditionalProbDist a subsclass of dict instead of defaultdict?  `__missing__` works for dict, and dict doesn't have default_factory in API. I think it won't worsen @heatherleaf's cleanup.

heatherleaf:
+1

alexrudnick:
@kmike, @heatherleaf  Oooh yeah! Good idea! (you mean ConditionalProbDistI, right? )

I'll do that. Patch forthcoming!

alexrudnick:
OK, I think this works! Now ConditionalProbDistI is a subclass of dict, and its subclasses have been changed accordingly. Also added a test.

@kmike , @heatherleaf -- what do you think?

heatherleaf:
Looking good, as far as I can tell

kmike:
Thanks @alexrudnick for idea & the fix and @heatherleaf for a review :+1: 

I didn't appreciate the importance of pickling until I start to work with multiprocessing :)

alexrudnick:
Great, thanks for the ideas and reviewing! :+1:

PR #399, opened 2013-05-03
make nltk.tag.simplify_brown_tag not map -- to the empty string

alexrudnick:
There are some tokens in the Brown corpus with the tag `--`, and this tag gets mapped to the empty string by nltk.tag.simplify_brown_tag.

``` python
## currently gives you empty string
nltk.tag.simplify_brown_tag("--")
```

But that shouldn't be! It's getting into the branch where it checks whether there's a dash in the tag. This patch avoids that branch by just adding `--` into the Brown tag map.


stevenbird:
Thanks Alex.

PR #400, opened 2013-05-04
Correct tnt.train() initialization (fix #398)

### LINKED ISSUE #398: Can't train a TnT tagger in Python 3
If you do:

``` python
from nltk.tag import tnt
tnt.demo()
```

... then it fails during training with an error like this, at least in Python 3:

```
  File "/home/alex/nltk/nltk/tag/tnt.py", line 484, in demo
    tagger.train(sents[200:1000])
  File "/home/alex/nltk/nltk/tag/tnt.py", line 166, in train
    self._compute_lambda()
  File "/home/alex/nltk/nltk/tag/tnt.py", line 199, in _compute_lambda
    for history in self._tri.conditions():
  File "/home/alex/nltk/nltk/probability.py", line 1871, in conditions
    return sorted(self.keys())
TypeError: unorderable types: str() < tuple()
```

Python 2.7 works for me, though.

rmalouf:
In train(), history should be a list of string/boolean pairs but it gets initialized to a list of strings.  This was okay (but still not right) in python2, but causes an exception in python3.

### DIFF
--- a/nltk/tag/tnt.py
+++ b/nltk/tag/tnt.py
@@ -140,7 +140,7 @@ class TnT(TaggerI):
             self._unk.train(data)
 
         for sent in data:
-            history = ['BOS', 'BOS']
+            history = [('BOS',False), ('BOS',False)]
             for w, t in sent:
 
                 # if capitalization is requested,


alexrudnick:
Great! Thanks, Rob!

PR #402, opened 2013-05-07
fix file descriptor leak when writing to megam training file

alexrudnick:
Currently if you train a lot of megam classifiers, you'll eventually run out of file handles!

This is because we weren't closing a file handle when we do a tempfile.mkstemp. This patch fixes that.


alexrudnick:
I'm just going to go ahead and merge this one, because it's definitely a bug and I'm pretty sure this is the right fix.

Let me know if this is a problem!

PR #404, opened 2013-05-10
Move all tkinter calls into main thread (fix #82)

### LINKED ISSUE #82: Loading corpora in concordance and collocations tools
<b>What steps will reproduce the problem?</b>

&gt;&gt;&gt; nltk.app.concordance()

and

&gt;&gt;&gt; nltk.app.collocations()

<b>What is the expected output? What do you see instead?</b>

out of stack space (infinite loop?)
Exception in thread English: Brown Corpus (Learned, simplified):
  File
&quot;/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/python2.6/threading.py&quot;,
line 525, in __bootstrap_inner
    self.run()
...
TclError: out of stack space (infinite loop?)

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=445</p><hr/><h2>earlier comments</h2>
<p><strong>sumukh.ghodke said, at 2009-09-19T02:06:59.000Z:</strong></p>

<p>I tested the latest nltk build on ubuntu 9.04 (python 2.6.2; Tcl/Tk 8.5.6) and Mac
OsX 10.5.8 (python 2.5.2; Tcl/Tk 8.4.7), but didn't encounter the above error.

This problem seems to happen during Tkinter initialisation from python code. Some say
that it occurs when Tcl/Tk is built without a parameter called --enable-threads.
[ref:http://bugs.python.org/issue3835] Can you please check if this happens even
after you recompile Tcl/Tk?

Some background on why threads are used in the concordance app:
The concordance app uses a Model-View architecture. the view is loaded using python's
mainloop thread, whereas data loading and search operations take place in independent
threads. In order to make sure that gui rendering always happen in the mainloop
thread I fire events from the data loading threads. The mainloop thread should be
informed of those events and accordingly render results.
This bug wouldn't have occurred if it was single threaded, but then the app would
have been less responsive too. If really required we could make it run in a single
thread.

  </p>

rmalouf:
The collocations and concordance app fail on systems where tcl/tk isn't compiled with thread support (issue #82).  The fix is to move all of the calls to tkinter into the main thread and using a queue rather than tkinter events for communication between sub-threads and the main thread.

### DIFF
--- a/nltk/app/collocations_app.py
+++ b/nltk/app/collocations_app.py
@@ -10,6 +10,10 @@
 import nltk.compat
 import threading
 import tkinter.font
+if nltk.compat.PY3:
+    import queue as q
+else:    
+    import Queue as q
 from tkinter import (Button, END, Frame, IntVar, LEFT, Label, Menu,
                      OptionMenu, SUNKEN, Scrollbar, StringVar,
                      Text, Tk)
@@ -22,6 +26,7 @@ from nltk.probability import FreqDist
 
 CORPUS_LOADED_EVENT = '<<CL_EVENT>>'
 ERROR_LOADING_CORPUS_EVENT = '<<ELC_EVENT>>'
+POLL_INTERVAL = 100
 
 _DEFAULT = 'English: Brown Corpus (Humor)'
 _CORPORA = {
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks Rob. I don't have an easy way to test this, so I'm going to merge anyway.

PR #405, opened 2013-05-12
tnt tagger probability multiplication change to log addition

isnowfy:
tnt tagger probability multiplication change to log addition to get more precision

### DIFF
--- a/nltk/tag/tnt.py
+++ b/nltk/tag/tnt.py
@@ -59,6 +59,12 @@ class TnT(TaggerI):
     from the training data. It is the set of all tags
     that exact word has been assigned.
 
+    To speed up and get more precision, we can use log addition
+    to instead multiplication, specifically:
+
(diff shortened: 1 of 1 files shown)


kmike:
I think this is a change in good direction, but it will require some work to get merged. E.g. variables with log probability shouldn't be named "prob" ("logprob" is better); docstrings should be also updated, e.g. this is no longer correct:

```
        :param current_states : List of possible tag combinations for
                                the sentence so far, and the probability
                                associated with each tag combination

```

isnowfy:
@kmike Thanks for your advices and I will change it

kmike:
Thanks! One more point: in other NLTK parts (e.g. `nltk.probability.ProbDistI.logprob(sample)`) base 2 logarithm is used instead of natural logarithm; I think it is better to be consistent here.

isnowfy:
@kmike Thanks for your reminder, I have changed the variable, docstrings and used base 2 log instead of natural log.

kmike:
Thanks!

PR #407, opened 2013-05-22
Fixes to internals.py to accomodate windows users

dougalg:
- Edited to fix a regex comment error from lines 58-64
- Added java.exe to binary names for java (to support windows)
- Modified find_file to:
  1) split the environment paths, and loop over them, 
  2) to check if os.path.join(dir, filename) exists, rather than just checking if the dir is a file

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -1,12 +1,11 @@
 # Natural Language Toolkit: Internal utility functions
 #
-# Copyright (C) 2001-2013 NLTK Project
-# Author: Steven Bird <stevenbird1@gmail.com>
+# Copyright (C) 2001-2012 NLTK Project
+# Author: Steven Bird <sb@csse.unimelb.edu.au>
 #         Edward Loper <edloper@gradient.cis.upenn.edu>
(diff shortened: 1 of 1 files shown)


kmike:
Hi Dougal,

It looks like re.VERBOSE flag was missing in the re.sub call you've fixed (see http://docs.python.org/2/library/re.html#re.VERBOSE). Could you try it? I'm asking because even with `(?# ... )` comments the regexp looks wrong (e.g. there are extra spaces and newlines).

There are many unrelated changes in your pull request; it looks like you patched a file from NLTK 2.x, put the changed file into NLTK master checkout and commited the changes; this effectively reverted all changes since previous release. Could you please create a patch/pull request on top of the current code?

Thanks for contributing!

dougalg:
OK, I'll close this and try again. Sorry, this is my first time doing this! Also, I thought that this would be for nltk2. Will changes here only go into NLTK3 or both 3 and 2?

alexrudnick:
Hey Dougal,

The master branch here is for the upcoming releases of NLTK, which will be called NLTK 3.

PR #409, opened 2013-05-23
Added java.exe as alternate binary name for Java

dougalg:
2 Commits:
1) Fix a bug where setting JAVA_HOME and/or JAVAHOME in Windows does not find the java executable (add java.exe as an alternate name for java).
2) environment vars were not split and looped

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -438,7 +438,6 @@ def find_file(filename, env_vars=(), searchpath=(),
     :param url: URL presented to user for download help.
     :param verbose: Whether or not to print path when a file is found.
     """
-    if file_names is None: file_names = [filename]
     file_names = [filename] + (file_names or [])
     assert isinstance(filename, compat.string_types)
     assert not isinstance(file_names, compat.string_types)
@@ -448,10 +447,9 @@ def find_file(filename, env_vars=(), searchpath=(),
 
     # File exists, no magic
     for alternative in file_names:
-        # Check if the 'alternative' is a file extension
         path_to_file = os.path.join(filename, alternative)
         if os.path.isfile(path_to_file):
-            if verbose: print('[Found %s: %s]' % (filename, alternative))
+            if verbose: print('[Found %s: %s]' % (filename, path_to_file))
             return path_to_file
         # Check the bare alternatives
         if os.path.isfile(alternative):
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -439,7 +439,7 @@ def find_file(filename, env_vars=(), searchpath=(),
     :param verbose: Whether or not to print path when a file is found.
     """
     if file_names is None: file_names = [filename]
-    elif filename not in file_names: file_names.append(filename)
+    file_names = [filename] + (file_names or [])
     assert isinstance(filename, compat.string_types)
     assert not isinstance(file_names, compat.string_types)
     assert not isinstance(searchpath, compat.string_types)
@@ -448,18 +448,34 @@ def find_file(filename, env_vars=(), searchpath=(),
 
     # File exists, no magic
     for alternative in file_names:
+        # Check if the 'alternative' is a file extension
+        path_to_file = os.path.join(filename, alternative)
+        if os.path.isfile(path_to_file):
+            if verbose: print('[Found %s: %s]' % (filename, alternative))
+            return path_to_file
+        # Check the bare alternatives
         if os.path.isfile(alternative):
-            if verbose: print('[Found %s: %s]' % (alternative, alternative))
+            if verbose: print('[Found %s: %s]' % (filename, alternative))
             return alternative
+        # Check if the alternative is inside a 'file' directory
+        path_to_file = os.path.join(filename, 'file', alternative)
+        if os.path.isfile(path_to_file):
+            if verbose: print('[Found %s: %s]' % (filename, path_to_file))
+            return path_to_file
 
     # Check environment variables
     for env_var in env_vars:
         if env_var in os.environ:
             for env_dir in os.environ[env_var].split(os.pathsep):
+                # Check if the environment variable contains a direct path to the bin
+                if os.path.isfile(env_dir):
+                    if verbose: print('[Found %s: %s]'%(filename, env_dir))
+                    return env_dir
+                # Check if the possible bin names exist inside the environment variable directories
                 for alternative in file_names:
                     path_to_file = os.path.join(env_dir, alternative)
                     if os.path.isfile(path_to_file):
-                        if verbose: print('[Found %s: %s]'%(alternative, path_to_file))
+                        if verbose: print('[Found %s: %s]'%(filename, path_to_file))
                         return path_to_file
 
     # Check the path list.
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -62,7 +62,8 @@ def convert_regexp_to_nongrouping(pattern):
         \\.           |  # Backslashed character
         \(\?P<[^>]*>  |  # Named group
         \(\?          |  # Extension group
-        \(               # Grouping parenthesis''', subfunc, pattern)
+        \(               # Grouping parenthesis
+        ''', subfunc, pattern)
 
 
 ##########################################################################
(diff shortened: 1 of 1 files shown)


stevenbird:
Won't this modification to internals.py break on non-Windows platforms?

dougalg:
No because now it searches first for 'java', and if it is found, sets it as the path, if not found, searches for the 'alternative' name 'java.exe'.

Unless you meant something else?

To be fair, I'd like a unix/mac user to test it as well. I couldn't get the JDK to install on my mac, so I've not yet been able to test it on a non-windows environment, but I ran a test against the function and it showed it checking for both 'java' and 'java.exe'.

dougalg:
Actually, while this doesn't break anything in other environments, a more full fix would be to also include 'java' in the binary_names variable so that that will be searched for by find_file. The reason is that for some reason while find_file() searches the JAVAHOME paths for BOTH filename and file_names, it only uses the 'which' command to search for file_names, but not filename.

This is easy to fix, and I will submit another commit and pull request unless there is a reason that it won't work that I am not aware of.

nltk/internals.py
```diff
@@ -90,7 +91,7 @@ def config_java(bin=None, options=None, verbose=True):
     :type options: list(str)
     """
     global _java_bin, _java_options
-    _java_bin = find_binary('java', bin, env_vars=['JAVAHOME', 'JAVA_HOME'], verbose=verbose)
+    _java_bin = find_binary('java', bin, env_vars=['JAVAHOME', 'JAVA_HOME'], verbose=verbose, binary_names=['java.exe'])
```
dougalg, line 94: This adds java.exe as a second optional binary name to search for

nltk/internals.py
```diff
@@ -438,6 +439,7 @@ def find_file(filename, env_vars=(), searchpath=(),
     :param verbose: Whether or not to print path when a file is found.
     """
     if file_names is None: file_names = [filename]
+    elif filename not in file_names: file_names.append(filename)
```
dougalg: This merges both the single binary name and any other optional names into a single list to streamline searching later

nltk/internals.py
```diff
@@ -445,37 +447,19 @@ def find_file(filename, env_vars=(), searchpath=(),
...
 
     # Check environment variables
     for env_var in env_vars:
         if env_var in os.environ:
-            path_to_file = os.environ[env_var]
-            if os.path.isfile(path_to_file):
-                if verbose: print('[Found %s: %s]' % (filename, path_to_file))
-                return path_to_file
-            else:
+            for env_dir in os.environ[env_var].split(os.pathsep):
```
dougalg, line 467: Often environment variables contain multiple directories. They should all be searched, but first they need to be split based on the correct path separator (usually : or ;)

nltk/internals.py
kmike: `file_names.append` is bad because of 2 reasons:

1) it puts `filename` to the end of the list (previously it was the first name to look for), so this change is not backwards-compatible;
2) if passed file_names is a list then append changes it (and we shouldn't); if passed file_names is a tuple then it will raise an exception.

I also think duplicates are harmless here, and it is up to user not to pass duplicate items. What about rewriting these 2 lines this way?

```
file_names = [filename] + (file_names or [])
```

---

Using a single list makes sense, +1 from me. But I don't understand the reason why some of the code below exists (I'll leave a comment), so we probably need somebody who touched it (@xim?) to confirm changes.

nltk/internals.py
```diff
@@ -445,37 +447,19 @@ def find_file(filename, env_vars=(), searchpath=(),
         env_vars = env_vars.split()
 
     # File exists, no magic
-    if os.path.isfile(filename):
-        if verbose: print('[Found %s: %s]' % (filename, filename))
-        return filename
     for alternative in file_names:
-        path_to_file = os.path.join(filename, alternative)
```
kmike, line 452: Removing this looks backwards incompatible. I don't understand why this code existed though.

nltk/internals.py
```diff
@@ -445,37 +447,19 @@ def find_file(filename, env_vars=(), searchpath=(),
...
-        if os.path.isfile(path_to_file):
-            if verbose: print('[Found %s: %s]' % (filename, path_to_file))
-            return path_to_file
-        path_to_file = os.path.join(filename, 'file', alternative)
```
kmike, line 456: Removing this is also backwards-incompatible.

nltk/internals.py
```diff
@@ -445,37 +447,19 @@ def find_file(filename, env_vars=(), searchpath=(),
...
-            return path_to_file
+        if os.path.isfile(alternative):
+            if verbose: print('[Found %s: %s]' % (alternative, alternative))
+            return alternative
 
     # Check environment variables
     for env_var in env_vars:
         if env_var in os.environ:
-            path_to_file = os.environ[env_var]
-            if os.path.isfile(path_to_file):
```
kmike, line 465: This removes support for passing full path to binary in an environment variable, so this is also backwards-incompatible.

kmike:
I like the idea behind these changes, and I agree that we should add java.exe as a secondary option, but some of these changes are backwards-incompatible and it is better to understand what are we breaking and for what reason.

stevenbird:
Thanks @dougalg for the contributions, and @kmike for the feedback.

dougalg:
@kmike, thank you for the extensive comments! I've restored the backwards compatibility and added your suggested modifications.

nltk/internals.py
```diff
@@ -438,6 +439,7 @@ def find_file(filename, env_vars=(), searchpath=(),
     :param verbose: Whether or not to print path when a file is found.
     """
     if file_names is None: file_names = [filename]
```
kmike: This line is not neccessary (and it causes file_names to be [filename, filename] instead of [filename] if file_names is None).

nltk/internals.py
```diff
@@ -445,14 +447,17 @@ def find_file(filename, env_vars=(), searchpath=(),
...
     # File exists, no magic
-    if os.path.isfile(filename):
-        if verbose: print('[Found %s: %s]' % (filename, filename))
-        return filename
     for alternative in file_names:
+        # Check if the 'alternative' is a file extension
         path_to_file = os.path.join(filename, alternative)
         if os.path.isfile(path_to_file):
-            if verbose: print('[Found %s: %s]' % (filename, path_to_file))
+            if verbose: print('[Found %s: %s]' % (filename, alternative))
```
kmike: I think it is better to leave `path_to_file` here because it is the path found

nltk/internals.py
```diff
@@ -445,14 +447,17 @@ def find_file(filename, env_vars=(), searchpath=(),
         env_vars = env_vars.split()
 
     # File exists, no magic
-    if os.path.isfile(filename):
-        if verbose: print('[Found %s: %s]' % (filename, filename))
-        return filename
     for alternative in file_names:
+        # Check if the 'alternative' is a file extension
         path_to_file = os.path.join(filename, alternative)
```
kmike, line 450: is the comment correct?

```
>>> import os
>>> os.path.join('foo', '.exe')
'foo/.exe'
```

Using "alternative" as a file extension could make sense, but it is not what the code does currently.

nltk/internals.py
```diff
@@ -461,19 +464,14 @@ def find_file(filename, env_vars=(), searchpath=(),
...
+                    if verbose: print('[Found %s: %s]'%(filename, env_dir))
+                    return env_dir
+                # Check if the possible bin names exist inside the environment variable directories
                 for alternative in file_names:
-                    path_to_file = os.path.join(os.environ[env_var],
-                                                alternative)
-                    if os.path.isfile(path_to_file):
-                        if verbose: print('[Found %s: %s]'%(filename, path_to_file))
-                        return path_to_file
-                    path_to_file = os.path.join(os.environ[env_var], 'file',
```
kmike, line 475: one more point :) this removes support for files inside 'file' subfolders of folders from environment variables, so this is backwards incompatible

dougalg:
Thank you. Sorry about all the broken backwards compatibility, this has been a good lesson for me.

kmike:
The code looks good, and it works for me on Mac OS X, so merging it.

Thanks for contributing! I hope the process was not too tedious and it won't discourage you from contributing in future.

dougalg:
Absolutely not, if I can find a bug I can fix, I'd like to contribute more. This is very helpful for me, since I've never had a chance to work on a collaborative project before. I'm just glad I was able to help a bit!

PR #410, opened 2013-05-23
Removed hardcoded classifier java class from weka module

tibidoh:
Hi.  I'm new to python so please double-check this.

### DIFF
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -95,7 +95,7 @@ class WekaClassifier(ClassifierI):
         'C4.5': 'weka.classifiers.trees.J48',
         'log_regression': 'weka.classifiers.functions.Logistic',
         'svm': 'weka.classifiers.functions.SMO',
-        'kstar': 'weka.classifiers.lazy.kstar',
+        'kstar': 'weka.classifiers.lazy.KStar',
         'ripper': 'weka.classifiers.rules.JRip',
         }
 
@@ -147,7 +147,7 @@ class WekaClassifier(ClassifierI):
                                      % stderr)
 
             # Parse weka's output.
-            return self.parse_weka_output(stdout.split('\n'))
+            return self.parse_weka_output(stdout.decode().split('\n'))
 
         finally:
             for f in os.listdir(temp_dir):
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -76,17 +76,49 @@ def _check_weka_version(jar):
         zf.close()
 
 class WekaClassifier(ClassifierI):
+
+    # [xx] full list of classifiers (some may be abstract?):
+    # ADTree, AODE, BayesNet, ComplementNaiveBayes, ConjunctiveRule,
+    # DecisionStump, DecisionTable, HyperPipes, IB1, IBk, Id3, J48,
+    # JRip, KStar, LBR, LeastMedSq, LinearRegression, LMT, Logistic,
+    # LogisticBase, M5Base, MultilayerPerceptron,
+    # MultipleClassifiersCombiner, NaiveBayes, NaiveBayesMultinomial,
+    # NaiveBayesSimple, NBTree, NNge, OneR, PaceRegression, PART,
+    # PreConstructedLinearModel, Prism, RandomForest,
+    # RandomizableClassifier, RandomTree, RBFNetwork, REPTree, Ridor,
+    # RuleNode, SimpleLinearRegression, SimpleLogistic,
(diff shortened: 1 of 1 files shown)


kmike:
Hi Alexander,

Thanks for contributing. I'm not a user of WekaClassifier myself, but it looks like it is better to make "classifier" a constructor parameter because:
- it is not a good practice to change interface in a subclass (`ClassifierI.batch_classify` and `ClassifierI.batch_prob_classify` don't have 'classifier' parameter);
- 'train' method already knows what classifier is used and it could pass it to a constructor.

Setting model parameters in a constructor is a common practice (for example, this is what scikit-learn does).

What I don't understand is how the demo works (it is at the bottom of the file). WekaClassifiers returned by WekaClassifier.train could be trained using any Weka classifier, but then it uses NB at prediction time. Could it be the case Weka doesn't care about classifier if model is already trained? I have troubles installing Weka on my computer for misterious reasons, does demo work for you?

tibidoh:
Sorry, it seems I will not have time to check the demo this week. Is it OK that pull request will hang for this time?

kmike:
Yes, there is no rush. Thanks!

stevenbird:
@tibidoh - if we can resolve this issue this week, it will make it into the next release.

tibidoh:
Hello again! 
I'm really sorry for a huge delay since the pull request opening. I have created a little test: https://gist.github.com/tibidoh/6139489 and corrected a couple of things:
1 - Stdout of Java was not decoded to Unicode, that broke python 3 compatibility in method split('\n'). Addition of encoding also affected the behavior for python 2.x. Do we need to consider such things?
2 - Name of java class 'KStar' was misspelled.

Also, I have a question about API design of WekaClassifier. Why do we weed to create an instance of it to classify things although we don't for building a model?

tibidoh:
Will play with the demo soon

tibidoh:
Kmike, you were right that weka is not caring about classifier when training. I have tried batch_classify without additional parameter which I introduced and it works perfectly :) 
I think I need to find some references for this behavior in weka documentation.

tibidoh:
So, what we have now: 
-Instead of putting the model to file Weka simply serializes classifier itself. It explains why 'incompatible' models (e.g. decision tree and NB) worked. When weka is invoked for classification the classifier object itself passes control to general runtime which deserializes classifier and invokes it. Type of class which was originally invoked makes no difference.
-Demo from weka.py works :)
-We still have incompatibility with py3 and misspelled class name.

I propose to close this request since most of its changes are not needed. Minor changes can be made in a separate one.

kmike:
@tibidoh thanks for checking this! I'll close the PR then.

A note about Python3 compat: you're decoding Weka stdout output with default encoding `sys.getdefaultencoding()`, which is often ASCII. Could you please explain why? Is WEKA output ASCII-only? `sys.getdefaultencoding()` is not always the encoding used for IO. Also, If I'm not mistaken, in Python 3 stdout should already return unicode.

tibidoh:
@kmike I have mentioned that I don't have solid experience with Python. Actually I made parsing of the output working for me leaving the details you mentioned out of my sight.
As for encoding: In the current version of weka most of the output is made via `PrintStream`'s `println()` method. Javadoc for that method says: "The string produced by the String.valueOf(Object) method is translated into bytes according to the platform's default character encoding..."

kmike:
@tibidoh those encoding issues are rather obscure, and I myself am not 100% sure I'm getting them right; your Python is fine. 

There are several default platform encodings, i.e. system encoding and console encoding. For example, in Russian Windows XP default system encoding is usually cp1251 and console encoding is usually cp866. Do you know which one Java's println uses? If it uses console encoding then bytes should be decoded using stdin.encoding; if it uses system encoding then just 'decode' is a correct way to decode bytes.

Unfortunately I can't check and test this myself because Weka doesn't install for me for arcane reasons.

kmike:
> if it uses system encoding then just 'decode' is a correct way to decode bytes.

.. or maybe locale.getpreferredencoding() - that's crazy :)

tibidoh:
I can't check this too because on my platform both `stdin.encoding` and `sys.getdefaultencoding()` both return UTF-8 and I don't have access to windows machine. Although I didn't find anything about it in java documentation, it would be reasonable to suggest using `stdin.encoding` in this situation because it would be wrong by design to write to stdout in encoding which is different from console's encoding by default. 
Some time ago I've worked with java on Russian WinXp and I didn't have any problems running `System.out.println("someRussianText")`

kmike:
I think it means Java uses console encoding for println indeed.

PR #412, opened 2013-05-25
Remove unnecessary conditional imports

kmike:
In this pull requests some conditional imports are removed. They are no longer necessary (as numpy imports are protected by try/except since https://github.com/nltk/nltk/commit/73906dcce7e1552de277dcfdcca30e358f433319). 

This should make exception messages more informative and prevent issues like http://stackoverflow.com/questions/15320894/python-importerror-maxentclassifier/15321784.

This change doesn't break any tests for me (including tests in environments without numpy & friends).


stevenbird:
So if I understand correctly, we now only fail at runtime (not load time) when dependencies are not installed.

Issue #395, opened 2013-04-23
WordNet: Lowest Common Hypernym


dabril:
I have seen a strange behavior when using the function to find the lowest common hypernym between two synsets. 
For instance, Synset('policeman.n.01') and Synset('firefighter.n.01')
## Example
- wn.synset('policeman.n.01').definition
  - 'a member of a police force'
- wn.synset('chef.n.01').definition
  - 'a professional cook'
- If we compute the lowest common hypernym we get the following synset _wn.synset('policeman.n.01').lowest_common_hypernyms(wn.synset('chef.n.01'))_
  - **[Synset('organism.n.01')]**

However, if we look at the hypernym paths
- _wn.synset('policeman.n.01').hypernym_paths()_
  1. [Synset('entity.n.01'), Synset('physical_entity.n.01'), Synset('object.n.01'), Synset('whole.n.02'), Synset('living_thing.n.01'), Synset('organism.n.01'), _**Synset('person.n.01')**, Synset('preserver.n.03'), Synset('defender.n.01'), Synset('lawman.n.01'), Synset('policeman.n.01')_],
  2. [Synset('entity.n.01'), Synset('physical_entity.n.01'), Synset('causal_agent.n.01'), _**Synset('person.n.01')**, Synset('preserver.n.03'), Synset('defender.n.01'), Synset('lawman.n.01'), Synset('policeman.n.01')_]
- _wn.synset('firemann.04').hypernym_paths()_
  1. [Synset('entity.n.01'), Synset('physical_entity.n.01'), Synset('object.n.01'), Synset('whole.n.02'), Synset('living_thing.n.01'), Synset('organism.n.01'), _**Synset('person.n.01')**, Synset('worker.n.01'), Synset('skilled_worker.n.01'), Synset('cook.n.01'), Synset('chef.n.01')_],
  2. [Synset('entity.n.01'), Synset('physical_entity.n.01'), Synset('causal_agent.n.01'), _**Synset('person.n.01')**, Synset('worker.n.01'), Synset('skilled_worker.n.01'), Synset('cook.n.01'), Synset('chef.n.01')_]

As can be seen in all the hypernym paths the lowest common ancestor should be Synset('person.n.01') instead of Synset('organism.n.01'), which is the immediately ancestor of one of the paths of the Synset('person.n.01').

[...]

PR #415, opened 2013-06-05
Proposed resolution for issue 395

dougalg:
See [Issue 395](https://github.com/nltk/nltk/issues/395)

Currently lowest common hypernyms do not return a result based on the
hypernym paths of the 2 synsets. So there may be a "better" result when
taking into account the hypernym paths.

```
>>> wn.synset('policeman.n.01').relative_lowest_common_hypernyms(wn.synset('chef.n.01'))
[Synset('person.n.01')]
>>> wn.synset('policeman.n.01').lowest_common_hypernyms(wn.synset('chef.n.01'))
[Synset('organism.n.01')]
```

I'm not sure if this will be considered useful or not, but I thought this approach might make sense. Of course, I'm open to changes and suggestions.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -298,7 +298,6 @@ class Synset(_WordNetObject):
     - root_hypernyms
     - common_hypernyms
     - lowest_common_hypernyms
-    - relative_lowest_common_hypernyms
 
     Note that Synsets do not support the following relations because
     these are defined by WordNet as lexical relations:
@@ -451,11 +450,12 @@ class Synset(_WordNetObject):
                            for other_synset in other_synsets)
         return list(self_synsets.intersection(other_synsets))
 
-    def lowest_common_hypernyms(self, other, simulate_root=False):
+    def lowest_common_hypernyms(self, other, simulate_root=False, use_max_depth=False):
         """Get a list of absolute lowest synset(s) that both synsets have as a hypernym.
 
-        This is calculated by finding the shortest paths for all synsets that are hypernyms
-        of both words, and returning that with the longest path."""
+        By default this is calculated by finding the shortest paths for all synsets that are hypernyms
+        of both words, and returning that with the longest path. By setting use_max_depth=True, lower 
+        hypernyms can be found."""
 
         fake_synset = Synset(None)
         fake_synset.name = '*ROOT*'
@@ -474,40 +474,14 @@ class Synset(_WordNetObject):
         synsets.intersection_update(others)
 
         try:
-            max_depth = max(s.min_depth() for s in synsets)
-            return [s for s in synsets if s.min_depth() == max_depth]
+            if use_max_depth:
+                max_depth = max(s.max_depth() for s in synsets)
+                return [s for s in synsets if s.max_depth() == max_depth]
+            else:
+                max_depth = max(s.min_depth() for s in synsets)
+                return [s for s in synsets if s.min_depth() == max_depth]
         except ValueError:
             return []
-
-    def relative_lowest_common_hypernyms(self, other, simulate_root=False):
-        """Get a list of the lowest synset(s) that both synsets have as a hypernym, relative
-        to the hypernym paths of this synset.
-
-        This compares the hypernym paths of this synset with those of the other word.
-        The lowest hypernym from each path of this word that appears in a hypernym path
-        of the other word is returned."""
-
-        fake_synset = Synset(None)
-        fake_synset.name = '*ROOT*'
-        fake_synset.hypernyms = lambda: []
-        fake_synset.instance_hypernyms = lambda: []
-
-        if simulate_root:
-            self_hypernyms = chain(self.hypernym_paths(), [[fake_synset]])
-            other_hypernyms = chain(other.hypernym_paths(), [[fake_synset]])
-        else:
-            self_hypernyms = self.hypernym_paths()
-            other_hypernyms = other.hypernym_paths()
-
-        results = []
-        for self_path in self_hypernyms:
-            for other_path in other_hypernyms:
-                for s in self_path[::-1]:
-                    if s in other_path:
-                        results.append(s)
-                        break
-
-        return list(set(results))
 
     def hypernym_distances(self, distance=0, simulate_root=False):
         """
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -453,9 +453,29 @@ class Synset(_WordNetObject):
     def lowest_common_hypernyms(self, other, simulate_root=False, use_max_depth=False):
         """Get a list of absolute lowest synset(s) that both synsets have as a hypernym.
 
-        By default this is calculated by finding the shortest paths for all synsets that are hypernyms
-        of both words, and returning that with the longest path. By setting use_max_depth=True, lower 
-        hypernyms can be found."""
+        By default this is calculated by finding the shortest paths for all synsets that 
+        are hypernyms of both words, and returning that with the longest path.
+
+        By setting the use_max_depth flag to True, lower hypernyms can be found by searching for the 
+        longest paths of each hypernym.
+
+        :type other: Synset
+        :param other: other input synset
+        :type simulate_root: bool
+        :param simulate_root: The various verb taxonomies do not
+            share a single root which disallows this metric from working for
+            synsets that are not connected. This flag (False by default)
+            creates a fake root that connects all the taxonomies. Set it
+            to True to enable this behavior. For the noun taxonomy,
+            there is usually a default root except for WordNet version 1.6.
+            If you are using wordnet 1.6, a fake root will need to be added 
+            for nouns as well.
+        :type use_max_depth: bool
+        :param use_max_depth: If True, will use the max_depth function to 
+            calculate the lowest common hypernyms giving results that should
+            be lower in the tree than when using the default settings.
+        :return: The synsets that are the lowest common hypernyms of both synsets
+        """
 
         fake_synset = Synset(None)
         fake_synset.name = '*ROOT*'
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -298,6 +298,7 @@ class Synset(_WordNetObject):
     - root_hypernyms
     - common_hypernyms
     - lowest_common_hypernyms
+    - relative_lowest_common_hypernyms
 
     Note that Synsets do not support the following relations because
     these are defined by WordNet as lexical relations:
@@ -451,7 +452,10 @@ class Synset(_WordNetObject):
         return list(self_synsets.intersection(other_synsets))
 
     def lowest_common_hypernyms(self, other, simulate_root=False):
-        """Get the lowest synset that both synsets have as a hypernym."""
+        """Get a list of absolute lowest synset(s) that both synsets have as a hypernym.
+
+        This is calculated by finding the shortest paths for all synsets that are hypernyms
+        of both words, and returning that with the longest path."""
 
         fake_synset = Synset(None)
         fake_synset.name = '*ROOT*'
@@ -474,6 +478,36 @@ class Synset(_WordNetObject):
             return [s for s in synsets if s.min_depth() == max_depth]
         except ValueError:
             return []
+
+    def relative_lowest_common_hypernyms(self, other, simulate_root=False):
+        """Get a list of the lowest synset(s) that both synsets have as a hypernym, relative
+        to the hypernym paths of this synset.
+
+        This compares the hypernym paths of this synset with those of the other word.
+        The lowest hypernym from each path of this word that appears in a hypernym path
+        of the other word is returned."""
+
+        fake_synset = Synset(None)
+        fake_synset.name = '*ROOT*'
+        fake_synset.hypernyms = lambda: []
+        fake_synset.instance_hypernyms = lambda: []
+
+        if simulate_root:
+            self_hypernyms = chain(self.hypernym_paths(), [[fake_synset]])
+            other_hypernyms = chain(other.hypernym_paths(), [[fake_synset]])
+        else:
+            self_hypernyms = self.hypernym_paths()
+            other_hypernyms = other.hypernym_paths()
+
+        results = []
+        for self_path in self_hypernyms:
+            for other_path in other_hypernyms:
+                for s in self_path[::-1]:
+                    if s in other_path:
+                        results.append(s)
+                        break
+
+        return list(set(results))
 
     def hypernym_distances(self, distance=0, simulate_root=False):
         """


dougalg:
Actually, I would rather see this replace the current implementation of the `lowest_common_hypernyms` function, but I did it this way initially for discussion.

Does anyone have any comments or objections?

tschier:
Hi, I'm a newcomer to the forum having just encountered Issue 395 myself (combinations of synsets `employee.n.01`, `customer.n.01`, `worker.n.01` and others all result in `organism.n.01` instead of `person.n.01`).

The additional solution I came up with is a bit simpler: in lines 478 and 479 replace `s.min_depth()` with `s.max_depth()`. This results in the (actually lower) lowest common hypernym producing `person.n.01` in the examples I gave above.

A difference in result (which I believe is a slight advantage) is that the hypernym paths to common hypernyms will at least be the same length. In your proposed solution it seems a common hypernym can appear anywhere in the `other` synset's hypernym path to still be selected.

Either way, these solutions may break some people's existing use of `lowest_common_hypernyms()`, so having either an additional method or additional flag parameter is probably the best way to go.

dougalg:
Thank you for the comment! I think this will have essentially the same effect as what I wrote, but the code will be clearer with your approach. It should be that same since the `max_depth` should always appear in at least one of the hypernym paths, if I understand correctly.

If this approach is acceptable, rather than duplicating most of the code in 2 functions, would it make sense to just add an additional True/False argument to the original function to toggle between the 2 behaviors? ie:

```
def lowest_common_hypernyms(self, other, simulate_root=False, use_max_depth=False):
    ...
    ...
    try:
         if use_max_depth:
             max_depth = max(s.max_depth() for s in synsets)
             return [s for s in synsets if s.max_depth() == max_depth]
         else: 
             max_depth = max(s.min_depth() for s in synsets)
             return [s for s in synsets if s.min_depth() == max_depth]
     except ValueError:
         return []
```

I wonder if anyone knows why `min-depth` is used by default? Does `max_depth` occasionally give incorrect output?

dougalg:
If accepted, new usage:

```
>>> wn.synset('policeman.n.01').lowest_common_hypernyms(wn.synset('chef.n.01'))
[Synset('organism.n.01')]
>>> wn.synset('policeman.n.01').lowest_common_hypernyms(wn.synset('chef.n.01'), use_max_depth=True)
[Synset('person.n.01')]
```

tschier:
That looks fine to me dougalg.

kmike:
"person" is in some sense more general than "organism" because its min_depth is 3 while organism has min_depth 5, so returning "organism" as a lowest synset is not that unreasonable, I don't see this as a bug. But person's max_depth is 6 which is greater than organism's max_depth==5, so in some sense person is more specific than organism.

I think the root cause of this issue is the ambiguity of "the lowest synset", because there are at least two depth measures in NLTK: min depth and max depth. There is a depth for each of the paths, so even min/max depth measures are arbitrary in some sense (what about e.g. average depth?), but min and max depth are commonly used, so let's consider only them. 

Changing default to max depth could break people's code and make results unreproducible between NLTK versions.
It may also be noted that min depth is used in Ted Petersen's Perl Wordnet module, see http://search.cpan.org/dist/WordNet-Similarity/lib/WordNet/Similarity/DepthFinder.pm (look at description of getLCSbyDepth method). It seems that compatibility with that implementation is somewhat desirable, see e.g. https://github.com/nltk/nltk/commit/101a73a75e969d85eb221dfaf3f61e70e1209fa2 - I don't know the whole story though. It also looks like the current code is compatible with Perl implementation regarding min_depth/max_depth. See also: http://acl.ldc.upenn.edu/N/N04/N04-3012.pdf.

A side note: docstring for Wu-Palmer similarity score (implemented in wup_similarity method) says:

> Where the LCS has multiple paths to the root, the longer path is used for the purposes of the calculation.

This is quite subtle, because currently for finding LCS (aka lowest common hypernym) min_depth is used, but then max_depth is used. 

I wasn't able to find any papers with min depth/max depth suggestions for LCS task, so unless somebody points us to a relevant research or @stevenbird (or somebody else) have something to say, I think that:

1) both metrics are reasonable;
2) there is no evidence that we should prefer max_depth to min_depth.

So, 
- +1 to introducing a flag for choosing between these two measures.
- +1 for leaving min_depth as a default metric to remain compatible with previous NLTK versions and with Ted Peterson's implementation. 

I also think that the extended docstring (by the way, kudos for the extended docstring!) should be changed to be a bit more explicit - "lower" and "lowest" looks ambiguous because there are several valid depth metrics. Otherwise this pull request looks good to me.

stevenbird:
Thanks for the contributions and discussion -- I agree with @kmike 's analysis, and will merge this and tweak the docstring to avoid ambiguity.

stevenbird:
I'm having second thoughts about this solution, as I write doctests for it. It seems strange to return organism as the default LCS for policeman and chef. I will ask for Ted Pedersen's input.

dougalg:
@kmike thanks for the discussion. This makes it much more clear why min_depth() was chosen for calculation.

I think the issue is that people expect a bottom-up approach, but we're providing a top-down one. Looking at the actual hypernym_paths(), in no case is `organism` lower than `person` in that path, which I think is why people are confused, even if comparing the two paths we can see that `organism` in the 2nd path is lower than `person` in the first path. I think the problem for people is that the current default decontextualizes things in order to get the results.

duluthted:
Greetings all,

Steven asked me what I thought about this issue, and I thought perhaps the best way to do that would be to try and share some of the ideas and intuitions that guided us in creating WordNet::Similarity. 

To summarize very briefly, indeed what we do in WordNet::Similarity is bottom up, and in the case of chef and policeman we would return person not organism. The argument in favor of that (to me) is that person is a more
specific concept, where specificity can be measured by depth of the concept, although you'll see that has a few wrinkles.

In WordNet::Similarity we find all the paths from the hypernyms of each word/sense in a pair being measured to the root, and then we look for the point of intersection (least common subsumer) between each possible pair of paths. It's important to find all the paths because there is a good deal of multiple inheritance, especially as you work your way up towards the root.

If there are multiple points of intersection (lcs) (and there often are) then we choose the "deepest" one, where the definition of depth is different depending on if we are using a path based measure (path, wup, lch) or an information content measure (lin, res, jcn).

In the case of the path based measure, we choose the candidate lcs that has the largest/greatest path length to the root. The intuition here is fairly straightforward and corresponds to a sort of graph theoretic view of what it means to be "deep" in a tree.

Now there's an important issue that comes up, and that's multiple inheritance. It turns out to be pretty common, and you can certainly find candidate lcs' with multiple paths to the root. In this case there's a question there about how you define the depth of such a candidate lcs. In that case, we choose the longest/deepest path rather
than the shortest path as the definition of depth. I think our feeling was motivated by our view of depth as a proxy for specificity, which I think is at the heart of the idea of a lcs. Our feeling was that you want the least common subsumer to be the most _specific_ concept that is an ancestor of both words/senses. We always want to make sure we get the most specific interpretation, so that's why we opt to look for the deeper value (in the case of multiple inheritance).

In the case of the information content measures, we choose the candidate lcs with the largest information content value. Oh my. :) The intuition here is much more subtle I think, but the idea is that if you are using info content measures, you are generally a bit skeptical of a graph theoretic view of depth (that's why you are
choosing to use the info content measures). Info content in general is a measure of specificity, and so a larger IC values means a more specific concept, which is what we consider to be "deeper" in this case. As a practical matter I think you will often end up choosing the same lcs using either the path based or info content approach, but it seemed important to handle these differently (and in ways consistent with the general family of measures). One nice thing about info content measures is that multiple inheritance is actually easier to handle, since you just get one IC value per concept/sense and so you don't have this issue of making a choice.

To generalize a bit, I think in both cases (path based and info content) we are viewing depth to be a measure of specificity - more specific concepts are deeper, and the least common subsumer should be the most specific concept that is an ancestor of a pair. In path based measures depth is a graph theoretic notion, whereas the info content measures take a more info content view...

So... I realize that's quite a lot to dump into the middle of this discussion, but I thought better to just get it all out there. I should say this took a long time to sort out in WordNet::Similarity, and there are a few places here where we made choices that could have possibly been done some other way. But, I think/hope overall that our ways of finding lcs are consistent and reasonable. 

I'm happy to  clarify as needed, and am even happier to be corrected if I've said something that seems really really off the wall. :)

Good luck,
Ted

stevenbird:
Thanks to Ted for a comprehensive discussion of the issues! The implications of this for our implementation of lowest_common_hypernyms is that it should be bottom-up, something like what @dougalg suggested in relative_lowest_common_hypernyms in https://github.com/dougalg/nltk/commit/86ede86608d9f52292c447999cd1be5b48f246ee

It would be nice to have confidence in our solution. Ideally it should be a port of the one in WordNet::Similarity, or it should include some edge cases (in nltk/test/wordnet.doctest).

stevenbird:
I would appreciate clarification on the following case. Suppose we have two synsets s1 and s2, each with two hypernym paths. Following @dougalg 's solution, we iterate up each of the hypernym paths from s2 looking for the first time a hypernym in s1 is encountered.

Suppose the first hypernym path from s1 intersects the first hypernym path from s2, and that the second hypernym path from s1 intersects the second hypernym path from s2. Do we return both (as in @dougalg 's) solution, or do we return the lowest (and if its the lowest, according to which criterion, since there could be multiple paths back from a common hypernym to the root).

dougalg:
@duluthted, thank you for the extensive explanation. @stevenbird, In the perl wordnet module cited above by @kmike they return the results as a list of `(synset, depth)` tuples. Maybe such a result could be returned here, with `depth` being any of: `min_depth()`, `max_depth()`, or a distance score based on the number of hypernyms between the key synset and this hypernym (or average of distances from the bottom synset in each relevant path)?

The intent of the original code is definitely to return multiple valid results, and for me, something along these lines seems to make at least intuitive sense.

duluthted:
Just to elaborate a little more, we normally take the position that WordNet::Similarity methods should return lists of results so that a developer has access to everything we know about behind the scenes. However, for code that a user might run and get a result from, we like to provide them with a single answer if at all possible. That's why having a sort of intuition behind whatever you choose as your preferred result seems useful. Most users don't actually ask "why do i get this" but it's nice for those that do. 

Anyway, probably a self evident kind of point but just wanted to add that to the comment made by @dougalg 

Also, at various times we have thought about some kind of averaging of depths to deal with multiple inheritance, but we shied away. As best I recall we did some experiments with some of the human reference standards out there (Miller and Charles pairs for example) and found that our current approach did better with those. (This is a little hazy though so take this as more anecdotal evidence). I do think that significant innovations in how measures are potentially computed should be experimented with a good bit before settling on a particular interpretation. To my mind wup (Wu and Palmer) is probably the measure most affected by these depth issues, so might be a good measure to evaluate should you go down that road.

stevenbird:
@dougalg - I would prefer not to change the interface in a way that breaks people's code. Optional flags are fine, but changing the return type is not.

I think we can all agree that the lowest common hypernym of 'policeman' and 'chef' is 'person', not 'organism'. Let's adopt this as one of our test cases, and make sure that any implementation gets this right.

How about coming up with some more test cases? These can live in our doctests (nltk/test), and can give us some confidence in our final solution. Does the WordNet::Similarity suite include tests that we can adopt?

@dougalg can you run your relative_lowest_common_hypernyms() function on some random cases and see if you can find instances where we get more than one solution? It would be interesting to discuss these with @duluthted, since I like his point that "for code that a user might run and get a result from, we like to provide them with a single answer if at all possible".

dougalg:
@stevenbird I'd be happy to. It'll take me a little time, as I'm very busy at the moment, but I imagine that's not a big problem, right?

stevenbird:
@dougalg - I hope to get a new release out this week - is that enough impetus to get on to this?

dougalg:
@stevenbird - I will have free time Wed and Thu this week when I was planning to work on this. Will that give enough time, or will it be too late? I'll try to work on it earlier if I get any additional time.

dougalg:
I've been writing tests, and preliminarily there looks to be a strong case to just use `min_depth` as the default. I ran through the synsets for the top 100 most common English nouns which ended up being about 190,000 hypernym paths. In 1332 cases there was a conflict between `min_depth` and `max_depth` in which the `max_depth` result was actually lower than the `min_depth` result in the relevant paths.

There were no cases where the reverse was true. To me this seems to show that `max_depth` is in all cases better than `min_depth` and I assume that this must be caused by the way the `hypernym_paths` function works. I think it must end up including the longest and shortest routes, so that the `max_depth` result is always there. But I am not 100% sure of that.

The algorithm I proposed initially is no good if we want to keep the results to 1 in most cases, I had thought at that time that returning multiple results would be useful, but it returns many more results than `min_depth` and `max_depth` (~18,300 out of ~200,000). Both min and max depth occasionally produce multiple results (16 and 96 times out of ~200,000, so min_depth has a slight edge here). However, I think that that is offset by the better quality of results in the edge cases.

If this makes sense then I'll proceded with writing a test file, updating the doctests, and setting max_depth to default with a boolean option to set it

While testing I came across this case:

```
>>> wn.synset('woman.n.01').lowest_common_hypernyms(wn.synset('girlfriend.n.02'), use_max_depth=True)
[Synset('woman.n.01')]
```

Is this expected behavior, or should the code ignore the synsets that you are searching on. I mean, is a synset its own lowest hypernym? That seems absurd since by definition a hypernym should be above, but it does make a kind of sense, when looking at `Synset('girlfriend.n.02')`.

The next LCHs for those two synsets (should be) `[Synset('adult.n.01'), Synset('female.n.02)]`. Any thoughts on this? This is an easy fix, either way, though. Personally, I think it makes sense to exclude the target synsets themselves as possible LCHs. 

Quick Note: It seems to be mainly words about people that give `min_depth` a hard time, interestingly. Other kinds of words seem to work generally fine.

duluthted:
Greetings all,

I might be misunderstanding something here as I'm not sure about the term
"lowest common hypernym" - I guess for this case I'm assuming we mean
something like least common subsumer (or most specific ancestor) but if
that's not the case then feel free to ignore the rest.

Anyway, I've never really thought that a single synset would have a least
common subsumer - rather I've always thought that an LCS is a binary
relationship if you will, where you have two synsets as input. In that
case, if you have synsets a and a, I think that their LCS is in fact a. So,
if you were looking for the lch of woman.n.1 and woman.n.1, I would say
that woman.n.1 is indeed their LCS  This makes sense when you think about
self-similarity according to various of the measures of similarity (like
wup).

wup (woman.n.1, woman.n.1) = 2*depth(woman.n.1.)/(depth(woman.n.1) +
depth(woman.n.1) = 1.0

If the LCS of woman.n.1 and woman.n.1 was something like adult.n.1 then
you'd get a self similarity value less than 1.0, which doesn't seem
reasonable for this measure.

Anyway, I might be a little off base here due to differences in
terminology, but just thought I'd raise this particular case since it seems
possibly relevant.

Cordially,
Ted

On Wed, Jul 10, 2013 at 9:20 AM, dougalg notifications@github.com wrote:

> I've been writing tests, and preliminarily there looks to be a strong case
> to just use min_depth as the default. I ran through the synsets for the
> top 100 most common English nouns which ended up being about 190,000
> hypernym paths. In 1332 cases there was a conflict between min_depth and
> max_depth in which the max_depth result was actually lower than the
> min_depth result in the relevant paths.
> 
> There were no cases where the reverse was true. To me this seems to show
> that max_depth is in all cases better than min_depth and I assume that
> this must be caused by the way the hypernym_paths function works. I think
> it must end up including the longest and shortest routes, so that the
> max_depth result is always there. But I am not 100% sure of that.
> 
> The algorithm I proposed initially is no good if we want to keep the
> results to 1 in most cases, I had thought at that time that returning
> multiple results would be useful, but it returns many more results than
> min_depth and max_depth (~18,300 out of ~200,000). Both min and max depth
> occasionally produce multiple results (16 and 96 times out of ~200,000, so
> min_depth has a slight edge here). However, I think that that is offset by
> the better quality of results in the edge cases.
> 
> If this makes sense then I'll proceded with writing a test file, updating
> the doctests, and setting max_depth to default with a boolean option to set
> it
> 
> While testing I came across this case:
> 
> > > > wn.synset('woman.n.01').lowest_common_hypernyms(wn.synset('girlfriend.n.02'), use_max_depth=True)
> > > > [Synset('woman.n.01')]
> 
> Is this expected behavior, or should the code ignore the synsets that you
> are searching on. I mean, is a synset its own lowest hypernym? That seems
> absurd since by definition a hypernym should be above, but it does make a
> kind of sense, when looking at Synset('girlfriend.n.02').
> 
> The next LCH for those two synsets (should be) Synset('adult.n.01') and
> Synset('female.n.02). Any thoughts on this? This is an easy fix, either
> way, though. Personally, I think it makes sense to exclude the target
> synsets themselves as possible LCHs.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/nltk/nltk/pull/415#issuecomment-20745161
> .

## 

Ted Pedersen
http://www.d.umn.edu/~tpederse

kmike:
@dougalg thanks for your work. I want to make a quick note: `lowest_common_hypernyms` method is used by similarity score methods. So if we'll make min_depth=True default IMHO it is better to explicitly pass min_depth=False in similarity scoring methods because Ted said the similarity score functions took long time to sort out and we'd better not change them.

@duluthted thanks for your ongoing participation here! I personally learned a lot from your comments (and expanded my "learning TODO" a lot).

dougalg:
@duluthted Thank you for the clarification. I think that that makes complete sense, and that maybe just the name "lowest common hypernym" is less clear in some ways that "least common subsumer". But I agree with you that it makes sense to do it that way, and I'll make a note in the docstring about it as well.

@kmike I'll do that. But, is it something we should consider checking out later? Is it possible that that might make the results slightly better in these edge cases? (doing a quick test, chef and policeman get a higher rating on wup_similarity [0.54... with min_depth vs 0.63... with max_depth] but woman and girlfriend do not) Is this something to think about revisiting later on? From a quick look at the code I think that it should only improve results in those edge cases where min and max depth disagreed, but I can't be sure.

stevenbird:
@dougalg thanks for the analysis! So the conclusion is that "max_depth is in all cases better than min_depth". This makes perfect sense to me. I agree with @duluthted that we should note the equivalence with "least common subsumer", and that LCH(a,a)=a.

@kmike suggests that we support old behaviour (though I'm confused by the suggestion of passing "min_depth=False in similarity scoring methods", when we have already said that max_depth is best in general -- so wouldn't you be asking for min_depth=True?). Anyway, I think that correctness wins out. I can't see the argument for using a different LCH for the path similarity metrics.

PR #417, opened 2013-06-13
Fix detect_encoding regex in corpus.reader.xmldocs.XMLCorpusView

wroberts:
corpus.reader.xmldocs.XMLCorpusView includes a function to read the file encoding out of the XML declaration on the first line of a corpus file.

Unfortunately, this function relies on regular expressions which contain a simple syntax error.  Instead of searching for r'<?xml' we should look for r'<\?xml' (i.e., the question mark character should be a literal).


kmike:
Thanks!

PR #421, opened 2013-06-22
Added checking for zero depth in lch_similarity

michelleful:
lch_similarity method raises a ZeroDivisionError when a synset has no hypernyms, since LCH distance is -log(p/2d), so when d == 0, it throws a ZeroDivisionError.

I added checking for when depth == 0 and had the function return None in this case, as LCH similarity is undefined in this instance.


stevenbird:
Thanks for this. But you also included wordnet.py~

michelleful:
I've now deleted wordnet.py~

stevenbird:
Thanks.

PR #433, opened 2013-07-11
made codebase sensitive to the current python version

stevenxxiu:
...ing on the current python version

use http://nltk.github.com/nltk_data/ or http://nltk.github.com/nltk_data_3/ as the download url depending on the current python version
reuse code from nltk.data to get the default download path, if no existing nltk data directories are present


stevenbird:
I want to discuss the best way of doing this before proceeding. The current solution does not support ongoing changes to the corpus collection that might be aligned with changes to the codebase.

stevenbird:
Closing at the request of @jskda

PR #436, opened 2013-07-12
open gzip files in text mode in the tadm classifier trainer

stevenxxiu:


stevenxxiu:
Thanks for pointing that out, your gzip_open_txt function is better since it's closer to the existing python 3.3 function in the gzip module.
I've fixed this in this pull request: https://github.com/nltk/nltk/pull/440

PR #441, opened 2013-07-15
nltk fixes

stevenxxiu:
fixed gzip open unicode issue
fixed broken data.doctest in windows, and normalization/speedups in data.py's url protocols

### DIFF
--- a/nltk/compat.py
+++ b/nltk/compat.py
@@ -61,6 +61,11 @@ else:
     reload = reload
     raw_input = raw_input
 
+    #python 2.7 pickle compatibility
+    import UserString
+    import collections
+    UserString.defaultdict = collections.defaultdict
+
     from itertools import imap, izip
 
     try:
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -61,7 +61,7 @@ from nltk import compat
 # Search Path
 ######################################################################
 
-paths = []
+path = []
 """A list of directories where the NLTK data package might reside.
    These directories will be checked in order when looking for a
    resource in the data package.  Note that this allows users to
@@ -69,23 +69,26 @@ paths = []
    (e.g., in their home directory under ~/nltk_data)."""
 
 # User-specified locations:
-paths += [d for d in os.environ.get('NLTK_DATA', str('')).split(os.pathsep) if d]
-if os.path.expanduser('~/') != '~/': paths += [
-    os.path.expanduser(str('~/nltk_data'))]
+path += [d for d in os.environ.get('NLTK_DATA', str('')).split(os.pathsep) if d]
+if os.path.expanduser('~/') != '~/':
+    path.append(os.path.expanduser(str('~/nltk_data')))
 
-# Common locations on Windows:
-if sys.platform.startswith('win'): paths += [
-    str(r'C:\nltk_data'), str(r'D:\nltk_data'), str(r'E:\nltk_data'),
-    os.path.join(sys.prefix, str('nltk_data')),
-    os.path.join(sys.prefix, str('lib'), str('nltk_data')),
-    os.path.join(os.environ.get(str('APPDATA'), str('C:\\')), str('nltk_data'))]
-
-# Common locations on UNIX & OS X:
-else: paths += [
-    str('/usr/share/nltk_data'),
-    str('/usr/local/share/nltk_data'),
-    str('/usr/lib/nltk_data'),
-    str('/usr/local/lib/nltk_data')]
+if sys.platform.startswith('win'):
+    # Common locations on Windows:
+    path += [
+        str(r'C:\nltk_data'), str(r'D:\nltk_data'), str(r'E:\nltk_data'),
+        os.path.join(sys.prefix, str('nltk_data')),
+        os.path.join(sys.prefix, str('lib'), str('nltk_data')),
+        os.path.join(os.environ.get(str('APPDATA'), str('C:\\')), str('nltk_data'))
+    ]
+else:
+    # Common locations on UNIX & OS X:
+    path += [
+        str('/usr/share/nltk_data'),
+        str('/usr/local/share/nltk_data'),
+        str('/usr/lib/nltk_data'),
+        str('/usr/local/lib/nltk_data')
+    ]
 
 
 ######################################################################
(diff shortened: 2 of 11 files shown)


nltk/compat.py
```diff
@@ -61,6 +61,11 @@ def u(s):
     reload = reload
     raw_input = raw_input
 
+    #python 2.7 pickle compatibility
+    import UserString
+    import collections
+    UserString.defaultdict = collections.defaultdict
```
kmike, line 67: Could you please elaborate more on this? What is it for?

stevenxxiu, line 67: I re-pickled the data files using python 3 by using code like the following:

```
pickle.dump(destpath,pickle.load(srcpath,encoding='latin-1'),protocol=2)
```

If there's a `collections.defaultdict` in the pickle dump, python 3.3 pickles it to `UserString.defaultdict` instead of `collections.defaultdict`. I'm however not sure why this is, but python 2.7 and 2.6 do not have a `defaultdict` class in `UserString`, so I've added that as a compatibility fix.

nltk/data.py
```diff
@@ -69,24 +69,74 @@
...
+        if protocol == 'file':
+            protocol = 'file:///'
+            name = normalize_resource_name(name)
+        elif protocol == 'http':
+            protocol = 'http://'
+        elif protocol == 'nltk':
+            protocol = 'nltk:'
+            name = normalize_resource_name(name)
+        else:
+            raise ValueError('Unknown resource protocol', protocol)
```
kmike: Am I missing something or we can just return a value from this branch instead of using ValueError exceptions for the control flow?

nltk/compat.py
kmike, line 67: Hmm, that's quite strange because Python 3.3 doesn't have UserString module.

nltk/data.py
stevenxxiu: You're right, I've fixed that in the latest commit on my branch, I'll add another pull request that does this part better.
Here's the new code:

```
def split_resource_url(resource_url):
    """
    Divide the resource name into "<protocol>:<path>".
    """
    return re.split(r':/*', resource_url, 1)

def normalize_resource_url(resource_url):
    try:
        protocol, name = split_resource_url(resource_url)
    except ValueError:
        # the resource url has no protocol, use the nltk protocol by default
        protocol = 'nltk'
        name = resource_url
    if protocol == 'file':
        protocol = 'file:///'
        name = normalize_resource_name(name)
    elif protocol == 'nltk':
        protocol = 'nltk:'
        name = normalize_resource_name(name)
    else:
        # handled by urllib
        protocol += '://'
    return ''.join([protocol,name])
```

nltk/compat.py
stevenxxiu, line 67: Yes indeed. Try the following:

```
pickle.dump(collections.defaultdict(),open('test.pickle','wb'),protocol=2)
```

Then dump the pickle file:

```
python -m pickletools test.pickle
```

kmike, line 67: Cool. I believe it is a Python bug. 

```
In [6]: dct = collections.defaultdict()

In [8]: pickle.dumps(dct, protocol=2)
Out[8]: b'\x80\x02cUserString\ndefaultdict\nq\x00)Rq\x01.'

In [10]: pickle.dumps(dct, protocol=2, fix_imports=False)
Out[10]: b'\x80\x02ccollections\ndefaultdict\nq\x00)Rq\x01.'
```

The cause of error is here: http://hg.python.org/cpython/file/7272ef213b7c/Lib/_compat_pickle.py#l80

We could try fix_imports=False - it will work as soon as we don't use moved functions from stdlib directly. It should work if we e.g. import urlencode from compat.py. 

Another way is to create workarounds in compat.py (as you've done in your pull request). I believe we should also add defaultdict to UserList in this case because REVERSE_IMPORT_MAPPING relies on dict order that can vary between runs. But these workarounds are weird. Maybe it is better to patch REVERSE_IMPORT_MAPPING directly because this makes workaround more clear. 

But I hope fix_imports=False would work.

kmike, line 67: reported here: http://bugs.python.org/issue18473

stevenxxiu, line 67: Yup you're right, good job on finding the bug.

kmike, line 67: I was wrong about "it will work as soon as we don't use moved functions from stdlib directly. It should work if we e.g. import urlencode from compat.py.", please disregard this

kmike, line 67: but fix_imports=False could still work and I think we should try it first

nltk/data.py
```diff
@@ -61,7 +61,7 @@
 # Search Path
 ######################################################################
 
-path = []
+paths = []
```
kmike: This is backwards-incompatible, because this variable was public and documented (see http://nltk.org/api/nltk.html?highlight=nltk.data#nltk.data.path), and people seems to use it, e.g.: http://dragonlist.wordpress.com/2011/04/30/nltk-add-new-corpus/, http://stackoverflow.com/questions/3522372/how-to-config-nltk-data-directory-from-code.

Singular "path" is common, e.g. Python's sys.path or PATH environment variable. I think we shouldn't change this name.

stevenxxiu: Fair enough I'll change that back.

kmike: Thanks!

stevenxxiu: What do you think about using a public `path` variable as an alias to a private `_paths` variable, as the latter feels more natural, especially in for loops?

kmike: I don't think it worths it. It'll introduce extra complexity and edge cases. For example, somebody could replace nltk.data.path variable with a new list, and we'll be still using old list via _paths reference. 

That said, if you think 'paths' argument is better for 'find' function then go for it. 

Btw, I think that find's 'paths' argument should be None by default, not 'path', because of the same reason (we'd better resolve this variable at run time, not at import time).

nltk/data.py
kmike: Some doctests for that (and for normalize_resource_name) would be great :)  (put them into doctstrings) 

The tricky part is that they'll depend on os.path.sep, but I'm not sure NLTK is currently testable under Windows, and there are no other doctests in nltk.data. I think it could be OK to just skip nltk.data doctests under Windows.

stevenxxiu: I'm actually testing it under windows, and most of the changes I introduced are for cross-platform compatibility. I think I now have the same tests passing under windows as linux. Normalize converts `os.path.sep` to forward slashes, as the file protocol requires that anyway.

nltk/data.py
stevenxxiu: Good point, I've fixed both. Thanks.

stevenbird:
@kmike thanks for your feedback

stevenbird:
Sorry about this. @jskda please take a look at these.

stevenxxiu:
Sorry I requested a merge without testing it properly, I'm trying to fix this asap.

Issue #445, opened 2013-07-17
test_lowest_common_hypernyms fails


kmike:
test_lowest_common_hypernyms introduced in https://github.com/nltk/nltk/commit/609fe05dfcf32c0b11a7de82124f02cf86f0434b fails for me (both under Python 2.7 and 3.3):

```
> tox -e py27-nodeps -- nltk.test.unit.test_wordnet
GLOB sdist-make: /Users/kmike/svn/nltk/setup.py
py27-nodeps inst-nodeps: /Users/kmike/svn/nltk/.tox/dist/nltk-3.0a2.zip
py27-nodeps runtests: commands[0]
F
======================================================================
FAIL: This tests wordnet's lowest_common_hypernyms method. It should
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/kmike/svn/nltk/.tox/py27-nodeps/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/kmike/svn/nltk/nltk/test/unit/test_wordnet.py", line 37, in test_lowest_common_hypernyms
    assert len(lch) == 1
AssertionError
```

I also think that this test could work better written as doctest (preferable) or as many small unit tests. In both cases we'll be able to see individual "small" failings - now the test stops working on first error. There is no need to put this test to wordnet.doctest (it is slow to run as whole); I think a separate file should be ok if we decide to rewrite it as a doctest.

dougalg:
I'd be happy to do this over the next couple of days if you'd like. Sorry for introducing the error, I was sure I'd run the test a final time before submitting, but I must have done something wrong.

stevenbird:
@kmike is this resolved?

kmike:
Yes, this is fixed by #465.

PR #451, opened 2013-07-26
Changed test_wordnet.py to wordnet_lch.doctest

dougalg:
Possible resolution to [issue 445](https://github.com/nltk/nltk/issues/445)

I tested this on OS X with Python 2.7, and with Python 3.3. In Python 3.3 the test fails with this error:

```
... line 45, in wordnet_lch.doctest
Failed example:
    wn.synset('woman.n.01').lowest_common_hypernyms(wn.synset('girlfriend.n.02'), use_min_depth=True)
Expected:
    [Synset('woman.n.01'), Synset('organism.n.01')]
Got:
    [Synset('organism.n.01'), Synset('woman.n.01')]
```

What is the best way to resolve this problem? I don't know why the list is coming back in a different order in Python 3, but should I just change it to something like this which should work for both?

```
set(wn.synset('woman.n.01').lowest_common_hypernyms(wn.synset('girlfriend.n.02'), use_min_depth=True))
```

Personally, I'd rather if it "just worked". Thank you for your advice. Also, I'm not sure that the version numbers I mentio are correct.


stevenbird:
+1 to make lowest_common_hypernyms() return a set

PR #454, opened 2013-08-03
Fix bug in lemma name parsing

nschneid:
Lemmas containing periods were causing an exception on full name lookup: wordnet.lemma('.45_caliber.a.01..45_caliber'), wordnet.lemma('aquinas.n.01.St._Thomas'), etc. Instead of splitting on the last period, we look for two digits (the sense number) surrounded by periods.


stevenbird:
Thanks for the fix.

PR #455, opened 2013-08-05
Added stdout decoding in weka module fo Py3 compatibility.  Corrected misspelled classifier java class.

tibidoh:
As discussed at PR #410.
Change is not backward compatible. Now we return Unicode strings on Py2 instead of regular ones.

### DIFF
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -9,7 +9,6 @@
 Classifiers that make use of the external 'Weka' package.
 """
 from __future__ import print_function
-
 import time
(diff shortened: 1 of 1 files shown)


kmike:
LGTM. That's desirable to return unicode in Python 2.x. 
I can't test this PR because of my Weka issues; merging it anyways :) 

We can get encodings wrong here (though I don't think so), but user code is also likely to be subtly wrong. After this change it will at least fail with an exception if assumptions are incorrect, and we'll be able to fix that.

PR #456, opened 2013-08-06
added new methods in data.py which should be exposed to __all__

stevenxxiu:


kmike:
Could you please give an example when would user want to use these functions directly? 

split_resource_url, normalize_resource_url and normalize_resource_name looks like internal functions to make nltk.data.load work; they are e.g. special-casing 'nltk' protocol. Instead of promoting them to `__all__` I'd make them private and prepend their names with "_". It won't remove anything from NLTK (these functions are still available for direct import), but we'll make less commitments this way: more focused public APIs means more freedom in future development.

PR #457, opened 2013-08-06
fixed some encoding problems portuguese_en.doctest

stevenxxiu:
...s in the nltk book were encoded incorrectly in the doctest file


kmike:
Thanks!

Issue #459, opened 2013-08-15
parse.generate.generate() Can't generate strings for non-binarized grammars


syntactic:
The generate() function, which generates strings top down from the start symbol of a CFG, gives errors when attempting to generate the strings from a grammar with ternary rules, such as S -> NP VP PP. The _multiply function in generate.py attempts to concatenate sets of constituents, but it will, when using a grammar with non-binarized rules, attempt to concatenate strings to lists, resulting in a TypeError. Here's the stack trace that I get:

``` python
Traceback (most recent call last):
  File "nltkTest.py", line 36, in <module>
    for sent in list(chain.from_iterable(generate.generate(grammar))):
  File "/Library/Python/2.7/site-packages/nltk/parse/generate.py", line 15, in generate
    return _generate_all(grammar, [start])[0]
  File "/Library/Python/2.7/site-packages/nltk/parse/generate.py", line 22, in _generate_all
    frags.append(_generate_all(grammar, prod.rhs()))
  File "/Library/Python/2.7/site-packages/nltk/parse/generate.py", line 22, in _generate_all
    frags.append(_generate_all(grammar, prod.rhs()))
  File "/Library/Python/2.7/site-packages/nltk/parse/generate.py", line 28, in _generate_all
    for frag in _multiply(frag1, frag2):
  File "/Library/Python/2.7/site-packages/nltk/parse/generate.py", line 40, in _multiply
    frags.append(f1+f2)
TypeError: can only concatenate list (not "str") to list
```

An example grammar where this fails is the following:
S -> RelativeMinutes OneToTwelve | OneToTwelve AMPM | OneToTwelve OClock | OneToTwelve O OneToNine | OneToTwelve TenToFiftyNine | OneToTwelve O OneToNine AMPM | OneToTwelve TenToFiftyNine AMPM
RelativeMinutes -> 'twenty to' | 'quarter to' | 'ten to' | 'five to' | 'half past'
OneToTwelve -> 'one' | 'two' | 'three' | 'four' | 'five' | 'six' | 'seven' | 'eight' | 'nine' | 'ten' | 'eleven' | 'twelve'
AMPM -> 'am' | 'pm'
OClock -> 'oclock'
OneToNine -> 'one' | 'two' | 'three' | 'four' | 'five' | 'six' | 'seven' | 'eight' | 'nine'
TenToFiftyNine -> ElevenToNineteen | Tens OneToNine
O -> 'o'
[...]

PR #460, opened 2013-08-18
Bugfix and rewrite of CFG generation

heatherleaf:
While fixing issue #459, I added some extra options, and docstrings.

I don't see any reason for keeping generate2.py, so I removed it.


stevenbird:
Thanks! It would be nice to have some tests in nltk/test if possible.

PR #461, opened 2013-08-18
Updates to scikit-learn wrapper in nltk.classify

larsmans:
- Dropped support for scikit-learn 0.11 in the example;
- added missing docstrings and generally improved the docs.

### DIFF
--- a/nltk/classify/scikitlearn.py
+++ b/nltk/classify/scikitlearn.py
@@ -133,7 +133,8 @@ class SklearnClassifier(ClassifierI):
                     pass
 
         shape = (i + 1, len(self._feature_index))
-        return coo_matrix((values, (i_ind, j_ind)), shape=shape, dtype=self._dtype)
+        return coo_matrix((values, (i_ind, j_ind)), shape=shape,
+                          dtype=self._dtype)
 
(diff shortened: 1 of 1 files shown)


nltk/classify/scikitlearn.py
```diff
@@ -69,24 +68,42 @@ def __repr__(self):
         return "<SklearnClassifier(%r)>" % self._clf
 
     def batch_classify(self, featuresets):
+        """Classify a batch of samples.
+
+        :param labeled_featuresets: A list of featuresets, each a dict
+            mapping strings to either numbers or booleans.
+        :return: The predicted class label for each input sample.
+        :rtype: List.
```
kmike: I think it is better to use type name as-is in `:rtype:` declarations ('list', not 'List')

kmike:
Looks good, thanks! +1 for dropping support for scikit-learn 0.11.

nltk/classify/scikitlearn.py
larsmans: Force-pushed a new version with this changed.

stevenbird:
Thanks

Issue #450, opened 2013-07-25
Use scikit-learn for nltk.classify.svm or deprecate/remove nltk.classify.svm


kmike:
I think that nltk.classify.svm does more harm than good currently, mainly because it is based on https://bitbucket.org/wcauchois/pysvmlight. It has the following issues:
- license restrictions (only for non-commercial use) - documentation ("For license information, see LICENSE.TXT") is incorrect regarding this, by the way;
- pysvmlight doesn't install in Python 3.x; another installation issue is reported [here](https://groups.google.com/d/msg/nltk-users/34fjbwCUHTI/kVfimh7xGyIJ);
- it segfaults. This was reported e.g. at https://github.com/nltk/nltk/issues/170, https://github.com/nltk/nltk/issues/261#issuecomment-6861718 and [here](https://groups.google.com/forum/?fromgroups#!topic/nltk-users/qEct8uzpjnY).

What do you think about removing it and suggesting to use nltk.classify.scikitlearn together with LinearSVC (or scikit-learn directly) instead?

Another option would be to support the existing interface, but change the implementation to use nltk.classify.scikitlearn. It looks like more work. They both have feature mapping implemented, but in different ways; I'm not sure it worths it to support the mapping method implemented in nltk.classify.svm.

leondz:
I think this is a good idea. Better integration with scikit-learn has given us many options for SVM classification, compared to the zero we had when I added this module. Pysvmlight appears to be unmaintained, and so the situation is unlikely to improve - on top of the license restrictions. We now have multiple choices for using SVM in NLTK, and this one is sub-optimal.

stevenbird:
+1 yes I think it is a good idea to drop nltk.classify.svm

Narts:
+1  scikit-learn has given us many options for SVM classification

larsmans:
If by feature mapping, you mean taking features with string values and converting them to one-of-K boolean features, that can be added in two ways:
- by adding it to the scikit-learn wrapper directly, or
- by using a scikit-learn `DictVectorizer` in the wrapper.

[...]

PR #462, opened 2013-08-19
Support for non-numeric features in scikit-learn wrapper, using DictVectorizer

larsmans:
As discussed at #450. Also simplified the code a great deal, fixed some errors in the docstrings and improved the example a little bit.

I did not expose the underlying data structures yet.


stevenbird:
This looks much cleaner.

kmike:
Thanks, looks very good! I was also thinking of changing this code to use DictVectorizer, but never got time to do that.

PR #463, opened 2013-08-19
Doctests for CFG generation

heatherleaf:
I create a new doctest file -- is there anything else I need to do, like modifying a list somewhere in another file?


stevenbird:
That's all we need, thank you!

PR #464, opened 2013-08-20
installation via puppet

deanmalmgren:
I'm not sure if this is the best place to put this or whether this is even the appropriate forum for it, but I thought that the NLTK user community would like to know about (and contribute to) the [puppet-nltk project](https://github.com/datascopeanalytics/puppet-nltk) that we are working on to make it easy to install nltk on many machines in an automated and reliable way. 

If there is a better place to include this or a better forum for this kind of thing, please let me know and I'm happy to use that alternative channel instead.

### DIFF
--- a/web/data.rst
+++ b/web/data.rst
@@ -47,3 +47,17 @@ Windows: Use the "Run..." option on the Start menu.  Windows Vista users need to
 Test the installation: Check that the user environment and privileges are set correctly by logging in to a user account,
 starting the Python interpreter, and accessing the Brown Corpus (see the previous section).
 
+.. _puppet_data_installation:
+
+Puppet installation
+-------------------
+
+If you are provisioning many machines with NLTK, it can be convenient
+to use a tool like puppet to download any necessary data on each of
+these machines in a predictable and automated way. The `puppet NLTK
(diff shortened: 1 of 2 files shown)


kmike:
Hi @deanmalmgren,

puppet-nltk looks nice! But I'm -1 on adding these instructions to docs - they looks a bit too specific. Most people don't need nltk-puppet, and those who need it may prefer chef or ansible or other orchestration framework. What nltk-puppet solves is definitely not a kind of problem first-time NLTK users solve.

@stevenbird what do you think about creating "related projects" page in Wiki, or just putting a list of related projects to the main wiki page?

deanmalmgren:
Completely understood, @kmike. If you and @stevenbird have any other ideas for places to mention things like this, I'd be happy to include it there.

stevenbird:
Thanks @deanmalmgren; I've added a link to puppet-nltk from the nltk github wiki.

deanmalmgren:
Sounds great, Steven. Thanks!

stevenbird:
Oh, note that the plural of "corpus" is "corpora" (not "corpi" as on your webpage).

deanmalmgren:
haha. just updated now. Thanks for the catch

Issue #458, opened 2013-08-12
wordnet lowest_common_hypernyms() returning different results in Python 2.7 vs 3.3 (doctest failing in py 3.3 on OS X)


dougalg:
As in [pull request 451](https://github.com/nltk/nltk/pull/451), sometimes wordnet lch returns differently ordered lists in python 2.7 vs 3.3 on OS X.

```
... line 45, in wordnet_lch.doctest
Failed example:
    wn.synset('woman.n.01').lowest_common_hypernyms(wn.synset('girlfriend.n.02'), use_min_depth=True)
Expected:
    [Synset('woman.n.01'), Synset('organism.n.01')]
Got:
    [Synset('organism.n.01'), Synset('woman.n.01')]
```

Possible fix is to return a set rather than a list, but that seems to be a major change for a small problem. Any suggestions? I'll try finding the source of the difference, anyway.

kmike:
It is most likely caused by iteration over dict or set somewhere. I think this can be solved either by sorting the results or by using OrderedDict instead of dict where it happens.

fcbond:
I think that conceptually, it should be a set: there is no specified order.  Therefore making it a set is a better solution than arbitrarily ordering them: the same solution as in 451.

pquentin:
Lines 511 and 514 indeed [iterate over a set](https://github.com/nltk/nltk/blob/3f1ece3fee6bc425c42dd858f36ca781bd496621/nltk/corpus/reader/wordnet.py#L514).

As said by @fcbond, @stevenbird suggested that the function should return a set in #451, but @kmike suggests sorting. In both cases, it's a simple fix: what should be done?

Since the behavior has changed (see `use_min_depth`), it could be a good idea to explicitely change the interface too.

kmike:
I'm fine with lowest_common_hypernyms returning set. I think it is better to be consistent: if we return sets then also return sets in Synset.common_hypernyms and maybe other methods.

stevenbird:
Yes, all of the synset and lemma methods would need to return sets. In most cases, the sequence comes straight from the wordnet file, so the above-mentioned problem would never arise, and we are doing extra processing for no reason (other than theoretical correctness). A downside is that user code would become non-deterministic. I prefer a more conservative fix, namely sorting the result of lowest_common_hypernyms().

fcbond:
On Thu, Aug 22, 2013 at 4:46 PM, Steven Bird notifications@github.com wrote:

[...]

PR #465, opened 2013-08-22
Sort lowest common hypernyms by Synset name

pquentin:
Proposal for sorting synsets in lowest_common_hypernyms: see issue #458


pquentin:
Is there anything else I need to do? This was "doctested" successfully on both Python 2.7 and Python 3.3.

stevenbird:
Sorry for the delay. I think this solution is a nice compromise (compared with using sets across the board, a major change).

PR #467, opened 2013-08-25
New data package versioning format, cf https://github.com/nltk/nltk/pull/359

stevenxxiu:


stevenxxiu:
I haven't tested this on linux, to test this the data packages for the new versioning format are required, see my branch of nltk_data: https://github.com/jskda/nltk_data

stevenxxiu:
updated the pull request

stevenbird:
I would prefer to use this path-mapping approach for corpus versioning and for coordinating corpus versions with NLTK versions, and not to use it for the Python version issue.

PR #469, opened 2013-08-26
Changed frame() to accept either a frame id or frame name

wooters:
I created two new functions: frame_by_name() and frame_by_id() and changed the frame() function to accept either a str or an int and it will call either frame_by_name() frame_by_id() depending on the type of arg. (Change suggested by Nathan Schneider.)

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -14,9 +14,219 @@ __docformat__ = 'epytext en'
 
 import os
 import re
+import textwrap
 from nltk.internals import ElementWrapper
 from nltk.corpus.reader import XMLCorpusReader, XMLCorpusView
+from nltk.compat import string_types
 
+def _pretty_longstring(defstr, prefix='', wrap_at=65):
+
+    """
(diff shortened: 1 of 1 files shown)


nltk/corpus/reader/framenet.py
```diff
@@ -263,30 +331,26 @@ def frame(self, fn_fid, ignorekeys=[]):
...
         try:
-            name = self._frame_idx[fn_fid]['name']
-        except TypeError:
-            self._buildframeindex()
-            name = self._frame_idx[fn_fid]['name']
-        except KeyError:
-            raise FramenetError('Unknown frame id: {0}'.format(fn_fid))
+            arg_is_strtype = isinstance(fn_fid_or_fname, basestring)
+        except NameError: # python3 doesn't have basestring
+            arg_is_strtype = isinstance(fn_fid_or_fname, str)
```
kmike: This is fine, but the code can be simplified a bit by using nltk.compat.string_types

wooters: Thanks. I didn't know about that. I'll fix it.

PR #473, opened 2013-09-01
framenet.py: pretty printing and API interface improvements

nschneid:


stevenbird:
Thanks @nschneid. I'll wait to see what @wooters says before merging these changes.

nltk/corpus/reader/framenet.py
```diff
@@ -241,12 +247,12 @@ class AttrDict(dict):
 
     >>> foo = {'a':1, 'b':2, 'c':3}
     >>> bar = AttrDict(foo)
-    >>> bar
+    >>> dict(bar)
```
kmike, line 250: these doctests could fail because dict order shouldn't be relied on

nltk/corpus/reader/framenet.py
```diff
@@ -275,6 +287,44 @@ def __str__(self):
...
+            parts.append(kv)
+        return '{'+', '.join(parts)+'}'
+
+class PrettyList(list):
+    """
+    Displays an abbreviated repr of only the first several elements, not the whole list.
+    """
+    # from nltk.util
+    _MAX_REPR_SIZE = 60
+    def __repr__(self):
```
kmike, line 315: `__repr__` returns unicode here, but unfortunately in Python 2.x `__repr__` must return bytestrings (and for work without issues these strings should be ASCII in Python 2.x). There is nltk.compat.python2_unicode_compatible class decorator that could help, but I'm not sure how will it sork for a list subclass without `__str__` method overridden.

nltk/corpus/reader/framenet.py
```diff
@@ -1018,53 +1049,55 @@ def frame_relations(self):
...
+        >>> from nltk.corpus import framenet as fn
         >>> ferels = fn.fe_relations()
-        >>> type(ferels)
-        <type 'list'>
+        >>> isinstance(ferels,list)
+        True
         >>> len(ferels)
         10020
-        >>> ferels[0]
+        >>> dict(ferels[0])
```
kmike, line 1076: doctest can fail because of dict ordering

nltk/corpus/reader/framenet.py
nschneid, line 250: If that's the case then it affects most of the doctests in this module, whether they are checking dicts themselves or lists produced by iterating over dict keys (without sorting them first).

kmike, line 250: Yes, you're right; I just noticed it here.

nltk/corpus/reader/framenet.py
nschneid, line 315: Not sure I follow...in Python 2.7 I'm getting `str` results, not `unicode`:

``` python
>>> type(framenet.frames('Pred'))
<class 'nltk.corpus.reader.framenet.PrettyList'>
>>> repr(framenet.frames('Pred'))
'[<frame ID=300 name=Predicament>, <frame ID=425 name=Predicting>]'
>>> type(framenet.frame('Predicament').FE)
<class 'nltk.corpus.reader.framenet.PrettyDict'>
>>> repr(framenet.frame('Predicament').FE)
"{'Domain': <fe ID=11950 name=Domain>, 'Place': <fe ID=9900 name=Place>, 'Experiencer': <fe ID=2589 name=Experiencer>, 'Time': <fe ID=2593 name=Time>, 'Situation': <fe ID=2590 name=Situation>, 'Cause': <fe ID=2591 name=Cause>}"
```

I'm not sure what happens with non-ASCII characters....

nschneid, line 315: As far as I can tell, there are no names (of frames, frame elements, lexical units, ...) in the FrameNet lexicon v. 1.5 containing a non-ASCII character. The FrameNet website has LUs fiancé.n and fiancée.n, but these are more recent than the 1.5 release.

kmike, line 315: That's a bit tricky. `repr(framenet.frames('Pred'))` returns str because `repr` built-in function converts the result of `__repr__` to str. And this conversion is an issue, because it is performed implicitly using `sys.getdefaultencoding()` (which is often ascii) and this could fail if default encoding can't encode unicode string `__repr__` returns. 

Users don't call `repr` daily, but they could check objects in REPL - `__repr__` is also used for that, and the result of `__repr__` is also decoded using `sys.getdefaultencoding()` in this case, but the correct encoding is `sys.stdout.encoding`, so checking object in REPL could fail.

Why do you need those `text_type` calls anyway? Won't `str` work?

kmike, line 315: .. and non-ascii bytestrings could also fail for `__repr__` in Python 2.x - there is a reason why `repr` returns ASCII for all built-in container types. In console the correct encoding is sys.stdout.encoding, and if we e.g. return utf8-encoded bytestrings in `__repr__` and stdout encoding is not utf8 (this is often the case e.g. in non-English Windows) then REPL will still be broken.

nschneid, line 315: The whole function is simply copied from https://github.com/nltk/nltk/blob/master/nltk/util.py#L712. Is there a reason `text_type` should be used there but not here?

kmike, line 315: Yes, class you're referring to uses `nltk.compat.python_2_unicode_compatible` decorator. This decorator takes care of encodings and requires Python 3.x semantics for `__str__` and `__repr__` (they must return unicode when this decorator is used).

nschneid, line 315: Ah. So is it better to apply that decorator here, or to remove the use of `text_type`?

kmike, line 315: It is better to apply the decorator. I realized that just removing `text_type` usages won't fix issues with non-ascii `__repr__` and REPL because bytestrings would be in incorrect encoding.

kmike, line 315: Default list's `__str__` (that is not overridden) returns bytes, but it shouldn't be an issue for `nltk.compat.python_2_unicode_compatible` decorator because these bytes are ASCII and implicit decoding would work as reliably as it could.

nschneid, line 315: Added the decorator in 9b66f59.

kmike, line 315: Thanks!

wooters:
These changes look great! Your approach to pretty printing is much better than mine @nschneid. As far as I'm concerned @stevenbird you can merge at your convenience.

@kmike regarding the issue with doc testing of dicts, what is the best way to do that? Would something like this be ok:

```
list(sorted(my_dict.items()))
```

Thanks very much!

kmike:
@wooters yes, `sorted(my_dict.items())` will work fine. Another way is to use pprint.pprint - it also sorts dict items by key, and doctests could look a bit nicer with pprint.pprint. Thanks for contributing!

wooters:
Ah, thanks @kmike !

stevenbird:
We have merge conflicts on corpus/reader/{semcor.py, nombank.py, **init**.py}

nschneid:
Sorry, those should not be relevant to this patch. Can you override them
and just merge framenet.py?

stevenbird:
Thanks @nschneid.
Resolved in https://github.com/nltk/nltk/commit/55cafbff53d167c4e6e1aaddba78da4a61e6c8c4

nltk/corpus/reader/framenet.py
nschneid, line 315: @kmike: In trying fixes for #486 we have found that the `python_2_unicode_compatible` decorator removes accents in the result of `__repr__()`. Is there some reason that non-ASCII `__repr__` results are problematic, or is that just a convention (as suggested [here](http://stackoverflow.com/questions/3627793/best-output-type-and-encoding-practices-for-repr-functions#comment3811415_3627835))? Ideally we would like `__repr__` to return UTF-8 strings, so users can see accented characters when entering variables at the interactive prompt.

kmike, line 315: The reason why UTF-8 `__repr__` is problematic is that `__repr__` is used in interactive prompt, and interactive prompt encoding is `sys.stdout.encoding`, not UTF-8. So e.g. in Windows objects with UTF-8 `__repr__` will often raise an exception when checked in console, and this is not something we could fix. We decided to play safe here in NLTK, and make interactive console work for all Python 2.x users. I wrote about that in more details here: http://kmike.ru/python-with-strings-attached/

In Python 3.x `__repr__` returns unicode string by contract (not binary UTF-8-encoded string), and interactive prompt properly encodes this string to `sys.stdout.encoding`, so in Python 3.x this is not a problem, and users will see those accents in Python 3.x.

PR #476, opened 2013-09-04
More efficient batch_parse for MaltParser

wvanlint:
This commit makes MaltParser parse in batches so the binary doesn't have to be executed and the model doesn't have to be loaded for each sentence.

Tested with the following code:
Results are the same except for the batch_parse time being halved.

``` python
import nltk
import timeit
print timeit.timeit('graph = parser.parse(txt)', 'import nltk; parser = nltk.parse.malt.MaltParser(working_dir="<working_dir>", mco="engmalt.linear-1.7", additional_java_args=["-Xmx512m"]); txt = nltk.tokenize.word_tokenize("This is a test sentence")', number=1)
print timeit.timeit('graph = parser.batch_parse([txt, txt])', 'import nltk; parser = nltk.parse.malt.MaltParser(working_dir="<working_dir>", mco="engmalt.linear-1.7", additional_java_args=["-Xmx512m"]); txt = nltk.tokenize.word_tokenize("This is a test sentence")', number=1)
txt1 = "This is a test sentence."
txt2 = "The cat ate the mouse."
txt1_tok = nltk.tokenize.word_tokenize(txt1)
txt2_tok = nltk.tokenize.word_tokenize(txt2)
parser = nltk.parse.malt.MaltParser(working_dir="<working_dir>", mco="engmalt.linear-1.7", additional_java_args=["-Xmx512m"])
graphs = parser.batch_parse([txt1_tok, txt2_tok])
print graphs[0].tree().pprint()
print graphs[1].tree().pprint()
graph = parser.parse(txt1_tok)
print graph.tree().pprint()
graph = parser.raw_parse(txt1)
print graph.tree().pprint()
```

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -11,8 +11,7 @@
 """
 Tools for reading and writing dependency trees.
 The input is assumed to be in Malt-TAB format
-(http://w3.msi.vxu.se/~nivre/research/MaltXML.html).
-Currently only reads the first tree in a file.
+(http://stp.lingfil.uu.se/~nivre/research/MaltXML.html).
 """
 from __future__ import print_function, unicode_literals
 
@@ -121,9 +120,11 @@ class DependencyGraph(object):
     def load(file):
         """
         :param file: a file in Malt-TAB format
+        :return: a list of DependencyGraphs
         """
         with open(file) as f:
-            return DependencyGraph(f.read())
+            return [DependencyGraph(tree_str) for tree_str in
+                                                  f.read().split('\n\n')]
 
     @staticmethod
     def _normalize(line):
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -97,8 +97,22 @@ class MaltParser(ParserI):
         :type sentence: list(str)
         :return: ``DependencyGraph`` the dependency graph representation of the sentence
         """
-        taggedwords = self.tagger.tag(sentence)
-        return self.tagged_parse(taggedwords, verbose)
+        return self.batch_parse([sentence], verbose)[0]
+
+    def batch_parse(self, sentences, verbose=False):
+        """
+        Use MaltParser to parse multiple sentence. Takes multiple sentences as a
+        list where each sentence is a list of words.
+        Each sentence will be automatically tagged with this MaltParser instance's
+        tagger.
+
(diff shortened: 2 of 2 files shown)


nltk/parse/dependencygraph.py
```diff
@@ -121,9 +120,11 @@ def __repr__(self):
     def load(file):
         """
         :param file: a file in Malt-TAB format
+        :return: a list of DependencyGraphs
         """
         with open(file) as f:
-            return DependencyGraph(f.read())
+            return [DependencyGraph(tree_str) for tree_str in
```
heatherleaf, line 126: This is not backwards compatible, and might be a problem. What does @stevenbird think?

heatherleaf:
Otherwise it looks good to me.

stevenbird:
@heatherleaf -- what's the precise compatibility problem please?

pquentin:
@stevenbird Previous version only parsed the first tree/sentence in the file, returning one `DependencyGraph`. This patch now reads all trees and returns a list of `DependencyGraph`s.

stevenbird:
Thanks, so its a change of API, and it is not consistent with the definition of parse() in nltk/parse/api.py.

heatherleaf:
The method in question is not MaltParser().parse(), but instead a static method DependencyGraph.load(). It reads a file in Malt-TAB format. Currently it only returns the first tree in the file, which is strange since the Malt-TAB files usually contains several trees. This pull request suggests to instead read all trees in the file, which is more intuitive to me. 

So, I think the new behaviour is better, but it is an API change. The alternative is to create a new method, load_all() or load_trees() or something.

wvanlint:
Thanks for the comments!
The MaltParser code stays compatible with nltk/parse/api.py.
The API change is as @heatherleaf said in the static method DependencyGraph.load() which now returns a list of `DependencyGraph`.
Previously only one sentence per Malt-TAB file was possible while the specification at http://stp.lingfil.uu.se/~nivre/research/MaltXML.html specifies that multiple sentences can be separated by two newlines.
Currently, in the nltk repository, the load() method is only called in one place namely in the MaltParser class which I edited.
Let me know whether you prefer a new method although I feel that the load method should obey the Malt-TAB specification.

stevenbird:
Thanks for the clarifications, and sorry for the delay.

Issue #261, opened 2012-05-27
Impossible to save SVM classifier (solution inside)


dchaplinsky:
At the moment it's impossible to pickle SVM classifier because it relies on C implementation and pickle is raising an error if you try it.

However I worked out the solution:

``` python

    classifier = SvmClassifier.train(feats)

    write_model(classifier._model, "svn_model.txt")
    fix_model("svn_model.txt")

    classifier._model = None
    with open("classifier.pickle", 'wb') as f:
        pickle.dump(classifier, f)

```

where fix_model routine is referring to this:
https://bitbucket.org/wcauchois/pysvmlight/issue/3/write_model-creates-model-which-is#comment-1470562

Loading is pretty straightforward too:

``` python
with open("models/classifier.pickle", 'r') as f:
    classifier = pickle.load(f)

classifier._model = read_model("svn_model.txt")

```

both read_model and write_model can be found in original pysvmlight package

dcampbell24:
Might a better solution be to wrap the svmlight library with ctypes?

dchaplinsky:
Yeah, probably, as original bindings are reason of often segfaults.

But this will definitely require more work than proposed idea.

kmike:
nltk.classify.svm is no longer in NLTK.

Issue #170, opened 2012-01-17
nltk.classify.svm segfaults on names_demo


alexrudnick:
<b>What steps will reproduce the problem?</b>

leon@ross:~/dev$ py 
Python 2.6.5 (r265:79063, Apr 16 2010, 13:09:56) 
[GCC 4.4.3] on linux2
Type &quot;help&quot;, &quot;copyright&quot;, &quot;credits&quot; or &quot;license&quot; for more information.
&gt;&gt;&gt; from nltk.classify.util import names_demo
&gt;&gt;&gt; from nltk import SvmClassifier
&gt;&gt;&gt; x = names_demo(SvmClassifier.train)
Training classifier...
Segmentation fault
leon@ross:~/dev$ 

<b>What is the expected output? What do you see instead?</b>

Would expect a standard nltk.classify.util.names_demo() output; instead, execution aborts with a Segmentation Fault.
# Test code + stack trace follows.

leon@ross:~/dev$ cat segf.py 
import nltk
from nltk.classify.util import names_demo
from nltk import SvmClassifier
x = names_demo(SvmClassifier.train)

leon@ross:~/dev$ gdb python
GNU gdb (GDB) 7.1-ubuntu
Copyright (C) 2010 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later &lt;http://gnu.org/licenses/gpl.html&gt;
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type &quot;show copying&quot;
and &quot;show warranty&quot; for details.
This GDB was configured as &quot;i486-linux-gnu&quot;.
For bug reporting instructions, please see:
&lt;http://www.gnu.org/software/gdb/bugs/&gt;...
Reading symbols from /usr/bin/python...Reading symbols from /usr/lib/debug/usr/bin/python2.6...done.
done.

(gdb) run segf.py
Starting program: /usr/bin/python segf.py
[Thread debugging using libthread_db enabled]
[New Thread 0xb7e74b70 (LWP 26653)]
Training classifier...

Program received signal SIGSEGV, Segmentation fault.
0x00390c1c in ?? () from /lib/tls/i686/cmov/libc.so.6
(gdb) bt
#0  0x00390c1c in ?? () from /lib/tls/i686/cmov/libc.so.6
#1  0x00393ecd in free () from /lib/tls/i686/cmov/libc.so.6
[...]

PR #480, opened 2013-09-06
deprecate nltk.classify.svm

kmike:
All code from nltk.classify.svm is removed in this PR without providing a compatibility layer. 

`nltk.classify.SvmClassifier` is no longer available (to not confuse people when `dir(nltk.classify)` is called). `nltk.classify.svm.SvmClassifier` is still there, but raises an exception on instantiation.

See GH-450, GH-170 and GH-261


stevenbird:
@kmike I'm accepting this pull request, though it would be ideal if we could use NLTK's standard approach to deprecation (using a decorator).

PR #482, opened 2013-09-07
framenet.py: link together objects rather than storing pointers; new propagate_semtypes() method

nschneid:
Various changes, with updated test cases.

Note that the aim was to link everything together without incurring too much overhead at runtime, so information that would be slow to load is loaded lazily. If they are needed, the semtype hierarchy and frame relations are loaded all at once. Typically names/IDs of frames and LUs are all loaded from index files, with details from the individual frame/LU files filled in as necessary. This is accomplished with a `Future` wrapper that loads its contents on demand.

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -1,7 +1,7 @@
 # Natural Language Toolkit: Framenet Corpus Reader
 #
 # Copyright (C) 2001-2013 NLTK Project
-# Author: Chuck Wooters <wooters@icsi.berkeley.edu>x
+# Author: Chuck Wooters <wooters@icsi.berkeley.edu>
 # URL: <http://www.nltk.org/>
 # For license information, see LICENSE.TXT
 from __future__ import print_function
@@ -12,13 +12,16 @@ Corpus reader for the Framenet 1.5 Corpus.
 
 __docformat__ = 'epytext en'
 
-import os
+import os, sys
 import re
 import textwrap
+from collections import defaultdict
 from pprint import pprint, pformat
 from nltk.internals import ElementWrapper
 from nltk.corpus.reader import XMLCorpusReader, XMLCorpusView
 from nltk.compat import text_type, string_types, python_2_unicode_compatible
+from nltk.util import AbstractLazySequence, LazyMap
+
 
 def _pretty_longstring(defstr, prefix='', wrap_at=65):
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @nschneid!
@wooters would you please test this?

wooters:
It looks like there are some issues running under python3. To test, I ran using:

```
$ python3 framenet.py
```

1) I think line 309 should be:

``` python
if not hasattr(self, '_type'):
```

2) I think line 1950 should be:

``` python
print('  ', [x.name for x in m_frame.lexUnit.values()][:5], '...')
```

instead of:

``` python
print('  ', [x.name for x in m_frame.lexUnit.values()[:5]], '...') 
```

3) This is the last one I ran into and I wasn't quite sure how to resolve it. It appears that `AttrDict().__getitem__()` is throwing a KeyError exception under python3:

```
  File "framenet.py", line 1990, in <module>
    demo()
  File "framenet.py", line 1959, in demo
    pprint(luinfo)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 55, in pprint
    printer.pprint(object)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 136, in pprint
    self._format(object, self._stream, 0, 0, {}, 0)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 193, in _format
    allowance + 1, context, level)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 193, in _format
    allowance + 1, context, level)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 187, in _format
    allowance + 1, context, level)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 193, in _format
    allowance + 1, context, level)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 159, in _format
    rep = self._repr(object, context, level - 1)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 249, in _repr
    self._depth, level)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 261, in format
    return _safe_repr(object, context, maxlevels, level)
  File "/usr/local/Cellar/python3/3.3.2/Frameworks/Python.framework/Versions/3.3/lib/python3.3/pprint.py", line 347, in _safe_repr
    rep = repr(object)
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 411, in __repr__
    pieces.append(elt._short_repr()) # key difference from inherited version: call to _short_repr()
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 296, in _short_repr
    if hasattr(self,'_type'):
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 288, in __getattr__
    return self[name]
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 290, in __getitem__
    v = super(AttrDict,self).__getitem__(name)
KeyError: '_type'
```

nschneid:
How about replacing instances of `hasattr(self, '_type')` with `'_type' in self`?

wooters:
Yup, that fixed it!

I'm getting another error now.... I'll update soon...

wooters:
Looks like it's having a problem with the generator on line 838, but I don't see the problem. Here is the error:

```
Searching for all Frames that have a lemma that matches the regexp: "^run.v$":
  File "framenet.py", line 1989, in <module>
    demo()
  File "framenet.py", line 1986, in demo
    pprint(fn.frames_by_lemma(r'^run.v$'))
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 838, in frames_by_lemma
    return PrettyList(f for f in self.frames() if any(re.search(pat, luName) for luName in f.lexUnit))
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 400, in __init__
    super(PrettyList, self).__init__(*args, **kwargs)
  File "/Users/wooters/src/nltk/nltk/corpus/reader/framenet.py", line 838, in <genexpr>
    return PrettyList(f for f in self.frames() if any(re.search(pat, luName) for luName in f.lexUnit))
  File "/Users/wooters/src/nltk/nltk/util.py", line 901, in iterate_from
    try: yield self._func(self._lists[0][index])
TypeError: 'dict_keys' object does not support indexing
```

nschneid:
I'm not sure of the exact cause here, but it may have something to do with `keys()`: http://stackoverflow.com/questions/8953627/python-dictionary-keys-error

Does eliminating calls to `keys()` help? I think most of them are unnecessary; e.g., `for attr in list(attr_dict.keys()):` can be replaced by `for attr in attr_dict`, right?

wooters:
I tried changing the line you suggested, but it still produces the same error.

The error is actually coming from the `LazyMap` class (in nltk/nltk/util.py) which `PrettyLazyMap()` inherits from. (A `PrettyLazyMap()` object is returned by both `frames()` and `lus()`.)

nschneid:
I think if you wrap all calls to `keys()` with `list()`, it should emulate the Python 2.x behavior.

wooters:
Yup! I did that for the `frames()` and the `lus()` functions and it seems to work just fine now.

So, a practical question - how do we get these changes into your pull request (sorry, still learning how to deal with pull requests...)

nschneid:
I'm not a GitHub expert, so maybe the easiest thing would be for you to email me the updated file and I'll update the pull request.

wooters:
Sounds good. 

But I just noticed something else. At line 1958 we are printing LU info that we obtained via a call to `fn.lu_basic()` which then calls `fn.lu()`. But it looks like under python3 this LU is being printed via `PrettyDict()` whereas under python 2 it is being printed by `_pretty_lu()`.

I suspect it has to do with whether or not the full LU info has been loaded. But I'm not sure what triggers that...

nschneid:
Hmm...what happens if you do `print()` instead of `pprint()`? Also worth checking that `type(luinfo)` is `AttrDict` and `luinfo._type=='lu'`.

wooters:
Ok, I think this may not be an issue. I think it comes down to the difference between `pprint()` under python3 and python2. I tried printing the `luinfo` object using the regular `print()` instead, and they both looked identical...

nschneid:
OK good.

nschneid:
Thanks for testing! Updating the patch now.

stevenbird:
Thanks @wooters and @nschneid. I'll merge this now.

(By the way, either of you could have used "git remote add" to add a new "remote" repository before pulling the other person's work into your working copy, then updating the code, and submitting a new pull request.)

PR #484, opened 2013-09-08
Fixed failing tests in test/framenet.doctest and updated doc strings in ...

wooters:
...framenet.py to match new API

I also made sure that that `test/framenet.doctest` worked for both python 2 and python 3.


stevenbird:
Thanks!

PR #485, opened 2013-09-12
framenet.py: enhance frame_relations(), fix coresets

nschneid:


wooters:
Looks great! Thanks Nathan!

nltk/corpus/reader/framenet.py
```diff
@@ -364,7 +374,9 @@ def __getattr__(self, key):
     def __str__(self):
         return self._data().__str__()
     def __repr__(self):
-        return self._data().__repr__()
+        s = self._data().__repr__()
+        assert isinstance(s,unicode)
```
kmike, line 378: There is no "unicode" in Python 3.x. Also, asserting that `__repr__` returns unicode feels wrong - `__repr__` must return str (UPD, was: bytes). What is this check for?

nltk/corpus/reader/framenet.py
```diff
@@ -1380,10 +1393,14 @@ def frame_relation_types(self):
             self._buildrelationindex()
         return self._freltyp_idx.values()
 
-    def frame_relations(self, frame=None):
+    def frame_relations(self, frame=None, frame2=None, type=None):
```
kmike, line 1396: What about renaming `type` argument to `relation_type`? 

`type` is a builtin, so code like `if not isinstance(type,dict)` reads worse than `if not isinstance(relation_type, dict)` because in the latter case it is clear that relation_type is an user-defined variable.

nschneid, line 1396: I went back and forth on this. Agreed that shadowing a builtin isn't ideal. In practice, because it is third, most calls that provide this argument are going to have to specify the keyword, so something shorter like `rtype` or `reltype` or `typ` strikes me as less annoying to type than `relation_type`. The advantage of `type` is that it's both short and memorable, as it matches the key in the data structure.

If it's just about readability of the function body, would it be better to keep `type` in the signature and then have `relation_type = type` as the first line in the body, so `relation_type` will replace `type` afterwards?

kmike, line 1396: Readability is a subjective question, but IMHO assigning relation_type to type won't improve readability, because the variable name inside the function body won't match the name of function argument, and so to understand where this variable came from developer will have to know the whole function body, not only function signature and the line he's interested in.

I don't think that `relation_type` is that annoying to type, and it reads better than 'rtype'. But I'm fine with rtype or reltype or typ if you think they are better - you're writing this code, so you're the one to paint this bike shed :)

nltk/corpus/reader/framenet.py
stevenbird, line 378: I'm going to revert this change for now.

nltk/corpus/reader/framenet.py
stevenbird, line 1396: For now I have kept the interface (with type=) but changed the implementation to use relation_type (to avoid the odd looking isinstance(type, dict)).

PR #487, opened 2013-09-18
word_tokenizer / TreebankWordTokenizer rather slow

deanmalmgren:
I'm tokenizing several million documents and need to tokenize each of them separately. When tokenizing each of these documents, the repeated calls to re.sub recompiles each of [all these regular expressions on each document](https://github.com/nltk/nltk/blob/master/nltk/tokenize/treebank.py#L61-L87) and is consequently rather slow. Is there a reason that each of these regular expressions aren't pre-compiled as class variables like the [contractions regular expressions](https://github.com/nltk/nltk/blob/master/nltk/tokenize/treebank.py#L46-L57)? 

My understanding is that [pre-compiling is 2-3x faster](http://stackoverflow.com/a/452142/564709).

### DIFF
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -42,6 +42,39 @@ class TreebankWordTokenizer(TokenizerI):
         ['They', "'ll", 'save', 'and', 'invest', 'more', '.']
     """
 
+    # compile a sequence of regular expressions to make the overall
+    # processing faster in situations where we are tokenizing many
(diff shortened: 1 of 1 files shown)


kmike:
I profiled it with line_profiler some time ago, and (if I recall properly) the bottleneck was in contractions handling. I guess that's the reason other regexes are not compiled.

deanmalmgren:
I'm pretty sure this should make things considerably faster.

kmike:
Thanks!

PR #488, opened 2013-09-18
somebody forgot the commas...

deanmalmgren:
Sorry...I forgot the commas and didn't test this before issuing PR #487. Total rookie mistake.

So, I thought I'd do a little speed testing and I was surprised that this wasn't too much faster. In fact, it was almost exactly the same speed.

After a bit more reading (ironically enough in the same SO answer...my bad!), it looks like [python actually caches the regexps](http://stackoverflow.com/a/452143/564709), which explains why this isn't so much faster. Apparently pre-compiling is a misnomer?

If you're inclined to backout the changes in favor of readability, I'd totally understand. Otherwise, this should fix the current push so it actually works.

### DIFF
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -49,30 +49,30 @@ class TreebankWordTokenizer(TokenizerI):
     # and the corresponding substitution pattern.
     SUBSTITUTIONS1 = [
         #starting quotes
-        (re.compile(r'^\"'), r'``')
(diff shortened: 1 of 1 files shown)


kmike:
Oops, I haven't checked it either, sorry about that! And our CI is not noisy enough.. I reverted your commit. Thanks for the heads up. Avoiding re cache can help sometimes (because cache is limited and there could be other regexes cached), but I don't think it matters here.

Issue #479, opened 2013-09-06
Framenet doctests fails


kmike:
To run them, execute `nltk/test/runtests.py nltk/test/framenet.doctest` from nltk checkout:

```
> nltk/test/runtests.py nltk/test/framenet.doctest
F
======================================================================
FAIL: Doctest: framenet.doctest
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/Cellar/python/2.7.5/Frameworks/Python.framework/Versions/2.7/lib/python2.7/doctest.py", line 2201, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for framenet.doctest
  File "/Users/kmike/svn/nltk/nltk/test/framenet.doctest", line 0

----------------------------------------------------------------------
File "/Users/kmike/svn/nltk/nltk/test/framenet.doctest", line 75, in framenet.doctest
Failed example:
    pprint(fn.frames(r'(?i)medical'))
Expected:
    [{'ID': 239, 'name': 'Medical_conditions'},
     {'ID': 256, 'name': 'Medical_specialties'},
     {'ID': 257, 'name': 'Medical_instruments'},
     {'ID': 255, 'name': 'Medical_professionals'}]
Got:
    [<frame ID=239 name=Medical_conditions>, <frame ID=256 name=Medical_specialties>, ...]
----------------------------------------------------------------------
File "/Users/kmike/svn/nltk/nltk/test/framenet.doctest", line 95, in framenet.doctest
Failed example:
    pprint([x.name for x in f.FE])
Exception raised:
    Traceback (most recent call last):
      File "/usr/local/Cellar/python/2.7.5/Frameworks/Python.framework/Versions/2.7/lib/python2.7/doctest.py", line 1289, in __run
        compileflags, 1) in test.globs
      File "<doctest framenet.doctest[11]>", line 1, in <module>
        pprint([x.name for x in f.FE])
    AttributeError: 'str' object has no attribute 'name'
----------------------------------------------------------------------
File "/Users/kmike/svn/nltk/nltk/test/framenet.doctest", line 97, in framenet.doctest
Failed example:
    pprint(f.frameRelation)
Expected:
    [{'_type': 'framerelation', 'relatedFrame': ['Cure'], 'type': 'Uses'}]
Got:
    [[relatedFrame] ['Cure']
    [_type] framerelation
    [type] Uses
    ]
----------------------------------------------------------------------
[...]

PR #491, opened 2013-09-20
framenet.py: doctest fixes for Python 3.x compatibility

nschneid:
see #479

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -1328,16 +1328,9 @@ class FramenetCorpusReader(XMLCorpusReader):
         >>> from nltk.corpus import framenet as fn
         >>> len(fn.documents())
         78
-        >>> pprint(set([x.corpname for x in fn.documents()]))
-        {'ANC',
-         'C-4',
-         'KBEval',
-         'LUCorpus-v0.3',
-         'Miscellaneous',
-         'NTI',
-         'PropBank',
-         'QA',
-         'SemAnno'}
+        >>> set([x.corpname for x in fn.documents()])==set(['ANC', 'C-4', 'KBEval', \
+                    'LUCorpus-v0.3', 'Miscellaneous', 'NTI', 'PropBank', 'QA', 'SemAnno'])
+        True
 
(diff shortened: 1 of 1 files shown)


nltk/corpus/reader/framenet.py
```diff
@@ -1328,16 +1328,16 @@ def documents(self, name=None):
...
-        set(['ANC',
-             'C-4',
-             'KBEval',
-             'LUCorpus-v0.3',
-             'Miscellaneous',
-             'NTI',
-             'PropBank',
-             'QA',
-             'SemAnno'])
+        >>> pprint({x.corpname for x in fn.documents()})
```
kmike: this syntax is not supported in Python 2.6

nltk/corpus/reader/framenet.py
```diff
@@ -1329,15 +1329,15 @@ def documents(self, name=None):
...
-        set(['ANC',
-             'C-4',
-             'KBEval',
-             'LUCorpus-v0.3',
-             'Miscellaneous',
-             'NTI',
-             'PropBank',
-             'QA',
-             'SemAnno'])
+        {'ANC',
```
kmike: I bet this breaks tests in Python 2.x :) Unfortunately set repr is different in 2.x and 3.x, and NLTK test runner doesn't fix that. It seems that comparing two sets is the sanest way to deal with this issue.

nltk/corpus/reader/framenet.py
```diff
@@ -1513,8 +1513,8 @@ def semtypes(self):
         >>> stypes = fn.semtypes()
         >>> len(stypes)
         73
-        >>> stypes[0].keys()
-        ['definition', '_type', 'name', 'rootType', 'abbrev', 'superType', 'subTypes', 'ID']
+        >>> list(sorted(stypes[0].keys()))
```
kmike: list(..) is unnecessary because `sorted` returns list

nschneid:
Oops, I wasn't testing the file I thought I was testing. Tests now pass on 2.7.

kmike:
Thanks! 
One test is still failing in Python 3.3 for me; I'll update #479.

PR #493, opened 2013-09-22
changed self references to super

robwahl:
relextract.py
connection.text_factory cannot be set

mace.py
    def valuation(mbc): return mbc.model('valuation')
first parameter of an accessor is usually called self

chartparser_app.py
            raise
            tkinter.messagebox.showerror('Error Saving Chart',
                                   'Unable to open file: %r' % filename)
filename undefined because it is referenced before assignment

weka.py (I had it unzip in docrunner.py)
def _check_weka_version(jar):
    try:
        zf = zipfile.ZipFile(jar)
    except SystemExit as KeyboardInterrupt:
        raise
    except:
        return None
    try:
        try:
            return zf.read('weka/core/version.txt')
        except KeyError:
            return None
    finally:
        zf.close()

What should the second except be?

inference.py changed for multicore
changed second prover to parallelProverBuilderCommand from BaseModelBuilderCommand
tp_thread = TheoremToolThread(lambda: BaseModelBuilderCommand(self, verbose), verbose, 'TP')
        mb_thread = TheoremToolThread(lambda: ParallelProverBuilderCommand.build_model(self, verbose), verbose, 'MB')

self issues
fixed missed constructor of super class throughout project
method's parameter self reassigned in drt.py
                    self = DRS(self.refs[:i]+[newvar]+self.refs[i+1:],
                               [cond.replace(ref, newvarex, True, alpha_convert)
                                for cond in self.conds],
                               consequent)
and logic.py
                self = self.alpha_convert(unique_variable(pattern=self.variable))
regexp.py
elif re.match('[^{}]_{[^{}]_}[^{}]_', rule):
                left, chunk, right = re.split('[{}]', rule)
changed to re.match('[^{}]\_{[^{}]_}[^{}]_', rule):
This regex is has a dangling metacharacter \* is a reserved character in regexps, so you need to escape it.

align.py return " ".join("%d-%d" % p[:2] for p in sorted(self)) treated as a string and it's not

viterbi.py
print('%11.4f%11d%19.14f' % (time, num_parses, average))
time.py is unexpected I replaced this with a function call

Lossy encoding floesta.txt file was loaded with wrong encoding UTF-8

suggestions:
basestring is not available in python3
compat.py

util.py
   except KeyError:
        print(node, end=' ')
        print("has no specification for the feature %s" % semkey)
        raise
indentation error (must be associated with except)


kmike:
Hi @robwahl! Thanks for raising this issue - you've found some real bugs in NLTK code. But the pull request went wrong - it seems that you've copied nltk folder to "nltk-master" and commited it all. Could you please resubmit this pull request? Just make your changes inplace and commit them. Also, it is easier to review & merge separate pull requests for each problem, so it'll be great if you split this PR into several PRs.

robwahl:
Thanks @kmike for your help. I'm new to Github so I appreciate your comment. I'm going to try the zip file, the connection.text_factory being unwriteable and reg-ex suggestions in the method you described. Thank you. Rob

PR #495, opened 2013-09-26
Sort output of all wordnet methods with undefined order. 

kmike:
This makes wordnet tests pass for all pythons. See also: #465 and #458.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -18,7 +18,7 @@ from collections import defaultdict
 from nltk.corpus.reader import CorpusReader
 from nltk.util import binary_search_file as _binary_search_file
 from nltk.probability import FreqDist
-from nltk.compat import xrange, python_2_unicode_compatible
+from nltk.compat import xrange, python_2_unicode_compatible, total_ordering
 
 ######################################################################
 ## Table of Contents
@@ -95,6 +95,7 @@ class WordNetError(Exception):
     """An exception class for wordnet-related errors."""
 
 
+@total_ordering
 class _WordNetObject(object):
     """A common base class for lemmas and synsets."""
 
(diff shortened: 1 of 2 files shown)


nltk/corpus/reader/wordnet.py
```diff
@@ -237,9 +241,9 @@ def __repr__(self):
 
     def _related(self, relation_symbol):
         get_synset = self._wordnet_corpus_reader._synset_from_pos_and_offset
-        return [get_synset(pos, offset).lemmas[lemma_index]
+        return sorted([get_synset(pos, offset).lemmas[lemma_index]
                 for pos, offset, lemma_index
-                in self.synset._lemma_pointers[self.name, relation_symbol]]
+                in self.synset._lemma_pointers[self.name, relation_symbol]])
```
kmike, line 246: self.synset._lemma_pointers is defaultdict(set) so order was undefined before

nltk/corpus/reader/wordnet.py
```diff
@@ -512,7 +522,7 @@ def lowest_common_hypernyms(self, other, simulate_root=False, use_min_depth=Fals
             else:
                 max_depth = max(s.max_depth() for s in synsets)
                 unsorted_lch = [s for s in synsets if s.max_depth() == max_depth]
-            return sorted(unsorted_lch, key=attrgetter('name'))
+            return sorted(unsorted_lch)
```
kmike, line 525: "key" argument is no longer needed because Synset got comparison methods

nltk/corpus/reader/wordnet.py
```diff
@@ -868,7 +878,7 @@ def __repr__(self):
     def _related(self, relation_symbol):
         get_synset = self._wordnet_corpus_reader._synset_from_pos_and_offset
         pointer_tuples = self._pointers[relation_symbol]
-        return [get_synset(pos, offset) for pos, offset in pointer_tuples]
+        return sorted([get_synset(pos, offset) for pos, offset in pointer_tuples])
```
kmike, line 881: self._pointers is defaultdict(set), so pointer_tuples is set and order was undefined here

stevenbird:
Thanks @kmike

PR #496, opened 2013-09-27
added numpy, scipy, and epydoc to requirements

robwahl:
Expected a string and got a Unicode character instead casted
added numpy, scipy and epydoc to requirements in setup.py
 Changed so "min" would not get confused with built in
toolbox.py wanted tree received list
changed sets data structure because sets is deprecated as of python 2.6
changed "min" to minimum so it would not get confused with shadow built in
changed "file" to "distancefiles" so it would not be confused with shadow built in
changed function type to super


stevenbird:
@robwahl thanks for all this work -- some of it looks very useful. However, it seems like something went wrong; some of these patches are huge. Would you please resubmit minimal patches that can easily be inspected?

PR #497, opened 2013-09-29
SemcorCorpusReader was returning incomplete sense information

nschneid:
semcor.py was creating lemma sense tags as strings of the form `<lemma>.<sensenum>`, which is incorrect: the sense number is actually POS-specific. With this patch, SemcorCorpusReader tries to retrieve the Lemma object directly from WordNet (using the sense key). If that fails, a string of the form `<lemma>.<pos>.<sensenum>` is given.

### DIFF
--- a/nltk/corpus/reader/semcor.py
+++ b/nltk/corpus/reader/semcor.py
@@ -170,7 +170,7 @@ class SemcorCorpusReader(XMLCorpusReader):
                 if sensenum is not None:
                     try:
                         sense = wordnet.lemma_from_key(sense_key)   # Lemma object
-                    except:
+                    except Exception:
                         # cannot retrieve the wordnet.Lemma object. possible reasons:
                         #  (a) the wordnet corpus is not downloaded;
                         #  (b) a nonexistant sense is annotated: e.g., such.s.00 triggers: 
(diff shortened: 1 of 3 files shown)


nltk/corpus/reader/semcor.py
```diff
@@ -162,9 +169,17 @@ def _word(xmlword, unit, pos_tag, sem_tag):
             else:
                 if sensenum is not None:
                     try:
-                        sense = '%s.%02d' % (lemma, int(sensenum))
-                    except ValueError:
-                        sense = lemma+'.'+sensenum  # e.g. the sense number may be "2;1"
+                        sense = wordnet.lemma_from_key(sense_key)   # Lemma object
+                    except:
```
kmike: It is better to use a more specific "except" statement (at least `except Exception`) because `except:` catches all exceptions, even those that are not subclasses of Exception (e.g. KeyboardInterrupt or SystemExit).

nschneid: Good to know! Fixed.

nschneid:
@kmike Can this be merged? It would be nice to have in 3.0 beta. :)

nschneid:
Is there a plan to merge this, @stevenbird?

heatherleaf:
@stevenbird This isn't merged, afaik

stevenbird:
Sorry, this pull request was automatically closed when I renamed the master branch to develop. I've recreated the master branch temporarily so I can reopen the pull request...

stevenbird:
Thanks -- this looks good.

Issue #483, opened 2013-09-07
module or protocol for POS tagset specification and conversion


nschneid:
Given the importance of part-of-speech tags in corpora and NLP applications, it seems that NLTK would benefit from a standard way to **encode, document, and convert among different tagsets**. For example, a module might be added for each tagset that lists all the tags, with a description and examples of each, and provides groupings/morphosyntactic features or mappings to more general POS categories.

Special cases of such functionality have been requested multiple times, e.g.:
- the "universal" tagset of Petrov, Das, & McDonald, #316 
- the BNC tagset as used in FrameNet vs. the Penn Treebank tagset

I often find myself writing one-off functions like `isNominal(pos)` or `isFunctionWord(pos)`, and in a few cases have attempted to systematize particular tagsets (see https://gist.github.com/nschneid/4231292 and https://gist.github.com/nschneid/6476715). @brendano similarly has a conversion script: http://brenocon.com/bnc_ptb.py But I know of no general solution to this problem. There are many subtleties that can arise in implementing such routines, so having a standard implementation of some of the common cases should promote comparability and replicability of systems that group or map POS tags.

To make this easily extensible as new corpora (or standalone tagsets) are added, perhaps there should be a way for a corpus reader to register a tagset with a master list of all NLTK tagsets, and specify conversions where applicable. Corpora and taggers could then make explicit which tagset they use.

Things to consider:
- What information should be available to converters in order to be as accurate as possible? Just the source tag/tagset and target tagset? The source word? Full sentences?
- The interaction between tagsets and tokenization schemes. For example, not all tagsets require clitics like _'s_ to be separated; and some but not all expect hyphens to be separated. Mismatches might call for modifying the tokenization as part of the tag conversion process, though the user should probably have the option to override this.

nschneid:
another resource along these lines: http://ufal.mff.cuni.cz/wiki/user:zeman:interset

kmike:
I don't have experience with English tagsets, but I have some experience with Russian tagsets (https://github.com/kmike/russian-tagsets). Readme is in Russian, sorry :) There are conversion functions for 5 tagsets implemented in that library. The idea is the following:
- each tagset module defines conversion routines to/from other tagsets - they are just functions that take a single tag from one tagset and return a single tag from an another tagset;
- these conversion functions are added to registry;
- there is a "main" function that for a given pair of tagsets returns a function that converts between these tagsets (maybe using an intermediate step(s) - it can find a conversion path because all conversion functions are registered);
[...]

PR #499, opened 2013-10-04
Added partial support for mapping of tagsets.

stevenbird:
Currently only works for the Universal Tagset (Petrov, Das, McDonald, 2012).
Existing support for a simplified tagset (via simplify_tags=True) is dropped.
Access new mappings via tagset='universal', optional flag for corpus reader
methods that return pos-tagged text.

https://github.com/nltk/nltk/issues/483


stevenbird:
I'm merging this, and will soon do another alpha release and get wider testing.

Issue #475, opened 2013-09-04
Consistency with properties


stevenbird:
We should audit the NLTK API to make sure we are being consistent with the use of properties. E.g. why do we have Tree.node and Tree.leaves()?

heatherleaf:
Because .node is a stored value and .leaves() are calculated.

stevenbird:
Right, though this distinction is not transparent to students whose only exposure to programming is a course on Python for NLP.

heatherleaf:
Another distinction that is perhaps easier is that properties are mutable but methods are not. 

But then we get into trouble with ProbabilisticTree.prob(), which is mutable and has a setter method .setprob(x). The reason for this is probably to be consistent with ProbDists which have methods .prob(x), but that is definitely not transparent.

The most consistent would be to get rid of all properties and start using get- and set-methods (.node() resp .setnode())

paulproteus:
@stevenbird , curious if you agree with @heatherleaf 's proposal.

stevenbird:
There's too many options here, and the goal is to keep NLTK pythonic, simple, and consistent. (Simplicity trumps consistency: http://www.jwz.org/doc/worse-is-better.html). Where setting a property requires some logic then I think it makes sense to use get- and set-methods (http://tomayko.com/writings/getters-setters-fuxors).

I don't understand the suggestion to get rid of all properties and start using get- and set-methods instead, since the latter still leads to user code that accesses member variables using the property syntax.

kmike:
Yes, that there are too many options here. Also, any solution that converts methods to properties will break a lot of code, so it is better to handle it before NLTK3. 

This is what we can do for Tree class:
- make height(), leaves(), productions() and pos() properties. We can't make subtrees() property as-is because it has a parameter. Other methods also have parameters, or they make no sense as properties. So still, some attributes will be properties and some will be methods. I'd make some of these functions properties if started from a clean sheet, but this doesn't fix the problem entirely, and changing them now could break a lot of code, including examples from a printed version of NLTK book (e.g. ch8, Weighted Grammar).
- rename some methods to become verbs: get_height(), get_leaves(), get_productions(), get_pos(), get_subtrees(), etc. By this logic students could learn that every method that computes something must be prefixed with 'get_' or named as verb. I think this assumption is wrong, and there are methods named after nouns without 'get_' prefix everywhere in NLTK, in standard library and in other Python packages. This is a respectable approach IMHO, but it could be hard to enforce in practice.
[...]

PR #502, opened 2013-10-09
change tree node property to a method

stevenbird:
changed tree node property to a method, and added a set_node method
(following the model of prob/set_prob etc in probability.py)


stevenbird:
This is to partially address https://github.com/nltk/nltk/issues/475

kmike:
I''m -0 on this because changing 'node' from variable to function is backwards-incompatible. That's why I wrote 'get_node' - we can keep backwards compatibility that way. 

As you mentioned, 'get_node' is not consistent with nltk.probability where 'get_' prefix is not used. 

What do you think about creating a function with some other name, e.g. label(), and keeping 'node' attribute? We could make this attribute property and issue a warning on first read or write access to this attribute.

I'm not an expert, but "node" definition in NLTK Tree looks unusual. Usually node is a data structure that contains some label (named either key or value, depending on application) and a pointer to children (in Python it would be a list of children). So what is called 'node' here is usually called 'node label'. Some references:
- In [Data Structures and Algorithms by Aho, Ullman and Hopcroft](http://www.amazon.com/Data-Structures-Algorithms-Alfred-Aho/dp/0201000237) they distinguish between 'node' and 'node label': nodes can be named, but these names must be unique, while 'node label' or 'node value' is some data associated with the node.
- In  Knuth vol 1 chapter 2 I haven't found this distinction explicit; node labels are called 'nodes' when this is not ambiguous. But when this distinction is important Knuth uses INFO(P) to denote information associated with node pointer P.
- [Wikipedia](http://en.wikipedia.org/wiki/Tree_%28data_structure%29) says that 'each node is a data structure consisting of a value, together with a list of references to nodes (the "children")'.

heatherleaf:
+1 to @kmike's suggestion: `label()` and `set_label()`, and deprecate the `node` property

[edit: of course it should be `set_label` and not `setlabel`, since it is `set_prob`]

stevenbird:
@kmike that's a good point -- the node includes the label, and they are not the same thing. I'll update the pull request (though its going to be a while now as I'm leaving for IJCNLP now.)

stevenbird:
New changes support label() method as proposed by @kmike

nltk/toolbox.py
```diff
@@ -219,7 +219,7 @@ def _record_parse(self, key=None, **kwargs):
     def _tree2etree(self, parent):
         from nltk.tree import Tree
 
-        root = Element(parent.node)
+        root = Element(parent._node)
```
kmike: Is it correct?

stevenbird: I missed that one. I'll fix it and do some more searching.

stevenbird:
In the absence of any controversy I'm merging this now.

PR #517, opened 2013-10-27
Use with ... open() where appropriate instead of just open()

stevenxxiu:
Use with ... open() where appropriate instead of just open(). Also made the names of the opend file handles a bit more consistent.


stevenbird:
This looks good to me.

kmike:
Agree, this is a good change.

There is a similar issue with NLTK corpus readers (they open a lot of files and never close them) - this prevents tests from running on PyPy, but it is unclear how to fix that.

PR #520, opened 2013-10-27
Implemented span_tokenize() method for PunktWordTokenizer.

huderlem:
I noticed that the span_tokenize() method was unimplemented in the PunktWordTokenizer class.  I modeled the implementation of this method after the span_tokenize() method in the PunktSentenceTokenizer class.


huderlem:
Sure thing.  I'll have time in a few hours.

huderlem:
I've never used doctest before, so please let me know if that wasn't what you meant.

stevenbird:
That's fine. Thanks for your contribution!

PR #523, opened 2013-10-28
Add support for unicode terms to nltk.chunk.named_entity.shape()

Arachnid:
Currently, shape(u'Monáe') returns 'other'. This patch fixes it to handle unicode terms better, causing it to return 'mixedcase'. This in turn causes NEChunkParserTagger to correctly tag tokens that have unicode characters in them.

### DIFF
--- a/nltk/chunk/named_entity.py
+++ b/nltk/chunk/named_entity.py
@@ -167,16 +167,17 @@ class NEChunkParser(ChunkParserI):
         return toks
 
 def shape(word):
-    if re.match('[0-9]+(\.[0-9]*)?|[0-9]*\.[0-9]+$', word):
+    if re.match('[0-9]+(\.[0-9]*)?|[0-9]*\.[0-9]+$', word, re.UNICODE):
         return 'number'
-    elif re.match('\W+$', word):
+    elif re.match('\W+$', word, re.UNICODE):
         return 'punct'
-    elif re.match('[A-Z][a-z]+$', word):
-        return 'upcase'
-    elif re.match('[a-z]+$', word):
-        return 'downcase'
-    elif re.match('\w+$', word):
-        return 'mixedcase'
+    elif re.match('\w+$', word, re.UNICODE):
+        if word.istitle():
+            return 'upcase'
+        elif word.islower():
+            return 'downcase'
+        else:
+            return 'mixedcase'
     else:
         return 'other'
 


kmike:
Hi @Arachnid,

re.UNICODE affects only `\w` and `\W` here, for other re.match calls it does nothing; 'upcase' and 'downcase' still won't be returned for appropriate versions of `u'Monáe'`.

Arachnid:
@kmike That's correct, but it's not possible to match on unicode character classes using Python re. This at least recognizes words containing unicode word characters as words, rather than as 'other'.

Arachnid:
I'm happy to remove re.UNICODE from the re calls that don't specifically require it, but I figured consistency was better. Let me know if you'd prefer that.

The regex module (https://pypi.python.org/pypi/regex) supports unicode character class assertions, but I figured adding a new library was disproportionate.

Possibly the relevant regular expressions could be rewritten in Python using the unicode module to look up character classes; I suspect it would be a lot slower.

kmike:
Some notes:
- If I'm not mistaken, NLTK built-in NE chunker uses this function. So if we update it model will use new features values at prediction time (different from the ones used for training) - it could be better to update pickled NE model. But this is not needed if corpus used for training didn't have those unicode words (function should return the same feature values for words in original training data, not for all words). Do you know if this is a problem? @stevenbird ?
- Python's `.istitle()` / `.title()`, `.islower()`/`.lower()` and `.isupper()`/`.upper()` work with unicode correctly, and I think it may be possible to replace regexes with some combination of these functions.

Arachnid:
I can't really comment on point 1, since my experience with NLTK doesn't extend that far. On point 2, I believe you're right; I can update the PR to incorporate this.

Arachnid:
I've updated the commit to use istitle and islower; it should now correctly return upcase/downcase/mixedcase for Unicode-containing words.

Arachnid:
Can you clarify why you're rejecting this? Is there a modification that would make it acceptable?

stevenbird:
Sorry, this pull request was automatically closed when I renamed the master branch to develop. I've recreated the master branch temporarily so I can reopen the pull request...

stevenbird:
@kmike thanks for pointing out that issue, but note that we already transliterated the built-in NE chunker to unicode in https://github.com/nltk/nltk_data/commit/a2d463f2af749983f96d61b5d88cb60390529706

kmike:
@stevenbird what do you mean by "transliterated"?

stevenbird:
@kmike, I should have said "transcoded" -- @jskda loaded the trained models and wrote them out again using utf8.

kmike:
I think that transcoding is orthogonal to the issue with incompatible features. 

Let's imagine there was `Monáe` word in training data. `shape()` function previously returned `'other'` for it, and classifier updated coefficients for `shape:other` feature based on `Monáe` observation. Then we change `shape()` function to return `upcase` for `Monáe` without re-training the classifier. Coefficients for `shape:other`, `shape:upcase` and other features are now incorrect: they won't be the same if we retrain classifier using new `shape()` function and exactly the same data.

This is not an issue if there was no words like `Monáe` in training data.

PR #524, opened 2013-10-28
Don't overwrite sys.meta_path in nltk.compat

Arachnid:
This breaks things for anything else that sets a finder in meta_path, such as pytest.


stevenbird:
Good idea, thanks.

Issue #535, opened 2013-11-01
app/wordnet_app breaks in Python3


stevenbird:
Run wordnet_app.py on the command line, to start up the wordnet browser, and search for "dog" and then click on the first link "domestic dog". The following stacktrace is produced in the terminal:

Exception happened during processing of request from ('127.0.0.1', 50299)
Traceback (most recent call last):
  ...
  File "./wordnet_app.py", line 719, in page_from_href
    return page_from_reference(decode_reference(href))
  File "./wordnet_app.py", line 687, in decode_reference
    string = base64.urlsafe_b64decode(string)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.2/lib/python3.2/base64.py", line 131, in urlsafe_b64decode
    return b64decode(s, b'-_')
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.2/lib/python3.2/base64.py", line 83, in b64decode
    raise TypeError("expected bytes, not %s" % s.__class__.**name**)
TypeError: expected bytes, not str

stevenxxiu:
`get_static_wx_help_page` does not exist anymore in wordnet_app.py, can it be removed?

PR #536, opened 2013-11-02
Fixed some encoding bugs in wordnet_app.py, cf https://github.com/nltk/nltk/issues/535

stevenxxiu:
Fixed some encoding bugs in wordnet_app.py, cf https://github.com/nltk/nltk/issues/535


stevenbird:
@jskda are you testing with both Python 2.6 and 3.2?

stevenxxiu:
@stevenbird: I've tested with python 2.7 and 3.2, I'll test with python 2.6 today.

stevenxxiu:
@stevenbird: I've now tested this with python 2.6 on linux and it works for me.

stevenbird:
Thanks.

Issue #529, opened 2013-10-30
Check find_file() in internals.py for byte vs. string issues


ewan-klein:
The following diff fixes one problem in part of find_file() (though change has not yet been merged in). It would be worth checking through the whole of the function to see if there are other instances where this problem arises.

```
             try:
                 p = subprocess.Popen(['which', alternative], stdout=subprocess.PIPE)
                 stdout, stderr = p.communicate()
-                path = stdout.strip()
+                path = stdout.strip().decode('utf8')
                 if path.endswith(alternative) and os.path.exists(path):
                     if verbose: print('[Found %s: %s]' % (filename, path))
                     return path
```

stevenbird:
@jskda please take a look at this

kmike:
By the way, shouldn't this be `stdout.decode(stdout.encoding).strip()`?

ewan-klein:
Thanks - I stand corrected :-)

PR #538, opened 2013-11-03
decode path byte string

japerk:
without decode(), path.endswith(alternative) will throw an exception, making the which lookup fail
decode() on python2 returns unicode from ascii
decode() on python3 returns string from bytes


stevenbird:
This was already dealt with in https://github.com/nltk/nltk/issues/529

PR #539, opened 2013-11-05
PLEASE CODE REVIEW. This is an implementation of Monetemurro and Zanette...

PeteBleackley:
This is an implementation of Montemurro and Zanette's algorithm for calculating the distribution entropy of words in a body of text.  http://www.worldscientific.com/doi/abs/10.1142/S0219525910002530 This code was originally written to evaluate the algorithm, and probably needs to be refactored before being included in NLTK - in particular, I's like suggestions for a better API

### DIFF
--- a/nltk/entropy/EntropyCalculator.py
+++ b/nltk/entropy/EntropyCalculator.py
@@ -1,0 +1,176 @@
+u"""Module to find keywords using the Montemurro and Zanette algorithm.
+Created by Dr Peter J. Bleackley """
+from math import log,exp,lgamma
+import re
+import codecs
+import urllib
+import sys
+
+
+
+def LogCombinations(x,y):
+    u"""Calculates the logarithm of a binomial coefficient.
+    This avoids overflows. Implemented with gamma functions for efficiency"""
+    result=lgamma(x+1)
+    result-=lgamma(y+1)
+    result-=lgamma(x-y+1)
+    return result
+
+class EntropyCalculator:
+    u"""Class that contains data and methods for calculating the entropy of
+        words in a text"""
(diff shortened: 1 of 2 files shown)


stevenbird:
Please explain why you feel this code belongs in NLTK. Note that any contribution needs to conform to our guidelines (e.g. this code does not conform with PEP-8, and it doesn't use appropriate docstrings, and it doesn't work with Python 3). https://github.com/nltk/nltk/wiki/Developers-Guide

PeteBleackley:
Long documents typically have structure on scales of 500 to 1000 words, corresponding to the development of topics within the discourse. This algorithm measures the information contributed by different words to that structure.

My implementation was originally written as a stand-alone app for evaluation purposes, and its current API reflects that. It needs some refactoring to make it useful as part of a bigger library, and I'd like the advice of the NLTK developer community on that. Obviously, I can address coding-style issues like the ones you have raised in the process. 

We can continue the discussion on nltk-dev

Pete Bleackley

Sent from Samsung Mobile

 Steven Bird notifications@github.com wrote: 

Please explain why you feel this code belongs in NLTK. Note that any contribution needs to conform to our guidelines (e.g. this code does not conform with PEP-8, and it doesn't use appropriate docstrings, and it doesn't work with Python 3). https://github.com/nltk/nltk/wiki/Developers-Guide

ewan-klein:
We evaluate APIs in terms of what functionality they expose for central use cases, for example as covered in doctests strings and HOWTOs (http://nltk.org/howto/). In order to review the distribution entropy code, we would need to see some illustrated use cases of how the library is called and what information it provides to the typical user. It would also be helpful if we could move this discussion to the nltk-dev list.

Issue #86, opened 2012-01-17
Reimplement FreqDist


alexrudnick:
Reimplement FreqDist to use collections.Counter, in NLTK 3.0.

http://docs.python.org/3.1/library/collections.html?highlight=counter#collections.Counter

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=456</p><hr/><h2>earlier comments</h2>
<p><strong>dchichkov said, at 2010-07-01T22:56:11.000Z:</strong></p>

A few comments:

1) Currently FreqDist is derived from dict(). While it is maybe convenient, it creates extra ambiguities in the interface.

2) Recalculating _N upon every inc() call is wasteful, especially in a way it is done now - via __setitem__ call.

3) For large data sets it is valuable to have a trimming method that deletes samples with frequencies/counts lower than a specified limit;

4) For large data sets it is valuable to have a trimming method that can be called during the data collection process and trims the FreqDist to a specified size (e.g. 100,000,000 samples). This trimming method should delete 'oldest' samples (FIFO basis) with the lowest frequencies. That can be implemented if an OrderedDict is used to store the samples (as an option during the construction).

5) Consider storing samples that occur only once (hapax legomena) in a separate bin (set). Potentially that could reduce memory usage and give overall performance boost. Something along these lines (the code is untested):

```
        self._d = dict()
        self._s = set()
        self._N = 0
        self._Nr_cache = None
        self._max_cache = None
        self._item_cache = None
        if samples:
            self.update(samples)

    def inc(self, sample, count=1):
        """
        Increment this C{FreqDist}'s count for the given
        sample.

        @param sample: The sample whose count should be incremented.
        @type sample: any
        @param count: The amount to increment the sample's count by.
        @type count: C{int}
        @rtype: None
        @raise NotImplementedError: If C{sample} is not a
               supported sample type.
        """
        if count == 0: return
        self._N += count
        if sample in self._d: self._d[sample] += count; return
        if sample in self._s: self._d[sample] = count + 1; self._s.remove(sample); return
        if count == 1: self._s.add(sample); return
        self._d[sample] = count

    def __getitem__(self, sample):
        if sample in self._d: return self._d[sample]
        if sample in self._s: return 1
        return 0

    def __setitem__(self, sample, value):
        """
        Set this C{FreqDist}'s count for the given sample.

        @param sample: The sample whose count should be incremented.
        @type sample: any hashable object
        @param count: The new value for the sample's count
        @type count: C{int}
[...]

PR #544, opened 2013-11-07
Remove caching from FreqDist

stevenbird:
This commit removes caching from FreqDist. This needs more testing, and classes that use FreqDist need to be updated. Cf https://github.com/nltk/nltk/issues/86

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -48,6 +48,7 @@ from itertools import islice
 from collections import defaultdict
 from functools import reduce
 from nltk import compat
+from nltk.compat import Counter
 
 from nltk.internals import raise_unorderable_types
 
@@ -57,11 +58,8 @@ _NINF = float('-1e300')
 ##  Frequency Distributions
 ##//////////////////////////////////////////////////////
 
-# [SB] inherit from defaultdict?
-# [SB] for NLTK 3.0, inherit from collections.Counter?
-
 @compat.python_2_unicode_compatible
-class FreqDist(dict):
+class FreqDist(Counter):
     """
     A frequency distribution for the outcomes of an experiment.  A
     frequency distribution records the number of times each outcome of
@@ -108,10 +106,8 @@ class FreqDist(dict):
             distribution with.
         :type samples: Sequence
         """
-        dict.__init__(self)
         self._reset_caches()
-        if samples:
-            self.update(samples)
+        Counter.__init__(self, samples)
 
     def inc(self, sample, count=1):
         """
@@ -125,8 +121,9 @@ class FreqDist(dict):
         :raise NotImplementedError: If ``sample`` is not a
                supported sample type.
         """
-        if count == 0: return
-        self[sample] = self.get(sample,0) + count
+        if count == 0:
+            return
+        self[sample] += count
 
     def __setitem__(self, sample, value):
         """
@@ -140,8 +137,8 @@ class FreqDist(dict):
         :raise TypeError: If ``sample`` is not a supported sample type.
         """
 
-        self._N += (value - self.get(sample, 0))
-        dict.__setitem__(self, sample, value)
+        self._N += value - self[sample]
+        Counter.__setitem__(self, sample, value)
 
         # Invalidate the caches
         self._reset_caches()
@@ -168,16 +165,6 @@ class FreqDist(dict):
         """
         return len(self)
 
-    def samples(self):
-        """
-        Return a list (or an iterator under Python 3.x) of all samples
-        that have been recorded as outcomes by this frequency distribution.
-        Use ``fd[sample]`` to determine the count for each sample.
-
-        :rtype: list
-        """
-        return self.keys()
-
     def hapaxes(self):
         """
         Return a list of all samples that occur once (hapax legomena)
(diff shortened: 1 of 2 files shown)


stevenxxiu:
I see that the `inc` method is now removed, should the classes be updated too regarding this?

stevenxxiu:
I've modified `SimpleGoodTuringProbDist` in the above commit so that it passes the unit tests in probability.doctest, I don't think it needs caching, as `_r_nr` is only used once on `__init__`, I've also changed some method names in `FreqDist` so they conform to PEP-8.

nltk/probability.py
```diff
@@ -264,9 +208,9 @@ def freq(self, sample):
         :type sample: any
         :rtype: float
         """
-        if self._N == 0:
+        if self.N() == 0:
```
kmike, line 211: Maybe calculate N once and store it in a local variable to match old behavior?

nltk/probability.py
```diff
@@ -490,19 +330,17 @@ def __repr__(self):
         """
         return '<FreqDist with %d samples and %d outcomes>' % (len(self), self.N())
 
-    def __str__(self):
+    def __str__(self, maxlen = 10):
```
kmike, line 333: I think it is better to have a separate method with 'maxlen' parameter and call it from `__str__`

nltk/probability.py
```diff
@@ -199,36 +160,19 @@ def Nr(self, r, bins=None):
...
-        """
-        if self._Nr_cache is None:
-            self._cache_Nr_values()
-        return sorted(self._Nr_cache.items())
-
-    def _cache_Nr_values(self):
-        Nr = defaultdict(int)
         for sample in self:
             c = self.get(sample, 0)
             if c == 0 and c not in Nr:
```
heatherleaf, line 171: `_Nr` not `Nr`

nltk/probability.py
```diff
@@ -186,12 +149,10 @@ def hapaxes(self):
         """
         return [item for item in self if self[item] == 1]
 
-    def Nr(self, r, bins=None):
+    def Nr(self, bins=None):
```
heatherleaf, line 152: Here are two equivalent definitions of `Nr(...)`:

```
_Nr = defaultdict(int)
for count in self.values():
    _Nr[count] += 1
# Special case for Nr[0]:
_Nr[0] = bins - self.B() if bins is not None else 0
```

Or this one, even shorter:

```
_Nr = Counter(self.values())
# Special case for Nr[0]:
_Nr[0] = bins - self.B() if bins is not None else 0
```

The latter is of course much nicer, but it's also 4x slower than the defaultdict version (using timeit on python 3.3).

nltk/probability.py
stevenbird, line 333: +1

nltk/probability.py
stevenbird, line 171: Thanks

stevenbird:
I've incorporated all these contributions (except caching of N()) into a new `fdist` branch. Further testing welcome.

Note that there's still a problem:
$ python ./probability.py
  ...
  File "./probability.py", line 897, in _calculate_Tr
    Tr[r] += self._heldout_fdist[sample]
IndexError: list index out of range

heatherleaf:
I got rid of that error by letting `Tr` be a defaultdict(float) instead of a list:

```
def _calculate_Tr(self):
    ....
    Tr = defaultdict(float)
    for sample in self._heldout_fdist:
        ...
```

 But then I got this error instead:

```
  File "/Users/peter/Projekt/NLTK/nltk/nltk/probability.py", line 879, in <listcomp>
    Nr = [base_fdist.r_Nr(r, bins) for r in range(self._max_r+1)]
TypeError: r_Nr() takes from 1 to 2 positional arguments but 3 were given
```

So I replace the list comprehension on line 879 with this:

```
    r_Nr = base_fdist.r_Nr(bins)
    Nr = [r_Nr[r] for r in range(self._max_r+1)]
```

 And now I get this error:

```
  File "/Users/peter/Projekt/NLTK/nltk/nltk/probability.py", line 951, in prob
    return self._estimate[r]
IndexError: list index out of range
```

I don't know that much about the HeldoutProbDist, and I don't have the time to continue right now.

stevenxxiu:
Think I got rid of the bug now, was caused by a bug in the `max()` implementation.

stevenbird:
Oh, well done @jskda

stevenxxiu:
Thanks! Forgot to remove a debug statement, fixed that now.

stevenbird:
Thanks. I've manually added the bugfixes to probability.py in the fdist branch at https://github.com/nltk/nltk/tree/fdist. Please take a look. There are still problems (to see them, run "python ./probability.py").

heatherleaf:
Isn't this a really slow way of calculating (line 879 in `HeldoutProbDist.__init__`)? Since `.Nr()` calls `.r_Nr()` for every r:

```
    Nr = [base_fdist.Nr(r, bins) for r in range(self._max_r+1)]
```

How about this instead:

```
    r_Nr = base_fdist.r_Nr(bins)
    Nr = [r_Nr[r] for r in range(self._max_r+1)]
```

heatherleaf:
The problems seem to be very old. I tried to checkout old NLTK versions, and all of them ran into the same problems:
- first a lot of `UserWarning: Probability distribution <SimpleGoodTuringProbDist based on 10 samples> sums to 0.9128893773581336; generate() is returning an arbitrary sample.`
- then there is an `AssertionError: Bins parameter must not be less than freqdist.B() + 1`

heatherleaf:
I first get the AssertionError after commit e5f2a9e, 3 years ago... Has noone tested the probability demo since then?

stevenxxiu:
I think the cause of the assertion error is that in the commit @heatherleaf mentioned, `SimpleGoodTuringProbDist` started using the total number of bins as the third parameter of `__init__` instead of using the "Katz cutoff threshold". I don't think the Katz cutoff exists in the nltk codebase anymore (I did a grep of "katz"), can this portion of the demo be removed?

I've added my suggested fix above.

stevenxxiu:
I've added `SUM_TO_ONE=False` to `SimpleGoodTuringProbDist` to fix the not summing to one issue, in the commit here on my branch: https://github.com/jskda/nltk/commit/5e541f6de892f5c1a97d34eb1456b9dfa64fa823, as I don't think sample probabilities in the simple good turing distribution sum to one.

stevenbird:
@jskda please submit a pull request against the fdist branch in the nltk repository.

PR #548, opened 2013-11-11
k-means bug len() of map object

ibarria0:
In python3, map evaluates to a map object instead of a list and you cant do len() on them.

This causes line 138 in kmeans.py to fail since self._means is a map object

``` python
for index in range(len(self._means)):
```

I changed the definition of self._means to a list comprehension in line 124

``` python
# recalculate cluster means by computing the centroid of each cluster
new_means = [self._centroid(cluster,mean) for cluster,mean in zip(clusters, self._means)]
...
self._means = new_means
```

### DIFF
--- a/nltk/cluster/kmeans.py
+++ b/nltk/cluster/kmeans.py
@@ -121,8 +121,8 @@ class KMeansClusterer(VectorSpaceClusterer):
                     #print '  mean', i, 'allocated', len(clusters[i]), 'vectors'
 
                 # recalculate cluster means by computing the centroid of each cluster
-                new_means = [self._centroid(cluster,mean) for cluster,mean in zip(clusters, self._means)]
-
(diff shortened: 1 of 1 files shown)


ibarria0:
created a pull request which references this

kmike:
@ibarria0, I think `list(map(...))` vs `[...]` is not very important (there is no consistent style for that in nltk). `list(map(...))` is shorter, but list comp is also fine as long as it follows pep8 (currently the line is too long and there is a missing space after comma). These issues are very minor, and you've fixed a real bug, but is better to fix the formatting.

ibarria0:
Yes, I will switch it to list(map(....)) immediately.

kmike:
Thanks!

heatherleaf:
@kmike, it seems you merged into the very old and obsolete(?) 2and3 branch.

kmike:
oops

PR #549, opened 2013-11-17
update to nltk.tag.brill 

muneson:
This is a proposed update to the brill tagger for nltk3, with 
main differences with respect to the one of nltk2 as indicated below.
It is (or at least was written to be) compatible with python 2.6, 2.7, 
3.2, 3.3. The provided demo will produce identical results under these
versions (well, given an identical baseline, of course -- at the time of
writing, tag.UnigramTagger() will produce subtly different
taggers with different python versions. My tests have 
just picked a baseline tagger, serialized that and read it 
from disk for all other versions). 

I realize this is a bit big for a pull request -- it is offered to nltk 
much as an afterthought, or it would have been come in more
digestible pieces. Still, I tried to redo my successive refactorization
(the largest change), into the first few commits with no new features, 
so that later commits can be mostly incremental and focus on single 
features. The demo should work after each commit (except when it 
doesn't ;-) ) Anyway, it is a point of departure.

**Modularization and isolation of application-specific code**
PR is split into subpackages (its predecessor is thoroughly monolithic). 
This should make it easier to implement new alternative training 
algorithms and other extensions, of which there are many. It is also 
less biased towards POS tagging, which is just a single thing
out of many that TBL is good for, and there should be natural homes
for future non-postagging applications.

Here is the present layout of the update:

```
brill/
  __init__.py       #desired exports

  nltk2/            #(for backwards compatibility)
    __init__.py
    demo.py         #demo
    postagging      #class ProximateTagsRule, ProximateWordsRule
    template        #class ProximateTokensTemplate,
                    # SymmetricProximateTokensTemplate,
                    # ProximateTokensRule

  application/
    __init__.py
    postagging      #class Word, Pos
    ...             #future implementations of other applications

  trainer/
    __init__.py     #desired exports and defaults for training algorithms
    fast            #class TaggerTrainer
    brillorig       #class TaggerTrainer
    ...             #future implementations of alternative trainers
                    #(e.g  Florian & Ngai, Samuel, Hepple, dos Santos ...)
                    #and other extensions (multidimensional TBL,
                    #probabilistic TBL, constraint-grammar-styled TBL,
                    #transformation-based regression, etc)
  demo/
    __init__.py
    postagging      #postagging demo
    postagging_templates #predefined template sets (e.g., from publications)
    ...             #future demos of other applications

  tagger            #class BrillTagger

  rule              #class BrillRule (abstract), Rule

  template          #class BrillTemplateI, Template, Feature (abstract)

  erroranalysis     #error_list
  ...               #future additions -- export rules to FST, graphing, etc
```

**Feature specification**
Any number of features is allowed at any number of positions (not 
necessarily contiguous). They are expressed so :

```
templates = [                  # or start, end args (as in nltk2):
    Template(Word([-3,-2,-1]), # Template(Word(-3, -1))
    Template(Word([1,2,3]),    # Template(Word(1, 3))
    Template(Tag([-1]), Tag([1])),
    Template(Tag([-1]), Word([1,2]))
]
```

**Feature- and Template-generating factory functions**
There is a class method Feature.expand(startat, winlen) which generates lists
of Features. There is also a companion method Template.expand([Featurelists]) which basically builds
the Cartesian product of such Feature lists (or manually specified ones),
with semantic duplicates removed. For instance, 

```
wordtpls = Word.expand([-1,0,1], [1,2], excludezero=False)
print(wordtpls)
```

[Word([-1]), Word([0]), Word([1]), Word([-1, 0]), Word([0, 1])]

```
tagtpls = Pos.expand([-2,-1,0,1], [1,2], excludezero=True)
print(tagtpls)
```

[Pos([-2]), Pos([-1]), Pos([1]), Pos([-2, -1])]

Now, 

```
templates = list(Template.expand([wordtpls, tagtpls, tagtpls], combinations=(1,3)))
print(tpls)
```

yields

```
[Template(Word([-1])), Template(Word([0])), Template(Word([1])), Template(Word([-1, 0])), 
Template(Word([0, 1])), Template(Pos([-2])), Template(Pos([-1])), Template(Pos([1])), 
Template(Pos([-2, -1])), Template(Pos([-2]),Word([-1])), Template(Pos([-1]),Word([-1])), 
...
Template(Pos([-2]),Pos([-1]),Word([-1, 0])), Template(Pos([-2]),Pos([1]),Word([-1, 0])),
Template(Pos([-1]),Pos([1]),Word([-1, 0])), Template(Pos([-2, -1]),Pos([1]),Word([-1, 0])),
Template(Pos([-2]),Pos([-1]),Word([0, 1])), Template(Pos([-2]),Pos([1]),Word([0, 1])),
Template(Pos([-1]),Pos([1]),Word([0, 1])), Template(Pos([-2, -1]),Pos([1]),Word([0, 1]))]
```

that is, the 53 semantically unique templates with 1..3 features each (all non-empty 
combinations of Word(wordpositions), Tag(...), Tag(...), where (for instance) 
wordpositions contains all the ways of starting of one of [-1, 0, 1] and looking 
at a window of length one of [1,2]). The excludezero flag omits all positions lists
containing 0 (the motivation is that the target Feature may get special treatment at position 
0, at least in many common implementations).
See docstring for more.

**Feature computation**
The abstract extract_token() method takes the entire current token sequence
and the current index as input, instead of a single token; this makes it possible 
to compute aggregate features (sequence length, say). 

**Accuracy threshold**
Rules with lower accuracy than a threshold will be discarded. This may hurt 
performance a bit, but will often produce rules which are more interesting read 
to a human (which in some cases may be the main interest).  For instance,
with min_acc = 0.96 rules such as the following may appear:

```
 200 202   2   2  | NN->VB if Pos:TO@[-1] & Pos:-NONE-@[-2]
 170 170   0   1  | NN->NNP if Word:Mr.@[-2,-1]
  82  82   0   0  | NN->VB if Pos:MD@[-1]
  54  56   2   3  | VBD->VBN if Pos:VBD@[-1]
```

**String representations of rules**
A rule now knows what template it was created from, which is useful
for statistical analysis. repr(rule) looks so:

```
 156 194  38  24  | Rule(029, 'NN', 'NNP', [(Pos([-2, -1]),'NNP')])
  92 100   8  19  | Rule(018, 'NN', 'VB', [(Word([-1]),'to')])
  51  53   2   1  | Rule(020, 'VBD', 'VBN', [(Pos([1]),'-NONE-'),
                  |   (Pos([2]),'IN')])
```

str(rule) returns a compact, logic-based representation:

```
 156 194  38  24  | NN->NNP if Pos:NNP@[-2,-1]
  92 100   8  19  | NN->VB if Word:to@[-1]
  51  53   2   1  | VBD->VBN if Pos:-NONE-@[1] & Pos:IN@[2]
```

The wordier str(rule) of nltk2 is also available

```
 156 194  38  24  | NN -> NNP if the Pos of words i-2...i-1 is "NNP"
  92 100   8  19  | NN -> VB if the Word of the preceding word is "to"
  51  53   2   1  | VBD -> VBN if the Pos of the following word is "-NONE-",
                  |   and the Pos of word i+2 is "IN"
```

**Template statistics**
There is a diagnostic tagger.print_template_statistics() method which ranks all
templates according to the total relative scores (in training or, optionally, testing) 
they have been involved in (useful to speed up training and/or suggest new 
templates). For instance, 

```
TEMPLATE STATISTICS (TEST AND TRAIN) (25 templates, 183 rules)
TEST  (   5251 tokens) initial  1100 0.7905 final:   763 0.8547 
TRAIN (  13207 tokens) initial  2653 0.7991 final:  1232 0.9067 
#ID | Score (test) | Score (train) |  #Rules     | Template
-----------------------------------------------------------
082 |   95   0.282 |   230   0.162 |   6   0.033 | Template(Pos([-2, -1]))
071 |   57   0.169 |   149   0.105 |   9   0.049 | Template(Word([-1]))
075 |   38   0.113 |    70   0.049 |   6   0.033 | Template(Pos([1]))
076 |   34   0.101 |   164   0.115 |  15   0.082 | Template(Pos([-1]))
073 |   28   0.083 |    96   0.068 |   7   0.038 | Template(Pos([1]),Pos([2]))
067 |   27   0.080 |   241   0.170 |  51   0.279 | Template(Word([0]))
060 |   21   0.062 |    78   0.055 |  16   0.087 | Template(Word([-2, -1]))
081 |   18   0.053 |    69   0.049 |   5   0.027 | Template(Pos([-3, -2, -1]))
065 |   11   0.033 |    17   0.012 |   2   0.011 | Template(Word([0]),Pos([1]))
074 |    3   0.009 |    26   0.018 |   7   0.038 | Template(Pos([-1]),Pos([-2]))
...
```

**Incremental tagging**
Statististics for individual rules is collected in training and can (optionally, slow) 
also be collected in testing by batch_tag_incremental() method. This gives 
statistics for individual rules which may be used to plot learning curves 
such as (green for incremental accuracy on training set, blue on testing)

learningcurve

**A demo with a documented function for each of the above**
in the form of a postag() function with many default arguments,
and different named wrappers around it.


nltk/tag/brill/nltk2/template.py
```diff
@@ -0,0 +1,346 @@
...
+        Each implentation of this method should correspond to an
+        implementation of the method with the same name in a subclass
+        of ``ProximateTokensTemplate``.
+
+        :param token: The token
+        :type token: tuple(str, str)
+        :return: The property
+        :rtype: any
+        """
+        assert False, "ProximateTokenRules must define extract_property()"
```
kmike: It is better to raise NotImplementedError here because assert will become no-op if python is executed with -O option.

muneson: (I must have messed up something with my github account -- now I get
"403 Forbidden -- Request forbidden by administrative rules"
seemingly everywhere. In the meantime, hope "reply directly" is a way
of answering that works)

Thanks, I'll commit a change of the assert.

However, note that this is not new code -- the code under
nltk.tag.brill.nltk2
is basically the previous version of nltk.tag.brill (sorry, should
have pointed that out more clearly in the pull request). Except the basic
splitting into
subpackages for clarity (and some automatized global changes to
header and docstring format),  I haven't changed anything.

All previous functionality (and more) is available in the proposal, but
the APIs aren't completely compatible. Still, the changes at the
callers' side are small, and maybe we should have a more general
discussion whether it is worth the maintenance effort to keep
the old nltk.tag.brill.nltk2 in the first place, before anyone puts
more effort into reviewing it. Thoughts?

Marcus

2013/11/19 Mikhail Korobov notifications@github.com

kmike: I see there were many asserts like this before your refactoring; I think it is a good chance to fix them. For example, asserts in `__eq__` and alike methods should be replaced with TypeErrors.

kmike:
Hi @muneson,

First: I'm not an expert in Brill tagging algorithm. 

What you're proposing looks like good changes, and the code looks good. But it is not easy to properly review this PR because original code is split into subpackages, so diffs no longer help in checking what is changed and what is not changed. It could be easier to review this PR without splitting the module into subpackages. I think that's what makes reviewing harder, not the size of PR.

I don't know if it belongs to this PR, and please tell me I'm wrong if I'm wrong, but it looks like a good idea to split the rule into condition and and action (basically, split "apply" and "applies" methods into different objects). Replacing the whole tag makes sense only if tagset is small (like in English); as far as I know, for morphologically rich languages like Russian or Polish which have large tagsets (thousands of tags) an usual way to apply Brill tagger is to replace some property of tag with some other property when condition is met, e.g. change only number or case when condition is satisfied. It looks like to do that now most of classes must be overridden.

muneson:
Hi,

thanks for your comments. I understand this is not an easy PR to review and
I welcome input to make it more digestible.

> because original code is split into subpackages, so diffs
> no longer help in checking

It was near-to-impossible to start work on a 1300-line module which mixed
data representation, training, output formats, application code etc in a
single place, so the refactoring was unavoidable. However,
my idea was to move all the subpackaging to the first few commits -- those
should have zero change in behaviour, just move around code. Thus, to assess
the actual changes in code, you could assume that commit #5 of the PR

https://github.com/muneson/nltk/commit/0830ebfb3e43175bb96e7e047f342d095f266e5f

works like the original; then you could take that as point of departure and
look at
the sequence of diffs starting at that point.

Probably the entire PR would have been better split into PR1:
refactorization with no change
in behaviour and PR2: addition of the rest. This was my first PR ever and I
didn't
understand how to do that, but if it is still meaningful to do so and
anyone can point
me to a good tutorial I can try to split it accordingly.

> I don't know if it belongs to this PR, and please tell me I'm wrong if I'm
> wrong,
> but it looks like a good idea to split
> the rule into condition and and action (basically, split "apply" and
> "applies" methods into different objects).
> Replacing the whole tag makes sense only if tagset is small (like in
> English);
> as far as I know, for morphologically >rich languages like Russian or
> Polish which have large tagsets (thousands of tags) an usual way to apply
> Brill
> tagger is to replace some property of tag with some other property when
> condition is met, e.g. change only number >or case when condition is
> satisfied. It looks like that to do that now most of classes must be
> overridden.

No, you are not wrong, there are several papers suggesting that mechanism
(see e.g. Beata Megyesi 1999, "Improving Brill's pos tagger for an
agglutinative language"
and Szymon Acedański 2010, "A morphosyntactic Brill tagger for inflectional
languages").
There are also several attractive extensions to the basic algorithm --
first in line, I think, is
multidimensional TBL, which allows the template set to target more than one
target feature (say, learning POS and NP chunking simultaneously).
(I have written a survey (ACM Computing surveys, forthcoming), on TBL;
I refer to that for an overview of TBL extensions.)

However, not all can be done at the same time, or there would never be any
commits.
The  main design choice here is to offer an API (Template, Feature) which
can later
be extended upon. My idea on how to make the API flexible is to allow for
Template
(and possibly Feature, too) to later take named arguments.
For multidimensional TBL, for instance:

Template(Chunk([-1]), Word([-1]), target=Chunk, srctagincondition=True).
Template(Chunk([-1]), Pos([-1]), target=Pos, srctagincondition=False)
Template(Chunk([-1]), Pos([-1]), target=Pos, srctagincondition=False)

The first would then work like before (conceptually, in the example of
Pos-tagging,
the src-tag Pos[0] is implicitly added to the conditions list); but the
second would
not add anything implicitly and thus allow for conditioning only on
Features which represent case, number or something else of Pos([0]).
It might even set a new Pos([0]) without conditioning on the current one at
all.
It may be that a new class, say MultiDimTemplate, should be added with the
appropriate defaults. However, I don't think the user would have to add any
overrides except subclassing Feature.

(Some of these later suggestions would require or at least be much
facilitated by a more
flexible token representation, probably based on NamedTuples. So rather than
the 2-tuple (word, tag), we might have
token.word, token.tag, token.chunk, token.tagtrue, token.chunktrue...
But this is another topic.)

Anyway, the proposal is what I have right now. It is mostly a matter of
submitting
existing code -- I can't say when I will have time to continue on it. I
figured it
is at least a reasonably flexible improvement to the current
implementation.

Hope this made sense, let me know otherwise.

cheers,

Marcus

2013/11/19 Mikhail Korobov notifications@github.com

stevenbird:
Thanks @muneson for a great contribution. I intend to review it properly in the next few days.

heatherleaf:
@stevenbird This isn't merged, afaik

stevenbird:
Sorry -- I'm not sure how this was closed!

stevenbird:
This pull request was automatically closed when I removed the branch on which it was based (master branch), as part of the process of migrating to the gitflow model. I have now pulled it into a new `brill` branch for review.

stevenbird:
Now that I've deleted the master branch again, this request has been automatically closed, and it cannot be reopened. I'm opening a new issue so we can continue the discussion there. https://github.com/nltk/nltk/issues/555

PR #554, opened 2013-11-25
Fixes #550

### LINKED ISSUE #550: Featstruct.py: Use of variable before check
In _parse_seq_value (line 2324) m.group(1) is attempted before checking if m is not None. This results in a crash with no information rather than a ValueError. Move this line after the check.

heatherleaf:
https://github.com/nltk/nltk/issues/550


stevenbird:
Thanks.

Issue #540, opened 2013-11-06
Eliminate yaml dependency


stevenbird:
Use json or pickle instead of yaml for serialisation, to remove the yaml dependency in favour of the built-in libraries. This effects a couple of tagger modules, and data.py and yamltags.py.

stevenbird:
We need to decide which of pickle or json to use.

kmike:
Big +1 for removing YAML dependency.

### JSON

Pros:
- there is no security risk in using json;
- compatibility is not an issue;

Cons:
- it'll require more code than pickle.

### Pickle

Pros:
- pickle support means multiprocessing support;
- less code to write.

Cons:
- there are different pickle protocols;
- pickling/unpickling is different in Python 2/3;
- pickled models disallow changing/moving classes, attributes, etc. - they require the same code to present at unpickling time;
- unpickling allows arbitrary code execution;
- pickled data is not readable without Python.

I vote for JSON even though it is more work.

heatherleaf:
Also +1 for JSON

turicas:
The idea is to remove all YAML support (like not being able to use `nltk.data.load` with a YAML file) or just remove from the official data distribution?

Curently the list of files that have `yaml` in its contents is:
- `ChangeLog`
- `nltk/data.py`
- `nltk/__init__.py`
- `nltk/tag/brill.py`
- `nltk/tag/sequential.py`
- `nltk/test/data.doctest`
- `nltk/yamltags.py`
- `pip-req.txt`
- `setup.py`
- `web/install.rst`

Can you provide more information about which files need to be changed and which not?

stevenbird:
It would be fine to leave yaml support in data.py, and data.py works fine even when yaml is not installed. However, I think yaml can be replaced everywhere else.

stevenxxiu:
Should the classes using yaml have any backwards compatibility for yaml files? I don't think any yaml files exist in the nltk data packages.

stevenbird:
No, we will give up backwards compatibility. We'll just retain yaml support in data.py in case someone wants to use it to load their own yaml files using the data loader.

kmike:
If yaml files don't exist in nltk data packages then we can live without json serialization. Pickle support is still desired because users may want to save models using pickle, and because it is needed for multiprocessing to work. Most of pickle disadvantages doesn't matter if it is user who creates a model and saves it.

heatherleaf:
[...]

PR #557, opened 2013-12-03
Pull request.json tags, cf https://github.com/nltk/nltk/issues/540

stevenxxiu:
Removed any yaml serialization schemes, converted `nltk.tag.sequential` to using a json serialization scheme, by using `nltk.jsontags` as a helper class.


kmike:
I like this solution - changes are minimal. 

Could you please add some tests and update install docs (as PyYAML is no longer needed)?

stevenbird:
Yes, good work @jskda.
+1 to @kmike's request.

stevenxxiu:
Thanks. I've added some tests for `nltk.jsontags`.

Issue #522, opened 2013-10-28
Interface to Stanford Parser


stevenbird:
Add interface to Stanford Parser.

stevenbird:
We will need to decide which trained model(s) to distribute with NLTK. We will also need to make sure that the model version is reflected in its filename. A given version of NLTK will only be expected to support one version of the Stanford Parser (i.e. we will hardcode the data loader to the model version).

stevenbird:
Note that we want to pick up the latest version of the parser that is available. This hardcodes 3.3.0, but it is now already at 3.3.1.

PR #560, opened 2013-12-11
Interface to stanford parser, cf https://github.com/nltk/nltk/issues/522

stevenxxiu:
Interface to stanford parser which works by calling java.exe on the commandline, cf https://github.com/nltk/nltk/issues/522


stevenbird:
@jskda do we need to add a trained model to the nltk_data collection?

stevenxxiu:
@stevenbird I've currently implemented this by adding the model data path as a java classpath: `nltk.parse.stanford.StanfordParser.model_path`, by using the model jar from the stanford parser.
Should I extract those models to nltk_data?

Issue #559, opened 2013-12-06
Gale-Church aligner


stevenbird:
Get the Gale-Church aligner in the align branch to work. It currently breaks when blocks have unequal length (i.e. aligning a source paragraph with n sentences with a target paragraph with m != n sentences).

stevenbird:
@tmarek: we are moving ahead with your implementation at last, and would welcome your participation if you have time

stevenbird:
Merged into develop in https://github.com/nltk/nltk/commit/e4df597462981b6cee927c0286b934ce18ddd57b

PR #562, opened 2013-12-12
Fixes to the align branch

stevenxxiu:
Some small bug fixes for the align branch and changed variable names to PEP-8 style variable names.

Fixed the gale church implementation so the included doctests pass & used log probabilities in the distance function. 
The implementation details has changed somewhat from @tmarek's original implementation as I couldn't get the same distance scores as the ones from the reference implementation in the the Gale-Church Aligner paper.

cf https://github.com/nltk/nltk/issues/559


stevenbird:
Nice work, thanks @jskda.

PR #563, opened 2013-12-12
Speed improvements and refactor of wordnet `common_hypernyms` functions

watercrossing:
The changes cause nltk to cache the sets of all hypernyms of a Synset, as its generation is expensive. This causes speed improvements of factor >20 if `common_hypernyms` or `lowest_common_hypernyms` are called repeatedly on combinations of Synsets, for example:

```
from itertools import combinations_with_replacement
p = wn.synsets("House")
[x.common_hypernyms(y) for x,y in combinations_with_replacement(p,2)]
```

Comparing speeds using `timeit`, the following results are achieved:
Current develop: 100 loops, best of 3: 4.82 ms per loop
This pull request: 10000 loops, best of 3: 86.6 μs per loop
74x faster.


stevenbird:
Thanks for the speed improvement @watercrossing

Issue #555, opened 2013-11-28
Brill tagger


stevenbird:
@muneson has contributed a new implementation of the Brill tagger in https://github.com/nltk/nltk/pull/549. I've merged it into a new feature branch `brill` and invite people to test it out and post comments here.

muneson:
_Suggestion: skip nltk2 compatibility_

At present, in order not to break any tests, you can say

```
from nltk.tag.brill import (
    BrillTagger,
    ProximateTagsRule,
    ProximateWordsRule,
    ProximateTokensRule,
    ProximateTokensTemplate,
    SymmetricProximateTokensTemplate
)
```

and get the exact nltk2 API. This unnecessarily pollutes the 
nltk.tag.brill namespace, however, so a moderate but
efficient improvement is to require the user to say

```
from nltk.tag.brill.nltk2 import (
    BrillTagger,
    ProximateTagsRule,
    ProximateWordsRule,
    ProximateTokensRule,
    ProximateTokensTemplate,
    SymmetricProximateTokensTemplate
)
```

Still, since there really is nothing which can be done with the older API only,
I would prefer to drop the nltk2 concerns entirely,
so that one need not worry about keeping legacy signatures synchronized
(say, with dummy arguments) when the signatures of their newer
counterparts are changed (e.g., by adding new keyword arguments).
If not when changing major version number, then when?

muneson:
_Warning: possibly too automated yaml serialization (but yaml is to disappear anyway)_

The earlier nltk2 version required user-defined (application) subclasses (of Rule, with that API) to specify 
a yaml tag. By contrast, with the suggested API, user-defined subclasses (of Feature) automatically constructs a 
yaml tag from the class name, if not specifically given by the user. 

There may be a reason for the original design which I have missed. Anyway, it seems yaml is scheduled for removal (https://github.com/nltk/nltk/issues/540) in favour of pickle and optional json, so I haven't put much effort into it.

muneson:
_A note on the place of 'brill tagging' in the namespace hierarchy_

In my view, both of the terms "brill" and "tagging" get increasingly inaccurate when extensions to the basic transformation-based learning (TBL) paradigm are considered. Brill invented the basic paradigm and certainly deserves all credit for that. However, TBL has been much developed since. In an upcoming survey on TBL, I review two dozen papers on later extensions to basic TBL, and only one of them is written by Brill. 

[...]

PR #564, opened 2013-12-14
Transformation-based learning

muneson:
This is basically the same pull request as https://github.com/nltk/nltk/pull/549, which mysteriously got automagically closed. It has slight emendations after some discussion there and in https://github.com/nltk/nltk/pull/555. (Not sure that a new pull request is the best way to do deal with automagic closings -- but how, otherwise?)

The feature description is really unchanged from https://github.com/nltk/nltk/pull/549. Here are comments to the further changes:
- the package now is called `nltk.tag.tbl`, since everyone seemed happy with the tbl-part of the change. @stevenbird proposed moving it to `nltk.classify.tbl` (as the original suggestion) while still making it accessible from `nltk.tag.brill`. To me, this seems like a good idea. At the moment, only the renaming to tbl is carried out, however, so nothing moved to `nltk.classify` before hearing other views. 

The current top-level packages seems to be a mixture of named problems (`tokenize`) and general methods to solve them (`classify.*`) -- it seems difficult to do that consistently. 

Below I will still call the package `nltk.tag.brill` to avoid confusion.
- the old code in the `nltk2` subpackage is removed and all references to it. External tests may still choke
  on it, of course. 
- `application` is renamed `task`, to better reflect the semantics; will hold subclasses of Feature and
  sets of Templates specific to a given problem (and problem encoding). @Heatherleaf suggested "feature"
  which also seems OKish (not clear that it is an obvious home for predefined templates).
- the demo package is gone and moved up into module `nltk.tag.brill.demo`. More demos are unlikely to be needed.
- it is probably reasonable to call the code "based on previous version"
  -- a very rough estimation is that 70% of the code is new. I added
  the following header to all files as per suggestion.
  
  ```
  Natural Language Toolkit: Transformation-based learning
  
  Copyright (C) 2001-2013 NLTK Project
  Author: Marcus Uneson <marcus.uneson@gmail.com>
    based on previous (nltk2) version by
    Christopher Maloof, Edward Loper, Steven Bird
  URL: <http://nltk.org/>
  For license information, see  LICENSE.TXT
  ```
- trainer and task subpackages are kept for now. I don't have more time right now, but I will return with stronger arguments a bit later. I disagree with the view that deeper module hierarchies by necessity are less transparent or maintainable than shallower ones. In my view, nltk needs more subpackages, not fewer. Anyway, irrespective of how that is solved, it won't affect the API. The user knows nothing about subpackages and may say for instance
  
  ```
  from nltk.tag.tbl import Template, Feature, TaggerTrainer, error_list
  class MyFeat1(Feature):  
      ...
  class MyFeat2(Feature): 
      ...
  templates = Template.expand([MyFeat1, MyFeat2])
  ...
  my_tagger = TaggerTrainer(..., templates)
  ```

For the record, here is sample of known TBL problem encodings (Features and Templates), all of which would be rather self-explaining if found in `nltk.tag.brill.tasks`,  but much less so in `nltk.tag.brill`:

```
    Part-of-speech tagging             Brill (1993c , 1995a)
    Unknown word guessing              Brill (1994); Mikheev (1997)
    Text chunking                      Ramshaw and Marcus (1995)
    Prepositional phrase attachment    Brill and  Resnik (1994)
    Parsing/grammar induction          Brill (1996)
    Morphological disambiguation       Oflazer and Tür (1996)
    Spelling correction                Mangu and Brill (1997)
    Word segmentation                  Palmer (1997)
    Message understanding              Day et al. (1997)
    Dialogue act tagging               Samuel et al. (1998)
    Prosody prediction                 Fordyce (1998)
    Ellipsis resolution                Hardt (1998)
    Word sense disambiguation          Dini et al. (1998)
    Document format processing         Curran and Wong (1999)
    Grapheme-phoneme conversion        Bouma (2000)
    Grammar correction                 Hardt (2001)
    Handwritten character segmentation Kavallieratou et al. (2000)
    Regression                         Bringmann et al. (2002)
    Hyphenation                        Bouma (2003)
    Named entity recognition           Florian et al. (2003)
    Compound segmentation              Park et al. (2004)
    Disfluency detection               Kim et al. (2004)
    Semantic role labeling             Williams et al. (2004)
    Word alignment                     Ayan et al. (2005)
    Information extraction             Nahm (2005)
    Biomedical term normalization      Tsuruoka et al. (2008)
    Human activity recognition         Landwehr et al. (2008)
```
- trainer.brillorig is also kept -- I have used that both for reference and pedagogical purposes.
- I included a convenience `_run_doctests.py` to facilitate testing on different systems; should be merged
  into the larger test suite later
- Yaml is gone and the demo has an example of simple serialization (with pickle). However, there is no serialization to JSON at the moment and I won't have time to do it in a while. JSON seems to be optional, but nice (https://github.com/nltk/nltk/issues/540). Maybe @jskda could chip in?


stevenbird:
Thanks @muneson. I've merged this into the `brill` branch in https://github.com/nltk/nltk/commit/cc7dfbee3d57a55d439cf2c8e7671f20431b6735
@jskda would you please do your yaml to json conversion?

muneson:
> @jskda https://github.com/jskda would you please do your yaml to json
> conversion?

@jskda

I probably should have left the yaml commented out in the code, rather than
deleting it -- sorry
about that. Also, the associated commit message could have been clearer.
Anyway, this is the one:
https://github.com/muneson/nltk/commit/494c7c265d9d932b157c9f369e58668e84f456fa

2013/12/15 Steven Bird notifications@github.com

stevenxxiu:
>  I probably should have left the yaml commented out in the code, rather than deleting it -- sorry about that. Also, the associated commit message could have been clearer. Anyway, this is the one: https://github.com/muneson/nltk/commit/494c7c265d9d932b157c9f369e58668e84f456fa 

@stevenbird: Sure but since @muneson has already removed any dependency on yaml in those newer commits, should I still remove yaml from the set of old commits?

stevenbird:
@jskda -- oh, ok, no need to rewrite history. Can you add json serialization support?

stevenxxiu:
@stevenbird Sure thing, I've added it in the above pull request, I had to merge with the develop branch first as the jsontags file was in there.

stevenbird:
This pull request is merged into https://github.com/nltk/nltk/tree/brill, pending resolution of the issues around code organisation.

stevenbird:
Sorry for the delay on this; the fact that this calls for a different package organisation makes it harder to deal with.

stevenbird:
Sorry for the delay in handling this pull request @muneson. I've now had a crack at reorganising the code to make it fit with NLTK's existing structure. I've tried to put tagging-specific code in `nltk.tag`, and general-purpose TBL code in a new top-level `nltk.tbl` package. A list of changes follows, including some outstanding items at the end. I hope this is acceptable. I would be grateful if someone could test this (see the `brill` branch).

This is one of the last tasks remaining before we finalise NLTK 3.
- nltk.tag.tbl -> nltk.tbl  (promoted, for consistency with the rest of NLTK re package nesting)
- nltk.tbl.tagger -> nltk.tag.brill  (tagger still belongs in nltk.tag)
- nltk.tbl.trainer.fast -> nltk.tag.brill_trainer
- nltk.tbl.trainer.brillorig -> nltk.tag.brill_trainer_orig
- nltk.tbl.task.api -> nltk.tbl.feature (consistency with rest of NLTK's use of api.py)
- nltk.tbl.task.postagging -> nltk.tag.brill (templates are specific to the brill tagger; nltk tasks are top-level packages such as tag)
- nltk.**init** imports nltk.tbl but not its contents (since names like Feature and Rule are too generic)
- nltk.tbl.api (created empty file, which is where the package API would normally be defined; but I don't think this package requires a public API; it's just going to be used by other modules)

To do:
- does nltk.tbl.demo belong in nltk.tag.brill_demo?
- nltk.tbl.**init** should have a better module docstring (original one was specific to Brill Tagging)
- remove nltk.tbl._run_doctests

stevenbird:
Were there any API changes that deserve a mention here: https://github.com/nltk/nltk/wiki/Porting-your-code-to-NLTK-3.0?

stevenbird:
The old Brill tagger includes a module-level docstring (https://github.com/nltk/nltk/blob/develop/nltk/tag/brill.py). It would be great if we could have the corresponding thing in the new implementation.

PR #567, opened 2013-12-17
Update internals.py

ltiao:
It might be a good idea to make the distinction between capturing and non-capturing groups since technically, they are all grouping parentheses. I've only changed the docstring for this function slightly but maybe the function name shld be updated accordingly if others have this confusion in the future.


stevenbird:
Good idea, thanks.

Issue #558, opened 2013-12-05
Stanford tokenizer


stevenbird:
Add support for Stanford tokenizer: http://nlp.stanford.edu/downloads/tokenizer.shtml

stevenxxiu:
Should I implement `span_tokenize`? I can't find a way to do this directly on the commandline using `edu.stanford.nlp.process.PTBTokenizer`, so I'll need to add some java code.

stevenbird:
No, this is too low priority.

PR #568, opened 2013-12-18
Added interface to the stanford tokenizer

stevenxxiu:
Added interface to the stanford tokenizer, cf https://github.com/nltk/nltk/issues/558

### DIFF
--- a/nltk/tokenize/stanford.py
+++ b/nltk/tokenize/stanford.py
@@ -7,7 +7,7 @@
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 
-from __future__ import unicode_literals
+from __future__ import unicode_literals, print_function
 
 import tempfile
 import os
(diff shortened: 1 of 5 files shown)


nltk/tokenize/stanford.py
```diff
@@ -0,0 +1,95 @@
...
+            input_file.flush()
+
+            cmd.append(input_file.name)
+
+            # Run the tagger and get the output.
+            stdout, stderr = java(cmd, classpath=self._stanford_jar,
+                                  stdout=PIPE, stderr=PIPE)
+            stdout = stdout.decode(encoding)
+            if (not compat.PY3) and encoding == 'ascii':
+                stdout = str(stdout)
```
kmike: why not use unicode in 2.x?

stevenxxiu: I copied some of the code from `nltk.tag.stanford`, but I'm not sure why either.
I've added a commit which removes this conversion in both this and in `nltk.tag.stanford`.
Thanks.

stevenbird:
Please include a usage example (cf nltk.tokenize.treebank).

stevenxxiu:
I've now added a demo similar to that of `nltk.tokenize.treebank` and another one for the 'americanize' option.

PR #573, opened 2014-01-08
Pick up the latest version of the stanford parser model that is available

stevenxxiu:
Pick up the latest version of the stanford parser model that is available, using some regexes to search for the model.


stevenbird:
Add parser model to nltk_data and test that it can be found.

PR #574, opened 2014-01-13
Regex remove groups bug

stevenxxiu:
Fix to nltk#545, by using the python's sre module to parse regexes so all corner cases are covered.


stevenbird:
Thanks for these fixes. Do we have any regression tests?

stevenxxiu:
I tested this with `tokenize.doctest` which includes some tests for `regexp_tokenize`, I don't know if there are any other tests.

Issue #580, opened 2014-01-20
compile_regexp_to_noncapturing is broken


kmike:
Check this line: https://github.com/nltk/nltk/commit/764fd52d12096ba254b97ec50288ecdbfb26f939#diff-6d3d9c7d1794a239ab676fc6fce3131fR49 - this identifier is undefined. It broke `tokenize.doctest`. Yay tests, let's run them! :)

PR #582, opened 2014-01-21
Pull request.regexp noncapturing

### LINKED ISSUE #578: nltk.internals.compile_regexp_to_noncapturing doctest is broken
In Python 2.7:

```
FAIL: Doctest: nltk.internals.compile_regexp_to_noncapturing
  File "/usr/local/Cellar/python/2.7.5/Frameworks/Python.framework/Versions/2.7/lib/python2.7/doctest.py", line 2201, in runTest
AssertionError: Failed doctest test for nltk.internals.compile_regexp_to_noncapturing
  File "/Users/kmike/svn/nltk/nltk/__init__.py", line 38, in compile_regexp_to_noncapturing

File "/Users/kmike/svn/nltk/nltk/__init__.py", line 44, in nltk.internals.compile_regexp_to_noncapturing
Failed example:
    compile_regexp_to_noncapturing('ab(c(x+)(z*))?d')
Expected:
    'ab(?:c(?:x+)(?:z*))?d'
Got:
    <_sre.SRE_Pattern object at 0x7fa09e70b2e0>
```

stevenxxiu:
Fixes https://github.com/nltk/nltk/issues/578 and https://github.com/nltk/nltk/issues/580

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -31,12 +31,8 @@ from nltk import compat
 
 def compile_regexp_to_noncapturing(pattern, flags=0):
     """
-    Convert all grouping parentheses in the given regexp pattern to
-    non-capturing groups, and return the result.  E.g.:
-
-        >>> from nltk.internals import compile_regexp_to_noncapturing
-        >>> compile_regexp_to_noncapturing('ab(c(x+)(z*))?d')
-        'ab(?:c(?:x+)(?:z*))?d'
+    Compile the regexp pattern after switching all grouping parentheses
+    in the given regexp pattern to non-capturing groups.
 
     :type pattern: str
     :rtype: str
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -46,7 +46,7 @@ def compile_regexp_to_noncapturing(pattern, flags=0):
         for key, value in parsed_pattern.data:
             if key == sre_constants.SUBPATTERN:
                 index, subpattern = value
-                value = (None, convert_regexp_to_noncapturing(subpattern))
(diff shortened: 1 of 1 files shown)


nltk/internals.py
```diff
@@ -31,12 +31,8 @@
 
 def compile_regexp_to_noncapturing(pattern, flags=0):
     """
-    Convert all grouping parentheses in the given regexp pattern to
-    non-capturing groups, and return the result.  E.g.:
-
-        >>> from nltk.internals import compile_regexp_to_noncapturing
-        >>> compile_regexp_to_noncapturing('ab(c(x+)(z*))?d')
-        'ab(?:c(?:x+)(?:z*))?d'
```
kmike, line 39: there is `regexp_obj.pattern` attribute with pattern as text - what do you think about keeping this doctest?

stevenxxiu, line 39: I would like to but I don't think `sre.compile` converts a parsed pattern back into text. For the current case the `pattern` attribute is set to `None`. I couldn't find a routine in `sre` to do this either.
Despite this drawback I still think this is better than the previous version which parses regexes manually though as this covers a lot more corner cases.

kmike, line 39: I've missed we're using `sre_compile.compile` instead of `re.compile`. Why do we use it? `re.compile` uses `sre_compile.compile` internally and implements a cache for recent patterns.

stevenxxiu, line 39: I don't think `re.compile` accepts parsed patterns, only strings or compiled patterns.

kmike, line 39: Thanks for the explanation!

Issue #581, opened 2014-01-20
nltk.test.unit.test_collocations is broken


kmike:
Using Python 2.7:

```
======================================================================
FAIL: nltk.test.unit.test_collocations.test_bigram2
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/kmike/svn/nltk/.tox/py27-nodeps/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/kmike/svn/nltk/nltk/test/unit/test_collocations.py", line 29, in test_bigram2
    [(('a', 'a'), 1), (('a', 'test'), 1), (('is', 'a'), 1), (('is', 'is'), 1), (('test', 'test'), 1), (('this', 'is'), 1), (('this', 'this'), 1)]
AssertionError

======================================================================
FAIL: nltk.test.unit.test_collocations.test_bigram3
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/kmike/svn/nltk/.tox/py27-nodeps/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/kmike/svn/nltk/nltk/test/unit/test_collocations.py", line 43, in test_bigram3
    [(('a', 'test'), 3), (('is', 'a'), 3), (('this', 'is'), 3), (('a', 'a'), 1), (('is', 'is'), 1), (('test', 'test'), 1), (('this', 'this'), 1)]
AssertionError

======================================================================
FAIL: nltk.test.unit.test_collocations.test_bigram5
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/kmike/svn/nltk/.tox/py27-nodeps/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/kmike/svn/nltk/nltk/test/unit/test_collocations.py", line 57, in test_bigram5
    [(('a', 'test'), 4), (('is', 'a'), 4), (('this', 'is'), 4), (('is', 'test'), 3), (('this', 'a'), 3), (('a', 'a'), 1), (('is', 'is'), 1), (('test', 'test'), 1), (('this', 'this'), 1)]
AssertionError
```

jnothman:
Oh! Any idea when this began to fail? It was only introduced recently, in
#414.

Also, that's not a very useful error message. Can I propose that we use
things like nose or unittest's assert_equal to get better output upon
failure?

[...]

PR #583, opened 2014-01-21
fixes unit testing bugs for https://github.com/nltk/nltk/issues/581

stevenxxiu:
Fixes unit testing bugs for https://github.com/nltk/nltk/issues/581

I added unittest2 as an import as python 2.6 doesn't have any list comparison asserts, so cleaner code can be used. Unittest2 is also available on python 3 as unittest2py3k.


kmike:
We don't depend on unittest2 currently - it should be added to tox.ini and to testing docs if we add this dependency.

But I think it isn't worths adding it here, and checking sorted output is fine.

stevenxxiu:
You're right, I've added `sorted` back now.

kmike:
Thanks!

PR #584, opened 2014-01-22
nltk tokenize doctest non-existent import

stevenxxiu:
Removed nltk tokenize doctest non-existent import.

### DIFF
--- a/nltk/tokenize/__init__.py
+++ b/nltk/tokenize/__init__.py
@@ -14,7 +14,7 @@ Tokenizers divide strings into lists of substrings.  For example,
 tokenizers can be used to find the list of sentences or words in a
 string.
 
-    >>> from nltk.tokenize import word_tokenize, wordpunct_tokenize, sent_tokenize, tokenize
+    >>> from nltk.tokenize import word_tokenize, wordpunct_tokenize, sent_tokenize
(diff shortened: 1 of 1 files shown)


kmike:
There are also references to this function later in docsting. It seems Steven wanted to rename `wordpunct_tokenize` to `tokenize`, but I'm not sure. @stevenbird - could you please clarify it?

kmike:
(I merged this PR too early, sorry for that)

stevenxxiu:
The doctest seems to work for me, @kmike which ones are you referring to?

kmike:
The references are in text, they don't affect test running:

> This is a simple regular-expression based tokenizer which does not handle
>  periods very well. We can also use a sentence tokenizer (Punkt) and followed
>  by a good sentence-internal word tokenizer (Treebank). The `tokenize()`
>  function combines both of these.

stevenbird:
Thanks for picking this up @kmike. Yes, I intended to create a `tokenize()` function that combined sentence tokenization and sentence-internal word tokenization. The name is unfortunate in colliding with the package name (`nltk.tokenize`) but I think its workable. I'll open a separate issue to make sure we deal with this before nltk3 beta is finalized.

Issue #577, opened 2014-01-20
StanfordTokenizer doctest is broken


kmike:
Here: https://github.com/nltk/nltk/blob/develop/nltk/tokenize/stanford.py#L29 

"\n" is interpreted as a new line _in a docstring_, not in a constant. This makes this doctest broken.

PR #587, opened 2014-01-23
Fixes to #577

stevenxxiu:
Fixed bugs in stanford tokenizer doctests.
Added code to skip doctests if the jar doesn't exist.
Tested the doctest for python 2.6 and 3.2, they should hopefully work now.

### DIFF
--- a/nltk/parse/stanford.py
+++ b/nltk/parse/stanford.py
@@ -27,11 +27,11 @@ class StanfordParser(ParserI):
     Interface to the Stanford Parser
 
     >>> parser=StanfordParser(
-    ...     model_path='edu/stanford/nlp/models/lexparser/englishPCFG.ser.gz'
+    ...     model_path="edu/stanford/nlp/models/lexparser/englishPCFG.ser.gz"
     ... )
     >>> parser.raw_batch_parse((
-    ...     'the quick brown fox jumps over the lazy dog',
-    ...     'the quick grey wolf jumps over the lazy fox'
+    ...     "the quick brown fox jumps over the lazy dog",
+    ...     "the quick grey wolf jumps over the lazy fox"
     ... ))
     [Tree('ROOT', [Tree('NP', [Tree('NP', [Tree('DT', ['the']), Tree('JJ', ['quick']), Tree('JJ', ['brown']),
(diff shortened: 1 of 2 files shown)


nltk/tokenize/stanford.py
```diff
@@ -21,20 +21,28 @@
 
 _stanford_url = 'http://nlp.stanford.edu/software/lex-parser.shtml'
 
+def setup_module(module):
```
kmike: Can we move it to the bottom of the file to make the source code more readable?

stevenxxiu: Sure thing, I've moved it now.

nltk/tokenize/stanford.py
```diff
@@ -26,15 +26,15 @@ class StanfordTokenizer(TokenizerI):
     Interface to the Stanford Tokenizer
 
     >>> from nltk.tokenize.stanford import StanfordTokenizer
-    >>> s = 'Good muffins cost $3.88\nin New York.  Please buy me\ntwo of them.\nThanks.'
+    >>> s = 'Good muffins cost $3.88\\nin New York.  Please buy me\\ntwo of them.\\nThanks.'
```
kmike: One option is to add 'r' before the whole docstring, it could be easier to read this way (outside the python shell)

stevenxxiu: Sure thing I've fixed that now.

nltk/parse/stanford.py
```diff
@@ -51,8 +41,8 @@ class StanfordParser(ParserI):
     Tree('JJ', ['lazy']), Tree('NN', ['fox'])])])])])])]
 
     >>> parser.batch_parse((
-    ...     'I \\\'m a dog'.split(),
-    ...     'This is my friends \\\' cat ( the tabby )'.split(),
+    ...     'I \'m a dog'.split(),
```
kmike: wow I'm picky today :) Can we use double quotes here?

stevenxxiu: Fixed that. How about now :)

kmike:
Looks great now, thanks! :)

stevenbird:
Thanks :)

PR #588, opened 2014-01-23
Fixes for jenkins

stevenxxiu:
Fixes for jenkins so the stanford parser interfaces can be tested.


stevenbird:
Thanks for this. Ideally https://github.com/jskda/nltk/commit/07f04403ff6152aa33647e3c579e7fa2500d476f would be separate!

stevenxxiu:
Thanks I'll do that next time.

PR #589, opened 2014-01-25
Small fix in chat80 doctest

stevenxxiu:
This fixes the chat80 doctest so it is up to date with 'book_grammars/sql0.fcfg'.


ewan-klein:
Thanks for this.

PR #515, opened 2013-10-26
Fix stem.doctest failures

usami:
I run doctest for stemmers and got the following failures:

```
$ python -m doctest nltk/test/stem.doctest
**********************************************************************
File "nltk/test/stem.doctest", line 53, in stem.doctest
Failed example:
    stemmer.stem("hurried")
Expected:
    'hurri'
Got:
    u'hurri'
**********************************************************************
File "nltk/test/stem.doctest", line 56, in stem.doctest
Failed example:
    stemmer.stem("advisable")
Expected:
    'advis'
Got:
    u'advis'
**********************************************************************
File "nltk/test/stem.doctest", line 69, in stem.doctest
Failed example:
    print(" ".join(SnowballStemmer.languages))
Expected:
    danish dutch english finnish french german hungarian italian
    norwegian porter portuguese romanian russian spanish swedish
Got:
    danish dutch english finnish french german hungarian italian norwegian porter portuguese romanian russian spanish swedish
**********************************************************************
1 items had failures:
   3 of  23 in stem.doctest
***Test Failed*** 3 failures.
```

The first and second failures were caused because doctest did not expect unicode output. The last failure was caused by missing NORMALIZE_WHITESPACE option.

This pull request resolves these issues and now stemmers can pass all tests.

PR #593, opened 2014-02-01
Changed URL to account for branch rename

pbronez:
The URL was broken when the "Master" branch was renamed to "Develop." This was pointed out in issue #515


kmike:
Thanks a lot!

stevenbird:
Thanks @pbronez

PR #605, opened 2014-02-12
Add initial version of BLLIP Parser wrapper

dmcc:
This covers the main functionality of BLLIP: 1-best and n-best with and without tag constraints. The `demo()` function downloads and install a basic parsing model to show these capabilities.

I haven't used `sphinx` before, so please let me know if I should fix anything with the docstrings (well, or anything else for that matter).


nltk/parse/bllip.py
```diff
@@ -0,0 +1,270 @@
...
+class BllipParser(ParserI):
+
+    """
+    An interface for parsing with the BLLIP Parser using a
+    SWIG wrapper.  Parsers can be constructed manually (using the
+    constructor) or with the ``BllipParser.from_unified_model_dir``
+    class method. The latter is generally easier if you have
+    a unified model directory which can be obtained with the
+    ``bllipparser.ModelFetcher.download_and_install_model`` function or
+    the command-line utility ``python -mbllipparser.ModelFetcher``.
```
kmike: I think `python -m bllipparser.ModelFetcher` reads better.

nltk/parse/bllip.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _check_bllip_import_success():
+        pass
+except ImportError, ie:
+    def _check_bllip_import_success(ie=ie):
+        raise RuntimeError("Couldn't import bllipparser module: %s" % ie)
+
+
+def _nbest_list_item_to_tree(nbest_list_item):
+    return Tree(str(nbest_list_item.ptb_parse))
```
kmike: It seems this won't work with unicode because you cast `nbest_list_item.ptb_parse` to str. What is the type of `nbest_list_item.ptb_parse`? This may raise an exception for non-ascii data.

kmike:
Good work!
- Is it is possible to make a trained model available in nltk data? It seems that ModelFetcher has the same purpose as `nltk.downloader`.
- Some automated tests would be great.
- What do you think about adding references to papers to a module or class docstring? I see there are references at https://pypi.python.org/pypi/bllipparser/, maybe copy some of the info from there?

nltk/parse/bllip.py
dmcc: `nbest_list_item.ptb_parse` is an `InputTree` object (wrapped SWIG object -- see https://github.com/BLLIP/bllip-parser/blob/master/first-stage/PARSE/swig/wrapper.i#L347 and https://github.com/BLLIP/bllip-parser/blob/master/first-stage/PARSE/InputTree.h). I'm don't think BLLIP Parser supports Unicode internally properly and a quick test indicates that it doesn't. What does this mean for NLTK wrapping? Would it be okay to have it raise an error if asked to parse non-ASCII text?

kmike: Do you know if BLLIP Parser supports non-ascii data encoded to some encoding (e.g. to utf8)? If so, a good way to work with it is to encode/decode data in NLTK wrapper or in BLLIP Python wrapper. If it dosn't support non-ascii data then raising an exception for non-ascii input (with some explanation) sounds like a good idea.

dmcc:
Thanks and thanks for the feedback! I agree with all those points.

What would I do to make the parsing model(s) available to the downloader? Would that involve just making a pull request to the nltk_data repo? The models are possibly big, depending on your scale (smallest is ~25MB, bigger ones are hundreds).

I'll put some automated tests into the main class's docstring (would be based on what's in `demo()` currently). This ties into the model downloading, though -- is there a module that needs to download something in order to test that I can use for a basis?

References to papers also make a lot of sense, will add those to the docstrings as well.

kmike:
Let's summon @stevenbird - he knows more about nltk data. Maybe we can make a subset of models available to the downloader. 25MB is certainly not a problem.

Creating a test based on demo is a good idea.

NLTK tests usually assume the data is already downloaded. You can skip the test if data is not available, but we are not doing this for existing tests. We usually skip a test if a dependency (i.e. BLLIP parser itself) is not available. Check https://github.com/nltk/nltk/blob/develop/web/dev/local_testing.rst for more info about NLTK testing - there are some non-obvious details.

nltk/parse/bllip.py
```diff
@@ -0,0 +1,270 @@
...
+from nltk.parse.api import ParserI
+from nltk.tree import Tree
+
+try:
+    from bllipparser import RerankingParser
+    from bllipparser.RerankingParser import get_unified_model_parameters
+
+    def _check_bllip_import_success():
+        pass
+except ImportError, ie:
```
stevenbird: Please use Python 3 syntax (backported to Python 2.6 and 2.7), `except ImportError as ie`, http://docs.python.org/3/howto/pyporting.html#capturing-the-currently-raised-exception

nltk/parse/bllip.py
stevenbird: +1 to testing out this idea

stevenbird:
Let's make the 25MB model available via nltk_data, to make it easy for users to get started. Please just tell me where to find it. The module docstring can document where users can download larger models. Would it be ok to drop ModelFetcher, since this isn't consistent with the rest of the toolkit?

dmcc:
Sounds great! The vanilla WSJ model is at [here](http://nlp.stanford.edu/~mcclosky/models/BLLIP-WSJ-no-AUX.tar.bz2) (22MB compressed, 52MB uncompressed).

By dropping `ModelFetcher`, do you mean removing it from `demo()` and replacing it with the NLTK downloader equivalent? That makes sense to me. I may still need to reference `ModelFetcher` as a method of finding the other models (though ideally, there'd be a webpage for this as well). If NLTK downloader can point to arbitrary URLs, then we may be able remove all references to `ModelFetcher`.

nltk/parse/bllip.py
dmcc: I'm not sure I understand the details of Unicode well enough to investigate completely, actually, so any help here would be greatly appreciated. There's the added complication here of possible interactions between C++, SWIG, and Python.

BLLIP does have a Chinese model which was built from [CTB5](http://catalog.ldc.upenn.edu/LDC2005T01) (I think) and uses GB encoding, so there seems to be some support for non-ASCII encoding. Still, I'm not confident in my ability to test this and will have to read up on this.

Here's a possible hack^Wworkaround for the English model -- if a token includes non-ASCII characters, BLLIP doesn't actually care too much about what the word is (since it will be an unknown word, only the initial capitalization, whether it has a hyphen, and its last two letters will matter). So, we could actually replace the word with it's nearest ASCII equivalent before parsing and then stick the original token text back into the NLTK `Tree`. Does this seem like too much of a hack? (After writing it, I'm beginning to think that it might be...)

dmcc:
The latest patch is improved (hopefully) but probably not final. It includes an attempt to use `nltk.download()` even though there's no BLLIP model in `nltk_data` yet. Currently, it expects the model file to be called `bllip-wsj`. Once I can test the `nltk.download()` code path, I'll remove references to `bllipparser.ModelFetcher` in `demo()` and the doctests (I think it should still be mentioned in the docstrings as a means for obtaining bigger models).

Also, should I add `bllipparser` as an optional dependency in `pip-req.txt`?

stevenbird:
Sorry for the delay.

stevenbird:
@dmcc would you please point me to the data so I can add `bllip-wsj` to `nltk_data`?

dmcc:
@stevenbird Sure, it's currently here: http://nlp.stanford.edu/~mcclosky/models/BLLIP-WSJ-no-AUX.tar.bz2

stevenbird:
Sorry this won't make it into 3.0b1. I hope to get it in before 3.0 is finalised.

dmcc:
Is there anything I can do to help get this pull request merged? IIRC, there have been some parsing API changes and I can work on adjusting this pull request. I'd also be happy to make a pull request in `nltk_data` which adds the BLLIP parsing model.

stevenbird:
@dmcc sorry for the lengthy delay! I've added the model, so you won't need the download functionality any more. Once you've updated the PR I'll pull it into a new branch for testing. Thanks.

dmcc:
@stevenbird No worries. I think I've done the merging correctly but let me know if there were any issues. Thanks!

stevenbird:

stevenbird:
Thans @dmcc. Sorry for the delay. I've finally had time to check this, but can't compile the parser (because of [this known issue](https://github.com/BLLIP/bllip-parser/issues/19)). The code in `bllip.py` meets our coding standards and implements `ParserI`, and doesn't interfere with anything else, so I'm just going to merge this. I'll remove the code that downloads the corpus since this would be very wasteful on our CI server.

PR #615, opened 2014-02-20
Extend the BNC corpus reader with more tags.

dimazest:
BNC is now available at http://www.ota.ox.ac.uk/desc/2554

These are the changes I had to make to be able to read this version of the BNC.

### DIFF
--- a/nltk/corpus/reader/bnc.py
+++ b/nltk/corpus/reader/bnc.py
@@ -5,10 +5,7 @@
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 
-"""
-Corpus reader for the XML version of the British National Corpus.
-"""
-__docformat__ = 'epytext en'
+"""Corpus reader for the XML version of the British National Corpus."""
 
 from nltk.corpus.reader.api import *
 from nltk.corpus.reader.util import *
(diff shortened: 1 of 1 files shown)


nltk/corpus/reader/bnc.py
```diff
@@ -16,11 +13,20 @@
...
     method.  For access to simple word lists and tagged word lists, use
     ``words()``, ``sents()``, ``tagged_words()``, and ``tagged_sents()``.
+
+    You can obtain the full version of the BNC corpus at
+    http://www.ota.ox.ac.uk/desc/2554
+
+    If you extracted the archive to a directory called `BNC`, then you can
+    instantiate the reder as::
+
+        BNCCorpusReader(root='BNC/Texts/', fileids=r'[A-K]/\w*/\w*\.xml')
```
kmike, line 28: should we make this value for fielids default?

stevenbird, line 28: Will this work across BNC versions?

dimazest, line 28: i'm not sure. maybe it worth to leave it the documentation, so there is a reasonable hint for the parameter.

dimazest, line 28: should i then remove the hint, or leave it as it is?

PR #619, opened 2014-02-21
ISRI stemmer: Corrections to logic (and many style fixes)

Tahnan:
There are two commits here. The first is a simple bug fix that resulted in the literal string "\u0627" being put into stemmed Arabic forms instead of the unicode character \u0627. The second is a more substantive change, fixing two fundamentally flawed points in the logic of the code, and cleaning up a lot of things stylistically.

Please do note, as I say in the latter commit, that I don't speak any Arabic, and I didn't compare this to the original paper laying out the ISRI stemmer. So while I left most of the substance of the code alone, I'm not trying to implicitly endorse it as correct; I was only changing things that I knew couldn't be correct.

### DIFF
--- a/nltk/stem/isri.py
+++ b/nltk/stem/isri.py
@@ -34,6 +34,7 @@ import re
 
 from nltk.stem.api import StemmerI
 
+
 class ISRIStemmer(StemmerI):
     '''
     ISRI Arabic stemmer based on algorithm: Arabic Stemming without a root dictionary.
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for this contribution -- I appreciate the rationale of fixing the logic and cleaning the code. Let me reach out to the original author Hosam Algasaier to see if he has any comments.

stevenbird:
I haven't heard back, so I'm merging this. Thanks for the great contribution.

PR #625, opened 2014-02-22
Remove 'SVN Revision' from the downloader GUI

alexrudnick:
... we might want to add a similar field later, but right now "SVN Revision" can never be set!

Tested this out: when this is removed from the list, "SVN Revision" simply doesn't appear in the dropdown of possible fields to display.


stevenbird:
Thanks @alexrudnick, well spotted.

PR #627, opened 2014-02-24
Parser bugfixes

heatherleaf:
Some bugfixes. Now all grammar and parser doctests pass: parse, grammar, grammartestsuites, featgram, featstruct

### DIFF
--- a/nltk/app/chartparser_app.py
+++ b/nltk/app/chartparser_app.py
@@ -1690,7 +1690,8 @@ class ChartParserApp(object):
         self._chart = self._cp.chart()
 
         # Insert LeafEdges before the parsing starts.
-        LeafInitRule().apply(self._chart, self._grammar)
+        for _new_edge in LeafInitRule().apply(self._chart, self._grammar):
+            pass
 
         # The step iterator -- use this to generate new edges
         self._cpstep = self._cp.step()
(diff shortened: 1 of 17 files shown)


stevenbird:
@heatherleaf thanks for all these fixes! Much appreciated. Do you think that `parse()` should return an iterator instead of a list? I had had second thoughts about that (see https://github.com/nltk/nltk/commit/cfd6dd54583e104ff475b579b73a201f93cd34b6) because in many cases there was no efficiency benefit. I don't feel strongly about this, but would like to capture the thinking here for future reference if possible, thanks.

adonoho:
Steven,

I would always bias NLTK towards using iterators instead of lists. Judiciously using memory is important to retaining high performance. While the array itself isn't typically that expensive, it is what needless allocation does to your working set in first and second level caches that matters. You can help beat the reputation NLTK unfortunately has for low performance by using iterators when possible.

Anon,
Andrew

heatherleaf:
@stevenbird Currently there's no space improvement when using iterators instead of lists, at least not for the chart parsers. The reason is that the `Chart` methods `parses` and `trees` return lists. But I was thinking I'd change them to return iterators instead, and then we will see much better memory footprints.

stevenbird:
@adonoho right, but other factors include consequences for the calling program, and the fact that few of our current parser implementations do anything but convert a list to an iterator as their last step.
@heatherleaf ok, let's stick with iterators

PR #634, opened 2014-02-26
more iterators in parsers

heatherleaf:
I changed almost all .parse() methods to return iterators. The exceptions are StanfordParser and MaltParser which I don't know how to handle.

I also changed the methods .parses() and .trees() to return iterators.

Finally I merged commit 8c5f9e7 (removing .prob_parse()), which is the reason why there are so many commits referred in this pull request. Sorry for that


heatherleaf:
I spent some hours trying to make the ChartParser even more "iterative", but I didn't succeed. The tree building (method Chart._trees()) is quite ingenious: it really creates a parse forest, since all subtrees that can be shared are shared, which means that it is quite compact already as it is.

To make it more iterative, I had to remove the memoization, which made it slower in almost all cases. I tried on a very ambiguous grammar (S -> S S), and got these results:
- when I collect all trees, the iterator version is 10 times slower and uses 8 times more memory
- when I just count the number of trees, the iterator version is still 10 times slower, but it uses constant memory (while the memoized version uses memory proportional to the number of trees)
- when I want the first tree, the iterator version is faster overall

My conclusion is that the current memoized version is better in most cases, so I think we should stick to that.

stevenbird:
Thanks for exploring the possibilities with the ChartParser code -- good to know it stands up to scrutiny.

Issue #50, opened 2012-01-17
corpus reader docstrings


alexrudnick:
The corpus reader docstrings should have a reasonable amount of
information, and be consistent from one corpus to the next, cf:

From issue 201:
&quot;nltk.corpus.verbnet.**doc** gives very little information, so it could be
good to describe that a little.&quot;

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=252</p>

paulproteus:
To illuminate the task here for a newcomer:
- Note that at https://github.com/nltk/nltk/blob/master/nltk/corpus/reader/verbnet.py , nothing says what Verbnet _is_.
- Note also that therefore, at http://nltk.org/api/nltk.corpus.reader.html?highlight=verbnet#nltk.corpus.reader.VerbnetCorpusReader , in the generated HTML, nothing says what Verbnet is.

Your task is to notice the kind of description that BNCCorpusReader has (which you can see on the BNCCorpusReader page), go surf the web to figure out what Verbnet is, and then a simple explanation for the Verbnet code that is of similar quality as the description of BNCCorpusReader.

pancho-villa:
I think this is a better link to that verbnet source code:
https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/verbnet.py

Now I'm off to figure out what is the definition of a verbnet.

zoe-hammar:
I just submitted a pull request addressing this issue.  Here's what I did.  I copied the existing docstring into the correct spot so that it would show up in the auto-generated documentation. I also quoted a couple of descriptive sentences from the existing link that explains what VerbNet is.

This is my first time doing this, so please let me know if anything I did could be improved.

zoe-hammar:
This issue is resolved (unless I'm missing something), so it should be closed.

erip:
What's the status of this issue? Should it be closed?

ewan-klein:
It looks to me as though @justinhammar's https://github.com/nltk/nltk/commit/a9fa3e3cbd83ea77eb8958f0a749970e0d82b4ab only addresses two instances of the issue, i.e. docstrings for WordNet and Verbnet readers. Are there other corpus readers still lacking docstrings?

stevenbird:
Right, this is a more general issue. Things would be significantly better if help(c) caused corpus c to be loaded. (To see why, do: `from nltk.corpus import c; help(c); c.method(); help(c)`)

PR #636, opened 2014-02-27
Add docstrings for wordnet and verbnet

dougalg:
In response to https://github.com/nltk/nltk/issues/50

Would it be better for me to go through more of the documentation tickets all at once and add them here?
Is this a sufficient description? I wasn't sure if the links were considered appropriate or not, but I thought they'd be useful.


stevenbird:
Thanks for this. If you would be able to contribute to other documentation tickets that would be great. You could just mention each ticket in your commit message, and we'll see which ones can be closed. Thanks.

PR #643, opened 2014-03-10
Unit test fixes

stevenxxiu:
Unit test fixes, including:
- deterministic ordering in probability.doctest


stevenbird:
Thanks @jskda

PR #645, opened 2014-03-14
Unit test fixes

stevenxxiu:
unit test fixes, including:
- fixed parse doctests by sorting the grammar productions

### DIFF
--- a/nltk/test/corpus.doctest
+++ b/nltk/test/corpus.doctest
@@ -697,11 +697,11 @@ distribution of part-of-speech tags for reduplicated words.
     ...     lexeme = record.find('lx').text
     ...     if re.match(r'(.*)\1$', lexeme):
     ...         redup_pos_freqdist[record.find('ps').text] += 1
-    >>> for item in sorted(redup_pos_freqdist.keys()):
-    ...     print(item, redup_pos_freqdist[item])
-    ??? 4
-    N 14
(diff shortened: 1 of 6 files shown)


nltk/test/corpus.doctest
```diff
@@ -697,11 +697,11 @@ distribution of part-of-speech tags for reduplicated words.
     ...     lexeme = record.find('lx').text
     ...     if re.match(r'(.*)\1$', lexeme):
     ...         redup_pos_freqdist[record.find('ps').text] += 1
-    >>> for item in redup_pos_freqdist.keys():
+    >>> for item in sorted(redup_pos_freqdist.keys()):
```
kmike: maybe use `.most_common()` method?

stevenxxiu: `.most_common()` has the problem that some frequencies might be the same (not here though), hence not be deterministic.

kmike: Yes, that's correct. On the other hand, doctests are not only tests, but also a documentation; `sorted(...)` is a bit less readable, and it makes a worse example - users are more likely to use `.most_common` than `sorted` in their own code.

stevenxxiu: Makes sense, I've changed it to `most_common()` now.

stevenbird:
Thanks @kmike, @jskda.

PR #647, opened 2014-03-23
Unit test fixes

stevenxxiu:
Unit test fixes

I've changed the naive bays behavior slightly, so that the 'None' feature is only added when there are actually missing features in some instances, so it doesn't appear when unnecessary, e.g. in unit tests.

Is this acceptable?

### DIFF
--- a/nltk/test/unit/test_hmm.py
+++ b/nltk/test/unit/test_hmm.py
@@ -69,7 +69,7 @@ def test_backward_probability():
     wikipedia_results = [
         # Forward-backward algorithm doesn't need b0_5,
         # so .backward_probability doesn't compute it.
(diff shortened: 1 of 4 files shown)


nltk/test/unit/test_hmm.py
```diff
@@ -72,10 +70,10 @@ def test_backward_probability():
         # Forward-backward algorithm doesn't need b0_5,
         # so .backward_probability doesn't compute it.
         # [0.3531, 0.6469],
```
kmike: I think this line also needs to be swapped

kmike:
I like HMM changes; haven't looked into NB changes.

PR #648, opened 2014-03-24
changed numpy.stdev = ... to stdev = ... in texttiling algorithm

cjrd:
no apparent reason to assign the std of the scores to the numpy object


kmike:
Thanks! No apparent reason indeed. If the intention was to make stdev value available for debugging there are better ways than changing a global variable.

Issue #658, opened 2014-04-22
Standardise method for initialisation from strings


stevenbird:
Several classes now have a `read()` method which can be used for initialising an instance of the class from a string (cf tree.py, featstruct.py, grammar.py chunk/regexp.py). We need to audit the various ways that a string can be used to initialise a class, and try to standardise on using `read()`.

kmike:
Sorry if I'm too late, but I kind of dislike `read()` name: in Python it is usually used for reading data from a stream, and its argument is usually a number of bytes to read. This is the case for file objects, and objects from standard io, urllib, csv, json, configparser, codecs, zipfile and many other libraries. 

What is called `read()` in NLTK can be called `fromstring()` (see array, xml standard modules), `loads()` or `load()` (see json, pickle, cookielib) or `parse` (see ast, robotparser, string) or `parsestr` (email, xml.minidom); there are a few other variants, but it is never called `read()`.

heatherleaf:
Good point. I vote for `fromstring()`, but humbly let @stevenbird make the final decision.

stevenbird:
Yes, good point, and I also prefer `fromstring()` to the alternatives.

PR #659, opened 2014-04-25
renamed read() methods to fromstring(), c.f. https://github.com/nltk/nltk/issues/658

stevenxxiu:
renamed read() methods to fromstring(), c.f. https://github.com/nltk/nltk/issues/658

### DIFF
--- a/examples/semantics/model0.py
+++ b/examples/semantics/model0.py
@@ -33,7 +33,7 @@ v = [('john', 'b1'),
 
 
 #Read in the data from C{v}
-val.read(v)
+val.fromstring(v)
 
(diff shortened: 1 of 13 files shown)


examples/semantics/model0.py
```diff
@@ -33,7 +33,7 @@
 
 
 #Read in the data from C{v}
-val.read(v)
+val.fromstring(v)
```
kmike, line 36: This is confusing: `v` is not a string here

stevenxxiu, line 36: I've fixed this including some other fixes here: https://github.com/nltk/nltk/pull/660

PR #666, opened 2014-05-08
Updates doctests to use fromstring() method instead of read().

Liechti:
Updates the doctests to reflect the changes made in [#658](https://github.com/nltk/nltk/issues/658). Also,  `semantics.doctest` was updated to account for when the `batch_interpret/evaluate/parse` functions were [renamed here](https://github.com/stevenbird/nltk/commit/1391c620cb86a7c7d8514279c00685eaf82d33a5) and moved into `nltk.sem.util`.


stevenbird:
Thanks @Liechti this is great.

PR #667, opened 2014-05-09
Doctest fixes for renaming in d566f23 and more read() -> fromstring() changes

Liechti:
Missed a few `read() -> fromstring()` changes in my previous commit 227c157.

The tests `featgram.doctest,  featstruct.doctest, relextract.doctest` all pass now.

For the renaming of  `mk_pairs() -> _tree2semi_rel()` in d566f23, does it make sense for the doctest to test an underscored method?


stevenbird:
Thanks @Liechti. @ewan-klein what do you advise re `_tree2semi_rel()` please?

PR #671, opened 2014-05-17
fixes to align.doctest

stevenxxiu:
fixes to align.doctest, got rid of aligned() as it doesn't seem to exist anymore in `nltk.align.ibm1`.

### DIFF
--- a/nltk/align/bleu.py
+++ b/nltk/align/bleu.py
@@ -89,8 +89,8 @@ class BLEU(object):
     @staticmethod
     def compute(candidate, references, weights):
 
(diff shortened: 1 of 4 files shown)


stevenxxiu:
I couldn't find the bleu test data, including 'newstest2012-ref.en' and 'newstest2012.fr-en.cmu-avenue'

stevenbird:
Those files can be found here: https://github.com/langtech/clir/tree/master/src/ibm_bleu. It would be better to use data already included with NLTK, like europarl_raw (https://github.com/nltk/nltk_data/blob/gh-pages/packages/corpora/europarl_raw.zip).

PR #674, opened 2014-05-19
MultiSvmClassifier

directorli:
I write a program to package svmlight into MultiClassifier.And I have test it on a dataset from UCI about education level. Also ,it has some problems,please help build it.
Many Thanks!


kmike:
Hi @directorli,

What is the advantage of this wrapper over scikit-learn SVMs (see https://github.com/nltk/nltk/blob/develop/nltk/classify/scikitlearn.py)?

kmike:
See also: https://github.com/nltk/nltk/issues/450

directorli:
It can classify several classes not only two classes.On some dataset it has better behavior than(naviebayes,maxent,decisiontree).I've done  some test.I can send to you if needed

kmike:
The question is not whether SVM is more suited to some problems than (naviebayes,maxent,decisiontree), the question is whether adding svmlight-based implementation to NLTK makes sense, given that there is already an interface to scikit-learn's SVMs. They also handle multiclass output, pysvmlight wrapper has issues (doesn't work in Python 3, sometimes segfaults);  scikit-learn license is more permissive (svmlight is free only for non-commercial usage).

directorli:
Well,I haven't install scikit.I don't know it has Svms.Maybe they are same

stevenbird:
Thanks for the proposed contribution @directorli. I'm going to close this for now, until there's some case for your approach cf scikit-learn.

Issue #675, opened 2014-05-28
Tree initializer


stevenbird:
The `Tree` class is unusual in NLTK in allowing initialization from lists or from strings:

```
>>> Tree('dp', [Tree('d', ['the']), Tree('np', ['dog'])]) == Tree('(dp (d the) (np dog))')
True
```

I propose that we drop the string initialization, for consistency with the rest of NLTK. Users can still do this:

```
>>> Tree.fromstring('(dp (d the) (np dog))')]
```

SocialfiPanda:
I totally agree. I think I can take this one.

stevenbird:
Actually, I'm keen to have some more discussion before going ahead with this change.

heatherleaf:
I'm all for consistency. But we need to decide on the behaviour of `Tree("a")`:
- either it's illegal
- or it means the same as `Tree("a", [])`

I think it should be illegal, or the risk will be that people start writing `Tree("(a b c)")` and don't get what they think.

SocialfiPanda:
Yeah, letting `Tree("(a b c)") == Tree("(a b c)", [])` can produce some hard to trace bug for the user, wondering why they do not get the tree structure they intend to get. At least for me, throwing an exception for Tree initialization with only one string argument is a sane way to go.

stevenbird:
In the absence of any objections I'm going to accept this pull request. Thanks @jw2013, @kmike, and @heatherleaf for your helpful work!

PR #677, opened 2014-06-01
Drop the string initialization of Tree

SocialfiPanda:
This is the pull request for fixing issue #675.

### DIFF
--- a/nltk/test/tree.doctest
+++ b/nltk/test/tree.doctest
@@ -39,39 +39,6 @@ tree object to one of several standard tree encodings:
             [.dp [.d the ] [.np dog ] ]
             [.vp [.v chased ] [.dp [.d the ] [.np cat ] ] ] ]
 
-Trees can be initialized from treebank strings:
-
-    >>> tree2 = Tree('(S (NP I) (VP (V enjoyed) (NP my cookie)))')
-    >>> print(tree2)
-    (S (NP I) (VP (V enjoyed) (NP my cookie)))
-
-Trees can be compared for equality:
-
-    >>> tree == Tree(str(tree))
-    True
-    >>> tree2 == Tree(str(tree2))
-    True
-    >>> tree == tree2
-    False
-    >>> tree == Tree(str(tree2))
-    False
-    >>> tree2 == Tree(str(tree))
-    False
-
-    >>> tree != Tree(str(tree))
-    False
-    >>> tree2 != Tree(str(tree2))
-    False
-    >>> tree != tree2
-    True
-    >>> tree != Tree(str(tree2))
-    True
-    >>> tree2 != Tree(str(tree))
-    True
-
-    >>> tree < tree2 or tree > tree2
-    True
-
 Tree Parsing
 ============
 
@@ -1074,5 +1041,4 @@ This used to discard the ``(B b)`` subtree (fixed in svn 6270):
 
     >>> print(Tree.fromstring('((A a) (B b))'))
     ( (A a) (B b))
-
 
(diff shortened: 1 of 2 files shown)


nltk/test/tree.doctest
```diff
@@ -39,39 +39,6 @@ tree object to one of several standard tree encodings:
             [.dp [.d the ] [.np dog ] ]
             [.vp [.v chased ] [.dp [.d the ] [.np cat ] ] ] ]
 
-Trees can be initialized from treebank strings:
-
-    >>> tree2 = Tree('(S (NP I) (VP (V enjoyed) (NP my cookie)))')
-    >>> print(tree2)
-    (S (NP I) (VP (V enjoyed) (NP my cookie)))
-
-Trees can be compared for equality:
```
kmike, line 48: Why drop comparison tests?

nltk/tree.py
```diff
@@ -51,21 +51,10 @@ class Tree(list):
         (VP (V saw) (NP him))
         >>> print(s[1,1])
         (NP him)
-        >>> t = Tree("(S (NP I) (VP (V saw) (NP him)))")
-        >>> s == t
-        True
-        >>> t[1][1].set_label('X')
```
kmike, line 57: This test showed how to work with Tree labels and how to move subtrees, why remove it?

nltk/tree.py
```diff
@@ -85,28 +74,21 @@ class Tree(list):
     specifying ``tree[i]``; or a sequence *i1, i2, ..., iN*,
     specifying ``tree[i1][i2]...[iN]``.
 
-    Construct a new tree.  This constructor can be called in one
-    of two ways:
+    Construct a new tree.  This constructor can be called in the following way:
```
kmike: What do you think about referring to `fromstring` method here? Something like "You can also construct a tree from a string; use :meth:fromstring for that".

nltk/test/tree.doctest
SocialfiPanda, line 48: The issue #675 asks for dropping the string initialization of the Tree. So, these two still work:

Tree('dp', [Tree('d', ['the']), Tree('np', ['dog'])])
Tree.fromstring('(dp (d the) (np dog))')

But this not:
Tree('(dp (d the) (np dog))')

If I understand the issue #675 correctly, after the pull request, we aren't supposed to be able to initialize a Tree by supplying it a string (line 44 should break).

kmike, line 48: Right; can we replace `Tree('(dp (d the) (np dog))')` with `Tree.fromstring('(dp (d the) (np dog))')` here and keep the tests?

kmike, line 48: I'm not asking to keep tests in lines 44-46, the question was about lines 50-...

SocialfiPanda, line 48: can we replace Tree('(dp (d the) (np dog))') with Tree.fromstring('(dp (d the) (np dog))') here and keep the tests?
- Sure. I will do that.

nltk/tree.py
SocialfiPanda, line 57: Again, now we can't initialize a tree by a string. I can change replace 
replace Tree("(S (NP I) (VP (V saw) (NP him)))") with Tree.fromstring("(S (NP I) (VP (V saw) (NP him)))") and keep line 57 for setting the label. You think it's good?

nltk/tree.py
SocialfiPanda: I think it sounds good. I will add that to the docstring.

Also, I just found out there were many places in nltk/tree.py still using old string initilization syntax. e.g. line 212:

">>> t = Tree('(S (NP (D the) (N dog)) (VP (V chased) (NP (D the) (N cat))))')"

Sorry I forget to change these. I will replace them all with Tree.fromstring instead, and make a new commit addressing all your concerns above. Really appreciate your help!

SocialfiPanda:
These two last commits add back the test cases I deleted in my previous commits. Also, it changes:

Tree(str(tree)) to Tree.fromstring(str(tree)) in the docstring of the tree.py and in the test cases of test/tree.doctest.

PR #679, opened 2014-06-08
CONLL format missed columns

nournia:
CONLL [Data Format](http://nextens.uvt.nl/depparse-wiki/DataFormat) consists of 10 columns which some of them are missed in current DependencyGraph. Adding these fields makes read/write of conll file consistent with its standard.

Missed columns are `lemma`, `ctag` (coarse-grained POS) and `feats` (syntactic or morphological features), which are supported in current read and write functions.


stevenbird:
Thanks @nournia, much appreciated.

PR #683, opened 2014-06-23
Added `parse` method to malt.py, fixes #682

### LINKED ISSUE #682: MaltParser not completely updated to new API
Because of API changes, MaltParser objects fail when raw_parse() is called, since MaltParser.raw_parse internally calls parse() with two arguments (in addition to self).

Error occurs in malt.py, line 127, though other issues may exist.

heatherleaf:
A quick fix to #682. Found an error in `.to_conll()` in dependencygraph.py too.

However, this does not make the Maltparser adhere to the API:
- `.parse()` returns a DependencyGraph, not an iterator of DependencyGraph
- similar for `.parse_sents()`, it returns a list of DependencyGraph, not an iterator of iterators
- `.parse_all()` and `.parse_one()` will raise an error

Stanford suffers from similar issues.

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -284,7 +284,7 @@ class DependencyGraph(object):
             elif style == 4:
                 lines.append('%s\t%s\t%s\t%s\n' % (word, tag, head, rel))
             elif style == 10:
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks for the fixes, and for the reminder about the existing API inconsistency.

PR #687, opened 2014-06-27
Portability / resource mangement

robwahl:
indian.py got rid of codecs to make it more portable
megam.py got rid of import os.path and import os, not needed and it increases stability without it
cleaned up imports for pt.py to make code more portable and more efficient
simplified code in relextract.py import nltk.sem import relextract is recusive
string_category.py removed import os and general from nltk import compat to make it run with less resources

Where is nose being defined?
nltk/setup.py What setup tools that are to be imported?
Wordfreq.py where is pylab coming from?


nltk/corpus/reader/indian.py
```diff
@@ -1,9 +1,12 @@
+# got rid of codecs because codecs not needed
```
kmike, line 1: Could you please remove this comment? It is not needed to understand the source code, and it makes `coding=utf8` no-op.

nltk/classify/megam.py
```diff
@@ -1,8 +1,9 @@
+# coding=utf-8
 # Natural Language Toolkit: Interface to Megam Classifier
 #
-# Copyright (C) 2001-2013 NLTK Project
-# Author: Edward Loper <edloper@gradient.cis.upenn.edu>
-# URL: <http://www.nltk.org/>
+# Copyright (C) 2001-2014 NLTK Project
+# Author: Edward Loper <edloper@gmail.com>
+# URL: <http://nltk.org/>
```
kmike, line 6: I think we should keep it as www.nltk.org because nltk.org is a redirect to www.nltk.org. We can't have nltk.org as a "main" domain because of some DNS / CDN issues.

Is @edloper OK with email changes?

nltk/classify/megam.py
```diff
@@ -135,7 +136,7 @@ def parse_megam_weights(s, features_count, explicit=True):
     for line in lines:
         if line.strip():
             fid, weight = line.split()
-            weights[int(fid)] = float(weight)
+            weights[int(fid)] = float()
```
kmike, line 139: why remove "weight"?

nltk/sem/relextract.py
```diff
@@ -86,17 +92,18 @@ def descape_entity(m, defs=htmlentitydefs.entitydefs):
 
 
     """
-    #s = 'mcglashan_&amp;_sarrail'
-    #l = ['mcglashan', '&amp;', 'sarrail']
-    #pattern = re.compile("&(\w+?);")
-    #new = list2sym(l)
-    #s = pattern.sub(descape_entity, s)
-    #print s, new
+    # s = 'mcglashan_&amp;_sarrail'
```
kmike, line 95: There is a common pattern: comments without a space between '#' and a text are commented out code, not real comments.

nltk/sem/relextract.py
```diff
@@ -112,9 +119,10 @@ def list2sym(lst):
     sym = sym.replace('.', '')
     return sym
 
-def mk_pairs(tree):
+
+def tree2semi_rel(tree):
```
kmike, line 123: could you please move API changes to another PR?

stevenbird:
@robwahl do you have any responses to the feedback?

stevenbird:
Adopted suggestion to get rid of codecs in indian.py in https://github.com/nltk/nltk/commit/31a927e1e21f0038b0e299903ef9b6dbc1969e7e

stevenbird:
Adopted suggestion to drop import of os in string_category.py in https://github.com/nltk/nltk/commit/173f32929ff22e6f42d03e6a73d1fac86e95f20d

stevenbird:
`pt.py` is dropped in https://github.com/nltk/nltk/commit/7c2a65da4d2b97d99f2a11384d00fac00cd6d598

stevenbird:
There are some useful cleanups here, but there's too many issues with this PR to accept it. I would be grateful for a separate PR which is just fixing the whitespace issues.

PR #697, opened 2014-07-09
(unfinished) updated install docs: Python 3.4, virtualenvs, wheels.

kmike:
Todo:
- [ ] check that Win 64 install instructions work;
- [ ] add Win 32 install instructions?
- [ ] finish Ubuntu install instaructions;
- [ ] add Debian install instructions?
- [ ] check that everyone is OK with recommending Python 3.4 + virtualenvs

### DIFF
--- a/web/install.rst
+++ b/web/install.rst
@@ -1,39 +1,137 @@
 Installing NLTK
 ===============
 
-NLTK requires Python versions 2.6-2.7.
-(A version supporting Python 3 is available at http://nltk.org/nltk3-alpha/).
+NLTK is a Python package. It supports Python 2.6, 2.7 or 3.2+.
 
-Mac/Unix
---------
+If you don't have Python installed you should install it first.
+We recommend using latest Python (3.4.1 at the time of writing).
 
-#. Install Setuptools: http://pypi.python.org/pypi/setuptools
-#. Install Pip: run ``sudo easy_install pip``
-#. Install Numpy (optional): run ``sudo pip install -U numpy``
-#. Install NLTK: run ``sudo pip install -U nltk``
-#. Test installation: run ``python`` then type ``import nltk``
+Below is a description of how to install NLTK with Python 3.4.
+There are other ways to install Python and get NLTK working with
+it (e.g. one can use Anaconda_, or install NLTK without virtualenv,
+or use Python 2.x); please check Wiki_ for other installation methods.
 
-Windows
--------
+Preparing the Environment
+-------------------------
 
-These instructions assume that you do not already have Python installed on your machine.
-If you do, you can skip to the final step and just install NLTK.
+When the environment is ready, NLTK should install the same regardless of
+an operating system. How to prepare the environment depends on OS.
 
-32-bit binary installation
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Windows 64 bit
+~~~~~~~~~~~~~~
 
-#. Install Python: http://www.python.org/download/releases/2.7.3/ (avoid the 64-bit versions)
-#. Install Numpy (optional): http://sourceforge.net/projects/numpy/files/NumPy/1.6.2/numpy-1.6.2-win32-superpack-python2.7.exe
-#. Install NLTK: http://pypi.python.org/pypi/nltk
-#. Test installation: ``Start>Python27``, then type ``import nltk``
+TODO: check that it works
 
-Source installation (for 32-bit or 64-bit Windows)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#. Download and install latest minor release of Python 3.4 from
+   https://www.python.org.
 
-#. Install Python: http://www.python.org/download/releases/2.7.3/
-#. Install Numpy (optional): http://www.lfd.uci.edu/~gohlke/pythonlibs/#numpy
-#. Install Setuptools: http://pypi.python.org/packages/2.7/s/setuptools/setuptools-0.6c11.win32-py2.7.exe
-#. Install Pip: ``Start>Run... c:\Python27\Scripts\easy_install pip``
-#. Install NLTK: ``Start>Run... c:\Python27\Scripts\pip install nltk``
-#. Test installation: ``Start>All Programs>Python27>IDLE``, then type ``import nltk``
+#. Check that the Python is installed correctly: open the terminal and run
+   ``python3.4`` - it should open the Python shell. Press Ctrl-D to close the
+   Python shell.
 
+#. It is a good practice to use a `virtual environment`_. Create and activate
+   it using the following commands::
+
+       pyvenv my-nltk-env
+       my-nltk-env/Scripts/activate.bat
+
+#. Install numpy. Some of the NLTK modules can be used without numpy, but
+   many of them require numpy; you'll need numpy to follow NLTK book.
+
+   ::
+
+       pip install -f https://nipy.bic.berkeley.edu/scipy_installers numpy
+
+   .. note::
+
+        Installing Numpy can be hard on Windows, especially if a virtual
+        environment is used. For 64 bit Windows there are binary wheels_
+        available at https://nipy.bic.berkeley.edu/scipy_installers - they
+        should work without a build environment.
+
+        If for some reason this doesn't work you may consider using Anaconda_.
+
+
+OS X
+~~~~
+
+#. Download and install latest minor release of Python 3.4 from
+   https://www.python.org. Python 3 installed using http://brew.sh/ is
+   also fine.
+
+#. Check that the Python is installed correctly: open the terminal and run
+   ``python3.4`` - it should open the Python shell. Press Ctrl-D to close the
+   Python shell.
+
+#. It is a good practice to use a `virtual environment`_. Create and activate
+   it using the following commands::
+
+       pyvenv ./my-nltk-env
+       source ./my-nltk-env/bin/activate
+
+#. Install numpy. Some of the NLTK modules can be used without numpy, but
+   many of them require numpy; you'll need numpy to follow NLTK book.
+
+   ::
+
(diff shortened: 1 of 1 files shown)


stevenxxiu:
On the windows install of python, `python3.4` does not open the python shell, `python` does however.

kmike:
Thanks @jskda. Do you know what happens if several Pythons from python.org are installed? Is it the latest one which is available via "python" name?

stevenbird:
Thanks for your work @kmike. I agree that virtualenv offers a very clean solution, but I think it is going to be quite difficult for novice users. We can tell them what commands to put into a command window, but they're likely to have problems downstream. E.g. they might forget how to use virtualenvs then post to our list to say NLTK isn't installed; they might have installed other packages with their system Python and hope to use them together with NLTK. I think virtualenv might be the solution when someone has conflicting Python installations and wants to start afresh.

What do you think about maximally simple instructions for Mac OSX and Windows, assuming a vanilla existing setup on the machine, and then describe virtualenv later in the instructions, in case of problems?

pquentin:
I agree with @stevenbird. What Python projects usually say is "pip install nltk" - people who know virtual environments will use one if they want, and those who don't will find the installation instructions to be very complicated. This page is still very useful, but I don't think it should be the first one to be shown to users willing to use nltk.

stevenxxiu:
@kmike I just did a clean install on windows 7, there is the option of adding python to the path in the installer, which is disabled by default. Otherwise whichever is installed last is inserted to the start of the `path` environment variable I think. It might be good to tell people to add the latest python to path if they haven't done so already.

I agree with @stevenbird on the virtual env issue too, virtualenv is probably best for people who want to use it.

kmike:
It seems the consensus is that virtualenvs are hard for novice users. Here is my attempt at convincing you they are not :)
- On OS X and on most Linux distributions users are very likely to have Python 2.x installed and used by default. This means that when they run `python` or `pip` they'll run Python 2.x. This also means that the software they'll install will most likely also be installed for Python 2.x. This will cause troubles if we recommend Python 3.x.
- Recommending Python 3.x is good because working with text _correctly_ is easier / more natural there.
- Users are very likely to have outdated software on their machines. Older versions of pip can't install wheels, so `pip install numpy` or `pip install scipy` or `pip install matplotlib` will fail for most users if executed outside a newly created virtualenv.
- Users may not have pip and setuptools installed at all. It is a common gotcha: pip/setuptools are Python packages, and to install Python packages you need pip/setuptools - what to do if they are not installed? In Python 3.4 pyenv makes them available by default.
- Installing software without virtualenv often requires sudo rights. So users run `pip install foo`, it fails with a permission error, they fix it by prefixing the command with sudo: `sudo pip install foo`. In case of virtualenv they fix it by activating the virtualenv. This is almost the same effort, but the advantage is that it doesn't require sudo right, it can't break system software that depends on particular package versions, and that it needs to be done only once, not before each command.

It is one datapoint, but I myself don't recall any problems using virtualenv when I started with it. If you are inside virtualenv then the command-line prompt becomes `(envname) >` instead of `>`, so it is easy to tell if the virtualenv is active or not. If the users follow the proposed install doc the prompt will be `(my-nltk-env) >` - NLTK will work if there is 'nltk' in command prompt and won't work if it is not; that's quite easy to remember.

IDEs may cause problems - you need to configure them to use the correct interpreter. But it is always a problem to solve if there is more than one Python interpreter installed. If we recommend Python 3.x then it will often be the case anyways (Python 2.x and Python 3.x). I'm using PyCharm; its free&open source Community Edition has good virtualenv support.

For ipython notebooks virtualenvs are not a problem - you activate the virtualenv and start IPython notebook.

Also, with virtualenv there is a single way to install packages - use pip. Without virtualenvs there are other options in addition to pip: .exe installers, .dmg installers, yum, apt, macports, etc.; each one have its own specifics, can fail in its own way and may not play well with other methods.  

I tried to search for "install" in nltk-users google groups. I think that at least the following users shouldn't have had problems if they use virtualenvs:
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/MWK-Wy0T5ac/iDSU0USZvlEJ (just activate the proper virtualenv as described in instructions)
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/YR7KDHQ-E0s/xyA7V3v6_1EJ (setuptools is already installed in virtualenv)
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/n1s6kUbIbQ4/CTyCYIWFAJ0J (user have 2 versions of Python installed at the same time, using virtualenv ensures the version you want is used);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/srI2j_MTeW8/slOgtzggXvAJ (the same as previous, multiple versions of Python, hard to tell which version runs when `python` is executed - explicitly activating the virtualenv solves it);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/yiYXoTRyrbM/nQGMB6py_EUJ (AFAIK with recent pip recent matplotlib should install from wheel out of box without a need to configure build environment; if it doesn't work it should be fixed in a near future);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/-UGjUbfFBfQ/4fKDJ_DzSy0J (install instructions are reasonable, but many steps are unnecessary now);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/5UAA44eaHL4/OF4Ek6FoJuIJ (user upgraded the computer, and NLTK installed in system location became broken, most likely because the system Python changed after upgrade. With virtualenv and Python installed from python.org you don't rely on system Python);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/sNuc1vBIfUc/_SFPwB5V3p4J (no need to worry about setuptools installation, it is available by default in pyvenv);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/cO8nPUOUpj8/Ods1mufNovkJ (two versions of Python in parallel, hard to tell which version runs when `python` is executed - explicitly activating the virtualenv solves it);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/CnSLjDbQ2X0/II--RIh4gfIJ (with recent pip numpy should install from wheel);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/3EMhzarAIag/mkLgcVk_KlMJ (multiple Python versions again);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/c3SlWkXXZj4/B10wKEsfAjMJ (multiple NLTK versions installed - one using apt-get and one NLTK using setuptools, hard to tell which one is used - virtualenv makes this explicit);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/LwvzZFurEfY/enrzuW30BmIJ (multiple NLTK versions; with virtualenv it is possible to keep old projects using old NLTKs if needed);
- https://groups.google.com/forum/#!searchin/nltk-users/install/nltk-users/v49WUzm5bOU/tCsoi3JEcmYJ ("pip install numpy" should work in a newly created virtualenv);
- https://groups.google.com/forum/#!searchin/nltk-users/install|sort:date/nltk-users/HbyhWIXDud0/MeT6_UgC26kJ (with pip & virtualenv instead of .exe installers there shouldn't be such problems)
- https://groups.google.com/forum/#!searchin/nltk-users/install|sort:date/nltk-users/Cdcwd6C9uy8/9am3Aw97ZtwJ (multiple Python versions after the upgrade);
- https://groups.google.com/forum/#!searchin/nltk-users/install|sort:date/nltk-users/qDX7dOeyZGo/IToVqhOq_JsJ (wrong architecture; with the instructions in this PR it should fail earlier, when the Python was installed, and there is no way to choose wrong architecture for Python packages like setuptools).

stevenbird:
I don't mind including instructions for virtualenvs, but I'm not convinced this should be the _only_ recommended installation. My preference is to recommend rather than dictate.

Note that there will be people who already have packages installed in their default Python instance, and this approach will require them to recreate their Python environment. They may prefer to use a single package manager (like apt or macports) wherever possible. Also, a quick search of other packages (scipy, matplotlib, ...) shows that they are not advocating virtualenvs as their only installation method.

What do you think about some lightweight instructions (such as what I already have there) with the obvious health warnings about conflicts and getting an older version of Python, and then your virtualenvs instructions for anyone who wants to set it up in the recommended way? (And the way we will endeavour to support.)

web/install.rst
```diff
@@ -1,39 +1,137 @@
...
+   .. note::
+
+        Numpy started to provide binary wheels_ for OS X; this means
+        'pip install numpy' should now work fast and without a build
+        environment.
+
+Ubuntu
+~~~~~~
+
+#. TODO: install Python 3.4 from https://launchpad.net/~fkrull/+archive/ubuntu/deadsnakes
```
dimazest, line 88: Maybe, it's better to use default python. For 12.04 it could be python3 from apt-get (given that ubuntu provides numpy, scipy and scikit-learn for it) or python2.

For newer ubuntus it should be python 3.

I would suggest using apt-get as much as possible and don't rely on virtualnv (as it requires knowledge about python deployment)

web/install.rst
```diff
@@ -1,39 +1,137 @@
...
+
+Installing NLTK
+---------------
+
+When the environment is ready, NLTK should install the same regardless of
+an operating system.
+
+#. Install NLTK::
+
+       pip install nltk
```
dimazest, line 117: `pip install --user nltk` might be a better idea because it doesn't require root access. Sometimes on a machine with a limited access one can ask to install system packages (`apt-get install numpy pip`), getting custom software installed is much harder.

though i'm confused now. I think it's better to remind the reader that a virtualenv should be activated.

stevenbird:
I'm closing this stale PR, much as it would be nice to have more detailed installation instructions for NLTK 3. I am not aware of any recent complaints about the [simple installation instructions](http://www.nltk.org/install.html).

PR #705, opened 2014-07-16
Continuous integration

dimazest:
This is an attempt to:
- make the test run green over all supported Python versions: 2.6, 2.7, 3.2, 3.3, 3,4
- configure Travis CI to run the tests https://travis-ci.org/dimazest/nltk
- configure coveralls.io to monitor test coverage https://coveralls.io/r/dimazest/nltk
- installation of Numpy and friends using wheels to speed up installation

As a side effect I configured vagrant to have Travis like environment. Ideally, that virtualmachine should be enough to play with NLTK: all dependencies, parsers, provers, taggers should be there :).

So far it's a proof of concept. There are quite a lot messy commits. I wanted to squash them into bigger, but git is not the friendliest piece of software.

Things to be done:
- deploy hunpos model
- fix different set representation acros Python versions set([1, 2, 3]) versus {1, 2, 3}
- update readme
- decide where to host custom wheel files for numpy and the rest.


stevenbird:
Thanks @dimazest! This looks good to me. @kmike or @xim do you have any feedback?

dimazest:
There are still some tests failing, i want to finish them before Europython. btw, is anyone going there?

kmike:
:+1: 
1. About Salt - it is a non-trivial piece of software, and it looks like simple apt-get / a shell script can cover all its uses here, and it will have similar readability. What do you think about removing Salt dependency?
2. As for wheels - on Linux wheels must match the distribution pretty closely. How are you building these wheels, and where will they work?

dimazest:
I'll try to answer all the points at one. Indeed, wheels need similar environment. To achieve this I use vagrant to spin off an Ubuntu 12.04 box, which runs salt to install all the software. So the eggs are  intended to be used only by travis or vagrant.

I'm afraid that a bash script will become messy, though I can try to replace it.

stevenbird:
@kmike or @xim do you have any further feedback?

stevenbird:
In the absence of input I'm afraid I need to close this. Please re-open if you want to keep working on it.

PR #706, opened 2014-07-17
BLUE score refactoring and more tests.

dimazest:
BLUE score is now stable across different python versions. Some functions are refactored to make them more readable.

More tests are added based on examples from Papineni et al.

### DIFF
--- a/nltk/align/bleu.py
+++ b/nltk/align/bleu.py
@@ -31,7 +31,7 @@ class BLEU(object):
 
     Part 2 - brevity penalty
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @dimazest this is great!

dimazest:
I've forgotten to mention, but I had to change the returned score in the tests. But I didn't calculate it by hand, nor got if from a paper.

stevenbird:
This happens all the time.

dimazest:
I mean, the returned score might be wrong :) but i'm not a MT expert.

PR #711, opened 2014-07-26
Added GDFA Symmetrization for nltk.align

alvations:
I have added a `grow-diag-final-and` symmetrization algorithm to retrieve symmetrized word alignments from both source-to-target and target-to-source word alignment (from http://www.statmt.org/moses/?n=FactoredTraining.AlignWords)

There might be some naming convention issues, since sometimes its call `grow-diag-final` without the `and`. Possibly `and` refers to the inclusion of original alignment points in the step 3 algorithm but maybe some people think the `final` subsumes the `and`.


stevenbird:
Thanks @alvations!

PR #712, opened 2014-07-27
WordNet paths

larsmans:
Since path computations in WordNet are currently the bottleneck in one of my applications, I decided to optimize them. This patch takes an average 44% off the running time of `Synset.shortest_path_distance` in my benchmarks.


stevenbird:
Thanks for converting the n² to nlogn.

I wonder about exposing the `sort=True` parameter at the next level up to avoid introducing a de facto convention `method()` (sorted) and `_method()` (unsorted).

PR #724, opened 2014-08-05
Added Phrase Extraction Algorithm for Phrase-based MT in nltk.align

alvations:
Added the phrase extraction algorithm from Philip Koehn (2009) Statistical Machine Translation book. 
Phrase extraction algorithm found on page 133. Also note that the original book has an error in the algorithm, see errata http://www.statmt.org/book/errata.html.

### DIFF
--- a/nltk/align/gdfa.py
+++ b/nltk/align/gdfa.py
@@ -23,7 +23,7 @@ def grow_diag_final_and(srclen, trglen, e2f, f2e):
             
     Step 3 - Adds all other alignment points thats not in the intersection,
              not in the neighboring alignments that met the criteria but in the 
-             original forward/backward alignment outputs.
+             original foward/backward alignment outputs.
     
     [in]:
     *e2f* is the forward word alignment outputs from source-to-target language.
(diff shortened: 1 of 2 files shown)


alvations:
Possibly, the `phrase_based.py` needs to also have a module that generates the phrase tables with the probabilities but for now, the `phrase_extraction()` is usable for people who wants to extract phrases and use other ways to normalize/calculate the phrase probabilities.

stevenbird:
@alvations thanks for the contribution -- would you please take a look at https://github.com/nltk/nltk/wiki/Developers-Guide and follow our coding practices?

alvations:
@stevenbird , really sorry about the bare commenting of the code. I forgot to push the with the proper doctest and comments. i'll push the code up in a min.

alvations:
about some of the weird and not so useful comment. It's because that's what shown in the algorithm in Koehn's SMT book. It might be useful for readers/learners to see the corresponding lines in the algorithm. http://i.stack.imgur.com/0j556.png.

stevenbird:
I agree that it would be good to retain Koehn's original comments.

stevenbird:
Would you also please update your code to work with Python 3 (use 2to3).
Can you avoid nested functions? I don't understand why you've put `extract()` inside `phrase_extraction()`.

alvations:
I guess the spaghetti functions are a little ugly but it's to avoid over-verbosity. It also shields users from using the wrong function in `phrase_based.py` and less parameters transfer from the `phrase_extraction()` to `extract()`. I've untangled the nested functions.

stevenbird:
This looks good now, thanks.

PR #727, opened 2014-08-07
handling of inline xml format and xml format

RahulKulhari:
json response of output


stevenbird:
This PR includes seems to be based on an outdated version of `stanford.py`. E.g. it changes copyright year range and project URL. @RahulKulhari would you please submit a PR based on the current version?

stevenbird:
I would like to include this work, but unfortunately cannot while it reverts other changes made to the code. Please submit a new PR based on the current version.

PR #733, opened 2014-08-15
Updated Paice documentation (Sphinx-style) & improved demo.

lade:
Parameters for paice.py documented Sphinx-style; demo now uses more non-trivial example.


stevenbird:
Great, thanks @lade

PR #734, opened 2014-08-15
Fixed a problem with jar finder

capdevc:
There's a control flow problem here. As is, if you specify path_to_jar, even if it's correct, the next iteration will raise the lookup exception.


stevenbird:
Well spotted, thanks.

PR #745, opened 2014-09-18
handling of inline xml format and xml format

RahulKulhari:


stevenbird:
Sorry, I can't accept this PR because it is based on an outdated revision of the code. Please open a new issue to document the problem you want to fix, then submit a PR using the current revision, thanks.

PR #748, opened 2014-10-01
Removing left-over PunktWordTokenizer import

pquentin:
Tests failed early because of this.


pquentin:
For what it's worth, `import nltk` also fails without this patch.

stevenbird:
Oh dear! Thanks.

PR #749, opened 2014-10-01
Remove remaining googlecode URLs

pquentin:
A few URLs from Google Code were remaining. I removed those. This patch is not perfect as I:
- could not find an alternative for the ccg.pdf link
- didn't update nltk/corpus/**init**.py docstring but changed xx to TODO for better visbility.

It's still a step in the right direction as nltk.googlecode.com is no longer accessible.


stevenbird:
Thanks.

PR #763, opened 2014-10-10
Avoid imports of tkinter in nltk.sem if tkinter is not available.

dimazest:
This is my second try of fixing NLTK tests. Instead of delivering a big pull request that tries to fix tests and configure CI, I decided to submit many small patches.


stevenbird:
Looks good, thanks.

PR #765, opened 2014-10-11
Make nltk.wsd.lesk() deterministic.

dimazest:


stevenbird:
I think it's a bug if the Lesk algorithm is not deterministic. We've had to do some work on other WordNet methods to ensure that they are deterministic (e.g. with sorting results).

dimazest:
Ok. I'll make it deterministic.

nltk/wsd.py
```diff
@@ -12,53 +12,61 @@
...
-        >>> lesk(sent, word, pos)
+        >>> lesk(sent, 'bank', 'n')
         Synset('bank.n.07')
-    
+
     :param context_sentence: The context sentence where the ambiguous word occurs.
     :param ambiguous_word: The ambiguous word that requires WSD.
     :param pos: A specified Part-of-Speech (POS).
-    :param dictionary: A list of words that 'signifies' the ambiguous word.
+    :param dict dictionary: A mapping of synsets to their definitions.
```
dimazest: I'm not sure whether it makes sense for dictionary to be of type dict, but i've left it to keep compatibility.

dimazest:
Hi,

I've cleaned up lesk() implementation, so all the tests pass now!!!

nltk/wsd.py
stevenbird: I'd prefer to break compatibility in favour of correctness, since NLTK 3 is not widely adopted yet. So what about a better choice of name and type for this parameter?

dimazest: i'll have a look to the paper to see how things are named there.

stevenbird: OK, thanks. Similarity with the published algorithm helps us read and maintain the code.

dimazest: I agree. Keeping code as close to a paper could be a good NLTK coding guide. Also, implementing examples in a paper as tests might be a good idea.

nltk/wsd.py
```diff
@@ -6,59 +6,41 @@
...
-    best_sense = _compare_overlaps_greedy(context_sentence,
-                                       dictionary, pos)
-    return best_sense
+    context = set(context_sentence)
+    if not synsets:
+        synsets = wordnet.synsets(ambiguous_word)
+
+    _, sense = max(
+        (len(context.intersection(ss.definition().split())), ss)
+        for ss in synsets if pos is None or str(ss.pos()) == pos
```
dimazest, line 40: I use `if pos is None *or* ...` so when no pos tag is passed all possible senses are evaluated.

PR #766, opened 2014-10-11
skip ccg doctests as they are broken.

dimazest:


stevenbird:
Fair enough. I wish I understood why they broke – the output looks the same as expected, but I guess it is not identical.

PR #770, opened 2014-10-14
Remove emacs directory

edran:
- The packages are **extremely** old and likely not to be compatible with the 24+ release (which was released in June 2012). 
- emacs 24 includes package.el and ELPA, which makes the operation of getting emacs extension pretty much trivial and painless. 
- python.el is now shipped with GNU emacs by default as opposed to python-mode.

I suggest we write a small development setup guide if really needed.


stevenbird:
Fair enough! Thanks.

edran:
No problem :)

PR #777, opened 2014-10-26
_PY3_DATA_UPDATES uses win path if on win

owenhaynes:
py3_data function does not work on windows as the item would never match a value in a windows path.

_PY3_DATA_UPDATES now has a windows version if windows is used

This is only a quick fix as having the list of theses data directories is not perfect.

### DIFF
--- a/nltk/compat.py
+++ b/nltk/compat.py
@@ -358,10 +358,19 @@ except ImportError: # python 2.6
 
 # The following datasets have a /PY3 subdirectory containing
 # a full copy of the data which has been re-encoded or repickled.
-_PY3_DATA_UPDATES = ["chunkers/maxent_ne_chunker",
-                     "help/tagsets",
-                     "taggers/maxent_treebank_pos_tagger",
-                     "tokenizers/punkt"]
+
+_PY3_DATA_UPDATES = []
+
+if sys.platform.startswith('win'):
+    _PY3_DATA_UPDATES = ["chunkers\maxent_ne_chunker",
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for this. While it's not the final solution it's a helpful step in that direction.

nltk/compat.py
```diff
@@ -358,10 +358,19 @@ def total_ordering(cls):
...
+if sys.platform.startswith('win'):
+    _PY3_DATA_UPDATES = ["chunkers\maxent_ne_chunker",
+                         "help\tagsets",
+                         "taggers\maxent_treebank_pos_tagger",
+                         "tokenizers\punkt"]
+else:
+    _PY3_DATA_UPDATES = ["chunkers/maxent_ne_chunker",
+                        "help/tagsets",
+                        "taggers/maxent_treebank_pos_tagger",
+                        "tokenizers/punkt"]
```
dimazest, line 373: os.path.join() should join fragments correctly, taking care of '/' and '\'.

stevenbird, line 373: Yes, that would be much cleaner, if you don't mind submitting a patch!

dimazest:
could you reopen the issue and assign it to me, so it doesn't get lost.

stevenbird:
@dimazest I've added you as a developer; once you accept, please assign #783 to yourself and submit a PR when you're ready, thanks!

dimazest:
cool! thanks

PR #780, opened 2014-10-31
Call gs directly in nltk.tree._repr_png_, no shell

kowey:
shell=True causes ghostscript to stick around with a prompt on
MacOS X (and apparently Linux too, as reported by another user)

```
GPL Ghostscript 9.06 (2012-08-08)
Copyright (C) 2012 Artifex Software, Inc.  All rights reserved.
This software comes with NO WARRANTY: see the file PUBLIC for details.
GS>
```

Note that the shell=True was deliberately introduced in 0cc3d8a
to avoid a console window from popping up, possibly
related to Windows? If so, this patch may undo that change, but
maybe we can have the best of both worlds by setting the
STARTUPINFO.wShowWindow flag instead.

http://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess

### DIFF
--- a/nltk/tree.py
+++ b/nltk/tree.py
@@ -715,8 +715,7 @@ class Tree(list):
             _canvas_frame.destroy_widget(widget)
             subprocess.call([find_binary('gs', binary_names=['gswin32c.exe', 'gswin64c.exe'], env_vars=['PATH'], verbose=False)] +
                             '-q -dEPSCrop -sDEVICE=png16m -r90 -dTextAlphaBits=4 -dGraphicsAlphaBits=4 -dSAFER -dBATCH -dNOPAUSE -sOutputFile={0:} {1:}'
-                            .format(out_path, in_path).split(),
-                            shell=True)
+                            .format(out_path, in_path).split())
             with open(out_path, 'rb') as sr:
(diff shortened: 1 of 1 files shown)


stevenbird:
@kowey, @moreymat – would one of you please add brief instructions for an MSWindows IPython user to test this change? What do they need to do to cause this repr method to be invoked?

kowey:
Of course, sorry for not thinking to include this.  Run a notebook with this cell.  You should see a graphical a(b c) tree that looks [like this](http://i.imgur.com/k14twDJ.png)

``` python
from nltk.tree import Tree
from IPython.display import display

tree = Tree('a', [Tree('b', []), Tree('c', [])])
display(tree)
```

On Windows, Mac, and Linux, what I expect to happen is that we see the tree, but a slightly annoying console window open along the side.  For Windows, the presence of the console window can be considered a regression from 0cc3d8a; but for Mac/Linux, annoying console window is a small price to pay to get graphics working in the first place.

stevenbird:
Well, there's been no complaints (https://groups.google.com/forum/#!topic/nltk-dev/FcCku24RvBk), so I'll go ahead and merge.

PR #794, opened 2014-11-12
ASCII visualization of trees

heatherleaf:
ASCII visualization of trees, modified from the disco-dop project by Andreas van Cranenburgh, https://github.com/andreasvc/disco-dop

```
>>> tree = nltk.Tree.fromstring('(S (NP Mary) (VP walks))')
>>> tree.visualize()
      S       
  ____|____    
 NP        VP 
 |         |   
Mary     walks
```

The module `treevisualize` contains one class `TreeVisualizer` with two methods `text()` and `svg()`. I also added a convenience method `visualize()` to the standard `Tree` class.

Some questions before merging this pull request:
- Is the module name `treevisualize` fine? And the class name `TreeVisualizer`?
  - van Cranenburgh's original module name is `treedraw` and class `DrawTree`, but I thought they interfered too much with the module `nltk.draw.tree`
- I added a the method `visualize()` to `Tree`, is that a good name? 
  - the method prints the tree on stdout, should it return a string instead?

### DIFF
--- a/nltk/test/tree.doctest
+++ b/nltk/test/tree.doctest
@@ -41,7 +41,7 @@ tree object to one of several standard tree encodings:
 
 There is also a fancy ASCII art representation:
 
-    >>> tree.visualize()
+    >>> tree.pretty_print()
                   s               
           ________|_____           
          |              vp        
@@ -52,7 +52,7 @@ There is also a fancy ASCII art representation:
      |       |    |     |       |  
     the     dog chased the     cat
 
-    >>> tree.visualize(unicodelines=True, nodedist=4)
+    >>> tree.pretty_print(unicodelines=True, nodedist=4)
                            s                        
             ┌──────────────┴────────┐                   
             │                       vp              
--- a/nltk/test/tree.doctest
+++ b/nltk/test/tree.doctest
@@ -39,6 +39,30 @@ tree object to one of several standard tree encodings:
             [.dp [.d the ] [.np dog ] ]
             [.vp [.v chased ] [.dp [.d the ] [.np cat ] ] ] ]
 
+There is also a fancy ASCII art representation:
+
+    >>> tree.visualize()
+                  s               
+          ________|_____           
+         |              vp        
+         |         _____|___       
+         dp       |         dp    
+      ___|___     |      ___|___   
+     d       np   v     d       np
+     |       |    |     |       |  
+    the     dog chased the     cat
+
+    >>> tree.visualize(unicodelines=True, nodedist=4)
+                           s                        
+            ┌──────────────┴────────┐                   
+            │                       vp              
+            │              ┌────────┴──────┐            
+            dp             │               dp       
+     ┌──────┴──────┐       │        ┌──────┴──────┐     
+     d             np      v        d             np
+     │             │       │        │             │     
+    the           dog    chased    the           cat
+
 Trees can be initialized from treebank strings:
 
     >>> tree2 = Tree.fromstring('(S (NP I) (VP (V enjoyed) (NP my cookie)))')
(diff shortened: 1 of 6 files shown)


stevenbird:
Thanks for this welcome contribution @heatherleaf.

The name `visualize` is in addition to the existing `draw` and `pprint` which are available for trees. And we also have some other classes with visualisation methods, e.g. `FreqDist.plot()`, `FreqDist.pprint()`, and `FreqDist.tabulate()`, `sem.drt.DrtExpression.pretty()`, `parse.chart.Chart.pp()`.

I think the new functionality is closest to "pretty printing". The term "visualize" is probably too generic. Python uses the name `pprint` for pretty printing, but this output is supposed to be [in a form which can be used as input to the interpreter](https://docs.python.org/3/library/pprint.html).

What do you think about `prettyprint` or `pretty_print` or `pretty`? We could then standardise on this across the library.

timClicks:
`pretty_print` would be best across the whole of NLTK IMO. It's consistent with the library's philosophy of being accessible to newcomers while being fine for seasoned pros.

jnothman:
Consider providing a _repr_svg_() method (or similar) on the tree itself to
present a tree in an IPython GUI.

alvations:
From other related libraries: 
- `lxml` : `pretty_print` as parameter (from http://lxml.de/tutorial.html)
- `BeautifulSoup`: `prettify()` (from http://www.crummy.com/software/BeautifulSoup/bs4/doc/)
- `pprint`: `pprint()` (from https://docs.python.org/3.2/library/pprint.html)

Base on personal preference and my 2cents worth, I would also go for `pretty_print()`.

dimazest:
I'm also prefer `pretty_print()` because it's the clearest option.

`prettify` suggests that a tree is changed in some way. On the other side, it might be useful to split `pretty_print` to two methods: one that returns a pretty string, and other, that prints it (well, this is what pirnt is for).

stevenbird:
Thanks for the input. Let's go with `pretty_print` then. I suggest that the internal names that users typically won't see, like the module and class name, could keep van Cranenburgh's names. They're not perfect, but replacing "draw" with "visualize" doesn't do much in my opinion. Any others have thoughts on this?

heatherleaf:
Thanks for the input. I've updated the files, like this:
- renamed the `Tree.visualise()` method to `Tree.pretty_print()`
- renamed the class `TreeVisualizer` to `TreePrettyPrinter` 
- renamed the file `treevisualizer.py` to `treeprettyprinter.py`

If you want to keep the previous classname and/or filename, I guess you can skip some of the commits when merging the pull request.

heatherleaf:
@jnothman, there is already a method `Tree._repr_png_()`, and I don't know which one IPython selects if there is another "conflicting" method, `_repr_svg_()`. 

Also, I don't have IPython, so I can't test it. But perhaps someone else (@jnothman?) can try it out and issue another pull request? I guess something like this would do:

```
def _repr_svg_(self):
    from nltk.treeprettyprinter import TreePrettyPrinter
    return TreePrettyPrinter(self).svg()
```

stevenbird:
Thanks @heatherleaf.

jnothman:
Yes, I'm a little behind the times so you're right there's PNG. I can't see
documentation on the precedence ordering, but svg trumps png on my instance.

andreasvc:
Cool to see this merged.

Here's an example of the SVG output compared to ASCII art: http://nbviewer.ipython.org/gist/andreasvc/21fe2bbf35e5e639b3a5
I don't know what the PNG output looks like, but the SVG output is definitely more useful than the ASCII art on big trees. Having multiple _repr methods doesn't "conflict" but provides more options.

jnothman:
The SVG output can be readily made more useful via interactivity. One could
click to collapse nodes, for instance, like in the graphical CCG output at
http://svn.ask.it.usyd.edu.au/trac/candc/wiki/Demo (which seems to be down
atm).

stevenbird:
@andreasvc – I agree; any contributions along these lines would be very welcome.

PR #796, opened 2014-11-15
WordNet: lemma name lookup bugfix (#454) was clobbered

nschneid:


stevenbird:
Sorry – I'm not sure how that happened. Thanks.

Issue #602, opened 2014-02-10
Serious problem with SimpleGoodTuringProbDist


htInEdin:
Given any fd with all singletons, SimpleGoodTuring becomes seriously confused, assigning _all_ probability mass to unseens:

> > > fd=FreqDist(['a','b','c'])
> > > sgt=SimpleGoodTuringProbDist(fd)
> > > sgt.prob('c')
> > > 0.0
> > > sgt.prob('d')
> > > 1.0

Since this actually happens a lot with real ngram data, it renders not only SGT unusable, but the default for NgramModel.

stevenbird:
Thanks for the bug report. I've asked for a volunteer from nltk-dev to look into this.
https://groups.google.com/forum/#!topic/nltk-dev/pvusfr7OFnY

kmike:
Both Good Turing and Simple Good Turing probability estimation methods assign `Nr(1)/N` probability mass to unknown events, and when all events are single this mass would be 1.0, leaving 0.0 for seen events. So I think it is a pathological case for SGT probability estimation method and you shouldn't use SGT for such data.

We may detect such cases in `_renormalize()` method and issue a warning (raising an exception seems too harsh).

There is also https://github.com/nltk/nltk/issues/383 which may cause troubles even if there are events with frequency > 1 - make sure you set `bins` parameter manually.

htInEdin:
Mikhail Korobov writes:

> Both Good Turing and Simple Good Turing probability estimation
> methods assign `Nr(1)/N` probability mass to unknown events, and
> when all events are single this mass would be 1.0, leaving 0.0 for
> seen events. So I think it is a pathological case for SGT
> probability estimation method and you shouldn't use SGT for such
> data.

That's just not a sufficient answer when NLTK ships with SGT as the
default estimator for NgramModel, and almost any NgramModel built from
nltk corpora _will_ contain such cases.

> We may detect such cases in `_renormalize()` method and issue a
> warning (raising an exception seems too harsh).

Shouldn't the warning be accompanied by a useful return value?
Returning 0 will, _inter alia_, cause an Exception if someone has
called logprob.

> There is also https://github.com/nltk/nltk/issues/383 which may
> cause troubles even if there are events with frequency > 1 - make
> sure you set `bins` parameter manually.

Again, if this is true then this default (from ngram.py)

def _estimator(fdist, bins):
    """
    Default estimator function using a SimpleGoodTuringProbDist.
    """
    # can't be an instance method of NgramModel as they
[...]

Issue #396, opened 2013-04-24
make some good doctests for nltk.model.NgramModel 


alexrudnick:
I don't think there are any tests for NgramModel at all right now...

But there should be some.

Tests should include the obvious things like creating a model, using some different estimators, sampling from it. And also pickling and loading from a pickle.

kmike:
There is a simple test in `NgramModel.__init__` docstring (it could be executed using runtests.py script), but it only checks that there are no syntax errors and that entropy calculation works; +1 for proper NgramModel tests (esp. given all recently opened NgramModel tickets).

iliakur:
@stevenbird this can also be closed while we're at it!

Issue #367, opened 2013-03-07
Error in NgramModel backoff smoothing calculation?


bcroy:
I believe there is an error in how backoff smoothing is calculated in NgramModel.
- Consider the sequence of "words": aaaababaaccbacb with words ['a','b','c']
- Build a bigram model (n=2). For simplicity use LidstoneProbDist smoothing
- Notably, this sequence does not contain all bigrams prefixed by 'b' or 'c'. Thus backoff is required to get the probability of bigram 'bb', 'bc' and 'ca'
- For context = ['a'], model.prob(w,context) looks good for all words and sums to 1
- For context = ['b'], model.prob(w,context) doesn't look right, sums to > 1

I thought that the backoff calculation should do the following, say for the context 'b':
- Calculate the total "missing" probability for unseen values in the 'b' context (ie. bb and bc), at the bigram level, call this Beta2
- Calculate the total unigram probability for these unseen values (ie. 'b' and 'c'), call this Beta1
- return (Beta2 / Beta1) \* backoff.prob()

This essentially swaps in the unigram probabilities for those words that were unobserved in the bigram context, scaled appropriately, to fill in the missing probability mass.

Am I missing something? The code in NgramModel does something rather different it seems, and I couldn't make sense of it.

dan-blanchard:
I believe you are correct that this is indeed a bug.

If we assume this setup:

``` python
from nltk.model import NgramModel
from nltk.probability import LidstoneProbDist

word_seq = list('aaaababaaccbacb')
words = ['a', 'b', 'c', '']

est = lambda freqdist, bins: LidstoneProbDist(freqdist, 0.2, bins=bins)
model = NgramModel(2, word_seq, True, True, est, 4)
```

We can see pretty quickly the discrepancies:

``` python
sum(model.prob(w, ['b']) for w in words)
Out[150]: 2.4583333333333335
sum(model.prob(w, ['a']) for w in words)
Out[151]: 1.0

[(w, model.prob(w, ['b'])) for w in words]
Out[152]: 
[('a', 0.6666666666666667),
 ('b', 0.875),
 ('c', 0.6666666666666667),
 ('', 0.25)]

[(w, model.prob(w, ['a'])) for w in words]
Out[153]: 
[('a', 0.47727272727272724),
 ('b', 0.25),
 ('c', 0.25),
 ('', 0.022727272727272728)]
```

[...]

PR #800, opened 2014-12-02
A stab at fixing the ngram model

iliakur:
The commit messages should give you a good idea of what I've done specifically, so I'm going to give a high-level overview here as well as ask some questions. 
1. The goal of this PR isn't necessarily to fix **everything** wrong with the ngram and probability modules but to get started on that task by enabling basic ngram functionality. 
2. I've implemented @bcroy's take on backoff from #367 and set Lidstone as the default estimator based on the discussion in #602. I suspect  both of these changes have been expected by the community and shouldn't spark much discussion.
3. Here comes the controversial part. I think I've found the cause of the bugs @afourney pointed out in #367. The problem was that nltk's implementation of conditional probability distributions by default breaks up the training sample space into sample spaces for each condition/context. Thus, for every condition/context the denominator is _not_ the **N** of the total number of outcomes _but_ instead the **N** of outcomes observed given a condition/context. This means that as soon as we start summing over outcomes observed during training _but not_ given a certain condition the result will be > 1! I have fixed this by computing an "override_N" value based on the **N** of the whole distribution and then passing that down to the estimator. As of this writing I am certain there's a more natural way to implement this given nltk's class topology. If folks are on board with the concept, I'd be willing to try out some alternative refactorings.
4. @kmike I know we went over this in #326, but I'm still very fuzzy on how testing works. I tried running "tox" from the root of my nltk repo even before I made any of my changes and got 3 to 5 errors right off the bat. I see myself working on the ngram and probability module for a while, so I might as well figure this testing stuff out now with this PR. Who knows, maybe we'll even solve #396!

### DIFF
--- a/nltk/model/ngram.py
+++ b/nltk/model/ngram.py
@@ -118,6 +118,33 @@ class NgramModel(ModelI):
         if n > 1:
             self._backoff = NgramModel(n-1, train, pad_left, pad_right,
                                        estimator, *estimator_args, **estimator_kwargs)
+            self._backoff_alphas = dict()
+            # For each condition (or context)
+            for ctxt in cfd.conditions():
+                prdist = self._model[ctxt] # prob dist for this context
+
+                backoff_ctxt = ctxt[1:]
+                backoff_total_pr = 0.0
+                total_observed_pr = 0.0
+                for word in cfd[ctxt]:
+                    # this is the subset of words that we OBSERVED
+                    # following this context
+                    total_observed_pr += prdist.prob(word)
+                    # we normalize it by the total (n-1)-gram probability of
+                    # words that were observed in this n-gram context
+                    backoff_total_pr += self._backoff.prob(word, backoff_ctxt)
+
+                assert (0 < total_observed_pr <= 1), total_observed_pr
+                # beta is the remaining probability weight after we factor out
+                # the probability of observed words
+                beta = 1.0 - total_observed_pr
+
+                # backoff total has to be less than one, otherwise we get
+                # ZeroDivision error when we try subtracting it from 1 below
+                assert (0 < backoff_total_pr < 1), backoff_total_pr
+                alpha_ctxt = beta / (1.0 - backoff_total_pr)
+
+                self._backoff_alphas[ctxt] = alpha_ctxt
 
     def prob(self, word, context):
         """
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @Copper-Head. I don't mind re-instating the `model` package once its behaving sensibly, even if there are known issues (point 1). I agree that it's fine to adopt Lidstone as the default estimator (point 2). Let's try to get some comments on the design and correctness of your fixes to `probability.py` (point 3) – @edloper, @heatherleaf?

stevenbird:
Thanks again @Copper-Head; I'm going to merge this into the `model` branch and call for feedback on nltk-dev.

stevenbird:
@Copper-Head – I had problems updating the `model` branch from `develop`, so I created a fresh branch of `develop` called `model2` and checked out the model package. This lost the commit history unfortunately, but I've identified you as a co-author of `model/ngram.py`.

When I run the doctests I see this error:

```
TypeError: <lambda>() got an unexpected keyword argument 'override_N'
```

Can you replicate this? I'd be grateful if you could look into this and suggest a fix please. (I'd like to retain the illustration of how to provide an estimator in `ngram.py:50`.)

stevenbird:
I'm not sure how this happened, but in the NgramModel initializer https://github.com/nltk/nltk/blob/model2/nltk/model/ngram.py#L40 we have keyword arguments followed by non-keyword arguments.

stevenbird:
I've done some cleaning up, deleting the model branch and renaming model2 back to model. I've taken a closer look at what's going on with [override_N](https://github.com/nltk/nltk/blob/model/nltk/model/ngram.py#L123). It turns out that [override_N is only defined for LidstoneProbDist](https://github.com/nltk/nltk/blob/model/nltk/probability.py#L649). However, `model.py` uses [override_N when calling ConditionalProbDist](https://github.com/nltk/nltk/blob/model/nltk/model/ngram.py#L125).

Should `override_N` be defined for all of the probability distributions?

stevenbird:
I've pulled this into a new branch [test](https://github.com/nltk/nltk/tree/test) and configured our [CI server](https://nltk.ci.cloudbees.com/job/nltk/) to work on this branch in addition to the develop branch. The log should be available in a few hours.

stevenbird:
Here is the [Jenkins log for the test branch](https://nltk.ci.cloudbees.com/job/test/)

iliakur:
Thanks a bunch for the log, I'll use it's output (and common sense) to update the doctests as best I can. I noticed there were some unrelated failures in there. I'm assuming I don't have to worry about them, correct?

stevenbird:
Thanks. Right, please just look at the relevant failures.

nschneid:
What is the status of this? I was hoping to reuse some course material that involves `NgramModel`—is there a current version of the API that the students can use? Or should we install an older version of NLTK?

stevenbird:
@nschneid – the older version was retired because it had issues. The `model` branch has the current version of the model package. I'll update it from `develop` and push it to the `test` branch so our CI server can report any problems. We may need to add more tests though.

iliakur:
I'm gonna try this again as a final project for a class :)

stevenbird:
@Copper-Head – good to hear. What's your timeframe? A great first step would be to provide some regression tests that demonstrate problems with the current implementation.

iliakur:
I'd say I should be done by the end of february. I'm still figuring out the scope of the project, thanks a bunch for the suggestion regarding regression tests, I'll include that in the list of concrete steps!

PR #802, opened 2014-12-05
Fixed NLTK Downloader's "About" dialog issue with Python 3.x

salamatin:
Attribute name "_top" was misspelled, which resulted in an error when trying to show a fallback ShowText.
tkMessageBox have been renamed to messagebox in Python 3.x. I have added another import to take this into account. From my understanding, the fallback ShowText is not needed now, but I left it just in case.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -1749,7 +1749,11 @@ class DownloaderGUI(object):
             from tkMessageBox import Message
             Message(message=ABOUT, title=TITLE).show()
         except ImportError:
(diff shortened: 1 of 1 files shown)


salamatin:
Looks like I've messed up. Is there still a way to split those two changes and make two different pull requests?

Sorry about this, I'm a git noob :(

iliakur:
Try the stuff people recommend here:
http://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits

stevenbird:
Thanks @salamatin

Issue #735, opened 2014-08-16
Stanford tokenizer -- request to add -outputFormat inlineXML


Liontooth:
The NLTK.tokenize.stanford class provides access to the stanford-postagger, with these options:

class nltk.tokenize.stanford.StanfordTokenizer(path_to_jar=None, encoding='UTF-8', options=None, verbose=False, java_options='-mx1000m')

``` python```
 >>> from nltk.tokenize.stanford import StanfordTokenizer
 >>> st = StanfordTokenizer(
 '/software/java/stanford-POS/stanford-postagger-2014-06-16/stanford-postagger.jar', 'utf-8')
 >>> st.tokenize(s)
 [u'Good', u'muffins', u'cost', u'$', u'3.88', u'in', u'New', u'York', u'.', u'Please', u'buy', u'me', u'two',
 u'of', u'them', u'.', u'Thanks', u'.']
```

This works, but there is no explicit sentence divider in this slashTags format. However, in -outputFormat  inlineXML, sentence boundaries are explicit. Is there a way to pass this option to the postagger?

It would seem the options= switch should work, but alas:

``` python
>>> st = StanfordTokenizer('/software/java/stanford-POS/stanford-postagger-2014-06-16/stanford-postagger.jar', 'utf-8', java_options='-mx300m', options='-outputFormat inlineXML')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/nltk/tokenize/stanford.py", line 50, in __init__
    self._options_cmd = ','.join('{0}={1}'.format(key, json.dumps(val)) for key, val in options.items())
AttributeError: 'str' object has no attribute 'items'
```

Cheers,
Dave

longdt219:
Please note that -outputFormat inlineXML is not the correct option for stanford tokenizer. It is the option for POS MaxEnt tagger.

Here is the list of valid options : 
https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/process/PTBTokenizer.java

For the purpose of keeping the newline we should do 
==> st = StanfordTokenizer('/software/java/stanford-POS/stanford-postagger-2014-06-16/stanford-postagger.jar', 'utf-8', java_options='-mx300m', options={"tokenizeNLs": True})
==> st.tokenize(s) 
[...]

PR #808, opened 2014-12-10
Fix bug  Stanford tokenizer -- request to add -outputFormat inlineXML #735 

longdt219:
The list of valid options for string tokenizer is 
https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/process/PTBTokenizer.java

Thus, for preserving the "newline character" we should use 
==> st = StanfordTokenizer('stanford-postagger.jar', 'utf-8', java_options='-mx300m', options='-tokenizeNLs true')
==> st.tokenize(s)
[u'Good', u'muffins', u'cost', u'$', u'3.88', u'*NL*', u'in', u'New', u'York', u'.', u'Please', u'buy', u'me', u'*NL*', u'two', u'of', u'them', u'.', u'*NL*', u'Thanks', u'.']
The "*NL*" is the new line character. 

Code modified accordingly to resolve this.

### DIFF
--- a/nltk/tokenize/stanford.py
+++ b/nltk/tokenize/stanford.py
@@ -46,9 +46,28 @@ class StanfordTokenizer(TokenizerI):
 
         self._encoding = encoding
         self.java_options = java_options
-        options = {} if options is None else options
-        self._options_cmd = ','.join('{0}={1}'.format(key, json.dumps(val)) for key, val in options.items())
-
+        #options = {} if options is None else options
+        
+        # Long Duong : fix bug #735 
+        options_str = options
+        options = {} 
+        if options_str is not None:
+            tokens = options_str.split()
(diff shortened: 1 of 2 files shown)


stevenbird:
I cherry-picked the bugfix in https://github.com/nltk/nltk/commit/9129b55f5e1aea58cd536092cf612c6a67e39d09

stevenbird:
Thanks @longdt219. Does this approach need to be applied to the other Java wrappers?

longdt219:
Hi @stevenbird, currently we use default option (not support for user option) for many other wrappers i.e. malt parser, stanford parser, humpos tagger, stanford tagger etc. That's why I don't think this approach need to be applied to other java wrappers yet.

stevenbird:
ok, that's fine, thanks

PR #811, opened 2014-12-11
Triples

rmalouf:
Two small additions to `DependencyGraph`:
- `zero_based` option to `__init__` and `load`, to allow input with zero-based node numbering. For some reason, [zpar](http://www.sutd.edu.sg/cmsresource/faculty/yuezhang/zpar.html) produces output that is almost in Malt-TAB format except that the first word is number 0 rather than 1.
- `triples()` method returns all the dependency triples in a `DependencyGraph`, where a triple has the form ((_head-word_, _head-ctag_), _relation_, (_dependent-word_, _dependent-ctag_)).

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -221,6 +221,23 @@ class DependencyGraph(object):
         deps = node['deps']
         return Tree(word, [self._tree(i) for i in deps])
 
+    def triples(self, node=None):
+        """
+        Extract dependency triples of the form: 
+        ((head word, head tag), rel, (dep word, dep tag))
+        """
+
+        if not node:
+            node = self.root
+
+        head = (node['word'], node['ctag'])
+        for i in node['deps']:
+            dep = self.get_by_address(i)
+            yield (head, dep['rel'], (dep['word'], dep['ctag']))
+            for triple in self.triples(node=dep):
+                yield triple
+
+
     def _hd(self, i):
         try:
             return self.nodelist[i]['head']
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -120,14 +120,14 @@ class DependencyGraph(object):
         return "<DependencyGraph with %d nodes>" % len(self.nodelist)
 
     @staticmethod
-    def load(file, zero_based=False):
+    def load(filename, zero_based=False):
         """
-        :param file: a file in Malt-TAB format
-        :param zero_based: nodes in the input file are numbered starting from 0 rather 
+        :param filename: a name of a file in Malt-TAB format
+        :param zero_based: nodes in the input file are numbered starting from 0 rather
             than 1 (as produced by, e.g., zpar)
         :return: a list of DependencyGraphs
         """
-        with open(file) as infile:
+        with open(filename) as infile:
             return [DependencyGraph(tree_str, zero_based=zero_based) for tree_str in
                                                   infile.read().split('\n\n')]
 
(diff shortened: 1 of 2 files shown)


nltk/parse/dependencygraph.py
```diff
@@ -221,6 +228,23 @@ def tree(self):
...
+    def triples(self, node=None):
+        """
+        Extract dependency triples of the form:
+        ((head word, head tag), rel, (dep word, dep tag))
+        """
+
+        if not node:
+            node = self.root
+
+        head = (node['word'], node['ctag'])
```
dimazest, line 240: There is no need for (...)

`head = node['word'], node['ctag']`

nltk/parse/dependencygraph.py
```diff
@@ -221,6 +228,23 @@ def tree(self):
...
+        for i in node['deps']:
+            dep = self.get_by_address(i)
```
dimazest, line 242: It might may sense to iterate over `self.nodelist` directly. `get_by_address()` does an iteration to find a needed address.

https://github.com/nltk/nltk/pull/811/files#diff-31ba76604fcce0dbd82cdfd1dba4233dL93

nltk/parse/dependencygraph.py
```diff
@@ -221,6 +228,23 @@ def tree(self):
         deps = node['deps']
         return Tree(word, [self._tree(i) for i in deps])
 
+    def triples(self, node=None):
+        """
+        Extract dependency triples of the form:
+        ((head word, head tag), rel, (dep word, dep tag))
```
dimazest, line 234: consider using named tuples here.
- dependency
  - head
  - relation
  - dependent
- head and dependent are also named tuples with the following fields
  - token
  - tag

though i'm not sure what the best names are, would be great to be consistent with malt documentation.

nltk/parse/dependencygraph.py
```diff
@@ -221,6 +228,23 @@ def tree(self):
...
+        """
+
+        if not node:
+            node = self.root
+
+        head = (node['word'], node['ctag'])
+        for i in node['deps']:
+            dep = self.get_by_address(i)
+            yield (head, dep['rel'], (dep['word'], dep['ctag']))
```
dimazest, line 243: no need of brackets

nltk/parse/dependencygraph.py
```diff
@@ -117,13 +120,15 @@ def __repr__(self):
         return "<DependencyGraph with %d nodes>" % len(self.nodelist)
 
     @staticmethod
-    def load(file):
+    def load(filename, zero_based=False):
```
dimazest, line 123: maybe it worth using _args and *_kwargs and pass them to DependencyGraph

dimazest:
looks good! are the tests?

rmalouf:
Thanks! Just added some tests...

nltk/parse/dependencygraph.py
stevenbird, line 234: Note that we haven't embraced named tuples in NLTK (yet). If we do, we'll do it systematically. https://github.com/nltk/nltk/issues/530

stevenbird:
Thanks @rmalouf, looks good.

Thanks for the stylistic feedback @dimazest. I for one don't mind extra parens around tuples, given that they are required in the corresponding mathematical notation. The other things you mention are moot – I'm happy to leave them open. I would welcome offers of help to audit the entire codebase for stale idioms, cf https://github.com/nltk/nltk/issues/815.

PR #817, opened 2014-12-18
nltk.parse tests are fixed for Python 2.6.

dimazest:
Broken test fix and file cleanup.

### DIFF
--- a/nltk/parse/depeval.py
+++ b/nltk/parse/depeval.py
@@ -7,116 +7,118 @@
 # For license information, see LICENSE.TXT
 import unicodedata
 
+
 class DependencyEvaluator(object):
     """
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @dimazest. I see that my attempt to rename `depeval` as `evaluate` wasn't successful and now we have duplication. I'll merge your PR then try to sort this out.

stevenbird:
I've deleted `depeval` since it should not have been in the repository. I cannot identify any 2.6-related error in `evaluate`.

dimazest:
hi, right the formatting changes dominate, though the main reason of this PR is line 64 where python 2.6 renders the score as 0.59999 and later python versions as 0.6.

PR #818, opened 2014-12-18
Standardize the method for replacing suffixes

### LINKED ISSUE #756: Remove word suffixes consistently (and only once) on snowball stemmers
While comparing the Portuguese and Spanish stemmers I noticed they use different methods to remove the suffixes. For example:
- [SpanishStemmer](https://github.com/nltk/nltk/blob/c23c18de204961b513947e5c593a196d06fdd6ee/nltk/stem/snowball.py#L3479-L3487) uses `word = word.replace(suffix, ...)`
- PortugueseStemmer, FrenchStemmer and others uses `word = "".join((word[:-len(suffix)], ...))`

I believe the code should use only the second form, both for consistency and because the first form can potentially replace more than one occurrence of the suffix (although that should not happen in practice).

he7d3r:
Fixes #756 .
Update a few occurrences by
- searching for `""\.join\(\((\w+)\[:-len\(suffix\)\], "(\w+)"\)\)` and
- replacing by `self._suffix_replace(\1, suffix, "\2")`

Other places were updated manually.

### DIFF
--- a/nltk/stem/snowball.py
+++ b/nltk/stem/snowball.py
@@ -23,6 +23,7 @@ from __future__ import unicode_literals, print_function
 from nltk import compat
 from nltk.corpus import stopwords
 from nltk.stem import porter
+from nltk.stem.util import suffix_replace
 
 from nltk.stem.api import StemmerI
 
@@ -282,8 +283,6 @@ class _StandardStemmer(_LanguageSpecificStemmer):
 
         return rv
 
-
-
 class DanishStemmer(_ScandinavianStemmer):
 
     """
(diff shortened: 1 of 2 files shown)


he7d3r:
I wonder if we should also search for `(\w+) = \1\[:-len\((\w+)\)\]` and replace by `\1 = self._suffix_replace(\1, \2, "")`, or if the overhead of calling a function is not worth it.

stevenbird:
Thanks @he7d3r this is a big improvement.

he7d3r:
I [updated](https://github.com/he7d3r/nltk/commit/fcb2fc60c7e04d3f36ef514ce80f1e1e6ac38e5d#diff-9ecaa583357a87d05ad2735b49d41583R1) the pull request.

stevenbird:
Thanks @he7d3r

Issue #543, opened 2013-11-07
Method for finding senna executable in senna.py should be improved


ewan-klein:
Code which wraps a 3rd party executable should adopt common methods for locating the executable. Existing wrappers for prover9/mace, malt parser, mallet etc. do a reasonably intelligent (though not optimal) job of doing this. senna.py has a hard-wired path which won't work for people on Windows or who don't have root access.

stevenbird:
Senna.py should probably use `internals.find_binary`

stevenbird:
We can check an environment variable `SENNA`. Also need to update the wiki instructions: https://github.com/nltk/nltk/wiki/Installing-Third-Party-Software

stevenbird:
Some issues with Senna are showing up on the [CI server](https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py26-jenkins/lastCompletedBuild/testReport/)

stevenbird:
Previously, Senna was not installed on the CI server, and so the doctest was being skipped. @longdt219 modified jenkins.sh to download Senna.

PR #820, opened 2014-12-23
Fix bug #543 (finding senna executable file)

longdt219:
Fixing bug #543, looking for binary executable file from environment variable "SENNA"

### DIFF
--- a/nltk/tag/senna.py
+++ b/nltk/tag/senna.py
@@ -55,7 +55,8 @@ class SennaTagger(TaggerI):
     misalignment errors.
 
     The input is:
-    - path to the directory that contains SENNA executables.
+    - path to the directory that contains SENNA executables. If the path is incorrect, 
+       SennaTagger will automatically search for executable file specified in SENNA environment variable
     - List of the operations needed to be performed.
     - (optionally) the encoding of the input data (default:utf-8)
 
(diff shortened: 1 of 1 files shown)


longdt219:
I modified some code to remove this hard-coded path for doctest

stevenbird:
Thanks @longdt219

nltk/tag/senna.py
```diff
@@ -77,29 +79,41 @@ class SennaTagger(TaggerI):
     def __init__(self, senna_path, operations, encoding='utf-8'):
         self._encoding = encoding
         self._path = path.normpath(senna_path) + sep
+        
+        # Long Duong : Fix bug #543 (looking for Senna executable) 
+        
+        # Verifies the existence of the executable on the self._path first    
+        #senna_binary_file_1 = self.executable(self._path)
+        exeFile1 = self.executable(self._path)
+        if not path.isfile(exeFile1):
```
kmike, line 88: 1. What about preferring SENNA environment variable? What if user has executable self._path file, but wants to use another?
2. exeFile1 and exeFile2 names are not PEP8.

longdt219, line 88: We must give preference for one of those, either SENNA environment variable or user defined path. In this case, I think we should give preference for user defined path. It will be more explicit for the user.

kmike, line 88: ah, I see - sorry for the noise! For some reason I thought self._path is a hardcoded path.

PR #822, opened 2014-12-28
Refactor DependencyGraph

dimazest:
Note that this is still work in progress.
- [x] DependencyGraph stores nodes in a list, where position of nodes is it's address (or index). In a situation when a node depends on an unseen node this is problematic (so variable `temp` was created to deal with this problem).
- [x] `input_` for `_parse` can be either a string (the old behavior) or an iterable of lines (so it's possible to use DependencyGraph to parse a large corpus, for example ukwac).
- [x] Tests don't cover all the functionality, I'll try to cover it as much as possible.
- [x] It's possible to specify cell separator.
- [ ] `DependencyGraph.nodes` should be ordered by address.
- [ ] `DependencyGraph.nodes[...]['deps']` should be an ordered dict as well, where keys are relations. This will make it easy to retrieve all the dependents of a node. Btw, are relations unique?

Some other points
- I'm not sure whether `remove_by_address`, `get_by_address` and `contains_address` are needed. They duplicate dictionary API.
- `right_children()` should return the list of children, computing the length of it is trivial.
- Maybe, it makes sense make `triples()` to return node, not just a word, its head and a ctag.

In general, the PR is not ready yet, I might change it a bit: restructure the code, write more tests and documentation.


stevenbird:
Thanks @dimazest.
@longdt219 you might be interested in monitoring this.

nltk/sem/lfg.py
```diff
@@ -37,45 +38,52 @@ def to_glueformula_list(self, glue_dict):
...
+                'word': self.pred[0],
+                'tag': self.pred[1],
+                'head': head,
+                'rel': rel,
+            }
+        )
 
-        for feature in self:
-            for item in self[feature]:
+        for feature in sorted(self):
```
dimazest, line 68: self is a dictionary, so the order of the keys is not determined, I use `sort` to make sure that the order is always the same.

longdt219:
It does affect my code. Thanks @stevenbird

dimazest:
I've updated modules that use DependencyGraph and added tests (based on demos). The PR is ready to be reviewed and merged.

nltk/parse/nonprojectivedependencyparser.py
```diff
@@ -69,22 +74,47 @@ def score(self, graph):
...
     for classification.
+
+    >>> from nltk.parse.dependencygraph import DependencyGraph, conll_data2
+
+    >>> graphs = [DependencyGraph(entry) for entry in conll_data2.split('\\n\\n') if entry]
+    >>> npp = ProbabilisticNonprojectiveParser()
+    >>> npp.train(graphs, NaiveBayesDependencyScorer())
+    >>> for parse_graph in npp.parse(['Cathy', 'zag', 'hen', 'zwaaien', '.'], ['N', 'V', 'Pron', 'Adj', 'N', 'Punc']):
+    ...     print(parse_graph)
+    {0: {'address': 0,
```
dimazest: it seems that nodes miss the `head` value. `to_conll()`` expects it.

nltk/parse/nonprojectivedependencyparser.py
```diff
@@ -69,22 +74,47 @@ def score(self, graph):
...
+         'ctag': 'TOP',
+         'deps': [2],
+         'feats': None,
+         'lemma': None,
+         'rel': 'TOP',
+         'tag': 'TOP',
+         'word': None},
+     1: {'address': 1, 'deps': [], 'rel': 'NTOP', 'tag': 'N', 'word': 'Cathy'},
```
dimazest: it seems that nodes miss the head value. `to_conll()` expects it.

`ProbabilisticNonprojectiveParser` should provide all required information.

nltk/parse/nonprojectivedependencyparser.py
```diff
@@ -69,22 +74,47 @@ def score(self, graph):
...
     It uses head-word, head-tag, child-word, and child-tag features
     for classification.
+
+    >>> from nltk.parse.dependencygraph import DependencyGraph, conll_data2
+
+    >>> graphs = [DependencyGraph(entry) for entry in conll_data2.split('\\n\\n') if entry]
+    >>> npp = ProbabilisticNonprojectiveParser()
+    >>> npp.train(graphs, NaiveBayesDependencyScorer())
+    >>> for parse_graph in npp.parse(['Cathy', 'zag', 'hen', 'zwaaien', '.'], ['N', 'V', 'Pron', 'Adj', 'N', 'Punc']):
+    ...     print(parse_graph)
```
dimazest: these tests are broken for Python 2...

dimazest:
I've added two more points about dependent node storage. Seems that order **is** important and just storing it in a dict may break many things.

stevenbird:
Thanks for all this work @dimazest. I've inspected the changes and they look fine. I'm going to merge and see how the testing goes at https://nltk.ci.cloudbees.com/job/nltk/

longdt219:
Hi @dimazest, 
Could you please check the following code 

```
    if not self.nodes[0]['deps']['ROOT']:
        raise DependencyGraphError(
            "The graph does'n contain a node "
            "that depends on the root element."
        )
```

It will fail for the following test case: 
gold_sent = DependencyGraph("""
Economic  JJ     2      ATT
news  NN     3       SBJ
has       VBD       0       PRED
little      JJ      5       ATT
effect   NN     3       OBJ
on     IN      5       ATT
financial       JJ       8       ATT
markets    NNS      6       PC
.    .      3       PU
""")

Probably you don't want to fix the root relationship as 'ROOT' since it might be predicate relationship i.e 'PRED'

stevenbird:
Some issues for the dependency parsers are showing up on the [CI server](https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py26-jenkins/lastCompletedBuild/testReport/). Would you please take a look @dimazest

dimazest:
Thanks for feedback and merge. I'll have a look to the tests and think how to get the relationship from a root node in a reliable way.

Do you think it's a problem that dg.nodes.values() are not ordered?

dimazest:
and a follow up question. @longdt219, how many nodes will depend on the ROOT node? I suspect there might be several of them (possibly with different relations). Does it make sense to have `.root` attribute, it seems it's not really a root.

nltk/corpus/reader/knbc.py
```diff
@@ -135,30 +155,33 @@ def _knbc_fileids_sort(x):
...
+    print(''.join(knbc.words()[:100]))
 
-    print('\n\n'.join( '%s' % tree for tree in knbc.parsed_sents()[:2] ))
+    print('\n\n'.join(str(tree) for tree in knbc.parsed_sents()[:2]))
 
     knbc.morphs2str = lambda morphs: '/'.join(
-        "%s(%s)"%(m[0], m[1].split(' ')[2]) for m in morphs if m[0] != 'EOS'
-        ).encode('utf-8')
+        "%s(%s)" % (m[0], m[1].split(' ')[2]) for m in morphs if m[0] != 'EOS'
+    ).encode('utf-8')
```
kmike, line 164: It is suspicious we have to encode string to utf8 - shouldn't corpus reader work on unicode? And does the demo work in Python 3?

nltk/parse/dependencygraph.py
```diff
@@ -78,74 +97,64 @@ def add_arc(self, head_address, mod_address):
...
-                                                  infile.read().split('\n\n')]
 
-    @staticmethod
-    def _normalize(line):
-        """
-        Deal with lines in which spaces are used rather than tabs.
         """
-        SPC = re.compile(' +')
-        return re.sub(SPC, '\t', line).strip()
+        with open(filename) as infile:
```
kmike, line 141: I think file should be opened in text mode, we should get unicode data from it and be explicit about the required encoding. Now `tree_str` is a bytestring.

nltk/parse/dependencygraph.py
```diff
@@ -153,70 +162,113 @@ def right_children(self, node_index):
...
-        self.root = self.nodelist[root_address]
+        if isinstance(input_, string_types):
+            input_ = (line for line in input_.split('\n'))
+
+        lines = (l.rstrip() for l in input_)
+        lines = (l for l in lines if l)
+
+        for index, line in enumerate(lines, start=1):
+            cells = line.split(cell_separator)
+            nrCells = len(cells)
```
kmike, line 211: maybe rename it to conform to PEP8?

nltk/parse/dependencygraph.py
```diff
@@ -244,48 +296,75 @@ def triples(self, node=None):
...
             for pair in new_entries:
                 distances[pair] = new_entries[pair]
                 if pair[0] == pair[1]:
-                    print(pair[0])
-                    path = self.get_cycle_path(self.get_by_address(pair[0]), pair[0]) #self.nodelist[pair[0]], pair[0])
+                    path = self.get_cycle_path(self.get_by_address(pair[0]), pair[0])
                     return path
-        return False  # return []?
 
+        return False  # return []?
```
kmike, line 360: Why not return [] indeed? The method can be renamed e.g. to get_cycles then. For backwards compatibility (if it is desired) we can add something like

```
def contains_cycle(self):
    return self.get_cycles() or False
```

longdt219:
Hi @dimazest, 
There might be many nodes have different relationship with root node. Here is what I think 
(1) The nodes[0] is always the ROOT node so you don't need the .root attribute. 
(2) Only keep the basic contrains about the dependency graph i.e. no-cycle. Remove all other contrains. This will make the dependency graph flexible 
(3) Keeping dependency graph simple. For example, we don't need the 'deps' attribute for each nodes. The 'head' is enough.

dimazest:
@longdt219  originally, `.root` is the node that the ROOT node points to. I think some code relies on this fact. However you are right that the dependency graph needs to be as flexible as possible, so it's capable of dealing with trees produced by parsers.

Regarding removing `deps` is a good idea to keep the code simple, but quite often one needs to know what are the children of some node. For example, to get the subject and the object of a verb. Without `.deps` that will require additional iteration over the graph.

Thank you all for the comments. I'll come back to this once I clear some university work, but I should be able to come back soon.

PR #823, opened 2014-12-29
Add case sensitivity

achembarpu:
Both the `Data dir` path and `Server URL` are `str.lower()`ed.

However, case-sensitivity depends on the file-system, and is handled by python's `os.path`.
Additionally, according to the [W3 spec](http://www.w3.org/TR/WD-html40-970708/htmlweb.html), URLs are case-sensitive as well.
Therefore, `nltk` shouldn't be implicitly lowercasing these strings.
Thoughts?

Note - This affects me, since I store `nltk_data` in `$HOME/Development/data/nltk`.


stevenbird:
Thanks @arvindch – this looks good. I don't see why we normalized these to lowercase.

PR #829, opened 2015-01-10
Update setup.py

robic:


stevenbird:
This duplicates https://github.com/nltk/nltk/pull/827

PR #832, opened 2015-01-11
framenet corpus reader: fes() method to search by frame element name

nschneid:
I have found myself searching FrameNet by FE name a lot; this method will make that more convenient. CC: @wooters


wooters:
Nice addition. Looks good to me!

stevenbird:
Thanks @nschneid

PR #833, opened 2015-01-12
fix which stderr

0ssifrage:
Avoid displaying `which` error message like:

``` bash
which: no java.exe in (/usr/local/sbin:...)
```

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -491,7 +491,8 @@ def find_file_iter(filename, env_vars=(), searchpath=(),
     if os.name == 'posix':
         for alternative in file_names:
             try:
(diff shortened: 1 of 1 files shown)


stevenbird:
Should it raise an error instead of sending stderr to the pipe?

0ssifrage:
It always display the `which` error because there is no `java.exe` on Linux, but the correct path of `java` on Linux has been found before. I think there is no need to raise the error.

0ssifrage:
Someone mentioned the same problem [here](https://groups.google.com/forum/#!topic/nltk-users/wfxsntOjybI).

stevenbird:
Thanks @0ssifrage.

Issue #694, opened 2014-07-08
Transition-based dependency parsing algorithms


stevenbird:
Implement Nivre's arc-eager and arc-standard algorithms.
http://aclweb.org/anthology/J08-4003.pdf
(Note that there have been some minor improvements made to the algorithms since these publications; please contact the author for details.)

charnugagoo:
I strongly recommend to implement this algorithm or shift-reduce parser since these linear parsing algorithm could be very fast comparing to bottom-up parser. For some situation, say parsing large data set, this would be very useful. And in academic, these linear parsers could be almost as good (in accuracy) as bottom-up parser.

I really want to work on this, but I don't have much after-work time. I wish I could help not as a coder, as a reviewer or at least in discussion.

Thanks!

stnatic:
I would be interested in contributing. I think that the biggest question is how to implement the classifier. We could either do the entire training process in Python or to start by importing a pre-trained model, e.g. from MaltParser.

@charnugagoo we could start a thread on nltk-dev group and hear some feedback from the authors of ntlk

stevenbird:
Our philosophy in NLTK is to provide both, i.e. pure Python implementation(s) that have educational value, and interfaces to optimised third-party implementations. If you can pick a pre-trained model that you would like included with the NLTK-data collection, I'm happy to take care of adding that.

charnugagoo:
@stnatic Personally, I recommend perceptron, which works well when I was writing my own shift-reduce parser. An average perceptron is very easy to implement. I searched there are some papers talking about perceptron + arc eager, so I think it would work in this case.

I agree. Could you send out an email to nltk-dev, please?

stnatic:
I had almost no time during the last 2 weeks so I apologize for not starting a mailing group thread.

I see that @longdt219 is now assigned to this issue. Although, it seems to be a rather big task assuming that the classifier would be written in pure Python. Is there still any room for contribution? If so, please let me know.

longdt219:
Hi @stnatic, I'm a bit confuse here, why we should re-implement the classifier given that nltk already provides many classifiers i.e. svm, maxent etc.

stnatic:
Hi @longdt219, I didn't mean implementing SVM from the ground up, I rather meant that converting a CoNLL treebank data to learning data for the learning black-box might take a solid amount of programming effort.

[...]

PR #836, opened 2015-01-13
Transition based parsing #694

longdt219:
Algorithm for performing arc-eager and arc-standard dependency parsing.
Refer to #694 for more detail discussion


stevenbird:
Thanks @longdt219.

Issue #761, opened 2014-10-09
NERTagger stops tagging at first: ? ! .


vaisov:
How to reproduce:

```
from nltk.tag.stanford import NERTagger
import sys, os

classifier = '/opt/stanford-ner/classifiers/hgc_175m_600.crf.ser.gz'
ner = '/opt/stanford-ner/stanford-ner.jar'
nt = NERTagger(classifier, ner, 'utf-8')

nt.tag('einen schönen. guten abend. freue mich sehr, dass sie dabei sind herzlich'.split())
```

output:

```
[(u'einen', u'O'), (u'sch\xf6nen', u'O'), (u'.', u'O')]
```

However POSTagger works without problem with same texts.

alvations:
Related to NER tagger, http://stackoverflow.com/questions/26352041/nltk-entity-extraction-difference-from-nltk-2-0-4-to-nltk-3-0

vaisov:
I tried Stanford's NER tagger using NLTK 2.0.4, outcome is the same, no tagging after .!?

stevenbird:
@desilinguist would you please advise?

desilinguist:
I don't know much about the NERTagger portion since I only wrote the POSTagger wrapper. However, I can take a more detailed look and report back.

stevenbird:
Oh, no problem. Before you do that, can @vaisov confirm that the Stanford NER tagger behaves in the desired manner when invoked directly, i.e. on the command line?

vaisov:
Sure. Here are results of using command line without python.

```
root@localhost:/opt/stanford-ner# cat 1
einen schönen guten abend. freue mich sehr, dass sie dabei sind. herzlich willkommen zu unserer sendung. schönen guten abend in die runde. freue mich sehr, dass ihr alle da seid.

root@localhost:/opt/stanford-ner# ./ner.sh 1
CRFClassifier invoked on Thu Oct 16 15:18:29 UTC 2014 with arguments:
   -loadClassifier ./classifiers/hgc_175m_600.crf.ser.gz -textFile 1 -outputFormat slashTags
loadClassifier=./classifiers/hgc_175m_600.crf.ser.gz
textFile=1
outputFormat=slashTags
Loading classifier from ./classifiers/hgc_175m_600.crf.ser.gz ... done [57.1 sec].
einen/O schönen/O guten/O abend/O ./O
freue/O mich/O sehr/O ,/O dass/O sie/O dabei/O sind/O ./O
[...]

PR #841, opened 2015-01-17
Fix bug #761, NERTagger stops tagging at first: ? ! .

longdt219:
Fix bug #761, 
(1) If output of tag is more than one sentence => join them 
(2) Disable Stanford tokenizer

### DIFF
--- a/nltk/tag/stanford.py
+++ b/nltk/tag/stanford.py
@@ -56,7 +56,13 @@ class StanfordTagger(TaggerI):
       raise NotImplementedError
 
     def tag(self, tokens):
-        return self.tag_sents([tokens])[0]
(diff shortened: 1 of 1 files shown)


stevenbird:
The implementation of `tag` looks inefficient:

```
    def tag(self, tokens):
        temp = self.tag_sents([tokens])
    # Handle the case where return more than one sentence.
        # Note that if it is the case, user should use tag_sents instead
        result = []
        for sent in temp:
            result += sent
    return result
```

I'm replacing it with the following:

```
    def tag(self, tokens):
        return list(self.tag_sents([tokens]))
```

Can you explain why the former is preferable?

PR #844, opened 2015-01-20
nkjp reader

ducki13:
Reader of The National Corpus of Polish


stevenbird:
Thanks @ducki13. I think it would be good to include the million-word subcorpus with NLTK's corpus collection, and to add an entry for it in the [corpus interface](https://github.com/nltk/nltk/blob/develop/nltk/corpus/__init__.py). What do you think?

ducki13:
Hello stevenbird. I am very glad that you want my code to be a part of NLTK :) I've already commited an entry. Is anything else what is necessary for right submission of NKJP?

stevenbird:
Nice work @ducki13, and I liked the decorator. We should consider using that elsewhere. I've tested this and will merge it now. Shall we add the million-word subcorpus to NLTK's corpus collection?

ducki13:
Yes, I'd like to submit the corpus and I've already made the pull request with NKJP, here is the link: https://github.com/nltk/nltk_data/pull/17 . Data is submited in taz.bz extension, because size of zip package was too big to make a commit, I wrote about this problem here: https://groups.google.com/forum/#!topic/nltk-dev/VMMCXejQJMQ .

Issue #107, opened 2012-01-17
Language ID module


alexrudnick:
It would be nice to have a language ID module in NLTK.   There are lots of
free implementations out there, most recycling the same couple of ideas:
http://www.let.rug.nl/~vannoord/TextCat/competitors.html

(including some in Python).

I am willing to donate character three-gram data for about 450 languages
assembled by the web crawler I've been running, here:

http://borel.slu.edu/crubadan/index.html

Table of cosine distances:
http://borel.slu.edu/crubadan/table.html

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=532</p><hr/><h2>earlier comments</h2>
<p><strong>StevenBird1 said, at 2010-04-16T21:09:55.000Z:</strong></p>
<p>The language ID data has been added to NLTK's corpus collection
(nltk_data/corpora/langid), currently item #25 in NLTK's corpus list
(http://nltk.googlecode.com/svn/trunk/nltk_data/index.xml).  We need a corpus reader
that loads the data for an individual language/writing-system into a FreqDist.  We
might also want programmatic access to the code table (table.txt), which contains the
mapping between the internal identifiers and ISO 639-3 codes.
</p>

davidsbatista:
Hi! I've been using NLTK for a while and I would like to contribute to the project. Is this task still necessary? If it is, I don't mind doing it. Are there any code guidelines specific for NLTK?

PR #845, opened 2015-01-23
Added corpus reader for n-gram frequencies

avitalp:
Hi there, I've created a corpus reader for n-gram frequencies provided by the "An Crubadan" project. 

As suggested by @stevenbird on https://github.com/nltk/nltk/issues/107 the reader can load individual language data (or all of them) and provide its corresponding n-gram distribution in a FreqDist. There are also multiple helper functions to map between Crubadan, ISO 639-3, and "human friendly" language names. Feedback appreciated.


stevenbird:
Thanks @avitalp. It's great to have this corpus reader. I've merged this into the [crubadan](https://github.com/nltk/nltk/tree/crubadan) branch. I've also renamed the corpus file to `crubadan`.

I'm getting the following error. Would you please take a look at this, and also add some documentation to `nltk/test/corpus.doctest`, and submit a new PR to the `crubadan` branch?

```
>>> from nltk.corpus import crubadan
>>> crubadan.fileids()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/nltk-3.0.2-py3.4.egg/nltk/corpus/util.py", line 99, in __getattr__
    self.__load()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/nltk-3.0.2-py3.4.egg/nltk/corpus/util.py", line 67, in __load
    corpus = self.__reader_cls(root, *self.__args, **self.__kwargs)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/nltk-3.0.2-py3.4.egg/nltk/corpus/reader/crubadan.py", line 45, in __init__
    self._load_lang_mapping_data()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/nltk-3.0.2-py3.4.egg/nltk/corpus/reader/crubadan.py", line 156, in _load_lang_mapping_data
    mapper_file = self.root + '/' + self._LANG_MAPPER_FILE
TypeError: unsupported operand type(s) for +: 'ZipFilePathPointer' and 'str'
```

avitalp:
Thanks @stevenbird. 

The error is due to "self.root" being a ZipFilePointer instead of FileSystemPathPointer, this is set by CorpusReader which the reader subclasses. This module assumes that the Crubadan data is installed (unzipped). Would you mind installing the data and retrying?

I've submitted a PR on crubadan.xml in the nltk_data repo to set the unzip to default on this corpus (https://github.com/avitalp/nltk_data/commit/befd49dc1a1f2f54060699a875a9ae1ba7f5ed08).

Meanwhile, I will add code to handle the not-installed case as well as some documentation and submit a PR.

avitalp:
Hi @stevenbird, I just submitted a PR addressing above: https://github.com/nltk/nltk/pull/884

stevenbird:

stevenbird:
I've done some clean-up work on the corpus reader, and made sure it works with Python 3 (https://github.com/nltk/nltk/commit/9c5fd33dd25b5e192a97ddab32bac55defd2b001). However, I see the initializer loads all languages. In common with other corpus readers, we should only load required languages on demand.

So I propose that the initializer create an empty dictionary `_all_lang_freq` which is private (i.e. not documented in `crubadan.doctest`), and just treated like a cache. When any method requires access to data for a particular language `lang`, a language-specific version of `_load_lang_mapping_data` is called to make sure it is loaded, then `_all_lang_freq[lang]` is accessed.

avitalp:
@stevenbird: clean-up looks good both in the code and docs, thank you :)

I'll re-code it with `_all_lang_freq` private, serving as a cache, populated as langs get added.

I would have to do it differently than you've described though:
The var  `_load_lang_mapping_data` must remain the same as it contains info about any potential language we may want to load on demand. It holds the contents of table.txt, a tiny txt file, so I don't see any performance benefit there. 

Instead, I would modify `lang_freq` to check the cache, if not in there, I'd follow up with a call to `_load_lang_ngrams` and a cache update. Obviously I'd remove `_load_all_ngrams` from the initializer, that's the slowest part of the module. 

Would it be reasonable to load English by default, so that the module is "ready-to-go" after instantiation?

stevenbird:
I agree that it's fine to load the language mapping data at initialization. But no need to load a language file until it is needed, and no need to treat English specially.

avitalp:
Ok, great, I'll change it as described without special treatment to English and resubmit a PR.

avitalp:
@stevenbird I've submitted a PR with the changes discussed above.

stevenbird:
Superseded by #924

PR #846, opened 2015-01-24
Fix definition of _rv_standard

he7d3r:
The test in the definition of RV is "If the first two letters are vowels"
but this is not what `word[:2] in vowels` tests. In fact, this is `False`:

``` python
'august'[:2] in 'aeiou'
```


stevenbird:
Thanks

Issue #104, opened 2012-01-17
Add a CRF POS tagger, replacing the old Mallet interface with CRFSuite


alexrudnick:
NLTK's mallet interface, nltk.classify.mallet, is for mallet 0.4.  It
should be upgraded to the current version (which has a different API).

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=523</p>

xim:
It might be a good idea to document that we only support Mallet 0.4 (current version is 2.0.7) or drop support entirely.

Current master fails as well:

```
Traceback (most recent call last):
  File "nltk/tag/crf.py", line 760, in <module>
    demo()
  File "nltk/tag/crf.py", line 746, in demo
    acc = nltk.tag.accuracy(crf, brown_test)
AttributeError: 'module' object has no attribute 'accuracy'
```

duhaime:
I know this is an old thread, but has this problem been fixed in the current master?

kmike:
No, this problem is not fixed.

rmunn:
This is a problem for packaging NLTK in Linux distrubutions, specifically Debian and Ubuntu. I worked on the Ubuntu package a while ago, and had to remove the mallet functionality entirely from the Ubuntu package because Ubuntu didn't carry Mallet 0.4 at all. And I was unable to get the NLTK package into Debian, for similar reasons: I started looking at whether I could create a mallet0 package for Debian, got bogged down in Java complexities I wasn't familiar with, and ended up abandoning the Debian NLTK package altogether.

I would like to create a Debian package for NLTK version 3 so that it could be available on both Debian and Ubuntu, but until this issue is resolved one way or the other, I doubt I'll be able to create that Debian package.

ewan-klein:
Much of the problem is that mallet.py calls internals.py which adds NLTK_JAR to the Java class path. This is intended to provide a more fine-grained access to parameters that were embedded in the Mallet 0.4 CRF Java code. It is not straightforward to update the Java code that compiles to NLTK_JAR, since it calls classes and fields that are no longer present in Mallet 2.x. I propose that we provide a much simpler interface to the Mallet 2.x CRF classifier which is just a wrapper for cc.mallet.fst.SimpleTagger. I have a crude version of this working now.

stevenbird:
If we need to modify the `ClassifierI` interface please move this to the nltk3beta milestone.

stevenxxiu:
[...]

PR #855, opened 2015-01-25
Add mallet CRF Tagger

longdt219:
Add mallet CRF tagger #104

The accuracy for CRF Tagger evaluated on part of English Penn Treebank 
Training data : 300 sentences 
Testing data : 214 sentences 
--> Accuracy : 81 % 

I also updated the instruction for using Mallet on Wiki.

### DIFF
--- a/nltk/tag/mallet.py
+++ b/nltk/tag/mallet.py
@@ -1,0 +1,249 @@
+# -*- coding: utf-8 -*-
+# Natural Language Toolkit: Interface to the Stanford NER-tagger
+#
+# Copyright (C) 2001-2015 NLTK Project
+# Author: Long Duong <longdt219@gmail.com>
+# URL: <http://nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""
+A module for interfacing with the Mallet CRF taggers.
+"""
+
+import os
+import tempfile
+from subprocess import PIPE
+import unicodedata
+import re 
+
+from nltk.internals import find_jar, config_java, java, _java_options
+from nltk.tag.api import TaggerI
+
+_mallet_url = 'http://mallet.cs.umass.edu/grmm/'
+
+class MalletTagger(TaggerI):
+    """
+    An interface to Mallet CRF taggers.
+    
+    >>> from nltk.tag.mallet import MalletTagger
+    >>> mt = MalletTagger('/Users/HeroAthen/Dropbox/Work/nltk/temp/mallet-2.0.7/dist/mallet.jar', 
+    ... '/Users/HeroAthen/Dropbox/Work/nltk/temp/mallet-2.0.7/dist/mallet-deps.jar')
+ 
+    >>> train_data = [[('University','Noun'), ('is','Verb'), ('a','Det'), ('good','Adj'), ('place','Noun')],
+    ... [('dog','Noun'),('eat','Verb'),('meat','Noun')]]
+    
+    >>> mt.train(train_data,'model.crf.tagger')
+    >>> mt.tag_sents([['dog','is','good'], ['Cat','eat','meat']])
+    [[('dog', 'Noun'), ('is', 'Verb'), ('good', 'Noun')], [('Cat', 'Noun'), ('eat', 'Verb'), ('meat', 'Noun')]]
+    
+    >>> gold_sentences = [[('dog','Noun'),('is','Verb'),('good','Adj')] , [('Cat','Noun'),('eat','Verb'), ('meat','Noun')]] 
+    >>> mt.evaluate(gold_sentences) - 0.83333 < 0.0001
+    True
+    
+    Test the function of finding jar file in the environment variable and setting learned model file  
+    >>> mt = MalletTagger() 
+    >>> mt.set_model_file('model.crf.tagger')
+    >>> mt.evaluate(gold_sentences) - 0.83333 < 0.0001
+    True 
+    
+    """
+    
+    _JAR = 'mallet.jar'
+    _JAR_DEPS = 'mallet-deps.jar'
+
+    def __init__(self, path_to_mallet_jar=None, path_to_mallet_deps_jar = None, verbose=False, java_options='-mx1000m'):
+
+        self._mallet_jar = find_jar(
+                self._JAR, path_to_mallet_jar,
+                env_vars=('MALLET',), searchpath=(), url=_mallet_url,
+                verbose=verbose)
+        
+        self._mallet_dep_jar = find_jar(
+                self._JAR_DEPS, path_to_mallet_deps_jar,
+                env_vars=('MALLET_DEPS',), searchpath=(), url=_mallet_url,
+                verbose=verbose) 
+        
+        self.java_options = java_options
+        self._model_file = ''
+    
+    def set_model_file(self, model_file):
+        self._model_file = model_file
+            
+    def _get_features(self,data):
+        """
+        Extract basic features about this word including 
+             - Current Word 
+             - Is Capitalized ?
+             - Has Punctuation ?
+             - Has Number ?
+             - Suffixes up to length 3 
+        :return : a string which contains the features   
+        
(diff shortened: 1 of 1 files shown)


nltk/tag/mallet.py
```diff
@@ -0,0 +1,249 @@
...
+from nltk.tag.api import TaggerI
+
+_mallet_url = 'http://mallet.cs.umass.edu/grmm/'
+
+class MalletTagger(TaggerI):
+    """
+    An interface to Mallet CRF taggers.
+    
+    >>> from nltk.tag.mallet import MalletTagger
+    >>> mt = MalletTagger('/Users/HeroAthen/Dropbox/Work/nltk/temp/mallet-2.0.7/dist/mallet.jar',
```
kmike, line 29: The test shouldn't depend on hardcoded Mallet paths - it will fail on CI server and on users' machines.

nltk/tag/mallet.py
```diff
@@ -0,0 +1,249 @@
...
+            
+    def _get_features(self,data):
+        """
+        Extract basic features about this word including 
+             - Current Word 
+             - Is Capitalized ?
+             - Has Punctuation ?
+             - Has Number ?
+             - Suffixes up to length 3 
+        :return : a string which contains the features
```
kmike, line 80: Is it an example Mallet wrapper for POS tagging? If so, where did the existing features came from? It could make sense to use features from some published paper, e.g. from http://www.inf.ed.ac.uk/resources/nlp/local_doc/mxpost_doc.pdf, with a reference. 

Alternatively, if it is not just an educational example, and if it is intended to be a production-quality POS tagger, we can make model use features which gives good results on a large treebank. 81% accuracy on POS tagging is far from the state of the arts; it may be explained by a small training data size (300/214 sentences is not a large treebank), but if so features shouldn't be chosen based on what gives best results on this dataset.

nltk/tag/mallet.py
```diff
@@ -0,0 +1,249 @@
...
+             - Current Word 
+             - Is Capitalized ?
+             - Has Punctuation ?
+             - Has Number ?
+             - Suffixes up to length 3 
+        :return : a string which contains the features   
+        
+        >>> from nltk.tag.mallet import MalletTagger
+        >>> mt = MalletTagger()
+        >>> mt._get_features(u'University1')
```
kmike, line 84: this doctest will fail in Python 3.2

nltk/tag/mallet.py
```diff
@@ -0,0 +1,249 @@
...
+        :rtype : list (list (tuple(str,str))) 
+        '''
+        
+        # We need the list of sentences instead of the list generator for matching the input and output  
+        sentences = list(sents) 
+        
+        # First, build the test file 
+        input_file = tempfile.NamedTemporaryFile(
+                prefix='mallet_tagger.test',
+                dir=tempfile.gettempdir(),
```
kmike, line 131: dir argument doesn't look necessary here

nltk/tag/mallet.py
```diff
@@ -0,0 +1,249 @@
...
+                delete=False)
+
+        for sent in sentences: 
+            for token in sent:
+                data = unicode(token)
```
kmike, line 136: What is this line for? If the input is unicode then unicode(token) does nothing; if it is not unicode then the text is decoded using some system-depending encoding which is not good. And there is no "unicode" symbol in Python 3.x, so it doesn't work in Python 3.x

nltk/tag/mallet.py
```diff
@@ -0,0 +1,249 @@
...
+                input_file.write(self._get_features(data) + '\n')
```
kmike, line 137: Unicode handling is broken here: `self._get_features` returns unicode, but input_file is open in binary mode.

kmike:
Hey @longdt219,

Thanks for the contribution!

My main question is why are we going with Mallet for linear-chain CRFs? Some of the problems with Mallet:
1. Mallet is not a Python package, it should be installed and configured. For one, it makes it harder to write unit tests.
2. It seems that Mallet is slower than other linear-chain CRF implementations. For example, here http://www.chokkan.org/software/crfsuite/benchmark.html author of CRFSuite claims Mallet is up to 100x slower than CRFSuite; also, as implemented, the tagger starts Mallet binary each time `tag_sents` is called.

Another question is what is this wrapper for. Now it looks like an example of how to write a POS tagger using Mallet. Are users supposed to define their own features? If so, the I think this interface should be public, and the example POS tagger class should be clearly separated from the Mallet wrapper itself.

stevenbird:
Thanks for the detailed feedback @kmike. I see that you're behind the [Python interface to CRFSuite](https://pypi.python.org/pypi/python-crfsuite/). The only reason to support Mallet is if there is other functionality that people want (like topic modelling, [mentioned by @ewan-klein](https://github.com/nltk/nltk/issues/104#issuecomment-34634881)) or existing trained models that people want to use (for replicating experiments). However, it looks like [there hasn't been any interest shown in Mallet for NLTK in the past 18 months](https://www.google.com/webhp?#q=mallet+nltk), so I would be happy to use CRFSuite for CRF tagging instead.

Separately from this, we can always come back to using Mallet for topic modelling, after considering the alternatives.

Doesn't your objection 1 above apply equally to CRFSuite, @kmike?

nltk/tag/mallet.py
longdt219, line 29: If mallet paths are not found, it will look at the environment variable MALLET and MALLET_DEPS.

kmike, line 29: Aha, I see, thanks for an explanation.

kmike:
Hi @stevenbird,

Yes, https://github.com/tpeng/python-crfsuite is one alternative for linear-chain CRFs; others are https://github.com/adsva/python-wapiti and https://github.com/larsmans/seqlearn, and maybe there are some other packages. Both python-crfsuite and python-wapiti bundle CRFsuite and Wapiti libraries, so users get them just by running 'pip install python-crfsuite' or 'pip install libwapiti'. There is no need to install CRFsuite or Wapiti separately. SeqLearn doesn't bundle any external non-Python libraries, but it requires numpy and scikit-learn (also, it doesn't implement CRFs, but there are similar models in seqlearn, and a PR for CRF training).

All of these libraries are based on Python C/C++ extensions. It means that while they can be installed with pip, users must have a working build chain on their machines (a compiler, etc). I'm not sure if it is a problem; many other Python packages need a working build environment to be installed, and this environment is relatively easy to setup unless you're on Windows. 

As for the topic modelling, gensim is a popular choice in Python. It has lots of useful stuff built-in like LSI or word2vec implementation or doc2vec (Quoc Le and Tomas Mikolov. Distributed Representations of Sentences and Documents) implementation. And it also has a wrapper for Mallet LDA, so by using gensim we're getting a wrapper for Mallet topic modelling with the same interface as other gensim models.

stevenbird:
Thanks @kmike. I'm happy to presume that anyone doing substantial machine learning tasks has a functioning build environment. Given the simplicity of installing CRFSuite, and the better speed than Mallet, I think we should just go with CRFSuite.

To interface with gensim, I think we just need to provide a Matrix Market view on NLTK corpora. Would that do it?

longdt219:
Hi @stevenbird,
Should I write a wrapper for CRFSuite. I don't think it is necessary given that <a href='https://github.com/tpeng/python-crfsuite' > someone did it for  python  </a> and the <a href='http://nbviewer.ipython.org/github/tpeng/python-crfsuite/blob/master/examples/CoNLL%202002.ipynb'> usage </a> is fairy straight forward.

stevenbird:
@longdt219, the idea is to have an implementation of `TaggerI` using CRFSuite, so that someone can switch between different taggers just by a change of import statement.

stevenbird:
We've agreed to use CRFSuite instead. The main advantage of Mallet is configurability, and this is not exposed via the NLTK wrapper. We prefer the faster speed of CRFSuite.

PR #862, opened 2015-01-29
Attempt to fix the CI Errors related to senna and stanford 

longdt219:
Hi @stevenbird, 
I modified jenkins.sh to download and creat environment variables which is needed for correctly finding the .jar file. Hope this will fix the problems listed in CI server for 
- Senna package
- Stanford package

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -34,6 +34,21 @@ if [[ ! -d ${stanford_tagger_package_name} ]]; then
 	ln -s ${stanford_tagger_package_name} 'stanford-postagger'
 fi
 
+# Download SENNA 
+senna_file_name=$(curl -s 'http://ml.nec-labs.com/senna/download.html' | grep -o 'senna-v.*.tgz' | head -n1)
+senna_folder_name='senna'
+if [[ ! -d $senna_folder_name ]]; then
+        wget -nv "http://ml.nec-labs.com/senna/$senna_file_name"
+        tar -xvzf ${senna_file_name}
+        rm ${senna_file_name}       
+fi
+
+# Setup the Enviroment variable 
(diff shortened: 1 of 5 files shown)


stevenbird:
Thanks @longdt219. But I see you are reverting some changes in senna.py and stanford.py that involve the `u` prefix for strings, that is not appropriate for Python 3, or for Python 2 where we have `from __future__ import unicode_literals`. Is your PR based on the head revision?

longdt219:
Hi @stevenbird, 
I run doc test locally but it said that the test fail because of `u` prefix for string. That's why I modify the doc test. Should I revert them to the original ?

stevenbird:
There should be no `u` prefixes, but your patch is adding `u` prefixes in some places (e.g. `stanford.py`).

kmike:
@longdt219 how are you running tests locally? Running them just with `doctest.testmod(optionflags=doctest.NORMALIZE_WHITESPACE | doctest.ELLIPSIS)` may not work; there are issues with doctests & Python 2/3. See https://github.com/nltk/nltk/blob/develop/web/dev/local_testing.rst for more info. There is a nose plugin (https://github.com/nltk/nltk/blob/develop/nltk/test/doctest_nose_plugin.py) that makes tests work if the output only differs in 'u'.

longdt219:
Thank you @kmike. Your comments are very helpful.

longdt219:
Hi @stevenbird, 
I saw that you pulled jenkins.sh file. Could you please pull other files. i.e.  
- nltk/internals.py (so that if stanford package fail again I will have some clues) 
- nltk/tokenize/stanford.py (fix a bug shown in CI)

PR #868, opened 2015-02-02
Fix ci errors

longdt219:
Hi @stevenbird 
I add some script to fix bug senna on CI server. I changed the JDK on building machine to 1.8 but it didn't solve the problem. I put some scripts on jenkins.sh to output some diagnosis for Stanford package.


stevenbird:
Thanks @longdt219

PR #869, opened 2015-02-03
Fix ci errors

longdt219:
Modify the code to comply with recent changes in dependency graph. Remove what I added for stanford package dianosis. Fix bug wrong exception for nltk.classify.senna


stevenbird:
Thanks @longdt219

PR #871, opened 2015-02-04
Second Stab at fixing NgramModel

iliakur:
This is a response to #800. I foolishly merged in changes from the develop branch. For clarity, @stevenbird, could you pull the latest develop changes into the "model" branch, so that only my commits are visible? Sorry to ask, this situation's my fault.

The changes I'm proposing are as follows:
- cleaned up FreqDist and it's attributes
- ConditionalFreqDist now inherits from FreqDist, solving the issue of `override_N` mentioned #800; it is also designed in a way that most ProbDist classes can use **either**.
- ConditionalProbDist is no longer used; NgramModel directly calls LidstoneProbDist or any other ProbDist class. This makes it conceptually clearer.


stevenbird:
Thanks @Copper-Head!

rmalouf:
It looks like this completely changes the API for ConditionalFreqDist and breaks all kinds of things (including the tests and all the documentation).  Now to get a count for a condition/sample you need `cfd[(condition, sample)]` rather than the old `cfd[condition][sample]`.  Also, now there's no way to easily get a FreqDist given a condition or to initialize a CFD from a list of (condition, sample) pairs. Is this what was intended?  Am I missing something?   

(Just to clarify, I'm looking at the version in the develop branch.)

stevenbird:
Oh dear, I didn't see that. There was no justification for changing the API in this way, and we'd never roll out such a pervasive change in a minor release.

@Copper-Head would you please leave `ConditionalFreqDist` as a dictionary of `FreqDist`?

stevenbird:
@rmalouf, @Copper-Head – I've undone these changes, since they are too pervasive and were never agreed. I shouldn't have merged. I hope the work on fixing ngram model can continue, but without modifying existing APIs.

iliakur:
Gotcha. I'll put something up that's not as drastic a change!

stevenbird:
Thanks @Copper-Head

PR #874, opened 2015-02-05
Improve spanish stemmer

Madrigal:
This is a refactor of things that smell bad on the spanish stemmer. No functionality was affected.

I verified that it doesn't break anything with the test files in https://github.com/snowballstem/snowball-data/tree/master/spanish. It would be a good idea to add those to the unit tests for spanish stemmer, but I didn't found any reference on how the nltk project handles this kind of testing. If you guide me through it I'll be happy to help.

### DIFF
--- a/nltk/stem/snowball.py
+++ b/nltk/stem/snowball.py
@@ -3407,36 +3407,13 @@ class SpanishStemmer(_StandardStemmer):
             if not (word.endswith(suffix) and rv.endswith(suffix)):
                 continue
 
-            if rv[:-len(suffix)].endswith(("i\xE9ndo",
-                                           "\xE1ndo",
-                                           "\xE1r", "\xE9r",
-                                           "\xEDr")):
-                word = (word[:-len(suffix)].replace("\xE1", "a")
-                                           .replace("\xE9", "e")
-                                           .replace("\xED", "i"))
-                r1 = (r1[:-len(suffix)].replace("\xE1", "a")
-                                       .replace("\xE9", "e")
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @Madrigal.

stevenbird:
You would be welcome to add tests to the module docstring (those that are useful to a user), or put them in `nltk/test/stem.doctest` (this is more for regression testing, not user documentation).

Madrigal:
I was thinking more on something like adding it to `nltk/test/unit/test_stem.py`

However, I don't see any kind of exhaustive testing there. [Snowball data](https://github.com/snowballstem/snowball-data) contains the reference test cases for all implementation of snowball stemmer, and Snowball project is licensed under BSD, (see http://snowball.tartarus.org/license.php) so I guess we can distribute the data for testing with nltk. Let me know if that works for the project, and where should I put it if so.

stevenbird:
I would be happy to add the snowball test data to the NLTK data collection. Then you could access it via a corpus reader, and put your test code under `test/unit` as you suggest. Would that work?

Madrigal:
Yes, that would be really nice :)

stevenbird:
Sure. I've opened an issue for this, https://github.com/nltk/nltk_data/issues/21

stevenbird:
I've uploaded the snowball test data. You probably need to set up a corpus reader like `PPAttachmentCorpusReader`. We could refactor things to have a corpus reader that handles files consisting of whitespace-delimited fields, and which calls a supplied factory function on the fields.

Issue #840, opened 2015-01-16
pos_tag in NLTK 3


stevenbird:
The model used by `pos_tag` in NLTK 3 should be set to `taggers/maxent_treebank_pos_tagger/PY3/english.pickle`, cf https://groups.google.com/forum/#!topic/nltk-users/e4ULOXFAOV8

PR #880, opened 2015-02-09
Remove the path compatibility support in tag/__init__.py

ziyuang:
There would be a duplicated "PY3" in the path, since the PY2/PY3 compatibility has been handled by `py3_data` in `nltk.compat`.


stevenbird:
This is a proposal to revert https://github.com/nltk/nltk/commit/00a3a610715a81d2c6677e719b3470b2d11ab60b (cf #840). I think that was the wrong solution, and that instead we should modify `data.load` to use `compat.py3_data`.

PR #881, opened 2015-02-10
Modify to work with python3

longdt219:
Hi @stevenbird, 
I modified the code to work with python 3.

### DIFF
--- a/nltk/parse/transitionparser.py
+++ b/nltk/parse/transitionparser.py
@@ -6,6 +6,9 @@
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
 import tempfile
 import pickle
 import os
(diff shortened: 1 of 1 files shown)


nltk/parse/transitionparser.py
```diff
@@ -41,9 +42,7 @@ def __init__(self, dep_graph):
         """
         # dep_graph.nodes contain list of token for a sentence
         self.stack = [0]  # The root element
-        self.buffer = range(
-            1, len(
-                dep_graph.nodes))  # The rest is in the buffer
+        self.buffer = list(range(1, len(dep_graph.nodes)))  # The rest is in the buffer
```
kmike, line 45: I see you're using `.buffer.pop(0)` in the later code. For Python lists it is O(N) operation which is bad. If you want to add/remove items from both ends then probably collections.deque is more appropriate; if you want to add/remove items only from the beginning it could make sense to use a reversed buffer: popping an item from the end (`buffer.pop()`) is O(1).

longdt219, line 45: Thank you @kmike

stevenbird, line 45: Thanks @kmike

Issue #882, opened 2015-02-11
NLTK downloader bug


stevenbird:
```
$ sudo python -m nltk.downloader
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/sb/git/nltk/nltk/downloader.py", line 2266, in <module>
    halt_on_error=options.halt_on_error)
  File "/Users/sb/git/nltk/nltk/downloader.py", line 655, in download
    self._interactive_download()
  File "/Users/sb/git/nltk/nltk/downloader.py", line 967, in _interactive_download
    DownloaderGUI(self).mainloop()
  File "/Users/sb/git/nltk/nltk/downloader.py", line 1227, in __init__
    self._fill_table()
  File "/Users/sb/git/nltk/nltk/downloader.py", line 1523, in _fill_table
    items = self._ds.collections()
  File "/Users/sb/git/nltk/nltk/downloader.py", line 499, in collections
    self._update_index()
  File "/Users/sb/git/nltk/nltk/downloader.py", line 854, in _update_index
    packages[child.id] = child
AttributeError: 'str' object has no attribute 'id'
```

kaselis:
The same here, started this morning only. nltk version 2.0.4

boffbowsh:
Ditto, with 3.0.0 and 3.0.1 here

alexlouden:
+1 same here

Looks like collections contains `lang_id` which doesn't have a package definition:

```
  <collections>
    <collection id="all-corpora" name="All the corpora">
      ....
      <item ref="langid" />
```

Here's the culprit: https://github.com/nltk/nltk_data/blob/fa7aae1c5c5e26b3cc3217dae15005c471184819/collections/all-corpora.xml#L29

I'll make a pull request

stevenbird:
I dealt with [the spurious langid line](https://github.com/nltk/nltk_data/blob/fa7aae1c5c5e26b3cc3217dae15005c471184819/collections/all-corpora.xml#L29) in [this commit](https://github.com/nltk/nltk_data/commit/dba94eb6d16f8367f1b3631eeee73a4bffb94ba2) but the downloader is still broken.

[...]

PR #883, opened 2015-02-11
Handle items in collection that don't have packages

alexlouden:
This is to handle this situation: https://github.com/nltk/nltk/issues/882


stevenbird:
Good idea, thanks @alexlouden

PR #891, opened 2015-02-19
Adding _repr_svg_ for Ipython notebook

oplatek:
The _repr_svg is implemented using dot and graphviz.
Usage in IPython notebook:

from IPython.display import display
import nltk.parse.dependencygraph as depgraph
dg = depgraph.DependencyGraph(depgraph.conll_data1)
display(dg)

Should draw the dependency graph for the sentence

draw_dot is based on @gmonce code https://github.com/pln-fing-udelar/pln_inco/blob/master/syntax_trees.py#L52

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -129,7 +129,7 @@ class DependencyGraph(object):
         """
         return node_address in self.nodes
 
-    def draw_dot(self):
+    def to_dot(self):
         """
         Returns a dot representation suitable for using with Graphviz
         @rtype C{String}
@@ -139,13 +139,14 @@ class DependencyGraph(object):
         s += 'edge [dir=forward]\n'
         s += 'node [shape=plaintext]\n'
         # Draw the remaining nodes
-        for head, h_node in self.nodes.iteritems():
-            s += '\n%s [label="%s (%s)"]' % (h_node['address'], h_node['address'], h_node['word'])
-            if head != 0:  # not TOP
-                if h_node['rel'] != '_':
-                    s += '\n%s -> %s [label="%s"]' % (h_node['head'], h_node['address'], h_node['rel'])
-                else:
-                    s += '\n%s -> %s ' % (h_node['head'], h_node['address'])
+        for node in sorted(self.nodes.values()):
+            s += '\n%s [label="%s (%s)"]' % (node['address'], node['address'], node['word'])
+            for rel, deps in node['deps'].iteritems():
+                for dep in deps:
+                    if rel != None:
+                        s += '\n%s -> %s [label="%s"]' % (node['address'], dep, rel)
+                    else:
+                        s += '\n%s -> %s ' % (node['address'], dep)
         s += "\n}"
         return s
 
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -132,7 +132,6 @@ class DependencyGraph(object):
     def draw_dot(self):
         """
         Returns a dot representation suitable for using with Graphviz
-        @type t:L{nltk.parse.dependencygraph.DependencyGraph}
         @rtype C{String}
         """
         # Start the digraph specification
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -18,6 +18,7 @@ from __future__ import print_function, unicode_literals
 from collections import defaultdict
 from itertools import chain
 from pprint import pformat
+import subprocess
 
 from nltk.tree import Tree
 from nltk.compat import python_2_unicode_compatible, string_types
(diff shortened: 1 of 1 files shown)


gmonce:
Great! I see you did not implement _repr_dot, but used the original private draw_dot method. I guess it should be good to produce a "raw text" dot representation... am I wrong? (I'm just learning the nltk coding standards!)

oplatek:
@gmonce I have no clue what nltk codings standards are, I just started to using nltk for dependency parsing so I want to add this functionality:-) 
I did not named it _repr_dot_ because this syntax is used by IPython and it cannot handle dot files.

I guess somebody with nltk experience should look at this

gmonce:
I'm trying to test the changes (and maybe add something), but the DependencyGraph class seems to have changed, with nltk 3.0.2 I cannot run parsed_sents() from conll2007 
as before, nor I can create a new graph using the Malt-tab format (I get an 'EOF in multi-line string' from IPython). With nltk 3.0.0 I have no problems...

oplatek:
Note: That I got another error for 3.0.2 from develop branch of nltk.
I fix that by replacing "root" to ROOT when I was loading universal
dependencies
(http://universaldependencies.github.io/docs/)

I also needed to rename few variable from your code "nodelist" > "nodes"
etc.

On Fri, Feb 20, 2015 at 4:32 AM, Guillermo Moncecchi <
notifications@github.com> wrote:

> I'm trying to test the changes (and maybe add something), but the
> DependencyGraph class seems to have changed, with nltk 3.0.2 I cannot run
> parsed_sents() from conll2007
> as before, nor I can create a new graph using the Malt-tab format (I get
> an 'EOF in multi-line string' from IPython). With nltk 3.0.0 I have no
> problems...
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/nltk/nltk/pull/891#issuecomment-75184611.

## 

Ondřej Plátek

stevenbird:
@dimazest – would you please check if there's any residual problems from your refactoring of the DependencyGraph class, that are causing the above issues?

dimazest:
Sure.

Yes, `nodelist` is renamed to `nodes` because it's not a list anymore (it's a dictionary), note that it's not sorted anymore. Actually, would it make sense to have a property `nodelist` which returns sorted nodes.

Replacing 'root' to 'ROOT' is ok.

dimazest:
Actually, taking into account how many issues the change from `.nodelist` to `.node` caused I propose the following:
1. Conceptually, DependencyGraph is a mapping from `node id` to `node`, it should have:
   - `keys()` that iterates over node ids
   - `values()` that iterates over nodes
   - `items()` that iterates over (node_id, node)
2. Introduce `.nodelist` property with a deprecation warning saying that DependencyGraph.values() should be used instead.
3. Do the same for `.nodes`.

Clearly, it's outside of this PR, and I can implement this inside of a new PR.

nltk/parse/dependencygraph.py
```diff
@@ -128,6 +129,41 @@ def contains_address(self, node_address):
...
+        Returns a dot representation suitable for using with Graphviz
+        @type t:L{nltk.parse.dependencygraph.DependencyGraph}
+        @rtype C{String}
+        """
+        # Start the digraph specification
+        s = 'digraph G{\n'
+        s += 'edge [dir=forward]\n'
+        s += 'node [shape=plaintext]\n'
+        # Draw the remaining nodes
+        for head, h_node in self.nodes.iteritems():
```
dimazest: will `.iteritems()` work in Python3?

dimazest:
The `root`, `ROOT` relation is a bit of a pain because some code expects that it's possible to access the root node (from the parse point of view), so when a graph is created the `.root` is set assuming that the relation is `ROOT` https://github.com/nltk/nltk/blob/develop/nltk/parse/dependencygraph.py#L264

Maybe, the code should be smarter and figure out what this relation actually is.

gmonce:
After a bit of debugging, I found that the conll2007 corpus cannot be load because some sentences do not follow the condition cell_number==3 and head==0:

https://github.com/nltk/nltk/blob/f1c6b3074dfcc6c25fd3836b4dec85bf5b5f97a3/nltk/parse/dependencygraph.py#L253-256

I do not know how to handle these cases... @dimazest

dimazest:
Do you get a `DependencyGraphError` exception saying that "The graph does'n contain a node that depends on the root element"?

dimazest:
```
1       "       "       PUNT    PUNT_BEREIZ     _       0       PUNC    _       
```

Is a strange sentence, I'm not even sure how to hand it...

dimazest:
It seems that it's bad idea to assume that there is a fake root (with index 0) and a "root" node, which is depends on it with some relation, usually 'ROOT'.

I would remove the `DependencyGraph.root` attribute.

stevenbird:
Can we try to resolve this in the next day or so please? Then I can release it in 3.0.2.

gmonce:
I understand that the original pull request is ok. 

The problem loading conll2007 is different, related (that is what I understook from @dimazest comment, at least) to how DependencyGraphs are loaded...

stevenbird:

stevenbird:

stevenbird:
@gmonce thanks for the clarification; I agree this can be merged.

PR #892, opened 2015-02-20
Populating each DependencyGraph node(aka dictionary) with all the keys

oplatek:
Hi @dimazest, and @stevenbird,
Some functions in`nltk.parse.dependencygraph.DependencyGraph`
 e.g `nltk.to_conll` expects that all keys exists for each node.
// I think that this is reasonable assumption
However, most of the code does not specify all the keys.
E.g. the dependency graphs generated by `nltk.parse.dependencygraph.ProbabilisticNonprojectiveParser`
contain nodes with missing `lemma` key.

As a solution, I changed the default behaviour for initializing each node
with all the expected keys and corresponding default values.
See the `DependencyGraph.defaultnode` in the PR.

Thanks for your work!

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -57,6 +57,7 @@ class DependencyGraph(object):
                                            'feats': None,
                                            'head': None,
                                            'deps': defaultdict(list),
+                                           'rel': None,
                                            })
 
         self.nodes[0].update(
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -18,7 +18,6 @@ from __future__ import print_function, unicode_literals
 from collections import defaultdict
 from itertools import chain
 from pprint import pformat
-from copy import deepcopy
 
 from nltk.tree import Tree
 from nltk.compat import python_2_unicode_compatible, string_types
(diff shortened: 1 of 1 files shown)


nltk/parse/dependencygraph.py
```diff
@@ -49,14 +60,11 @@ def __init__(self, tree_str=None, cell_extractor=None, zero_based=False, cell_se
         are split by whitespace.
 
         """
-        self.nodes = defaultdict(lambda: {'deps': defaultdict(list)})
+        self.nodes = defaultdict(lambda: deepcopy(DependencyGraph.default_node))
```
dimazest: I would not use deepcopy (it feels like magic to me) here, instead I would just specify all the needed items inside of lambda, so

``` python
{'deps': defaultdict(list)}
```

becomes

``` python
{
    'deps': defaultdict(list)},
    'address': None,
    'word': None,
    ...
}
```

alternatively one could define a function that returns a default node and pass it to the default dict.

Ideally, I think, node should be an object, not a dictionary.

stevenbird:
I'm still concerned about the simplicity and  efficiency of our new dependency graph representation.

nltk/parse/dependencygraph.py
oplatek: Ok, I will rewrite it back to lambda function.

I thought that the property of  `default_dict` is nice self documentation, but I do not care.

dimazest:
I don't remember the details, but I did some profiling when I've started using DependencyGrpah.
I'll think of a representative performance test and report the results.

stevenbird:
Can we consider using `namedtuple` instead of `dict` here please?

dimazest:
Yes, that a good idea.

Additionally, we should move dependency information from node to `DependencyGraph.deps` (for example) which is a mapping from a node address to a list of id of dependents.

stevenbird:

stevenbird:
I'm ready to release 3.0.2. Is there anything we can do to resolve this so it can be included?

dimazest:
The request looks fine to be merged, given that there is code that expects some of the keys present.

Replacing the dict with a nametuple can be done in a next release as it most probably will break a lot of code and will require additional testing.

stevenbird:
OK, thanks @dimazest

PR #893, opened 2015-02-21
Added TextCat module

avitalp:
- TextCat now inherits 'object' for consistent style with the rest of NLTK.
- Added check so that NLTK can be loaded but only fails at runtime when the regex dependency isn't met.
- Trivial variable name correction.


stevenbird:
Thanks @abitalp

PR #896, opened 2015-02-23
Upgrade mallet #104

longdt219:
Hi @stevenbird, 

Here is the code for using CRFSuite for tagging to resolve the #104. 
I added the script to automatically download and build the necessary file. I also modified the wiki for using CRFSuite. 

Running on part of Spanish treebank (ConLL dataset) when training on 3000 sentences and testing on the ConLL Spanish test data achieved the accuracy of 93.6 %

### DIFF
--- a/nltk/tag/crfsuite.py
+++ b/nltk/tag/crfsuite.py
@@ -11,28 +11,17 @@ A module for interfacing with the CRFSuite taggers.
 """
 from __future__ import absolute_import
 from __future__ import unicode_literals
-from os import path, environ, sep
-import os
-from platform import system
-import tempfile
-from subprocess import PIPE
 import unicodedata
 import re 
-from subprocess import Popen
 from nltk.tag.api import TaggerI
+import pycrfsuite
 
-_crfsuite_url = 'http://www.chokkan.org/software/crfsuite'
+_py_crfsuite_url = 'https://pypi.python.org/pypi/python-crfsuite'
 
-class Error(Exception):
-    """Basic error handling class to be extended by the module specific
-    exceptions"""
-    
-class ExecutableNotFound(Error):
-    """Raised if the crfsuite executable does not exist"""
 
 class CRFTagger(TaggerI):
     """
-    An interface to CRFSuite taggers. http://www.chokkan.org/software/crfsuite/tutorial.html
+    An interface to Python CRFSuite taggers. https://pypi.python.org/pypi/python-crfsuite
     
     >>> from nltk.tag.crfsuite import CRFTagger
     >>> ct = CRFTagger()
@@ -58,31 +47,8 @@ class CRFTagger(TaggerI):
     
     def __init__(self, file_path=''):
                
-        self._path = path.normpath(file_path) + sep
-        exe_file_1 = self.executable(self._path)
-        # Verifies the existence of the executable on the self._path first    
-        if not path.isfile(exe_file_1):
-            # Check for the system environment 
-            if 'CRFSUITE' in environ:  
-                self._path = path.normpath(environ['CRFSUITE']) + sep 
-                exe_file_2 = self.executable(self._path)
-                if not path.isfile(exe_file_2):
-                    raise ExecutableNotFound("CRFSuite executable expected at %s or %s but not found" % (exe_file_1,exe_file_2))
-                
         self._model_file = ''
     
-    def executable(self, base_path):
-        """
-        The function that determines the system specific binary that should be
-        used in the pipeline. In case, the system is not known the default crfsuite binary will
-        be used.
-        """ 
-        os_name = system()
-        if os_name == 'Linux':
-            return path.join(base_path, 'crfsuite')
-        if os_name == 'Windows':
-            return path.join(base_path, 'crfsuite.exe')
-        return path.join(base_path, 'crfsuite')
 
     def set_model_file(self, model_file):
         self._model_file = model_file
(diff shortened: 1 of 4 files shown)


stevenbird:
I've pulled this into the `test` branch, to permit regression testing (https://nltk.ci.cloudbees.com/job/test/).

stevenbird:
@kmike do you have any feedback please?

kmike:
Well, I like python-crfsuite approach more. Here users are required to download & install two C++ libraries (crfsuite and liblbfgs), and then tagging is performed by running an external executable file (so a constant overhead is big). 

With python-crfsuite manual installation of C++ libraries is not needed, there is no need to create temporary files with input features, start external binaries and parse their output - python-crfsuite works like a regular Python library. It is also somewhat better tested - there is an existing testing suite (not big, 40-50 automated tests for various edge cases - still much better than a single test in this PR), and it was used in production already.

stevenbird:
Thanks for your feedback @kmike. I agree. @longdt219 would you please use python-crfsuite instead?

longdt219:
Hi @stevenbird , 

I modify the code so that it now work for Python CrfSuite. I also added the dependency to the pip-req.txt file. Trained  on 3000 English sentences and test on the official Conll English data file, the accuracy is 94.3

stevenbird:
Thanks @longdt219; I've added some stylistic comments.
@kmike does this look ok now?

longdt219:
Hi @stevenbird , 
I modify the code for it look more pythonic.

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
...
+        Tag a list of sentences. NB before using this function, user should specify the mode_file either by 
+                       - Train a new model using ``train'' function 
+                       - Use the pre-trained model which is set via ``set_model_file'' function  
+        :params sentences : list of sentences needed to tag. 
+        :type sentences : list(list(str))
+        :return : list of tagged sentences. 
+        :rtype : list (list (tuple(str,str))) 
+        '''
+        tagger = pycrfsuite.Tagger()
+        tagger.open(self._model_file)
```
kmike: The parser shouldn't load and realod the tagger each time tag_sents is called. It makes more sense to do it in a constructor.

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
...
+        # We need the list of sentences instead of the list generator for matching the input and output
+        result = []  
+        for sent in sents:
+            features = [self._get_features(data) for data in sent]
+            labels = tagger.tag(features)
+                
+            if len(labels) != len(sent):
+                raise Exception(' Predicted Length Not Matched, Expect Errors !')
+            
+            tagged_sent = [(sent[i],labels[i]) for i in range(len(labels))]
```
kmike: list(zip(sent, labels))

nltk/tag/crfsuite.py
stevenbird: Good point, and we have plenty of examples of this already in nltk

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
+# -*- coding: utf-8 -*-
+# Natural Language Toolkit: Interface to the CRFSuite Tagger
+#
+# Copyright (C) 2001-2015 NLTK Project
+# Author: Long Duong <longdt219@gmail.com>
+# URL: <http://nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""
+A module for interfacing with the CRFSuite taggers.
```
kmike: Currently it is a module for POS tagging using CRFsuite. 

There is no explicit documentation about how to use your own features (and `_get_features` method is private), and there is no way to set trainer parameters - regularization and other constants, training algorithms. If one wants to develop e.g. a NER system using this tagger then all of these is helpful: custom features are must, default constants should be tuned, and it is good to have more information about the training process (verbose=True parameter, usage of the holdout dataset, etc).

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
...
+"""
+from __future__ import absolute_import
+from __future__ import unicode_literals
+import unicodedata
+import re 
+from nltk.tag.api import TaggerI
+import pycrfsuite
+
+_py_crfsuite_url = 'https://pypi.python.org/pypi/python-crfsuite'
```
kmike: what is this constant for?

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
...
+    
+    def __init__(self, file_path=''):
+               
+        self._model_file = ''
+    
+
+    def set_model_file(self, model_file):
+        self._model_file = model_file
+            
+    def _get_features(self,data):
```
kmike: It could be more clear to name the argument "token".

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
...
+        :return : a string which contains the features   
+        
+        """ 
+        feature_list = []  
+        # Capitalization 
+        if data[0].isupper():
+            feature_list.append('CAPITALIZATION')
+        
+        # Number 
+        pattern = re.compile('\\d')
```
kmike: It makes no sense to compile a pattern every time we extract a feature. Actually, it should be faster if you just remove .compile call because there is an internal regexp cache in re module. 

`r'\d'` looks better than `'\\d'`.

nltk/tag/crfsuite.py
```diff
@@ -0,0 +1,162 @@
...
+        '''
+        X_train = []
+        y_train = [] 
+        
+        for sent in train_data:
+            features = [self._get_features(data) for data,label in sent]
+            labels = [label for data,label in sent]
+            
+            X_train.append(features)
+            y_train.append(labels)
```
kmike: I think this method can be made smaller if `trainer.append` is called as we go. 'features' is the same as 'xseq', 'labels' is the same as 'yseq', there is no reason to build and unroll X_train and y_train lists.

longdt219:
Hi @stevenbird  and @kmike, 

Thank you very much for your comments. I modified the code accordingly.

stevenbird:
Thanks @longdt219. I'm going to merge this, so it makes the 3.0.2 release. Any further issues can be dealt with later.

PR #918, opened 2015-03-14
* Language frequency distributions are now loaded on-demand rather

avitalp:
Language frequency distributions are now loaded on-demand rather than all at init time, as discussed in https://github.com/nltk/nltk/pull/845#issuecomment-80700112

### DIFF
--- a/nltk/corpus/reader/crubadan.py
+++ b/nltk/corpus/reader/crubadan.py
@@ -34,22 +34,21 @@ class CrubadanCorpusReader(CorpusReader):
     """
     
     _LANG_MAPPER_FILE = 'table.txt'
-    all_lang_freq = {}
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @avitalp. I think we can also drop `_load_all_ngrams`. There's no harm in someone calling `_load_lang_ngrams` for each language if they really need it, right?

avitalp:
No harm; in fact, `_load_all_ngrams` calls `_load_lang_ngrams` itself. So, if necessary, one would just get an exhaustive list via `langs` and loop over those with `_load_lang_ngrams` (or call `lang_freq`).

Submitted a PR with method removed.

stevenbird:
Thanks @avitalp

Issue #924, opened 2015-03-18
merge crubadan branch


stevenbird:
Cf #107, #845, #884, #890.

stevenbird:
Closed by https://github.com/nltk/nltk/commit/af059ef31ae79ceb673e3858ae5907c787a74875

PR #927, opened 2015-03-20
Changes to reflect all modifications discussed in nltk#924

avitalp:
@stevenbird I've taken the liberty of updating this module as well to reflect changes in https://github.com/nltk/nltk/issues/924. The `demo` method requires an arbitrary number of "sample_*.txt" files with some sample text in any of the supported languages, do I provide these or let the user create his/her own?

### DIFF
--- a/nltk/classify/textcat.py
+++ b/nltk/classify/textcat.py
@@ -171,23 +171,24 @@ class TextCat(object):
                     'srp':'Serbian',
                     'epo':'Esperanto'}
         
-        # Generate a sample text of the language
         for cur_lang in langs:
-          raw_sentences = udhr.sents(cur_lang)
-          rows = len(raw_sentences) - 1
-          cols = map(len, raw_sentences)
+            # Get raw data from UDHR corpus
+            raw_sentences = udhr.sents(cur_lang)
+            rows = len(raw_sentences) - 1
+            cols = map(len, raw_sentences)
 
-          sample = ''
+            sample = ''
           
-          for i in range(0, rows):
-            cur_sent = ''
-            for j in range(0, cols[i]):
-              cur_sent += ' ' + raw_sentences[i][j]
+            # Generate a sample text of the language
+            for i in range(0, rows):
+                cur_sent = ''
+                for j in range(0, cols[i]):
+                    cur_sent += ' ' + raw_sentences[i][j]
             
-            sample += cur_sent
+                sample += cur_sent
           
-          # Try to detect what it is
-          print('Language snippet: ' + sample[0:140] + '...')
-          guess = self.guess_language(sample.encode('utf8'))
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @avitalp. The demo could use literal strings for its test data (cf dependency graph demo). Or it could use the udhr2 corpus.

avitalp:
Thanks for the hint @stevenbird! I just committed a new demo method with no external dependency on sample text files, but using the udhr corpus.

stevenbird:
Thanks @avitalp. Could I ask you to please use 4-space indentation, following PEP-8 style?

avitalp:
@stevenbird oops! Changed my IDE indentation for another lang and it carried over when I typed up this method, fixed now in https://github.com/nltk/nltk/commit/0805e246e10c1039d38870cb28da271e1810c9d1

stevenbird:
Thanks @avitalp. I've just done another review and noticed that you define your own version of `trigrams` instead of using the built in one at https://github.com/nltk/nltk/blob/develop/nltk/util.py#L445. Would you mind using that version instead please, unless there's some particular reason not to?

nltk/classify/textcat.py
```diff
@@ -144,37 +152,43 @@ def guess_language(self, text):
...
+            raw_sentences = udhr.sents(cur_lang)
+            rows = len(raw_sentences) - 1
+            cols = map(len, raw_sentences)
+
+            sample = ''
+          
+            # Generate a sample text of the language
+            for i in range(0, rows):
+                cur_sent = ''
+                for j in range(0, cols[i]):
```
kmike, line 179: cols[i] won't work in Python 3 because map doesn't return list

nltk/classify/textcat.py
```diff
@@ -144,37 +152,43 @@ def guess_language(self, text):
...
+                    cur_sent += ' ' + raw_sentences[i][j]
+            
+                sample += cur_sent
+          
+            # Try to detect what it is
+            print('Language snippet: ' + sample[0:140] + '...')
+            guess = self.guess_language(sample.encode('utf8'))
```
kmike: Why is language detection working on bytestrings instead of unicode? It could make sense to work on bytes if there is also some encoding detection, or an ability to handle arbitrary bytes data, but this is not the case here - text is decoded from utf8 in `remove_punctuation` function.

avitalp:
@stevenbird & @kmike: I believe newest commit fixes both issues, if you could kindly take another look. Quick testing locally shows `demo()` running correctly under Python 2.7 and 3.2.

stevenbird:
Thanks @avitalp. I'm glad to see that the decoding is happening at load time. I'm going to merge this now and test it out in time for the 3.0.3 release.

PR #932, opened 2015-03-27
Fix example for lin scored_synonyms

chrisji:
- The demo is currently repeating previous example of displaying synonyms without scores.


stevenbird:
Thanks @chrisji

PR #933, opened 2015-03-30
Add support for Google App Engine

sadovnychyi:
We cannot use nor filesystem nor subprocesses under GAE, but by simple
overriding these methods we will allow it to work under GAE.

There's a lot of people who would like to use it under GAE, and with this PL it will be much easier to do.
Without this PL its not even possible to `import nltk` under GAE.

http://stackoverflow.com/questions/1286301/using-the-python-nltk-2-0b5-on-the-google-app-engine

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -73,7 +73,7 @@ path = []
 
 # User-specified locations:
 path += [d for d in os.environ.get('NLTK_DATA', str('')).split(os.pathsep) if d]
-if not os.environ.get('APPENGINE_RUNTIME') and os.path.expanduser('~/') != '~/':
+if 'APPENGINE_RUNTIME' not in os.environ and os.path.expanduser('~/') != '~/':
     path.append(os.path.expanduser(str('~/nltk_data')))
 
 if sys.platform.startswith('win'):
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -73,7 +73,7 @@ path = []
 
 # User-specified locations:
 path += [d for d in os.environ.get('NLTK_DATA', str('')).split(os.pathsep) if d]
-if os.path.expanduser('~/') != '~/':
+if not os.environ.get('APPENGINE_RUNTIME') and os.path.expanduser('~/') != '~/':
     path.append(os.path.expanduser(str('~/nltk_data')))
 
 if sys.platform.startswith('win'):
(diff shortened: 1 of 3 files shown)


nltk/data.py
```diff
@@ -73,7 +73,7 @@
 
 # User-specified locations:
 path += [d for d in os.environ.get('NLTK_DATA', str('')).split(os.pathsep) if d]
-if os.path.expanduser('~/') != '~/':
+if not os.environ.get('APPENGINE_RUNTIME') and os.path.expanduser('~/') != '~/':
```
dimazest: maybe `'APPENGINE_RUNTIME' in os.environ` is a clearer version.

However, if it's set to an empty string, behavior will be different.

nltk/downloader.py
```diff
@@ -924,6 +924,10 @@ def default_download_dir(self):
         permission: ``/usr/share/nltk_data``, ``/usr/local/share/nltk_data``,
         ``/usr/lib/nltk_data``, ``/usr/local/lib/nltk_data``, ``~/nltk_data``.
         """
+        # Check if we are on GAE where we cannot write into filesystem.
+        if os.environ.get('APPENGINE_RUNTIME') is not None:
```
dimazest: `'APPENGINE_RUNTIME' in os.environ`

sadovnychyi: Just fixed both of them

dimazest:
Looks good to me. Thanks. @stevenbird could you have a look, please.

sadovnychyi:
Any updates on this? @dimazest @stevenbird

dimazest:
It looks good to me.

However, I thought a bit more about it, and it might be better to throw an exception instead of printing to stdout (because, in principle, you might not have it available).

sadovnychyi:
You're talking about this part right?

``` python
# Override missing methods on environments where it cannot be used like GAE.
import subprocess
if not hasattr(subprocess, 'PIPE'):
  subprocess.PIPE = lambda *args, **kwargs: print(
    'subprocess.PIPE is not supported.')
if not hasattr(subprocess, 'Popen'):
  subprocess.Popen = lambda *args, **kwargs: print(
    'subprocess.Popen is not supported.')
```

Well if we will throw an exception here it will fail to load on GAE.

We may just do not output anything to stdout and maybe with additional checking if we are under GAE environment. (but it seems useless as we are patching it only if the attribute is not available)

Ideas?

dimazest:
I meant something like:

``` python
def fake_PIPE(*args, **kwargs):
    raise NotImplementedError('subprocess.PIPE is not supported.')

if not hasattr(subprocess, 'PIPE'):
    subprocess.PIPE = fake_PIPE

```

so when somebody uses it under GAE an exception is thrown

sadovnychyi:
Yeah, you were right -- as its being only **imported**, but not actually **called** -- we should raise an exception there.

Fixed!

stevenbird:
Thanks @sadovnychyi – this looks very good.

PR #938, opened 2015-04-05
Add warning for when SimpleGoodTuring doesn't work

smilli:
An attempt to address #602.  I added a warning for when SimpleGoodTuring doesn't find a proper best-fit line.  In order for the Good-Turing estimated frequency to be less than the actual seen frequency, the slope of the best-fit line must be less than -1.

From [Good-Turing Smoothing Without Tears](http://www.d.umn.edu/~tpederse/Courses/CS8761-FALL02/Code/sgt-gale.pdf):
"Thus if the estimate for the slope, b, is greater than -1, the method presented should be regarded as not
applicable."

The warning is triggered for the example in #602, but doesn't appear when there is enough data (for example using a frequency distribution made out of brown.words()).

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -1272,6 +1272,11 @@ class SimpleGoodTuringProbDist(ProbDistI):
             xy_cov += (x - x_mean) * (y - y_mean)
             x_var += (x - x_mean)**2
         self._slope = (xy_cov / x_var if x_var != 0 else 0.0)
+        if self._slope >= -1:
+            warnings.warn("SimpleGoodTuring did not find a proper best fit "
+                          "line for smoothing probabilities of occurrences. "
+                          "The probabilities are likely to be unreliable "
+                          "estimates.")
         self._intercept = y_mean - self._slope * x_mean
 
     def _switch(self, r, nr):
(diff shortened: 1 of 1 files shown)


iliakur:
@smilli do you think you know how to fix SimpleGoodTuring instead?

smilli:
I don't see anything wrong with SimpleGoodTuring?

I agree with @kmike:

> Both Good Turing and Simple Good Turing probability estimation methods assign Nr(1)/N probability mass to unknown events, and when all events are single this mass would be 1.0, leaving 0.0 for seen events. So I think it is a pathological case for SGT probability estimation method and you shouldn't use SGT for such data.

I did add another warning for when `bins` isn't passed in because of #383 though.

iliakur:
ah, sorry about the confusion. I hadn't properly read #602 and got the mistaken impression there was something _wrong_ with SGT. Now that I understood what's up, :+1: from me :)

stevenbird:
Thanks @smilli. I've gone with the first warning but not the second; I don't see the reason for giving this warning only to proceed with a value for the bins parameter that's likely to be a bad choice. Instead, I'd be happy to add the warning to the docstring, or even to require an explicit bins parameter.

PR #940, opened 2015-04-06
Fix param name in KneserNey

smilli:
The param name was documented incorrectly in KneserNeyProbDist.


stevenbird:
Thanks @smilli

Issue #943, opened 2015-04-08
Stabalize MaltParser once and for all


alvations:
MaltParser API in NLTK has been breaking on several occasions:
- http://stackoverflow.com/questions/14009330/how-to-use-malt-parser-in-python-nltk
- http://stackoverflow.com/questions/20091698/malt-parser-throwing-class-not-found-exception
- http://stackoverflow.com/questions/21815891/dependency-parser-using-nltk-and-maltparser
- http://stackoverflow.com/questions/29513187/maltparser-not-working-in-python-nltk

I'm somehow using the interface and it's breaking for me too, so let's try to fix MaltParser once and for all. Shouldn't be more than a day's of work @ https://github.com/alvations/nltk/blob/patch-1/nltk/parse/malt.py

alvations:
A few things we need to forgo:
- the `working_dir` parameter isn't very useful, especially when it's not necessary to run MaltParser from command line. 
- finding malt parser binaries shouldn't be part of the API in `/usr/local/bin`, it should be like how stanford tagger API works in `nltk.tag.stanford`, let the user give the path to the maltparser and we do a quick `find_jar`.
- command creation should not be using `-jar` should be using `--classpath` so as to satisfy other dependencies that maltparser wants.

alvations:
Let's start with allowing only maltparser-1.8 to work then work out how to fit all possible malts.

alvations:
Current, the patched malt.py works for parsing with pre-trianed models: https://github.com/alvations/nltk/blob/patch-1/nltk/parse/malt.py

We need to rename some of the functions because it's getting a little confusing:
 - `tagged_parse_sents`: parse multiple sentences with POS tagged.
 - `tagged_parse` : parse one sentence with POS tagged
 - `parse_sents`: parse multiple sentence without POS tags...

Any suggestions for the name refactoring?

alvations:
Moved to #944

PR #944, opened 2015-04-09
Stabilized MaltParser API

alvations:
From #943, 

MaltParser was requiring all sorts of weird `os.environ` to make it find the binary and then call jar file with environment java classpath.
- The new API requires only where the user saves his/her installed version of maltparser and finds the jar files using `os.walk` and uses full classpath and `org.maltparser.Malt` to call Maltparser instead of -jar
- Also the `generate_malt_command` makes updating the API to suit Maltparser easier.

I've tried with `Maltparser-1.7.2` and `Maltparser-1.8`


alvations:
However there remain problems with `DependencyGraph` and how it reads the maltparser output files. 

Pre-trained models from http://www.maltparser.org/mco/mco.html  outputs uncased chunk labels, e.g. nsubj, null, dobj, poss:

```
1    I    _    PRP    PRP    _    2    nsubj    _    _
2    shot    _    VBD    VBD    _    0    null    _    _
3    an    _    DT    DT    _    4    det    _    _
4    elephant    _    NN    NN    _    2    dobj    _    _
5    in    _    IN    IN    _    2    prep    _    _
6    my    _    PRP$    PRP$    _    7    poss    _    _
7    pajamas    _    NN    NN    _    5    pobj    _    _
```

But DependencyChart is expecting nice chunk tags, e.g. ROOT, SUBJ, SPEC, OBJ. E.g.

```
1    John    _    NNP   _    _    2    SUBJ    _    _
2    sees    _    VB    _    _    0    ROOT    _    _
3    a       _    DT    _    _    4    SPEC    _    _
4    dog     _    NN    _    _    2    OBJ     _    _
```

The demo is fine with we parse using a trained model from NLTK. So the awkward `find_binary` and NLTK's job to call MaltParser to retrieve the output is seamless.

But there's still problem when reading the parses from a pre-trained model in NLTK:

```
from nltk.parse import malt
from nltk import word_tokenize, sent_tokenize

indir = '/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/'
modelfilepath = '/home/alvas/engmalt.linear-1.7.mco'
maltParser = malt.MaltParser(path_to_maltparser=indir, model=modelfilepath)

sentences = [word_tokenize(sent) for sent in sent_tokenize('I shot an elephant in my pajamas. This is a foobar sentence')]
tagged_sentences = [nltk.pos_tag(sentence) for sentence in sentences]

maltParser.tagged_parse_sents(tagged_sentences)
```

[out]:

```
  File "/home/alvas/git/nltk/test_malt.py", line 15, in <module>
    print maltParser.tagged_parse_sents(tagged_sentences)
  File "/home/alvas/git/nltk/newermalt.py", line 132, in tagged_parse_sents
    DependencyGraph.load(output_file.name))
  File "/home/alvas/git/nltk/nltk/parse/dependencygraph.py", line 156, in load
    for tree_str in infile.read().split('\n\n')
  File "/home/alvas/git/nltk/nltk/parse/dependencygraph.py", line 72, in __init__
    cell_separator=cell_separator,
  File "/home/alvas/git/nltk/nltk/parse/dependencygraph.py", line 260, in _parse
    "The graph does'n contain a node "
nltk.parse.dependencygraph.DependencyGraphError: The graph does'n contain a node that depends on the root element.
```

Although, there was an outputfile created from MaltParser if we add `print output_file.read()` before https://github.com/alvations/nltk/blob/develop/nltk/parse/malt.py#L158

`output_file.read()` prints:

```
1   I   _   PRP PRP _   2   nsubj   _   _
2   shot    _   VBD VBD _   0   null    _   _
3   an  _   DT  DT  _   4   det _   _
4   elephant    _   NN  NN  _   2   dobj    _   _
5   in  _   IN  IN  _   2   prep    _   _
6   my  _   PRP$    PRP$    _   7   poss    _   _
7   pajamas _   NN  NN  _   5   pobj    _   _
8   .   _   .   .   _   2   punct   _   _

1   This    _   DT  DT  _   5   nsubj   _   _
2   is  _   VBZ VBZ _   5   cop _   _
3   a   _   DT  DT  _   5   det _   _
4   foobar  _   NN  NN  _   5   nn  _   _
5   sentence    _   NN  NN  _   0   null    _
```

alvations:
@dhgarrette , @kmike, @heatherleaf , @stevenbird . 

Any idea why the pre-trained model outputs is unreadable by `DependencyChart.load()`? 
Or are there some secret options in maltparser that can make it readable to `DependencyChart`?

I'll leave this as it is now and let someone else deal with the dependency parses. I'll go back to the `translate`, `model` and `align` packages =)

Santosh-Gupta:
Thanks Alvations!!

I was wondering if you could give an example of how to use it in python.

alvations:
@Santosh-Gupta , the `demo()` shows how you can train a parser and then use it. But loading the pre-trained model is still messy because of DependencyChart objects

stevenbird:
@alvations, that error message was introduced in https://github.com/nltk/nltk/commit/e0f0630155eec7e0df5b907ee523ad7ed8a04f5b#diff-31ba76604fcce0dbd82cdfd1dba4233d.

@dimazest it looks like this change gets in the way of loading pre-trained models. Are you able to investigate please?

stevenbird:
Just pinging you again @dimazest

dimazest:
Sorry, I somehow missed the first mention, I'll have a look to this right now...

stevenbird:
@dimazest thanks for the PR. @alvations, are you able to load pre-trained models now?

alvations:
Sorry for the late reply. @dimazest  thanks for the fix!! @stevenbird, now the malt API works with pre-trained model. 

I'm not sure why it only works with `malt.MaltParser.parse_one(sentence)`: 

```
_path_to_maltparser = '/home/alvas/maltparser-1.8/dist/maltparser-1.8/'
_path_to_model= '/home/alvas/engmalt.linear-1.7.mco'     
>>> mp = MaltParser(path_to_maltparser=_path_to_maltparser, model=_path_to_model)
>>> sent = 'I shot an elephant in my pajamas'.split()
>>> print(mp.parse_one(sent).tree())
(pajamas (shot I) an elephant in my)
```

But when i tried to do `malt.MaltParser.parse_sents(sentences)` for multiple sentence, it didn't return me an iterable of `DependencyGraph` but a `listiterator`:

```
_path_to_maltparser = '/home/alvas/maltparser-1.8/dist/maltparser-1.8/'
_path_to_model= '/home/alvas/engmalt.linear-1.7.mco'     
>>> mp = MaltParser(path_to_maltparser=_path_to_maltparser, model=_path_to_model)
>>> sent = 'I shot an elephant in my pajamas'.split()
>>> sent2 = 'Time flies like banana'.split()
>>> print(mp.parse_one(sent).tree())
(pajamas (shot I) an elephant in my)
>>> print(next(mp.parse_sents([sent,sent2])))
<listiterator object at 0x7f0a2e4d3d90> 
>>> print(next(next(mp.parse_sents([sent,sent2]))))
[{u'address': 0,
  u'ctag': u'TOP',
  u'deps': [2],
  u'feats': None,
  u'lemma': None,
  u'rel': u'TOP',
  u'tag': u'TOP',
  u'word': None},
 {u'address': 1,
  u'ctag': u'NN',
  u'deps': [],
  u'feats': u'_',
  u'head': 2,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NN',
  u'word': u'I'},
 {u'address': 2,
  u'ctag': u'NN',
  u'deps': [1, 11],
  u'feats': u'_',
  u'head': 0,
  u'lemma': u'_',
  u'rel': u'null',
  u'tag': u'NN',
  u'word': u'shot'},
 {u'address': 3,
  u'ctag': u'AT',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'AT',
  u'word': u'an'},
 {u'address': 4,
  u'ctag': u'NN',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NN',
  u'word': u'elephant'},
 {u'address': 5,
  u'ctag': u'NN',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NN',
  u'word': u'in'},
 {u'address': 6,
  u'ctag': u'NN',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NN',
  u'word': u'my'},
 {u'address': 7,
  u'ctag': u'NNS',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NNS',
  u'word': u'pajamas'},
 {u'address': 8,
  u'ctag': u'NN',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NN',
  u'word': u'Time'},
 {u'address': 9,
  u'ctag': u'NNS',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NNS',
  u'word': u'flies'},
 {u'address': 10,
  u'ctag': u'NN',
  u'deps': [],
  u'feats': u'_',
  u'head': 11,
  u'lemma': u'_',
  u'rel': u'nn',
  u'tag': u'NN',
  u'word': u'like'},
 {u'address': 11,
  u'ctag': u'NN',
  u'deps': [3, 4, 5, 6, 7, 8, 9, 10],
  u'feats': u'_',
  u'head': 2,
  u'lemma': u'_',
  u'rel': u'dep',
  u'tag': u'NN',
  u'word': u'banana'}]
```

alvations:
With help from http://goo.gl/TpW1iY, I manage to get a tree from `parse_sents()` by calling `print(next(next(mp.parse_sents([sent,sent2]))).tree())`.  Somehow the `parse_sents()` looks to be broken, it was combining two sentences into one instead of parsing them separately.

```
    # Initialize a MaltParser object with a pre-trained model.
    mp = MaltParser(path_to_maltparser=path_to_maltparser, model=path_to_model) 
    sent = 'I shot an elephant in my pajamas'.split()
    sent2 = 'Time flies like banana'.split()
    # Parse a single sentence.
    print(mp.parse_one(sent).tree())
    print(next(next(mp.parse_sents([sent,sent2]))).tree())
```

[out]:

```
(pajamas (shot I) an elephant in my)
(shot I (banana an elephant in my pajamas Time flies like))
```

alvations:
@dimazest @stevenbird: Fixed at last, now we can easily malt any sentences with the API. And i'll be able to use this for tree2string models in `nltk.translate`

stevenbird:
Thanks @alvations and @dimazest.

If either of you has time it would be nice to include a doctest with little demonstration in the docstring for the MaltParser class, cf: https://github.com/nltk/nltk/blob/develop/nltk/tag/stanford.py#L120

alvations:
Sorry for the late reply, I was out for 1 month. 

@dimazest, previously the outputs were converting into trees but now it's getting strange. After parsing with the pre-trained model, I get the following output file:

```
1   I   _   NN  NN  _   2   nn  _   _
2   shot    _   NN  NN  _   0   null    _   _
3   an  _   AT  AT  _   2   dep _   _
4   elephant    _   NN  NN  _   7   nn  _   _
5   in  _   NN  NN  _   7   nn  _   _
6   my  _   NN  NN  _   7   nn  _   _
7   pajamas _   NNS NNS _   3   dobj    _   _
```

Somehow the POS tagger outputs gets erroneous and tags everything as NN...

And when I read it into a `nltk.parse.DependencyGraph`, I couldn't convert it into a tree.

```
>>> from nltk.parse import DependencyGraph
>>> dg_str = """1       I       _       NN      NN      _       2       nn      _       _
... 2   shot    _       NN      NN      _       0       null    _       _
... 3   an      _       AT      AT      _       2       dep     _       _
... 4   elephant        _       NN      NN      _       7       nn      _       _
... 5   in      _       NN      NN      _       7       nn      _       _
... 6   my      _       NN      NN      _       7       nn      _       _
... 7   pajamas _       NNS     NNS     _       3       dobj    _       _
... """
>>> 
>>> DependencyGraph(dg_str)
/usr/local/lib/python2.7/dist-packages/nltk/parse/dependencygraph.py:307: UserWarning: The graph doesn't contain a node that depends on the root element.
  "The graph doesn't contain a node "
<DependencyGraph with 8 nodes>
>>> dg = DependencyGraph(dg_str)
>>> dg.tree()
  File "/usr/local/lib/python2.7/dist-packages/nltk/parse/dependencygraph.py", line 340, in tree
    word = node['word']
TypeError: 'NoneType' object has no attribute '__getitem__'
>>> dg
<DependencyGraph with 8 nodes>
>>> print dg
defaultdict(<function <lambda> at 0x7f87f0b46cf8>, {0: {u'ctag': u'TOP', u'head': None, u'word': None, u'deps': defaultdict(<type 'list'>, {u'null': [2], u'ROOT': []}), u'lemma': None, u'tag': u'TOP', u'rel': u'TOP', u'address': 0, u'feats': None}, 1: {u'ctag': u'NN', u'head': 2, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'NN', u'address': 1, u'word': u'I', u'lemma': u'_', u'rel': u'nn', u'feats': u'_'}, 2: {u'ctag': u'NN', u'head': 0, u'deps': defaultdict(<type 'list'>, {u'dep': [3], u'nn': [1]}), u'tag': u'NN', u'address': 2, u'word': u'shot', u'lemma': u'_', u'rel': u'null', u'feats': u'_'}, 3: {u'ctag': u'AT', u'head': 2, u'deps': defaultdict(<type 'list'>, {u'dobj': [7]}), u'tag': u'AT', u'address': 3, u'word': u'an', u'lemma': u'_', u'rel': u'dep', u'feats': u'_'}, 4: {u'ctag': u'NN', u'head': 7, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'NN', u'address': 4, u'word': u'elephant', u'lemma': u'_', u'rel': u'nn', u'feats': u'_'}, 5: {u'ctag': u'NN', u'head': 7, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'NN', u'address': 5, u'word': u'in', u'lemma': u'_', u'rel': u'nn', u'feats': u'_'}, 6: {u'ctag': u'NN', u'head': 7, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'NN', u'address': 6, u'word': u'my', u'lemma': u'_', u'rel': u'nn', u'feats': u'_'}, 7: {u'ctag': u'NNS', u'head': 3, u'deps': defaultdict(<type 'list'>, {u'nn': [4, 5, 6]}), u'tag': u'NNS', u'address': 7, u'word': u'pajamas', u'lemma': u'_', u'rel': u'dobj', u'feats': u'_'}})
```

alvations:
Given that the POS tagger outputs the correct POS:

Using `nltk.pos_tag` MaltParser outputs this file:

```
1   I   _   PRP PRP _   2   nsubj   _   _
2   shot    _   VBD VBD _   0   null    _   _
3   an  _   DT  DT  _   4   det _   _
4   elephant    _   NN  NN  _   2   dobj    _   _
5   in  _   IN  IN  _   2   prep    _   _
6   my  _   PRP$    PRP$    _   7   poss    _   _
7   pajamas _   NN  NN  _   5   pobj    _   _
8   .   _   .   .   _   2   punct   _   _
```

And the more "powerful" regex tagger does outputs this:

```
1   I   _   PRP PRP _   2   nsubj   _   _
2   shot    _   NN  NN  _   0   null    _   _
3   an  _   DT  DT  _   4   det _   _
4   elephant    _   NN  NN  _   2   dobj    _   _
5   in  _   IN  IN  _   4   prep    _   _
6   my  _   NN  NN  _   7   nn  _   _
7   pajamas _   NNS NNS _   5   pobj    _   _
8   .   _   .   .   _   2   punct   _   _
```

The DependencyGraph is still not accepting the fact that there is no word for the `TOP` node:

```
>>> from nltk.parse import DependencyGraph
>>> dg_str = """1       I       _       PRP     PRP     _       2       nsubj   _       _
... 2   shot    _       VBD     VBD     _       0       null    _       _
... 3   an      _       DT      DT      _       4       det     _       _
... 4   elephant        _       NN      NN      _       2       dobj    _       _
... 5   in      _       IN      IN      _       2       prep    _       _
... 6   my      _       PRP$    PRP$    _       7       poss    _       _
... 7   pajamas _       NN      NN      _       5       pobj    _       _
... 8   .       _       .       .       _       2       punct   _       _
... """
>>> 
>>> dg = DependencyGraph(dg_str)
/usr/local/lib/python2.7/dist-packages/nltk/parse/dependencygraph.py:307: UserWarning: The graph doesn't contain a node that depends on the root element.
  "The graph doesn't contain a node "
>>> dg 
<DependencyGraph with 9 nodes>
>>> print dg
defaultdict(<function <lambda> at 0x7f5bbc323c80>, {0: {u'ctag': u'TOP', u'head': None, u'word': None, u'deps': defaultdict(<type 'list'>, {u'null': [2], u'ROOT': []}), u'lemma': None, u'tag': u'TOP', u'rel': u'TOP', u'address': 0, u'feats': None}, 1: {u'ctag': u'PRP', u'head': 2, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'PRP', u'address': 1, u'word': u'I', u'lemma': u'_', u'rel': u'nsubj', u'feats': u'_'}, 2: {u'ctag': u'VBD', u'head': 0, u'deps': defaultdict(<type 'list'>, {u'punct': [8], u'dobj': [4], u'prep': [5], u'nsubj': [1]}), u'tag': u'VBD', u'address': 2, u'word': u'shot', u'lemma': u'_', u'rel': u'null', u'feats': u'_'}, 3: {u'ctag': u'DT', u'head': 4, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'DT', u'address': 3, u'word': u'an', u'lemma': u'_', u'rel': u'det', u'feats': u'_'}, 4: {u'ctag': u'NN', u'head': 2, u'deps': defaultdict(<type 'list'>, {u'det': [3]}), u'tag': u'NN', u'address': 4, u'word': u'elephant', u'lemma': u'_', u'rel': u'dobj', u'feats': u'_'}, 5: {u'ctag': u'IN', u'head': 2, u'deps': defaultdict(<type 'list'>, {u'pobj': [7]}), u'tag': u'IN', u'address': 5, u'word': u'in', u'lemma': u'_', u'rel': u'prep', u'feats': u'_'}, 6: {u'ctag': u'PRP$', u'head': 7, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'PRP$', u'address': 6, u'word': u'my', u'lemma': u'_', u'rel': u'poss', u'feats': u'_'}, 7: {u'ctag': u'NN', u'head': 5, u'deps': defaultdict(<type 'list'>, {u'poss': [6]}), u'tag': u'NN', u'address': 7, u'word': u'pajamas', u'lemma': u'_', u'rel': u'pobj', u'feats': u'_'}, 8: {u'ctag': u'.', u'head': 2, u'deps': defaultdict(<type 'list'>, {}), u'tag': u'.', u'address': 8, u'word': u'.', u'lemma': u'_', u'rel': u'punct', u'feats': u'_'}})
>>> dg.tree()
  File "/usr/local/lib/python2.7/dist-packages/nltk/parse/dependencygraph.py", line 340, in tree
    word = node['word']
TypeError: 'NoneType' object has no attribute '__getitem__'
```

But if I manually substituted the `null` from the MaltParser output to `TOP`, it sort of work:

```
>>> from nltk.parse import DependencyGraph
>>> dg_str = """1   I   _   PRP PRP _   2   nsubj   _   _
... 2   shot    _   VBD VBD _   0   ROOT    _   _
... 3   an  _   DT  DT  _   4   det _   _
... 4   elephant    _   NN  NN  _   2   dobj    _   _
... 5   in  _   IN  IN  _   2   prep    _   _
... 6   my  _   PRP$    PRP$    _   7   poss    _   _
... 7   pajamas _   NN  NN  _   5   pobj    _   _
... 8   .   _   .   .   _   2   punct   _   _"""
>>> dg = DependencyGraph(dg_str)
>>> dg
<DependencyGraph with 9 nodes>
>>> dg.tree()
Tree('shot', ['I', Tree('elephant', ['an']), Tree('in', [Tree('pajamas', ['my'])]), '.'])
>>> print dg.tree()
(shot I (elephant an) (in (pajamas my)) .)
```

@dimazest , How could we fix that?

stevenbird:
@alvations, @dimazest: I just confirmed that this works fine in NLTK 2.0.4. Perhaps we should go back to that version of the dependency parser code?

alvations:
@stevenbird, do you mean using the `DependencyGraph` code from NLTK 2.0.4 with this patched MaltParser? If so, I'm not sure whether it breaks other things. Maybe @dimazest has a better idea about that.

stevenbird:
@alvations: I mean, reverting to the old version of DependencyGraph.

dimazest:
I'll have a closer look to this. It seems that we should be able to fix the bug, though, it might be difficult to figure out the label of the root node (`TOP`, `null`, or something else).

dimazest:
Is it standard Malt behavior to label the root relation as `null`?

alvations:
The malt output looks erroneous to me too. 

Given the input file `/tmp/malt_input.conll6PGXhB`:

```
1   I   _   PRP PRP _   0   a   _   _
2   shot    _   NN  NN  _   0   a   _   _
3   an  _   DT  DT  _   0   a   _   _
4   elephant    _   NN  NN  _   0   a   _   _
5   in  _   IN  IN  _   0   a   _   _
6   my  _   PRP$    PRP$    _   0   a   _   _
7   pajamas _   NNS NNS _   0   a   _   _
8   .   _   .   .   _   0   a   _   _
```

And output goes to `/tmp/malt_output.conllu1jwnT` with the command running on the `/home/alvas/` directory:

```
java -cp /home/alvas/maltparser-1.7.2/maltparser-1.7.2.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/maltparser-1.7.2.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/lib/log4j.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/lib/libsvm.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/lib/liblinear-1.8.jar:/home/alvas/maltparser-1.7.2/lib/log4j.jar:/home/alvas/maltparser-1.7.2/lib/libsvm.jar:/home/alvas/maltparser-1.7.2/lib/liblinear-1.8.jar org.maltparser.Malt -c engmalt.linear-1.7.mco -i /tmp/malt_input.conll6PGXhB -o /tmp/malt_output.conllu1jwnT -m parse
```

Given that maltparser is saved in `/home/alvas/maltparser-1.7.2` and the model `engmalt.linear-1.7.mco` save in the '/home/alvas/engmalt.linear-1.7.mco`

It outputs:

```
1   I   _   PRP PRP _   2   nsubj   _   _
2   shot    _   NN  NN  _   0   null    _   _
3   an  _   DT  DT  _   4   det _   _
4   elephant    _   NN  NN  _   2   dobj    _   _
5   in  _   IN  IN  _   2   prep    _   _
6   my  _   PRP$    PRP$    _   7   poss    _   _
7   pajamas _   NNS NNS _   5   pobj    _   _
8   .   _   .   .   _   2   punct   _   _
```

I'm not sure why malt does that.

---

Even if I use `pos_tag` and a VBD pos appears in the tagged sentence, the output still gets `null` at node 0:

```
1   I   _   PRP PRP _   2   nsubj   _   _
2   shot    _   VBD VBD _   0   null    _   _
3   an  _   DT  DT  _   4   det _   _
4   elephant    _   NN  NN  _   2   dobj    _   _
5   in  _   IN  IN  _   2   prep    _   _
6   my  _   PRP$    PRP$    _   7   poss    _   _
7   pajamas _   NN  NN  _   5   pobj    _   _
8   .   _   .   .   _   2   punct   _   _

```

dimazest:
I'm trying right now to make it possible to parametrize the top relation, so it's possible to tell what the label is when the dependency graph is constructed.

I've almost got the code working...

dimazest:
I got an idea. In https://github.com/nltk/nltk/blob/develop/nltk/parse/dependencygraph.py#L302 we depend on the fact that the parse root node depends on the graph fake root node via the `ROOT` dependency. This is problematic because the relation is not always labeled as `ROOT`, sometimes it's `null`.

To find the parse root node, we can rely on the fact, that there should be only one node that depends on the graph fake root node, so `self.nodes[0]['deps']` should contain only one relation pointing to one node.

However, it might be the case that there are several nodes that depend only on graph root node (in case of an incorrect parse, for example). Moreover, DG code depends on this assumption, see for example, https://github.com/nltk/nltk/blob/develop/nltk/parse/dependencygraph.py#L338

I propose to identify the root node in a more relaxed way: just take the node in `self.nodes[0]['deps'][...]` which in most cases is correct. However, if we got a graph with several sentence root nodes, then we can't reliably say what the root node is and should throw an exception. Thus, DG.root should be a property, so if a graph with two root nodes is being build, nothing fails until the moment when a single root node is expected.

It's a bit trickier, since sometime the root code is set outside of DG, for example here https://github.com/nltk/nltk/blob/develop/nltk/sem/lfg.py#L53 I suspect that in this particular case it will be unnecessary to set the root node explicitly. There are other places in code, where I'm not sure that figuring out the root node will work, I need to check.

alvations:
Maybe there's a way to force a root in MaltParser.  @jnivre, would that be possible? https://groups.google.com/forum/?hl=en#!topic/nltk-dev/bA-DQAZSf6s

If we confirm that MaltParser never gives more than one 0th indices, then we can force all `self.nodes[0]` to `ROOT` relation.

dimazest:
I would not force things, as it might be the case that some corpus will have strange parse. I would prefer to be flexible because DependencyGraph is a general datastructure.

alvations:
@dimazest , if we're training a new model that has ROOT, the malt parser has no problems returning dependency chart with `ROOT` and the `nltk.parse.malt.demo()` works for training and parsing with trained model.. 

It seems like the `null` issue only happens when loading the pre-trained model. So forcing the 0th indices when it gives `null`, might only be used for certain `.mco` model that the MaltParser team provides.

dimazest:
Let see what malt people say.

alvations:
@dimazest Given @jnivre reply, i think it's best that the `nltk.parse.malt` handles the 0 indices with "null" and convert them to 'ROOT' before instantiating the malt outputs as DependencyGraphs.

I'll do the following minor fixes and the `malt.py` should be good to merge:
- Use the MALT_PARSER environment variable here, cf https://github.com/nltk/nltk/wiki/Installing-Third-Party-Software?
- Add a new environment variable MALT_PARSER_MODELS, cf STANFORD_MODELS in https://github.com/nltk/nltk/wiki/Installing-Third-Party-Software
- Move demo code inside an if **name** == '**main**' block.
- Include a doctest with little demonstration in the docstring for the MaltParser class, cf: https://github.com/nltk/nltk/blob/develop/nltk/tag/stanford.py#L120

jnivre:
If you only want to use the pre-trained model for English that we provide on the MaltParser website, this will work fine. But if you also want to allow people to use models for other languages (for example, Czech), you have to allow any label to occur with HEAD=0, and simply converting to ROOT will in this case destroy information.

dimazest:
Alternatively, we can pass the label of the root dependency explicitly. Please have a look to https://github.com/nltk/nltk/pull/1049/files

Though, we still need to decide the label, which is not as nice as guessing, but as a first solution it might fine.

alvations:
I'm trying out the french `.mco` from http://www.maltparser.org/mco/french_parser/fremalt.html and it's not looking good. 

@jnivre Converting all index 0 to root is indeed a bad idea. But converting `0\tnull` in `0\tROOT` would only affect the module when using the pre-trained English `.mco` from http://www.maltparser.org/mco/english_parser/engmalt.html. Once the DependencyGraph object allows parameter for TOP label specification, the "hack" should disappear =)

Other than deciding a label for the top node, I've been trying out with the french pre-trained model and it somehow gives multiple `0\troot` outputs. 

Given input:

```
1   Nous    _   CLS CLS _   0   a   _   _
2   prions  _   V   V   _   0   a   _   _
3   les _   DET DET _   0   a   _   _
4   cineastes   _   NC  NC  _   0   a   _   _
5   et  _   CC  CC  _   0   a   _   _
6   tous    _   ADJ ADJ _   0   a   _   _
7   nos _   DET DET _   0   a   _   _
8   lecteurs    _   NC  NC  _   0   a   _   _
9   de  _   P   P   _   0   a   _   _
10  bien    _   ADV ADV _   0   a   _   _
11  vouloir _   VINF    VINF    _   0   a   _   _
12  nous    _   CLS CLS _   0   a   _   _
13  en  _   CLO CLO _   0   a   _   _
14  excuser _   VINF    VINF    _   0   a   _   _
15  .   _   PUNCT   PUNCT   _   0   a   _   _
```

And command:

```
java -cp /home/alvas/maltparser-1.7.2/maltparser-1.7.2.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/maltparser-1.7.2.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/lib/log4j.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/lib/libsvm.jar:/home/alvas/maltparser-1.7.2/dist/maltparser-1.7.2/lib/liblinear-1.8.jar:/home/alvas/maltparser-1.7.2/lib/log4j.jar:/home/alvas/maltparser-1.7.2/lib/libsvm.jar:/home/alvas/maltparser-1.7.2/lib/liblinear-1.8.jar org.maltparser.Malt -c fremalt-1.7.mco -i /tmp/malt_input.conll19gl5b -o /tmp/malt_output.conllB1di4J -m parse
```

Malt using `fremalt-1.7.mco`, it outputs this:

```
1   Nous    _   CLS CLS _   2   suj _   _
2   prions  _   V   V   _   0   root    _   _
3   les _   DET DET _   2   dep _   _
4   cineastes   _   NC  NC  _   2   mod _   _
5   et  _   CC  CC  _   2   dep _   _
6   tous    _   ADJ ADJ _   2   dep _   _
7   nos _   DET DET _   2   mod _   _
8   lecteurs    _   NC  NC  _   2   mod _   _
9   de  _   P   P   _   2   de_obj  _   _
10  bien    _   ADV ADV _   0   root    _   _
11  vouloir _   VINF    VINF    _   10  dep _   _
12  nous    _   CLS CLS _   10  dep _   _
13  en  _   CLO CLO _   10  dep _   _
14  excuser _   VINF    VINF    _   0   root    _   _
15  .   _   PUNCT   PUNCT   _   14  mod _   _
```

I'm not sure what to do with them but with the `pretrained_model_sanity_checks` hacks. The DependencyGraphs accepts it but the tree looks terrible:

```
(prions Nous les cineastes et tous nos lecteurs de)
```

alvations:
@dimazest Variable TOP dependency label in DependencyGraph sounds good.

For now, the pre-trained models from http://www.maltparser.org/mco/mco.html:
- `engmalt-linear-1.7.mco`  uses `null`
- `fremalt-1.7.mco` uses `root`
- `swemalt-1.7.2.mco` uses `ROOT`
- `espmalt-1.0.mco` seems to use `ROOT` 

@jnivre other than the pre-trained models from the maltparser site, are there other downloadable models that we can use to test on the Malt API in NLTK?

alvations:
@stevenbird I couldn't find something like `find_directory` in `nltk.internals` or maybe i've missed something. 

If not, could I add a function to `find_directory` given a name like that of `nltk.internals.find_file()`? 

It's because it's easier to use the default path of where the MaltParser tarball saves the .jar dependencies than to find multiple .jars from various locations to satisfy the MaltParser call. 

Currently, the MaltParser command generation is dependent on `malt.find_jars_within_path` which needs to know where the user's extracted the MaltParser directory is located.

dimazest:
The French example has 2 roots, so just one root node is picked up and the tree looks bad.

alvations:
I would think the problem lies in the pre-trained model so we can't do much about that.

The example sentence is hard to parse too, with `prions - to urge` and `excuser - to be excused`.  I think it's making errors on "de bien vouloir" too. And the parser assigning same index for the roots is also wrong. 

If the parser assigns different indices to the root at different level, the DependencyGraph gives a better output:

```
>>> from nltk.parse import DependencyGraph
>>> dg_str = """1   Nous    _   CLS CLS _   2   suj _   _
... 2   prions  _   V   V   _   0   ROOT    _   _
... 3   les _   DET DET _   2   dep _   _
... 4   cineastes   _   NC  NC  _   2   mod _   _
... 5   et  _   CC  CC  _   2   dep _   _
... 6   tous    _   ADJ ADJ _   2   dep _   _
... 7   nos _   DET DET _   2   mod _   _
... 8   lecteurs    _   NC  NC  _   2   mod _   _
... 9   de  _   P   P   _   2   de_obj  _   _
... 10  bien    _   ADV ADV _   3   ROOT    _   _
... 11  vouloir _   VINF    VINF    _   10  dep _   _
... 12  nous    _   CLS CLS _   10  dep _   _
... 13  en  _   CLO CLO _   10  dep _   _
... 14  excuser _   VINF    VINF    _   1   ROOT    _   _
... 15  .   _   PUNCT   PUNCT   _   14  mod _   _"""
>>> dg = DependencyGraph(dg_str)
>>> dg.tree()
Tree('prions', [Tree('Nous', [Tree('excuser', ['.'])]), Tree('les', [Tree('bien', ['vouloir', 'nous', 'en'])]), 'cineastes', 'et', 'tous', 'nos', 'lecteurs', 'de'])
>>> print dg.tree()
(prions
  (Nous (excuser .))
  (les (bien vouloir nous en))
  cineastes
  et
  tous
  nos
  lecteurs
  de)
```

jnivre:
Sorry for butting in, but it seems to me that a fundamental problem is that you assume that the dependency graph should be a tree when excluding the artificial root node. In other words, you assume that exactly one word in the sentence should dominate everything else. This assumption is NOT made by all dependency treebanks and dependency parsers. In general, it is only assumed that the dependency graph is a tree rooted at the artificial root node indexed 0. Therefore, there is nothing anomalous about having several words with HEAD=0, because there is still a unique root of the graph, namely the artificial root node. It is true that some treebanks and parsers in addition assume that the artificial root node will have a single child, but this is not true in general. So as long as you make the more restrictive assumption, you are going to run into problems when dealing with arbitrary treebanks and parsing models. Just trying to help.

stevenbird:
Thanks @jnivre – I'm grateful for your input.
@alvations: please go ahead and add `find_dir` to `nltk.internals`. If `find_file` can share any code with `find_dir` that would be a plus.

dimazest:
@jnivre thank you for your comment. I think refactoring dependency graph to support several nodes with HEAD=0 is outside of this PR and should be done separately.

I'll have a look what can be done, but untested code might get broken... By the way, do you think it's a god idea that the trees are printed with the artificial root node, currently we don't print is because of the assumption that there is only one word that depends on the root node.

alvations:
@jnivre thanks for the precious input, for a moment I was too focused on getting trees out of the graphs =)

@stevenbird I'll work on the find_jars and the environment variables and that should wrap up this PR.

@dimazest, agreed. Malt outputs with multiple roots to DependencyGraph object in NLTK would be another PR.

alvations:
@stevenbird I think it's ready for a code review before pull-merge. I hope I didn't miss out anything

@dimazest Once the DependencyGraph object and other parser object in NLTK improves, we can remove the hacks in `nltk.parse.malt.MaltParser.pretrained_model_sanity_checks()`.

@jnivre I hope this python API in NLTK helps to ease python coder when using MaltParser. Thank you for your precious input once again. If you spot anything wrong/missing, feel free to give us some  pointers.

nltk/internals.py
```diff
@@ -429,7 +430,7 @@ def find_file_iter(filename, env_vars=(), searchpath=(),
     if isinstance(env_vars, compat.string_types):
         env_vars = env_vars.split()
     yielded = False
-
+
```
dimazest: Please remove a whitespace.

nltk/internals.py
```diff
@@ -452,9 +453,14 @@ def find_file_iter(filename, env_vars=(), searchpath=(),
             yielded = True
             yield path_to_file
 
+
```
dimazest: This new line is unnecessary.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+	6    .    _    .    .    _    0    a    _    _
+
+	1    Is    _    VBZ    VBZ    _    0    a    _    _
+	2    that    _    IN    IN    _    0    a    _    _
+	3    right    _    JJ    JJ    _    0    a    _    _
+	4    ?    _    .    .    _    0    a    _    _
+	"""
+	for sentence in sentences:
+		for (i, (word, tag)) in enumerate(sentence, start=1):
+			input_str = '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' \
```
dimazest: Maybe

``` python
input_str = '\t'.join(i, word, '_', tag, tag, '_', '0', 'a', '_', '_', '\n')  # I like this more
```

or

``` python
input_str = '{i}\t{word}\t_\t{tag}\t{tag}\t_\t0\ta\t_\t_\n'.format(i=i, word=word, tag=tag)
```

is nicer.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
-    maltParser.train([dg1,dg2], verbose=verbose)
-
-    maltParser.parse_one(['John','sees','Mary'], verbose=verbose).tree().pprint()
-    maltParser.parse_one(['a','man','runs'], verbose=verbose).tree().pprint()
-    
-    next(maltParser.tagged_parse([('John','NNP'),('sees','VB'),('Mary','NNP')], verbose)).tree().pprint()
-    
+def taggedsent_to_conll(sentences):
+	"""
+	A module to convert the a POS tagged document stream
```
dimazest: It's a function, strictly speaking.

It's related to https://github.com/nltk/nltk/blob/develop/nltk/parse/dependencygraph.py#L434 It might worth making this function more general. A sentence contains of tuples of different length, so a correct template is picket up. Then the DependencyGraph.to_connl() uses this function.

nltk/internals.py
```diff
@@ -411,7 +412,7 @@ def get(self):
 ##########################################################################
 
 def find_file_iter(filename, env_vars=(), searchpath=(),
-        file_names=None, url=None, verbose=True):
+        file_names=None, url=None, verbose=True, finding_dir=False):
```
dimazest: What do you use to indent the code? It looks that use tabs, please use 4 spaces.

nltk/internals.py
```diff
@@ -478,6 +483,7 @@ def find_file_iter(filename, env_vars=(), searchpath=(),
                         yielded = True
                         yield path_to_file
 
+
```
dimazest: Please remove this newline

nltk/internals.py
```diff
@@ -523,6 +529,12 @@ def find_file(filename, env_vars=(), searchpath=(),
     return next(find_file_iter(filename, env_vars, searchpath,
                                file_names, url, verbose))
 
+
+def find_dir(filename, env_vars=(), searchpath=(),
+        file_names=None, url=None, verbose=True):
+    return next(find_file_iter(filename, env_vars, searchpath,
+                               file_names, url, verbose, finding_dir=True))
+
```
dimazest, line 537: Please add a new line, as between the top level functions there should be 2 newlines.

nltk/internals.py
```diff
@@ -661,6 +673,11 @@ def find_jar(name_pattern, path_to_jar=None, env_vars=(),
         searchpath=(), url=None, verbose=True, is_regex=False):
     return next(find_jar_iter(name_pattern, path_to_jar, env_vars,
                          searchpath, url, verbose, is_regex))
+
```
dimazest: Please add a new line.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+	3    right    _    JJ    JJ    _    0    a    _    _
+	4    ?    _    .    .    _    0    a    _    _
+	"""
+	for sentence in sentences:
+		for (i, (word, tag)) in enumerate(sentence, start=1):
+			input_str = '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' \
+			% (i, word, '_', tag, tag, '_', '0', 'a', '_', '_')
+			yield input_str
+		yield '\n\n'
+
```
dimazest: Please add a newline.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+	malt_dependencies = ['','','']
+	_malt_jars = set(find_jars_within_path(_malt_dir))
+	_jars = set(jar.rpartition('/')[2] for jar in _malt_jars)
+	malt_dependencies = set([u'log4j.jar', u'libsvm.jar', u'liblinear-1.8.jar'])
+					
+	assert malt_dependencies.issubset(_jars)
+	assert any(filter(lambda i: i.startswith('maltparser-') and \
+								i.endswith('.jar'), _jars))
+	return list(_malt_jars)
+
```
dimazest, line 77: Please add a new line.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+	A module to find pre-trained MaltParser model.
+	"""
+	if model_filename == None:
+		return 'malt_temp.mco'
+	elif os.path.exists(model_filename): # If a full path is given.
+		return model_filename
+	else: # Try to find path to malt model in environment variables.
+		return find_file(model_filename, env_vars=('MALT_MODEL',), 
+						verbose=False)
+
```
dimazest, line 91: Please add a newline.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+        >>> mp = malt.MaltParser('maltparser-1.7.2', 'engmalt.linear-1.7.mco') # doctest: +SKIP
+        >>> mp.parse_one('I shot an elephant in my pajamas .'.split()).tree() # doctest: +SKIP
+        (shot I (elephant an) (in (pajamas my)) .)
+        >>> # Without MALT_PARSER and MALT_MODEL environment.
+        >>> mp = malt.MaltParser('/home/alvas/maltparser-1.7.2/', '/home/alvas/engmalt.linear-1.7.mco') # doctest: +SKIP
+        >>> mp.parse_one('I shot an elephant in my pajamas .'.split()).tree() # doctest: +SKIP
+        (shot I (elephant an) (in (pajamas my)) .)
+	"""
+	def __init__(self, parser_dirname, model_filename=None, tagger=None, 
+				 additional_java_args=[]):
```
dimazest: Please make `additional_java_args=None` and add this

``` python
if additional_java_args is None:
    additional_java_args = []
```

as having mutable default parameters might lead to obscure bugs.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		one can use when calling Maltparser, usually this is the heapsize 
+		limits, e.g. `additional_java_args=['-Xmx1024m']`
+		(see http://goo.gl/mpDBvQ)
+		:type additional_java_args: list
+		"""
+		
+		# Find all the necessary jar files for MaltParser.
+		self.malt_jars = find_maltparser(parser_dirname)
+		# Initialize additional java arguments.
+		self.additional_java_args = additional_java_args
```
dimazest: or

``` python
self.additional_java_args = additional_java_args if additional_java_args is not None else []
```

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		# Initialize model.
+		self.model = find_malt_model(model_filename)
+		self._trained = False if self.model == 'malt_temp.mco' else True
```
dimazest: could this be replaced to

``` python
self._trained = self.model != 'malt_temp.mco'
```

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		# Set the working_dir parameters i.e. `-w` from MaltParser's option.
+		self.working_dir = tempfile.gettempdir() 
+		# Initialize POS tagger.
+		if tagger is not None:
```
dimazest: since we use a lot of inline ifs, what about

``` python
self.tagger = tagger if tagger is not None else malt_regex_tagger()
```

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		"""
+		Performs sanity checks and replace oddities in pre-trained model
+		outputs from http://www.maltparser.org/mco/english_parser/engmalt.html
+		
+		:param tree_str: The CONLL output file for a single parse
+		:type tree_str: str
+		:return: str
+		"""
+		# Checks for oddities in English pre-trained model.
+		if '\t0\tnull\t' in tree_str and \
```
dimazest: I would rewrite this to avoid `\`

``` python
if (
    '\t0\tnull\t' in tree_str
     and self.model.endswith(('engmalt.linear-1.7.mco', 'engmalt.poly-1.7.mco'))
):
```

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		if tagger is not None:
+			self.tagger = tagger
+		else:
+			self.tagger = malt_regex_tagger()	
+
+	def pretrained_model_sanity_checks(self, tree_str):
+		"""
+		Performs sanity checks and replace oddities in pre-trained model
+		outputs from http://www.maltparser.org/mco/english_parser/engmalt.html
+
```
dimazest: Maybe add a note what has to be done, so this hacks can go away.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		:return: str
+		"""
+		# Checks for oddities in English pre-trained model.
+		if '\t0\tnull\t' in tree_str and \
+		(self.model.endswith('engmalt.linear-1.7.mco') or
+		self.model.endswith('engmalt.poly-1.7.mco')
+		):
+			tree_str = tree_str.replace('\t0\tnull\t','\t0\tROOT\t')
+		# Checks for oddities in French pre-trained model.
+		if '\t0\troot\t' in tree_str and \
```
dimazest, line 164: Would this fit to one line?

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+			# missing .jars or strange -w working_dir problem.
+			_current_path = os.getcwd() # Remembers the current path.
+			try: # Change to modelfile path
+				os.chdir(os.path.split(self.model)[0]) 
+			except:
+				pass
+			ret = self._execute(cmd, verbose) # Run command.
+			os.chdir(_current_path) # Change back to current path.
+
+			if ret != 0:
```
dimazest: Just `if ret:`

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		Use MaltParser to parse multiple sentences. 
+		Takes a list of sentences, where each sentence is a list of words.
+		Each sentence will be automatically tagged with this 
+		MaltParser instance's tagger.
+
+		:param sentences: Input sentences to parse
+		:type sentence: list(list(str))
+		:return: iter(DependencyGraph)
+		"""
+		tagged_sentences = [self.tagger(sentence) for sentence in sentences]
```
dimazest: could it be an iterator:

``` python
tagged_sentences = (self.tagger(sentence) for sentence in sentences)
```

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		parsed_sentences = self.parse_tagged_sents(tagged_sentences, verbose)
```
dimazest: maybe just return this

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		if os.path.exists(self.model): # when parsing
+			cmd+= ['-c', os.path.split(self.model)[-1]] 
+		else: # when learning
+			cmd+= ['-c', self.model]
+
+		cmd+= ['-i', inputfilename]
+		if mode == 'parse':
+			cmd+= ['-o', outputfilename]
+		cmd+= ['-m', mode] # mode use to generate parses.
+		#print(" ".join(cmd))
```
dimazest: Please remove.

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		:param depgraphs: list of ``DependencyGraph`` objects for training input data
+		:type depgraphs: DependencyGraph
+		"""
+		input_file = tempfile.NamedTemporaryFile(prefix='malt_train.conll',
+				                                 dir=self.working_dir, mode='w',
+				                                 delete=False)
+		try:
+			input_str = ('\n'.join(dg.to_conll(10) for dg in depgraphs))
+			input_file.write(text_type(input_str))
+			input_file.close()
```
dimazest: should this go to the finally clause?

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		return p.wait()
+
+	def train(self, depgraphs, verbose=False):
+		"""
+		Train MaltParser from a list of ``DependencyGraph`` objects
+
+		:param depgraphs: list of ``DependencyGraph`` objects for training input data
+		:type depgraphs: DependencyGraph
+		"""
+		input_file = tempfile.NamedTemporaryFile(prefix='malt_train.conll',
```
dimazest: Try to use with statement:

``` python
with tempfile.NamedTemporaryFile(...) as input_file:
    input_str = ('\n'.join(dg.to_conll(10) for dg in depgraphs))
    input_file.write(text_type(input_str))
    input_file.flush()  # You should be able to flush the contents of the file to disc, closing it seems to be unnecessary.
    self.train_from_file(input_file.name, verbose=verbose)
```

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		"""
+		Train MaltParser from a file
+		:param conll_file: str for the filename of the training input data
+		:type conll_file: str
+		"""
+
+		# If conll_file is a ZipFilePathPointer, 
+		# then we need to do some extra massaging
+		if isinstance(conll_file, ZipFilePathPointer):
+			input_file = tempfile.NamedTemporaryFile(prefix='malt_train.conll',
```
dimazest: Again i would use context manager instead of `try ... finally`

nltk/parse/malt.py
```diff
@@ -5,245 +6,492 @@
...
+		if ret != 0:
+			raise Exception("MaltParser training (%s) "
+				            "failed with exit code %d" %
+				            (' '.join(cmd), ret))
+		self._trained = True
+
+	
 if __name__ == '__main__':
-    demo()
+	'''
```
dimazest, line 318: Could you write a (maybe shorter) version of this demo as a doctest.

nltk/parse/malt.py
alvations: In this case, it's `if ret is not 0:` right? 

When:

```
cmd = os.system(...)
self._execute(cmd, verbose)
```

`self._execute()` returns 1 or 0 integer  instead of boolean.

dimazest: 0 is equivalent to false, here you check whether ret is not zero.

``` python
>>> if 0: print('0')
...
>>> if 1: print('1')
...
1
>>> if not 0: print ('not 0')
...
not 0
>>> if not 1: print('not 1')
...
```

so it should be `if ret`.

nltk/internals.py
alvations, line 537: I'm not sure why git didn't detect the changes but it's done, see https://github.com/alvations/nltk/blob/patch-1/nltk/internals.py#L533

dimazest, line 537: Great!

alvations:
@dimazest , @stevenbird , all done =)

nltk/parse/malt.py
alvations: Can this be in `nltk.parse.util` instead? It's should be a general function that one can use even after `nltk.pos_tag` or stemming or lemmatization.

dimazest: Yes, sounds good to me.

alvations:
@dimazest , @stevenbird It's all patched up.

stevenbird:
Thanks @dimazest for the code review, and @alvations for all this work. It's looking good to me, so I'm going to merge.

PR #951, opened 2015-04-17
date limit in streaming requests

lrnzcig:
initial implementation of parameter for streaming requests that stops
when a date of the tweets stream is reached, regardless of the tweet
count

To be considered/enhanced:
- date is now a python date parameter, should probably accept a string
  instead, what format?
- limit in number of counts is still mandatory, consider if it is
  necessary

### DIFF
--- a/nltk/twitter/api.py
+++ b/nltk/twitter/api.py
@@ -9,13 +9,14 @@
 """
 Provides an interface for TweetHandlers.
 """
+import pytz
 
 class TweetHandlerI(object):
     """
     Interface class whose subclasses should implement a handle method that
     Twitter clients can delegate to.
     """
-    def __init__(self, limit=20):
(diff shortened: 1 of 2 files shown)


lrnzcig:
Thanks @ewan-klein 

I've modified line 342 as per your suggestion, it makes it indeed more clear.

As for the `repeat` parameter, indeed I had not understood it properly. Therefore I think it makes sense to leave it as it is -actually it would not make sense _not_ to specify `limit`.

Finally, I know it is my own suggestion, but the more I think about it the less I'm convinced about the text parameter for `date_limit`. At the end, it is not asking that much to the user to create a python date; it is one line of code and it can be documented via example. Anyhow, if we decided to do it, the code would indeed be better as you suggest.

PR #952, opened 2015-04-17
tagset mapping support for treebank_chunk, ChunkedCorpusReader

japerk:
Modified tagstr2tree to take optional source_tagset & target_tagset arguments


stevenbird:
@japerk in our existing use of `map_tag`, the source tagset is implicit (we know it from the corpus) and the target tagset is the only one specified, using the `tagset` argument. Can you see any way to work within this restriction here? E.g. picking up the source tagset from `self._tagset`?

japerk:
@stevenbird I did that for nltk.corpus.treebank_chunk & ChunkedCorpusReader, but when looking at the internals of ChunkedCorpusReader, it first creates parse trees, then selectively modifies the trees to produce tagged_words or chunked_sents. I thought the simplest implementation would be to get the mapped tags into the tree first, instead of modifying the tree after the fact, so that's why I modified tagstr2tree to take the source & target tagsets.

stevenbird:
@japerk I'm still trying to get my head around this, and would appreciate more feedback. I agree that you need both parameters in `tagstr2tree`. It makes sense that the various `ChunkedCorpusReader` methods pass their `tagset` argument down to `ChunkedCorpusView`.

However, you never specify a `source_tagset` when calling `ChunkedCorpusView`, even though its `read_block` method needs it in order to do any tag mapping. So how would the `source_tagset` ever be set?

Ideally, the source tagset would be known from context (and available in the `_tagset` member variable), and `ChunkedCorpusView` would only need to know the target tagset, which could be specified in a parameter just called `tagset`.

Note that there's a loosely related issue in nltk/nltk_book#110.

japerk:
@stevenbird In ChunkedCorpusReader.**init**, I added the tagset keyword argument. This is then added to the self._cv_args tuple, which is used in other methods when constructing the ChunkedCorpusView. That's how ChunkedCorpusView gets the source_tagset keyword argument. I could have made an explicit self._tagset variable, but since self._cv_args was already there & being used, it seemed easier to add the tagset to self._cv_args, instead of adding the keyword argument to every call to ChunkedCorpusView.

stevenbird:
@japerk sorry this is taking me a while to deal with...

stevenbird:
Sorry @japerk – I'll try to deal with this before 3.1.

stevenbird:
Thanks @japerk – it's taken me too long to get my head around this!

PR #954, opened 2015-04-20
Fix bug shown in jenkin server ...

longdt219:
Hi @stevenbird, 
Simple fix for errors shown in Jenskin related to transition parser.


stevenbird:
Thanks @longdt219

PR #956, opened 2015-04-21
compatibility with python 2.7

lrnzcig:
use pytz to obtain UTC timezone in python <> 3, standard datetime in python 3


lrnzcig:
Please advice on the implementation of the local timezone avoiding the use of non-standard packages pytz and tzlocal

Issue #931, opened 2015-03-25
nltk_tgrep


stevenbird:
@wroberts would you be open to the idea of incorporating [nltk_tgrep](https://github.com/wroberts/nltk_tgrep) into NLTK? (I learnt about nltk_tgrep from [this post on nltk-users](https://groups.google.com/forum/#!topic/nltk-users/iyCelNfsZ0U)).

wroberts:
Hi Dr. Bird,

Thanks for your message!  I'd be happy to see `nltk_tgrep` go into NLTK.  It seems to me that `nltk_tgrep` fits well into the NLTK project, and should be added to NLTK if possible.  I would be willing to continue maintaining the code, if you'd like.

The module is currently useable and has fairly good unit test coverage; I'm in contact with Prof. Emily Bender, which might result in a few more tgrep2 search features getting implemented.  She also mentions the documentation being a bit sparse, so perhaps that's something that will also need to be improved in the future.

A few thoughts about integration with NLTK:
- Where would the module go?  `nltk.parse.tgrep.py`, or maybe `nltk.tgrep.py`?  There's an opportunity here to rename things or otherwise make the interface different to fit better with NLTK.
- What should we do about `nltk_tgrep`'s requirements (pyparsing, future)?

Best,
Will

stevenbird:
Thanks @wroberts! I think it can live in `nltk/tgrep.py`. It would be great if you would continue to maintain it there. It needs to be ported to Python 3 – is that something you would have time to do? The pyparsing dependency should only be an issue at runtime. I.e.:

```
try:
    import pyparsing
except ImportError:
    pass
```

wroberts:
`nltk_tgrep` currently does py26, py27, py32, py33, and py34 (I'm using `future` to provide compatibility between py2 and py3).  I suppose `future` can be imported the same way as `pyparsing`, by simply throwing out ImportError.

I'll remove the license statement from the source (I'm currently distributing `nltk_tgrep` under MIT, but Apache-2.0 would work fine, too), `filter-branch` and send you a pull request as soon as I get a chance.

stevenbird:
@wroberts what do you think about modifying the interface to `tgrep_positions` and `tgrep_nodes` to support a list of trees as the first parameter? Then the overheads for processing the query expression would be amortised over all the trees being queried.

wroberts:
Hi @stevenbird that's a nice thought.

[...]

PR #964, opened 2015-04-27
tgrep: tgrep_positions and tgrep_nodes can take lists of trees

wroberts:
Change `tgrep_positions` and `tgrep_nodes` to be able to take a list of trees.  Update the docstrings for the functions to document this.

Mentioned in #931

### DIFF
--- a/nltk/tgrep.py
+++ b/nltk/tgrep.py
@@ -120,6 +120,7 @@ except ImportError:
     print('Warning: nltk.tgrep will not work without the `pyparsing` package')
     print('installed.')
 import re
+import types
 
 class TgrepException(Exception):
     '''Tgrep exception type.'''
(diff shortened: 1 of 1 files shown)


stevenbird:
@wroberts I haven't looked closely, but are you accepting a tree or list of trees for searching? I think it would be a cleaner design, and more consistent with the rest of NLTK, to only accept a list of trees, and to expect the calling program to wrap a single tree inside a list.

stevenbird:
I'm going to merge this, but tweak the API to require a sequence of trees for searching. If someone wants to search a single tree, they can trivially wrap it in the list constructor.

stevenbird:
@wroberts I've modified the API in https://github.com/nltk/nltk/commit/851304a422616ae3e655f5ef3eca9c3e39e5a8ce. Please let me know if you see any problems with this, thanks.

PR #966, opened 2015-04-28
Warn a user when there is no ROOT node, instead of throwing an exception...

dimazest:
This should resolve issues faced at #944. However, there is code that
depends on a fake root node, for example the tree visualisation code reads this and FStructure.to_depgraph() sets it.


stevenbird:
Thanks @dimazest

PR #967, opened 2015-04-28
Modified sem/relextract.py extract_rels() to accept lists instead of strings, added 'NE' to NE_CLASSES

ghost:
Modified extract_rels() to accept lists instead of strings for subjclass and objclass and matches against all entity types in the lists.  Also included 'NE' in all NE_CLASSES values so that binary ne_chunk tree results can also be used.

### DIFF
--- a/nltk/sem/relextract.py
+++ b/nltk/sem/relextract.py
@@ -30,10 +30,10 @@ from nltk.compat import htmlentitydefs
 # Dictionary that associates corpora with NE classes
 NE_CLASSES = {
     'ieer': ['LOCATION', 'ORGANIZATION', 'PERSON', 'DURATION',
-            'DATE', 'CARDINAL', 'PERCENT', 'MONEY', 'MEASURE'],
-    'conll2002': ['LOC', 'PER', 'ORG'],
+            'DATE', 'CARDINAL', 'PERCENT', 'MONEY', 'MEASURE', 'NE'],
+    'conll2002': ['LOC', 'PER', 'ORG', 'NE'],
     'ace': ['LOCATION', 'ORGANIZATION', 'PERSON', 'DURATION',
-            'DATE', 'CARDINAL', 'PERCENT', 'MONEY', 'MEASURE', 'FACILITY', 'GPE'],
+            'DATE', 'CARDINAL', 'PERCENT', 'MONEY', 'MEASURE', 'FACILITY', 'GPE', 'NE'],
     }
 
 # Allow abbreviated class labels
@@ -181,10 +181,10 @@ def extract_rels(subjclass, objclass, doc, corpus='ace', pattern=None, window=10
     Named Entities to particular types (any of 'LOCATION', 'ORGANIZATION',
     'PERSON', 'DURATION', 'DATE', 'CARDINAL', 'PERCENT', 'MONEY', 'MEASURE').
 
-    :param subjclass: the class of the subject Named Entity.
-    :type subjclass: str
-    :param objclass: the class of the object Named Entity.
-    :type objclass: str
+    :param subjclass: list of possible classes of the subject Named Entity.
+    :type subjclass: list
+    :param objclass: lsit of possible classes of the object Named Entity.
+    :type objclass: list
     :param doc: input document
     :type doc: ieer document or a list of chunk trees
     :param corpus: name of the corpus to take as input; possible values are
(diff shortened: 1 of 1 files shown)


stevenbird:
Does this look ok @ewan-klein?

nltk/sem/relextract.py
```diff
@@ -198,17 +198,20 @@ def extract_rels(subjclass, objclass, doc, corpus='ace', pattern=None, window=10
...
-            subjclass = _expand(subjclass)
-        else:
-            raise ValueError("your value for the subject type has not been recognized: %s" % subjclass)
-    if objclass and objclass not in NE_CLASSES[corpus]:
-        if _expand(objclass) in NE_CLASSES[corpus]:
-            objclass = _expand(objclass)
-        else:
-            raise ValueError("your value for the object type has not been recognized: %s" % objclass)
+    
+    for i, s in enumerate(subjclass):
```
dimazest, line 202: What if `subjclass` is a set, instead of a list. Then containment check in line 225 is sublienear.

Maybe, instead of modifying the passed `subjclass` it makes sense to build a new:

``` python
#  Calculate _expand() for every subjclass, but also keep the original
expanded_subjclass = (
  (_expand(s), s)
  for s in subjclass
  if s and s not in NE_CLASSES[corpus]
)
#  Select between an expanded and the original subjclass
subjclass = set(
  e if _expand(s) in NE_CLASSES[corpus] else s
  for e, s in expanded_subjclass
)
# BUT WE STILL NEED TO RAISE AN EXCEPTION!
```

nltk/sem/relextract.py
```diff
@@ -198,17 +198,20 @@ def extract_rels(subjclass, objclass, doc, corpus='ace', pattern=None, window=10
...
+        if s and s not in NE_CLASSES[corpus]:
+            if _expand(s) in NE_CLASSES[corpus]:
+                subjclass[i] = _expand(s)
+            else:
+                raise ValueError("your value for the subject type has not been recognized: %s" % s)
+
+    for i, o in enumerate(objclass):
```
dimazest, line 209: Looks very similar to lines 202-207, would it be possible to move this code to a helper function?

Once it's moved to a generator function, it will be possible to raise an exception.

nltk/sem/relextract.py
```diff
@@ -30,10 +30,10 @@
 # Dictionary that associates corpora with NE classes
 NE_CLASSES = {
     'ieer': ['LOCATION', 'ORGANIZATION', 'PERSON', 'DURATION',
```
dimazest, line 32: Because of containment checks, it would be nice to use a set instead of a list.

I guess the order of classes is not important here.

nltk/sem/relextract.py
ghost, line 202: @dimazest  can you clarify what you mean by the containment check being sublinear for a set instead of a list?

nltk/sem/relextract.py
ghost, line 209: @dimazest it could be moved to a helper function but I'm not sure any other piece of code would use that helper function.  Sorry to ask again but can you clarify why you think this needs an exception?  If the class is in the corpus list of accepted classes then it passes through so we don't need an extra exception right?  The point of these checks are to see if we need to expand the selected class or not.  Maybe I'm not understanding your question here.

nltk/sem/relextract.py
dimazest, line 202: @helshowk, consider this code,

``` python
# check whether 3 is in a list
3 in [1, 2, 4, 5, 3]
```

it will iterate over the list to figure out whether there is `3` or not. In the worst case scenario it will make as many comparison checks as there are in the list.

If a set is build, then the check

``` python
3 in set([1, 2, 4, 5, 3])
```

in the worst case will take less comparisons than there are elements in the list, thus the check will be sublinear.

nltk/sem/relextract.py
dimazest, line 209: Lines 202-207 and 210-214 seem to do the same thing but on different variables. If the code is moved to a function, that that function has to be called with different argument.

I'm talking about the `ValueError` (lines 214 and 207), the code i've written doesn't do such a check.

dimazest:
@helshowk I hope my comments are clear now.

nltk/sem/relextract.py
ghost, line 202: I don't disagree that it would be faster but I wasn't clear on why you think we need to optimize these containment checks and change the structure of the code.  Although your code is more compact I think it's less readable but I agree it would also be a bit faster to use a set instead of a list here (there are maybe a dozen elements here so I don't think it would really make a huge difference).  I think it's cleaner to leave the for loop with nested if's and you can explicitly see that the execution won't get to 216 unless all elements of subjclass (or their expansions) are recognized.  I agree the two loops can be pulled out to a helper function if you think that would be helpful.

dimazest, line 202: What about something like this, it's still readable and avoids code duplication.

``` python
def extract_rels(...):

    def expand(classes):
        for class_ in classes:
            if class_ and class_ is not in NE_CLASSES[corpus]:
                exapnded = _expand(class_)

                if expanded in NE_CLASSES[corpus]
                    yield expanded
                else:
                    raise ValueError("Your value for the type has not been recognized: %s" % o)

    subjclass = set(expand(subjclass))
    objclass = set(expand(objclass))

```

ghost, line 202: That looks good to me and still handles the ValueError.  Sorry for the delay I've been away, want to put that in?

dimazest, line 202: Cool. Could you please update the peer review.

ghost, line 202: Sorry but I don't know what you mean by update the peer review?  Do you mean make those modifications and update?

nltk/sem/relextract.py
stevenbird, line 209: I agree with @dimazest that it makes sense to abstract those lines into a helper function.

nltk/sem/relextract.py
stevenbird, line 202: I'm also confused about what peer review means here

ghost:
hi @stevenbird, what would you like me to do to get this closed?

stevenbird:
@ewan-klein – do you have any advice here?

ewan-klein:
Sorry, I'll try to look at this in the next day or so.

stevenbird:
Thanks @ewan-klein – then I can include this in 3.1

ewan-klein:
@helshowk Apologies for not addressing your PR sooner. My preference is to keep the `relextract` module backwards-compatible, in the sense that all the tests in `test/relextract.doctest` should still pass in the updated version of the code. Unfortunately this won't be true if `extract_rels` requires the classes to be lists of strings.

I've adopted @dimazest's suggestion for introducing an `expand()` function, and also modified the code so that `extract_rels` allows both strings and lists as arguments. I've put this into a new branch `relextract`. Can you have a look at the result? It would be great if you could update `test/relextract.doctest`  with an example which demonstrates a case where having a _list_ of classes gives useful results. You could either send me a working example, or fork the `relextract` branch and do another PR against it.

stevenbird:
Thanks @ewan-klein, @dimazest

stevenbird:
@helshowk – thanks for your contribution. I'm going to close this, but encourage you to follow up on @ewan-klein's suggestion and submit a new PR.

PR #968, opened 2015-04-29
convert twitter additional entities to csv + date_limit for not streaming accesses

lrnzcig:
1. The function to convert twitter additional entities to csv is now tested for all the possible entities of a tweet (hashtags, user_mentions, urls, media), of a user (urls), and for place (including bounding box)
2. Includes date limit functionality to tweets method for rest api (i.e. not streaming). Api of the method does not change. The param date_limit has different meaning when either streaming or not; in streaming it is a date in the future, if not, it is a date in the past

### DIFF
--- a/nltk/twitter/api.py
+++ b/nltk/twitter/api.py
@@ -72,4 +72,6 @@ class TweetHandlerI(object):
         (default implementation should be enough in most cases)
         """
         for item in data_chunk:
-            self.handle(item)
+            if self.handle(item) == False:
+                return False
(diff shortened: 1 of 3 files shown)


ewan-klein:
Thanks @lrnzcig.  It would be good to have some demo code for this new functionality, either in the `twitter_demo.py` file or else in the iPython notebook. I might try to restructure `twitter_demo.py` to make it easier to add new examples while still making it possible to select which demo to run.

PR #971, opened 2015-05-05
Code to load/train and run demo for word embedding

longdt219:
Hi @stevenbird, 

Here is the code for Word Embedding. Look at demo() of word2vec.py. I also committed the pruned model file which is part of the model trained on 100 billions words (and the code to get the pruned model prune_word2vec_model.py). Could you please put file word_embeddings.zip to the corpora list ?


stevenbird:
Thanks @longdt219, I'll take a look

nltk/embedding/prune_word2vec_model.py
```diff
@@ -0,0 +1,60 @@
...
+    # Only output word that appear in the Brown corpus 
+    from nltk.corpus import brown
+    dict = {}  
+    for word in brown.words():
+        dict[word] = 1 
+    print (len(dict.keys()))
+    
+    # Only print out the model that contain word in dict
+    out_file = 'pruned.word2vec.txt'
+    f = open(out_file,'wb')
```
dimazest: Please consider using a context manager to open a file, so you don't need to close it manually.

``` python
with open(out_file, 'wb') as f:
    ...
```

nltk/embedding/prune_word2vec_model.py
```diff
@@ -0,0 +1,60 @@
...
+import gensim
+from gensim.models.word2vec import Word2Vec
+
+if __name__ == '__main__':
+    # Load the C pre-trained model file 
+    model = Word2Vec.load_word2vec_format('GoogleNews-vectors-negative300.bin.gz', binary = True);
+    
+    # Only output word that appear in the Brown corpus 
+    from nltk.corpus import brown
+    dict = {}
```
dimazest: Please consider using a different name, for example `words`, `dict` is a builtin name for dictionaries.

Probably, set is a more suitable datastructure.

``` python
words = set(brown.words())
```

nltk/embedding/prune_word2vec_model.py
```diff
@@ -0,0 +1,60 @@
...
+    dict = {}  
+    for word in brown.words():
+        dict[word] = 1 
+    print (len(dict.keys()))
+    
+    # Only print out the model that contain word in dict
+    out_file = 'pruned.word2vec.txt'
+    f = open(out_file,'wb')
+    count = 0
+    for key in dict.keys():
```
dimazest:
``` python
words_in_model = words.intersection(model.vocab.keys())
```

lines 32-36 are not needed, if we really need to report the missing words, then

``` python
missing_words = words.difference(words_in_model)
```

nltk/embedding/prune_word2vec_model.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            temp = model[key]
+            count +=1
+        except KeyError:
+            print (' Not containing key ' + key)
+                       
+    f.write(str(count) + ' ' + str(len(model['word'])) + '\n')
```
dimazest:
``` python
f.write('{} {}\n'.format(len(words_in_model), len(model['word']))))
```

nltk/embedding/prune_word2vec_model.py
```diff
@@ -0,0 +1,60 @@
...
+    
+    for key in dict.keys():
```
dimazest:
``` python
for word in words_in_model:
   f.write('{} {}\n'.format(word, ' '.join(str(value) for value in model[word]))
```

by using `words_in_model`, it's guaranteed that `KeyError` won't happen.

nltk/embedding/prune_word2vec_model.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            f.write(key + ' ' + ' '.join(str(value) for value in model[key]) + '\n')
+        except KeyError:
+            pass
+    
+    f.close()
+    
+    # Reload the model from text file 
+    new_model = Word2Vec.load_word2vec_format(out_file, binary = False);
```
dimazest:
``` python
new_model = Word2Vec.load_word2vec_format(out_file, binary=False);
```

unnecessary spaces in `binary = False`

nltk/embedding/word2vec.py
```diff
@@ -0,0 +1,134 @@
...
+- Training new model
+- Load the pre-trained model from nltk.corpus 
+- Run some demo tasks  
+"""
+from __future__ import absolute_import, division, unicode_literals
+import gensim
+
+
+
+class Word2Vec():
```
dimazest:
``` python
class Word2Vec(object):
```

I'm actually not sure whether a wrapper brings any value in comparison of using the wordsim directly.

It would be beneficial to wrap the gensim implementation to make it compatible with  some API. However, I find the gensim interface for accessing vectors nicer. 

Catching exceptions and returning `-1` or `None` instead may lead to strange bugs.

`get_vector()` should be `__getitem__()`, so it would be possible to access a word vector using `model[word]`.

nltk/embedding/word2vec.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        self._check_model()
+        try:
+            return self.model.most_similar(positive, negative, topn)
+        except KeyError:
+            return None 
+    
+        
+def demo():
+    ####### DEMO TRAINING ####### #######
```
dimazest: maybe, it's better to write a demo as doctest, so the results are also visible

nltk/embedding/__init__.py
```diff
@@ -0,0 +1,15 @@
+# Natural Language Toolkit: word embedding package
+#
+# Copyright (C) 2001-2015 NLTK Project
+# Author: Long Duong <longdt219@gmail.com>
+# URL: <http://nltk.org/>
+# For license information, see LICENSE.TXT
+
+# Import Gensim if installed
+try:
```
dimazest: To make installations easier, it might be useful to define extra dependencies, see https://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

something like

``` python
setup(
    ...
    extras_require = {
        'everything':  ['gensim'],
    }
)
```

then if one wants to install NLTK with all it's dependencies, `nltk[everyting]` is installed. This is just a proposal, which is out of the scope of this PR, but clearly is relevant.

nltk/embedding/prune_word2vec_model.py
longdt219: Thanks @dimazest

nltk/embedding/prune_word2vec_model.py
longdt219: Thanks @dimazest

nltk/embedding/prune_word2vec_model.py
longdt219: Thanks @dimazest

nltk/embedding/prune_word2vec_model.py
longdt219: Thanks @dimazest

nltk/embedding/prune_word2vec_model.py
longdt219: Thanks @dimazest

nltk/embedding/prune_word2vec_model.py
longdt219: Thanks @dimazest

nltk/embedding/word2vec.py
longdt219: Hi @stevenbird and @dimazest 
I have the similar concern too. Currently, we haven't add any value to the wrapper.We might be able to do something more for the wrapper, e.g. visualization of word embedding using <a href = "http://lvdmaaten.github.io/tsne/"> t-SNE </a>. The other option is just write the demo to show how user can train gensim model using nltk data or load the pre-trained model from nltk.corpus (and probably visualize it using t-SNE). What do you think ?

nltk/embedding/word2vec.py
longdt219: Hi @dimazest, 
Probably I should convert the part using pre-trained model to the doctest. I assume that for each training, the vector representation might be slightly different.

kmike:
- I'm not sure it is OK to import gensim because itis LGPL while NLTK uses Apache license. 
- Returning -1 for similarity value may cause problems if similarity is used in some formula. For example, if similarities are multiplied, the result can silently become negative, or change the sign semi-randomly (if many words are missing). With an exception or None value users will get an error, and it'll require them to handle this case explicitly.
- Another question is what is the value of the proposed Word2Vec wrapper? gensim is also Python, and its API is also easy to use.

nltk/embedding/__init__.py
kmike: I like extras_require, +1 to add it. Maybe we shouldn't call it "everything" because many NLTK dependencies can't be installed with pip as they're non-Python.

nltk/embedding/word2vec.py
dimazest: Hello,

That's a good point!

longdt219:
Hi @stevenbird , 
I remove the package and convert it to the doctest. The pruned model path is also updated. The code for visualization using t-SNE and extracting pruned model are also included.

nltk/draw/tsne.py
```diff
@@ -0,0 +1,173 @@
+#
+#  tsne.py
+#  
+# Implementation of t-SNE in Python. The implementation was tested on Python 2.5.1, and it requires a working 
+# installation of NumPy. In order to plot the results of this example, a working installation of matplotlib is required.
+# The example can be run by executing: ipython tsne.py -pylab
+#
+#
+#  Created by Laurens van der Maaten on 20-12-08.
+#  Copyright (c) 2008 Tilburg University. All rights reserved.
```
kmike: Another option is to use http://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html 
It is an extra dependency, but the advantage is that we won't be adding more code to NLTK, and that we'll be teaching users to use scikit-learn - it is a good instrument to have in a toolbelt, and it is already NLTK dependency. 

I'm not quite comfortable with adding non-trivial math-heavy code with zero unit tests to NLTK; this module also doesn't follow pep8, and it is not clear it works in Python 3. Its license is also unclear.

dimazest: +1 for scikit-learn. It's a wide-spread and actively developed library, so using it will make NLTK maintenance much simpler.

longdt219:
Hi @stevenbird , 
I changed the code to use t-SNE from scikit-learn instead of adding another file to nltk. Thanks @kmike  and @dimazest for pointing this out.

stevenbird:
Thanks @longdt219. Also thanks @kmike and @dimazest.

stevenbird:
Oh dear, this PR contained a 45MB binary file! I need to try and rewrite history to avoid bloating this repository.

longdt219:
Oh sorry @stevenbird , I should have removed it.

stevenbird:
No problem, fixed now.

Issue #963, opened 2015-04-26
Searching for jar files


stevenbird:
It looks like we only find jar files when they are listed as files in the `CLASSPATH`:

https://github.com/nltk/nltk/blob/develop/nltk/internals.py#L587

However, we have a more general way of finding binaries, given a path that is just a directory that contains the binaries:

https://github.com/nltk/nltk/blob/develop/nltk/internals.py#L466

I think [find_jar_iter](https://github.com/nltk/nltk/blob/develop/nltk/internals.py#L548) needs to work when the `CLASSPATH` contains directories. I.e. `find_jar_iter(foo.jar)` should work if `foo.jar` exists in `/path/to/directory`.

With this fix, the [Stanford NER Tagger example](https://github.com/nltk/nltk/blob/develop/nltk/tag/stanford.py#L149) should work, rather than generating the error `NLTK was unable to find stanford-ner.jar!`.

PR #974, opened 2015-05-06
Fix bug #963 finding jar files

longdt219:
Hi @stevenbird, 
Here's the simple bug fix for #963. I also modified the <a href = 'https://github.com/nltk/nltk/wiki/Installing-Third-Party-Software' > wiki </a> for 3rd party software installation so that all Stanford packages use the consensus global environment.


stevenbird:
Thanks @longdt219

PR #988, opened 2015-05-26
Update boxer.py

prpfialho:
as per boxer rev 2554
- improved variable parse ( for topic(e1,p1), which erroneously yielded topic(x1,x1) )
- addded "duplex" parser
- changes to handle_named


stevenbird:
@dhgarrette would you please review this?

Issue #981, opened 2015-05-18
tgrep doctests


stevenbird:
Lots of tgrep doctests are failing. @wroberts would you mind taking a look please? It might just be a matter of converting a generator into a list in various places.

https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/testReport/

sahutd:
Hi, I have trouble running doctest for tgrep. I see that there are no .doctest file for tgrep module in nltk/test. Would you please explain how I should run the doctests for tgrep module?

wroberts:
Hi @sahutd !  You can run the doctests and unit tests with:

```
nltk/test/runtests.py -v nltk.tgrep
nltk/test/runtests.py -v nltk.test.unit.test_tgrep
```

nltk-webdeploy:
I think the failing doctests occur because a generator is returned instead
of a list, or vice versa.

sahutd:
Thanks. Will work on it now.

wroberts:
Hi @stevenbird I've fixed up some issues in pull request #990.  There's something quite wrong with tgrep though; the search functions return very long lists of empty list items.  Could be confusion between generators and lists, which can also be confused with tgrep trees (which are also a kind of list!).  I'll try looking through the history to see if I can figure out where things went wrong.

sahutd:
@wroberts  : I saw your pull request, and it overlaps with what I worked on just now. I will wait for your PR to be merged. Then depending on whats failing, I will work on that. For now, I will look for other issues to work on.

wroberts:
OK, got it. I hadn't realised the order of the arguments to tgrep_nodes and tgrep_positions had changed. Should be good to go now.

wroberts:
@sahutd Sorry to step on your toes!

sahutd:
@wroberts No issues! Learnt something :)

PR #990, opened 2015-05-28
Tgrep doctests

wroberts:
Some changes to stop so many tgrep doctests failing.  Problem description in #981


wroberts:
OK, got it.  I hadn't realised the order of the arguments to tgrep_nodes and tgrep_positions had changed.  Should be good to go now.

stevenbird:
Thanks @wroberts. Sorry I didn't think to fix the doctests corresponding to the interface changes I made.

Issue #945, opened 2015-04-14
Error when trying to change sentence from menu widget in nltk.app.rdparser_app.app()


hyp3rv1p3r:
This is the error:

Exception in Tkinter callback
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/tkinter/**init**.py", line 1533, in **call**
    return self.func(*args)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/nltk-3.0.2-py3.4.egg/nltk/app/rdparser_app.py", line 859, in edit_sentence
    EntryDialog(self._top, sentence, instr, self.set_sentence, title)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/nltk-3.0.2-py3.4.egg/nltk/draw/util.py", line 1944, in __init__
    self._entry = Entry(entryframe, width=width)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/tkinter/__init__.py", line 2514, in **init**
    Widget.**init**(self, master, 'entry', cnf, kw)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/tkinter/**init**.py", line 2122, in **init**
    (widgetName, self._w) + extra + self._options(cnf))
_tkinter.TclError: expected integer but got "43.5"

alvations:
Resolved in #1003

PR #1003, opened 2015-06-04
Fix error when changing sentence in rdparser_app

sahutd:
Cast resulting window width to int to avoid error
#945


stevenbird:
Thanks @sahutd

PR #1010, opened 2015-06-08
issues with automatic test json2csv

lrnzcig:
- carriage return depends on machine OS
- made a faster test with a smaller input file (actually first 100
  lines of the corpus)
- python 2.7 compatibility of TemporaryDirectory()


ewan-klein:
Thanks @lrnzcig

ewan-klein:
@lrnzcig, I had to make a few changes to get the code running under 3.4. Can you check that it still works for you? 

I'm still getting a Fail for `test_tweet_usermention()`.

lrnzcig:
@ewan-klein Sorry, indeed I should have tested it for python3. I've tested now for both 2.7 and 3, and for me it works in both!

I think I can guess what the problem is: different order in the user mentions. But I've modified the test so that it gives info on the line difference (PR #1013). Could you please run it and send me the output?

Thanks.

Issue #986, opened 2015-05-23
IBMModel1 aligner fails when aligning novel data (Python 3)


alexisdimi:
The method `IBMModel1.align()` fails when trying to align unseen data, evidently due to the tightened rules in Python 3 for comparing objects of different types:

``` python
from nltk import IBMModel1 
from nltk.corpus import comtrans
data = comtrans.aligned_sents("alignment-de-en.txt")
model = IBMModel1(data[:10], 20)

model.align(data[15])   # boom!
```

The training set is really tiny, so the error is inevitable. A real training set will probably lower the probability of an exception but not eliminate it completely.

Traceback:

```
    ...
    File "/Applications/anaconda/lib/python3.4/site-packages/nltk/align/ibm1.py", line 123, in
     align(self.probabilities[en_word][fr_word], i))
    TypeError: unorderable types: int() > NoneType()
```

This error does not arise under python 2 (with NLTK 3), since python 2 will never give an error when comparing objects of different types (it just compares the type names alphabetically, i.e. `NoneType < int`). Whether the resulting alignment decisions make any sense, I have not checked.

PR #1012, opened 2015-06-09
fix IBM Model 1 comparison of incomparable types

sahutd:
PS: This produces the same result as on python2 before the fix.  I am reading up on how to verify if this is indeed correct. Will confirm when done. If someone has expertise on text alignment, please do the  needful 
#986

### DIFF
--- a/nltk/align/ibm1.py
+++ b/nltk/align/ibm1.py
@@ -118,6 +118,8 @@ class IBMModel1(object):
             # Initialize the maximum probability with Null token
             max_align_prob = (self.probabilities[en_word][None], None)
             for i, fr_word in enumerate(align_sent.mots):
+                max_align_prob = (max_align_prob[0], max_align_prob[1] if
+                                  max_align_prob[1] else 0)
                 # Find out the maximum probability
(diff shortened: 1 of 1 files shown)


nltk/align/ibm1.py
```diff
@@ -118,6 +118,8 @@ def align(self, align_sent):
             # Initialize the maximum probability with Null token
             max_align_prob = (self.probabilities[en_word][None], None)
             for i, fr_word in enumerate(align_sent.mots):
+                max_align_prob = (max_align_prob[0], max_align_prob[1] if
+                                  max_align_prob[1] else 0)
```
hoontw, line 122: This probably isn't what you want, since it will break the comparison to None a few lines down. And it will still blow up lthe next statement in the case when max_align_prob[0] == self.probabilities[en_word][fr_word]

There is also another unrelated issue with this align method in handling OOV words (out of vocabulary), because there is no smoothing. Perhaps there needs to be a rethink about how to handle words that were not seen during training.

stevenbird:
Thanks @sahutd. I believe this is now outdated by @hoontw's work on the ibm modules.

PR #1013, opened 2015-06-09
test json2csv adding info for debugging

lrnzcig:
For debugging problems in #1010

### DIFF
--- a/nltk/test/unit/test_json2csv_corpus.py
+++ b/nltk/test/unit/test_json2csv_corpus.py
@@ -29,7 +29,7 @@ def are_files_identical(filename1, filename2):
     with open(filename1, "rb") as a:
         with open(filename2, "rb") as b:
             result = True
-            for lineA, lineB in izip(a.readlines(), b.readlines()):
+            for lineA, lineB in izip(sorted(a.readlines()), sorted(b.readlines())):
                 if (lineA.strip() != lineB.strip()):
                     print("Error while comparing files. First difference below.")
                     print("=> Output file line: '" + str(lineA) + "'")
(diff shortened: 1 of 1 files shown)


lrnzcig:
If the problem is related to order, already solved in commit 577ceb9

ewan-klein:
@lrnzcig: I'm still getting this:

```
........Error while comparing files. First difference at line below.
=> Output file line: b'#tomorrowspaperstoday #bbcpapers http://t.co/ts9ZnULDwr\xe2\x80\x9d",21910500,suttonnick\r\n'
=> Refer. file line: b'\n'
F..
FAIL: test_tweet_usermention (__main__.Test)
  File "/Users/ewan/git/nltk/nltk/test/unit/test_json2csv_corpus.py", line 105, in test_tweet_usermention
    'tweets.20150430-223406.usermention.csv.ref'), 'Error in csv file'
AssertionError: Error in csv file

Ran 11 tests in 0.237s

FAILED (failures=1)
Process terminated with an exit code of 1
```

Is there a spurious blank line in the reference file?

lrnzcig:
@ewan-klein I've been looking at the ref file and it looks ok.

The only particular thing about this area of the file is that this is a tweet with carriage returns in the text of the tweet. Looking at the binary content of the file, the EOL of the files is `'\r\n'` but the EOL of the text of the tweet is `'\n'`. I don't know exactly what platform you are using, but I've been browsing and python3.4 does have by default support for "universal newlines". Besides, this same tweet with the same text is present in some other reference files. So I don't really have an good explanation for the problem.

What OS are you using? Could you send me the file you generate, so that I take a look to the binary content and try to figure out what's the difference? (Although I see the problem seems to happen when reading the reference file).

Thanks.

lrnzcig:
Sorry @ewan-klein forge my last comment. I did not realize I can reproduce the problem in python3. Let me debug it and do yet another pull request.

Issue #965, opened 2015-04-27
A bug in texttiling code in python 3


mhjang:
In texttiling.py,

The function identify_boundaries has a bug if you're running this on python 3 in the below piece of code.

```
  for dt in hp:
       boundaries[dt[1]] = 1
       for dt2 in hp: #undo if there is a boundary close already
               if dt[1] != dt2[1] and abs(dt2[1]-dt[1]) < 4 \
                         and boundaries[dt2[1]] == 1:
               boundaries[dt[1]] = 0
 return boundaries
```

Because lambda filter is not anymore a list unlike in python 2, the inner loop (for dt2 in hp) is done after a single pass, so it would run only one time for the first entry of the outer loop. This results in just one tile. To fix this, you need to add a explicit list conversion of the filter.

```
hp = list(filter(lambda x:x[0]>cutoff, depth_tuples))
```

before the loop. Thanks.

PR #1014, opened 2015-06-10
Cast filter object to list in texttiling

sahutd:
Added doctest which catches the current regression across python2 and
## python3

In the added doctest, I have chosen the smallest text(that I could get from demo() such that it fails without casting to list() on python2 and python3, but gets consistent results across python2 and python3 with the fix.
#965


stevenbird:
Thanks @sahutd.

PR #1016, opened 2015-06-10
json2csv_entities bug fix

lrnzcig:
- making sure that the function gets the info from the main object,
  e.g. from the tweet metadata instead of the retweeted metadata
- a couple of lines of the reference files were actually wrong

@ewan-klein I'd say this is the "final" fix, but please tell me if you find problems
Tested for both 3.4 and 2.7


ewan-klein:
Thanks @lrnzcig, the tests all pass for me now. I've done a bit of cosmetic refactoring and moved the reference files into a subdirectory. 

One thing I've noticed is that if one test fails, then it seems to cascade to a few other tests. I suspect this may be due to the TempDirectory not getting closed when there's an exception raised inside the `with` context. Since I don't know how to fix this, and it doesn't seem to be that serious, I'm inclined to leave it.

PR #1027, opened 2015-06-26
Added --resolve option to Boxer to disable Anaphora resolution

redsk:


redsk:
Sorry, you're right. I changed the order of the parameter.

stevenbird:
Thanks @redsk

PR #1034, opened 2015-07-03
Twitter compressed output files + generator (comments PR #1021)

lrnzcig:
for both streaming and rest apis

tested for python2.7 and 3


ewan-klein:
Thanks @lrnzcig

PR #1035, opened 2015-07-05
Twitter tokenizer normalization functions

fievelk:
Add options to normalize tweets:
- reduce character sequences of length 3 (or more) to sequences of length 3
- strip Twitter handles for usernames

### DIFF
--- a/nltk/tokenize/casual.py
+++ b/nltk/tokenize/casual.py
@@ -205,7 +205,7 @@ def _replace_html_entities(text, keep=(), remove_illegal=True, encoding='utf-8')
     is". For more information see the tests.
     Always returns a unicode string (with the entities removed).
     """
-    
+
     def convert_entity(m):
         entity_body = m.group(3)
         if m.group(1):
@@ -232,9 +232,9 @@ def _replace_html_entities(text, keep=(), remove_illegal=True, encoding='utf-8')
                 return unichr(number)
             except ValueError:
                 pass
-    
+
         return u'' if remove_illegal else m.group(0)
-    
+
     return _ent_re.sub(convert_entity, _str_to_unicode(text, encoding))
 
 
@@ -242,8 +242,9 @@ def _replace_html_entities(text, keep=(), remove_illegal=True, encoding='utf-8')
 
 class TweetTokenizer:
     """Tokenize Tweets"""
-    def __init__(self, preserve_case=True):
+    def __init__(self, preserve_case=True, normalize=False):
         self.preserve_case = preserve_case
+        self.normalize = normalize
 
     def tokenize(self, text):
         """
@@ -254,6 +255,8 @@ class TweetTokenizer:
         """
         # Fix HTML character entities:
         text = _replace_html_entities(text)
+        if self.normalize:
+          text = normalize(text)
         # Tokenize:
         words = WORD_RE.findall(text)
         # Possibly alter the case, but avoid changing emoticons like :D into :d:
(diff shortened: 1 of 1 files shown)


nltk/tokenize/casual.py
```diff
@@ -262,16 +270,34 @@ def tokenize(self, text):
                               x.lower()), words))
         return words
 
+######################################################################
+# Normalization Functions
+######################################################################
+
+def reduce_lengthening(text):
+    '''
+    Replace character sequences of length 3 or greater with sequences of length 3.
```
jnothman: I couldn't understand what you were talking about until I read the code. Please say "repeated character sequences", and add a doctest example.

nltk/tokenize/casual.py
```diff
@@ -262,16 +270,34 @@ def tokenize(self, text):
...
+    '''
+    pattern = re.compile(r"(.)\1{2,}")
+    return pattern.sub(r"\1\1\1", text)
+
+def remove_handles(text):
+    '''
+    Remove twitter username handles from text.
+    '''
+    pattern = re.compile(r"(?<=^|(?<=[^a-zA-Z0-9-\.]))@([A-Za-z_]+[A-Za-z0-9_]+)")
```
jnothman: Nitpicks:
- Don't nest look-behinds: `^` doesn't need a lookbehind, and can just be a normal (non-capturing) group.
- The capturing group for the username is unnecessary.
- Should `_` be included in `[^a-zA-Z0-9-\.]`? If so, this can be written as `[^\w.-]`, just as `[A-Za-z0-9_]` can be written `\w`. I suspect, though, that the lookbehind expression should actually include all unicode word characters, while the username expression should be ASCII exclusively.
- a parameter for substituting with a placeholder (like `'<USER>'`) rather than deleting might make this more useful (and could be allowed as a value for `strip_handles` param rather than True).
- this will result in double-spaces after deletions; do we care? Handling them could be fussy.
- `strip_handles` param and `remove_handles` function should probably have the same name
- again, a doctest example would be nice.

jnothman: Also, why not extend this to 2-character repetitions like `hahahaha` and `lolololol`

nltk/tokenize/casual.py
jnothman: Also, why not extend this to 2-character repetitions like `hahahaha` and `lolololol`?

nltk/tokenize/casual.py
fievelk: - I removed nested look-behind and the capturing group, and I refactored the regex using shorthands.
- Regarding double-spaces, `TweetTokenizer` result should not affected by them; nonetheless, I agree with you on the fact that it would be better to handle them.
- I personally prefer to differentiate between function and parameter names. In [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments) something similar is suggested for reserved keywords:
  
  > If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss . (Perhaps better is to avoid such clashes by using a synonym.)

jnothman: Yes, it comes down to personal preference, but I think this is a different
case to reserved words. Arguably, that dictum agrees with my position:
choose something with a canonical name, and if the system forbids it, mark
it deterministically, rather than choosing some arbitrary alternation (i.e.
strip <-> remove, where I would not be able to remember which is which).

ewan-klein:
Thanks @fievelk. 

Following comment above by @jnothman, I think at some point it would be nice to add the option for replacing user handles with a dummy token such as `<USER>`, and similarly for URLs and emoticons.

PR #1036, opened 2015-07-06
Fix comment in parseLexicon, in CCG

bryant1410:
Regarding to the target category, In [parseLexicon function, in CCG](https://github.com/nltk/nltk/blob/28d7d363769958f3608de574400b20ea3fdc9ad9/nltk/ccg/lexicon.py#L186), shouldn't it be "the first one" (or "the first category", or similar) instead of "the first line"?


stevenbird:
@bryant1410 – yes you're right, thanks

PR #1038, opened 2015-07-08
Support for Unicode in CCG

bryant1410:
Changes the definition of words in the regex of the lexicon parsing in CCG to support Unicode and a test is added too.


stevenbird:
Thanks very much @bryant1410

bryant1410:
You're welcome. I'm glad to have helped :smile:

PR #1040, opened 2015-07-10
Sentiment Analysis approaches

fievelk:
This Pull Request is a work-in-progress for different Sentiment Analysis approaches.


ewan-klein:
Thanks @fievelk. I had to delete a few stale files which were left over from previous version of twitter branch.

PR #1073, opened 2015-08-09
Enhancement/documentation contributing

turicas:
This is not a feature-pull-request: it just organizes the general documentation, authors and how to contribute.


stevenbird:
Thanks @turicas

turicas:
@stevenbird, you're welcome! ;-)

I've made some fixes -- the only missing information now is the Slack URL, I think.

stevenbird:
Thanks @turicas. I think I need to create a public channel first...

stevenbird:
It turns out that Slack doesn't support this, and they even [blocked a work-around](http://wiredcraft.com/blog/using-slack-as-a-public-chat/), so I'm removing the reference to slack.

PR #1076, opened 2015-08-11
Sentiment Analysis support

fievelk:
This Pull Request aims to add new corpora and lexicons specifically built for Sentiment Analysis purposes. Their integration has been discussed with authors and annotators, and we agreed for licensing terms that respect NLTK Apache License.

New corpora and Lexicons:
- **_Opinion Lexicon by Bing Liu**_
  - **URL**: http://www.cs.uic.edu/~liub/FBS/sentiment-analysis.html#lexicon
  - **PAPERS**: [Mining and summarizing customer reviews (Hu, Minqing and Liu, Bing)](http://www.cs.uic.edu/~liub/publications/kdd04-revSummary.pdf)
  - **CONTACT**: Bing Liu (liub@cs.uic.edu)
  - **CORPUS NAME**: `opinion_lexicon`
- **_Customer Review Dataset**_ (Product reviews)
  - **URL**: http://www.cs.uic.edu/~liub/FBS/sentiment-analysis.html#datasets
  - **PAPERS**: [Mining and summarizing customer reviews (Hu, Minqing and Liu, Bing)](http://www.cs.uic.edu/~liub/publications/kdd04-revSummary.pdf)
  - **CONTACT**: Bing Liu (liub@cs.uic.edu)
  - **NOTES**: Title of review, product feature, positive/negative label with opinion strength, other info (comparisons, pronoun resolution, etc.)
  - **CORPUS NAME**: `product_reviews_1`
- **_Additional Review Dataset**_ (Product reviews)
  - **URL**: http://www.cs.uic.edu/~liub/FBS/sentiment-analysis.html#datasets
  - **PAPERS**: [A holistic lexicon-based approach to opinion mining (Ding, Xiaowen and Liu, Bing and Yu, Philip S)](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.155.4017&rep=rep1&type=pdf)
  - **CONTACT**: Bing Liu (liub@cs.uic.edu)
  - **NOTES**: Title of review, product feature, positive/negative label with opinion strength, other info (comparisons, pronoun resolution, etc.)
  - **CORPUS NAME**: `product_reviews_2`
- **_Pros and Cons Dataset**_ (Pros and cons sentences)
  - **URL**: http://www.cs.uic.edu/~liub/FBS/sentiment-analysis.html#datasets
  - **PAPERS**: [Mining Opinions in Comparative Sentences (Ganapathibhotla, Murthy and Liu, Bing)](http://www.cs.uic.edu/~liub/FBS/Coling-2008-camera-ready.pdf)
  - **CONTACT**: Bing Liu (liub@cs.uic.edu)
  - **NOTES**: A list of sentences tagged `<pros>` or `<cons>`
  - **CORPUS NAME**: `pros_cons`
- **_Comparative Sentences**_ (Reviews)
  - **URL**: http://www.cs.uic.edu/~liub/FBS/sentiment-analysis.html#datasets
  - **PAPERS**: [Identifying Comparative Sentences in Text Documents (Nitin Jindal and Bing Liu)](http://www.cs.uic.edu/~liub/publications/sigir06-comp.pdf), [Mining Opinion Features in Customer Reviews (Minqing Hu and Bing Liu)](http://www.cs.uic.edu/~liub/publications/aaai04-featureExtract.pdf)
  - **CONTACT**: Bing Liu (liub@cs.uic.edu)
  - **NOTES**: Sentence, POS-tagged sentence, entities, comparison type (non-equal, equative, superlative, non-gradable)
  - **CORPUS NAME**: `comparative_sentences`
- **_Sentence Polarity Dataset 1.0**_
  - **URL**: http://www.cs.cornell.edu/People/pabo/movie-review-data/
  - **PAPERS**: [Seeing stars: Exploiting class relationships for sentiment categorization with respect to rating scales (Pang, Bo and Lee, Lillian)](http://arxiv.org/pdf/cs/0506075.pdf)
  - **CONTACT**: [Bo Pang](http://research.yahoo.com/Bo_Pang), [Lillian Lee](http://www.cs.cornell.edu/home/llee)
  - **CORPUS NAME**: `sentence_polarity`
- **_Subjectivity Dataset 1.0**_
  - **URL**: http://www.cs.cornell.edu/People/pabo/movie-review-data/
  - **PAPERS**: [A sentimental education: Sentiment analysis using subjectivity summarization based on minimum cuts (Pang, Bo and Lee, Lillian)](http://arxiv.org/pdf/cs/0409058.pdf)
  - **CONTACT**: [Bo Pang](http://research.yahoo.com/Bo_Pang), [Lillian Lee](http://www.cs.cornell.edu/home/llee)
  - **CORPUS NAME**: `subjectivity`

Additionally, `sentiment_analyzer.py` contains methods can be used as a sort of wrapper around common NLTK functionalities to ease Sentiment Analysis tasks, especially for teaching purposes.

Several demos and helper functions have been implemented in `util.py` to provide examples using different datasets.


ewan-klein:
@stevenbird Can you please have a look at this PR? Just for clarity, I should point out that the sentiment-relatted resources/corpora listed above have already been installed in `nltk_data`. 

@fievelk: as far as possible, I converted all the encodings of the sentiment corpora to UTF-8. Can you make sure that all your tests are based on the versions that are installed by `nltk.download()` rather than any local versions you might have installed? If I've failed to convert any of the corpora properly, please raise an issue.

fievelk:
@ewan-klein: Thank you. I deleted old local files and re-downloaded the corpora using `nltk.download()` before doing the PR. Tests using `tox` succeeded with this data, so I think there should not be problems.

stevenbird:
@fievelk – thanks for the huge amount of work you have done here.

stevenbird:
I noticed a file `nltk/corpus/reader/categ_sents.py` and we've generally erred on the side of unabbreviated words in module names. How about `categorized_sents.py`. Note that you also already have `comparative_sents.py` (and "categorized" and "comparative" both have 11 characters). Note that "sents" is already widely used as an abbreviation in NLTK.

nltk/corpus/reader/opinion_lexicon.py
```diff
@@ -0,0 +1,115 @@
...
+- Bing Liu, Minqing Hu and Junsheng Cheng. "Opinion Observer: Analyzing and
+    Comparing Opinions on the Web". Proceedings of the 14th International World
+    Wide Web conference (WWW-2005), May 10-14, 2005, Chiba, Japan.
+"""
+
+from nltk.compat import string_types
+from nltk.corpus.reader import WordListCorpusReader
+from nltk.corpus.reader.api import *
+
+class IgnoreReadmeCorpusView(StreamBackedCorpusView):
```
stevenbird, line 35: This might be better in `util.py`, with a function parameter to the initializer that specifies how to skip the readme.

nltk/corpus/reader/product_reviews.py
```diff
@@ -0,0 +1,324 @@
...
+from nltk.tokenize import *
+
+TITLE = re.compile(r'^\[t\](.*)$') # [t] Title
+FEATURES = re.compile(r'((?:(?:\w+\s)+)?\w+)\[((?:\+|\-)\d)\]') # find 'feature' in feature[+3]
+NOTES = re.compile(r'\[(?!t)(p|u|s|cc|cs)\]') # find 'p' in camera[+2][p]
+SENT = re.compile(r'##(.*)$') # find tokenized sentence
+
+
+@compat.python_2_unicode_compatible
+class Review(object):
```
stevenbird: The choice of name here, `Review`, makes sense, given that reviews are not necessarily about products. Could this corpus reader be applied to other kinds of reviews? If not, this class should be `ProductReview`. But if so, then the module name and corpus reader name could omit "product".

stevenbird:
@fievelk can you explain the thinking behind including an ipython notebook in the source tree, alongside the module. Is there a precedent for this? How would we expose it to users?

Ultimately, this could have implications for what we do with the other documentation that is currently in doctest format in `nltk/test` (and accessible via http://www.nltk.org/howto/).

stevenbird:
I think the import of `twython` needs to be inside a `try` block (cf `numpy` in other modules). This gives us more flexibility in importing `sentiment` and `twitter` modules from the top level `__init__.py`.

fievelk:
@stevenbird Thank you very much for your detailed feedback.
- I think it is possible to safely change names for both `categ_sents.py` and `ProductReviewsCorpusReader`. The latter corpus can actually be used to parse other kinds of reviews, so I will manage to update its name as soon as possible.
- Unfortunately I was not yet involved in the project when `sentiment_plotter.ipynb` was added to the repository. I will ask for information about it so that I will be able to answer your question.
- It is probably better to create a separate PR for `twython` imports. I think this also explains issues with `tox` no-deps environments.

ewan-klein:
@fievelk the IPython notebook is a residue from work I did in an earlier development phase. I have moved it to `nltk/test`. However, we should review the content of the notebook to check that it still makes sense to include it.

fievelk:
@ewan-klein @stevenbird : Regarding wrapping `twython` imports within `try` blocks, something similar is being discussed in [malt.py #1080](https://github.com/nltk/nltk/issues/1080) regarding `six` imports. @kmike said:

>  It should be definitely solved as a part of our test running setup, not as a code change.

Do you think this is the case for `twython` as well? In this case I would wait for `tox` fixes.

ewan-klein:
@fievelk the `twitterclient.py` module has

```
try:
    from twython import Twython, TwythonStreamer
    from twython.exceptions import TwythonRateLimitError, TwythonError
except ImportError as err:
    import textwrap
    MSG = """The NLTK twitterclient module requires the Twython package. See\
    https://twython.readthedocs.org/ for installation instructions."""
    err.msg = textwrap.fill(MSG)
    raise
```

The rationale is not specifically for testing, but just in case a user wants to use some of the core NLTK packages without installing twython.

fievelk:
@ewan-klein But if the exception is re-raised, am I wrong when I say that users would not be able to use NLTK anyway? If the exception is not raised, instead, it should mean that users already have `twython`, or that this `try/except` is never reached (so they could use other NLTK packages without worrying anyway).

Throughout NLTK, `numpy` imports are usually handled like this:

```
try:
    import numpy as np
except ImportError:
    pass
```

However, there are also cases<sup>1</sup> in which these exceptions are specifically handled for testing with `nose`:

```
from nose import SkipTest
try:
    import numpy
except ImportError:
    raise SkipTest("segmentation.doctest requires numpy")
```

Maybe it would be good to choose a specific strategy for the future and apply it to `twython` as well.

---

<sup>1</sup><sub>Specifically: `segmentation.py`, `classify_fixt.py`, `probability_fixt.py`, `segmentation_fixt.py`, `test_hmm.py`, `test_tag.py`.</sub>

ewan-klein:
@stevenbird Are we going to update `nltk/__init__.py` to import `twitter` and `sentiment` packages? If not, then would it be best to place `try` blocks to wrap the import of `twython` separately in each package?

stevenbird:
@ewan-klein: yes, I agree, we should do it that way.

stevenbird:
@ewan-klein – is this ready to be merged, for inclusion in 3.0.5?

fievelk:
@ewan-klein @stevenbird : Using a `try/except` block with a simple `pass` statement for every Twython import would not unfortunately fix `tox` tests. For example, wrapping the following line

```
from twython import Twython, TwythonStreamer
```

would produce failures when defining `class Streamer(TwythonStreamer)`. I think that these tests should be skipped when `twython` is not available in a test environment (`nodeps`), and an exception should be raised when in a production environment instead.

If `nodeps` dependency problems are supposed to be solved within Tox, I think we should only handle import exceptions raising them with a meaningful error message (not for tests purposes). This would actually be their correct output in a production environment.

Following [NLTK Testing](http://www.nltk.org/dev/local_testing.html) I tried to add a `setup_module()` function in the modules that use `twython` to skip `nose` tests, but tests were not skipped for some reason.

kmike:
> For example, wrapping the following line `from twython import Twython, TwythonStreamer` would produce failures when defining class Streamer(TwythonStreamer).

You're right that the solution for twython shouldn't be only for tests. If twython is optional we should raise an error if user tries to use related modules, but twython is not available.

> I think that these tests should be skipped when twython is not available in a test environment (nodeps), and an exception should be raised when in a production environment instead.

The question is when to raise this exception. There are 2 options:

1) at import time;
2) when an object which uses twython is instantiated.

(1) is easier to implement, and maybe provides a better feedback for user, but we should be careful not to import modules which depend on twython from nltk top-level namespaces. Protecting twython imports with try/except is not necessary if we go with this option (it is better to raise a regular ImportError than a custom exception because it provides more information). But it could make sense to use try/except in upper-level imports which expose modules which use twython.

(2) For (2) we can provide a dummy Streamer implementation which raises an error on instantiation. This implementation will be used if twython is not available.

To run `setup_module()` Python needs to import the module `setup_module()` is defined at; if this module imports packages which are not available (like twython) then error is raised. Is it the problem you're facing?

ewan-klein:
@stevenbird We've addressed everything apart from the `twython` import issue that you raised. The points raised by @kmike are also applicable to the`twitter` package, so it would be good to have an agreed resolution on his point (1). Could you clarify which of the following options you would prefer:
- Handle `twython` import at top level or package level?
- If package level, use 'try/except' block or just let an `ImportError` be raised?

I'm not sure about 

> it is better to raise a regular ImportError than a custom exception because it provides more information.

The counter-argument is that we can make a custom exception message more use-friendly for novices than a standard stack trace.

fievelk:
@kmike :

> To run setup_module() Python needs to import the module setup_module() is defined at; if this module imports packages which are not available (like twython) then error is raised. Is it the problem you're facing?

Yes, I think this is the case. I tried to define `setup_module()` in `__init__.py` and in the specific modules that use `twython`. So the error is raised, ignoring `setup_module()`.

stevenbird:
I prefer the simple solution, conditional import of modules that have non-standard dependencies (@kmike's option 1 I guess). This is what we do for the `tkinter` dependency in the `app` and `draw` packages.

This approach does not preclude doing something fancier in future (like option 2).

ewan-klein:
Thanks @stevenbird. So I'll do something analogous to `draw/__init__.py`  for both the `twitter` and `sentiment` packages. E.g.,

```
try:
    import twython
except ImportError:
    import warnings
    warnings.warn("nltk.sentiment package not loaded "
                  "(please install twython library).")
```

This is independent, I believe, of how we address the issues with testing raised by @fievelk.

ewan-klein:
@stevenbird Issue about `twython` import is addressed in https://github.com/nltk/nltk/commit/798984059a4eb352026b0f11f92ef3e98a0391b3. So I think we're ready to merge this now.

fievelk:
@ewan-klein @stevenbird @kmike : If I am not mistaken, I think it would be however necessary to solve `tox` problems with `nodeps` environments: we are currently handling ImportError exceptions, but tests will fail anyway for `six`, `twython` and `requests` imports (this is not specifically related to this PR).
For example: 

```
tox -e py26-nodeps
```

will fail when trying to import `six` in `malt.py` (see [#1080](https://github.com/nltk/nltk/issues/1080)). After fixing `six` errors (just for test purposes) `tox` tests will succeed until they meet `twython` imports. Fixing `twython` imports (again, just so that tests succeed) will make `tox` tests succeed until they face `requests` import errors (see [twitterclient.py #L30](https://github.com/nltk/nltk/blob/821ab10a4ee3c83bbe1239a2a22cbf1532a661cc/nltk/twitter/twitterclient.py#L30)).
Therefore I think that our priority should be that of fixing `tox` tests for `nodeps` environments.

alvations:
`six` is a weird issue because when we soon drop the py2.6 support, it would naturally be resolved since py2.7.x and py3.x comes shipped with `six`. Is that right?

dimazest:
I'm not sure whether it's included to py27 and py3. Instead, six is a one file distribution, what if we ship it together with NLTK and import it as `from nltk import six`.

kmike:
See https://github.com/nltk/nltk/issues/1080#issuecomment-136670517. tox issue is trivial to fix - just add six to -nodeps environments like it is added to other environments. Alternatively, we can figure out why aren't `install_requires` dependencies installed in tox.

fievelk:
I try to write a brief recap of the issue.
The `twitter` package contains some functions that need `twython` and `requests` libraries to work (they are contained in `twitter/util.py`). It is therefore expected that users get an ImportError when trying to use these functions without having `twython` installed.
The `sentiment` module needs to use some of the `twitter` functions that do not require `twython` and `requests`. However, since these functions are in the same `twitter/util.py` module, trying to import `sentiment` without having `twython` installed will produce an ImportError as well. I think we should fix this.

My suggestion is to split `twitter/util.py` in two modules (e.g. `twython_util.py` and `common.py`), one containing the functions that require `twython` and `requests`, and another one containing the other functions. In this way, we would be able to do the following in `twitter/__init__.py`:

```
try:
    import twython
except ImportError:
    import warnings
    warnings.warn("twython library is not installed. "
                  "Some Twitter functionalities will not be available.")
else:
    import twython_util.py

import common.py
```

This will allow users to correctly import the `sentiment` package. When they will try to load the `twitter` package, they will receive a warning in case they do not have `twython` installed.

Another possible option might be to move functions that require `twython` directly inside `twitter/__init__.py` and . However, the [Package Structure guide](https://github.com/nltk/nltk/wiki/Package-Structure) does not mention a similar use.

Do you think this could work? Do you have alternative options (also regarding module names)?

stevenbird:
Merged into `test` for regression testing at https://nltk.ci.cloudbees.com/job/test/

ewan-klein:
@stevenbird OK if I resolve the conflicts by merging `develop` back into this branch? They're due I think to some recent updates to `twitter` package.

stevenbird:
@ewan-klein good idea. I also need to update the tests to use Python 2.7 and 3.4.

stevenbird:
Thanks @fievelk and @ewan-klein for all this work!

PR #1077, opened 2015-08-12
Introduce base IBM model and minimum probability values

hoontw:
Move shared code and documentation to a new module, `ibm_model`.
Ensure a minimum value for probabilities in order to avoid underflow, singularities, and precision errors. In practical terms, the minimum value is small enough to be regarded as zero.

Other smaller changes for Model 3:
- Expose `p1` instead of `p0`.
- Swap order of `l` and `m` in `alignment_table` and `distortion_table`.
- Fix bug in updating fertilities when looking for neighboring alignments. Fertility tables were shared among some alignments, instead of each alignment having its own table.
- Add additional training data to doctest, so that the correct alignments can be converged. The original expected result was incorrect.

### DIFF
--- a/nltk/align/ibm_model.py
+++ b/nltk/align/ibm_model.py
@@ -37,6 +37,85 @@ Translation: Parameter Estimation. Computational Linguistics, 19 (2),
 263-311.
 """
 
+from abc import ABCMeta
+from collections import defaultdict
+
+
+class IBMModel(metaclass=ABCMeta):
+    """
+    Base class for all IBM models
+    """
+
+    # Avoid division by zero and precision errors by imposing a minimum
+    # value for probabilities. Note that this approach is theoretically
+    # incorrect, since it may create probabilities that sum to more
+    # than 1. In practice, the contribution of probabilities with MIN_PROB
+    # is tiny enough that the value of MIN_PROB can be treated as zero.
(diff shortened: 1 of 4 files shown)


nltk/align/ibm_model.py
```diff
@@ -0,0 +1,163 @@
...
+Robert L. Mercer. 1993. The Mathematics of Statistical Machine
+Translation: Parameter Estimation. Computational Linguistics, 19 (2),
+263-311.
+"""
+
+from abc import ABCMeta
+from collections import defaultdict
+
+
+class IBMModel(metaclass=ABCMeta):
```
kmike: Hey @hoontw,

I'm not doing a proper review now, but could you please check that your changes work in Python 2.x as well as 3.x? This is a Python 3-only syntax for metaclasses. It seems we can just remove ABCMeta here because its features are not used.

hoontw: Good point. I was actually planning to use the abstract class features in upcoming work, but now I realize getting it to work with both Python 2 and 3 is going to be a pain.

I will remove the use of abstract classes and add a note in the docstring that this is meant to be used as an abstract class.

stevenbird: Thanks @kmike

stevenbird:
Nice work @hoontw. I particularly appreciate the attention to good design and good documentation

Issue #1080, opened 2015-08-14
malt.py


stevenbird:
@alvations, there's some minor issues with `malt.py`. Would you please take a look? It breaks with Python 2, and it has some non PEP-8 formatting (e.g. missing whitespace around operators)

```
  File "./malt.py", line 185
    dir=self.working_dir, mode='w', delete=False) as input_file, \
                                                               ^
SyntaxError: invalid syntax
```

alvations:
@stevenbird I've fixed the problem for `taggedsent_to_conll()` reported on #1049 and PEP8 the `malt.py` and some minor doctest fixes with #1082. I am not having problems with python2. Could you post the error traceback from your machine?

alvations:
I just realized that `with` statements in python2.6 doesn't do multiple inputs. 

```
alvas@ubi:~$ python2.6
Python 2.6.9 (default, Oct 22 2014, 19:56:52) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> with io.open('in.txt', 'r', encoding='utf8') as fin:
...     print fin.read()
... 
After Du died of suffocation, her boyfriend posted a heartbreaking message online: "Losing consciousness in my arms, your breath and heartbeat became weaker and weaker. Finally they pushed you out of the cold emergency room. I failed to protect you."

Li Na, 23, a migrant worker from a farming family in Jiangxi province, was looking forward to getting married in 2015.

>>> with io.open('in.txt', 'r', encoding='utf8') as fin, io.open('in.txt', 'r') as fout:
  File "<stdin>", line 1
    with io.open('in.txt', 'r', encoding='utf8') as fin, io.open('in.txt', 'r') as fout:
                                                       ^
SyntaxError: invalid syntax
>>> 
```

I've fix it so that it's py2.6.x compatible, https://github.com/alvations/nltk/commit/9df9371b55fd3d829e77886d23ef2fff02d2b457

stevenbird:
Thanks @alvations. Please note that we use four-space indentation (recommended by PEP-8).

alvations:
@stevenbird #1084 changed the tabs to four spaces. Sorry about the tabs, it was set on automatically through my IDE.

fievelk:
@stevenbird @alvations 
After merging `develop` I am getting errors for **Python 3.2** and for all **no-deps** environments, relatively to `malt.py`, using `tox`. For example:

```
py32 inst-nodeps:
=============================================
[...]

PR #1082, opened 2015-08-14
PEP8-ify the malt.py

alvations:
Fixing #1080


stevenbird:
Thanks very much @alvations.

PR #1098, opened 2015-08-28
Implement IBM Model 4

hoontw:
Other noteworthy refactorings and fixes:
- Modify sampling algorithm to that of the original paper by Brown et al.
- More unit tests
- Change target language sentence to be 1-indexed internally, instead of 0-indexed
- Uniform initialization of Model 3 distortion probabilities
- Move distortion table from abstract IBMModel class to Model3 class, since it is only used there

Generation of word classes is not covered in this feature branch. The user is expected to provide a data structure with the words labelled with their classes.


stevenbird:
Thanks @hoontw – solid work.

PR #1100, opened 2015-08-31
Implement IBM Model 5

hoontw:
Other changes:
- Extract common methods for Models 4 and 5
- Remove hacky way of running tests from console


stevenbird:
Thanks @hoontw – great to have model 5!

PR #1107, opened 2015-09-04
Change English code from 'en' to 'eng' and a bug in lemma count() fun…

letuananh:
Change English code from 'en' to 'eng' and a bug in lemma count() function (currently lemma counting only works in English)

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -252,7 +252,7 @@ class Lemma(_WordNetObject):
         self._frame_ids = []
         self._lexname_index = lexname_index
         self._lex_id = lex_id
-        self._lang = "en"
+        self._lang = 'eng'
 
         self._key = None # gets set later.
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Why do you propose to change this code from ISO 639-1?

alvations:
I think it's because Open Multilingual Wordnet is using the 3 characters language code.

stevenbird:
Thanks @alvations. @fcbond do you have any advice please?

fcbond:
G'day,

as Liling said, all the languages except English use the three letter
codes.   As English is the default, and this doesn't change, almost no
behaviour changes here.  And the bug in lemma.count() makes it hard to
write monolingual applications.

On Sat, Sep 5, 2015 at 12:34 AM, Steven Bird notifications@github.com
wrote:

> Thanks @alvations https://github.com/alvations. @fcbond
> https://github.com/fcbond do you have any advice please?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/nltk/nltk/pull/1107#issuecomment-137784758.

## 

Francis Bond http://www3.ntu.edu.sg/home/fcbond/
Division of Linguistics and Multilingual Studies
Nanyang Technological University

stevenbird:
I see, thanks @fcbond

PR #1111, opened 2015-09-06
Multi-word expression tokenizer

rmalouf:
Given a lexicon of MWEs, merge instances of MWEs in tokenized text into single tokens:

```
    >>> tokenizer = MWETokenizer([('hors', "d'oeuvre")])
    >>> tokenizer.tokenize("An hors d'oeuvre tonight, sir?".split())
    ['An', "hors_d'oeuvre", 'tonight,', 'sir?']
```

### DIFF
--- a/nltk/tokenize/__init__.py
+++ b/nltk/tokenize/__init__.py
@@ -71,7 +71,8 @@ from nltk.tokenize.sexpr    import SExprTokenizer, sexpr_tokenize
 from nltk.tokenize.treebank import TreebankWordTokenizer
 from nltk.tokenize.stanford import StanfordTokenizer
 from nltk.tokenize.texttiling import TextTilingTokenizer
-from nltk.tokenize.casual import (TweetTokenizer, casual_tokenize)
+from nltk.tokenize.casual   import (TweetTokenizer, casual_tokenize)
+from nltk.tokenize.mwe      import MWETokenizer
 
 # Standard sentence tokenizer.
 def sent_tokenize(text, language='english'):
--- a/nltk/tokenize/mwe.py
+++ b/nltk/tokenize/mwe.py
@@ -66,7 +66,7 @@ class MWETokenizer(TokenizerI):
         We represent the trie as a dict of dicts:
 
             >>> tokenizer = MWETokenizer([('a', 'b'), ('a', 'b', 'c'), ('a', 'x')])
-            >>> print tokenizer._mwes
+            >>> tokenizer._mwes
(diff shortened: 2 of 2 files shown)


stevenbird:
Great idea @rmalouf.

I wonder if it's worth putting the trie functionality in nltk.util, and providing both a pure python implementation and wrapping a third-party one. I see that @kmike has written about this: http://kmike.ru/python-data-structures/

nschneid:
Any thoughts on what [multiword expression](http://people.eng.unimelb.edu.au/tbaldwin/pubs/handbook2009.pdf) dictionary would be preferable? Just foreign phrases, or also multiword named entities (_North Dakota_), compound nominals (_traffic light_), compound prepositions (_instead of_), idiomatic verbal expressions (_make up_, _watch out_, _come across_, _pay attention_, _pull strings_, _kick the bucket_), fixed utterances (_You're welcome_)? Would it be able to match MWEs containing gaps, or just contiguous ones?

I have built a Python/Cython MWE identifier ([AMALGrAM](https://github.com/nschneid/pysupersensetagger)) that uses a discriminative sequence model. For feature extraction there is a lexicon-matching [segmenter implementation](https://github.com/nschneid/pysupersensetagger/blob/master/src/segment.py) that allows for gaps within the MWE.

ewan-klein:
For both parsing and clustering, being able to take the output of NER would be great.

rmalouf:
For working with NER output, would you want to use a MWE lexicon, or rather retokenize a chunked sentence?

alvations:
- CLIPS pattern has a list of entities: https://github.com/clips/pattern/blob/master/pattern/text/en/en-entities.txt
- Most likely WordNet's MWE should be included in that list too.
- Also, taking Freebaes list would be great but it's most likely humongous 
- Extract NNPs from existing tagged resources in NLTK would also be nice.

---

Any thoughts on also inducing a list of MWE like what `gensim` does? https://github.com/piskvorky/gensim/blob/develop/gensim/models/phrases.py

rmalouf:
I have a simple term extraction module that I'll contribute soon.  

It's hard to say what's the right source for MWEs without knowing the application -- for what I'm doing, a pre-built lexicon would be useless, but that's just me.  Probably having a couple of general lists available in nltk_data (like the stopword lists) would be helpful for people getting started.

ewan-klein:
Although telling an NER system to do re-tokenization (rather than tagging) is a good idea, there may be applications such as microblogs where there's not enough context to do good feature-based NER and where a lexicon-based approach _might_ yield better results for specific domains.

PR #1112, opened 2015-09-06
Multext east corpus reader

stieglma:
Added a corpus reader for the newly introduced multext-east corpus. (nltk_data#30)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -20,6 +20,7 @@
 - Robert Berwick
 - Dan Blanchard
 - Nathan Bodenstab
+- Alexander Böhm
 - Francis Bond
 - Paul Bone
 - Jordan Boyd-Graber
@@ -151,6 +152,7 @@
 - Ceri Stagg
 - Peter Stahl
 - Oliver Steele
+- Thomas Stieglmaier
 - Jan Strunk
 - Liling Tan
 - Claire Taylor
@@ -169,6 +171,7 @@
 - Patrick Ye
 - Geraldine Sim Wei Ying
 - Jason Yoder
+- Thomas Zieglier
 - 0ssifrage
 - ducki13
 - kiwipi
--- a/nltk/corpus/__init__.py
+++ b/nltk/corpus/__init__.py
@@ -147,6 +147,8 @@ movie_reviews = LazyCorpusLoader(
     'movie_reviews', CategorizedPlaintextCorpusReader,
     r'(?!\.).*\.txt', cat_pattern=r'(neg|pos)/.*',
     encoding='ascii')
+multext_east = LazyCorpusLoader(
+    'mte_teip5', MTECorpusReader, r'(oana).*\.xml', encoding="utf-8")
 names = LazyCorpusLoader(
     'names', WordListCorpusReader, r'(?!\.).*\.txt', encoding='ascii')
 nkjp = LazyCorpusLoader(
--- a/nltk/corpus/reader/__init__.py
+++ b/nltk/corpus/reader/__init__.py
@@ -96,6 +96,7 @@ from nltk.corpus.reader.sentiwordnet import *
 from nltk.corpus.reader.twitter import *
 from nltk.corpus.reader.nkjp import *
 from nltk.corpus.reader.crubadan import *
+from nltk.corpus.reader.mte import *
 
 # Make sure that nltk.corpus.reader.bracket_parse gives the module, not
 # the function bracket_parse() defined in nltk.tree:
@@ -131,5 +132,6 @@ __all__ = [
     'TimitTaggedCorpusReader', 'LinThesaurusCorpusReader',
     'SemcorCorpusReader', 'FramenetCorpusReader', 'UdhrCorpusReader',
     'BNCCorpusReader', 'SentiWordNetCorpusReader', 'SentiSynset',
-    'TwitterCorpusReader', 'NKJPCorpusReader', 'CrubadanCorpusReader'
+    'TwitterCorpusReader', 'NKJPCorpusReader', 'CrubadanCorpusReader',
+    'MTECorpusReader'
 ]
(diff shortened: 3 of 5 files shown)


stevenbird:
@stieglma would you mind adding some brief information about this corpus and examples to `test/corpus.doctest` please?

stevenbird:
Thanks @stieglma

stevenbird:
@stieglma – mte.py breaks in Python 3.4. Would you please submit a patch?

ewan-klein:
A couple of other issues:
- the reader introduces a new dependency, namely `lxml`. This should at least be wrapped in a try/except block
- PEP8 advises against wildcard imports

stieglma:
I'm not sure if I can do it today, but at least tomorrow it should be fixed.
What should i be doing about the lxml import? Putting in into a try/except and therefore catching the ImportError does not help, as it would only swallow the exception. If this is the way to go, should I instead print a warning like `For the Multext-East corpus the lxml package is needed, please install it.`. And if yes, is there a logging utility especially for such things in nltk, or should I simply use a print statement?

jwacalex:
Python 3.4 breaks because of PEP 3113.  We are doing tuple unpacking at `map (lambda (w,t) : (w, MTETagConverter.msd ]…`. I've started a new [branch](https://github.com/nltk/nltk/compare/develop...jwacalex:fixes_1112) to fix  the issues.

kmike:
@stieglma how hard would it be to replace lxml with https://docs.python.org/2/library/xml.etree.elementtree.html?

jwacalex:
@kmike We have decided to use lxml instead of etree, because parsing the files with etree is much slower then lxml.

kmike:
@jwacalex what about using lxml when it is available and falling back to ElementTree if it is not available? Their interfaces are very close to each other.

jwacalex:
@kmike would be an idea to do so. @stieglma and I've to evaluate if a simple swap is possible or there are two calls to be managed

stevenbird:
Our [CI Server](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py34-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.corpus.reader.mte/MTECorpusReader/__init__/) points out a problem with the way the `MTECorpusReader` initializer is being called.

jwacalex:
`__init__(self, root=None, fileids=None, encoding='utf8'):` is the signature of `MTECorpusReader`. The desired tagset is specified not while initialising the corpus reader but while getting the data.

For example refer to the signature of `tagged_words`: `(self, fileids=None, tagset="msd", tags=None)`

stieglma:
@stevenbird we can introduce the `tagset` parameter to the init method of the `MTECorpusReader` but there are still other parameters in the init method of `TaggedCorpusReader` that make no sense at all for our corpus reader, do they have to be there, too? If yes then we should probably think about extending `CorpusReader` insteadof `TaggedCorpusReader`

stieglma:
I just saw that the example in the docstring of the init method is wrong, there we have the parameter for the tagset, which is actually not really there (see line 5, last parameter of the corpus reader, it should definitely be removed), can this lead to the unit-test error? (I am quite surprised that there even is one, as we did not submit any unit tests with our pull requests, so I assume you parse them somehow out of the docstring of the methods.)
Here is the docstring of the init method, if it is enough to fix the sample then I think this should be the way to go.

``` python
"""
Construct a new MTECorpusreader for a set of documents
located at the given root directory.  Example usage:
    >>> root = '/...path to corpus.../'
    >>> reader = MTECorpusReader(root, 'oana-*.xml', 'utf8', "msd") 
:param root: The root directory for this corpus. (default points to location in multext config file)
:param fileids: A list or regexp specifying the fileids in this corpus. (default is oana-en.xml)
:param enconding: The encoding of the given files (default is utf8)
"""
```

I am not there from today in the evening up to 20th of september, so I will most likely commit the fix not before 21st, I hope this is ok.

stieglma:
@stevenbird did you fix the error yourself or are you still waiting for us? If you are waiting for us, it would be great if you could answer the questions of my last two messages here, as it is still unclear for me where the error comes from, and how I can fix it best.

stevenbird:
@stieglma – thanks for following up. Tox runs the doctests, which is why we see an error (see below). It would be enough to fix the sample, thanks.

```
Failed example:
    reader = MTECorpusReader(root, 'oana-*.xml', 'utf8', "msd") 
Exception raised:
      File "/scratch/jenkins/python/python-3.5.0-x86_64/lib/python3.5/doctest.py", line 1320, in __run
        compileflags, 1), test.globs)
      File "<doctest nltk.corpus.reader.mte.MTECorpusReader.__init__[1]>", line 1, in <module>
        reader = MTECorpusReader(root, 'oana-*.xml', 'utf8', "msd")
    TypeError: __init__() takes from 1 to 4 positional arguments but 5 were given
```

stieglma:
@stevenbird  The docstring is fixed with #1142.
But I have another question: I still can't find the Multext-East Corpus in the NLTK downloader, although it was integrated some weeks ago. Did I miss a step or is there some cache that simply has to be cleared?

stevenbird:
@stieglma – I agree, this is unexpected. The MTE entry shows up in [the index](https://github.com/nltk/nltk_data/blob/gh-pages/index.xml#L40), but not (yet) on the [published version](nltk.github.com/nltk_data/). I've pushed an [empty commit](https://github.com/nltk/nltk_data/commit/f73cbc4b9b0ecce0c87b1f09211dcf6a8b567176) to rebuild the pages, to no avail.

stieglma:
@stevenbird thanks, Multext-East is now listed in the downloader, and the download works. However, after the download finished, the zipfile is not unzipped, which is necessary for our corpus reader, I guess `unzip="0"` should be replaced with `unzip="1"` in the xml file.

ewan-klein:
I updated the `unzip` value in e42ec2f. It seems to unzip OK for me now.

PR #1049, opened 2015-07-21
Allow variable top relation label in a dependency graph.

dimazest:
This is  a possible (partial) fix for #944

Issue #620, opened 2014-02-22
ParserI interface


stevenbird:
The behavior of `ParserI.parse` is to return the first (or best) parse or else `None` if no parses exist, while `iter_parse` iterates over all parses, and `nbest_parse` returns a list of n parses. All implementations of `nbest_parse` find all parses before pruning the set. I propose to simplify these to a single `parse` method.

heatherleaf:
Fixed in pull request #627

heatherleaf:
I suggest that we keep the `n` argument from `nbest_parse` in the `parse` method:

```
def parse(self, sent, n=None):
    """
    :param n: The maximum number of trees to return.
    :type n: int
    """
```

stevenbird:
We might want to use a decorator to achieve this (in `api.py`), and that decorator may be useful for any NLTK method that returns an iterator.

heatherleaf:
Good idea. Can such an iterator add to the docstring too, and would we want that?

stevenbird:
It would be good to update the docstring too. Looks like that's not too difficult: http://stackoverflow.com/questions/13314054/python-decorator-that-maintains-signature-allows-modification-to-docstring-and

stevenbird:
However, there's another argument that a programmer can very well select the first n from an iterator without built-in support.

heatherleaf:
We already have a function `internals._add_epytext_field()` for this...

heatherleaf:
Yes, the programmer can select the first n, but then s/he needs to import itertools:

```
itertools.islice(myparser.parse(sent), N)
```

Not a big problem, but an unexperienced programmer might well try this instead, which is _bad_:

```
list(myparser.parse(sent))[:N]
```

As a comparison:

```
myparser.parse(sent, n=N)
```

stevenbird:
OK I'm convinced this is a good idea.

heatherleaf:
Here's a decorator that works, but it has one problem:

```
def nbest_iterator(f):
    @functools.wraps(f)
    def wrapper(*args, **kwargs):
        n = kwargs.pop('n', None)
        iterator = f(*args, **kwargs)
        if n is not None:
            iterator = itertools.islice(iterator, n)
        return iterator
    _add_epytext_field(wrapper, 'n', "return at most this many results")
    return wrapper
```

Here's an example:

```
>>> @nbest_iterator
... def string_iterator(s):
...     """
...     Converts a string to an iterator
...     @s: the string
...     """
...     return iter(s)
...
[...]

PR #1114, opened 2015-09-06
Stanford dependency parser

rmalouf:
Modify interface to the Stanford Parser to return either constituency trees or typed
dependency graphs.


nltk/parse/dependencygraph.py
```diff
@@ -334,6 +334,7 @@ def extract_10_cells(cells):
             # Make sure that he fake root node has labeled dependencies.
             if (cell_number == 3) and (head == 0):
                 rel = 'ROOT'
+            if rel == 'root': rel = 'ROOT'  # Stanford parser doesn't capitalize ROOT
```
dimazest: Could you please split it to several lines (according to pap8 guidelines):

``` python
# Stanford parser doesn't capitalize ROOT
if rel == 'root':
    rel = 'ROOT'  
```

stevenbird:
@rmalouf – nice to have the extra doctests, thanks.

nltk/parse/dependencygraph.py
rmalouf: Fixed

stevenbird:
@rmalouf – this might be the first instance of a parser which can return phrase-structure trees or dependency trees, which means I'd like to be sure that ad hoc polymorphism is the best way to go (instead of, say, providing `parse` and `depparse` methods or some such solution). Do you want to argue for the polymorphism approach? @heatherleaf do you have an opinion, in light of your earlier input to the parser API in #620?

rmalouf:
No, I don't.  The version of this that I've been using actually had two classes, StanfordParser and StanfordDependencyParser.  I combined them just to minimize the changes, but I'm happy to split them again.  Any preferences?

stevenbird:
I like that solution. So two classes both inheriting `ParserI`, but where `StanfordParser.parse` vs `StanfordDependencyParser.parse` return `Tree` vs `DependencyGraph`.

nltk/parse/stanford.py
```diff
@@ -237,6 +188,114 @@ def _execute(self, cmd, input_, verbose=False):
...
+    ... ))],[]) # doctest: +NORMALIZE_WHITESPACE
+    [[((u'jumped', u'VBD'), u'nsubj', (u'fox', u'NN')), ((u'fox', u'NN'), u'det', (u'The', u'DT')),
+    ((u'fox', u'NN'), u'amod', (u'quick', u'JJ')), ((u'fox', u'NN'), u'amod', (u'brown', u'JJ')),
+    ((u'jumped', u'VBD'), u'nmod', (u'dog', u'NN')), ((u'dog', u'NN'), u'case', (u'over', u'IN')),
+    ((u'dog', u'NN'), u'det', (u'the', u'DT')), ((u'dog', u'NN'), u'amod', (u'lazy', u'JJ'))]]
+
+    """
+
+    _OUTPUT_FORMAT = 'conll2007'
+    _TO_TREE = DependencyGraph
```
dimazest: this assignment look strange, there might be problems due to bound methods. Maybe defining a method is a nicer way of doing this.

nltk/parse/dependencygraph.py
```diff
@@ -334,6 +334,9 @@ def extract_10_cells(cells):
             # Make sure that he fake root node has labeled dependencies.
             if (cell_number == 3) and (head == 0):
                 rel = 'ROOT'
+            # Stanford parser doesn't capitalize ROOT
```
dimazest: Actually, #1049 might be a cleaner solution.

nltk/parse/stanford.py
rmalouf: Strange how?

nltk/parse/dependencygraph.py
rmalouf: That would be better! But, it looks like only malt.py got changed and the necessary changes to dependencygraph.py didn't make it in to #1049.

dimazest: That's strange, I've created #1131 to restore that code.

stevenbird:
Thanks @rmalouf

PR #1115, opened 2015-09-07
Speed up collocations

rmalouf:
With the new FreqDist implementation based on Counter, N() is fairly expensive, and the collocation finding methods were calling it for every ngram.  This made the collocation finders unusably slow, but caching N speeds things up dramatically.


stevenbird:
Great idea, thanks

PR #1118, opened 2015-09-07
MNT: pep8ify files

pratapvardhan:
Made pep8 changes only to 
- _init_.py
- book.py
- collocations.py
- compat.py
- data.py

Plenty of other files, would need pep8. However, I thought, that might break other open PRs.

In any case, let me know, if you're hard on pep8, I can clean up few more files before you can merge **or** if you think pep8ification is unnecessary, please go ahead and close this.

### DIFF
--- a/nltk/__init__.py
+++ b/nltk/__init__.py
@@ -19,9 +19,9 @@ from __future__ import print_function, absolute_import
 
 import os
 
-##//////////////////////////////////////////////////////
-##  Metadata
-##//////////////////////////////////////////////////////
+# //////////////////////////////////////////////////////
+# Metadata
+# //////////////////////////////////////////////////////
 
 # Version.  For each new release, the version number should be updated
 # in the file VERSION.
@@ -35,7 +35,7 @@ except NameError:
 except IOError as ex:
     __version__ = "unknown (%s)" % ex
 
-if __doc__ is not None: # fix for the ``python -OO``
+if __doc__ is not None:  # fix for the ``python -OO``
     __doc__ += '\n@version: ' + __version__
 
 
@@ -84,7 +84,7 @@ __classifiers__ = [
     'Topic :: Text Processing :: General',
     'Topic :: Text Processing :: Indexing',
     'Topic :: Text Processing :: Linguistic',
-    ]
+]
 
 from nltk.internals import config_java
 
@@ -178,6 +178,7 @@ from nltk import align, ccg, chunk, classify, collocations
 from nltk import data, featstruct, grammar, help, inference, metrics
 from nltk import misc, parse, probability, sem, stem, wsd
 from nltk import tag, tbl, text, tokenize, tree, treetransforms, util
+
 
 # override any accidentally imported demo
 def demo():
--- a/nltk/collocations.py
+++ b/nltk/collocations.py
@@ -57,27 +57,27 @@ class AbstractCollocationFinder(object):
         self.ngram_fd = ngram_fd
 
     @classmethod
-    def _build_new_documents(cls,documents, window_size, pad_left=False, pad_right=False, pad_symbol=None):
+    def _build_new_documents(cls, documents, window_size, pad_left=False, pad_right=False, pad_symbol=None):
         '''
-        Pad the document with the place holder according to the window_size 
+        Pad the document with the place holder according to the window_size
         '''
         padding = (pad_symbol,) * (window_size - 1)
         if pad_right:
             return _itertools.chain.from_iterable(_itertools.chain(doc, padding) for doc in documents)
         if pad_left:
             return _itertools.chain.from_iterable(_itertools.chain(padding, doc) for doc in documents)
-    
+
     @classmethod
     def from_documents(cls, documents):
         """Constructs a collocation finder given a collection of documents,
         each of which is a list (or iterable) of tokens.
         """
         #return cls.from_words(_itertools.chain(*documents))
-        return cls.from_words(cls._build_new_documents(documents, cls.default_ws, pad_right=True)) 
+        return cls.from_words(cls._build_new_documents(documents, cls.default_ws, pad_right=True))
 
     @staticmethod
     def _ngram_freqdist(words, n):
-        return FreqDist(tuple(words[i:i+n]) for i in range(len(words)-1))
+        return FreqDist(tuple(words[i:i + n]) for i in range(len(words) - 1))
 
     def _apply_filter(self, fn=lambda ngram, freq: False):
         """Generic filter removes ngrams from the frequency distribution
@@ -140,8 +140,8 @@ class BigramCollocationFinder(AbstractCollocationFinder):
     association measures. It is often useful to use from_words() rather than
     constructing an instance directly.
     """
-    default_ws = 2 
-    
+    default_ws = 2
+
     def __init__(self, word_fd, bigram_fd, window_size=2):
         """Construct a BigramCollocationFinder, given FreqDists for
         appearances of words and (possibly non-contiguous) bigrams.
@@ -163,7 +163,7 @@ class BigramCollocationFinder(AbstractCollocationFinder):
 
         for window in ngrams(words, window_size, pad_right=True):
             w1 = window[0]
-            if w1 is None: 
+            if w1 is None:
                 continue
             wfd[w1] += 1
             for w2 in window[1:]:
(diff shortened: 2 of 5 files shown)


nltk/data.py
```diff
@@ -1111,13 +1140,14 @@ def readline(self, size=None):
             if len(lines) > 1:
                 line = lines[0]
                 self.linebuffer = lines[1:]
-                self._rewind_numchars = len(new_chars)-(len(chars)-len(line))
+                self._rewind_numchars = len(
+                    new_chars) - (len(chars) - len(line))
```
kmike: Such change makes readability worse, not better: the expression is split at a random point. Have you used some automated tools? 

Instead of splitting long expressions into multi lines it is usually better to introduce local varaibles. Here a variable for `len(chars)-len(line)` can be introduced - what should be its name?

nltk/data.py
```diff
@@ -335,7 +346,7 @@ class BufferedGzipFile(GzipFile):
     ``BufferedGzipFile`` is useful for loading large gzipped pickle objects
     as well as writing large encoded feature files for classifier training.
     """
-    SIZE = 2 * 2**20
+    SIZE = 2 * 2 ** 20
```
kmike: I kind of liked `2 * 2**20` more because operation priorities are more clear this way. Maybe it worths it to add a variable:

```
MB = 2 ** 20
SIZE = 2 * MB
```

nltk/data.py
```diff
@@ -236,6 +243,7 @@ def normalize_resource_name(resource_name, allow_relative=True, relative_path=No
 ######################################################################
 
 class PathPointer(object):
+
```
kmike: pep8 doesn't require a blank line before a docstring, and pep 0257 says: "There's no blank line either before or after the docstring. "

nltk/data.py
```diff
@@ -212,7 +217,8 @@ def normalize_resource_name(resource_name, allow_relative=True, relative_path=No
     >>> windows or normalize_resource_name('/dir/file', True, '/') == '/dir/file'
     True
     """
-    is_dir = bool(re.search(r'[\\/.]$', resource_name)) or resource_name.endswith(os.path.sep)
+    is_dir = bool(
+        re.search(r'[\\/.]$', resource_name)) or resource_name.endswith(os.path.sep)
```
kmike: This change makes code less clear because the expression is split at a random point. With this indentation I'd expect bool() to be applied to the whole expression.

pratapvardhan:
Fair points @kmike . I ought to be careful using autopep8 then!

nltk/data.py
```diff
@@ -72,7 +72,8 @@
    (e.g., in their home directory under ~/nltk_data)."""
 
 # User-specified locations:
-path += [d for d in os.environ.get('NLTK_DATA', str('')).split(os.pathsep) if d]
+path += [d for d in os.environ.get('NLTK_DATA',
+                                   str('')).split(os.pathsep) if d]
```
kmike: I think it is not good to split the line at this point - it is not clear which expression is an argument of which function. Add a local variable, e.g like this:

```
_paths_from_env = os.environ.get('NLTK_DATA', str('')).split(os.pathsep)
path += [d for d in _paths_from_env if d]
```

nltk/compat.py
```diff
@@ -544,6 +568,8 @@ def add_py3_data(path):
 
 # for use in adding /PY3 to the second (filename) argument
 # of the file pointers in data.py
+
+
 def py3_data(init_func):
```
kmike, line 561: Comments belong to the function; it is better to either add a blank line above the comment or convert comments to a docstring

nltk/compat.py
```diff
@@ -264,7 +277,8 @@ def _rmtree(self, path):
             for name in self._listdir(path):
                 fullname = self._path_join(path, name)
                 try:
-                    isdir = self._isdir(fullname) and not self._islink(fullname)
+                    isdir = self._isdir(
+                        fullname) and not self._islink(fullname)
```
kmike: weird line split

nltk/compat.py
```diff
@@ -171,7 +183,8 @@ def __init__(self, f, dialect=csv.excel, encoding="utf-8", errors='replace', **k
             self.queue = cStringIO.StringIO()
             self.writer = csv.writer(self.queue, dialect=dialect, **kwds)
             self.stream = f
-            self.encoder = codecs.getincrementalencoder(encoding)(errors=errors)
+            self.encoder = codecs.getincrementalencoder(
+                encoding)(errors=errors)
```
kmike:
```
encoder_cls = codecs.getincrementalencoder(encoding)
self.encoder = encoder_cls(errors=errors) 
```

nltk/collocations.py
```diff
@@ -363,7 +367,8 @@ def demo(scorer=None, compare_scorer=None):
         print('\t', [' '.join(tup) for tup in cf.nbest(scorer, 15)])
         print('\t Correlation to %s: %0.4f' % (compare_scorer.__name__,
                                                spearman_correlation(
-                                                   ranks_from_scores(cf.score_ngrams(scorer)),
+                                                   ranks_from_scores(
+                                                       cf.score_ngrams(scorer)),
                                                    ranks_from_scores(cf.score_ngrams(compare_scorer)))))
```
kmike: I think it is better to compute spearman_correlation before string formatting

nltk/collocations.py
```diff
@@ -66,18 +67,18 @@ def _build_new_documents(cls,documents, window_size, pad_left=False, pad_right=F
...
-    
+
     @classmethod
     def from_documents(cls, documents):
         """Constructs a collocation finder given a collection of documents,
         each of which is a list (or iterable) of tokens.
         """
-        #return cls.from_words(_itertools.chain(*documents))
-        return cls.from_words(cls._build_new_documents(documents, cls.default_ws, pad_right=True)) 
+        # return cls.from_words(_itertools.chain(*documents))
```
kmike: There is a convention: for commented out lines there is no space between `#` and code; it allows to visually distinguish them from comments.

nltk/__init__.py
```diff
@@ -180,5 +180,7 @@ def _fake_Popen(*args, **kwargs):
 from nltk import tag, tbl, text, tokenize, tree, treetransforms, util
 
 # override any accidentally imported demo
+
+
 def demo():
```
kmike, line 184: comment belongs to a function, it is better to add blank line above the comment

pratapvardhan:
@kmike -- weirdly enough I haven't seen autopep8 go this bad before -- should have seen the changes more closely!

kmike:
Thanks @pratapvardhan! I think code can benefit from a cleanup; +1 to fix whitespace issues and make other changes. But it is better to do it manually; automated tools don't understand the code well enough :)

pratapvardhan:
Thanks @kmike for painfully going through the changes, I'll send a patch in a while. Manually this time =)

nltk/collocations.py
stevenbird: +1

nltk/collocations.py
stevenbird: Agreed – I find this helpful.

nltk/compat.py
stevenbird: Yes, this is much more readable

stevenbird:
Thanks for your contribution @pratapvardhan. In general, clean-ups are very welcome. There's just some subtleties...

pratapvardhan:
@kmike -- committed the patch.

nltk/compat.py
```diff
@@ -474,36 +483,43 @@ def iterkeys(d):
...
             '__gt__': [('__lt__', lambda self, other: not (self > other or self == other)),
-                       ('__ge__', lambda self, other: self > other or self == other),
+                       ('__ge__', lambda self, other: self >
+                        other or self == other),
                        ('__le__', lambda self, other: not self > other)],
             '__ge__': [('__le__', lambda self, other: (not self >= other) or self == other),
-                       ('__gt__', lambda self, other: self >= other and not self == other),
+                       ('__gt__', lambda self, other: self >=
+                        other and not self == other),
                        ('__lt__', lambda self, other: not self >= other)]
```
kmike, line 513: total_ordering decorator was a verbatim copy from stdlib: https://hg.python.org/cpython/file/2.7/Lib/functools.py#l53 - I think we shouldn't change indentation

pratapvardhan, line 513: Thought about it before if I should revert or not. Reverted to old style as is.

stevenbird:
Looks better, thanks @pratapvardhan. There's just a couple of small things to fix still I think. Thanks.

pratapvardhan:
@stevenbird - changed to `corr`, any other pressing parts? would you want me to rebase commits into one?

stevenbird:
Thanks @pratapvardhan.

stevenbird:
Thanks for the good work @pratapvardhan

PR #1121, opened 2015-09-08
max phrase length added

barancik:
Added maximal length of extracted phrases, one redundant variable deleted.

### DIFF
--- a/nltk/align/phrase_based.py
+++ b/nltk/align/phrase_based.py
@@ -8,7 +8,7 @@
 
 def extract(f_start, f_end, e_start, e_end, 
             alignment, f_aligned,
-            srctext, trgtext, srclen, trglen, max_phrase_lenght):
+            srctext, trgtext, srclen, trglen, max_phrase_length):
     """
     This function checks for alignment point consistency and extracts 
     phrases using the chunk of consistent phrases.
@@ -62,7 +62,7 @@ def extract(f_start, f_end, e_start, e_end,
     phrases = set()
     fs = f_start
     while True:
-        fe = min(f_end, f_start + max_phrase_lenght - 1)
+        fe = min(f_end, f_start + max_phrase_length - 1)
         while True:
             # add phrase pair ([e_start, e_end], [fs, fe]) to set E
             # Need to +1 in range  to include the end-point.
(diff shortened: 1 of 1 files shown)


alvations:
@barancik thanks for the improvement.

@stevenbird, @barancik has made one of the suggested improvement to the NLTK code as listed on https://github.com/alvations/mt4nltk. 

The +1 , -1 indices in the `phrase_extraction` is a little confusing but that was the attempt to implement the Koehn's algorithm verbatim. Possibly the next improvement is the remove the +1 -1 indices in the code and make it more natural to read.

Should we make the changes for the indices before merging? Or merge the max phrase length added from this PR and make another PR for the indices changing?

alvations:
@stevenbird , the default max_phrase_length 7 is because it's Moses default. But after continuing work on the phrase extraction. I realize setting the default to -1 to extract all phrases would be more appropriate.

Suggesting change from (@barancik  PR):

```
def phrase_extraction(srctext, trgtext, alignment, max_phrase_length=7):
    # ... 
        # Index e_end from e_start to len(e) - 1 or the length of maximum phrase
        max_idx = min(srclen, e_start + max_phrase_length)
```

to:

```
def phrase_extraction(srctext, trgtext, alignment, max_phrase_length=-1):
    # ... 
        if max_phrase_length is -1: # Extracts phrases of all length.
            max_idx = max_phrase_length = srclen
        else:
            # Index e_end from e_start to len(e) - 1 or the length of maximum phrase
            max_idx = min(srclen, e_start + max_phrase_length)
```

When emulating Moses, users should set it to 7. We'll add that as a note in the docstring.

I'm adding more features to the phrase extraction so that we can get phrase probabilities and lexicalized weights and reordering from the internal alignments  The max_phrase_length is necessary to reduce vocab size `|V|` so that the smoothing don't get to negligible. 

**Should the default `max_phrase_length` be set to -1 or 0?**

stevenbird:
@alvations, note that we now presume at least Python 2.7, so `xrange` is fine.

stevenbird:
Thanks @barancik. I'm going to merge this, and change back to using xrange.

alvations:
@stevenbird , thanks for the note and the merge!

PR #1123, opened 2015-09-08
fixes problems stated in nltk/issue #1112

jwacalex:
fixing the issues in  #1112: lxml is optional, python3.4 compatible (pep 3113)


stieglma:
The import of CorpusReader was even not necessary as we are extending TaggedCorpusReader. The import of the latter should now be according to the guidelines.
Additionally I implemented the readme() method with some functionality, before it was only an empty method.

stevenbird:
Thanks @stieglma

PR #1127, opened 2015-09-11
Added features to nltk.align.util and nltk.align.phrase_based

alvations:
- Changed the function name to match the other `alignment2pharaohtext()`
- Removed extra comment that is already in the docstring of the function.
- Remove the not-so useful slice position of phrases in `phrase_extraction()` output.
- Added the phrase-internal lexical alignments to `phrase_extraction()` output.
- Change the doctest for `phrase_extraction()` accordingly

### DIFF
--- a/nltk/align/phrase_based.py
+++ b/nltk/align/phrase_based.py
@@ -6,6 +6,22 @@
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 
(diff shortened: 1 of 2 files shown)


alvations:
Shouldn't be merged, we're merging some other code that changes the same file. Another PR will be requested later. Sorry for the complication.

PR #1129, opened 2015-09-11
Adjusted the function name (pharaohtext2alignment).

AnnaGarbar:


stevenbird:
I think it would be good making this a factory function, e.g. `Alignment.fromstring()`, cf all the other `fromstring` methods we have.

stevenbird:
How about something like the following:

```
    @classmethod
    def fromstring(cls, s):
        """
        Read a giza-formatted string and return an Alignment object.

            >>> Alignment.fromstring('0-0 2-1 9-2 21-3 10-4 7-5')
            Alignment([(0, 0), (2, 1), (7, 5), (9, 2), (10, 4), (21, 3)])

        :type s: str
        :param s: the positional alignments in giza format
        :rtype: Alignment
        :return: An Alignment object corresponding to the string representation ``s``.
        """

        return Alignment([_giza2pair(a) for a in s.split()])
```

Note that the existing print method for `Alignment` already does the work proposed in `alignment2pharaohtext`.

stevenbird:
There's an issue with the `Alignment` initializer which currently accepts a string or a list. We would be changing the API to remove the polymorphism:

```
>>> a = Alignment([(0,0), (1,1)])
>>> b = Alignment.fromstring("0-0 1-1")
```

I prefer this, but there's consequences for the way that `AlignedSent` is initialized, and for some methods like `AlignedSent.precision` which permit a list or string gold alignment. Here, I would prefer the calling program to explicitly do the conversion where it is required. I think it's better practice for programmers to be aware of the types of things, and avoid the risk of unnecessary type-checks in the library.

```
>>> alignedsent.precision(list_or_string) # current
>>> alignedsent.precision(alignment) # proposed
>>> alignedsent.precision(Alignment(list)) # proposed
>>> alignedsent.precision(Alignment.fromstring(string)) # proposed
```

alvations:
Removing polymorphism looks good, I would prefer either:

```
>>> alignedsent.precision(alignment) # proposed
>>> alignedsent.precision(Alignment.fromstring(string)) # proposed
```

Possibly, we might move to [type hints](https://docs.python.org/3.6/library/typing.html#module-typing) that @gvanrossum was [talking about](https://www.youtube.com/watch?v=2wDvzy6Hgxg) with py3.5  but that might break in py2.7+.

If it's alright, I'll "take over" some of the coding for the `nltk.align.api` since the hackathon ended and @AnnaGarbar will help out whenever she's free.

alvations:
@stevenbird the polymorphism `AlignedSent.precision()` is rather "special" since it type cast the input (list/string) into Alignment before calculating the precision. So it's allowing `Alignment`, string or list and that prevents breakage from different input types.

If we only allow Alignment objects as input then we can simply remove https://github.com/alvations/nltk/blob/develop/nltk/align/api.py#L182 and L183. 

I'll check which other functions/classes interacts with the `Alignment` and list them out later this week.

hoontw:
Other than in api.py, I haven't seen any use of the string argument version of `Alignment` (or `precision`) in the align module.

As a side note, another way to implement the precision family of methods is static methods. Something like

```
AlignedSent.precision(reference_sentence, test_sentence)
```

alvations:
There're several options but I'm not sure which one is the preferred.

Original with "super" polymorphism, allows `AlignedSent`, `Alignment`, str or list(str).

```
def precision(self, reference):
    """
    Return the precision of an aligned sentence with respect to a
    "gold standard" reference ``AlignedSent``.

    :type reference: AlignedSent or Alignment
    :param reference: A "gold standard" reference aligned sentence.
    :rtype: float or None
    """
    # Get alignments in set of 2-tuples form
    # The "possible" precision is used since it doesn't penalize for finding
    # an alignment that was marked as "possible" (NAACL corpus)

    align = self.alignment
    if isinstance(reference, AlignedSent):
        possible = reference.alignment
    else:
        possible = Alignment(reference)

    return precision(possible, align)
```

Only allows `Alignment` object:

```
def precision(self, reference):
    """
    Return the precision of an aligned sentence with respect to a
    "gold standard" reference ``Alignment``.

    :type reference: Alignment
    :param reference: A "gold standard" reference ``Alignment``.
    :rtype: float or None
    """
    # Get alignments in set of 2-tuples form
    # The "possible" precision is used since it doesn't penalize for finding
    # an alignment that was marked as "possible" (NAACL corpus)
    return precision(reference, self.alignment)
```

Only allows `AlignedSent` object:

```
def precision(self, reference):
    """
    Return the precision of an aligned sentence with respect to a
    "gold standard" reference ``Alignment``.

    :type reference: AlignedSent
    :param reference: A "gold standard" reference ``Alignment``.
    :rtype: float or None
    """
    # Get alignments in set of 2-tuples form
    # The "possible" precision is used since it doesn't penalize for finding
    # an alignment that was marked as "possible" (NAACL corpus)
    return precision(reference.alignment, self.alignment)
```

Static method that only allows `Alignment` object:

```
@staticmethod
def precision(reference_alignment, test_alignment):
    """
    Return the precision of an aligned sentence with respect to a
    "gold standard" reference ``Alignment``.

    :type reference_alignment: Alignment
    :param reference_alignment: A "gold standard" reference ``Alignment``.
    :type test_alignment: Alignment
    :param test_alignment: A hypothesis test ``Alignment``.
    :rtype: float or None
    """
    # Get alignments in set of 2-tuples form
    # The "possible" precision is used since it doesn't penalize for finding
    # an alignment that was marked as "possible" (NAACL corpus)
    return precision(possible, align)
```

Static method that only allows `AlignedSent` object:

```
@staticmethod
def precision(reference_sent, test_sent):
    """
    Return the precision of an aligned sentence with respect to a
    "gold standard" reference ``Alignment``.

    :type reference_sent: AlignedSent
    :param reference_sent: A "gold standard" reference ``AlignedSent``.
    :type test_sent: AlignedSent
    :param test_sent: A hypothesis test ``AlignedSent``.
    :rtype: float or None
    """
    # Get alignments in set of 2-tuples form
    # The "possible" precision is used since it doesn't penalize for finding
    # an alignment that was marked as "possible" (NAACL corpus)
    return precision(reference_sent.alignment, test_sent.alignment)
```

From a user perspective, the "super" polymorphism is the most flexible. Maybe the input type for `AlignedSent.precision` should be resolved in another PR.  

This PR deleted thefunctions from the `nltk.align.util` and moved it into `nltk.align.api.Alignment.fromstring` and it didn't break anything else that uses `Alignment` =)

hoontw:
Personally, I prefer the versions that accept only the `Alignment` object. It will be then the responsibility of the application programmer to ensure that the arguments are the right type before using the `precision` method. I am beginning to realize that function overloading isn't really a big thing in Python as in other languages like C++ or Java. And it saves the tedious type checking inside `precision`.

Between the method and static method versions, I have a slight preference for the static method. In fact, it could be a utility method not associated with any class. After all, precision and recall are general metrics that can work with alignments, information retrieval, probabilistic parsing, and other NLP contexts. All we need is a set of reference elements and a test set of elements.

stevenbird:
Good point @hoontw. How bad would it be for people to use [nltk.metrics.precision](https://github.com/nltk/nltk/blob/develop/nltk/metrics/scores.py#L42) directly? Then the align package could just include an example of how to use it. What do you think @alvations?

Note that we've tended to avoid polymorphism in NLTK, and encouraged users to be clear about the types of the objects they are manipulating. E.g. taggers tag a list of tokens (one list per sentence) with `tag()`, and if you want to tag a text, you can't pass it a list of lists of tokens, you have to call `tag_sents()`. Trees are initialised with lists. If you want to initialise a tree from a string, you use `Tree.fromstring()`.

hoontw:
Personally, I don't think it is too big a deal for people to use precision directly.

```
# before
test_sentence.precision(ref_sentence)

# after
precision(ref_sentence.alignment, test_sentence.alignment)
```

From a software design perspective, precision doesn't seem to belong to the `AlignedSent` class. It feels more like a generic function like sort, average, etc. Whenever I see something like `A.precision(B)`, I can't for the life of me remember which is the test sentence and which is the reference, whereas `precision(reference=A, test=B)` is much less ambiguous.

alvations:
@stevenbird @hoontw I agree `precision` isn't nececessary in`AlignSent`, since `precision()` in `AlignSent` is a ducktype of `nltk.metrics.precision`. I think it's safe if we remove it and then give an example of how to use it in the docstring before the `__init__()`. Would that be okay? 

That would also kind of enforce users to use the idiom: 

```
precision(ref_sentence.alignment, test_sentence.alignment)
```

or

```
precision(Alignment.fromstring("..."), Alignment.fromstring("..."))
```

hoontw:
I am fine with removing the precision and recall methods. The documentation in `test/align.doctest` also needs to be updated as well. `alignment_error_rate` could be restricted to accept just `AlignedSent` and made a static function, but that is probably out of scope for this PR.

stevenbird:
+1 for removing precision and recall and updating `test/align.doctest`

@hoontw – I think you're suggesting that `alignment_error_rate` be a module function (not a method or classmethod), so that we keep the convention of `evaluation_metric(reference, test)`. Sounds good to me.

hoontw:
@AnnaGarbar @alvations Note that `nltk/align` is now `nltk/translate`, so the changes in this PR need to be updated to those files.

stevenbird:
@alvations, @AnnaGarbar, @hoontw – please comment on #1160 as a solution to this issue. (We would still need to drop `AlignedSent.precision`).

alvations:
Sorry for joining in the conversation this late, was a little busy with some work stuff. 

@hoontw , sorry the author omission was a mistake, Anna hasn't do a pull merge to her branch when I pushed to hers. 

@stevenbird , i think it's easier to work with my fork since the MT marathon ended and Anna might not be free to do merging from her branch. So it's easier to close this, wait for the #1160 merge before I work on it using my fork.

hoontw:
@alvations #1160 has the same changes as this branch, except that the functions in `util.py` are not removed.

I have a bit of time before I am busy with work. Do you want me to remove the precision and recall methods from `AlignedSent`, or are you comfortable with doing that yourself?

alvations:
@hoontw i think it's easier to point out in #1160 and let @stevenbird delete the code from `align.util.py` and commit in the same PR. Since the `Alignment.fromstring()` would replace the code in `util.py`.  

Just realized that the @stevenbird 's alignment branch has already deleted the `util.py`. 

And that would also give closure to this PR =)

alvations:
Resolved in #1160

stevenbird:
Thanks for the review @alvations

PR #1133, opened 2015-09-15
Languages list should include English ('eng') by default as it's always there

letuananh:
Languages list should include English ('eng') by default as it's always there (Princeton WordNet).


stevenbird:
Good point, thanks @letuananh

PR #1136, opened 2015-09-19
Python 3.5 Support. Fix Issue #1106.

### LINKED ISSUE #1106: Python 3.5 support
Folks,

I am seeing the following error message when attempting to work through the eBook:

Python 3.5.0rc2 (v3.5.0rc2:cc15d736d860, Aug 25 2015, 04:45:41) [MSC v.1900 32 b
it (Intel)] on win32

> > > import nltk
> > > nltk.corpus.brown.words()
> > >   File "<stdin>", line 1, in <module>
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\util.py", line 664, in **repr**
> > >     for elt in self:
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\corpus\reader\util.py", line 394, in iterate_from
> > >     for tok in piece.iterate_from(max(0, start_tok-offset)):
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\corpus\reader\util.py", line 291, in iterate_from
> > >     tokens = self.read_block(self._stream)
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\corpus\reader\tagged.py", line 243, in read_block
> > >     for sent_str in self._sent_tokenizer.tokenize(para_str):
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\tokenize\regexp.py", line 126, in tokenize
> > >     self._check_regexp()
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\tokenize\regexp.py", line 121, in _check_regexp
> > >     self._regexp = compile_regexp_to_noncapturing(self._pattern, self._flags)
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\internals.py", line 55, in compile_regexp_to_nonca
> > > pturing
> > >     return sre_compile.compile(convert_regexp_to_noncapturing_parsed(sre_parse.p
> > > arse(pattern)), flags=flags)
> > >   File "C:\Users\shenry\AppData\Local\Programs\Python\Python35-32\lib\site-packa
> > > ges\nltk-3.0.4-py3.5.egg\nltk\internals.py", line 51, in convert_regexp_to_nonca
> > > pturing_parsed
> > >     parsed_pattern.pattern.groups = 1
> > > AttributeError: can't set attribute

beeblebroccoli:
I was able to fix [this problem](https://github.com/nltk/nltk/issues/1106) by going into the `internals.py` file in the `nltk` directory and ignoring the exception caused by the `parsed_pattern.pattern.groups = 1` line.

My rationale behind this, after doing a bit of code reading, was that the [original version](http://svn.python.org/projects/python/branches/release22-maint/Lib/sre_parse.py) of `sre_parse.py` that NLTK appears to have been designed to work with stored `groups` as an attribute of an instance of the `sre_parse.Pattern` class. The version that comes with Python 3.5 stores `groups` as a property which returns (I'm not too familiar with properties, but this is what I presume it does) the length of a `subpattern` list. [The code I'm talking about is here at about line 75.](https://github.com/python/cpython/blob/master/Lib/sre_parse.py) Since setting the `groups` attribute to 1 was most likely a sort of resetting mechanism or similar (I haven't read that far yet), having the code just work as it does in the newer version with the built in length checker seems to be fine.

This does solve the problem that the OP and I had.

```
In [1]: import nltk

In [2]: nltk.corpus.brown.words()
Out [2]: ['The', 'Fulton', 'County', 'Grand', 'Jury', 'said', ...]

In [3]: 
```

What I don't know is what the long term effects of doing this will be, I came up with this solution just by tracing through the code, I haven't looked at what bugs this may cause in the long run.

Would this cause other problems? If so, is there a better solution?

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -49,6 +49,10 @@ def compile_regexp_to_noncapturing(pattern, flags=0):
                 raise ValueError('Regular expressions with back-references are not supported: {0}'.format(pattern))
             res_data.append((key, value))
         parsed_pattern.data = res_data
+        try:
+            parsed_pattern.pattern.groups = 1
+        except AttributeError:
+            pass
         parsed_pattern.pattern.groupdict = {}
         return parsed_pattern
 
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -49,7 +49,6 @@ def compile_regexp_to_noncapturing(pattern, flags=0):
                 raise ValueError('Regular expressions with back-references are not supported: {0}'.format(pattern))
             res_data.append((key, value))
         parsed_pattern.data = res_data
-        parsed_pattern.pattern.groups = 1
         parsed_pattern.pattern.groupdict = {}
         return parsed_pattern
 
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -182,3 +182,4 @@
 - wvanlint
 - Álvaro Justen <https://github.com/turicas>
 - bjut-hz
+- Zenohm


jennifer-yan:
Yes, it works anyway, thanks

nltk/internals.py
```diff
@@ -49,7 +49,10 @@ def convert_regexp_to_noncapturing_parsed(parsed_pattern):
                 raise ValueError('Regular expressions with back-references are not supported: {0}'.format(pattern))
             res_data.append((key, value))
         parsed_pattern.data = res_data
-        parsed_pattern.pattern.groups = 1
+        try:
+            parsed_pattern.pattern.groups = 1
+        except AttributeError:
+            pass
```
hoontw, line 55: By swallowing the error, the parsed_pattern could be left in an inconsistent state. If there are capturing groups, `parsed_pattern.pattern.groups` will be more than 1, yet `parsed_pattern.pattern.groupdict` will be empty.

Instead of trying to fix the symptom (pattern.groups cannot be assigned), why not try finding a better solution to compile_regexp_to_noncapturing? IMHO, the current implementation knows too much about the internals of sre_parse.py. Whenever the implementation of sre_parse changes, as with Python 3.5, it could break this function.

Another approach is to question the need for this function. This seems to be used only in `nltk.tokenize.RegexpTokenizer`. Instead of removing capture groups silently, why not just throw an exception if there are capture groups? Note that this is an API change and will require discussion with the NLTK community.

stevenbird, line 55: I agree about the current implementation knowing too much about the internals of sre_parse, and that we should reconsider the need for this. In addition to throwing an exception, we could provide a flag so that anyone who depends on the old behaviour can easily get it back, at least for a while.

stevenbird, line 55: Here's my proposed solution: https://github.com/nltk/nltk/commit/7e43ac05c41d5ffcce9197952e1bf46831d08d90

stevenbird:
NB there are comments here: https://github.com/nltk/nltk/commit/7e43ac05c41d5ffcce9197952e1bf46831d08d90

Issue #1070, opened 2015-08-07
Skipgram Implementation


alvations:
Could we add [this skipgram](http://stackoverflow.com/questions/31847682/how-to-compute-skipgrams-in-python) implementation to `nltk.util`?

alvations:
Ah there's a bug there where it's generating extra ngrams.

stevenbird:
An implementation of skipgrams would be a welcome addition.

alvations:
Sorry for the open and closing, I wasn't sure whether i've got it right but here's a working implementation (But it requires the knowledge of `len(sequence)` so it can't take an iterable. This is because of the need of skipgram to know whether `n+k > len(sequence)`).

Making `pad_sequence` into a separate function would relief the repeated code from ngram and skipgram but also it's needed so that different pad sequence can be in `NgramLanguageModel` later on where it's necessary to pad differently `<s>` on the left and `</s>` on the right.

```
from itertools import chain, combinations
import copy
from nltk.util import ngrams

def pad_sequence(sequence, n, pad_left=False, pad_right=False, pad_symbol=None):
    """
    Pads a sequence on the left or/and right with a specified pad_symbol before
    generating ngrams for a given sequence.
    >>> n = 3 
    >>> list(pad_sequence(iter(text), n)) # no pad
    ['Insurgents', 'killed', 'in', 'ongoing', 'fighting']
    >>> list(pad_sequence(iter(text), n, pad_left=True, pad_symbol='<s>')) # pad left
    ['<s>', '<s>', 'Insurgents', 'killed', 'in', 'ongoing', 'fighting']
    >>> list(pad_sequence(iter(text), n, pad_right=True, pad_symbol='</s>')) # pad right
    ['Insurgents', 'killed', 'in', 'ongoing', 'fighting', '</s>', '</s>']
    >>> list(pad_sequence(padded_left, n, pad_right=True, pad_symbol='</s>')) # pad left right
    ['<s>', '<s>', 'Insurgents', 'killed', 'in', 'ongoing', 'fighting', '</s>', '</s>']
    :param sequence: the source data to be converted into ngrams
    :type sequence: iter
    :param n: the degree of the ngrams
    :type n: int
    :param pad_left: whether the ngrams should be left-padded
    :type pad_left: bool
    :param pad_right: whether the ngrams should be right-padded
    :type pad_right: bool
    :param pad_symbol: the symbol to use for padding (default is None)
    :type pad_symbol: any
[...]

PR #1138, opened 2015-09-20
Added an Everygrams and Skipgrams implementation

alvations:
Added @jnothman implementation of skipgrams from #1070 and also an "everygrams" implementation to extract every ngram given a sequence.


alvations:
Thanks @stevenbird for the fixes and merge!

Issue #296, opened 2012-09-17
Corpora should have the same interface


turicas:
I'm writing a function that receives a corpus object but it doesn't know which corpus is it. In this function I need to get all the words in the corpus; when I try with `machado` or `stopwords` corpora it works, but not with `shakespeare`. The problem is that these objects does not share the same interface - and I think it's a bug.
I think all corpora methods should have the same interface regardless of how the corpus is retrieved, so we can reuse code without worrying about compatibility.
The problem can be illustrated with the code below (`nltk.__version__` = `'2.0.2'`):

``` python
>>> from nltk.corpus import shakespeare, machado, stopwords
>>> temp = machado.words() #works
>>> temp2 = stopwords.words() #works
>>> temp3 = shakespeare.words() #doesn't work!
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/alvaro/.virtualenvs/bug-nltk/local/lib/python2.7/site-packages/nltk/corpus/reader/xmldocs.py", line 63, in words
    elt = self.xml(fileid)
  File "/home/alvaro/.virtualenvs/bug-nltk/local/lib/python2.7/site-packages/nltk/corpus/reader/xmldocs.py", line 44, in xml
    raise TypeError('Expected a single file identifier string')
TypeError: Expected a single file identifier string
```

When reading the help of each method, I found the inconsistencies on parameters:
- `machado.words(self, fileids=None, categories=None)` (from `nltk.corpus.reader.plaintext.PortugueseCategorizedPlaintextCorpusReader`)
- `stopwords.words(self, fileids=None)` (from `nltk.corpus.reader.wordlist.WordListCorpusReader`)
- `shakespeare.words(self, fileid=None)` (from `nltk.corpus.reader.xmldocs.XMLCorpusReader`) -- even with `fileid=None`, it'll raise an exception if I don't pass it.

turicas:
I did some tests about corpora encoding in [this gist](https://gist.github.com/3801534) (and reported results on #299). These corpora below have some interface problem (probably with methods `words` or `raw`):
- `cmudict`
- `conll2007`
- `ieer`
- `ipipan`
- `nombank`
- `ppattach`
- `propbank`
- `qc`
- `senseval`
- `switchboard`
- `timit`
- `toolbox`
- `udhr`
- `wordnet`
- `wordnet_ic`
- `ycoe`

fcbond:
[...]

Issue #31, opened 2012-01-17
Update corpus metadata files


alexrudnick:
Check that the license and other information in the corpus index files is correct.

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=96</p>

fcbond:
It would be nice to have a standard interface to access the license and citation interface, as we do for the readme.

I suggest adding to the API  something like:

``` python
   def license(self):
        """
        Return the contents of the corpus LICENSE file, if it exists.
        """
        if os.path.exist(self._root.join("LICENSE")):
            return self.open("LICENSE").read()
        else:
            return "No LICENSE found for this corpus (maybe check the README)"
```

And the same for '''citation' which would look for "citation.bib".

Then we need to add the LICENSE and citation info for each corpus.  I will do it for the open multilingual wordnet and wordnet.

stevenbird:
Good idea. I've added this, but without the conditional, since I think it's fine let Python generate an error message. I see that there are many corpora containing `README.txt` instead of `README`, so that also needs to be standardized.

PR #1139, opened 2015-09-21
Updates to improve the handling of wordnets in multiple languages.

fcbond:
@stevenbird  here are some extensions: the first one makes it easier to manipulate lemmas even if some don't exist for some languages in some languages.  The second is for code readability.  The third is to make it easier for people to cite and find out about the individual resources (and to call the module with 'words()'), this helps with issues #31 and #296).  Finally, the last change makes the contributors more visible.


stevenbird:
Thanks @fcbond

PR #1140, opened 2015-09-22
Allow multiword terminal nodes in BracketParseCorpusReader

dwymark:
Added 'multiword_terminals' parameter to BracketParseCorpusReader.__init__. If multiword_terminals=True, terminal nodes separated by spaces will not be truncated. Default is set to False, so backwards compatibility will be maintained.

I made this change because I have access to the full Penn Treebank v2 through my university, which I found to be incompatible with the original BracketParseCorpusReader. The format of the parsed sentences that I have access to are identical to those found in the 'treebank/parsed' folder included with NLTK.

BracketParseCorpusReader._normalize was truncating any terminal nodes with spaces separating them, i.e. '(NP Pierre Vinken)' => '(NP Pierre)'. With the 'multiword_terminals' parameter set to True, the problem is solved.

### DIFF
--- a/nltk/corpus/reader/bracket_parse.py
+++ b/nltk/corpus/reader/bracket_parse.py
@@ -29,8 +29,8 @@ class BracketParseCorpusReader(SyntaxCorpusReader):
     trees.
     """
     def __init__(self, root, fileids, comment_char=None,
-                 detect_blocks='unindented_paren', encoding='utf8',
-                 tagset=None):
+                 detect_blocks='unindented_paren', multiword_terminals=False,
+                 encoding='utf8', tagset=None):
         """
         :param root: The root directory for this corpus.
         :param fileids: A list or regexp specifying the fileids in this corpus.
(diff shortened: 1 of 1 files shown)


stevenbird:
It looks like the substitution below is based on treebanks that include a word root as part of the annotation of leaves. I would prefer to favour the Penn Treebank setting as the default, and if a treebank contains this morphological information, that would require the flag.

I wonder if anyone knows which treebank corpus motivated this forgetful substitution:

```
t = re.sub(r"\(([^\s()]+) ([^\s()]+) [^\s()]+\)", r"(\1 \2)", t)
```

Do you know, @edloper, @heatherleaf, @desilinguist?

stevenbird:
@dwymark – you are using the "parsed" section of Penn Treebank which omits POS tags. So it's not correct to say that `(NP Pierre Vinken)` is a multiword terminal node; it is a preterminal node with two children. In the "combined" section of Treebank it shows up as `(NP (NNP Pierre) (NNP Vinken) )`. Would you please use that section instead, and report back how you got on?

dwymark:
Is there a standard way for dealing with the "parsed" section of Penn Treebank? For some reason that is all that I have full access to through my institution.

PR #1141, opened 2015-09-23
Hotfix/fix encoding cess (with test)

zeehio:
This fix amends the encoding of teh cess_cat and cess_esp corpora.

It is split in two commits:

The first commit enhances the TestCess to assert for some words different in ISO-8859-2 and ISO-8859-15 in both corpus. As a result of the test enhancement, after this first commit the: `nosetests nltk.test.unit.test_corpora.TestCess` command fails.

The second commit changes the encoding of both corpora and then the enhanced test passes.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -182,3 +182,4 @@
 - wvanlint
 - Álvaro Justen <https://github.com/turicas>
 - bjut-hz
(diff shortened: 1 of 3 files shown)


stevenbird:
Thanks @zeehio

PR #1142, opened 2015-09-25
Removed invalid parameter from docstring example

stieglma:
As proposed in #1112 i fixed the example in the docstring, there was still an out-of-date parameter that we had earlier but then got removed sometime without updating the docstring.


stevenbird:
Thanks @stieglma

Issue #1122, opened 2015-09-08
Perceptron tagger


stevenbird:
Port the [TextBlob PerceptronTagger](https://github.com/sloria/textblob-aptagger/blob/master/textblob_aptagger/taggers.py) to NLTK. Assume NLTK-style input, i.e. an iterator of sentences, where each sentence is a list of tokens. Train on Penn Treebank and create a trained model for addition to the NLTK data collection. Request review from @honnibal before merging. cf #1110

PR #1143, opened 2015-09-27
Add average perceptron tagger (fix bug #1122)

longdt219:
Hi @stevenbird  and @honnibal

This is the attempt to port Honnibal POS Average perceptron POS tagger to NLTK which borrows heavily from <a href='https://github.com/sloria/textblob-aptagger/blob/master/textblob_aptagger/taggers.py'> TextBlob Tagger </a>. 
When train on WSJ section 2-11 and test on section 23, the accuracy is 97.2 % 
The pretrained model is also included (the zip file). Could you put this file to the folder <b> /nltk_data/models/ </b>  @stevenbird 
Probably we can use this one as the default POS tagger for NLTK. 

from nltk.tag.perceptron import PerceptronTagger
tagger = PerceptronTagger()
tagger.tag("The quick brown fox jumps over the lazy dog".split())
output: [('The', 'DT'), ('quick', 'JJ'), ('brown', 'NN'), ('fox', 'NN'), ('jumps', 'VBZ'), ('over', 'IN'), ('the', 'DT'), ('lazy', 'JJ'), ('dog', 'NN')]
tagger.tag("The red cat".split())
output: [('The', 'DT'), ('red', 'JJ'), ('cat', 'NN')]


honnibal:
I had a look at this to make sure it's correct — looks good.

A couple of points:
- You'll probably want to break out the averaged perceptron code into its own file. You might want to reuse the implementation, particularly if you're taking on the dependency parser too. Actually averaged perceptron is a nice demonstration algorithm anyway, I think. It was what really helped me understand the difference between Naive Bayes and Max Ent. It's also useful to explain neural network concepts.
- You could consider implementing the learner with numpy, instead of the current dict of strings. This improves efficiency, particularly memory efficiency. The downside is it's harder to inspect the model. I'd probably keep it like this.
- There are a couple of constants you could consider exposing, e.g. the tag dictionary look up constants. You'll probably also want to give users control over the tag dictionary, as it's heavily data dependent. For instance, "Now" probably lands in the tag dictionary as an RB, but the user's data might have Google Now, and want it tagged as an entity.
- The best URL to cite for the blog post is here: http://spacy.io/blog/part-of-speech-POS-tagger-in-python/ . My Wordpress blog is no longer maintained, and the link might rot.
- I think the copyright notice in the header is incorrect --- see here: https://github.com/sloria/textblob-aptagger/blob/dev/LICENSE

kmike:
What it `nltk/tag/perceptron_tagger.zip`?

stevenbird:
@honnibal – thanks for the feedback. Would it be ok to include the following in the top of the module?

```
This module is a port of the Textblob Averaged Perceptron Tagger
https://github.com/sloria/textblob-aptagger

Copyright 2013 Matthew Honnibal
NLTK modifications Copyright 2015 The NLTK Project

This module is provided under the terms of the MIT License.
```

If you prefer, we can include the full text of the license.

honnibal:
That notice looks fine, thanks.

stevenbird:
@kmike – the zipfile is the pre-trained model.
@longdt219 – it is now in nltk_data, cf https://github.com/nltk/nltk_data/commit/c9887b508db9728d5acf6787de47af958e459c39

stevenbird:
Thanks @honnibal

longdt219:
Hi @stevenbird, 
I modified the code to reflect above comments.

kmike:
@longdt219 is the zip file still needed if it is in nltk_data? If it is not needed, it is better to remove this file and squash commits.

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+                total += (self.i - self._tstamps[param]) * weight
+                averaged = round(total / float(self.i), 3)
+                if averaged:
+                    new_feat_weights[clas] = averaged
+            self.weights[feat] = new_feat_weights
+        return None
+
+    def save(self, path):
+        '''Save the pickled model weights.'''
+        return pickle.dump(dict(self.weights), open(path, 'w'))
```
kmike: We should use context managers to close file properly.

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def load(self, path):
+        '''Load the pickled model weights.'''
+        self.weights = pickle.load(open(path))
```
kmike: Please use context manager to close the file properly. Also, it should be opened in binary mode.

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+                    prev = guess
+                    c += guess == tags[i]
+                    n += 1
+            random.shuffle(sentences)
+            logging.info("Iter {0}: {1}/{2}={3}".format(iter_, c, n, _pc(c, n)))
+        self.model.average_weights()
+        # Pickle as a binary file
+        if save_loc is not None:
+            pickle.dump((self.model.weights, self.tagdict, self.classes),
+                         open(save_loc, 'wb'), -1)
```
kmike: the file is never closed

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+        return None
+
+    def load(self, loc):
+        '''
+        :param loc: Load a pickled model at location.
+        :type loc: str 
+        '''
+        try:
+            w_td_c = pickle.load(open(loc, 'rb'))
```
kmike: file should be closed

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+        except IOError:
+            msg = ("Missing trontagger.pickle file.")
+            raise Exception(msg)
```
kmike: Why is this needed? IOError may be caused not only by a missing file, and Exception is less specific than IOError, so users can't catch this exception easily. I'm also not sure about 'trontagger.pickle' - it should be loc, not 'trontagger.pickle'.

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+        self.model.weights, self.tagdict, self.classes = w_td_c
+        self.model.classes = self.classes
+        return None
```
kmike: return None is pointless :)

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def _normalize(self, word):
```
kmike: what about making it a classmethod?

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+            if n >= freq_thresh and (float(mode) / n) >= ambiguity_thresh:
+                self.tagdict[word] = tag
+
+
+def _pc(n, d):
+    return (float(n) / d) * 100
+
+def _load_data_conll_format(filename):
+    print ('Read from file: ', filename)
+    fin = open(filename,'rb')
```
kmike: it is better to use context managers to close the file because it handles exceptions properly

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+    sentences = []
+    sentence = []
+    for line in fin.readlines():
+        line = line.strip()
+        #print line
+        if len(line) ==0:
+            sentences.append(sentence)
+            sentence = []
+            continue
+        tokens = line.split('\t')
```
kmike: does it work in Python 3?

nltk/tag/perceptron.py
```diff
@@ -0,0 +1,317 @@
...
+        Make a tag dictionary for single-tag words.
+        :param sentences: A list of list of (word, tag) tuples.
+        '''
+        counts = defaultdict(lambda: defaultdict(int))
+        for sentence in sentences:
+            for word, tag in sentence:
+                counts[word][tag] += 1
+                self.classes.add(tag)
+        freq_thresh = 20
+        ambiguity_thresh = 0.97
```
kmike, line 269: what about making these constants .train arguments?

nltk/tag/perceptron.py
kmike: To clarify: I'd prefer not to handle this IOError at all.

honnibal: +1 to all that.

stevenbird:
Thanks @kmike

kmike:
@longdt219 It is not enough to remove the file in a commit - if this PR will get merged as-is we'll have 2.5MB file sitting in nltk history forever. Could you please squash the commits?

kmike:
@stevenbird no problems :)

kmike:
Regarding the pretrained file: we should make sure it is loaded fine in Python 2 and 3, or that there are separate versions for 2 and 3. It could also make sense to use JSON instead of pickle, just to be sure we can always load it.

longdt219:
Thank you for your comments @kmike 
The pickle works fine for both python 2 and python 3

kmike:
hey @longdt219! the commit which adds a binary file is still in the git history - could you please remove it?

stevenbird:
Well, I'm confused about what happened to the git history, but it's not that important, the file is there, and there's no data file.

stevenbird:
Thanks @longdt219, @honnibal, @kmike

PR #1145, opened 2015-09-28
Implement stack decoder for machine translation

hoontw:
Basic implementation of a stack decoder.

This only implements the "engine" of translation. Some major components still need to be done before machine translation is viable: language model (that's a big one), extracting phrase tables from corpora, reading phrase tables generated by Moses, not loading the entire phrase table in memory.

@alvations If you have the time, I would appreciate it if you looked at the interface and see how it compares with your monotone decoder.


stevenbird:
@hoontw – with this PR you're proposing a new package `translate`. Do you think the `align` package contents would be more appropriate in this package?

hoontw:
I was thinking of putting this in the `align` package myself, but decided not to. A translation decoder is not really aligning anything; it has all the alignments it needs. Its main goal is to generate a sentence based on a sentence in another language.

I don't like creating new root-level packages. How about renaming the `align` package to `translate`? The word and phrase alignment modules can be left there, or put one subdirectory down. So, something like `nltk.translate.ibm1` or `nltk.translate.align.ibm1`.

stevenbird:
@hoontw – yes, that's what I was proposing, that the contents of `align` should be moved into `translate`. I prefer to avoid subpackages, so I think everything can be immediately under `translate`, not `translate/align`. Are you happy to go ahead and do this? Thanks.

hoontw:
@stevenbird Sure, I don't mind combining the `align` package with `translate`, so that only the `translate` package is left.

nltk/translate/stack_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+    >>> language_prob[('!',)] = log(0.1)
+    >>> language_model = type('',(object,),{'probability_change': lambda self, context, phrase: language_prob[phrase], 'probability': lambda self, phrase: language_prob[phrase]})()
+
+    >>> stack_decoder = StackDecoder(phrase_table, language_model)
+
+    >>> stack_decoder.translate(['niemand', 'erwartet', 'die', 'spanische', 'inquisition', '!'])
+    ['nobody', 'expects', 'the', 'spanish', 'inquisition', '!']
+
+    """
+    def __init__(self, phrase_table, language_model):
```
alvations, line 79: `stack_size` parameter should be initialize-able here but the default 5 or 100 can set be set. 

`_distortion_factor` too should be initialize-able.

These are important parameters in MT decoding.

nltk/translate/stack_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+
+        self.stack_size = 100
+        """
+        int: Maximum number of hypotheses to consider in a stack.
+            Higher values increase the likelihood of a good translation,
+            but increases processing time.
+        """
+
+        self.__distortion_factor = 0.5
+        self.__compute_log_distortion()
```
alvations, line 119: It's better if `self.__log_distortion_factor = self.__compute_log_distortion()`.

nltk/test/unit/translate/test_stack_decoder.py
```diff
@@ -0,0 +1,300 @@
...
+            [],
+            [7]
+        ]
+
+        # act
+        phrase_spans = StackDecoder.valid_phrases(all_phrases_from, hypothesis)
+
+        # assert
+        self.assertEqual(phrase_spans, [(0, 1), (1, 2), (3, 5), (4, 5), (4, 6)])
+
```
alvations, line 140: Possibly the `create_fake...()` will disappear when `phrase_extraction()` can generate a real `PhraseTable` object.

nltk/test/unit/translate/test_stack_decoder.py
```diff
@@ -0,0 +1,300 @@
...
+        score = stack_decoder.distortion_score(hypothesis, (8, 10))
+
+        # assert
+        # expansion from empty hypothesis always has zero distortion cost
+        self.assertEqual(score, 0.0)
+
+    def test_compute_future_costs(self):
+        # arrange
+        phrase_table = TestStackDecoder.create_fake_phrase_table()
+        language_model = TestStackDecoder.create_fake_language_model()
```
alvations, line 69: Possibly this would also be changed once `nltk.model` is stable and is able to produce `LanguageModel` object

nltk/translate/stack_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+                            src_phrase_span=src_phrase_span,
+                            trg_phrase=translation_option.trg_phrase,
+                            previous=hypothesis
+                        )
+                        new_hypothesis.future_score = self.future_score(
+                            new_hypothesis, future_score_table, sentence_length)
+                        total_words = new_hypothesis.total_translated_words()
+                        stacks[total_words].push(new_hypothesis)
+
+        if not stacks[sentence_length]:
```
alvations, line 182: When encounter Out-of-Vocabulary word, it should return the source language word and skip to the next. But this will also cause problems since you don't know where to put these OOV words =)

The complete solution would be to put it everywhere in the current hypothesis and then calculate sum perplexity of the phrases before the OOV and after the OOV and then minimize the perplexity. 

In actual fact, people just put it in where they see it and skip to the next word and treat the current position where the OOV exists is where the target language word would exists without distortion =)

alvations:
@stevenbird , @hoontw , I think it should remain separate since there are many alignment methods in NLP that don't necessary leads to or aims to translate. But that's just my personal preference, the line is fine between them.

Personally, I would like to keep things modular:

```
align/
    IBM models
    alignment wrappers
    Alignment API

model/
    NgramModel API
    smoothing
    LanguageModel API (this might not be an Ngram based model)   

translate/
    phrase extraction
    reordering extraction
    decoders
    Phrase table API
    Rule table API
```

@stevenbird , I think we should poll this to the devs and users. 

@hoontw , I read through the decoder code. It's surely much more documented and more object orientated than mine =)

Have you tried decoding the sample model from https://github.com/alvations/nltk/blob/translate/nltk/translate/util.py#L128 ? That should be the minimal gauge of whether the code is use-able. 

For my code `heapq` does all the ranking magic and all i need to do to prune is to use the optimized `heapq.nlargest` https://github.com/alvations/nltk/blob/translate/nltk/translate/stack_decoder.py#L38 I'm not sure how much speed I've gained but i like the idea of using native python objects but I wasn't considering other features like reordering and such so it's a lot simpler than your code.

nltk/test/unit/translate/test_stack_decoder.py
hoontw, line 69: Yes, that's right. Once this PR is accepted, I will work on a preliminary abstract class for language models. It can replace all the ugly language model stubs in this module.

nltk/test/unit/translate/test_stack_decoder.py
hoontw, line 140: For doctests, yes, a `PhraseTable` will be constructed from real data. The goal is to demonstrate how the different components of the align and translate modules can be used on a real world translation problem.

For unit tests, the goal is to test a component in isolation. So, a fake `PhraseTable` will still be manually constructed here.

nltk/translate/stack_decoder.py
hoontw, line 79: `stack_size` and `distortion_factor` are public attributes that can be set after constructing the `StackDecoder`. I did not want to have a large number of parameters in the `__init__`` and only included the two that were mandatory for any translation to happen.

Typical usage will look like this

```
phrase_table = read_phrase_table()
language_model = create_language_model()
decoder = StackDecoder(phrase_table, language_model)
# set optional parameters
decoder.stack_size = 5
decoder.distortion_factor = 1
decoder.word_penalty = 3
decoder.translate(['hello', 'world', '!'])
```

nltk/translate/stack_decoder.py
hoontw, line 119: The purpose of  `self.__log_distortion_factor` is to store the pre-computed value of `log(self.distortion_factor)`. This is an optimization, so that the log value doesn't have to be recomputed O(n^2) times a new hypothesis is considered.

`self.__log_distortion_factor` will be updated when `self.distortion_factor` is changed through the setter.

nltk/translate/stack_decoder.py
hoontw, line 182: I agree with you that the handling of OOV words could be much better. As this is meant to be an initial implementation, I will leave this as "future work" :smile: Incremental pull requests are easier to review than giant pull requests of a new feature.

hoontw:
@alvations Thanks for the review. I actually took a look at [your branch](https://github.com/alvations/nltk/tree/translate/nltk/translate) before writing this module. I learned some neat things like the use of `namedtuple`. I think you succeeded in making it a well-rounded solution, covering language model and file I/O work. My module still has a long way to go before it can be useful -- it only works on hand-coded models. When a language model and a phrase table reader is implemented, I can test it on real world data as you suggested.

For ranking, I kept it simple with the built-in `sort()` for Python lists. If performance is an issue, your suggestion for using heaps will definitely be well-taken.

As for package structure, I believe that `stack_decoder` should not be in `align`. The question then is whether the contents of `align` should be in `translate`. I understand that word alignments are not necessarily used for translation, but I have a slight preference for grouping them all under `translate`. After all, the modules in `align` are usually introduced in the context of machine translation in NLP textbooks (Jurafsky, Manning, etc.). I don't mind putting this to a poll if there are strong opinions about this.

stevenbird:
https://groups.google.com/forum/#!topic/nltk-dev/oeRirL87AbA (@hoontw's post on nltk-dev)

stevenbird:
I recommend that we go ahead and incorporate the existing contents of the align package into the new translate package.

PR #1153, opened 2015-09-29
Fix MTECorpusReader test error

fievelk:
Test is now skipped: doctest is only used for documentational purposes. This solution has been chosen following examples provided in `reader/tagged.py` and `reader/aligned.py`.


stevenbird:
Thanks @fievelk

PR #1156, opened 2015-10-06
fix lemmatize in rte_classify.py

pafnucy:
in _nltk.corpus.wordnet.morphy_, _pos='verb'_ argument produces a _KeyError_

constant for verb is _nltk.corpus.wordnet.VERB_, which evaluates to _'v'_


stevenbird:
Thanks @pafnucy

PR #1159, opened 2015-10-07
Create API for language model

hoontw:
First stab at creating a generic interface for language models. Used to replace hacky code in translate module.

A concrete implementation is still lacking. See #602 and https://github.com/nltk/nltk/commit/73f7e7b9ec301c39dc11035f138a27feb3c436cf for discussion regarding previous work on ngram models, which have been dropped because of major issues.

### DIFF
--- a/nltk/model/__init__.py
+++ b/nltk/model/__init__.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Natural Language Toolkit: Language Models
+#
+# Copyright (C) 2001-2015 NLTK Project
+# Author: Tah Wei Hoon <hoon.tw@gmail.com>
+# URL: <http://nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""
+Experimental features for language models
+These interfaces are prone to change.
+"""
+
+from nltk.model.api import LanguageModel
--- a/nltk/model/api.py
+++ b/nltk/model/api.py
@@ -1,0 +1,58 @@
+# Natural Language Toolkit: Language Model
+#
+# Copyright (C) 2001-2015 NLTK Project
+# Author: Tah Wei Hoon <hoon.tw@gmail.com>
+# URL: <http://nltk.org/>
+# For license information, see LICENSE.TXT
+
+
+class LanguageModel(object):
+    """
+    Abstract class for language model
+
+    A language model gives the probability of occurrence for a
+    sequence of words
+    """
+    def probability(self, words):
+        """
+        Log probability of occurrence for a sequence of words
+
+        :param words: Sequence of words. For example,
+            ``['ham', 'and', 'spam']``.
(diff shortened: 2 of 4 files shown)


stevenbird:
@jordanbg, @edloper, @alvations - please comment

stevenbird:
I've merged this into a new branch, [model2](https://github.com/nltk/nltk/tree/model2). I think that we need to see more functionality (cf the [model](https://github.com/nltk/nltk/tree/model) branch), before merging it with the main development branch.

stevenbird:
@hoontw: it looks like [SRILM](http://www.speech.sri.com/projects/srilm/manpages/Prob.3.html) distinguishes `Prob` from `LogP`. It looks like [IRSTLM](http://hermes.fbk.eu/people/bertoldi/teaching/lab_2010-2011/img/irstlm-manual.pdf) calls it `lprob`. A quick poke around the [KenLM](https://kheafield.com/code/kenlm/) docs doesn't reveal a naming choice.

So is there a case, based on practice with other toolkits, for a method called `probability` or `prob` that returns a log-probability?

alvations:
I think one reason why log prob is used more often is because of [underflow](https://en.wikipedia.org/wiki/Arithmetic_underflow). Also when doing `prod(probabilities)`, one could do `sum(logprob)` instead. 

If i'm not wrong KenLM uses log probabilities https://github.com/kpu/kenlm/blob/master/python/kenlm.cpp#L457  but I think it was using log_e instead of log_10. 

But I remember ARPA file dumps out the probabilities in log_10 (https://github.com/alvations/nltk/blob/translate/nltk/translate/util.py#L68) and every time KenLM dumps to ARPA, it converts to log_10.  @kpu would be the best person to tell us =)

kpu:
As far as users can tell, everything in KenLM is log base 10.

hoontw:
Based on the discussions above, I am inclined to rename the `probability` method to `logprob`. This makes it consistent with other parts of nltk, and also minimizes confusion. I am not sure how useful a regular `probability` method will be, since LMs usually deal with log probabilities. As @alvations mentioned, ARPA files use log10 probabilities.

SRILM does distinguish between `Prob` and `LogP` internally, but user commands like [ngram](http://www.speech.sri.com/projects/srilm/manpages/ngram.1.html) seem to deal with log probabilities. Maybe someone who is more familiar with the different language models out there can comment on the best practices?

Implementing a language model in NLTK is a huge undertaking that I unfortunately don't have time for this year. That would be a nice TODO for [2016](http://www.nltk.org/contribute.html). I am OK with closing this PR or leaving it dormant until a concrete implementation is provided.

hoontw:
Closing this until I have time to create a more substantial language model module.

stevenbird:
Ok, thanks @hoontw

PR #1160, opened 2015-10-07
add Alignment.fromstring; remove polymorphism in AlignedSent

stevenbird:
This is a possible resolution for #1129


nltk/translate/api.py
```diff
@@ -20,16 +20,17 @@ class AlignedSent(object):
...
-        ...     ['the', 'house', 'is', 'small'], '0-2 1-3 2-1 3-0')
+        ...     ['the', 'house', 'is', 'small'], Alignment.fromstring('0-2 1-3 2-1 3-0'))
         >>> algnsent.words
         ['klein', 'ist', 'das', 'Haus']
         >>> algnsent.mots
         ['the', 'house', 'is', 'small']
         >>> algnsent.alignment
         Alignment([(0, 2), (1, 3), (2, 1), (3, 0)])
-        >>> algnsent.precision('0-2 1-3 2-1 3-3')
+        >>> precision(Alignment.fromstring('0-2 1-3 2-1 3-3'), algnsent.alignment)
```
hoontw: When `precision` is removed from `AlignedSent`, this doctest should be removed. Example integrations with `nltk.metrics` and other modules could be shown in `test\translate.doctest`.

nltk/translate/api.py
```diff
@@ -275,19 +274,33 @@ class Alignment(frozenset):
...
         >>> b.issubset(a)
         True
-        >>> c = Alignment('0-0 0-1')
+        >>> c = Alignment.fromstring('0-0 0-1')
         >>> b == c
         True
     """
 
-    def __new__(cls, string_or_pairs):
-        if isinstance(string_or_pairs, string_types):
```
hoontw, line 284: Since `string_types` is no longer used, the import name in https://github.com/stevenbird/nltk/blob/ae2d80f24b6cdf739bdca7ef32584e9f5839abd2/nltk/translate/api.py#L12 can be removed.

hoontw:
Overall, looks good to me.

alvations:
@stevenbird Thanks for the new PR. It looks good. 

I'm not sure whether it's github that's not tracking the util.py removal. But if it hasn't been removed, would it be possible to also do a git remove for `util.py` in this PR?

nltk/translate/api.py
```diff
@@ -46,7 +47,7 @@ class AlignedSent(object):
     :type alignment: Alignment
     """
 
-    def __init__(self, words=[], mots=[], alignment='', encoding='utf8'):
+    def __init__(self, words=[], mots=[], alignment=[], encoding='utf8'):
```
kmike: It'd be good to avoid mutable defaults.

alvations: Also, `encoding` isn't used in the rest of the code, so i think it's safe to delete it.

stevenbird: @kmike – that's true – I don't see any use for these defaults so I'll remove them.
@alvations – I'll drop `encoding` too, thanks.

stevenbird:
@alvations – I think I inadvertently removed util.py in a different commit. Sorry about that.

stevenbird:
`AlignedSent.alignment_error_rate()` is another polymorphic method; it works with both `AlignedSent` and `Alignment`. I think it should only be the latter.

nltk/translate/api.py
```diff
@@ -163,51 +157,6 @@ def invert(self):
...
-        if isinstance(reference, AlignedSent):
-            sure = reference.alignment
-        else:
-            sure  = Alignment(reference)
-
-        # Call NLTKs existing functions for recall
-        return recall(sure, align)
-
-
     def alignment_error_rate(self, reference, possible=None):
```
hoontw: Just a stray observation. Why is there a default argument for `possible`? Probably should be removed

nltk/translate/api.py
```diff
@@ -163,51 +157,6 @@ def invert(self):
...
-        align = self.alignment
-        if isinstance(reference, AlignedSent):
-            possible = reference.alignment
-        else:
-            possible = Alignment(reference)
-
-        return precision(possible, align)
-
-
-    def recall(self, reference):
```
hoontw, line 188: Since `recall()` is removed, the doctests in `test\translate.doctest` will have to be updated.

hoontw:
@stevenbird Agree that `AlignedSent.alignment_error_rate()` should just work with `Alignment`. #1129 also discusses the possibility of making it a module function. That could be done in this PR, or I could write something after this PR goes in.

stevenbird:
Thanks @hoontw, I will do that.

stevenbird:
I've added some assertions to produce more helpful error messages in case people pass an alignment string instead of an Alignment to AlignedSent.

stevenbird:
I don't understand why `nltk.metrics.recall` is not returning `None` here:

```
File "...nltk/test/translate.doctest", line 187, in translate.doctest
Failed example:
    recall(Alignment([]), als.alignment)
Expected:
    None
Got nothing
```

NB `len(Alignment([])) == 0`

hoontw:
@stevenbird Because a `None` return value is not printed in the console. Try `print(recall(Alignment([]), als.alignment))`

stevenbird:
I see, thanks @hoontw.
@alvations – would you mind reviewing this PR and merging it if it's ok please?

nltk/translate/__init__.py
```diff
@@ -19,5 +19,6 @@
 from nltk.translate.ibm4 import IBMModel4
 from nltk.translate.ibm5 import IBMModel5
 from nltk.translate.bleu_score import bleu
```
alvations, line 21: Possibly out of scope of this PR but BLEU should be considered in `nltk.translate.metrics` or maybe at some point `nltk.translate.metrics` will become a package by itself because there's a wide range of MT metrics in the literature.

alvations:
@stevenbird Looks great! Thanks for cleaning up the polymorphism and the related functions. 

@hoontw @kmike Thanks for the reviews.

---

Now let me try and see whether I can fit in [RIBES](http://www.kecl.ntt.co.jp/icl/lirg/ribes/) into the BLEU code, it's sort of a BLEU variant with Kendall's tau prior on unigram precision. I sort of need it too =) And also the `PhraseTable` class with reordering.

PR #1161, opened 2015-10-07
Temporarily disable gensim test

longdt219:
Since Gensim is not working with python 3.2, we temporarily remove it. It might be added back once we get answer from Gensim dev team.


kmike:
I think that if we dropped 2.6 support it should be fine to drop 3.2 as well.

nltk/test/gensim_fixt.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+
+def setup_module(module):
+    from nose import SkipTest
+    try:
+        import gensim
+    except ImportError:
+        raise SkipTest("classify.doctest requires numpy")
```
kmike: the error message looks incorrect

stevenbird:
Thanks @kmike. In that case, @longdt219 let's replace 3.2 with 3.3 on the CI server. I'll accept this PR but reinstate the gensim dependency.

nltk/test/gensim_fixt.py
stevenbird: Thanks @kmike – I'll fix this

longdt219:
Replace Python 3.2 by python3.3 through PR #1162

PR #1163, opened 2015-10-11
Add flag to use Stanford neural dependency parser

futurulus:
Adds a flag (`use_nndep=True`) to `nltk.parse.stanford.StanfordDependencyParser` to use @danqi's neural network&ndash;based dependency parser (http://nlp.stanford.edu/software/nndep.shtml). Also makes `DependencyGraph` correctly read CONLL representations with non-consecutive token ids (needed for correct output on one of the existing Stanford dependency parser doctests).

Note the jar requirement and speed tradeoffs in the first commit.


stevenbird:
Thanks @futurulus. Is there a reason to prefer a flag in the initializer for selecting the parser, as compared with deriving another class? I think it would be more in keeping with the other NLTK parsers to have a one-to-one correspondence between external parsers and NLTK parser classes. In this case, both parsers could inherit from a superclass. Then `tagged_parse_sents` could be specified as not implemented in the subclass.

futurulus:
No strong preference here. Splitting it into a separate class would also probably make it cleaner.

I'm assuming we'd want all three parsers to inherit from `GenericStanfordParser`. (As opposed to making a common superclass for the two dependency parsers, which in turn inherits from `GenericStanfordParser`&mdash;I can't think of any code that would usefully go in the dependency-only superclass.)

stevenbird:
This sounds like a good way to go. Do you think you could update your PR?

nltk/parse/dependencygraph.py
```diff
@@ -282,19 +282,24 @@ def _parse(self, input_, cell_extractor=None, zero_based=False, cell_separator=N
 
         def extract_3_cells(cells):
             word, tag, head = cells
-            return word, word, tag, tag, '', head, ''
+            return None, word, word, tag, tag, '', head, ''
```
dimazest: I've been thinking of this, problem, maybe it's better to make extractors to have an  argument `index`, with which the index from the loop below is passed, then it's up to the function to use it or not. something like:

``` python
def extract_3_cells(cells, index):
    word, tag, head = cells
    return index, word, word, tag, tag, '', head, ''

...

def extract_10_cells(cells, index):
    index, word, lemma, ctag, tag, feats, head, rel, _, _ = cells
    return index, word, lemma, ctag, tag, feats, head, rel
```

nltk/parse/dependencygraph.py
```diff
@@ -321,10 +326,15 @@ def extract_10_cells(cells):
                         'CoNLL(10) or Malt-Tab(4) format'.format(cell_number)
                     )
 
-            word, lemma, ctag, tag, feats, head, rel = cell_extractor(cells)
+            line_index, word, lemma, ctag, tag, feats, head, rel = cell_extractor(cells)
```
dimazest: But we need to be careful here, as we might get a cell_extractor which doesn't take index as an argument.

``` python
index, word, lemma, ctag, tag, feats, head, rel = cell_extractor(cells, index=index)
```

is there a reason to have `line_index`?

nltk/parse/dependencygraph.py
```diff
@@ -321,10 +326,15 @@ def extract_10_cells(cells):
...
+            except TypeError: pass  # line_index is probably None, use default
+            except ValueError: pass  # line_index is not an integer, use default
 
+            if head == '_': continue
             head = int(head)
-            if zero_based:
+            if zero_based and line_index is None:
                 head += 1
 
             self.nodes[index].update(
```
dimazest, line 354: should we use `line_index` here in the original proposal?

nltk/parse/dependencygraph.py
```diff
@@ -321,10 +326,15 @@ def extract_10_cells(cells):
...
                     )
 
-            word, lemma, ctag, tag, feats, head, rel = cell_extractor(cells)
+            line_index, word, lemma, ctag, tag, feats, head, rel = cell_extractor(cells)
+            try:
+                index = int(line_index)
+            except TypeError: pass  # line_index is probably None, use default
+            except ValueError: pass  # line_index is not an integer, use default
 
+            if head == '_': continue
```
dimazest: please use

``` python

if head == '_':
    continue
```

as PEP8 suggests.

nltk/parse/dependencygraph.py
futurulus: Sounds sensible to me. I've changed it to use this scheme.

nltk/parse/dependencygraph.py
futurulus: I've wrapped this call in a try/except that should make sure the change is backwards compatible (and added a doctest to verify this). See if you agree or if you see any potential problems.

nltk/parse/dependencygraph.py
futurulus, line 354: Yes, you're right. Changed it to replace index, which I think is easier to read anyway.

nltk/parse/dependencygraph.py
futurulus: Done.

futurulus:
Added new commits to split off the extra class and address @dimazest's comments. I imagine it will make sense to squash some of these together, but I'll leave the history intact until you think the PR can be wrapped up.

stevenbird:
Thanks @futurulus and @dimazest

dimazest:
:+1:

stevenbird:
@futurulus – our CI server is reporting some errors here, and I'd be grateful if you could take a look please.
https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py33-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.parse/stanford/StanfordNeuralDependencyParser/

futurulus:
Ah, after I moved the doctests into the new class, they're no longer correctly skipped if the right jar isn't present. (The neural dependency class needs stanford-corenlp-x.x.x.jar and stanford-corenlp-x.x.x-models.jar from http://nlp.stanford.edu/software/corenlp.shtml , which are not the same as the parser jars.)

We could either:
1. Skip all `stanford.py` doctests if _either_ jar is missing.
2. Move the neural parser doctests back to `test/` and give them a fixture file.
3. Some other way of conditionally skipping doctests that I don't know about?

Recommendations welcome.

stevenbird:
@futurulus – thanks for looking into this. How about adding the extra model here: https://github.com/nltk/nltk/blob/develop/jenkins.sh#L16

PR #1165, opened 2015-10-12
Cleaning up the translate package

alvations:
**Refactoring "candidate" as "hypothesis" in bleu_score.py**
- In the literature, it's more common to refer to machine translated sentences as `hypothesis` instead of `candidate`. 
- `candidate` vs `hypothesis` is a technical term difference but this should be kept consistent with the variable names used in https://github.com/nltk/nltk/blob/develop/nltk/translate/stack_decoder.py  

**Refactoring "test" as "hypothesis in metrics.py**
- Using the consistent term "hypothesis" to refer to automatic alignment or machine translation

**Rearrange all the parameter orders for MT metrics to suit `nltk.metrics`**
- Reordered all the param to keep the order of `func(references, hypothesis, **kwargs)` consistent with functions in `nltk.metrics` (Note: `references` is a list of references and if there's only one reference, explicitly casting to list is necessary).


alvations:
Another point is the clean up where the `PhraseTable` class is moved to `translate.api`.

But because of the use of `namedtuple`. I'm not sure what is the pythonic way to put it in `api.py`.  Would it be:
- create a duck class for the `PhraseTableEntry` namedtuple? 
- create a proper class for  `PhraseTableEntry`?

alvations:
All cleaned up =)

Note: The BLEU doctest and unittest is similar but I think redundancy is good for testing.

stevenbird:
Thanks @alvations

stevenbird:
@alvations – please note that there's a couple of housekeeping issues in the translate module reported on the CI server https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/testReport/

PR #1171, opened 2015-10-13
fix bug shown in CI related to wordnet

longdt219:
Hi @stevenbird, 
Here's the attempt to fix bug related to wordnet shown in CI server. 
Note that I deleted 2 tests involving unicode character as described in the development guideline.


stevenbird:
Thanks @longdt219

PR #1174, opened 2015-10-14
Fix CI issue with Stanford CoreNLP jar 

futurulus:
Makes Jenkins download both Stanford jars, and skips the Stanford parser doctests if either of the jars is missing.

Fixes the CI issue that cropped up after merging #1163.


stevenbird:
Great, thanks @futurulus

Issue #1172, opened 2015-10-14
BLEU unit tests should not be sensitive to floating point precision


stevenbird:
@alvations: the [BLEU unit tests](https://github.com/nltk/nltk/blob/develop/nltk/test/unit/translate/test_bleu.py#L59) produce errors on our CI server like the following. Do you think you could find a way to test with less precision please?

```
Traceback (most recent call last):
  File "/scratch/jenkins/python/python-3.4.2-x86_64/lib/python3.4/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/scratch/jenkins/python/python-3.4.2-x86_64/lib/python3.4/unittest/case.py", line 577, in run
    testMethod()
  File "/scratch/jenkins/workspace/nltk/TOXENV/py34-jenkins/jdk/jdk8latestOnlineInstall/nltk/test/unit/translate/test_bleu.py", line 59, in test__modified_precision
    assert (_modified_precision(references, hyp1, n=1) == 0.9444444444444444)
AssertionError
```

alvations:
@stevenbird sorry for the slow response, I was getting intermittent internet connection while travelling. Hopefully i made it in time for 3.1.

PR #1175, opened 2015-10-15
Fix unittest for bleu_score.py

alvations:
The main problem  [this error](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py33-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.test.unit.translate.test_bleu/TestBLEU/test__modified_precision/) is that the typo in the reference that had additional fullstops so the BLEU scores were different. 

From #1172, I agree that the checking the floating point for all 4 bytes of decimals is not very helpful. I've added the test for rounded BLEU scores too.


stevenbird:
Thanks @alvations

PR #1178, opened 2015-10-19
Replace full double floating precision in BLEU unittest

### LINKED ISSUE #1172: BLEU unit tests should not be sensitive to floating point precision

alvations:
Resolve #1172 by replacing full double floating precision in unittest to using unittest.Testcase.assertAlmostEqual

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -82,6 +82,3 @@ class TestBLEU(unittest.TestCase):
     def test_brevity_penalty(self):
         pass
     
-    
-if __name__ == '__main__':
-    unittest.main()
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -25,7 +25,7 @@ class TestBLEU(unittest.TestCase):
         hyp1_unigram_precision =  _modified_precision(references, hyp1, n=1)
         assert (round(hyp1_unigram_precision, 4) == 0.2857)
         # With assertAlmostEqual at 4 place precision.
-        self.assertAlmostEqual(hyp1_unigram_precision, 0.2857142857142857, places=4)
+        self.assertAlmostEqual(hyp1_unigram_precision, 0.28571428, places=4)
         
         # Testing modified bigram precision.
         assert(_modified_precision(references, hyp1, n=2) == 0.0)
@@ -62,8 +62,8 @@ class TestBLEU(unittest.TestCase):
         hyp1_unigram_precision = _modified_precision(references, hyp1, n=1)
         hyp2_unigram_precision = _modified_precision(references, hyp2, n=1)
         # Test unigram precision with assertAlmostEqual at 4 place precision.
-        self.assertAlmostEqual(hyp1_unigram_precision, 0.9444444444444444, places=4)
-        self.assertAlmostEqual(hyp2_unigram_precision, 0.5714285714285714, places=4)
+        self.assertAlmostEqual(hyp1_unigram_precision, 0.94444444, places=4)
+        self.assertAlmostEqual(hyp2_unigram_precision, 0.57142857, places=4)
         # Test unigram precision with rounding.
         assert (round(hyp1_unigram_precision, 4) == 0.9444)
         assert (round(hyp2_unigram_precision, 4) == 0.5714)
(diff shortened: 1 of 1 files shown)


nltk/test/unit/translate/test_bleu.py
```diff
@@ -59,23 +61,27 @@ def test__modified_precision(self):
...
         assert (round(hyp1_bigram_precision, 4) == 0.5882)
         assert (round(hyp2_bigram_precision, 4) == 0.0769)
         
     def test_brevity_penalty(self):
         pass
-    
\ No newline at end of file
+    
+    
+if __name__ == '__main__':
```
hoontw: These two lines are not necessary to run the unit test. I made the same mistake once :) One way to run this file as a unit test is `python -m unittest test_bleu`.

nltk/test/unit/translate/test_bleu.py
```diff
@@ -22,8 +22,10 @@ def test__modified_precision(self):
         references = [ref1, ref2] 
         
         # Testing modified unigram precision.
-        hyp1_unigram_precision =  _modified_precision(references, hyp1, n=1) 
+        hyp1_unigram_precision =  _modified_precision(references, hyp1, n=1)
         assert (round(hyp1_unigram_precision, 4) == 0.2857)
+        # With assertAlmostEqual at 4 place precision.
+        self.assertAlmostEqual(hyp1_unigram_precision, 0.2857142857142857, places=4)
```
hoontw: I like the use of `assertAlmostEqual(a, b, 4)`, but is there a need to keep the test for `assert(round(a, 4), b)`?

alvations: I think it's nice to show that they are the both valid. That's why I left it there.

nltk/test/unit/translate/test_bleu.py
alvations: Ah thanks for the note on on the unittest. whoops, I forgot to remove it. I used that when i use my IDE, i'll remove that =)

nltk/test/unit/translate/test_bleu.py
hoontw: Personally, I think that 16 decimal places is a bit much. considering that the value is going to be rounded to 4 decimal places in `assertAlmostEqual`. 5 or 6 decimal places should be sufficient.

alvations:
@hoontw I'll change the unittest outputs to 8 decimal points since this is only for testing the function  =)

But just a comment of why keeping the full float is important when using the function in other parts of the main code:

```
>>> import math
>>> from operator import mul

>>> lst = [2/9.0, 1/7.0, 34/298.0, 723/2134.0]
>>> reduce(mul, lst, 1)
0.001227145280283037

>>> lst2 = [round(2/9.0, 4), round(1/7.0, 4), round(34/298.0, 4), round(723/2134.0, 4)]
>>> reduce(mul, lst2, 1)
0.0012274542938504

>>> math.pow(reduce(mul, lst, 1), 1/4.0) * 100
18.71647227854365
>>> math.pow(reduce(mul, lst2, 1), 1/4.0) * 100
18.717650439330172
```

Significance value is too much overestimated in machine translation (https://www.aclweb.org/anthology/W/W14/W14-3333.pdf and http://homepages.inf.ed.ac.uk/pkoehn/publications/bootstrap2004.pdf), so it's kind of important that intermediate steps don't start rounding. Given no brevity penaly, when rounding once it makes a +0.001 BLEU increase and rounding it off to one less decimal would make a +0.01 BLEU which in many scientific papers are counted as "significance" depending on test data size and the resulting p-value.

hoontw:
Yes, I agree that the full float value should be returned in real usage. It's just that in a unit test, it is unwieldy to compare directly to a high level of precision. I think it looks all right now.

stevenbird:
Thanks for the feedback @hoontw.

Issue #1183, opened 2015-10-31
Character ranges in TweetTokenizer regular expressions


erickrf:
As reported in the nltk-dev mailing list, I found that the TweetTokenizer splits words at accented characters. Checking the code, I found at line 159:

```
(?:[a-z][a-z'\-_]+[a-z])       # Words with apostrophes or dashes.
```

A simple solution would be replacing the character ranges by `\w`:

```
(?:\w[\w'\-_]+\w)       # Words with apostrophes or dashes.
```

stevenbird:
@fievelk – would you mind taking a look at this please?

fievelk:
@stevenbird Sure. 
@erickrf Thank you for pointing this out.

fievelk:
@erickrf could you please provide some examples please? I am not able to reproduce your issue using common accented characters.

erickrf:
Sure:

```
tokenizer = nltk.tokenize.TweetTokenizer()
tokens = tokenizer.tokenize(u'testing tokenizer with resumé, españa, münchen, россия')
for token in tokens: print token,
```

Output:

```
testing tokenizer with resum é , espa ña , münchen , россия
```

The regexp groups the first three letters, accented or not. The splits occurs when the accents are after the first three.

fievelk:
**Edit:** I am now able to reproduce your results:

```
>>> import nltk
>>> tokenizer = nltk.tokenize.TweetTokenizer()
>>> tokenizer.tokenize(u'testing tokenizer with resumé, españa, münchen, россия')
['testing', 'tokenizer', 'with', 'resum', 'é', ',', 'espa', 'ña', ',', 'münchen', ',', 'россия']
```

I will make a pull request soon.

ewan-klein:
Resolved in 2f7acb81a496044077af934817e02616c76badce

PR #1185, opened 2015-11-02
Fix issues with accented characters (issue #1183)

fievelk:
Fix issues with accented characters (see issue #1183)

### DIFF
--- a/nltk/tokenize/casual.py
+++ b/nltk/tokenize/casual.py
@@ -156,7 +156,7 @@ REGEXPS = (
 
     # Remaining word types:
     r"""
-    (?:[a-z][a-z'\-_]+[a-z])       # Words with apostrophes or dashes.
+    (?:\w[\w'\-_]+\w)              # Words with apostrophes or dashes.
     |
     (?:[+\-]?\d+[,/.:-]\d+[+\-]?)  # Numbers, including fractions, decimals.
     |


ewan-klein:
Thanks @fievelk

nltk/tokenize/casual.py
```diff
@@ -156,7 +156,7 @@
 
     # Remaining word types:
     r"""
-    (?:[a-z][a-z'\-_]+[a-z])       # Words with apostrophes or dashes.
+    (?:\w[\w'\-_]+\w)              # Words with apostrophes or dashes.
```
nschneid, line 159: Note that `\w` [includes](https://docs.python.org/2/library/re.html#regular-expression-syntax) all alphanumeric characters in the locale, plus `[0-9_]`. So the underscore here, and many of the underscores elsewhere in the file, are redundant.

If the intention is to match alphabetic characters only, `[^\W\d_]` is one solution: http://stackoverflow.com/a/12349512

fievelk, line 159: @nschneid Thank you. Do you think that, following your suggestion, this would be a better solution for the whole line?

```
(?:[^\W\d_](?:[^\W\d_]|['\-_])+[^\W\d_])
```

nschneid, line 159: If the goal is to match alphabetic sequences optionally containing `-`, `'`, or `_` in the middle.

It might also be worth including the curly right single quote/apostrophe `’` at the beginning (_’twas_) or middle of the word. But I would guess that straight `'` at the beginning is more likely to be an open quote than an apostrophe.

ewan-klein, line 159: We ought to be making our assumptions about `casual.py` explicit for regression testing in a file, i.e. in `test/tokenize.doctest`. @fievelk do you have time to do a first pass on this?

fievelk:
Sure, I'll work on it.

nltk/tokenize/casual.py
fievelk, line 159: Sure, I'll work on it.

PR #1190, opened 2015-11-05
Ptb tokenize with offsets

leondz:
The string_span_tokenize() functionality is very useful, critical for anyone working with standoff annotations. This simple function, intended originally for just the PTB tokenizer but actually tokenizer-independent, creates span tuples for any matching token list / source sentence pair.

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -115,4 +115,31 @@ Post-hoc alignment of tokens with a source string
     >>> from nltk.tokenize.util import *
     >>> list(align_tokens(word_tokenize(s1), s1))
     [(0, 2), (3, 4), (5, 6), (6, 12), (13, 21), (22, 24), (25, 27), (28, 33), (34, 36), (37, 38), (39, 46), (46, 47), (48, 51), (52, 59), (60, 67), (68, 73), (74, 76), (77, 78), (78, 84), (84, 85)]
-    
+    >>> list(align_tokens([''], ""))
+    [(0, 0)]
+    >>> list(align_tokens([''], " "))
+    [(0, 0)]
+    >>> list(align_tokens([], ""))
+    []
+    >>> list(align_tokens([], " "))
+    []
+    >>> list(align_tokens(['a'], "a"))
+    [(0, 1)]
+    >>> list(align_tokens(['abc', 'def'], "abcdef"))
+    [(0, 3), (3, 6)]
+    >>> list(align_tokens(['abc', 'def'], "abc def"))
+    [(0, 3), (4, 7)]
+    >>> list(align_tokens(['ab', 'cd'], "ab cd ef"))
+    [(0, 2), (3, 5)]
+    >>> list(align_tokens(['ab', 'cd', 'ef'], "ab cd ef"))
+    [(0, 2), (3, 5), (6, 8)]
+    >>> list(align_tokens(['ab', 'cd', 'efg'], "ab cd ef"))
+    Traceback (most recent call last):
+    ....
+    ValueError: ('Sentence ended while scanning for token start', 2)
+    >>> list(align_tokens(['ab', 'cd', 'ef', 'gh'], "ab cd ef"))
+    Traceback (most recent call last):
+    ....
+    ValueError: ('Sentence ended while scanning for token start', 3)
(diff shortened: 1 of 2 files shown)


jnothman:
Did you mean

``` python
def align_tokens(tokens, s):
    start = 0
    for token in tokens:
        try:
            start = s.index(token, start)
        except ValueError:
            raise 'Substring %r not found following offset %d' % (token, start)
        yield start, start.len(token)
        start += len(token)
```

?

In practice I've found one requires a more complex alignment algorithm because of noise and normalisation in the tokenised text.

I agree that tools for managing standoff annotation could be a useful addition to NLTK. Additionally, standoff representations of XML can be useful: `xml_to_standoff('hello <b class="foo">world</b>') == ('hello world', ['b', {'class': 'foo'}, 6, 10]))`; `standoff_to_xml` would reverse this transformation (assuming provided tags were properly nested).

leondz:
Sure, this seems fine too (perhaps avoiding the placeholders %).

Indeed, a lack of standoff support for something like this breaks a lot of
our research - with e.g. TEI docs, JSON - and is something considered
essential, and appreciated, in our GATE toolkit. After all, distributing
CONLL docs with an enforced tokenisation is rather lossy and a one-way
process; one cannot reliably reconstruct the source document given a
sequence of tokens. There's really not much case for not using standoff,
from here!

Supporting tokenizers that normalize halfway seems a thankless and slightly destructive task; it's not very scientific to change the observation, and in our case, the observation is the text. So I don't mind raising exceptions for that.

stevenbird:
@leondz sorry for the delay in processing your PR. Since you don't object to raising exceptions for mismatches, would you mind updating your PR with @jnothman's suggestion please?

alvations:
There're several things that will happen when stitching back tokens to the original strings. As @jnothman  noted, there are several string replacements and normalization that can occurring during tokenization and they need extra work before you can align the tokenized string to the original string.

E.g. at https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L49, quotation marks are replaced by strings not found in the original string. 

Usually, to get "detokenizing power", the offsets needs to be saved when tokenizing, like how Stanford or REPP tokenizers do it. Then stitching back the tokens will be easy. [Stanford Untokenizer](https://github.com/stanfordnlp/CoreNLP/blob/8b5f36aa7b83e5954bd94974de934156af325403/src/edu/stanford/nlp/process/PTBTokenizer.java) might be a good reference. 

But i think it's still a valid try to add detokenizing functionality but users must be warned about that it will not work 100%. Otherwise, we'll slowly be adding more and more little hacks to this detokenizer and it might lead to a long tail of regexes. 

@leondz  Can you add some tests on tokenizing and detokenizing these sentences:

```
text = u'The https://github.com/jonsafari/tok-tok/blob/master/tok-tok.pl is a website with/and/or slashes and sort of weird : things'
text = u'Is 9.5 or 525,600 my favorite number?'
text = u'This, is a sentence with weird\xbb symbols\u2026 appearing everywhere\xbf'
text = u"‘It is a capital mistake to theorize before one has data.'" # Watson's quote.
text = u'"Insensibly one begins to twist facts to suit theories, instead of theories to suit facts.’" # Watson's quote.

```

leondz:
Yes, I think untokenising is really rather an undesirable approach, with all the issues it can bring. The best approach is, hands-down, generating offsets within the original tokenisation process. However, to cater as a baseline catch-all for tokenisers, and get _something_ in place with stand-off, it's a start. The next step would be to have PTBTokenizer (or whatever is default) generate offsets natively.

Thanks for the tests - I'll reword this pull req and update accordingly as soon as I get a few free hours :)

stevenbird:
Thanks @alvations.
@leondz: I'm happy to proceed with this once you've had a chance to do the updates, thanks.

stevenbird:
@leondz: I'm hoping we can finalise this in time for NLTK 3.3 in February.

alvations:
@leondz I think it's still good to have options for detokenization using the function you've written. 
Could you do a git pull from the fresh develop branch of `nltk` and then maybe call it something like `leodz_untokenize()`? 

I'm bad with names, if there's a paper you've used the detokenization, it'll be good to just rename the function based on that =)

PR #1192, opened 2015-11-05
Fix TweetTokenizer problems with special characters and add tests

fievelk:
This PR fixes regular expression for words with accented characters in `TweetTokenizer` (see #1183) and adds relative tests.
- Many doctest lines have been moved from `tokenize/casual.py` to `test/tokenize.doctest`.
- Following [NLTK - Local Testing](http://www.nltk.org/dev/local_testing.html), tests relative to non-ascii characters are provided as unittests in `test/unit/test_tokenize.py`.

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -147,9 +147,9 @@ It is possible to specify `strip_handles` and `reduce_len` parameters for a Twee
     >>> s7 = '@_willy65: No place for @chuck tonight. Sorry.'
     >>> tknzr.tokenize(s7)
     [':', 'No', 'place', 'for', 'tonight', '.', 'Sorry', '.']
-    >>> s8 = '@mar_tin is a great developer. Contact him at mar_tin@email.com'
+    >>> s8 = '@mar_tin is a great developer. Contact him at mar_tin@email.com.'
     >>> tknzr.tokenize(s8)
-    ['is', 'a', 'great', 'developer', '.', 'Contact', 'him', 'at', 'mar_tin', '@email', '.', 'com']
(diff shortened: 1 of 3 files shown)


nltk/test/tokenize.doctest
```diff
@@ -109,3 +109,51 @@ A simple sentence tokenizer '\.(\s+|$)'
...
+    >>> tknzr = TweetTokenizer(strip_handles=True, reduce_len=True)
+    >>> s6 = '@remy: This is waaaaayyyy too much for you!!!!!!'
+    >>> tknzr.tokenize(s6)
+    [':', 'This', 'is', 'waaayyy', 'too', 'much', 'for', 'you', '!', '!', '!']
+    >>> s7 = '@_willy65: No place for @chuck tonight. Sorry.'
+    >>> tknzr.tokenize(s7)
+    [':', 'No', 'place', 'for', 'tonight', '.', 'Sorry', '.']
+    >>> s8 = '@mar_tin is a great developer. Contact him at mar_tin@email.com'
+    >>> tknzr.tokenize(s8)
+    ['is', 'a', 'great', 'developer', '.', 'Contact', 'him', 'at', 'mar_tin', '@email', '.', 'com']
```
nschneid: Shouldn't it be possible to use a regex for email addresses to avoid this behavior?

fievelk: I agree that it would be good to have such a regex. I think that Twitter usernames cannot contain dots, so we could use this information to distinguish usernames from email addresses domains.

fievelk:
I added a regex for email addresses, as suggested by @nschneid, and tests are passing. I would be grateful if someone could please double-check it :mag_right: :blush:

stevenbird:
Thanks @fievelk – I checked it and it looks good to me.

Issue #1194, opened 2015-11-07
ngram regression


stevenbird:
The CI server reports errors such as the following for the ngram functions (build [456](https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py27-jenkins,jdk=jdk8latestOnlineInstall/456/) following merge of #1179).

```
Failed example:
    list(bigrams([1,2,3,4,5]))
Exception raised:
    Traceback (most recent call last):
      File "<doctest nltk.util.bigrams[1]>", line 1, in <module>
        list(bigrams([1,2,3,4,5]))
      File "/scratch/jenkins/workspace/nltk/TOXENV/py27-jenkins/jdk/jdk8latestOnlineInstall/nltk/util.py", line 483, in bigrams
        for item in ngrams(sequence, 2, **kwargs):
      File "/scratch/jenkins/workspace/nltk/TOXENV/py27-jenkins/jdk/jdk8latestOnlineInstall/nltk/util.py", line 460, in ngrams
        history.append(next(sequence))
    TypeError: list object is not an iterator
```

stevenbird:
@alvations – would you be able to have a look at this please?

alvations:
@stevenbird , sorry for missing that out in the previous merge. The fix #1195  removes the checks and converts into iterable when padding.

stevenbird:
Thanks @alvations

PR #1195, opened 2015-11-07
Enforcing padding on ngrams

alvations:
This is a fix for #1194

Previously, all sequences/iterators were converted into iterables before proceeding with ngram extraction, from https://github.com/nltk/nltk/blob/554a1710ef1011ba65168d442d038763c517e292/nltk/util.py#L411

But after abstracting the `pad_sequence()` away from the ngrams function, the `iterator->iterable` conversion happens [there](https://github.com/alvations/nltk/blob/develop/nltk/util.py#L407) and that would make the checks for pad_left and pad_right within the `ngram()` obsolete because now `pad_sequence()` does the check. 

Sorry for missing this out in the previous merge.

### DIFF
--- a/nltk/util.py
+++ b/nltk/util.py
@@ -451,7 +451,7 @@ def ngrams(sequence, n, pad_left=False, pad_right=False,
     :rtype: sequence or iter
     """
     sequence = pad_sequence(sequence, n, pad_left, pad_right,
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for the prompt fix @alvations

PR #1197, opened 2015-11-09
Add further sphinx docs to read_* methods in nltk.internals

mrecachinas:
I was going through nltk/internals.py, specifically `read_str`, `read_int`, and `read_number`, and I saw they didn't have any :Example:'s for `doctest` as well as :param: and :return: (and their corresponding :type:'s), so I added them. I don't believe in this case it is overkill, as explicitly labeling type and providing examples can only help clarify the use.


stevenbird:
Thanks @mrecachinas. Please feel free to add your name to AUTHORS.md.

mrecachinas:
Thanks! See #1199.

Issue #1166, opened 2015-10-12
Trie implementation


stevenbird:
Consolidate `tokenize.mwe.MWETokenizer.add_mwe` and `sem.logic.StringTrie` into `util.Trie`.

mrecachinas:
I'm currently working on ensuring the `util.Trie` I've adapted from `sem.logic.StringTrie` maintains compatability with both `sem.logic.LogicParser.process` and `tokenize.mwe.MWETokenizer.add_mwe`.

stevenbird:
+1

stevenbird:
@mrecachinas – our [CI Server](https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py34-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/) reports test failures that involve the new Trie object:

```
AttributeError: 'Trie' object has no attribute 'iteritems'
```

Would you be able to track down the problem and submit a new PR please?

mrecachinas:
Very odd. I'll check it out.

mrecachinas:
@stevenbird -- In Python 3, `dict.iteritems` was removed because `dict.items` has optimized what `iteritems` had done and now returns a `memoryview` (an iterator that does much more). 

There are a few different ways to fix this that will ensure compatibility between 2 and 3:
- Add an `iteritems` method to `Trie` that will simply `return self.items()`.
  - Downside: Performance hit in Python2 because we're no longer using an iterator; we're now building a full list of tuples unpacked from the `dict` or `defaultdict` (base class of `Trie`).
  - Upside: Full compatibility between 2 and 3
- Change the Python3 unit test to call `items()`
  - Downside: Disconnect in the API?
  - Upside: `items` is in both 2 and 3, so the tests remain consistent.

What are your thoughts?

erip:
Any status on this? @stevenbird

iliakur:
This is coming a little bit out of left field, but would it be worth considering using @kmike's [DaTrie Implementation](https://github.com/kmike/datrie) for its speed? Or is it a problem to have such an external dependency?

mrecachinas:
+1 @Copper-Head This is a good question. I'd be curious about it too. cc @stevenbird @kmike

alvations:
+1 for @kmike's Trie dependency.

stevenbird:
[...]

PR #1200, opened 2015-11-10
Trie implementation

mrecachinas:
As per #1166, consolidated `tokenize.mwe.MWETokenizer.add_mwe` and `sem.logic.StringTrie` into `util.Trie`.

### DIFF
--- a/nltk/sem/logic.py
+++ b/nltk/sem/logic.py
@@ -17,6 +17,7 @@ import operator
 from collections import defaultdict
 from functools import reduce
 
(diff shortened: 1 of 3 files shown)


stevenbird:
Thanks @mrecachinas – this is looking really good. I just picked up a couple of little things.

stevenbird:
Thanks very much for the extra work @mrecachinas

mrecachinas:
Glad to do it :+1:

Issue #1203, opened 2015-11-11
Multext reader returns map object


stevenbird:
The `tagged_sents` and `tagged_words` methods should return a sequence type not a map object.

```
>>> nltk.corpus.brown.tagged_words()
>>> [('The', 'AT'), ('Fulton', 'NP-TL'), ...]
>>> nltk.corpus.multext_east.tagged_words("oana-en.xml", "universal")
 <map object at 0x10b34dfd0>
>>> 
```

stevenbird:
@stieglma: would you please look into this?

stieglma:
This seems to be a problem with Python 3. It can easily be fixed by wrapping the return value of the mentioned functions in `list(...)`. This is also what I would do. But I have a question on this: When executing the tagged_words() function on the brown corpus it only shows the first two elements and not the whole list. With the multext-east reader it will always print the whole list. I couldn't figure out how you achieve this. Is it necessary that it works like that, or is it ok if the whole list is printed?

nschneid:
@stieglma, that's done with the lazy sequence API in http://www.nltk.org/api/nltk.html#module-nltk.util. I think lazy loading should be preferred except for very small corpora.

stieglma:
I created the pull request quite a while ago, did you just forget to merge it, or is something wrong with it?

stevenbird:
Thanks @stieglma! I'm getting to it...

stevenbird:
Thanks @stieglma

PR #1207, opened 2015-11-15
Fixed issue  #1203

stieglma:
In python 3, instead of a list/sequence some methods (e.g. filter/reduce) return iterators (c.f. http://stackoverflow.com/questions/1303347/getting-a-map-to-return-a-list-in-python-3-x) this was fixed by surrounding these calls with list(...). Additionally I implemented lazy corpus loading.

This fixes issues #1203 and the mentioned problem with displaying the whole corpus at once (in contrast to other corpus readers where it works in a different way)


stevenbird:
Thanks @stieglma

stevenbird:
Sorry for the delay. Thanks for the solid work.

PR #1208, opened 2015-11-17
Added a re-implementation of RIBES

alvations:
RIBES is a popular MT evaluation metric, especially for [Asian language pairs](http://lotus.kuee.kyoto-u.ac.jp/WAT). The official scripts can be found at http://www.kecl.ntt.co.jp/icl/lirg/ribes/

This re-implementation provides an educational documentation of the various parts of the algorithm described in the paper. 

Since the licenses differs for NLTK and RIBES, having RIBES code in verbatim would be technically wrong. Users are still encouraged to use the official script when evaluating their MT system.

Note that the emulation code was removed at https://github.com/alvations/nltk/commit/520a6760b3012aa0fd245264741ad4d9563645da at the suggestion of the original RIBES maintainers.

### DIFF
--- a/nltk/translate/ribes_score.py
+++ b/nltk/translate/ribes_score.py
@@ -102,7 +102,7 @@ def word_rank_alignment(reference, hypothesis, character_based=False):
     w.r.t. the list of reference words.
     
     Below is (H0, R0) example from the Isozaki et al. 2010 paper, 
-    note the examples are indexed from 1th but the results here starts from 0th:
+    note the examples are indexed from 1 but the results here are indexed from 0:
     
         >>> ref = str('he was interested in world history because he '
         ... 'read the book').split()
--- a/nltk/translate/ribes_score.py
+++ b/nltk/translate/ribes_score.py
@@ -304,8 +304,3 @@ def spearman_rho(worder, normalize=True):
         return (rho + 1) /2
     else: # Otherwise, the rho outputs falls between -1.0 to +1.0
         return rho
-
-
-if __name__ == "__main__":
-    import doctest
-    doctest.testmod()
--- a/nltk/translate/ribes_score.py
+++ b/nltk/translate/ribes_score.py
@@ -9,11 +9,13 @@
 """ RIBES score implementation """
 
 from itertools import islice
+import math
 
 from nltk import ngrams
-from bleu_score import _brevity_penalty
 
-def ribes(references, hypothesis, alpha=0.25, beta=1.0):
+
+def ribes(references, hypothesis, alpha=0.25, beta=0.10, 
+          character_based=False, emulate_official=False):
     """
     The RIBES (Rank-based Intuitive Bilingual Evaluation Score) from 
     Hideki Isozaki, Tsutomu Hirao, Kevin Duh, Katsuhito Sudoh and 
@@ -34,23 +36,39 @@ def ribes(references, hypothesis, alpha=0.25, beta=1.0):
     :type alpha: float
     :param beta: hyperparameter used as a prior for the brevity penalty.
     :type beta: float
+    :param character_based: a flag to emulate the original character-based RIBES
+    :type character_based: boolean 
     """
-    bp = _brevity_penalty(references, hypothesis)
     _best_ribes = -1.0
     # Calculates RIBES for each reference and returns the best score.
     for reference in references:
-        # Collects the *worder* from the ranked correlation alignments.
-        worder = word_rank_alignment(reference, hypothesis)
+        # To emulate the official character-based RIBES implementation
+        if character_based:
+            reference = list("".join(reference))
+            hypothesis = list("".join(hypothesis))
+        
+        # Calculates the normalized kendall Tau, *nkt*
+        if emulate_official: # Emulates the official RIBES v1.03.1
+            # Collects the *worder* from the ranked correlation alignments.
+            worder = _ribes_word_rank_alignment(reference, hypothesis)
+            nkt = _ribes_kendall_tau(worder)
+        else:
+            # Collects the *worder* from the ranked correlation alignments.
+            worder = word_rank_alignment(reference, hypothesis)
+            nkt = kendall_tau(worder)
+            
+        # Calculates the brevity penalty
+        bp = min(1.0, math.exp(1.0 - 1.0 * len(reference)/len(hypothesis)))
+        
         # Calculates the unigram precision, *p1*
         p1 = 1.0 * len(worder) / len(hypothesis)
-        # Calculates the normalized kendall Tau, *nkt*
-        nkt = kendall_tau(worder)
-        _ribes = nkt * (alpha**p1) *  (beta**bp)
+        
+        _ribes = nkt * (p1 ** alpha) *  (bp ** beta)
         
         if _ribes > _best_ribes: # Keeps the best score.
             _best_ribes = _ribes
-            
-    return _best_ribes
+        
+    return _best_ribes, p1, bp, worder, nkt
 
 
 def position_of_ngram(ngram, sentence):
@@ -79,9 +97,9 @@ def position_of_ngram(ngram, sentence):
         # Returns the index of the word when ngram matches.
         if ngram == sublist:
             return i
-        
 
-def word_rank_alignment(reference, hypothesis):
+
+def word_rank_alignment(reference, hypothesis, character_based=False):
     """    
     This is the word rank alignment algorithm described in the paper to produce
     the *worder* list, i.e. a list of word indices of the hypothesis word orders 
@@ -157,7 +175,8 @@ def word_rank_alignment(reference, hypothesis):
                     if num_times_in_ref == num_times_in_hyp == 1:
                         # Find the position of ngram that matched the reference.
                         pos = position_of_ngram(left_context_ngram, reference)
-                        worder.append(pos)  # Add the positions of the ngram.
+                        # Add the positions of the ngram.
+                        worder.append(pos+ len(left_context_ngram) -1)  
                         break
     return worder
 
@@ -187,6 +206,7 @@ def choose(n, k):
         return ntok // ktok
     else:
         return 0
+
     
 def find_increasing_sequences(worder):
     """
@@ -211,6 +231,7 @@ def find_increasing_sequences(worder):
                 yield tuple(result)
             result = [b]
 
+
 def kendall_tau(worder, normalize=True):
     """
     Calculates the Kendall's Tau correlation coefficient given the *worder*
@@ -229,8 +250,16 @@ def kendall_tau(worder, normalize=True):
         >>> round(kendall_tau(worder),3)
         0.382
     
+    To compare with the _ribes_kendall_tau():
+    
+        >>> worder = [0, 1, 8, 9, 10, 11, 12, 13, 14, 15, 24, 17, 18, 19, 20, 25, 26]
+        >>> round(kendall_tau(worder), 3)
+        0.265
+    
     :param worder: The worder list output from word_rank_alignment
-    :param type: list(int)
+    :type worder: list(int)
+    :param normalize: Flag to indicate normalization
+    :type normalize: boolean
     """
     worder_len = len(worder)
     # Extract the groups of increasing/monotonic sequences.
@@ -241,7 +270,6 @@ def kendall_tau(worder, normalize=True):
     num_possible_pairs = choose(worder_len, 2)
     # Kendall's Tau computation.
     tau = 2 * num_increasing_pairs / num_possible_pairs -1
-    
     if normalize: # If normalized, the tau output falls between 0.0 to 1.0
         return (tau + 1) /2
     else: # Otherwise, the tau outputs falls between -1.0 to +1.0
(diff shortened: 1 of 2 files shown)


stevenbird:
@hoontw – would you mind reviewing this contribution please?

hoontw:
I will review this over the next few days.

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+    Please note that this re-implementation differs from the official
+    RIBES implementation and though it emulates the results as describe
+    in the original paper, there are further optimization implemented 
+    in the official RIBES script.
+    
+    Users are encouraged to use the official RIBES script instead of this 
+    implementation when evaluating your machine translation system. Refer
+    to http://www.kecl.ntt.co.jp/icl/lirg/ribes/ for the official script.
+    
+    :param reference: a reference sentence
```
hoontw: Should this be `references`?

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+    :type reference: list(str)
+    :param hypothesis: a hypothesis sentence
+    :type hypothesis: list(str)
+    :param alpha: hyperparameter used as a prior for the unigram precision.
+    :type alpha: float
+    :param beta: hyperparameter used as a prior for the brevity penalty.
+    :type beta: float
+    :param character_based: a flag to emulate the original character-based RIBES
```
hoontw: `character_based` doesn't seem to be relevant anymore

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+    :type character_based: boolean
```
hoontw: Would it help to have a doctest here to illustrate usage and the expected output?

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    _best_ribes = -1.0
+    # Calculates RIBES for each reference and returns the best score.
+    for reference in references:
+        # Collects the *worder* from the ranked correlation alignments.
+        worder = word_rank_alignment(reference, hypothesis)
+        nkt = kendall_tau(worder)
+            
+        # Calculates the brevity penalty
+        bp = min(1.0, math.exp(1.0 - 1.0 * len(reference)/len(hypothesis)))
```
hoontw, line 57: `bp = max(1.0, math.exp(1.0 - 1.0 * len(reference)/len(hypothesis)))` seems to make more sense to me

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+        
+        # Calculates the unigram precision, *p1*
+        p1 = 1.0 * len(worder) / len(hypothesis)
+        
+        _ribes = nkt * (p1 ** alpha) *  (bp ** beta)
+        
+        if _ribes > _best_ribes: # Keeps the best score.
+            _best_ribes = _ribes
+        
+    return _best_ribes, p1, bp, worder, nkt
```
hoontw: Some documentation about the return values would be nice :)

nltk/translate/ribes_score.py
alvations: Brevity penalty is to reduce the score by penalizing the long hypothesis that gets higher kendall's tau unigram precision. So smaller bp means lesser penaly and that means that the hypothesis is close to the reference in terms of length, thus `min(...)` here is right. `max(1.0, ...)` would always result in 1.0 because ` 0 < math.exp(1.0 - 1.0 * len(reference)/len(hypothesis)) < 1.0` is true for all length of hypothesis and reference.

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def word_rank_alignment(reference, hypothesis, character_based=False):
+    """    
+    This is the word rank alignment algorithm described in the paper to produce
+    the *worder* list, i.e. a list of word indices of the hypothesis word orders 
+    w.r.t. the list of reference words.
+    
+    Below is (H0, R0) example from the Isozaki et al. 2010 paper, 
+    note the examples are indexed from 1th but the results here starts from 0th:
```
hoontw: grammar nit: `are indexed from 1 but the results here are indexed from 0:`

nltk/translate/ribes_score.py
alvations, line 57: `min(...)` is right here because brevity penalty penalizes short hypothesis because precision is calculated as such  `precision = 1.0 * n / len(hyp)`. So the shorter the `len(hyp)` the higher the precision. These example will illustrate the penalty.

When hypothesis is close to the reference in length, we don't penalize the unigram precision as much, e.g.:

```
>>> import math
>>> lenref, lenhyp = 10, 9
>>> n = 5
>>> precision = 1.0 * n / lenhyp
>>> bp = min(1.0, math.exp(1.0 - 1.0 * lenref/lenhyp))
>>> precision
0.5555555555555556
>>> bp
0.8948393168143697
>>> bp * precision 
0.49713295378576094
```

But when it's much shorter, we should penalize it more:

```
>>> import math
>>> lenref, lenhyp = 10, 6
>>> n = 5
>>> precision = 1.0 * n / lenhyp
>>> precision
0.8333333333333334
>>> bp = min(1.0, math.exp(1.0 - 1.0 * lenref/lenhyp))
>>> bp
0.513417119032592
>>> bp * precision
0.4278475991938267
```

And if the hypothesis is super short, precision is can still be relatively high by it didn't match the longer reference so it needs to be penalize

```
>>> lenref, lenhyp = 10, 4
>>> n = 3
>>> precision = 1.0 * n / lenhyp
>>> precision
0.75
>>> bp = min(1.0, math.exp(1.0 - 1.0 * lenref/lenhyp))
>>> bp * precision 
0.16734762011132237
```

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+    :param hypothesis: a hypothesis sentence
+    :type hypothesis: list(str)
+    """
+    worder = []
+    hyp_len = len(hypothesis)
+    # Stores a list of possible ngrams from the reference sentence.
+    # This is used for matching context window later in the algorithm.
+    ref_ngrams = []
+    hyp_ngrams = []
+    for n in range(1, len(reference)+1):
```
hoontw, line 139: The original paper only considered unigrams and bigrams. Do current implementations consider ngrams of any length?

alvations, line 139: The current implementation from the official maintainers considers ngrams of any length. 

The results reported in the paper considers all ngram lengths too, otherwise it will be hard to achieve the correlation results they've gotten. Only in the examples, they simplify the explanation and they use unigrams and bigrams.

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+    for i, h_word in enumerate(hypothesis):
+        # If word is not in the reference, continue.
+        if h_word not in reference:
+            continue
+        # If we can determine one-to-one word correspondence for unigrams that 
+        # only appear once in both the reference and hypothesis.
+        elif hypothesis.count(h_word) == reference.count(h_word) == 1:
+            worder.append(reference.index(h_word))
+        else:
+            # Note: range(1, max(i, hyp_len-i+1)) is the range of window sizes.
```
hoontw: nit: could replace comment with a desciptive variable. Example:

```
max_window_size = max(i, hyp_len-i+1)
for window in range(1, max_window_size):
```

nltk/translate/ribes_score.py
alvations: Whoops that's my mistake, it should only return _best_ribes. The rest of the return values was used for debugging the code =)

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+                    if num_times_in_ref == num_times_in_hyp == 1:
+                        # Find the position of ngram that matched the reference.
+                        pos = position_of_ngram(left_context_ngram, reference)
+                        # Add the positions of the ngram.
+                        worder.append(pos+ len(left_context_ngram) -1)  
+                        break
+    return worder
+
+ 
+def choose(n, k):
```
hoontw: Wasn't a nCr function introduced into NLTK recently?

alvations: Yep, I should import this properly. It's the same function.

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,311 @@
...
+        >>> worder = [7, 8, 9, 10, 6, 0, 1, 2, 3, 4, 5]
+        >>> list(find_increasing_sequences(worder))
+        [(7, 8, 9, 10), (0, 1, 2, 3, 4, 5)]
+    
+    :param worder: The worder list output from word_rank_alignment
+    :param type: list(int)
+    """
+    items = iter(worder)
+    a, b = None, next(items, None)
+    result = [b]
```
hoontw, line 193: If `worder` is `[]`, is it ok for this function to return `[None]`?

alvations, line 193: If `worder == []`, `find_increasing_sequence()` will not return `None` or `[None]` but an empty generator because it doesn't return a list but a generator. And the generator will be iterated later to calculate the kendall's tau, so when doing a `sum(iter([]))` it returns a 0. E.g.

```
>>> worder = []
>>> len(worder)
0
>>> def find_increasing_sequences(worder):
...     """
...     Given the *worder* list, this function groups monotonic +1 sequences. 
...     
...         >>> worder = [7, 8, 9, 10, 6, 0, 1, 2, 3, 4, 5]
...         >>> list(find_increasing_sequences(worder))
...         [(7, 8, 9, 10), (0, 1, 2, 3, 4, 5)]
...     
...     :param worder: The worder list output from word_rank_alignment
...     :param type: list(int)
...     """
...     items = iter(worder)
...     a, b = None, next(items, None)
...     result = [b]
...     while b is not None:
...         a, b = b, next(items, None)
...         if b is not None and a + 1 == b:
...             result.append(b)
...         else:
...             if len(result) > 1:
...                 yield tuple(result)
...             result = [b]
... 
>>> find_increasing_sequences([])
<generator object find_increasing_sequences at 0x7f9fbe71ca00>
>>> increasing_sequences = find_increasing_sequences([])
>>> sum(i for i in increasing_sequences)
0
>>> sum(iter([]))
0
```

nltk/translate/ribes_score.py
hoontw, line 139: I see, thanks!

alvations:
@hoontw thanks for looking through the code! Please continue to review if you have yet finished but I've got to go offline for a while. And I'll get back to the edits tomorrow.

nltk/translate/ribes_score.py
hoontw, line 57: Sorry, you're right. I was thinking about the case when length of hypothesis > length of reference. My brain isn't working at this time of night :)

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,280 @@
...
+
+def kendall_tau(worder, normalize=True):
+    """
+    Calculates the Kendall's Tau correlation coefficient given the *worder*
+    list of word alignments from word_rank_alignment(), using the formula:
+    
+        tau = 2 * num_increasing_pairs / num_possible pairs -1
+    
+    Note that the no. of increasing pairs can be discontinuous in the *worder*
+    list and each increase sequence can be tabulated as choose(len(seq), 2) no.
```
hoontw: grammar: each increasing sequence

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,280 @@
...
+def kendall_tau(worder, normalize=True):
+    """
+    Calculates the Kendall's Tau correlation coefficient given the *worder*
+    list of word alignments from word_rank_alignment(), using the formula:
+    
+        tau = 2 * num_increasing_pairs / num_possible pairs -1
+    
+    Note that the no. of increasing pairs can be discontinuous in the *worder*
+    list and each increase sequence can be tabulated as choose(len(seq), 2) no. 
+    of increasing pair, e.g.
```
hoontw: grammar: increasing pairs

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,280 @@
...
+    
+        >>> worder = [7, 8, 9, 10, 6, 0, 1, 2, 3, 4, 5]
+        >>> number_possible_pairs = choose(len(worder), 2)
+        >>> round(kendall_tau(worder, normalize=False),3)
+        -0.236
+        >>> round(kendall_tau(worder),3)
+        0.382
+    
+    To compare with the _ribes_kendall_tau():
```
hoontw: I believe `_ribes_kendall_tau` has been removed

nltk/translate/ribes_score.py
```diff
@@ -0,0 +1,280 @@
...
+    
+        rho = 1 - sum(d**2) / choose(len(worder)+1, 3)  
+        
+    Given that d is the sum of difference between the *worder* list of indices
+    and the original word indices from the reference sentence.
+    
+    Using the (H0,R0) and (H5, R5) example from the paper
+    
+        >>> worder =  [7, 8, 9, 10, 6, 0, 1, 2, 3, 4, 5]
+        >>> sum((wi - i)**2 for wi, i in zip(worder, range(len(worder))))
```
hoontw: This line doesn't seem to belong in the doctest because it is about internal calculations. `worder` is also defined twice, so one of them could be removed.

nltk/translate/ribes_score.py
hoontw, line 193: Thanks for the detailed explanation and verifying it still works with edge cases. I'm rusty on Python generators.

nltk/translate/ribes_score.py
alvations, line 57: No problem =) I just realized, I forgot this example that will show the rationale behind brevity penalty. Even when all ngrams in hypothesis matches, i.e. `n == 4; lenhyp == 4`, it is still quite a bad translation if it only translate a small bit of the sentence. 

```
>>> lenref, lenhyp = 10, 4
>>> n = 4
>>> precision = 1.0 * n / lenhyp
>>> bp = min(1.0, math.exp(1.0 - 1.0 * lenref/lenhyp))
>>> bp * precisio
NameError: name 'precisio' is not defined
>>> bp * precision
0.22313016014842982
```

nltk/translate/ribes_score.py
alvations: I've removed the extra `worder` initialization but the spearman line (sum((wi - i)**2 for wi, i in zip(worder, range(len(worder))))`) that outputs 350 comes from the example from the paper e.g. (H5, R5).

hoontw: I think the extra worder initialization is still here.
For the sum of squares, I am not sure if users will be interested in the intermediate steps in deriving spearman's rho. I do not feel strongly about this, so if you want to leave it in, that's ok.

alvations:
@hoontw Thanks again for the reviews. I think I've fixed most of the suggested edits, I hope I didn't miss anything. Sorry that it took a while to revise because I was busy.

PR #1212, opened 2015-11-26
Replace None checks in skipgram right padding

alvations:
From #1070 and #1179, the `None` checks for the end of the inner skipgram loop is replaced with a sentinel object.


stevenbird:
Thanks @alvations

Issue #1211, opened 2015-11-25
Stanford parser unicodeDecodeError on phone numbers


Newtonvolt:
Python 3.4.3 windows 7
The stanford parser (3.5.2) will replace spaces with 0xa0 (which I guess is unicode non-breaking space), in the event of a phone number appearing. The stanford nltk module will not handle this well, and will crash with a UnicodeDecodeError.

To reproduce:

``` python
from nltk.parse import stanford
import os
os.environ['STANFORD_PARSER']='C:\stanford-parser\stanford-parser-full-2015-04-20'
os.environ['STANFORD_MODELS']='C:\stanford-parser\stanford-parser-full-2015-04-20'
stanford_parser = stanford.StanfordParser(model_path='C:\stanford-parser\stanford-parser-full-2015-04-20/englishPCFG.ser.gz')
x = stanford_parser.raw_parse('This is going to crash: 01 111 555.') 
```

result:

```
Traceback (most recent call last):
  File "C:/stanford-parser/crash.py", line 6, in <module>
    x = stanford_parser.raw_parse('This is going to crash: 01 111 555.')
  File "C:\Python34\lib\site-packages\nltk\parse\stanford.py", line 128, in raw_parse
    return next(self.raw_parse_sents([sentence], verbose))
  File "C:\Python34\lib\site-packages\nltk\parse\stanford.py", line 146, in raw_parse_sents
    return self._parse_trees_output(self._execute(cmd, '\n'.join(sentences), verbose))
  File "C:\Python34\lib\site-packages\nltk\parse\stanford.py", line 214, in _execute
    stdout = stdout.decode(encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 167: invalid start byte
```

stevenbird:
Thanks @Newtonvolt

longdt219:
Hi @Newtonvolt  and @stevenbird, 
This bug is caused by non-break char inserted between numbers. I added the code to remove this char. Tested on windows.

alvations:
[...]

PR #1213, opened 2015-11-30
Fix bug #1211 Stanford parser non break pause

longdt219:
Remove the non-break pause which is added by new stanford parser


stevenbird:
Thanks @longdt219

Issue #1116, opened 2015-09-07
specifying the type of dictionary values


stevenbird:
In sphinx documentation, specify not just the key type but also the value type, either as `dict(key_type): value_type` or as `dict(key_type -> value_type)`

PR #1215, opened 2015-12-02
Feature/update sphinx typing

mrecachinas:
Changed several sphinx `:rtype: dict` to be more specific as per #1116. Also, changed `list` to `list(str)`, `tuple` to `tuple(float)` and such, where applicable.


stevenbird:
Thanks @mrecachinas – very welcome contribution!

stevenbird:
@mrecachinas – we would welcome more of this housekeeping if you have time!

PR #1220, opened 2015-12-04
Update the default download URL to use https://

reedloden:
The index file that the downloader uses is currently served from GitHub
Pages (via http://www.nltk.org), but while GitHub Pages doesn't directly
offer https://, they provide an alternate way to easily download files
securely over https:// via their `raw.githubusercontent.com` FQDN.

This helps protect the index and the file URLs listed in it from being
MITM'd and modified to something malicious and just helps with ensuring
file integrity.


reedloden:
The other half of this is https://github.com/nltk/nltk_data/pull/40.

stevenbird:
Great idea, thanks @reedloden.

PR #1222, opened 2015-12-07
Use nltk.data.load to load model (to take advantage of caching)

rmalouf:
This PR modifies the perceptron tagger so that it uses nltk.data.load rather than pickle.load to load stored models.  This has a few advantages, but the biggest is that nltk.data.load automatically caches loaded data files, which means that the tagger only gets loaded once.  After the first use of nltk.pos_tag, subsequent calls are much faster.


stevenbird:
Oh great, thanks @rmalouf

Issue #1130, opened 2015-09-12
Adopt abstract base class syntax


stevenbird:
NLTK has plenty of abstract base classes. Now that Python has support for these, we need to audit the codebase and adopt this syntax.

cf https://docs.python.org/3.0/whatsnew/2.6.html#pep-3119

stevenbird:
Thanks @mrecachinas and @kmike for https://github.com/nltk/nltk/pull/1223. It looks like we have a model that can be applied across the toolkit now.

souravsingh:
@stevenbird Is the issue fixed?

gmotzespina:
@stevenbird it seems that this issue has been fixed, it might be useful to close it.

stevenbird:
Thanks @souravsingh, @gmotzespina!

PR #1223, opened 2015-12-08
Add abstract base class syntax to tokenize/api.py

mrecachinas:
As per #1130.

### DIFF
--- a/nltk/tokenize/api.py
+++ b/nltk/tokenize/api.py
@@ -11,13 +11,13 @@ Tokenizer Interface
 """
 
 from abc import ABCMeta, abstractmethod
+from six import add_metaclass
 
 from nltk.internals import overridden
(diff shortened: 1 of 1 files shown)


kmike:
This is Python 2 - only syntax for abc, in Python 3 it does nothing. Check six.with_metaclass or six.add_metaclass (https://pythonhosted.org/six/#six.with_metaclass)

mrecachinas:
@kmike Ah good catch! Fixing this...

stevenbird:
Thanks @mrecachinas, @kmike

mrecachinas:
@stevenbird Ready for merge or should I add any other changes?

stevenbird:
Thanks @mrecachinas. Sorry for the delay. Looks like we have a model that could be applied across the toolkit now.

PR #1224, opened 2015-12-08
Fix a bug on sentiwordnet

jihunchoi:
In the current implementation, sentiment synsets of some adjectives
are not found, due to the fact that in SentiWordNet corpus
there exist no 's' tag, which stands for 'adjective satellite.'
This modifies the code to treat 's' tag as 'a', so that it can find
synsets of adjectives correctly.
To see the change, check synsets of the word 'crazy.'


stevenbird:
Thanks @jihunchoi

PR #1225, opened 2015-12-08
Add missing import in Brill Tagger Trainer documentation.

mrdrozdov:


stevenbird:
Thanks for picking this up @mrdrozdov

PR #1229, opened 2015-12-09
Added function to calculate Corpus-level BLEU and RIBES

alvations:
In the process of adding functions to calculate corpus-level BLEU and RIBES, the following changes were made:
- **Set the default uniform weights for 4grams when calculating BLEU scores.**
  - The original paper also set default weights and limits to 4grams, thus the default weights were at at `[0.25] * 4`.
- **Added function for corpus-level BLEU by calculating mirco-average precision**
  - The original BLEU paper (Papineni et al. 2002) was meant to calculate micro-average corpus-level score. So the modified precision and brevity penalty scores does return a float but the sum of numerators and denominators. Simply averaging sentence level BLEU (i.e. marco-average) will not be the same as the BLEU as intended in the paper. 
- **Change `_modified_precision()` to return numerator and denominator instead of a single-sentence BLEU score, i.e. `corpus_bleu()`**
  - `sentence_modified_precision()` will return the floating point BLEU score for a single sentence.
  - Also, by disentangling the float return from the precision and penalty functions, smoothing on the precision can be easily extended, now that `_modified_precision()` and `_brevity_penalty()` returns numerators and denominators. E.g. when someone wants  to implement BLEU smoothing from http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf. 
- **Change the `nltk.translate` namespaces**
  - To preserve current user's code. the `tranlsate.__init__.py` imports the `sentence_bleu` as the default `nltk.translate.bleu`. So from the user's end, BLEU still works the same way and calculating corpus-level bleu will require importing `from nltk.translate.bleu_score import corpus_bleu`. 
- **Added function for corpus-level RIBES score, i.e. `corpus_ribes()`**
  - Different from BLEU's micro-average precision, RIBES calculates the macro-average precision by averaging the best RIBES score for each pair of hypothesis and its corresponding references. So we can safely sum and divide by `len()`. See line 307 from RIBES v1.03.1 (http://www.kecl.ntt.co.jp/icl/lirg/ribes/)

### DIFF
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -66,6 +66,8 @@ def sentence_bleu(references, hypothesis, weights=[0.25, 0.25, 0.25, 0.25]):
     :type hypothesis: list(str)
     :param weights: weights for unigrams, bigrams, trigrams and so on
     :type weights: list(float)
+    :return: The sentence-level BLEU score.
+    :rtype: float
     """
     # Calculates the modified precision *p_n* for each order of ngram.
     p_ns = [] 
@@ -137,6 +139,8 @@ def corpus_bleu(list_of_references, hypotheses, weights=[0.25, 0.25, 0.25, 0.25]
     :type hypotheses: list(list(str))
     :param weights: weights for unigrams, bigrams, trigrams and so on
     :type weights: list(float)
+    :return: The corpus-level BLEU score.
+    :rtype: float
     """
     p_numerators = Counter() # Key = ngram order, and value = no. of ngram matches.
     p_denominators = Counter() # Key = ngram order, and value = no. of ngram in ref.
@@ -253,6 +257,8 @@ def _modified_precision(references, hypothesis, n):
     :type hypothesis: list(str)
     :param n: The ngram order.
     :type n: int
+    :return: BLEU's modified precision for the nth order ngram.
+    :rtype: Fraction
     """
     counts = Counter(ngrams(hypothesis, n))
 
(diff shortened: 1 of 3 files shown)


alvations:
@hoontw Thanks for the review! The `Fraction` return type suggestion is great! That simplifies all the silly duck functions I was creating. The revised version of the code looks much cleaner now.

hoontw:
Looks much better than before! Just a few more minor tweaks and it is good to go.

alvations:
@hoontw sorry for the late commits, I was moving around and at last I've reached where I'm supposed to be =)

hoontw:
Thanks @alvations for the changes. It's great to have corpus-level metrics for MT.

alvations:
Now we can do some tuning using the corpus-level metrics, possibly something like https://github.com/alvations/mosesdecoder/blob/master/scripts/training/propy/pro.py.

nltk/translate/bleu_score.py
```diff
@@ -11,21 +11,19 @@
...
+from fractions import Fraction
+from collections import Counter
 
-from nltk.tokenize import word_tokenize
-from nltk.compat import Counter
 from nltk.util import ngrams
 
 
-def bleu(references, hypothesis, weights):
+def sentence_bleu(references, hypothesis, weights=[0.25, 0.25, 0.25, 0.25]):
```
dimazest, line 20: better set weights to a tuple, not a list, to avoid mutable default parameters.

nltk/translate/bleu_score.py
```diff
@@ -47,41 +45,148 @@ def bleu(references, hypothesis, weights):
...
     :type weights: list(float)
+    :return: The sentence-level BLEU score.
+    :rtype: float
     """
-    p_ns = (
-        _modified_precision(references, hypothesis, i)
-        for i, _ in enumerate(weights, start=1)
-    )
+    # Calculates the modified precision *p_n* for each order of ngram.
+    p_ns = []
```
dimazest, line 73: list comprehension would look much better here.

alvations:
@dimazest Thanks for the suggestions! Changes made in #1238

PR #1231, opened 2015-12-13
Small fixes for FreqDist and ConditionalFreqDist

rmalouf:
Fixes for some minor annoyances with FreqDists that came up in class this semester:
- Adjust tabulate column widths so that samples and counts line up.
- Add mathematical operators to FreqDist.  They're inherited from Counter and, since that is a built-in type, they returned Counter objects rather than FreqDists.  That's now fixed.
-  Add mathematical operators to ConditionalFreqDist.

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -292,6 +292,44 @@ class FreqDist(Counter):
         """
         return self.__class__(self)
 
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @rmalouf – these contributions are great

PR #1233, opened 2015-12-17
Feature/casual tokenize fix.1158

clarkgrubb:
Avoid text which causes nltk.tokenize.casual.casual_tokenize() to hang by shortening long sequences of punctuation (actually, all characters matching [^a-zA-Z0-9]) to a sequence of just three of the same character.

```
>>> import nltk
>>> nltk.tokenize.casual.casual_tokenize("Photo: Aujourd'hui sur http://t.co/0gebOFDUzn Projet... http://t.co/bKfIUbydz2.............................. http://fb.me/3b6uXpz0L")
[u'Photo', u':', u"Aujourd'hui", u'sur', u'http://t.co/0gebOFDUzn', u'Projet', u'...', u'http://t.co/bKfIUbydz2', u'...', u'http://fb.me/3b6uXpz0L']
```

On a sample of 246M tweets, tweets which cause casual_tokenize() to hang were about a 1 in 5M occurrence.  After this change, I was able to process 246M tweets successfully.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -61,6 +61,7 @@
 - Dougal Graham
 - Brent Gray
 - Simon Greenhill
+- Clark Grubb
 - Eduardo Pereira Habkost
 - Masato Hagiwara
 - Lauri Hallila
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -160,7 +160,7 @@ The `preserve_case` parameter (default: True) allows to convert uppercase tokens
 
 It should not hang on long sequences of the same punctuation character.
 
-    >>> tknzer = TweetTokenizer()
+    >>> tknzr = TweetTokenizer()
     >>> s10 = "Photo: Aujourd'hui sur http://t.co/0gebOFDUzn Projet... http://t.co/bKfIUbydz2.............................. http://fb.me/3b6uXpz0L"
     >>> tknzr.tokenize(s10)
-    [u'photo', u':', u"aujourd'hui", u'sur', u'http://t.co/0gebOFDUzn', u'projet', u'...', u'http://t.co/bKfIUbydz2', u'...', u'http://fb.me/3b6uXpz0L']
+    [u'Photo', u':', u"Aujourd'hui", u'sur', u'http://t.co/0gebOFDUzn', u'Projet', u'...', u'http://t.co/bKfIUbydz2', u'...', u'http://fb.me/3b6uXpz0L']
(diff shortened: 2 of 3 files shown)


ewan-klein:
@fievelk If you have time, could you review this PR?

fievelk:
Sure!

fievelk:
Thank you @clarkgrubb.
- There is a typo that should be fixed in `tokenize.doctest`:
  https://github.com/clarkgrubb/nltk/commit/098d4a661d798a393c153b324efa371e98a09f36#diff-c3036c642dd8a7591f4780613ec8bdf8R163

---
- This fix should work in practice. However, it should be noted that repeated sequences of characters did not always cause problems; before @clarkgrubb's PR, in fact, we had:
  
  ```
  >>> tknzr = TweetTokenizer()
  # This will work properly
  >>> s5 = "http://t.co/bKfIUbydz2 .............................. http://fb.me/3b6uXpz0L"
  >>> tknzr.tokenize(s5)
  ['http://t.co/bKfIUbydz2', '..............................', 'http://fb.me/3b6uXpz0L']
  
  # This will hang
  >>> s6 = "http://t.co/bKfIUbydz2.............................. http://fb.me/3b6uXpz0L"
  >>> tknzr.tokenize(s6)
  ```
  
  Moreover, the underscore character `_` did not seem to cause any problem, at least testing with the following:
  
  ```
  >>> s7 = "http://t.co/bKfIUbydz2___________________________________ http://fb.me/3b6uXpz0L"
  >>> tknzr.tokenize(s7)
  ['http://t.co/bKfIUbydz2___________________________________', 'http://fb.me/3b6uXpz0L']
  ```

The provided fix seems to work at the moment by the way. In any case, it would be nice to have a better understanding of the problem.

clarkgrubb:
The Python regex engine uses backtracking, so there is a known risk of this kind of problem.  Russ Cox discussed the problem here: https://swtch.com/~rsc/regexp/regexp1.html

I did some experiments, and none of the subcomponents of the WORD_RE regular expression hang on the bad example I used above.  It is only when we stitch all of the subcomponents together inside "( )|( )|...|()".  That suggests that you could ditch the monster regular expression and instead write a little loop which tests each subcomponent in turn to pull off the next token.

fievelk:
Thank you @clarkgrubb, this is interesting.
In my opinion we can keep your current fix at the moment, and eventually improve it in the future.

stevenbird:
Thanks @clarkgrubb and @fievelk

PR #1238, opened 2015-12-22
Improvement to BLEU scores

alvations:
Spillover improvements from #1229 
- Use tuple instead of list in `sentence_bleu` default parameter
- Use list comprehension to get modified precision for n orders of ngrams.

### DIFF
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -81,6 +81,7 @@ def sentence_bleu(references, hypothesis, weights=(0.25, 0.25, 0.25, 0.25)):
             s.append(w * math.log(p_n))
         except ValueError: # some p_ns is 0
             s.append(0)
+    s = math.fsum(s)
 
     # Calculates the brevity penalty.
     # *hyp_len* is referred to as *c* in Papineni et. al. (2002)
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -161,7 +161,7 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     bp = _brevity_penalty(ref_lengths, hyp_lengths)
     
     # Calculate sum of corpus-level modified precisions.
-    s = (0 if p_numerators[i] == 0 else
+    s = (0 if (p_numerators[i] | p_denominators[i]) == 0 else
          (w* math.log(p_numerators[i] / p_denominators[i]))
          for i, w in enumerate(weights, start=1))
         
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -70,17 +70,14 @@ def sentence_bleu(references, hypothesis, weights=(0.25, 0.25, 0.25, 0.25)):
     :rtype: float
     """
     # Calculates the modified precision *p_n* for each order of ngram.
-    p_ns = [float(_modified_precision(references, hypothesis, i))
+    p_n = [float(_modified_precision(references, hypothesis, i))
             for i, _ in enumerate(weights, start=1)]
 
     # Calculates the overall modified precision for all ngrams.
     # By sum of the product of the weights and the respective *p_n*
     s = []
-    for w, p_n in zip(weights, p_ns):
-        try:
-            s.append(w * math.log(p_n))
-        except ValueError: # some p_ns is 0
-            s.append(0)
+    for w, p_i in zip(weights, p_n):
+        p_i = 0 if p_i == 0 else w * math.log(p_i)
     s = math.fsum(s)
 
     # Calculates the brevity penalty.
@@ -169,11 +166,9 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     # Calculate corpus-level modified precision.
     p_n = []
     for i, w in enumerate(weights, start=1):
-        pn = p_numerators[i] / p_denominators[i]
-        try:
-            p_n.append(w* math.log(pn))
-        except ValueError:
-            p_n.append(0)
+        p_i = p_numerators[i] / p_denominators[i]
+        p_i = 0 if p_i == 0 else w * math.log(p_i) 
+        p_n.append(p_i)
         
     return bp * math.exp(math.fsum(p_n))
 
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -74,7 +74,7 @@ def sentence_bleu(references, hypothesis, weights=(0.25, 0.25, 0.25, 0.25)):
 
     # Calculates the overall modified precision for all ngrams.
     # By sum of the product of the weights and the respective *p_n*
-    s = (0 if p_i == 0 else (w * math.log(p_i)) 
+    s = (w * math.log(p_i) if p_i else 0 
          for w, p_i in zip(weights, p_n))
     
     # Calculates the brevity penalty.
@@ -161,8 +161,8 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     bp = _brevity_penalty(ref_lengths, hyp_lengths)
     
     # Calculate sum of corpus-level modified precisions.
-    s = (0 if (p_numerators[i] | p_denominators[i]) == 0 else
-         (w * math.log(p_numerators[i] / p_denominators[i]))
+    s = (0 if not(p_numerators[i] and p_denominators[i]) else
+         w * math.log(p_numerators[i] / p_denominators[i])
          for i, w in enumerate(weights, start=1))
         
     return bp * math.exp(math.fsum(s))
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -7,7 +7,6 @@
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 """BLEU score implementation."""
-
 from __future__ import division
 
 import math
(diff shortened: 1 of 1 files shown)


dimazest:
great and looks good!

alvations:
A few question on implementation choices:

 - Would a this list comprehension `math.fsum(w * math.log(p_n) if p_n != 0 else 0 for w, p_n in zip(weights, p_ns))` be better than https://github.com/alvations/nltk/blob/develop/nltk/translate/bleu_score.py#L169 ?

 - Would this one-liner better than 
`max_counts = reduce(or_, [Counter(ngrams(ref, n)) for ref in references])`  be better than this: https://github.com/alvations/nltk/blob/develop/nltk/translate/bleu_score.py#L268? (Additional imports are needed though: `from functools import reduce; from operator import or_`)

alvations:
@hoontw, it's actually rather strange. I [rewrote the BLEU scores for another project](https://gist.github.com/alvations/838cb021712ad66e7768) and I realized that the original implementation can be simplified quite a lot. But it will sacrificed readability. 

@everyone Should I rewrite it as how I did in the other project or is it better to leave it as it is now.

hoontw:
@alvations Is there an advantage in rewriting? Simplicity is good if it makes the algorithm more readable. If the simplification makes the code more terse and tricky, it is more debatable. Both implementations look comparable to me.

alvations:
@hoontw I rewrote BLEU because I wanted a version suitable for MT tuning and a standalone script. I find I had used too many tricks to go for speed and "standalone-ness". Possibly the only useful things that's different in the re-written version are:
- how `max_count` for `modified_precision` is computed; using [counter set theory](https://gist.github.com/alvations/838cb021712ad66e7768#file-bleu-py-L56) vs [normal loops](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L268)
- how [`sentence_bleu` is an instance of `corpus_bleu`](https://gist.github.com/alvations/838cb021712ad66e7768#file-bleu-py-L205); in NLTK, it's [a separate implementation](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L20) because the sentence bleu was written before corpus bleu (which is rather strange because the original Papineni et al (2000) was meant for corpus bleu)
- Chen and Cherry (2014) [smoothing techniques for `sentence_bleu`](https://gist.github.com/alvations/838cb021712ad66e7768#file-bleu-py-L119). This we can easily add to NLTK's version.

I'm not sure whether there's an advantage in my rewritten (speed wise, it would be the same if I materialize the `list_of_references` and `hypotheses`. In the rewritten version, it depends heavily on generators from file and `zip` in python3, in python2 it would be the same speed as NLTK version of the BLEU.

nltk/translate/bleu_score.py
```diff
@@ -70,18 +70,15 @@ def sentence_bleu(references, hypothesis, weights=[0.25, 0.25, 0.25, 0.25]):
...
-        s = math.fsum(w * math.log(p_n) for w, p_n in zip(weights, p_ns))
-    except ValueError:
-        # some p_ns is 0
-        return 0
+    p_n = [float(_modified_precision(references, hypothesis, i))
+            for i, _ in enumerate(weights, start=1)]
+
+    # Calculates the overall modified precision for all ngrams.
+    # By sum of the product of the weights and the respective *p_n*
+    s = []
```
dimazest: you can use list comprehension here as well.

(i'm sorry to be picky ;)

nltk/translate/bleu_score.py
```diff
@@ -70,18 +70,15 @@ def sentence_bleu(references, hypothesis, weights=[0.25, 0.25, 0.25, 0.25]):
...
-        p_ns.append(p_n) 
-
-    try:
-        # Calculates the overall modified precision for all ngrams.
-        # By taking the product of the weights and the respective *p_n*
-        s = math.fsum(w * math.log(p_n) for w, p_n in zip(weights, p_ns))
-    except ValueError:
-        # some p_ns is 0
-        return 0
+    p_n = [float(_modified_precision(references, hypothesis, i))
```
dimazest: maybe, for speed reasons it's better to create a generator here, not a list:

``` python
p_n = (float(...) for i, _ in ... )
```

nltk/translate/bleu_score.py
```diff
@@ -166,13 +163,14 @@ def corpus_bleu(list_of_references, hypotheses, weights=[0.25, 0.25, 0.25, 0.25]
     # Calculate corpus-level brevity penalty.
     bp = _brevity_penalty(ref_lengths, hyp_lengths)
     
-    # Calculate corpus-level modified precision.
-    p_n = []
+    # Calculate sum of corpus-level modified precisions.
+    s = []
     for i, w in enumerate(weights, start=1):
```
dimazest: here also use a generator.

nltk/translate/bleu_score.py
```diff
@@ -166,13 +163,14 @@ def corpus_bleu(list_of_references, hypotheses, weights=[0.25, 0.25, 0.25, 0.25]
     # Calculate corpus-level brevity penalty.
     bp = _brevity_penalty(ref_lengths, hyp_lengths)
     
-    # Calculate corpus-level modified precision.
-    p_n = []
+    # Calculate sum of corpus-level modified precisions.
+    s = []
```
dimazest: here also use a generator

alvations:
@dimazest thanks for the suggestion! I've changed them to generators.

alvations:
@stevenbird @hoontw @dimazest Could one of you help to review the code and merge it if there's no new modification that needs to be made? After which we can close also #1241 and #1243, and we can also move on to port [`chen and cherry smoothing`](https://gist.github.com/alvations/838cb021712ad66e7768#file-bleu-py-L119) for BLEU. Thanks in advance!

nltk/translate/bleu_score.py
```diff
@@ -49,14 +48,14 @@ def sentence_bleu(references, hypothesis, weights=[0.25, 0.25, 0.25, 0.25]):
...
-    0
+    0.39692877231857493
 
     The default BLEU calculates a score for up to 4grams using uniform
     weights. To evaluate your translations with higher/lower order ngrams, 
     use customized weights. E.g. when accounting for up to 6grams with uniform
     weights:
 
-    >>> weights = [0.1666, 0.1666, 0.1666, 0.1666, 0.1666]
+    >>> weights = (0.1666, 0.1666, 0.1666, 0.1666, 0.1666)
```
dimazest, line 58: strictly speaking you don't need brackets here.

nltk/translate/bleu_score.py
```diff
@@ -166,13 +160,12 @@ def corpus_bleu(list_of_references, hypotheses, weights=[0.25, 0.25, 0.25, 0.25]
     # Calculate corpus-level brevity penalty.
     bp = _brevity_penalty(ref_lengths, hyp_lengths)
     
-    # Calculate corpus-level modified precision.
-    p_n = []
-    for i, w in enumerate(weights, start=1):
-        pn = p_numerators[i] / p_denominators[i]
-        p_n.append(w* math.log(pn))
+    # Calculate sum of corpus-level modified precisions.
+    s = (0 if (p_numerators[i] | p_denominators[i]) == 0 else
```
dimazest: what does `(p_numerators[i] | p_denominators[i]) == 0` mean? does it check whether `p_numerators[i]` or `p_denominators[i]` is 0?

if yes, i would rewrite to something like `0 if not(p_numerators[i] and p_denominators[i]) else ...`

nltk/translate/bleu_score.py
```diff
@@ -166,13 +160,12 @@ def corpus_bleu(list_of_references, hypotheses, weights=[0.25, 0.25, 0.25, 0.25]
...
+         (w* math.log(p_numerators[i] / p_denominators[i]))
```
dimazest: `w * math.log(...)` a missing space before math.

nltk/translate/bleu_score.py
```diff
@@ -70,29 +69,24 @@ def sentence_bleu(references, hypothesis, weights=[0.25, 0.25, 0.25, 0.25]):
...
-        s = math.fsum(w * math.log(p_n) for w, p_n in zip(weights, p_ns))
-    except ValueError:
-        # some p_ns is 0
-        return 0
+    p_n = (float(_modified_precision(references, hypothesis, i))
+            for i, _ in enumerate(weights, start=1))
 
+    # Calculates the overall modified precision for all ngrams.
+    # By sum of the product of the weights and the respective *p_n*
+    s = (0 if p_i == 0 else (w * math.log(p_i))
```
dimazest: if  `p_i` is integer, then there is no need to compare with 0, I would write:

``` python
s = (w * math.log(p_i)) if p_i else 0
```

dimazest:
Please see my comments, looks good.

nltk/translate/bleu_score.py
stevenbird, line 58: It appears that [PEP-8](https://www.python.org/dev/peps/pep-0008/) is silent about this. My own preference is to retain the parens, but I don't feel strongly about it.

nltk/translate/bleu_score.py
stevenbird: +1

alvations: `(p_numerators[i] | p_denominators[i]) == 0` checks whether numerator or denominator is zero. In the case of numerator zero, it will raise `math.log(0)` math domain error and when denominator is 0, it raises `ZeroDivisionError`.

I'm not sure whether `0 if not variable else ...` is a good thing, that will allow `None` or "nully` types to pass through, e.g. 

```
>>> x, y = None, []
>>> 0 if not(x and y) else x / y
0
>>> x, y = None, 0
>>> 0 if not(x and y) else x / y
0
>>> x, y = 0, 1
>>> 0 if not(x and y) else x / y
0
>>> x, y = 2, ""
>>> 0 if not(x and y) else x / y
0
>>> x, y = 2, 1
>>> 0 if not(x and y) else x / y
2
```

Given the different types, python's will raise error and it might be easier to debug the code in the further when extending the BLEU functions:

```
>>> x, y = 2, ""
>>> 0 if (x|y) == 0 else x / y
TypeError: unsupported operand type(s) for |: 'int' and 'str'
```

But still 0 is problematic because it lets boolean type slip through:

```
>>> x,y = False, False
>>> 0 if (x|y) == 0 else x / y
0
```

nltk/translate/bleu_score.py
alvations: As above, I still think allowing nully types to pass through might be problematic when the function is extended and numbers get overlooked. It's easy to skip through tests when the function returns floating points and if they're zero, without a closer look, one might just say it's okay.

nltk/translate/bleu_score.py
alvations: Turns out in any case, the checks of zero, it's heavily dependent on the return type of `_modified_precision` ;P 

I'll change it to the `not variable` checks.

dimazest:
`(x|y) == 0` looks cryptic to me as I have to think what `|` does. However, it does look cool!

It's ready to be merged.

dimazest:
I've pressed a wrong button!

It looks good and is ready to be merged!

stevenbird:
Thanks @alvations for the contribution, and @dimazest, @hoontw for review.

PR #1243, opened 2016-01-02
Fixed bug in summation of modified precision in BLEU score

zephyrzilla:
Fixed the bug in summation of modified precision in BLEU score, which previously showed `ValueError` in case of log(0). BLEU score  equals zero if any one of the underlying score is zero. The `try` and `except` block handles this.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -189,3 +189,4 @@
 - Sergio Oller
 - Will Monroe
 - Elijah Rippeth
+- Surajit Dasgupta <https://github.com/surajit-techie>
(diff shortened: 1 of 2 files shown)


alvations:
@surajit-techie  Thanks! but actually, it's already fixed in #1238 with https://github.com/alvations/nltk/commit/3c5cf4a1cf192dbfc58ad510795ca4c9511441c5 and https://github.com/alvations/nltk/commit/96ddfbd09bf5401f2e8b8c08f0033d387ad7a99c ;P

stevenbird:
Thanks @surajit-techie. Resolved by #1238

PR #1226, opened 2015-12-08
Use `os.path.join()` to concatenate paths in compat.py

mrecachinas:
This offers a fix for #783.

Issue #783, opened 2014-10-31
paths in compat.py


stevenbird:
Use `os.path.join()` to concatenate paths, in `compat.py`, cf https://github.com/nltk/nltk/pull/777

Dobatymo:
was fixed here https://github.com/nltk/nltk/pull/1226 wasn't it?

alvations:
Yes, it resolved in #1226. Thanks for catching this! 

And also thank you for spotting the other lingering issues on the tracker that should have been resolved =)

stevenbird:
Yes, thanks @Dobatymo

PR #1245, opened 2016-01-04
Prevent "\t" being interpreted as a tab in compat.py

philiptzou:


kmike:
:+1:

dimazest:
maybe it worth using os.path.join https://docs.python.org/2/library/os.path.html#os.path.join to get rid of the if statement

mrecachinas:
@dimazest -- the `os.path.join` is addressed in patch #1226 as a fix for #783 :)

philiptzou:
Perhaps we can just use `os.path.sep` which is much concise to me.

mrecachinas:
@philiptzou -- As @dimazest suggested, it's a bit more concise and considered standard practice to use `os.path.join` to ensure cross-os support. Incorporating `os.path.join` would look like

``` python
ne_chunker = os.path.join("chunkers", "maxent_ne_chunker")
help_tagsets = os.path.join("help", "tagsets")
treebank_tagger = os.path.join("taggers", "maxent_treebank_pos_tagger")
punkt_tok = os.path.join("tokenizers", "punkt")
```

This removes the need for checking the platform or `/` direction (i.e., `if os.path.sep != '/':` or `if sys.platform.startswith('win'):`). This also removes the need for making each path a raw-string because `os.path.join` will properly escape the string.

erip:
:+1:

philiptzou:
Actually the if os.path.sep != '/': is just an optimization judgement.
Removing that line (and fix the indent of following lines) won't affect the
result. :)

Michael Recachinas <notifications@github.com>于2016年1月6日 周三19:53写道：

philiptzou:
@mrecachinas Actually the `if os.path.sep != '/':` is just an optimization judgement.
Removing that line (and fix the indent of following lines) won't affect the
result. :)

The thing I don't like in #1226's approachment, is the four new useless variables and four new method calls to `os.path.join`. This is more clean and pythonic:

``` python
_PY3_DATA_UPDATES = [('chukers', 'maxent_ne_chunker'),
                     ('help', 'tagsets'),
                     ('taggers', 'maxent_treebank_pos_tagger'),
                     ('tokenizers', 'punkt')]
_PY3_DATA_UPDATES = [os.path.join(*args) for args in _PY3_DATA_UPDATES]
```

But I still prefer the replacing way since it is faster than concating strings.

mrecachinas:
@philiptzou --

> Actually the `if os.path.sep != '/':` is just an optimization judgement. Removing that line (and fix the indent of following lines) won't affect the result.

Understood. I'm only suggesting that the more idiomatic solution is to use `os.path.join` rather than replace all instances of `/` with `os.path.sep` (see http://stackoverflow.com/questions/16789714/differences-between-use-of-os-path-join-and-os-sep-concatenation).

> But I still prefer the replacing way since it is faster than concating strings.

I agree with the assessment that `replace` is faster than `os.path.join`. To compare the times to see exactly how much faster `replace` is, I wrote a quick script to time both:

``` python
from __future__ import division
import os.path
import time

num_runs = 1000000
t1 = 0
t2 = 0
join_tot = 0
for i in xrange(num_runs):
    t1 = time.time()

    DATA_UPDATES = [("chunkers", "maxent_ne_chunker"),
                    ("help", "tagsets"),
                    ("taggers", "maxent_treebank_pos_tagger"),
                    ("tokenizers", "punkt")]

    _PY3_DATA_UPDATES = [os.path.join(*path_list) for path_list in DATA_UPDATES]

    t2 = time.time()
    join_tot += (t2 - t1)

replace_tot = 0
for i in xrange(num_runs):
    t1 = time.time()

    _PY3_DATA_UPDATES = ["chunkers/maxent_ne_chunker",
                         "help/tagsets",
                         "taggers/maxent_treebank_pos_tagger",
                         "tokenizers/punkt"]
    if os.path.sep != '/':
        _PY3_DATA_UPDATES = [s.replace('/', os.path.sep)
                             for s in _PY3_DATA_UPDATES]

    t2 = time.time()
    replace_tot += (t2 - t1)

join_avg = join_tot / num_runs
replace_avg = replace_tot / num_runs

print "{0} {1}".format(join_avg, replace_avg)
```

The output I got was `4.57689023018e-06` for `os.path.join` and `3.92367839813e-07` for `replace`. This means `replace` is about 91% faster than `os.path.join` in the case that `os.path.sep` is `/`. For the case where `os.path.sep` is `\`, I measured `4.4388897419e-06` for `os.path.join` and `2.15146183968e-06` for `replace`, which is only a 50% speedup. In a performance critical module, a 50% or 90% speedup would be unbelievably nontrivial; however, in this case, the runtimes are on the order of microseconds and hundreds of nanoseconds (both inconsequential magnitudes of time). Therefore, for the purposes of this module (i.e., a non-time-critical compatibility module), using `replace` for speed is a mere micro-optimization that ultimately reduces portability and is less Pythonic (and IMO less readable).

> The thing I don't like in #1226's approachment, is the four new useless variables and four new method calls to os.path.join. This is more clean and pythonic: ...

Good point! I agree this is more Pythonic than assigning several variables. I've edited #1226 to mirror your suggestion. :smile:

stevenbird:
I strongly agree that readable, Pythonic, maintainable code is the preference.

Issue #397, opened 2013-05-02
PunktTrainer fails with ZeroDivisionError on some text.


bmurr:
Running the following code will result in a ZeroDivisionError:

```
from nltk.tokenize.punkt import *
trainer = PunktTrainer()
trainer.train('1.\n2.\n3.')
```

Traceback:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ben/.virtualenvs/nltk/lib/python2.7/site-packages/nltk/tokenize/punkt.py", line 710, in train
    self.finalize_training(verbose)
  File "/Users/ben/.virtualenvs/nltk/lib/python2.7/site-packages/nltk/tokenize/punkt.py", line 797, in finalize_training
    for (typ1, typ2), ll in self._find_collocations():
  File "/Users/ben/.virtualenvs/nltk/lib/python2.7/site-packages/nltk/tokenize/punkt.py", line 1118, in _find_collocations
    col_count, self._type_fdist.N())
  File "/Users/ben/.virtualenvs/nltk/lib/python2.7/site-packages/nltk/tokenize/punkt.py", line 1058, in _col_log_likelihood
    p2 = 1.0 * (count_b - count_ab) / (N - count_a)
ZeroDivisionError: float division by zero
```

I've also run into a `ValueError` in _col_log_likelihood, when it attempted to do `math.log(0)`.

alvations:
@mrecachinas @bmurr is this fixed by #1247 ?

mrecachinas:
@alvations Correct. There should be no way to get a ZeroDivisionError or ValueError (for `math.log(0)`) with #1247.

PR #1247, opened 2016-01-08
Fix zero division & log(0) in PunktTrainer._col_log_likelihood

mrecachinas:
This commit proposes a possible fix for #397.

I followed Dunning's remarks in http://tdunning.blogspot.com/2008/03/surprise-and-coincidence.html to avoid `log(0)`. This includes converting
- `log(p)` to `log(p + (p == 0))`
- `log(1.0 - p)` to `log(1.0 - p + (p == 1))`

Also, for the original error:

```
p2 = (count_b - count_ab) / (N - count_a)
```

I capped it to 1 as in

```
p2_denom = N - count_a
if p2_denom == 0:
    p2 = 1
else:
    p2 = (count_b - count_ab) / p2_denom
```

Let me know what you think, and if this is a valid solution, I will readjust the test cases and add a new one to cover `p2_denom == 0`.

### DIFF
--- a/nltk/tokenize/punkt.py
+++ b/nltk/tokenize/punkt.py
@@ -1081,13 +1081,13 @@ class PunktTrainer(PunktBaseClass):
         except ValueError as e:
             summand2 = 0
 
-        if count_a == count_ab or p1 == 0 or p1 == 1:
+        if count_a == count_ab or p1 <= 0 or p1 >= 1:
             summand3 = 0
         else:
             summand3 = (count_ab * math.log(p1) +
                         (count_a - count_ab) * math.log(1.0 - p1))
 
-        if count_b == count_ab or p2 == 0 or p2 == 1:
+        if count_b == count_ab or p2 <= 0 or p2 >= 1:
             summand4 = 0
         else:
             summand4 = ((count_b - count_ab) * math.log(p2) +
--- a/nltk/tokenize/punkt.py
+++ b/nltk/tokenize/punkt.py
@@ -1080,7 +1080,7 @@ class PunktTrainer(PunktBaseClass):
         if count_a == count_ab:
             summand3 = 0
         else:
-            summand3 = (count_ab * math.log(p1 or 0) +
+            summand3 = (count_ab * math.log(p1 or 1) +
                         (count_a - count_ab) * math.log(1.0 - p1) if p1 < 1 else 0)
 
         if count_b == count_ab:
--- a/nltk/tokenize/punkt.py
+++ b/nltk/tokenize/punkt.py
@@ -1081,17 +1081,17 @@ class PunktTrainer(PunktBaseClass):
         except ValueError as e:
             summand2 = 0
 
-        if count_a == count_ab:
+        if count_a == count_ab or p1 == 0 or p1 == 1:
             summand3 = 0
         else:
-            summand3 = (count_ab * math.log(p1 or 1) +
-                        (count_a - count_ab) * math.log(1.0 - p1) if p1 < 1 else 0)
+            summand3 = (count_ab * math.log(p1) +
+                        (count_a - count_ab) * math.log(1.0 - p1))
 
-        if count_b == count_ab:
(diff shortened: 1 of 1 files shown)


nltk/tokenize/punkt.py
```diff
@@ -1062,25 +1062,32 @@ def _col_log_likelihood(count_a, count_b, count_ab, N):
...
-                    (count_a - count_ab) * math.log(1.0 - p))
+        p2_denom = N - count_a
+        if p2_denom == 0:
+            p2 = 1
+        else:
+            p2 = (count_b - count_ab) / p2_denom
+
+        print (p, p1, p2, N, count_a, count_b, count_ab)
+
+        summand1 = (count_ab * math.log(p + (p == 0)) +
```
dimazest: I would replace `math.log(p + (p == 0))` with `math.log(p or 1)`

Alternatively, the whole term could be replaced to

```
count_ab * math.log(p) if p else 0
```

nltk/tokenize/punkt.py
```diff
@@ -1062,25 +1062,32 @@ def _col_log_likelihood(count_a, count_b, count_ab, N):
...
+                    (count_a - count_ab) * math.log(1.0 - p + (p == 1)))
```
dimazest: is this equivalent to

```
(count_a - count_ab) * math.log(1.0 - p) if p < 1 else 0
```

stevenbird:
@mrecachinas – would you like to respond to the suggestions made by @dimazest?

nltk/tokenize/punkt.py
mrecachinas: Yes, good call. I'll add this later today.

nltk/tokenize/punkt.py
mrecachinas: I think you're right. I'll try it later.

mrecachinas:
@stevenbird -- Replied to @dimazest and made the changes. Alternatively, since we're clamping each term to 0 to avoid `log(0)`, we could catch the `ValueError` thrown. Thoughts?

nltk/tokenize/punkt.py
```diff
@@ -1062,25 +1062,32 @@ def _col_log_likelihood(count_a, count_b, count_ab, N):
...
-                    (N - count_a - count_b + count_ab) * math.log(1.0 - p))
+        summand2 = ((count_b - count_ab) * math.log(p or 1) +
+                    (N - count_a - count_b + count_ab) * math.log(1.0 - p) if p < 1 else 0)
 
         if count_a == count_ab:
             summand3 = 0
         else:
-            summand3 = (count_ab * math.log(p1) +
-                        (count_a - count_ab) * math.log(1.0 - p1))
+            summand3 = (count_ab * math.log(p1 or 0) +
```
dimazest: Good work! I believe it should be `math.log(p1 or 1)` because we want to avoid taking the log of 0.

Btw, do we have tests for this module?

mrecachinas: > Good work! I believe it should be math.log(p1 or 1) because we want to avoid taking the log of 0.

Ah whoops! Good catch. 👍 46e813a should fix that.

> Btw, do we have tests for this module?

None testing the `log(0)` that I'm aware of. I hadn't written any when I submitted my original  PR. I'm not really sure how this would come up.

stevenbird:
@mrecachinas: thanks for this. I think the code would be simpler and more maintainable if we catch `ValueError` as you suggest. If you agree, would you please make these further changes?

mrecachinas:
@stevenbird -- no problem. Let me know what you think. I opted to not catch the ValueError in the last two and instead I just checked for the cases that would cause a `log(0)` to occur. I suppose it may be worth adding the catch in the event p < 0 or p > 1, but if that's the case, I can just change the conditional to check for p <= 0 or p >= 1. For robustness, I'll do that now.

stevenbird:
Thanks @mrecachinas I think this is more readable.

stevenbird:
It's nice to have this resolved; thanks for your patience 👍

PR #1257, opened 2016-01-19
Allow corenlp_options argument for StanfordParser to be a list

almath123:
I have made a very small change to allow the corenlp_options argument to the StanfordParser initializer to be a list.

When adding corenlp arguments you might want to have arguments like '-nthreads 4'. Adding the arguments as a string doesn't work because when Popen is called to start the java process it interprets the whole string '-nthreads 4' as an argument which is unrecognised. Allowing corenlp_options to be a list allows the arguments to be added as ['-nthreads', '4'].

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -189,3 +189,4 @@
 - Sergio Oller
 - Will Monroe
 - Elijah Rippeth
+- Alexander Mathews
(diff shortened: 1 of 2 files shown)


alvations:
This might be superseded by #1249

almath123:
I have had a look and this will be superseded by #1249 . There was a commit to #1249 yesterday which added similar functionality. Thanks.

stevenbird:
Thanks @almath123, @alvations

PR #1262, opened 2016-01-22
Added function to convert Stanford JSON output to NLTK DependencyGraph object

alvations:
Added functionality to convert Stanford JSON output to NLTK DependencyGraph object, by:
- Taking the JSON output and converting it into CONLL format with `stanfordjson2conll()`
- Then, iteratively read the CONLL format into the DepdendencyGraph in `fromstanfordjson()`
- And a demo function `stanford_json_demo()` that tries to read in the graph from stanford-corenlp-full-2014-08-27 and stanford-corenlp-full-2015-04-20 outputs

(Source: http://www.eecs.qmul.ac.uk/~dm303/stanford-dependency-parser-nltk-and-anaconda.html)

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -1,8 +1,8 @@
 # Natural Language Toolkit: Dependency Grammars
 #
 # Copyright (C) 2001-2015 NLTK Project
-# Author: Jason Narad <jason.narad@gmail.com>
-#         Steven Bird <stevenbird1@gmail.com> (modifications)
+# Authors: Jason Narad <jason.narad@gmail.com>
+#          Steven Bird <stevenbird1@gmail.com> (modifications)
 #
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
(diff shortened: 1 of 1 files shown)


nltk/parse/dependencygraph.py
```diff
@@ -543,6 +543,109 @@ def nx_graph(self):
...
+        ...                               u'Lemma': u'.',
+        ...                               u'PartOfSpeech': u'.'}]]}
+        >>> list(stanfordjson2conll(stanfordjson)) # stanford-corenlp-full-2015-04-20.
+        [(2, '_', 'loves', 'love', 'VBZ', 'VBZ', '_', '0', 'ROOT', '_', '_'), (1, '_', 'John', 'John', 'NNP', 'NNP', '_', '2', 'nsubj', '_', '_'), (3, '_', 'Mary', 'Mary', 'NNP', 'NNP', '_', '2', 'dobj', '_', '_'), (4, '_', '.', '.', '.', '.', '_', '2', 'punct', '_', '_')]
+        
+        :param sentence: The JSON output from Stanford CoreNLP tool.
+        :type sentence: dict
+        :rtype: generator
+        """
+        for node in jsonparse['dependencies']:
```
dimazest, line 614: I believe that in recent corenlp version the JSON format is different, or the wrapper outputs something else.

Have a look to https://github.com/nltk/nltk/pull/1249/files#diff-3ddcbc41d8654bc0748b8afb087c55aeR633 it would be cool to have one function that does the job.

nltk/parse/dependencygraph.py
```diff
@@ -543,6 +543,109 @@ def nx_graph(self):
...
+            # Parses the node for different stanford-corenlp versions.
+            if len(node) == 5: # stanford-corenlp-full-2015-04-20.
+                rel, _, head, word, n = node
+            elif len(node) == 3: # For stanford-corenlp-full-2014-08-27.
+                rel, head, word = node
+                word, n = word.split('-')
```
dimazest, line 620: is it save to split like this, in case a token consists a `-`. `word.rsplit('-', maxsplit=1)` should be safer.

nltk/parse/dependencygraph.py
```diff
@@ -543,6 +543,109 @@ def nx_graph(self):
...
+        This function takes a JSON output from Stanford CoreNLP and converts 
+        it into a DependencyGraph object.
+    
+        :param jsonparse: The JSON output from Stanford CoreNLP tool.
+        :type sentence: dict
+        :rtype: DependencyGraph
+        """
+        # Converts Stanford JSON outputs to CONLL
+        conllparse = DependencyGraph.stanfordjson2conll(jsonparse)
+        return DependencyGraph([' '.join(items[1:]) for
```
dimazest, line 647: set `cell_separator=' '` CoreNLP uses unbreakable space inside of a token, which might break DependencyGraph.

nltk/parse/dependencygraph.py
alvations, line 614: I think it's okay to support multiple JSON formats. Since JSON is dynamic. I'll take a look and at functionalities for the new format.

alvations, line 614: It seems like the JSON output is a thing from the wrapper output instead of what Stanford CoreNLP natively outputs.  I'll close this issue and then rewrite the function.

alvations:
I think I should wait for a stable version from #1249 to be merged first. Meanwhile, I'll go back and work on the `translate` module. At some point, the Stanford tools got to work smoothly in NLTK for the dependency related MT methods to work =)

PR #1270, opened 2016-01-25
Added smoothing to BLEU scores

alvations:
- Added Chen and Cherry (2014) smoothing methods for BLEU. 
- Note that the `p_n` now needs to be materialized as opposed to how it was previously stored in a generator. It's because the `p_n` needs to be modified by the `smooth_bleu()` function. 

---

Additionally, fix related test failures from `bleu_score.py`:
- https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py34-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/nltk.test.unit.translate.test_bleu/TestBLEU/test__modified_precision/ 
- https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py34-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/(root)/bleu_doctest/bleu_doctest/

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -22,7 +22,7 @@ class TestBLEU(unittest.TestCase):
         references = [ref1, ref2] 
         
         # Testing modified unigram precision.
-        hyp1_unigram_precision =  _modified_precision(references, hyp1, n=1)
+        hyp1_unigram_precision =  float(_modified_precision(references, hyp1, n=1))
         assert (round(hyp1_unigram_precision, 4) == 0.2857)
         # With assertAlmostEqual at 4 place precision.
         self.assertAlmostEqual(hyp1_unigram_precision, 0.28571428, places=4)
@@ -59,8 +59,8 @@ class TestBLEU(unittest.TestCase):
         references = [ref1, ref2, ref3]
         
         # Unigram precision.
-        hyp1_unigram_precision = _modified_precision(references, hyp1, n=1)
-        hyp2_unigram_precision = _modified_precision(references, hyp2, n=1)
+        hyp1_unigram_precision = float(_modified_precision(references, hyp1, n=1))
+        hyp2_unigram_precision = float(_modified_precision(references, hyp2, n=1))
         # Test unigram precision with assertAlmostEqual at 4 place precision.
         self.assertAlmostEqual(hyp1_unigram_precision, 0.94444444, places=4)
         self.assertAlmostEqual(hyp2_unigram_precision, 0.57142857, places=4)
@@ -70,8 +70,8 @@ class TestBLEU(unittest.TestCase):
         
         
         # Bigram precision
-        hyp1_bigram_precision = _modified_precision(references, hyp1, n=2)
-        hyp2_bigram_precision = _modified_precision(references, hyp2, n=2)
+        hyp1_bigram_precision = float(_modified_precision(references, hyp1, n=2))
+        hyp2_bigram_precision = float(_modified_precision(references, hyp2, n=2))
         # Test bigram precision with assertAlmostEqual at 4 place precision.
         self.assertAlmostEqual(hyp1_bigram_precision, 0.58823529, places=4)
         self.assertAlmostEqual(hyp2_bigram_precision, 0.07692307, places=4)
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -423,32 +423,32 @@ class SmoothingFunction:
         ...               'Party', 'commands']
                 
         >>> chencherry = SmoothingFunction()
-        >>> print (sentence_bleu([reference1], hypothesis1))
-        0.411803763569
+        >>> print (sentence_bleu([reference1], hypothesis1)) # doctest: +ELLIPSIS
+        0.4118...
         >>> chencherry = SmoothingFunction()
-        >>> print (sentence_bleu([reference1], hypothesis1, smoothing_function=chencherry.method0))
-        0.411803763569
+        >>> print (sentence_bleu([reference1], hypothesis1, smoothing_function=chencherry.method0)) # doctest: +ELLIPSIS
+        0.4118...
         >>> chencherry = SmoothingFunction()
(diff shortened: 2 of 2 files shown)


stevenbird:
Thanks @alvations – nice addition!

alvations:
@stevenbird I've added the `# doctest: +ELLIPSIS` option the long floating points doctest. Hopefully that will prevent regression failures.

nltk/test/unit/translate/test_bleu.py
```diff
@@ -22,7 +22,7 @@ def test__modified_precision(self):
         references = [ref1, ref2] 
         
         # Testing modified unigram precision.
-        hyp1_unigram_precision =  _modified_precision(references, hyp1, n=1)
+        hyp1_unigram_precision =  float(_modified_precision(references, hyp1, n=1))
         assert (round(hyp1_unigram_precision, 4) == 0.2857)
```
dimazest, line 26: why not to use only self.assertAlmostEqual

nltk/test/unit/translate/test_bleu.py
```diff
@@ -22,7 +22,7 @@ def test__modified_precision(self):
...
         # With assertAlmostEqual at 4 place precision.
         self.assertAlmostEqual(hyp1_unigram_precision, 0.28571428, places=4)
```
dimazest, line 28: if we care only for 4 digits, `0.28571428` should become `0.2857`

nltk/translate/bleu_score.py
```diff
@@ -379,3 +398,132 @@ def _brevity_penalty(closest_ref_len, hyp_len):
...
+    :type epsilon: float
+    :param alpha: the alpha value use in method 6
+    :type alpha: int
+    :param k: the k value use in method 4
+    :type k: int
+    :return: a list of smoothed modified precisions
+    :rtype: list(float)
+    """
+    # No smoothing.
+    if method == 0:
```
dimazest: maybe methods should be functions, so it's possible to use custom smoothing method by passing a custom function.

dimazest: also, it might be better to use labels instead of numbers, e.g `none`, `add_epsilon` and so on.

nltk/test/unit/translate/test_bleu.py
alvations, line 26: There's `self.assertAlmostEqual` afterwards =)

nltk/test/unit/translate/test_bleu.py
alvations, line 28: I think it's okay to show that the `0.28571428` number doesn't necessarily need to be fit to the `places` parameter.

nltk/translate/bleu_score.py
alvations: I thought of using classes and then functions within a "smoothing" class but OOP might not be a good idea given the object overheads. In MT tuning this function would have to be called at least 7.5 million times e.g. 100 (n-best size) \* 3000 (no. of sentences in dev set) \* 25 (no. of cycles).

alvations: Actually the numbers should be preferred over boolean labels because the default values are set empirically from experiments. So i would expect users to try out multiple values instead of using the values set from Chen and Cherry.

alvations:
@dimazest I've changed the smoothing methods into functions instead. The advantage is now we can use other smoothing functions and untangle the smoothing from the BLEU implementation.

I've tried lambdas but it's rather crazy and unreadable so I think the only way is to use an object. I guess if the user really wants speed, he/she will create his own BLEU duck-type or SmoothingFunction duck-class.

nltk/translate/bleu_score.py
```diff
@@ -379,3 +400,147 @@ def _brevity_penalty(closest_ref_len, hyp_len):
...
+        self.epsilon = epsilon
+        self.alpha = alpha
+        self.k = k
+        
+    def method0(self, p_n, **kwargs):
+        # No smoothing.
+        return p_n
+        
+    def method1(self, p_n, **kwargs):
+        # Smoothing method 1: Add *epsilon* counts to precision with 0 counts.
```
dimazest: Convert comments to docstrings.

nltk/translate/bleu_score.py
```diff
@@ -379,3 +400,147 @@ def _brevity_penalty(closest_ref_len, hyp_len):
...
+        """
+        self.epsilon = epsilon
+        self.alpha = alpha
+        self.k = k
+        
+    def method0(self, p_n, **kwargs):
+        # No smoothing.
+        return p_n
+        
+    def method1(self, p_n, **kwargs):
```
dimazest: add *args

dimazest:
Looks good, just two suggestions.

Maybe it's better to define methods as classes that have `__call__` method:

``` python

class Smoothing(object):

    def __init__(self, epsilon=0.1, alpha=5, k=5):
        ... 

class Method1(Smoothing):

    def __call__(self, p_n, *args, **kwargs):  # might be a good idea to include args as well
        return [(p_i.numerator + self.epsilon)/ p_i.denominator 
                    if p_i.numerator == 0 else p_i for p_i in p_n]]

```

alternatively, closures can be used:

``` python
def method1(epsilon=0.1, alpha=5, k=5):
    def inner(p_n, *args, **kwargs):
        return [(p_i.numerator + epsilon)/ p_i.denominator  if p_i.numerator == 0 else p_i for p_i in p_n]]

    return inner
```

alvations:
Using `__call__` would look a little weird, since the user would have to create N different objects to use N no. of different smoothing techniques. There's be another few more papers on BLEU smoothing and it might turn into a explosion of classes if we use `__call__` and naming the classes might be quite hard =)

For now, we can still call it `Smoothing`/`SmoothFunction` but I think when more smoothing methods are added this might be called `ChenAndCherrySmoothing`. And I would expect to use something like `cc = ChenAndCherrySmoothing(); cc.method1(); cc.method2()` instead of  `ChenAndCherrySmoothingMethod1(); ChenAndCherrySmoothingMethod2()`.

Any thoughts?

dimazest:
that makes sense.

stevenbird:
Thanks @alvations, and also @dimazest for timely feedback.

Issue #1252, opened 2016-01-16
Tox fails with "ERROR: Failure: ImportError (No module named 'six')"


ExplodingCabbage:
When I try to run the tests with Tox (on Ubuntu) from within a local clone of the repo, it manages to install the dependencies but blows up when trying to import things from within NLTK.

I imagine I can work around this by figuring out how to manually run just the tests I care about, but it's inconvenient.

I'm not sure whether I'm doing something dumb or whether the Tox setup is broken; if the former, the CONTRIBUTING docs should probably mention what needs to be done besides just running Tox; if the latter, it should probably be fixed.

Here's the full output (had to pastebin it due to GitHub's post length limit):

http://pastebin.com/ENuCLnv6

PR #1274, opened 2016-01-26
Fix probability doctest

goodmami:
This PR does two things:
- fixes an import error which was causing some tests to fail (see #1252)
- fixes probability.doctest so it passes

The first thing is unrelated, but I fixed it so I could actually run the unit tests to ensure the second change worked. The problem with the doctests was easy: dictionary order is arbitrary and thus dictionary representations don't work well with doctests.


stevenbird:
Thanks @goodmami – nice solution

Issue #1237, opened 2015-12-21
Automatically adding classpath for Stanford Tools 


alvations:
The [new Stanford NER tool](http://nlp.stanford.edu/software/stanford-ner-2015-12-09.zip) and [parser](http://nlp.stanford.edu/software/lex-parser.shtml) requires more `.jar` files to be added to the classpath. Possibly, it's easier to modify the `nltk.tag.stanford`  to do something like what 
- `nltk.parse.malt` is doing with the dependencies check using `find_jar_within_paths`  (e.g. https://github.com/nltk/nltk/blob/develop/nltk/parse/malt.py#L66) https://github.com/nltk/nltk/blob/develop/nltk/internals.py#L722. 

More details on http://stackoverflow.com/questions/34361725/nltk-stanfordnertagger-noclassdeffounderror-org-slf4j-loggerfactory-in-windo/

alvations:
Given the new corenlp v3.6, maybe that's the way to go wrap around the JSON output: https://github.com/smilli/py-corenlp and something like https://github.com/alvations/nltk/blob/stanfordnlp/nltk/parse/dependencygraph.py#L636

That way we don't have to play catch up with Stanford NLP tools all the time and say that we only support tagger/parsers/tools versions from corenlp

alvations:
@stanfordnlp Any thoughts on whether third-party wrappers to Stanford tools should wrap the individual tools or the Stanford CoreNLP? If we're wrapping Stanford CoreNLP, what's an easy way to specify language or a specific model for a specific task to the server query?

longdt219:
Hi @stevenbird, 
All the bugs related to Stanford packages shown in the CI server is due to the new logging requirements such that <b> slf4j-api.jar</b> and <b> slf4j-simple.jar </b> are needed. We can do some hackings as suggested by @alvations  <a href = 'https://gist.github.com/alvations/e1df0ba227e542955a8a'> here </a> to add these additional .jar files to the classpath through the `find_jar_within_paths` function. What do you think ?

dimazest:
here is what I did to make it working https://github.com/nltk/nltk/pull/1249/files#diff-3ddcbc41d8654bc0748b8afb087c55aeR40

longdt219:
Vote for @alvations'solution since we don't have to manually specify the jar file names.

rapaliwal:
Hi,
I am also facing the same issue. I believe its not fixed yet? Please comment in case it is or there is any way to get it working. Thanks

[...]

PR #1280, opened 2016-02-02
Adding logging jar to the Stanford CP

longdt219:
Hacking to add logging jar file to the classpath for Stanford POS tagger, parser, tokenizer, NER as discussed in #1237

### DIFF
--- a/nltk/tokenize/stanford.py
+++ b/nltk/tokenize/stanford.py
@@ -15,7 +15,7 @@ import json
 from subprocess import PIPE
 
 from nltk import compat
-from nltk.internals import find_jar, config_java, java, _java_options
+from nltk.internals import find_jar, config_java, java, _java_options, find_jars_within_path
 
 from nltk.tokenize.api import TokenizerI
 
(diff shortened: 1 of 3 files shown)


nltk/tokenize/stanford.py
```diff
@@ -43,7 +43,11 @@ def __init__(self, path_to_jar=None, encoding='utf8', options=None, verbose=Fals
             searchpath=(), url=_stanford_url,
             verbose=verbose
         )
-
+        
+        # Adding logging jar files to classpath 
+        stanford_dir = self._stanford_jar.rpartition('/')[0]
```
alvations, line 48: Use `os.path.split(self._stanford_jar)[0]` instead of `self._stanford_jar.rpartition('/')[0]` instead. That will return the parent path regardless of OS, using `/` is problematic for Windows users. My quick hack on the [gist](https://gist.github.com/alvations/e1df0ba227e542955a8a) was made within little thoughts on library maintenance =)

alvations:
Although I would prefer to wait for @dimazest 's API to be stable and be merged but I think there's no harm putting in the hack until the Stanford CoreNLP API is stable. That'll also resolve some of the regression issues on CI server (i think).

stevenbird:
Thanks @longdt219. Yes, that's the hope @alvations.

alvations:
@longdt219 @stevenbird you've missed the `os.path.split()` edits on the other 2 files from `nltk.tag.stanford` and `nltk.parse.stanford` =) Sorry I was too lazy and only commented on one of them.

longdt219:
Hi @stevenbird , I created another PR #1281

PR #1281, opened 2016-02-02
Adding os.path.split() for finding the jar folder

longdt219:
Regarding PR #1280, using os.path.split() instead of rpartition


alvations:
Thanks @longdt219 !

alvations:
@stevenbird sorry, I've closed this mistakenly.

PR #1287, opened 2016-02-17
added lesk using sentiwordnet.

drevicko:
Added sentiwordnet version of lesk.

This is useful as synsets between wordnet and sentiwordnet are not all the same. Word senses returned from wordnet may not exist in sentiwordnet and so sentiment scores are thus unavailable (an example is "like", an important sentiment indicator). 

imports of both wordnet and sentiwordnet are attempted, but an exception from import problems only occurs when lesk or seltilesk functions are called. Perhaps not ideal - happy for suggestions on a better way.

### DIFF
--- a/nltk/wsd.py
+++ b/nltk/wsd.py
@@ -7,8 +7,15 @@
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 
-from nltk.corpus import wordnet
+try:
+    from nltk.corpus import wordnet
+except ImportError:
+    pass
 
+try:
+    from nltk.corpus import sentiwordnet
+except ImportError:
+    pass
 
 def lesk(context_sentence, ambiguous_word, pos=None, synsets=None):
     """Return a synset for an ambiguous word in a context.
@@ -33,6 +40,34 @@ def lesk(context_sentence, ambiguous_word, pos=None, synsets=None):
     Systems Documentation. ACM, 1986.
     http://dl.acm.org/citation.cfm?id=318728
     """
+    return _lesk(context_sentence, ambiguous_word, pos, synsets, wordnet)
+
+def sentilesk(context_sentence, ambiguous_word, pos=None, synsets=None):
+    """Return a synset for an ambiguous word in a context using sentiwordnet as a source of synsets.
+
+    :param iter context_sentence: The context sentence where the ambiguous word
+    occurs, passed as an iterable of words.
+    :param str ambiguous_word: The ambiguous word that requires WSD.
+    :param str pos: A specified Part-of-Speech (POS).
+    :param iter synsets: Possible synsets of the ambiguous word.
+    :return: ``lesk_sense`` The Synset() object with the highest signature overlaps.
+
+    This function is an implementation of the original Lesk algorithm (1986) [1].
+
+    Usage example::
+
+        >>> lesk(['I', 'went', 'to', 'the', 'bank', 'to', 'deposit', 'money', '.'], 'bank', 'n')
+        SentiSynset('savings_bank.n.02')
+
+    [1] Lesk, Michael. "Automatic sense disambiguation using machine
+    readable dictionaries: how to tell a pine cone from an ice cream
+    cone." Proceedings of the 5th Annual International Conference on
+    Systems Documentation. ACM, 1986.
+    http://dl.acm.org/citation.cfm?id=318728
+    """
+    return _lesk(context_sentence, ambiguous_word, pos, synsets, sentiwordnet)
+
+def _lesk(context_sentence, ambiguous_word, pos=None, synsets=None, wordnet):
 
     context = set(context_sentence)
     if synsets is None:


alvations:
Is there a reason to ducktype the `lesk()` -> `sentilesk()`? Essentially, SentiWordNet and Princeton WordNet shares the same index of Synsets, no? 

Unless, there are specific things in SentiWordNet that helps the disambiguation of the sense, (e.g. more example sentences, additional synset relations) and `sentilesk()` makes uses of these information to improve disambiguation, I am not sure whether duck-typing lesk would be a good idea. 

Adding a `wordnet` parameter to `_lesk()` may be a good idea, but I'm not sure whether we can use "other" wordnet since the assumption now is that all wordnet is covered under a shared under one `Inter-Lingual-Index` so technically, there should be only one "global" wordnet and language specific wordnets are subset of that. But I'm not the best person to talk about this. @fcbond and the OMW team can advise more on that.

---

Actually, there's no discrepancies between the no. of synsets =)

``` python
>>> from nltk.corpus import sentiwordnet as swn
>>> len(list(swn.all_senti_synsets()))
117659
>>> len(list(wn.all_synsets()))
117659
```

drevicko:
Apparently wordnet and sentiwordnet don't share the same synsets. 

```
from nltk.corpus import sentiwordnet as swn
from nltk.wsd import lesk

sent = 'I went to the nice bank to deposit my lovely money'
ambiguous = 'lovely'

synset = lesk(word_tokenize(sent), ambiguous)
print(synset.name())
print(swn.senti_synset(synset.name()))
print(list(swn.senti_synsets(ambiguous)))
```

Outputs:

> lovely.s.01
> None
> [SentiSynset('cover_girl.n.01')]

btw: There's some things need fixing for this pull request to work as senti-synsets don't have all the methods of synsets (but contain a synset attribute that does).

alvations:
I'm fetching it from SentiWordNet though:

``` python
>>> import nltk
>>> nltk.__version__
'3.1'
>>> from nltk.corpus import sentiwordnet as swn
>>> wn.synset('lovely.n.1')
Synset('cover_girl.n.01')
>>> ss_name = wn.synset('lovely.n.1').name()
>>> swn.senti_synset(ss_name)
SentiSynset('cover_girl.n.01')
```

Which version of NLTK are you using? What is the output from `import nltk; print nltk.__version__`?

drevicko:
'3.1'
It's from anaconda...
I downloaded swn yesterday (using `nltk.download("sentiwordnet"))` )
Do you know if there's a way to check the versions of wordnet/sentiwordnet?

what do you get from `wn.synset('lovely.s.01')`? That's the one that was giving me trouble...

btw: I'm using python 3.5, in case that makes a difference...

alvations:
It still works with `lovely.s.01` on python3.5:

```
>>> from nltk.corpus import sentiwordnet as swn
>>> ss = wn.synset('lovely.s.01')
>>> swn.senti_synset(ss.name())
SentiSynset('lovely.s.01')
```

Maybe @cgpotts has an idea of what went wrong.

drevicko:
Well, that's strange! I guess we've somehow got different versions of wordnet?

In my ipython notebook:

```
from nltk.corpus import sentiwordnet as swn
ss = wn.synset('lovely.s.01')
print(swn.senti_synset(ss.name()))
print(nltk.__version__)
!python --version
```

> None
> 3.1
> Python 3.5.1 :: Anaconda 2.4.1 (x86_64)

Mine is anaconda python, but I don't suppose that would make a difference?

In `~/nltk_data/corpora/wordnet/README` I see WordNet 3.0, for sentiwordnet I find SentiWordNet 3.0

alvations:
@drevicko Can you check whether you have the same output in your `sentiwordnet.zip`:

``` python
>>> import zipfile
>>> swnzip = '/home/alvas/nltk_data/corpora/sentiwordnet.zip'
>>> [zifo.CRC for zifo in zipfile.ZipFile(swnzip).infolist()]
[0, 30618389, 2969739728]
```

If you don't have the same number, deleted the sentiwordnet folder and the zipfile in your `nltk_data`, then re-download with `nltk.download('sentiwordnet')`. Then check the output from the code above again with the newly downloaded `sentiwordnet.zip`. 

If you can't get the same output after you've deleted and redownloaded the `sentiwordnet.zip`, then I think it's a packaging issue on Continuum `conda`, they would have to update their NLTK package and the data links that goes with it. You can either wait for them to patch/update their distribution of NLTK or use `pip` instead of `conda` to install NLTK (i.e. Do `pip install -U nltk`)

Otherwise, if the output from the new zipfile is the same as above, re-run these code to see whether you get the same output:

```
>>> from nltk.corpus import sentiwordnet as swn
>>> ss = wn.synset('lovely.s.01')
>>> swn.senti_synset(ss.name())
SentiSynset('lovely.s.01')
```

alvations:
@stevenbird I think it's safe to close this until further issues raised with regards to sentiwordnet not mapping to wordnet. They should match, as shown in the examples above.

stevenbird:
Thanks @alvations.

PR #1295, opened 2016-02-22
Use OS specific classpath separator

alvations:
Windows uses `;` semi-colons to separate Java classpaths and Linux/MacOS uses `:` colons.

(See http://stackoverflow.com/questions/4528438/classpath-does-not-work-under-linux)


stevenbird:
Thanks @alvations

PR #1300, opened 2016-02-24
Fix implementation error on Method3 Smoothing

alvations:
The smooth should only affect the precision count not the ngram precision itself, so the `p_i.denominator` needs to be considered, i.e.`(1/2**k) / p_i.denominator`. Simplified mathematically to `1/ (2**k * p_i.denominator)`


stevenbird:
Thanks @alvations

Issue #1299, opened 2016-02-24
zipfile problem for Python 3.5


stevenbird:
The bulk of the CI errors concern zipfiles in Python 3.5:

https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/

alvations:
The issue happens when automatically opening a zipfile when it hasn't been previously extracted in the `nltk_data` directory, e.g.

```
>>> from nltk.corpus import stopwords
>>> stopwords.words('english')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/alvas/git/nltk/nltk/corpus/reader/wordlist.py", line 20, in words
    return [line for line in line_tokenize(self.raw(fileids))
  File "/home/alvas/git/nltk/nltk/corpus/reader/wordlist.py", line 26, in raw
    return concat([self.open(f).read() for f in fileids])
  File "/home/alvas/git/nltk/nltk/corpus/reader/wordlist.py", line 26, in <listcomp>
    return concat([self.open(f).read() for f in fileids])
  File "/home/alvas/git/nltk/nltk/corpus/reader/api.py", line 210, in open
    stream = self._root.join(file).open(encoding)
  File "/home/alvas/git/nltk/nltk/data.py", line 505, in open
    data = self._zipfile.read(self._entry)
  File "/home/alvas/git/nltk/nltk/data.py", line 982, in read
    self.close()
  File "/usr/lib/python3.5/zipfile.py", line 1610, in close
    self._fpclose(fp)
  File "/usr/lib/python3.5/zipfile.py", line 1714, in _fpclose
    assert self._fileRefCnt > 0
AssertionError

```

The issue may seem to come from the changes from python3.4 -> 3.5: https://fossies.org/diffs/Python/3.4.3_vs_3.5.0/Lib/zipfile.py-diff.html which cause the code to break at https://github.com/nltk/nltk/blob/develop/nltk/data.py#L505

But there isn't a difference in the API in how python opens and read a zipfile in the buffer:

```
alvas@ubi:~$ python3.4
Python 3.4.1 (default, Jun  4 2014, 11:27:44) 
[GCC 4.8.3] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import zipfile
>>> zf = zipfile.ZipFile('/home/alvas/nltk_data/corpora/stopwords.zip')
[...]

Issue #1204, opened 2015-11-11
Zipfile corpus reader breaks in Python 3.5


stevenbird:
The zipfile corpus reader breaks in Python 3.5.

```
>>> from nltk.corpus import propbank
>>> propbank.instances()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/nltk-3.0.5-py3.5.egg/nltk/util.py", line 716, in __repr__
    for elt in self:
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/nltk-3.0.5-py3.5.egg/nltk/corpus/reader/util.py", line 282, in iterate_from
    self._open()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/nltk-3.0.5-py3.5.egg/nltk/corpus/reader/util.py", line 209, in _open
    self._stream = self._fileid.open(self._encoding)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/nltk-3.0.5-py3.5.egg/nltk/data.py", line 505, in open
    data = self._zipfile.read(self._entry)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/nltk-3.0.5-py3.5.egg/nltk/data.py", line 982, in read
    self.close()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/zipfile.py", line 1607, in close
    self._fpclose(fp)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/zipfile.py", line 1711, in _fpclose
    assert self._fileRefCnt > 0
AssertionError
```

dlukes:
Any idea why? And is anyone currently investigating? Or should I try to fix it myself (I was planning to use the zipfile corpus reader in a project) and submit a pull request if I manage to get it working again?

Thanks so much for NLTK by the way! :)

alvations:
The same thing was reported in #1256. @JiHongyu suggested a fix with a +1. But I'm not sure what exactly is wrong. So sorry that I can't fix this for now because I'm busy with something else.

dlukes:
Thanks for the reference!

[...]

PR #1301, opened 2016-02-25
Fixing the ZipFile issue in Python3.5

alvations:
Fixing the assert issue when ZipFile object closes in Python3.5 from #1204 and #1299 

I tried to emulate `zipfile.ZipFile.read(self, name)` without the setting the mode but maybe we should add `mode=r` (but i think that's just a personal preference for explicit over implicit)


stevenbird:
Thanks @alvations!!

alvations:
Seems like it only resolves part of the problem =(

PR #1306, opened 2016-02-26
Add +NORMALIZE_WHITESPACE in a CCG test

tanin47:
Otherwise the unicode support test would fail because there are a few white spaces after the first two lines of the result.

I'm not sure why other tests in `ccg.doctest` doesn't need NORMALIZE_WHITESPACE. Weird.

I'd also like to mention that, Python 2.7, it fails because `ó` is printed as a different character (apart from the additional whitespace). In Python 3, the character is printed correctly.


stevenbird:
Thanks @tanin47

Issue #1285, opened 2016-02-16
Weird issue with bleu scores


spijs:
I discovered a weird issue when trying to calculate bleu scores for individual sentences with evaluation code [https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py](here).
The method that contains the weird behaviour is : `sentence_bleu`
The problem occurs whenever the pn value becomes 1.

When for example calculating the bleu 1 score for the following candidate sentence:
_[this,is,an,example]_
When given two reference sentences:
_[this,is,something]
[this,is,an,example]_

`_modified_precision` correctly gives 1 as an answer, since the sentence is exactly the reference sentence. Then the logarithm gets taken which results in a zero value. This still seems correct. Next all the different p_n's gets summed which in this case is just one zero value. So the sum is still 0.

Then however the following code seems rather odd:
`if sum_s == 0:
        return 0`

This way the resulting bleu-score is 0. I don't understand why this needs to be here.
Without this if-clause, the zero-value correctly gets put in the rest of the formula and a correct 1 is returned.

alvations:
Now this is a wacky bug that has to do with the nature of BLEU and not the implementation ;P

The if [`sum_s == 0: return 0`](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L94) is a hack to handle cases where 0 words in the ngram matches the reference sentences. So that part is correct.

The problem lies with the [summing of the precision scores](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L91):

``` python
    # Calculates the overall modified precision for all ngrams.
    # By sum of the product of the weights and the respective *p_n*
    s = (w * math.log(p_i) if p_i else 0 
         for w, p_i in zip(weights, p_n))
    sum_s = math.fsum(s)
```

In BLEU equation in the [original BLEU paper](http://www.aclweb.org/anthology/P02-1040.pdf) section 2.3:
- Note the `log(p_n)` here, if `p_n == 1`, you will get a 0 although your precision is 1.
- Then note the `exp(sum(w_n * p_n))`, if `sum(w_n * p_n) == 0` meaning that none of the ngrams in the hypothesis appears in the references, it returns a BLEU of 1. 

---

If we do some debugging:

``` python
>>> import math
>>> weights = [0.25] * 4
>>> p_n = [1] * 4
>>> s = [w * math.log(p_i) if p_i else 0 for w, p_i in zip(weights, p_n)]
>>> s
[...]

PR #1319, opened 2016-03-04
Fixed the issue of the denominator always being 1 

josiahwang:
Fixed the issue of the denominator always being 1 if the numerator is 0, as Python normalizes fractions by default. This affects smoothing methods 3 and 4.

e.g. Fraction(0,100) always returns Fraction(0,1) if "_normalize = False" is not set.

Otherwise, the previous fix for smoothing method 3 does not do anything because the denominator is always 1!

### DIFF
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -291,7 +291,7 @@ def _modified_precision(references, hypothesis, n):
     numerator = sum(clipped_counts.values())
     denominator = sum(counts.values())  
     
-    return Fraction(numerator, denominator)  
+    return Fraction(numerator, denominator, _normalize=False)  
     
 
 def _closest_ref_length(references, hyp_len):
@@ -490,7 +490,7 @@ class SmoothingFunction:
         """
         incvnt = 1 # From the mteval-v13a.pl, it's referred to as k.
         for i, p_i in enumerate(p_n):
-            if p_i == 0:
+            if p_i.numerator == 0:
                 p_n[i] = 1 / (2**incvnt * p_i.denominator)
                 incvnt+=1
         return p_n
@@ -505,7 +505,7 @@ class SmoothingFunction:
         """
         incvnt = 1 
         for i, p_i in enumerate(p_n):
-            if p_i == 0 and hyp_len != 0:
+            if p_i.numerator == 0 and hyp_len != 0:
                 p_n[i] = incvnt * self.k / math.log(hyp_len) # Note that this K is different from the K from NIST.
                 incvnt+=1
         return p_n
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -172,7 +172,7 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     bp = _brevity_penalty(ref_lengths, hyp_lengths)
     
     # Collects the various precision values for the different ngram orders.
-    p_n = [Fraction(p_numerators[i], p_denominators[i]) 
+    p_n = [Fraction(p_numerators[i], p_denominators[i], _normalize=False) 
            for i, _ in enumerate(weights, start=1)]
     
     # Smoothen the modified precision.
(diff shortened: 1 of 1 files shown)


alvations:
@josiahwang Thanks for the fix! Sorry that we've missed that out.

For consistency, could you also add the `_normalized=False` for https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L175? (for corpus_bleu, we need the same just in case denominator is 0)

@anyone_interested We really need to relook at the BLEU implementation to imitate `mteval-13a.pl` and proper tests and comparisons to resolve #1285 properly. I'll work on that once the tokenizers PR is completed. 

---

Bug verification:

``` python
>>> from fractions import Fraction
>>> x = Fraction(0,250)
>>> x.denominator
1
```

alvations:
Could you also remove https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L279 and L280? (I think) This is from the original implementation of BLEU in NLTK.

Removing L279 and L280 will make the script a little slower when facing empty translation. But this is to account for the smoothing that should happen later, So this is yet another theoretic problem with BLEU, when denominator is 0 and when numerator is 0, an error would occur. It's not discussed in the original Chen and Cherry paper so I guess we'll have to decide this for ourselves and normalize the denominator when both `numerator == denominator == 0`.

josiahwang:
Python will throw a ZeroDivisionError if we try to create Fraction(0,0). So removing L279 and L280 might not be wise. I guess it's just a choice of whether to check this earlier or later...

On another note, Fraction(0) will produce Fraction(0,1). So the current code will assume denominator == 1 when it is 0.

alvations:
Having `Fraction(0,1)` would technically be wrong. 

E.g. if the hypothesis has N words, we would like `Fraction(0,N)` instead of `Fraction(0,1)`

`Fraction(0,1)`:
- Smoothing 2  -> `(0+1)/(1+1)`
- Smoothing 3 -> `(1/ (2**1 * 1))` 

`Fraction(0,N)`:
- Smoothing 2  -> `(0+1)/(1+1)`
- Smoothing 3 -> `(1/ (2**1 * N))`, given than N >> 1, we get a better smooth.

---

Maybe change https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L279 to:

```
if not count and not hypothesis:
```

might be better and we assign `Fraction(0,1)` only when both denominator and numerator is zero (which should not be the case, given that denominator should never be zero unless it's an empty reference sentence).

alvations:
Ah after some thought and tests, if `len(hypothesis)` less than `len(weights)`, denominator also is equals to zero... It'll be another theory vs implementation wack-a-mole.

josiahwang:
>  if the references has N words, we would like Fraction(0,N) instead of Fraction(0,1)

I assume you mean the hypothesis/candidate has N words.

> Ah after some thought and tests, if len(hypothesis) less than len(weights), denominator also is equals to zero...

Yes, I've encountered this with the data I'm working on, when the hypothesis is too short. E.g. "good dog". So you will get a 0 when computing BLEU-3 or BLEU-4 because there are no 3-grams/4-grams => denominator==0

For example:

```
hypothesis: good dog
reference: not a bad animal

To compute BLEU-3
p_1 = 0/2; smoothed_method3(p_1) = 1/(2^^1 * 2) = 1/4
p_2 = 0/1; smoothed_method3(p_2) = 1/(2^^2 * 1) = 1/4
p_3 = 0/0; smoothed_method3(p_3) = 1/(2^^3 * 0) = NaN 
```

I _think_ mteval-v13a.pl ignores smoothing for such cases, and just sets it to 0. Or rather set log(NaN)=0.

```
BLEU-3 = brevity_penalty * exp(1/3 * ( log(1/4) + log(1/4) + 0 ) )
```

They've also set brevity_penalty to 0 if it is an empty hypothesis (no unigrams). Which I think should probably also be modified in https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L401 to avoid division-by-zero errors.

I'm not familiar enough with Perl, so someone should double check to confirm.

alvations:
I think it's okay to merge this for now. But we need to sit down and look through the equations seriously to come up with:
1. Code to emulate `mteval-13a.pl`
2. NLTK implementation of the BLEU that makes most sense to us.
3. Resolve the wack-a-mole fringe cases with a lot more unittests

stevenbird:
Thanks @josiahwang.
@alvations - I'll go ahead and merge this; feel free to open a new issue or two so we don't lose track of what still needs to happen, thanks.

stevenbird:
I'm seeing lots of new doctest failures on our CI server after this merge

PR #1321, opened 2016-03-08
Add optional CCG semantics computation

tanin47:
Here are important things to mention:
- The syntax looks like this: `has => (S\\NP)/NP {\\x y.have(y, x)}`.  `Lexicon.fromstring(lex_str, include_semantics=False)` allows user to use the semantic computation.
- Previously, a leaf node only contained a category, which is of type `String`. Now a leaf node contains a `Token`, which contains a category (`String`) and a semantic (`Expression`). It enables us to compute semantics.
- I have removed `SKIP` from all tests in `nltk/test/ccg.doctest`. All tests pass.
- I'd like to create tests according to certain published papers to prove its correctness. I wonder if anyone could suggest me one. I saw [this paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.48.4937&rep=rep1&type=pdf), but it needs a variable category type, which is not implemented yet.

It might be easier to understand if you read the tests first: https://github.com/tanin47/nltk/blob/tanin-add-semantic-2/nltk/test/ccg/chart.doctest

Please let me know if you have any suggestion. Thank you!

### DIFF
--- a/nltk/ccg/chart.py
+++ b/nltk/ccg/chart.py
@@ -37,12 +37,15 @@ from nltk.parse import ParserI
 from nltk.parse.chart import AbstractChartRule, EdgeI, Chart
 from nltk.tree import Tree
 
-from nltk.ccg.lexicon import fromstring
+from nltk.ccg.lexicon import fromstring, Token
 from nltk.ccg.combinator import (ForwardT, BackwardT, ForwardApplication,
                                  BackwardApplication, ForwardComposition,
                                  BackwardComposition, ForwardSubstitution,
                                  BackwardBx, BackwardSx)
 from nltk.compat import python_2_unicode_compatible, string_types
+from nltk.ccg.combinator import *
+from nltk.ccg.logic import *
+from nltk.sem.logic import *
 
 # Based on the EdgeI class from NLTK.
 # A number of the properties of the EdgeI interface don't
@@ -73,14 +76,14 @@ class CCGLeafEdge(EdgeI):
     '''
     Class representing leaf edges in a CCG derivation.
     '''
-    def __init__(self, pos, categ, leaf):
+    def __init__(self, pos, token, leaf):
         self._pos = pos
-        self._categ = categ
+        self._token = token
         self._leaf = leaf
-        self._comparison_key = (pos, categ, leaf)
+        self._comparison_key = (pos, token.categ(), leaf)
 
     # Accessors
-    def lhs(self): return self._categ
+    def lhs(self): return self._token.categ()
     def span(self): return (self._pos, self._pos+1)
     def start(self): return self._pos
     def end(self): return self._pos + 1
@@ -91,7 +94,8 @@ class CCGLeafEdge(EdgeI):
     def is_incomplete(self): return False
     def nextsym(self): return None
 
-    def categ(self): return self._categ
+    def token(self): return self._token
+    def categ(self): return self._token.categ()
     def leaf(self): return self._leaf
 
 @python_2_unicode_compatible
@@ -202,8 +206,8 @@ class CCGChartParser(ParserI):
 
         # Initialize leaf edges.
         for index in range(chart.num_leaves()):
-            for cat in lex.categories(chart.leaf(index)):
-                new_edge = CCGLeafEdge(index, cat, chart.leaf(index))
+            for token in lex.categories(chart.leaf(index)):
+                new_edge = CCGLeafEdge(index, token, chart.leaf(index))
                 chart.insert(new_edge, ())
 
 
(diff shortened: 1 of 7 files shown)


stevenbird:
@jasonbaldridge – do you have any feedback on this, or know anyone else who could advise please?

jasonbaldridge:
Assuming that what you would like are a series of tests, I'd look at the testbeds that have been created at OpenCCG. Probably the easiest way to start with those is to look at the DSL we created for defining CCG grammars, which includes testbeds with every grammar. Here's the main directory for those:

https://github.com/OpenCCG/openccg/tree/master/ccg-format-grammars

For the very simplest grammar and testbed, see this:

https://github.com/OpenCCG/openccg/blob/master/ccg-format-grammars/tinytiny/tinytiny.ccg

E.g.:

testbed {
  the policemen eat:    2;
  the boys eat: 2;
  the boys eat the peaches: 1;
  the policeman sleeps: 1;
  the policemen sleeps: 0;
  the policemen sleep: 1;
  the policeman sleeps the peach: 0;
  the policeman saw me: 1;
  the policeman saw I: 0;
  I see the book: 1;
  I sees the book: 0;
  I see a book: 1;
  I see a books: 0;
}

The numbers indicate the number of analyses the grammar should produce for the string.

tanin47:
@jasonbaldridge Thank you for the suggestion. I'll take a look at it soon.

stevenbird:
@jasonbaldridge – thanks!
@tanin47 – I'm going to merge this since it's passing existing tests including ones we had been skipping; we can think about any extra tests in a separate PR

stevenbird:
I aggregated the doctests in https://github.com/nltk/nltk/commit/daa24998ed9017cb71d9b38eb8edc279b3f9b4ef
since I prefer to keep them at the top level.

stevenbird:
@tanin47 – are the idempotent leaf productions meant to be there in the output? This breaks our tests on the CI server, [e.g.](https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py27-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/%28root%29/ccg_doctest/ccg_doctest/)

tanin47:
Sorry about forgetting to mention it. I did remove the leaves from the output because they are redundant. I did remove those leaves from the tests as well (https://github.com/nltk/nltk/pull/1321/files#diff-7ecd46ce416cfc997522bcc265c2cefaL352). And that's why the tests fail.

Here is the line that skip the leaves: https://github.com/nltk/nltk/blob/develop/nltk/ccg/chart.py#L338. Anyway, the tests shouldn't break. I'll investigate.

---

The rationale is that:

```
     you    prefer      that   cake
     NP   ((S\NP)/NP)  (NP/N)   N
                      -------------->
                            NP
         --------------------------->
                   (S\NP)
    --------------------------------<
```

is much shorter than:

```
     you    prefer      that   cake
     NP   ((S\NP)/NP)  (NP/N)   N
    -----Leaf
     NP
         -------------Leaf
          ((S\NP)/NP)
                      --------Leaf
                       (NP/N)
                              ------Leaf
                                N
                      -------------->
                            NP
         --------------------------->
                   (S\NP)
    --------------------------------<
                   S
```

while both contain the same information. (I'm happy to roll this part back if it doesn't make sense.)

stevenbird:
I agree we want the short format. There's a mismatch between the code and doctests, according to our CI server. Thanks.

tanin47:
The current `develop` branch's tests are fine for me because I used Python3. From the CI, it only fails for Python2.7. My bet is that `if op is 'Leaf'` works differently on Python2.7. I'll fix it.

Another thing is that `ccg_semantics.doctest` fails on the CI for every python version:

```
~/Project/nltk on develop o [16:52:15]
$ python3 -m doctest nltk/test/ccg_semantics.doctest
**********************************************************************
File "nltk/test/ccg_semantics.doctest", line 452, in ccg_semantics.doctest
Failed example:
    print(str(lexicon.fromstring(
        '''
        :- S,NP

        IntransVsg :: S\\NP[sg]

        sleeps => IntransVsg {\\x.sleep(x)}
        eats => S\\NP[sg]/NP {\\x y.eat(x,y)}

        and => var\\var/var {\\x y.x & y}
        ''',
        True
    )))
Expected:
    and => ((_var1\_var1)/_var1) {(\x y.x & y)}
    eats => ((S\NP['sg'])/NP) {\x y.eat(x,y)}
    sleeps => (S\NP['sg']) {\x.sleep(x)}
Got:
    and => ((_var2\_var2)/_var2) {(\x y.x & y)}
    eats => ((S\NP['sg'])/NP) {\x y.eat(x,y)}
    sleeps => (S\NP['sg']) {\x.sleep(x)}
```

var1 and var2 are different depending on how many times the lexicon is run. The number is incremented.

I'll also find a way to resolve this.

nltk-webdeploy:
Thanks. Perhaps it can be done using an equality check. But note we have
other tests that fail for the same reason, and no satisfactory resolution.

tanin47:
Both of them are fixed by #1354

tanin47:
And thank you for the merge. I'll continue to work on the @jasonbaldridge's suggestion on the tests.

tanin47:
Hi @jasonbaldridge, I'm a bit confused. Does https://github.com/OpenCCG/openccg/blob/master/ccg-format-grammars/tinytiny/tinytiny.ccg contains semantic predicates? Does OpenCCG support semantic computation? Thanks!

jasonbaldridge:
@tanin47 Yes, it does! It uses hybrid logic semantic representations:

http://www.aclweb.org/anthology-new/P/P02/P02-1041.pdf

That's the original paper where Geert-Jan Kruijff and I used hybrid logics with CCG, but for more extensive and more updated perspectives, you should look to Mike White's papers:

http://u.osu.edu/white.1240/papers/

I'd especially look at this paper of Mike's:

http://www.ling.ohio-state.edu/~mwhite/papers/White-RoLC-2004-to-appear.pdf

Here's an example of an intransitive verb declaration:

def verbcat(args, sem) {
  # A \* here corresponds to <prop name="[*DEFAULT*]"/>.
  # Similar use of \* appears above in [lex=_] in prepositions,
  # and <det>_ in determiners.
  s<1>[E] \ np<2>[X NUM PERS nom] args:
     E:action(\* <Actor>X:animate-being sem)
}

The " E:action(\* <Actor>X:animate-being sem)" is the semantic part, and it is instantiated by unification with the E and X variables attached to the s and np categories. BTW, it's in this grammar: https://github.com/OpenCCG/openccg/blob/master/ccg-format-grammars/tiny/tiny.ccg

This uses the DotCCG format that is not especially well supported at this point. Here's a paper with the full details on it:

http://web.stanford.edu/group/cslipublications/cslipublications/GEAF/2007/papers/geaf07baldridgeetal.pdf

I mainly used DotCCG for teaching at the University of Texas at Austin, and folks using OpenCCG to do real parsing and realization work mostly use the XML formats for grammars.

tanin47:
Thank you @jasonbaldridge!

Issue #1330, opened 2016-03-13
BLEU Issues


alvations:
The BLEU implementation has sort of generated quite a lot of "wack-a-mole" situations where the fringe cases of using BLEU cause errors or unexpected output values from the following issues: #1328 #1288 #1285 #1268 #775 #789 #922 #926 #989 #1172 #1241

To resolve the issues, possibly our best chance is to emulate [`mteval-13a.perl`](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl) from @moses-smt. 

As a best practice, it should accompanied with a lot more test cases. 

**Any takers on writing/thinking of test cases?**

Contributors can also try to writing wrapper code or port code to NLTK from the following libraries:
- https://github.com/odashi/mteval
- https://github.com/moses-smt/mosesdecoder/blob/master/mert/sentence-bleu.cpp

alvations:
The "mini-roadmap" would be to:
- #1331: Revert code using `fractions.Fraction._normalize` (from #1319) 
- #1332: Create some sort of `nltk.compat` object to handle `_normalize` parameter in fractions.Fraction for Python < 3.5.
- #1332: Put back the `_normalize` patches in #1319 
- #1332: Write tests with fringe cases 
- #1383 Validate these fringe cases by changing the BLEU implementation accordingly
- #1477 Debug implementation of Chen and Cherry smoothing Method 6. 
- #1488 ~~Write a wrapper for or port `mteval-13a.pl`~~ Read `mteval-13a.pl` outputs and write tests to assert that our implementation corresponds to `mteval-13a.pl`'s implementation.

alvations:
After the mini-roadmap, a wrapper to Moses MERT's BLEU would be great since it's top of the "state-of-art" list in terms of human correlations score as reported in https://www.computing.dcu.ie/~ygraham/graham-acl15.pdf

stevenbird:
+1 to this: "Write a wrapper for or port mteval-13a.pl and write tests to assert that our implementation corresponds to mteval-13a.pl's implementation."

A shortcut would be to save output from `mteval-13a.pl` and test against that.

alvations:
While debugging why Chen and Cherry's method 6 didn't work as expected, it seems like we had an error in the `p_n` indexing at https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L538.

[...]

PR #1331, opened 2016-03-13
Reverting to bleu scores with normalized Fraction

alvations:
This is to revert the changes made in #1319. The `_normalize` parameter in the `Fraction` type will only work in Python>=3.5 and NLTK supports >=2.7. This PR should resolve the doctests and BLEU related issues on CI jenkins. 

Additionally, reduce the `sentence_bleu` to an instance of `corpus_bleu` where there's only one hypothesis and its references. This way, we only need to maintain one function. 

Also refractored the functions starting with underscores since these function are not "temporary" and they will be used by function MT evaluation metrics that are derived from BLEU.

This is the first of several PR to resolve issues with BLEU listed on #1330

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -79,6 +79,9 @@ class TestBLEU(unittest.TestCase):
         assert (round(hyp1_bigram_precision, 4) == 0.5882)
         assert (round(hyp2_bigram_precision, 4) == 0.0769)
         
+    def test_fringe_cases(self):
+        pass
+        
     def test_brevity_penalty(self):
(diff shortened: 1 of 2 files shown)


alvations:
@stevenbird could you help to review the code and then merge and check that it resolves the BLEU related issues on CI? Thanks in advance!

After that, I'll go on to the next step to patch #1330

stevenbird:
Looks good, thanks @alvations

stevenbird:
The CI server is much happier now, thanks

alvations:
It still needs #1332 for the implementation to be accurate though, the `_normalize` feature is rather important.

PR #1333, opened 2016-03-14
Refactored text.concordance to return list 

story645:
Refactored text.concordance so that it returns a list (while still retaining the print functionality). Written so that the list is optional and the old behavior is default, but I think it might make more sense to do this the other way around (list of concordances by default, print on request). 

The unittests and doctest are based on the behavior described in http://www.nltk.org/book/ch01.html
Doctest relies on ellipses because of invisible character issues. 

Would love more test cases.


story645:
Hi, just want to follow up with this. Just wondering what I could do to improve the PR. Thank you

stevenbird:
@story645 – the issue is a design one – the `Text` module was intended to demonstrate functionality that individuals might like to develop, not to provide functionality. Otherwise, methods would have returned their results and we would have left it to users to print. Modifying the behaviour of the concordance functionality would lead to divergence with the other functionality in this module.

I think a preferable solution would be to create a separate concordance module, and to modify the text module to use it. Then the text module retains its status as a demonstration. A text-based concordance module could support a similar interface to the graphical one that we already have (`concordance_app.py`). What do you think?

story645:
Hmm, I mean, I dunno that your users know that as I stumbled on this because a linguist was teaching Text Analysis using nltk and using the Text module & concordance to do it as that's what she used to do her analysis...and there's loads on google about people essentially rewriting concordance to print lists...

I'm seeing your point about essentially factoring out concordance altogether and then letting text use it in the way it always have-at least I I think that's what you're saying? ... but now I also want to propose a subclass of Text (TextA or something) that mirrors A but provides full functionality since that's what people are doing with it anyway...

AUTHORS.md
```diff
@@ -192,9 +192,13 @@
 - Will Monroe
 - Elijah Rippeth
 - Emil Manukyan
+<<<<<<< HEAD
```
dimazest: Could you please clean up the result of a merge.

story645: fixed, sorry about that. And I tried to rebase but I think my branch is way too enmeshed with develop at this point.

stevenbird:
@story645: it looks like you've changed the interface. Previously, we wrote:

```
>>> text1.print_concordance(word, width, lines)
```

And now to get the same behaviour, one has to write:

```
>>> text1.concordance(word, width)
>>> text1.print_concordance(lines)
```

Instead of changing the interface, can preserve the existing behaviour, and just have `print_concordance` as a wrapper for `concordance`?

story645:
How did that work, since `print_concordance` isn't a method on text:
https://github.com/nltk/nltk/blob/develop/nltk/text.py#L174

Unless this behavior is on a different branch?
ETA: I don't think I've changed the interface, as the example in the book:

```
text.concordance(word, width, lines)
```

works the same as always.

stevenbird:
Ah, I see, sorry, too rushed. I'll try to get back to this soon.

story645:
If API consistency is important, I can create a wrapper function in `ConcordanceIndex` that maintains the old api of  the `ConcordanceIndex.print_concordance` method and then make the new methods `_print_concordance` and `_list_concordance`

stevenbird:
@story645: sorry for the long delay – I've been doing fieldwork and no-one else has had time to respond further either. Each time I look at the [main diff](https://github.com/nltk/nltk/pull/1333/commits/1896401068f115ae120157d9d76592024fc47de6) I get confused about what was done.

If you have time to do some more work on this I'll try to get it included in the next release in a week or so.

For a minor release, we need to maintain the existing interface. So all we can do is add new functionality and refactor things to avoid duplicating code. Note that we can't have flags like this: `print_concordance=True` that change the signature of the method.

story645:
Hmm,  I'm adding new functionality though with the ability to return a list. Is a new kwarg a change in signature?

stevenbird:
It changes the return type, right?

story645:
Yup. But I wanted a way to support the existing functionality of printing too. (The current function doesn't return anything, just prints). I guess I could have it print + return a list, but that seems unhelpful. Unless it's okay to change the return type?

nltk/text.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, tokens, key=lambda x:x):
         self._offsets = defaultdict(list)
         """Dictionary mapping words (or keys) to lists of offset
            indices."""
-
+
```
dimazest: Could you please clean up whitespaces.

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,92 @@
...
+            of Whale - Bones ; for Whales of a monstrous size are oftentimes cast up dead u"""
+        
+        with stdout_redirect(StringIO()) as stdout:
+            self.text.concordance(self.query, stdout=True)
+        
+        clean = lambda rstr: rstr.lower().split()
+        
+        self.assertListEqual(clean(print_out), 
+                             clean(stdout.getvalue()))
+
```
dimazest: The empy line in the end of the file are redundant.

dimazest (commented):
Looks good, just a couple of formatting things.

alvations:
@story645 Could you help to make the changes that @dimazest suggested and do a fresh git pull from the latest develop branch? 

After that the automated test would kick in and possibly we can do a last bit of quick check and merge this =)

story645:
Sorry this dropped off my radar. Made the changes and added some more tests and cleanups, but I didn't really address @stevenbird's concern about return types.

nltk/text.py
```diff
@@ -322,18 +341,37 @@ def __len__(self):
...
+        :param lines: The number of lines to display (default=25)
+-       :type lines: int
+        :param stdout: print concordance [True] or return list [False]
+        :type stdout: bool
+
+        :return: concordance, either printed to stdout or as a list
+        :rtype: None, list
+
+        :param stdout: print concordance to stdout
+        :type stdout: bool
```
alvations: This is not a good variable name and it will conflict with `sys.stdout` if user imports using namespaces.

nltk/text.py
```diff
@@ -173,36 +171,57 @@ def __repr__(self):
         return '<ConcordanceIndex for %d tokens (%d types)>' % (
             len(self._tokens), len(self._offsets))
 
-    def print_concordance(self, word, width=75, lines=25):
-        """
-        Print a concordance for ``word`` with the specified context window.
+    def generate_concordance(self, word, width=80):
+        """Generate a concordance list of tuples(left, token, right)
+        for ``word`` with the specified context window.
```
alvations: ~Changing from print to generate might break e.g. from the books.~

nltk/text.py
```diff
@@ -173,36 +171,57 @@ def __repr__(self):
...
+        :rtype: list(str)
+        :return: list concordances as strings
+        :param length:  The number of concordances to return (default=None)
+        :type length: int
+        """
+        return [" ".join([left, token, right]) for
+                (left, token, right) in self._concordance[:lines]]
+
+    def print_concordance(self, lines=25):
+        """
```
alvations: hmmm. having print and generate existing together is a little spaghetti code like. Maybe using some sort of parameter like @stevenbird suggested might be a better way to go. 

@story645 Would you like to give it a try to redesign the code? Or would you prefer that I try to fork your branch and see whether we can redesign the code together?

nltk/text.py
```diff
@@ -322,18 +341,37 @@ def __len__(self):
...
+            # print("Building index...")
             self._concordance_index = ConcordanceIndex(self.tokens,
                                                        key=lambda s:s.lower())
+        self._concordance_index.generate_concordance(word, width)
 
-        self._concordance_index.print_concordance(word, width, lines)
+        if stdout:
+            self._concordance_index.print_concordance(lines)
+        else:
+            return self._concordance_index.list_concordance(lines)
```
alvations: L365-L374 shows the effect of the spaghetti code where lots if-else is required to patch how we control what is generated and what is printed. 

Let's try and work out the better way of adding a generate vs print functionality as a parameter to the function than keeping multiple functions in the codebase.

alvations:
@story645 Thanks for pulling the latest branch and merging the code. 

I've left some code review, do you want to give it a try to modify the code given the comments? They are mostly similar to what @stevenbird talked about with the parameters and return types. 

Or would you prefer some hand-holding to redesign the function?

nltk/text.py
alvations: To be fixed.

nltk/text.py
alvations: `to_print`/`to_generate` might be a possible choice.

nltk/text.py
alvations: Clean up on return types is needed.

alvations (dismissed):

nltk/text.py
story645: I'm kinda confused. generate_concordance is just the name I gave the functionality that was inside [print_concordance](https://github.com/nltk/nltk/blob/develop/nltk/text.py#L176) because it needs to get pulled out to be used for both printing and making the list.

nltk/text.py
story645: I'm fine changing it 'cause your library and all, and I honestly agree that the kwarg is clunky, but seriously it's one `if else` block based on a kwarg. L365 was [always in the code](https://github.com/nltk/nltk/blob/develop/nltk/text.py#L325)

nltk/text.py
alvations: Let me try to see whether we can come up with a code design for this and make it better =)

But I'll have to do this after work hours though (~7-8 hours later)

nltk/text.py
alvations: Yeah, sorry about that. We try to change It in this PR (if possible).

story645:
It'd be really helpful if you or @stevenbird gave some guidance on what this function should return. In the latest commit,  I made generate_concordance explicitly private (`__concordance__`) since it's an internal function (and really needs a unit test...) and removed concordance_list and just moved that bit of prettying up to text.concordance. 

Currently it's now returning a list and printing by default...but it's easily changeable.

alvations:
[CI: retest]

alvations:
@story645 sorry for coming back on this after several days. I've tried several times to redesign the code but it's a little tricky because of (i) the "print" word in function names and (ii) we need to check whether the changes made breaks anything else.

So first we have to check the `ConcordanceIndex` object and see whether it breaks anything in the user space. The nice thing is `ConcordanceIndex` is shielded from the users ;P

```
~/git-stuff/nltk$ grep -n -r "ConcordanceIndex" *
Binary file nltk/__pycache__/text.cpython-35.pyc matches
nltk/text.py:121:class ConcordanceIndex(object):
nltk/text.py:173:        return '<ConcordanceIndex for %d tokens (%d types)>' % (
nltk/text.py:329:        :seealso: ``ConcordanceIndex``
nltk/text.py:333:            self._concordance_index = ConcordanceIndex(self.tokens,
nltk/text.py:616:           "ConcordanceIndex",
Binary file nltk/text.pyc matches
```

So most of our changes made to the `Text.concordance()` function should be made in the `ConcordanceIndex` object which you did so =)

To incorporate what @stevenbird commented on using the return types, we might need to change some function name

On the `Text` object, we can try to change `print_concordance` function of the `ConcordanceIndex` object to `show_concordance` that takes in the `print_out` parameter, something like this:

```python
class Text:
    ... 
    def concordance(self, word, width=79, line=25, print_out=True):
        # Builds an index if it's not previously built.
        if '_concordance_index' not in self.__dict__:
             self._concordance_index = ConcordanceIndex(self.tokens,
                                                       key=lambda s:s.lower())
        return self._concordance_index.show_concordance(word, width, lines, print_out)
```

To double check that nothing breaks the function name change, only `Text.concordance` uses it, so it's safe: 

```
$ grep -n -r "print_concordance" *
Binary file nltk/__pycache__/text.cpython-35.pyc matches
nltk/text.py:176:    def print_concordance(self, word, width=75, lines=25):
nltk/text.py:336:        self._concordance_index.print_concordance(word, width, lines)
Binary file nltk/text.pyc matches
```

On more point of check is to check against the NLTK book, https://github.com/nltk/nltk_book/search?utf8=✓&q=print_concordance&type= . Now it's safe to make the changes, we refactor the `ConcordanceIndex.print_concordance` function to `ConcordanceIndex.show_concordance` or you could also keep the name `ConcordanceIndex.__concordance__`.

We let `show_concordance` always return the list of concordance as the default but prints when the `print_out` parameter is True, i.e.

```python
class ConcordanceIndex:
    ...
    def show_concordance(self, word, width=75, lines=25, to_print=True):
        """
        Show a concordance for ``word`` with the specified context window.

        :param word: The target word
        :type word: str
        :param width: The width of each line, in characters (default=80)
        :type width: int
        :param lines: The number of lines to display (default=25)
        :type lines: int
        :return: The lines of concordances.
        :rtype: list(tuple(str))
        """
        half_width = (width - len(word) - 2) // 2
        context = width // 4 # approx number of words of context

        offsets = self.offsets(word)
        header_line = "Displaying %s of %s matches:" % (lines, len(offsets))
        concordance_lines = []
        if offsets: # Collects the lines of concordances.
            lines = min(lines, len(offsets))
            for i in offsets:
                if lines <= 0:
                    break
                left = (' ' * half_width +
                        ' '.join(self._tokens[i-context:i]))
                right = ' '.join(self._tokens[i+1:i+context])
                left = left[-half_width:]
                right = right[:half_width]
                concordance_lines.append(left, self._tokens[i], right)
                lines -= 1
        else:
            header_line = "No matches"
        if to_print:
            print('\n'.join([header_line] + concordance_lines))
        return concordance_lines
```

In effect, calling `Text.concordance(word)` with the above changes should print the concordance and to get the list of concordance, users can easily do `word_concordance = Text.concordance(word)`. 

@story645 do you agree with the proposed changes?

@stevenbird is this okay?

alvations:
@story645 if @stevenbird okays the changes proposed in the previous comments, could you do a fresh pull-merge off latest develop branch, make the changes proposed? And we'll let the server checks run before we do a last quick check on this PR =)

story645:
Honestly, I think my approach of factoring the internal concordance  building  into `__concordance_` and out of the presentation layer is more robust because it means those internals can be tested-yes those tests don't yet exists-and possibly refactored for efficiency without affecting the presentation layer.  

`On more point of check is to check against the NLTK book,`

Which is why I wrote tests against the book... the failing tests are because I didn't update the doctests, which I'll do now. 


`In effect, calling Text.concordance(word) with the above changes should print the concordance and to get the list of concordance, users can easily do word_concordance = Text.concordance(word).`

That's what my last PR does...

alvations:
@story645 Thanks for the clarification. I've no strong objections to separating concordance generating and printing. I guess @stevenbird has the say on this =)

I'll re-review the code again outside of the code design.

nltk/text.py
```diff
@@ -173,36 +171,47 @@ def __repr__(self):
...
-            lines = min(lines, len(offsets))
-            print("Displaying %s of %s matches:" % (lines, len(offsets)))
             for i in offsets:
-                if lines <= 0:
-                    break
                 left = (' ' * half_width +
-                        ' '.join(self._tokens[i-context:i]))
-                right = ' '.join(self._tokens[i+1:i+context])
+                        ' '.join(self._tokens[i - context:i]))
+                right = ' '.join(self._tokens[i + 1:i + context])
```
alvations: PEP8 pet peeves on operators in slices https://www.python.org/dev/peps/pep-0008/#id27 =)

Avoid spaces in slices.

nltk/text.py
```diff
@@ -173,36 +171,47 @@ def __repr__(self):
         return '<ConcordanceIndex for %d tokens (%d types)>' % (
             len(self._tokens), len(self._offsets))
 
-    def print_concordance(self, word, width=75, lines=25):
-        """
-        Print a concordance for ``word`` with the specified context window.
+    def __concordance__(self, word, width=80):
+        """Generate a concordance list of tuples(left, token, right)
```
alvations: Dunder functions are little pricky. Try not to create new ones that are non-canonical call them something else, e.g. `generate_concordance()` or anything without a dunder.

See @amontalenti guide on http://www.pixelmonkey.org/2013/04/11/python-double-under-double-wonder

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,114 @@
...
+
+@contextlib.contextmanager
+def stdout_redirect(where):
+    sys.stdout = where
+    try:
+        yield where
+    finally:
+        sys.stdout = sys.__stdout__
+
```
alvations, line 26: Actually, this context manager can be avoided using the `file` parameter in Python3 style `print_function`. Use `from __future__ import print_function` and 

```
print(somestr, file=somewhere)  # prints to *somewhere*
print(somestr, file=sys.stdout)  # default setting prints to stdout.
```

See https://docs.python.org/3/library/functions.html#print

alvations: My mistake. You're capturing the printed string. So this is okay.

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,114 @@
...
+                    "they might scout at Moby Dick as a monstrous fable , or still worse and more de",
+                    "th of Radney .'\" CHAPTER 55 Of the monstrous Pictures of Whales . I shall ere l",
+                    "ing Scenes . In connexion with the monstrous pictures of whales , I am strongly",
+                    "ere to enter upon those still more monstrous stories of them which are to be fo",
+                    "ght have been rummaged out of this monstrous cabinet there is no telling . But",
+                    "of Whale - Bones ; for Whales of a monstrous size are oftentimes cast up dead u"]
+
+        concordance_out = self.text.concordance(self.query, print_out=False)
+        self.assertListEqual(clean_list(list_out), clean_list(concordance_out))
+        return
```
alvations: I think there's no need to return on the unit tests unless there's other meta test checking for outputs of the individual tests later on.

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_concordance_width(self):
+        list_out = ["monstrous", "monstrous", "monstrous",
+                    "monstrous", "monstrous", "monstrous",
+                    "monstrous", "monstrous", "monstrous",
+                    "monstrous", "monstrous"]
+
+        concordance_out = self.text.concordance(self.query, width=0,
+                                                print_out=False)
+        self.assertListEqual(clean_list(list_out), clean_list(concordance_out))
+        return
```
alvations: Similar to above, unnecessary `return`.

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_concordance_lines(self):
+        list_out = ["ong the former , one was of a most monstrous size . ... This came towards us ,",
+                    "ON OF THE PSALMS . \" Touching that monstrous bulk of the whale or ork we have r",
+                    "ll over with a heathenish array of monstrous clubs and spears . Some were thick",]
+
+        concordance_out = self.text.concordance(self.query, lines=3,
+                                                print_out=False)
+
+        self.assertListEqual(clean_list(list_out), clean_list(concordance_out))
+        return
```
alvations: Similar to above, unnecessary `return`.

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,114 @@
...
+            ing Scenes . In connexion with the monstrous pictures of whales , I am strongly
+            ere to enter upon those still more monstrous stories of them which are to be fo
+            ght have been rummaged out of this monstrous cabinet there is no telling . But
+            of Whale - Bones ; for Whales of a monstrous size are oftentimes cast up dead u
+            """
+
+        with stdout_redirect(StringIO()) as stdout:
+            self.text.concordance(self.query, print_out=True)
+        self.assertEqual(clean(print_out), clean(stdout.getvalue()))
+        return
```
alvations: Similar to above, unnecessary `return`.

nltk/test/unit/test_concordance.py
```diff
@@ -0,0 +1,114 @@
...
+
+def clean(raw_str):
+    """Remove spaces and case"""
+    return raw_str.lower().replace(" ", "")
+
+
+def clean_list(input):
+    """Removes spaces and case in list elements"""
+    return [clean(raw_str) for raw_str in input]
+
```
alvations: Is there a motivation of the `clean` and `clean_list` function? Wouldn't checking the string values for equality work too?

alvations (changes requested):

nltk/test/unit/test_concordance.py
story645: invisible white space, inconsistent new lines, and the like. Will try to refactor into something that doesn't need cleaning.

story645:
test failure doesn't seem to have to do with my code:
```
Failed doctest test for nltk.tag
  File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py35-jenkins/jdk/jdk8latestOnlineInstall/nltk/tag/__init__.py", line 8, in tag

File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py35-jenkins/jdk/jdk8latestOnlineInstall/nltk/tag/__init__.py", line 54, in nltk.tag
Failed example:
    tagger.evaluate(brown.tagged_sents(categories='news')[500:600])
Expected:
    0.73...
Got:
    0.7401981904351572

Stacktrace

  File "/scratch/jenkins/python/python-3.5.0-x86_64/lib/python3.5/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/scratch/jenkins/python/python-3.5.0-x86_64/lib/python3.5/unittest/case.py", line 597, in run
    testMethod()
  File "/scratch/jenkins/python/python-3.5.0-x86_64/lib/python3.5/doctest.py", line 2189, in runTest
AssertionError: Failed doctest test for nltk.tag
  File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py35-jenkins/jdk/jdk8latestOnlineInstall/nltk/tag/__init__.py", line 8, in tag

File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py35-jenkins/jdk/jdk8latestOnlineInstall/nltk/tag/__init__.py", line 54, in nltk.tag
Failed example:
    tagger.evaluate(brown.tagged_sents(categories='news')[500:600])
Expected:
    0.73...
Got:
    0.7401981904351572
```

alvations:
@story645 thanks for the updated code! Don't worry about the error, I think it's some strange staff from floating point rounding.

alvations:
[CI: retest]

alvations (approved):

alvations:
@story645 Now let's wait for @stevenbird to get back on this issue and we should be good to merge if he's okay with the changes to ConcordanceIndex =)

stevenbird:
@story645 – thanks for pursuing this. I'm happy with what you're proposing for `ConcordanceIndex`, but not happy with `Text.concordance` which optionally prints its result in addition to returning it. All of the methods in `Text` are designed to print their results, merely to demonstrate functionality that people could create for themselves; here you're proposing to change just one of them, cf my [earlier comment](https://github.com/nltk/nltk/pull/1333#issuecomment-227161507).

I didn't follow your earlier suggestion to subclass `Text`. I agree one could create a similar class which had many of the same methods, but which returned them. I think that would be overkill (and possibly confusing). I still think that a separate concordance module is the way to go. Then `Text.concordance` can just use it.

story645:
>I still think that a separate concordance module is the way to go. Then Text.concordance can just use it.

I'm confused by what you mean by this. Can you please give me an example/stub? My concern with a separate module was that I wrote this code precisely because of a complaint I was getting from people using text.concordance and so shuffling this functionality off to a different place would not solve the issue it was meant to solve. But maybe that's solvable in docs? 

The last refactor prints and returns a list, and I could make it so that it always prints and the list is the optional return - or that neither is optional.

stevenbird:
@story645 – the text module already contains examples of this pattern, e.g. see how it uses the metrics and collocations functionality.

story645:
Ok, I kinda get this now, I think...but then do I factor ConcordanceIndex into the new concordance module (either leaving a stub or depreciation warning) or replicate most of its behavior in the new module?

stevenbird:
@story645 – you could move `ConcordanceIndex` into the new module, and then import it from there back into the text module.

alvations:
@story645 Did you get a chance to refactor the `ConcordanceIndex`? Would you require more guidance on the refactoring?

story645:
Sorry, been really busy with school work and the like and hadn't had a chance to get back to this yet.

alvations (changes requested):
Please merge https://github.com/story645/nltk/pull/1 so that we can move towards merging this PR =)

alvations (approved):

alvations (changes requested):
@story645 please merge the pull-request made to your repo from me.

stevenbird:
Thanks @alvations, @story645 for helping resolve this long-standing issue.

PR #1341, opened 2016-03-23
CHILDES corpus reader: lazy loading (fixes #1340)

### LINKED ISSUE #1340: Problem with lazy corpus loading in CHILDES?
I would expect the following to display the beginning of the list fairly quickly:

``` py
>>> from nltk.corpus.reader.childes import CHILDESCorpusReader
>>> childes = CHILDESCorpusReader('/Users/nschneid/nltk_data/corpora/childes/data-xml/Eng-USA', '.*.xml')
>>> childes.tagged_words()
```

But it hangs, suggesting that it's trying to load the entire corpus. Is there a more efficient way to implement this?

nschneid:
This uses lazy loading by default in `words()`, `tagged_words()`, `sents()`, `tagged_sents()`, `participants()`, `age()`, and `MLU()`. Respects the setting in `self._lazy`.

Not sure if this is the most elegant solution, but it seems to work.

### DIFF
--- a/nltk/corpus/reader/childes.py
+++ b/nltk/corpus/reader/childes.py
@@ -16,7 +16,7 @@ __docformat__ = 'epytext en'
 import re
 from collections import defaultdict
 
-from nltk.util import flatten
+from nltk.util import flatten, LazyMap, LazyConcatenation
 from nltk.compat import string_types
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Looks good to me, thanks @nschneid

PR #1343, opened 2016-03-25
Ensure that proper exceptions are caught correctly

pombredanne:


stevenbird:
Thanks @pombredanne

pombredanne:
You are welcomed. That was an easy one.

PR #1346, opened 2016-03-27
Fix parsing of Chinese models output from StanfordPOSTagger

### LINKED ISSUE #1345: Problem with parsing Chinese models output from StanfordPOSTagger
The parser for the StanfordPOSTagger relies on a Separator to receive the output.

This is usually an underscore ('_'), but the output for the Chinese models for some reason uses '#'. This can be fixed by resetting the _SEPARATOR attribute before using the tagger.

```
> english_postagger.tag('What is the airspeed of an unladen swallow ?'.split())

[('What', 'WP'),
 ('is', 'VBZ'),
 ('the', 'DT'),
 ('airspeed', 'NN'),
 ('of', 'IN'),
 ('an', 'DT'),
 ('unladen', 'JJ'),
 ('swallow', 'VB'),
 ('?', '.')]

> chinese_postagger.tag('骑 自行车 那 个 人 是 谁 ？'.split())

[('', '骑#VV'),
 ('', '自行车#NN'),
 ('', '那#DT'),
 ('', '个#M'),
 ('', '人#NN'),
 ('', '是#VC'),
 ('', '谁#PN'),
 ('', '？#PU')]

> print ('Default separator:', chinese_postagger._SEPARATOR)

Default separator: _

> chinese_postagger._SEPARATOR = '#'
> chinese_postagger.tag('骑 自行车 那 个 人 是 谁 ？'.split())

[('骑', 'VV'),
 ('自行车', 'NN'),
 ('那', 'DT'),
 ('个', 'M'),
 ('人', 'NN'),
 ('是', 'VC'),
 ('谁', 'PN'),
 ('？', 'PU')]
```

Setup like this:

```
from nltk.tag import StanfordPOSTagger as POSTagger
from nltk.internals import find_jars_within_path

def fix_path(path):
    """Fix the jar-path for the tagger._stanford_jar.

    See: https://github.com/nltk/nltk/issues/1304
    """
    stanford_dir = path.rpartition('/')[0]
    stanford_jars = find_jars_within_path(stanford_dir)
    string_jars = ':'.join(stanford_jars)
    return string_jars

tagger_dir = HOME_DIR + 'nlp/stanford-postagger-full-2015-12-09/'
tagger_jar     = tagger_dir + 'stanford-postagger.jar'
english_tagger = tagger_dir + 'models/english-bidirectional-distsim.tagger'
# Same result for both Chinese models:
chinese_tagger = tagger_dir + 'models/chinese-distsim.tagger'
#chinese_tagger = tagger_dir + 'models/chinese-nodistsim.tagger'

english_postagger = POSTagger(english_tagger, tagger_jar)
chinese_postagger = POSTagger(chinese_tagger, tagger_jar)

english_postagger._stanford_jar = fix_path(english_postagger._stanford_jar)
chinese_postagger._stanford_jar = fix_path(chinese_postagger._stanford_jar)
```

casperlehmann:
Fix #1345

### DIFF
--- a/nltk/tag/stanford.py
+++ b/nltk/tag/stanford.py
@@ -133,6 +133,11 @@ class StanfordPOSTagger(StanfordTagger):
     _JAR = 'stanford-postagger.jar'
 
     def __init__(self, *args, **kwargs):
+        # Standford POSTagger uses '#' as separator.
+        if ('chinese-distsim.tagger' in args[0] or
+            'chinese-nodistsim.tagger' in args[0]) and \
+           'stanford-postagger.jar' in args[1]:
+                self._SEPARATOR = '#'
         super(StanfordPOSTagger, self).__init__(*args, **kwargs)
 
     @property
--- a/nltk/tag/stanford.py
+++ b/nltk/tag/stanford.py
@@ -133,7 +133,7 @@ class StanfordPOSTagger(StanfordTagger):
     _JAR = 'stanford-postagger.jar'
 
     def __init__(self, *args, **kwargs):
-        # Standford POSTagger uses '#' as separator.
+        # Standford POSTagger uses '#' as separator for Chinese texts.
         if ('chinese-distsim.tagger' in args[0] or
             'chinese-nodistsim.tagger' in args[0]) and \
            'stanford-postagger.jar' in args[1]:


nltk/tag/stanford.py
```diff
@@ -133,6 +133,11 @@ class StanfordPOSTagger(StanfordTagger):
     _JAR = 'stanford-postagger.jar'
 
     def __init__(self, *args, **kwargs):
+        # Standford POSTagger uses '#' as separator for Chinese texts.
+        if ('chinese-distsim.tagger' in args[0] or
+            'chinese-nodistsim.tagger' in args[0]) and \
+           'stanford-postagger.jar' in args[1]:
```
alvations, line 139: Try to use explicit variable names instead of `args[0]`

This if statement can be simplified to:

``` python
if os.path.split(self.model_filename)[1].startswith('chinese-'):
    self._SEPARATOR = '#'
```

The `os.path.split()[1]` is necessary to ensure that we only get the filename. If environment variable is set the `if self.model_filename.startswith('chinese-')` would be sufficient but we should keep it such that we allow full path without environ variables, so we use `os.path.split()`

The assumption that `StanfordPOSTagger` is initialized should assume that `'stanford-postagger.jar' in self.path_to_jar` so that is not needed.

alvations:
Personally, I would prefer to use `self._separator='#'` without changing the code and report the problem to @stanfordnlp so that they standardize their separator upstream. 

```
chinese_postagger = POSTagger(chinese_tagger, tagger_jar)
chinese_postagger._SEPARATOR = '#'
chinese_postagger.tag(u'骑 自行车 那 个 人 是 谁 ？'.split())
[(u'\u9a91', u'VV'), (u'\u81ea\u884c\u8f66', u'NN'), (u'\u90a3', u'DT'), (u'\u4e2a', u'M'), (u'\u4eba', u'NN'), (u'\u662f', u'VC'), (u'\u8c01', u'PN'), (u'\uff1f', u'PU')]
```

But I think it's fine to have the separator language dependent.

---

One thing to note, the `find_jars_within_dir` "hack" is already incorporated into NLTK 3.2, maybe upgrading your NLTK would simplify your code =)

casperlehmann:
I'm not a fan of setting self._SEPARATOR outside of the class. I think that should be avoided to keep the code pythonic.

I slept on it, and came to the conclusion that it should probably be set as an optional parameter at initialization, since there might be all sports of reasons to choose a specific separator, model side, and we would want to be able to adapt to that in production without having to resort to patching the source code.

How about this?

```
if kwargs.get('separator') is not None:
    self._SEPARATOR = kwargs.get('separator')
```

In any case, I'll send a wire to Stanford about their Chinese models.

alvations:
Using `kwargs` is a good solution. But possibly an easier fix is to pass the separator explicitly to Stanford with `-tagSeparator _` : 

``` python
class StanfordPOSTagger(StanfordTagger):
    _SEPARATOR = '_'
    _JAR = 'stanford-postagger.jar'

    def __init__(self, *args, **kwargs):
        super(StanfordPOSTagger, self).__init__(*args, **kwargs)

    @property
    def _cmd(self):
        return ['edu.stanford.nlp.tagger.maxent.MaxentTagger',
                '-model', self._stanford_model, '-textFile',
                self._input_file_path, '-tokenize', 'false',
                '-outputFormatOptions',  'keepEmptySentences', 
                '-tagSeparator', '_']
```

Note, some quirky behavior of how Stanford changes the token if `_` is in the token:

``` python
 >>> from nltk.tag import StanfordPOSTagger
>>> st = StanfordPOSTagger('chinese-distsim.tagger')
>>> st.tag(u'骑 自行车 那 个 人 是 谁 ？'.split())
[('骑', 'VV'), ('自行车', 'NN'), ('那', 'DT'), ('个', 'M'), ('人', 'NN'), ('是', 'VC'), ('谁', 'PN'), ('？', 'PU')]
>>> st.tag(u'骑 自行车 那 个 人 是 www.nltk.org ？'.split())
[('骑', 'VV'), ('自行车', 'NN'), ('那', 'DT'), ('个', 'M'), ('人', 'NN'), ('是', 'VC'), ('www.nltk.org', 'URL'), ('？', 'PU')]
>>> st.tag(u'骑 自行车 那 个 人 是 www.nltk_data.org ？'.split())
[('骑', 'VV'), ('自行车', 'NN'), ('那', 'DT'), ('个', 'M'), ('人', 'NN'), ('是', 'VC'), ('www.nltkdata.org', 'URL'), ('？', 'PU')]
>>> st.tag(u'骑 自行车 那 个 人 是 Michael_Jordan ？'.split())
[('骑', 'VV'), ('自行车', 'NN'), ('那', 'DT'), ('个', 'M'), ('人', 'NN'), ('是', 'VC'), ('MichaelJordan', 'NR'), ('？', 'PU')]
```

casperlehmann:
I didn't know that was an option, but that is of course a decent solution. I'd say, however, that we'd want to pass self._SEPARATOR in stead of hard coding the -tagSeparator, which in turn brings back the question of whether to allow users the option of setting it at initialization.

If we don't, any underscores in words will get removed during processing.

```
> chinese_postagger.tag(
...    '骑_Sentences_Containing_Underscores_Will_Get_Mangled 自行车 那 个 人 是 谁 ？'.split())
[('骑SentencesContainingUnderscoresWillGetMangled', 'VV'),
 ('自行车', 'NN'),
 ('那', 'DT'),
 ('个', 'M'),
 ('人', 'NN'),
 ('是', 'VC'),
 ('谁', 'PN'),
 ('？', 'PU')]
```

Alternatively, we could change the self.parse_output method to do an rsplit in stead of a split, to preserve the words on the left side.

```
def parse_output(self, text, sentences = None):
    # Output the tagged sentences
    tagged_sentences = []
    for tagged_sentence in text.strip().split("\n"):
        sentence = []
        for tagged_word in tagged_sentence.strip().split():

            #word_tags = tagged_word.strip().split(self._SEPARATOR)
            word_tags = tagged_word.strip().rsplit(self._SEPARATOR, 1)

            #sentence.append((''.join(word_tags[:-1]), word_tags[-1]))
            sentence.append((word_tags[0], word_tags[-1]))
```

In that case, manually setting self._SEPARATOR becomes redundant.

casperlehmann:
Submitted New PR #1348

PR #1347, opened 2016-03-27
address of free software foundation updated

Lightyagami1:
Address is now 51 Franklin St, Boston, MA 02110-1301. Please update


stevenbird:
Thanks @Lightyagami1

PR #1350, opened 2016-03-29
Stanseg

### LINKED ISSUE #1349: Stanford Word Segmenter
Issue #101 Concerned the lack of a Word Segmenter for Chinese Text. This issue relates specifically to the Stanford implementation.

There is some old code (2014) floating around for interfacing with the Stanford Word Segmenter, and it seems that quite a few people have implemented similar solutions over the years, but for some reason it has not yet been included in NLTK.

http://textminingonline.com/dive-into-nltk-part-vi-add-stanford-word-segmenter-interface-for-python-nltk

I see @alvations has already asked permission (in the comments) to include it in NLTK.

There are, however, a few problems, which I suppose result from the updated stanford-segmenter-3.6.0, and the use of abstract base class syntax in tokenize/api.py: https://github.com/nltk/nltk/commit/b40ebee0e1238f44984fb13dd1de6b9f77740b73.

casperlehmann:
slf4j-api.jar is required in the CLASSPATH, but it is not selected by default.
I have added a second call to nltk.internals.find_jar in order to add it to self._stanford_jar.
I also added a super().tokenize() call to deal with TokenizerI as an abstract class.

Initial commit is taken directly from source. Diff to last one for details.

This should fix #1349.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -192,3 +192,4 @@
 - Will Monroe
 - Elijah Rippeth
 - Emil Manukyan
+- Casper Lehmann-Strøm
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -2,8 +2,9 @@
 # -*- coding: utf-8 -*-
 # Natural Language Toolkit: Interface to the Stanford Chinese Segmenter
 #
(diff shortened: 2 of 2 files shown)


stevenbird:
Thanks @Laspimon – great to have this.

Issue #1342, opened 2016-03-25
Discussion: Resurrecting the Ngram Model


iliakur:
Hi folks!

I'm working on making sure the Ngram Model module could be added back into NLTK and would like to bring up a couple of issues for discussion.

**Issue 1**
[Here](https://github.com/nltk/nltk/issues/367#issuecomment-15716162) @afourney said it would be nice to add interpolation as an alternative to the default Katz backoff as a way of handling unseen ngrams. I've been thinking about this and I might have an idea how this could work. I'd like to run it by all interested parties.

The current class structure of the `model` module is as follows:
- `model.api.ModelI` -> this is supposed to be an Abstract class or an Interface, I guess.
- `model.ngram.NgramModel` -> extends above class, contains current implementation of the ngram model.

Here's what I propose:
- `model.api.Model` -> I'm honestly not sure I see the point of this, ambivalent on whether to keep it or ditch it.
- `model.ngram.BasicNgramModel` -> This is the same as `NgramModel`, **minus** everything that has to do with backoff. Basically, it can't handle ngrams unseen in training. "Why have this?" - you might ask. I think this would be a great demo of the need for backoff/interpolation. Students can simply try this out and see how badly it performs to convince themselves to use the other classes.
- `model.ngram.BackoffNgramModel` -> Inherits from `BasicNgramModel` to yield the current implementation of `NgramModel`, except that it's more explicit about the backoff part.
- `model.ngram.InterpolatedNgramModel` -> Also inherits from `BasicNgramModel`, but uses interpolation instead of backoff.

The long-term goals here are:

a) to allow any `ProbDist` class to be used as a probability estimator since interpolation/backoff are (mostly) agnostic of the smoothing algorithm being used.
b) to allow anyone who wants to _optimize_ an NgramModel for their own purposes to be able to easily inherit some useful defaults from the classes in NLTK.

**Issue 2**
Unfortunately the probability module has it's own problems (eg. #602 and (my) Kneser-Ney implementation is wonky). So for now I'm only testing correctness with `LidstoneProbDist`, since it is easy to compute by hand. Should I be worried about the lack of support for the more advanced smoothing methods? Or do we want to maybe proceed this way to ensure at least that Ngram Model works, and _then_ tackle the problematic probability classes separately?

**Python 3 `super()`**
When calling `super()`, do I need to worry about supporting python 2? See [this](http://stackoverflow.com/a/10483204/4501212) for context.

anttttti:
[...]

PR #1351, opened 2016-03-30
Bringing Language Modeling back

iliakur:
Based on the discussion in #1342  I think we've finally struck some decent ideas for how to fix the model package. I'm submitting here a patch that implements the ideas proposed in that thread.

Since I'm considerably changing how the `model` module works, please let me know if you'd like some examples of how I see this being used in the field.


nltk/model/counter.py
```diff
@@ -0,0 +1,99 @@
...
+    """Stores language model vocabulary.
+
+    Satisfies two common language modeling requirements for a vocabulary:
+    - When checking membership and calculating its size, filters items by comparing
+      their counts to a cutoff value.
+    - Adds 1 to its size so as to account for "unknown" tokens.
+    """
+
+    def __init__(self, unknown_cutoff, *counter_args):
+        Counter.__init__(self, *counter_args)
```
alvations, line 41: `super.__init__(...)` here would be appropriate =)

iliakur, line 41: Sure! To be honest, I got confused [by this thread](http://stackoverflow.com/a/5066411/4501212) as to whether `super()` is good or bad practice...
BTW, are we trying to maintain backward compatibility with Python 2.x where possible?

nltk/model/counter.py
```diff
@@ -0,0 +1,99 @@
...
+from collections import Counter, defaultdict
+from itertools import chain
+
+from nltk.util import ngrams
+from nltk.probability import FreqDist, ConditionalFreqDist
+from nltk import compat
+
+
+@compat.python_2_unicode_compatible
+class LanguageModelVocabulary(Counter):
```
ezubaric: Doesn't this also need a membership function that returns false for terms that don't match the frequency cutoff?

nltk/model/counter.py
```diff
@@ -0,0 +1,99 @@
...
+        return sum(1 for item in self if item in self) + 1
+
+
+@compat.python_2_unicode_compatible
+class NgramCounter(object):
+    """Class for counting ngrams"""
+
+    def __init__(self, highest_order, unknown_cutoff, training_text,
+                 start_symbol="<s>", end_symbol="</s>", unknown_label="<UNK>"):
+        """
```
alvations, line 79: It might be more intuitive if the `training_text` should be the first param.

```
def __init__(self, training_text, highest_order, unknown_cutoff,
                    start_symbol="<s>", end_symbol="</s>", unknown_label="<UNK>")
```

nltk/model/counter.py
```diff
@@ -0,0 +1,99 @@
...
+    - Adds 1 to its size so as to account for "unknown" tokens.
+    """
+
+    def __init__(self, unknown_cutoff, *counter_args):
+        Counter.__init__(self, *counter_args)
+        self.cutoff = unknown_cutoff
+
+    def __contains__(self, item):
+        """Only consider items with counts GE to cutoff as being in the vocabulary."""
+        return self[item] >= self.cutoff
```
iliakur, line 57: @ezubaric see here for cutoff being used for membership checking.

nltk/model/ngram.py
```diff
@@ -1,247 +1,72 @@
...
+    Currently untested.
+    """
 
-                # backoff total has to be less than one, otherwise we get
-                # an error when we try subtracting it from 1 in the denominator
-                assert (0 <= backoff_total_pr < 1), backoff_total_pr
-                alpha_ctxt = beta / (1.0 - backoff_total_pr)
+    def __init__(self, highest_order, unknown_cutoff, training_text,
+                 **ngram_counter_kwargs):
+        self.counter = NgramCounter(highest_order, unknown_cutoff, training_text,
```
alvations: Same intuitive initialization here should be used (spill over from the refactoring the `NgramCounter.__init__`)

```
 def __init__(self, training_text, highest_order, unknown_cutoff,
                  **ngram_counter_kwargs):
```

alvations:
@Copper-Head Maybe `self._normalize`, `norm_word` and `norm_context`, `normed_text` isn't a good name here, since we might confuse it with normalizing the probabilities. 

Maybe something like `self._patch_unknown`, `word_patch_unknown`, `context_patch_unknown`, `text_patch_unknown`? I'm bad at naming. Someone else should think of a better name =)

nltk/model/counter.py
ezubaric, line 57: How did I miss that?  Sorry!

nltk/model/counter.py
alvations, line 41: I would say, `super()` here makes sense otherwise, `class LanguageModelVocabulary(Counter)` would be redundant since we can do `class LanguageModelVocabulary(object)` and still do `Counter.__init__(self, *counter_args)`. But I think someone else from a more pythonic background should comment on this.

As for backwards compat, I guess so, NLTK still supports Python2.7, so we have to use old style `super()`, e.g. https://github.com/nltk/nltk/blob/develop/nltk/compat.py#L684 (`Fraction` is a type, so it uses `cls` instead of `self`).

iliakur:
Naming **is** hard and I whipped up that `MLENgramModel` class literally in 5 min just as a demo of how I see the `NgramCounter` being consumed. `NgramCounter`, `LanguageModelVocabulary`, and `NgramModelI` are the parts I would like focus on first and once we iron those out I'll revisit that variable naming and (I hope) add a couple more ngram model classes.

nltk/model/counter.py
iliakur, line 57: No worries! Revisiting that actually made me aware of a potential subtle bug: if someone (for whatever crazy reason) sets the cutoff to `0`, the vocabulary will silently start returning `True` even for words unseen during training.
I wonder if the best way to mitigate this is by making sure the value of `cutoff` is at least equal to 1 and raising some clear exception if that's not the case so as to stop the user in their tracks before things get weird.

alvations, line 57: Possibly, not-the-usual-usage fringe cases should be added to regression tests, e.g.
- `cutoff = 0`
- highest_order > max(len(sent) for sent in sents)
- empty `training_data`
- etc.

alvations:
@Copper-Head Using `from __future__ import division` should make the code cleaner too, but remember to use `//` instead of `/` if the division needs to output an integer. Looks a like a good start to a working `NgramModel` =)

nltk/model/counter.py
iliakur, line 57: I actually wonder if empty `training_data` may not be such a fringe case. Imagine this scenario:
We use different texts for creating the vocabulary (`vocab_data`) and for training (`training_data`). We'd like to be able to do something along the following lines:

``` python
>>> counter = NgramCounter(highest_order=3, unk_cutoff=2)
>>> counter.build_vocabulary(vocab_data)
>>> counter.train_counts(training_data)
```

ezubaric, line 57: @Copper-Head : That's a very common use case.  And to be able to train on multiple training sets without having to merge them together.  (e.g., add train_counts twice)

iliakur, line 57: Sounds like we should make `__init__`'s `training_data` argument optional (set it to `None` by default) and move all the vocabulary and training code into "public" methods (to the extent that concept applies to Python).

iliakur:
Folks, I've implemented the changes we discussed a couple of days ago. `NgramCounter` can now be initialized with no training data and vocab setting/ngram counting have been moved into separate methods.

iliakur:
After some thought I've come to the realization that it makes a lot of sense to completely de-couple vocabulary creation from ngram counting. For instance, imagine someone wants to test several _unknown word cutoff_ values and pick the one with the best results (whatever **that** means!). What they are in effect doing is using the same vocabulary but filtering its elements differently using the cutoff. I think it would be silly/inefficient for them to have to create a whole new vocabulary just to count ngrams with respect to a different cutoff value. Not to mention that really the input types for vocabulary training and ngram counting are slightly different.

The changes I've pushed make the following usage possible:

``` python
>>> from nltk.model import build_vocabulary, NgramCounter
>>> text1 = ["colorless", "green", "ideas", "sleep", "furiously"]
>>> text2 = "abcdef...ah, screw it!"
>>> vocab = build_vocabulary(2, text1, text2)  # specify as many texts as you like
>>> vocab
NgramModelVocabulary({"a": 1, "b": 1, "c": 2, "colorless": 1, ... "!": 1})
>>> from nltk.corpus import brown
>>> training_corpus = brown.sents()
>>> tri_counter = NgramCounter(3, vocab)
>>> tri_counter.count_ngrams(training_corpus)
>>> higher_cutoff_tri_counter = NgramCounter(3, vocab, cutoff=5)
>>> higher_cutoff_tri_counter.count_ngrams(training_corpus)
>>> # compare the two counters somehow...
```

I think this is an improvement over initial setup because it's more modular and thus easier to adapt to different workflows people might have.

At the same time I wonder if there's a way to also hide the NgramCounter class behind some easy-to-use function corresponding to `build_vocab`.

As always, thoughts on all this are appreciated!

stevenbird:
Thanks for this new work @Copper-Head. Shall I pull it into a new branch and encourage people to try it out? Then we might get new PRs against that branch...

iliakur:
@stevenbird I think with the last batch of commits we might be approaching the point where we can show this to testers.
Before we let them have at it, should I write up some doctests/docstrings? I haven't been very diligent about this stuff because the module is only just now starting to take on a stable form.
Obviously, in addition to documentation, I'll be closely monitoring github and the google group and try to answer any questions folks might field as they use the package.

nltk-webdeploy:
Yes, it would be great to have test code now, since that will help us deal
with contributions from others.

stevenbird:
I've merged this into the `model` branch, since this will surely help anyone who wants to test it. I've posted a new message to nltk-dev to encourage feedback. https://groups.google.com/forum/#!topic/nltk-dev/q_R_VaMKI9o

clayfink:
Looking at the model branch and there doesn't seem to be support for 1-gram models. NgramCounter creates a unigram FreqDist, but the various model classes don't support it. I could roll my own bit I'm looking for the 1-gram functionality that the older versions of nltk supported.

iliakur:
Hmm, let me take a quick look this weekend to see what needs to be adjusted in the model classes to make this work. I have to admit this usecase kinda slipped through the cracks when I was thinking about this initially :)

ezubaric:
This might be a rough corner case for some of the estimation/backoff methods. :)

iliakur:
Right. @clayfink, do keep in mind that most of the back-off/interpolation estimators don't make much sense to have with only unigrams. What kind of models are you primarily interested in?

PR #1352, opened 2016-03-31
Bigram fixer

melvyniandrag:
I have made a simple change which is to make bigram scoring depend on the FreqDist of bigrams, and not on the frequency of words, and added myself to the AUTHORS file. 

Ive run the tox for py34 and py27 and get a number of errors, but I get the same errors for both the code with my changes and the code without my changes. 

Also I wrote this simple test to make sure frequency distributions are calculated correctly:

```
from nltk.collocations import BigramCollocationFinder
from nltk.metrics.association import NgramAssocMeasures
import unittest
"""
Bigrams: 
    (Dogs, are)
    (are, animals)
    (animals, Giraffes)
    (Giraffes, are)
    (are, animals)
    (animals, too)
 """
class BigramTest(unittest.TestCase):
    def test_freq_dist1(self):
        text = 'Dogs are animals Giraffes are animals too'.split()
        bcf = BigramCollocationFinder.from_words(text)
        scorefn = NgramAssocMeasures.raw_freq
        tol = 1e-10
        val = bcf.score_ngram(scorefn, 'are', 'animals')
        desired_val = 1.0 / 3.0
        diff = abs(val - desired_val)
        self.assertTrue(diff < tol)
    def test_freq_dist2(self):
        text = 'Dogs are animals Giraffes are animals too'.split()
        bcf = BigramCollocationFinder.from_words(text)
        scorefn = NgramAssocMeasures.raw_freq
        tol = 1e-10
        val = bcf.score_ngram(scorefn, 'Dogs', 'are')
        desired_val = 1.0 / 6.0
        diff = abs(val - desired_val)
        self.assertTrue(diff < tol)

unittest.main()
```

### DIFF
--- a/nltk/collocations.py
+++ b/nltk/collocations.py
@@ -176,12 +176,14 @@ class BigramCollocationFinder(AbstractCollocationFinder):
         function.  Following Church and Hanks (1990), counts are scaled by
         a factor of 1/(window_size - 1).
         """
-        n_all = self.N
-        n_ii = self.ngram_fd[(w1, w2)] / (self.window_size - 1.0)
+        n_all = self.ngram_fd.N()
+        n_ii = self.ngram_fd[(w1, w2)]/(self.window_size - 1.0)
         if not n_ii:
             return
-        n_ix = self.word_fd[w1]
-        n_xi = self.word_fd[w2]
+        nix_list = [tup for tup in self.ngram_fd if tup[0] == w1]
+        nxi_list = [tup for tup in self.ngram_fd if tup[1] == w2] 
+        n_ix = sum([self.ngram_fd[tup] for tup in nix_list])
+        n_xi = sum([self.ngram_fd[tup] for tup in nxi_list])
         return score_fn(n_ii, (n_ix, n_xi), n_all)
 
 
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -49,6 +49,7 @@
 - Nikhil Dinesh
 - Liang Dong
 - David Doukhan
+- Melvyn Drag
 - Rebecca Dridan
 - Pablo Duboue
 - Long Duong


stevenbird:
@jothman do you agree that `n_all` should be based on the [number of bigrams rather than unigrams](https://github.com/nltk/nltk/pull/1352/commits/ceaaa949da4d307b4b9d214f88eb370c1039effd#diff-f88a44fe16a5ce656ebd7da67fa69dc3R179).

nltk/collocations.py
```diff
@@ -176,12 +176,14 @@ def score_ngram(self, score_fn, w1, w2):
         function.  Following Church and Hanks (1990), counts are scaled by
         a factor of 1/(window_size - 1).
         """
-        n_all = self.N
-        n_ii = self.ngram_fd[(w1, w2)] / (self.window_size - 1.0)
+        n_all = self.ngram_fd.N()
+        n_ii = self.ngram_fd[(w1, w2)]/(self.window_size - 1.0)
```
jnothman, line 180: Why this change?

nltk/collocations.py
```diff
@@ -176,12 +176,14 @@ def score_ngram(self, score_fn, w1, w2):
...
         if not n_ii:
             return
-        n_ix = self.word_fd[w1]
-        n_xi = self.word_fd[w2]
+        nix_list = [tup for tup in self.ngram_fd if tup[0] == w1]
```
jnothman, line 183: Stylistically, I'd much prefer to see the following than introduce another var:

``` python
        n_ix = sum(freq for (ng1, ng2), freq in iteritems(self.ngram_fd)
                   if ng == w1)
```

However, this code is irrelevant. The approach adds unjustifiable computational complexity. The incumbent approach is an approximation, one suggested by textbooks, etc., for efficiency; and yes there appear to be cases where it breaks.

Since a common use-case applies this function for a large quantity of n-grams, we need to get `n_ix` for a particular n-gram in O(1), or maybe at worst O(number of contexts for w1), but the proposed approach uses O(vocabulary) which is too much. So if we go down this route numerically, we need to pre-store `n_ix` and `n_xi`; or my preferred solution is to actually store the number (where greater than 0) that one should subtract from `word_fd[w1]` (`[w2]`) to get `n_ix` (`n_xi`); the data structure will be sparsely populated.

melvyniandrag:
Ok, will reconsider and try something else. I've seen this approximation in many papers (it seems that this is so accepted that authors don't even bother to use the 'approximately equal to' sign anymore and go right for the '='). So far I think that  it only breaks for very small texts, but there might be other cases as well. If it turns out that all users understand that they can't expect accurate results when finding collocations in a single sentence, then I guess it's okay. 

When I was experimenting with nltk I was working with very short bodies of text, and that's why this came up. I think the code should either alert the user if  self.N is too small that results might not be accurate, or maybe even have an if/else depending on self.N, so that the approximation is used only for sufficiently large bodies of text. The cost of the 'if' will be negligible.

I'll continue to see if there are other scenarios where the approximation breaks down. If anyone knows off the top of their head please leave a comment!

jnothman:
I think the issue of "small N gives bad results" is true of all statistical NLP. We should print it on the NLTK book cover.

Corpora of very small texts are a use-case of (increasing?) interest. Does this remain, statistically, an issue for a _corpus_ of tweets?

melvyniandrag:
I have a solution and will push in a few days after a couple more tests.

jnothman:
Great, thanks!

melvyniandrag:
Joel we wrote a short note about how to do this because so far as I can tell no one handles this situation yet and I'd like to get my name 'in the books' and I'd like to get more professionally involved in NLP. Corpora of short texts seems to be a fertile area for nlp. I'm planning on putting it on the arxiv or submitting it as a note to a journal. I think we should put your name on it too, so shoot me an email if you want to review this couple of paragraphs we threw together. The solution we have  takes a little reworking of how the BigramCollocationFinder works (or at least how I understand how it works.)
melvyniandrag@gmail.com

jnothman:
Hey @melvyniandrag, did you come up with a feasible solution?

PR #1353, opened 2016-04-01
Added a verification for empty tokens

gnardari:
An empty token makes the classifier crash. I've ran into this issue multiple times and this simple if statement fixes that.


stevenbird:
Thanks @guivn

PR #1358, opened 2016-04-10
Fixed bug: corpus_bleu was expecting Fractions from smoothing_function

almath123:
The `corpus_bleu` method was keeping the precisions as fractions until the overall precision for all n-grams was calculated. If a `smoothing_function` was used then the precisions were converted to floats before this; which caused a crash in the overall precision calculation. I have modified `corpus_bleu` to use floats after `smoothing_function` is called. This fixes the problem. Also added a doctest for this case.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -195,3 +195,4 @@
 - Casper Lehmann-Strøm
 - Andrew Giel
 - Tanin Na Nakorn
+- Alexander Mathews
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -66,6 +66,12 @@ def sentence_bleu(references, hypothesis, weights=(0.25, 0.25, 0.25, 0.25),
     >>> weights = (0.1666, 0.1666, 0.1666, 0.1666, 0.1666)
     >>> sentence_bleu([reference1, reference2, reference3], hypothesis1, weights)
     0.45838627164939455
+
+    The example below shows how to use a smoothing function:
+
+    >>> ref_short = ['It', 'is', 'a']
+    >>> sentence_bleu([ref_short], hypothesis1, smoothing_function = SmoothingFunction().method1) # doctest: +ELLIPSIS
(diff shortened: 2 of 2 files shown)


alvations:
@almath123 thanks for the attempt but the smoothing and float casting stems from a larger problem which comes from #1285 and #1330 

The `if-else` you have added is fine but I think the more appropriate solution would be to have a default smoothing to emulate `mteval-13a.pl`. But there hasn't been enough response as to how should `nltk` handle the fringe cases =(

I have also tried to poll on my twitter feed but it seems like it's still unclear how we should handle the default smoothing of BLEU, https://twitter.com/alvations/status/717647857768144896

alvations:
The patched up version that handles fringe cases from #1383 would resolve this issue. 

Thanks @almath123 , if you would like to, you can help to add more test cases for the BLEU code in https://github.com/nltk/nltk/blob/develop/nltk/test/unit/translate/test_bleu.py and when it breaks the code, you can use the `@unittest.skip("Skipping breaking cases for BLEU.")` decorators and do new PRs to BLEU related issues.

stevenbird:
Thanks @alvations for the guidance. @almath123 – would you have a chance to add some more test cases as suggested?

almath123:
Thanks @alvations the new fixes to the Bleu methods are awesome. You seem to have a good selection of tests too. I will see if there are any more I can add.

stevenbird:
@almath123 or @alvations – would you please let me know where this stands? Is it just waiting some more tests before merging?

alvations:
@stevenbird I think we should close this PR since #1383 resolves the problem. Unless @almath123 wants to pull-merge the current branch and writes more tests, in that case, i think it'll be better for him to open another PR.

almath123:
Yes this PR should be closed. As I understand it this issue was resolved by @alvations in a different PR. I can open another PR to contribute tests etc...

alvations:
Thanks @almath123, looking forward to more tests =)

PR #1360, opened 2016-04-11
framenet.py: remove unnecessary str() call that causes problems for names with Unicode characters

### LINKED ISSUE #993: FrameNet: Subcorpus name with non-ascii character can't be handled
This is probably an issue with the underlying xml reader, but here are the repro steps:

```
minimum = fn.lu(14294)
id = minimum.ID  # okay
sc = minimum.subCorpus  # throws UnicodeDecodeError
```

Looks like one of the elements in the xml file is `<subCorpus name="01-£10,000-(1)"/>`

nschneid:
Fixes #993

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -1886,7 +1886,7 @@ class FramenetCorpusReader(XMLCorpusReader):
         """Load a subcorpus of a lexical unit from the given xml."""
         sc = AttrDict()
         try:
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @nschneid

fsfszm256:
Hey friend, 


I just wanted to say  hi and to show you something really  amazing,  that stuff made my  day! Just take   a look <http://praveltonte.live5linx.com/e7hal>

Cheers, Adult Dating

Issue #1361, opened 2016-04-13
SyntaxError


abojchevski:
On Travis CI the following error occurs for `Python 3.2.5`
Just by calling `from nltk import download`
Actually it even happens with just `import nltk`

```
Traceback (most recent call last):
  File "/opt/python/3.2.5/lib/python3.2/runpy.py", line 161, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/opt/python/3.2.5/lib/python3.2/runpy.py", line 74, in _run_code
    exec(code, run_globals)
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nalaf/download_corpora.py", line 10, in <module>
    from nltk import download
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nltk/__init__.py", line 177, in <module>
    from nltk import ccg, chunk, classify, collocations
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nltk/ccg/__init__.py", line 21, in <module>
    from nltk.ccg.chart import CCGEdge, CCGLeafEdge, CCGChartParser, CCGChart
  File "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/nltk/ccg/chart.py", line 338
    if op == u'Leaf':
                   ^
SyntaxError: invalid syntax
```

tanin47:
That's odd. It seems to work on Python 3.5 and 2.7:

```
~ [19:51:19]
$ python3
Python 3.5.1 (default, Jan 22 2016, 08:52:08) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> if u'a' == u'a':
...   print("sdfdsf")
... 
sdfdsf
>>> 
```

and 

```
~ [19:55:40]
$ python
Python 2.7.10 (default, Jul 14 2015, 19:46:27) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> if u'a' == u'a':
...   print("SDfsdf")
... 
SDfsdf
>>> 
```

I'll investigate. Maybe something is specific with python 3.2.5

tanin47:
[...]

PR #1362, opened 2016-04-15
remove `u` because it doesn't work in Python 3.0 to 3.2

tanin47:
And we don't really need it if it passes the CI. So, how do I get CI to run tests on this PR?

My machine only has python 2.7 and 3.5, and tests are green on both versions.

Here's the current issue with `u`: #1361


stevenbird:
Thanks @tanin47

PR #1363, opened 2016-04-15
Try to fix unicode problem on python3.2 by removing `u`

tanin47:
Here's the problem: #1361 

I'd like the CI to run on this branch to make sure that it works on all python versions.

I hope this is the right way to test a branch on CI. If not, please let me know.


stevenbird:
We can't process this PR as it contains too many extraneous commits. I think https://github.com/nltk/nltk/pull/1362 covers the substance of this PR anyway.

Issue #1339, opened 2016-03-22
dispersion_plot incorrectly imported in nltk/text.py


Mithrillion:
The issue is around line 444
`from nltk.draw import dispersion_plot`
The function is actually defined in nltk/draw/dispersion.py
The above line causes a failure to import name error in a NLTK book chapter one example
The correct import should be:
`from nltk.draw.dispersion import dispersion_plot`

dennisobrien:
I ran across this problem as well with Python 2.7.11 and Python 3.5.1.  The problem seems to be in `nltk.draw.__init__.py` which is attempting to `import tkinter` first and only if this is successful does it bring `nltk.draw.dispersion.dispersion_plot` into the `nltk.draw` namespace.
https://github.com/nltk/nltk/blob/develop/nltk/draw/__init__.py#L17

`dispersion_plot` does not depend on `tkinter` so it can be moved out of this `else` block.

PR #1364, opened 2016-04-16
moved the import for dispersion_plot out of the else block

dennisobrien:
Moved the import for `dispersion_plot` out of the else block since it does not depend on tkinter.

This addresses issue #1339 where attempting to call `dispersion_plot` on a `Text` object would fail if `tkinter` was not installed.


stevenbird:
Thanks @dennisobrien

Issue #1293, opened 2016-02-22
Gale-Church aligner raises TypeError


cassidylaidlaw:
Running

```
nltk.translate.gale_church.align_blocks([10] * 20, [200, 400])
```

raises

```
TypeError: 'NoneType' object is not iterable
```

---

Example:

``` python
>>> from nltk.translate import gale_church
>>> gale_church.align_blocks([5,5,5], [7,7,7])
[(0, 0), (1, 1), (2, 2)]
>>> gale_church.align_blocks([10,] * 20, [5])
[(18, 0), (19, 0)]
>>> gale_church.align_blocks([10,] * 20, [5, 10])
[(16, 0), (17, 0), (18, 1), (19, 1)]
>>> gale_church.align_blocks([10] * 20, [5, 10])
[(16, 0), (17, 0), (18, 1), (19, 1)]
>>> gale_church.align_blocks([10] * 20, [200, 400])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "nltk/translate/gale_church.py", line 168, in align_blocks
    return trace(backlinks, source_sents, target_sents)
  File "nltk/translate/gale_church.py", line 84, in trace
    s, t = backlinks[pos]
TypeError: 'NoneType' object is not iterable
```

alvations:
Looks like a quirky bug:

``` python
>>> gale_church.align_blocks([500, 400], [500])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "nltk/translate/gale_church.py", line 168, in align_blocks
    return trace(backlinks, source_sents, target_sents)
  File "nltk/translate/gale_church.py", line 84, in trace
    s, t = backlinks[pos]
TypeError: 'NoneType' object is not iterable
>>> gale_church.align_blocks([500, 300], [500])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "nltk/translate/gale_church.py", line 168, in align_blocks
    return trace(backlinks, source_sents, target_sents)
  File "nltk/translate/gale_church.py", line 84, in trace
    s, t = backlinks[pos]
TypeError: 'NoneType' object is not iterable
>>> gale_church.align_blocks([500, 200], [500])
[(0, 0), (1, 0)]
>>> gale_church.align_blocks([500, 201], [500])
[(0, 0), (1, 0)]
>>> gale_church.align_blocks([500, 250], [500])
[(0, 0), (1, 0)]
[...]

PR #1366, opened 2016-04-18
Patch None TypeError in align_blocks()

alvations:
Patching #1293 with additional checks for non-valid alignments and None backlinks.


stevenbird:
Thanks @alvations

Issue #1380, opened 2016-04-21
LazyIteratorList


nschneid:
In working on enhancements to the FrameNet corpus reader, I found it would be convenient to define a lazy sequence type that is instantiated with an iterator (such as a generator expression). I came up with the following:

``` py
class LazyIteratorList(AbstractLazySequence):
    """
    Wraps an iterator, loading its elements on demand 
    and making them subscriptable.
    __repr__ displays only the first few elements.
    """
    def __init__(self, it, cache_limit=None, known_len=None):
        self._it = it
        self._len = known_len
        self._cache = []
        self._cache_limit = cache_limit
        self._i = 0 # Number of items consumed so far

    def __len__(self):
        if self._len:
            return self._len
        for x in self.iterate_from(len(self._cache)):
            pass
        return len(self._cache)

    def iterate_from(self, start):
        while self._i<start:
            v = next(self._it)
            if self._cache_limit is None or len(self._cache)+1<self._cache_limit:
                self._cache.append(v)
            self._i += 1
        i = start
        while i<len(self._cache):
            yield self._cache[i]
            i += 1
        while True:
            v = next(self._it)
            if self._cache_limit is None or len(self._cache)+1<self._cache_limit:
                self._cache.append(v)
            yield v

    def __add__(self, other):
        """Return a list concatenating self with other."""
        return type(self)(itertools.chain(self, other))

    def __radd__(self, other):
        """Return a list concatenating other with self."""
        return type(self)(itertools.chain(other, self))
```

which enables things like:

``` py
LazyIteratorList(aset for sent in self.ft_sents() for aset in sent.annotationSet[1:] if luNamePattern is None)
```

Before I check in the changes to the corpus reader, should this class go in https://github.com/nltk/nltk/blob/develop/nltk/util.py? The alternative of using `LazyMap` seems far less intuitive, especially with nested loops and conditions.

stevenbird:
@nschneid this is really great. It's always bothered me that we've had to convert a corpus iterator into a list in order to index it.

I'm thinking that this might warrant a new module, e.g. `nltk/collections.py`.
Things like `OrderedDict` from util.py could go there too.

nschneid:
Yeah, util.py is a bit crowded, so moving classes into collections.py seems sensible.

PR #1382, opened 2016-04-24
Collections refactor

nschneid:
As suggested by #1380, adds `LazyIteratorList` in a new collections.py, along with collection classes (for lazy sequences, as well as `OrderedDict` and `Trie`) moved from util.py. util.py imports them, which should preserve backward compatibility.


stevenbird:
Thanks @nschneid

nschneid:
It turns out that in Python 2.x, collections.py can shadow the `collections` module in the standard library. Should we be using `from __future__ import absolute_import`, which would solve this? Or should we choose a filename that doesn't clash with a standard library module?

nschneid:
If we keep the name collections.py, it would be necessary to add the `absolute_import` to (at least) the following modules in the root NLTK directory, as well as collection.py itself:
- compat.py
- text.py
- treeprettyprinter.py
- util.py

Or, we could just have the `absolute_import` in collection.py, and there import `defaultdict`, `deque`, and `Counter` from the standard library for use by the other modules. I've committed this minimal fix, but am happy to go with any of the other solutions.

alvations:
How about refactoring `collections` to `nltk_collections`?

nschneid:
`nltk_collections` seems like a clunky name, but it may be the simplest strategy.

There might be some magic to structure the package such that `nltk.collections` or `nltk.util.collections` is the only way to refer to the new module. But I'm not enough of a Python package wizard to know how to do this.

alvations:
I'm no python wizard / ninja but I think you can create a new directory call `util` under the main source directory then change `util.py` to `__init__.py` and put `collections.py` in the same directory.

So the `nltk` repo should look like this:

```
\nltk
     \nltk
         \app
         \ccg
         ....
         \util
             __init__.py # The old nltk/nltk/util.py
            collections.py
     \tools
     \web
       ...
```

The same namespace problem would occur if user does `import collections; from nltk.util import collections`

nschneid:
OK, I think that is working. Will push the changes when I get a chance.

iliakur:
I think it would be more "pythonic" to restrict `__init__.py` to imports and keep the actual code in a proper python script.
Here's the structure:

```
\nltk
       \nltk
              \util
                    __init__.py # import stuff here
                    util.py         # contents of old util.py here, use __init__.py to maintain interface
                    collections.py
```

stevenbird:
I agree that `nltk_collections` is a bit clunky. I find `util/util.py` a bit strange as well. What about breaking up util a bit:
- sys.py – `in_idle`, `py2?`, `set_proxy`
- file.py – `filestring`, `guess_encoding`, `binary_search_file`,
- print.py – `pr`, `print_string`, `tokenwrap`, `re_show`, `elementtree_indent`
- collections.py – `Index`, `unique_list`, `invert_dict`, `flatten`
- grams.py – `pad_sequence`, `ngrams`, `trigrams`, `bigrams`, `everygrams`, `skipgrams`
- graph.py – `transitive_closure`, `invert_graph`
- math.py – `choose`
- misc.py – `clean_html`, `clean_url`
- ../tree.py – `breadth_first`

Of course, I'm hitting other system names like `sys` and `math`, so this presumes we already avoided this problem in the case of `collections`.

alvations:
As long as `nltk.util.__init__.py` imports the submodules within the file without importing the file itself we can avoid the namespace clash, so we have to be a little hardworking by doing something like this in `nltk.util.__init__.py`:

``` python
from nltk.util.sys import *
from nltk.util.file import *
from nltk.util.collections import *
...
```

I think we can't be lazy and do:

``` python
from nltk.util import *
```

iliakur:
Afaik this pattern is actually quite common in packages: break up the module into submodules for ease of maintenance, expose only the top-level module for ease of use.

alvations:
Someday, we should break up this too: https://github.com/nltk/nltk/blob/develop/nltk/probability.py. one of the biggest file in NLTK =)

nschneid:
Sounds like a good idea to me. I'm a bit busy this week—@alvations or @Copper-Head, feel free to take a stab at this if you like.

iliakur:
Hahaha, I'm learning how to build neural networks this semester. I think wrapping up the Ngram Model module will be the extent of what I have time for :(
And yes, `probability.py` is huge and (imho) full of copy/paste code...

PR #1383, opened 2016-04-25
Handles fringe cases in BLEU

alvations:
Previously, `nltk.translate.bleu_score` is brittle and breaks in cases where there're:
- empty hypothesis with non-empty reference
- empty reference(s) with non-empty hypothesis
- ngram order > len(hypothesis) 

To handle these cases, we put in place mathematical solutions that will:
- Returns score of 0 when there's an empty hypothesis, regardless of whether the reference is empty. In the case where hypothesis is empty, we can consider this as "inadequate" translation, thus BLEU = 0.
- Returns score of 0 when there's a non-empty hypothesis in the case of empty reference. Usually this shouldn't happen because of the nature of how MT generates sentence dependent on the source and when source is non-empty but reference is empty, usually `UNK` (unknown) is generated. And in that case, the non-empty hypothesis to an empty reference implies a superfluous translation, thus BLEU = 0

These techniques to handle the fringe cases corresponds to how `multi-bleu.pl` works where it generates BLEU=0 for weird cases. And [`in the recent straw poll`](https://twitter.com/alvations/status/717647857768144896), this behavior would be set as the default and users can easily substitute the `smoothing_function` parameter to emulate other shades of BLEU.

Now we can move on to the last part of the mini-roadmap in #1330 to give an example that emulates `mteval-13a.perl`.

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -106,33 +106,40 @@ class TestBLEU(unittest.TestCase):
         self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.4729, places=4)
             
 
-@unittest.skip("Skipping fringe cases for BLEU.")
+#@unittest.skip("Skipping fringe cases for BLEU.")
 class TestBLEUFringeCases(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks for pushing forwards here!

PR #1386, opened 2016-05-03
Allow specifying substitution cost in edit distance

dnc1994:
Sometimes it makes sense to assign greater penalties for substitution edits when computing edit distance. For example when using a keyboard a substitution edit is actually consisted of 1 deletion and 1 insertion. Therefore I introduced a new parameter to `edit_distance`.

Thanks.


stevenbird:
Thanks @dnc1994 – this is good to have.

Issue #1392, opened 2016-05-08
Doctests for model package


stevenbird:
We need doctests for the new model package, as pointed out by @Copper-Head in #1351.

iliakur:
Ah, crap, I'm actually about to sit down to write some :)

iliakur:
@stevenbird should we close this for now? The doctests are present and exercise the existing code.

stevenbird:
Good point, thanks @Copper-Head

PR #1393, opened 2016-05-08
Adds model doctests and related refactorings/fixes

iliakur:
Aims to address #1392 and #396

### DIFF
--- a/nltk/model/counter.py
+++ b/nltk/model/counter.py
@@ -8,6 +8,7 @@
 from __future__ import unicode_literals
 
 from collections import Counter, defaultdict
+from copy import copy
 from itertools import chain
 
 from nltk.util import ngrams
@@ -65,6 +66,9 @@ class NgramModelVocabulary(Counter):
         # the if-clause here looks a bit dumb, should we make it clearer?
         return sum(1 for item in self if item in self) + 1
 
+    def __copy__(self):
+        return self.__class__(self._cutoff, self)
+
 
 @compat.python_2_unicode_compatible
 class EmptyVocabularyError(Exception):
@@ -97,7 +101,7 @@ class NgramCounter(object):
         # While allowing whatever the user passes to override them
         self.ngrams_kwargs.update(ngrams_kwargs)
 
-        self.vocabulary = vocabulary
+        self.vocabulary = copy(vocabulary)  # copy needed to prevent state sharing
         if unk_cutoff is not None:
             # If cutoff value is provided, override vocab's cutoff
             self.vocabulary.cutoff = unk_cutoff
@@ -117,7 +121,7 @@ class NgramCounter(object):
 
         for sent in training_text:
             checked_sent = (self.check_against_vocab(word) for word in sent)
-            for ngram in self.padded_ngrams(checked_sent):
+            for ngram in self.to_ngrams(checked_sent):
                 context, word = tuple(ngram[:-1]), ngram[-1]
                 for trunc_index, ngram_order in self._enumerate_ngram_orders():
                     trunc_context = context[trunc_index:]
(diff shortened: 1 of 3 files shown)


stevenbird:
Great work, thanks @Copper-Head!

stevenbird:
@Copper-Head – I'll deploy this to the website soon. How complete would you say these doctests are now?

iliakur:
I'd say they cover **all** of the API intended for "public" use. Edge cases aren't really covered but this should definitely showcase how to use the module.
Also, the unit test coverage is less thorough, in case that's relevant.

alvations:
@Copper-Head could you do some unit stress test? E.g. no. of tokens vs the `NgramCounter` size? I think a normal user might start breaking their own machines if NgramCounter is maxing out their RAM when they feed a sizable corpus to it.

iliakur:
Sure thing!
I've never done something like this before though, so I have a couple of questions regarding how to go about this.
1. Is the stress test explicitly part of the suite (i.e. will it be present in `test_model.py`)?
2. What is the intended output of this test? Do we want an approximate mapping from no. of tokens to NgramCounter size? For instance, I cound run this on all the largest NLTK corpora and put the results in a table for folks to extrapolate from, no?
3. Just to make sure I understand, size is in binary storage units (MB, GB, etc), right?

alvations:
**Location of Stress Test**

I am not sure where a stress test should be placed, but i would expect `nltk/nltk/test/unit/` to be a good place. And if it's necessary creating a directory within that like how the `nltk/nltk/test/unit/translate/` is located might be appropriate.

**Intended Output of Stress Test**

In web dev, i think they do load testing to make sure that their site can handle heavy flow (e.g. http://locust.io/). I think for MT and LM, it might be appropriate to warn/predict/estimate how much resources might be used by the model, especially when we're talking about phrase table of tens to hundreds of GBs and for LM it can go up to TBs of space required. 

I think lmplz, does some estimations after speed-optimized counting and then if user's machine cannot save the size of the LM, it will throw a system exit. I'm not sure how we should do it in NLTK though but using the several NLTK corpora on ngram order of 1 to 5 would be good. At least, we can also take some time estimation as well as record how big the models get. 

Then possibly, users should read this table of speed/memory tests in the docs before using the `NgramCounter`, just in case someone tries an NgramCounter with sizeable corpus and a large ngram order and start crashing their machines =)

**Size in binary storage units?** 

Yep, i think it can get up to TB if you do something like http://data.statmt.org/ngrams/

iliakur:
Ok, I'll take a shot at this in my next window of opportunity. Honestly, I didn't really have any _real world_ applications in mind when writing this code, since anyone who's training big models these days is gonna probably work with neural networks to begin with. Or write custom code in a truly "optimized" language like C++.
The primary intended audience as I understood were students and people who didn't need to work with "big data". But maybe times have changed and these audiences don't exist anymore...

alvations:
I think training 5-grams model on Europarl can lead to some big numbers in terms of RAM usage if it's fully loaded, so you have to be careful when testing too. 

I would expect the first thing NLTK users would do is to learn 5,6,7-grams on Europarl =)

iliakur:
Haha, ok. Any pointers for largest nltk corpora in terms of sheer numbers of tokens? Parses/metadata/transcriptions/translations shouldn't matter in this case.

alvations:
Maybe these would be a good start, https://github.com/nltk/nltk/blob/develop/nltk/corpus/europarl_raw.py. I would expect the different languages giving different sizes too but they should be similar or at least linearly correlated in terms of no. of token and size of model across languages. Testing on 1 to 5-grams for several languages would be good.

Another good "baseline" to test on would be to also try on `nltk.corpus.brown`. 

Maybe someone else have other suggestions to test on too. 

@Copper-Head Thanks for all the work on Ngram model!! Maybe we'll try to piece it up with the other translate code once we get stabilize the other translation related features in NLTK =)

PR #1402, opened 2016-05-20
Fix conll.py

germanferrero:
Change non-existent method call 'parse' with fromstring method on _get_parsed_sent.


stevenbird:
Thanks @germanferrero. Please add yourself to our contributors list

RokoMijic:
This is still broken for me on nltk version 3.2.1 as of November 2nd 2016.

fievelk:
@RokoMijic `3.2.1` is the latest NLTK version so far (see [ChangeLog](https://github.com/nltk/nltk/blob/develop/ChangeLog) and [NLTK News](http://www.nltk.org/news.html)). The problem has been fixed but this was after `3.2.1` was released, so the fix will be available with the next NLTK version. 

If you need this feature you could install NLTK cloning it from the `develop` branch.

PR #1403, opened 2016-05-20
Docs/ngrams

ColCarroll:
Old message was left over from `ingrams`, updated to suggest users wrap `ngrams` in `list` to get a list.


stevenbird:
Thanks @ColCarroll

PR #1405, opened 2016-05-24
fixed bug: StackDecoder would throw exception if it's unable to finish…

hikui:
According to line 182, the decoder is designed to return an empty list if it's unable to translate all words. However, because `stacks[sentence_length]` is a _Stack, `if not stacks[sentence_length]:` will always return false.

I added __bool__ and __nonzero__ to make this line of code work


stevenbird:
Thanks @hikui. Please add yourself to `AUTHORS.md` if you like.

PR #1408, opened 2016-05-29
Fix/bug loading empty files

tsubic:
When creating a corpora by loading a bunch of plain text files it can happen than some text files are empty. They get loaded into the corpus but can cause trouble later on. For example len(reared.words()) won't work; it will give an AssertionError. This is a quick fix to ignore empty files so that it doesn't even include the fileid of the file in the corpus.

Example what happens if there are empty files in the directory all/:

```
>>> reader = CategorizedPlaintextCorpusReader('all/', r'.*\.txt', cat_pattern=r'([^_]*).*\.txt')
>>> len(reader.words())
  File "/usr/lib/python3.4/site-packages/nltk/corpus/reader/util.py", line 372, in __len__
    for tok in self.iterate_from(self._offsets[-1]): pass
  File "/usr/lib/python3.4/site-packages/nltk/corpus/reader/util.py", line 394, in iterate_from
    for tok in piece.iterate_from(max(0, start_tok-offset)):
  File "/usr/lib/python3.4/site-packages/nltk/corpus/reader/util.py", line 336, in iterate_from
    assert self._len is not None
AssertionError
```

With the fix, it works just fine.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -197,3 +197,4 @@
 - Tanin Na Nakorn
 - Linghao Zhang
 - Colin Carroll
+- Tomislav Subic <https://github.com/tsubic>
--- a/nltk/corpus/reader/util.py
+++ b/nltk/corpus/reader/util.py
@@ -760,7 +760,7 @@ def find_corpus_fileids(root, regexp):
         for dirname, subdirs, fileids in os.walk(root.path, **kwargs):
             prefix = ''.join('%s/' % p for p in _path_from(root.path, dirname))
             items += [prefix+fileid for fileid in fileids
-                      if re.match(regexp, prefix+fileid)]
+                      if re.match(regexp, prefix+fileid) and os.path.getsize(root+fileid) > 0]
             # Don't visit svn directories:
             if '.svn' in subdirs: subdirs.remove('.svn')
         return sorted(items)


alexisdimi:
The bug is insidious because the `AssertionError` is really uninformative; but it should be fixed by handling empty files correctly, not by dropping them from fileids. The proposed patch would treat empty files differently from files that only contain empty space (which do not trigger an assertion error, and would not get dropped by the patch.)  

Also, the patch adds a gratuitous trip to the filesystem to check file size-- that's a whole lot of extra size checks for large corpora. The file has been read, so find out its size a different way if you really must.

stevenbird:
@alexisdimi, @tsubic – is the problem resolved by #1411?

tsubic:
@stevenbird Hey, yes, I tried out the fix from https://github.com/nltk/nltk/pull/1411 
The problem is resolved and this is certainly a better solution than mine.

alexisdimi:
I think so, it passed my tests in any case. 

Incidentally, calling `words()` or `sents()` on a corpus with no files at all raises an unrelated error: "concat() expects at least one object!“. I’m not sure it needs to be fixed, but if it’s an error to have no files then it should be raised by `fileids()` as well (it isn’t).

Alexis

stevenbird:
@alexisdimi – good point; ideally we would check that a corpus contains at least one file, but I'm happy to make this assumption for now, and let the developer of a corpus/corpus-reader interpret that error if they see it.

PR #1409, opened 2016-06-03
Much faster Vader sentiment tagging

georgeberry:
Drastic reduction of time complexity for Vader sentiment analysis. Existing `SentiText` code was `O(N^4)`, where `N` is the number of words in the corpus. New `SentiText` class is `O(N)`. On the `scikit-learn` 20 Newsgroups dataset, Vader sentiment tagging from 45 minutes to 30 seconds.

### DIFF
--- a/nltk/sentiment/vader.py
+++ b/nltk/sentiment/vader.py
@@ -5,6 +5,7 @@
 # Author: C.J. Hutto <Clayton.Hutto@gtri.gatech.edu>
 #         Ewan Klein <ewan@inf.ed.ac.uk> (modifications)
 #         Pierpaolo Pantone <24alsecondo@gmail.com> (modifications)
+#         George Berry <geb97@cornell.edu> (modifications)
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 #
@@ -24,6 +25,7 @@ import codecs
 import math
 import re
 import string
+from itertools import product
 import nltk.data
 
 ##Constants##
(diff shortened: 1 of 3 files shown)


georgeberry:
I found it difficult to use the builtin testing tools. Specifically, the deprecated `--download-cache` is called in `tox.ini`, and `vader_lexicon.txt` is not included by default in the `nltk` repo. To test, I copied `vader_lexicon.txt` from my `site-packages` to the appropriate location, and used `test/run_tests.py` on `sentiment.doctest`.

georgeberry:
Can I get a review on this? This is a small addition that offers huge performance bonuses.

nltk/sentiment/vader.py
```diff
@@ -160,37 +162,38 @@ def __init__(self, text):
...
+        """
+        no_punc_text = REGEX_REMOVE_PUNCTUATION.sub('', self.text)
         # removes punctuation (but loses emoticons & contractions)
-        words_only = text_mod.split()
-        # get rid of empty items or single letter "words" like 'a' and 'I'
-        words_only = [word for word in words_only if len(word) > 1]
-        return words_only
+        words_only = no_punc_text.split()
+        # remove singletons
+        words_only = set([w for w in words_only if len(w) > 1])
```
dimazest: you don't need to consruct a list here.

``` python
words_only = set(w for w in words_only if len(w) > ) 
```

dimazest:
Looks good.

nltk/sentiment/vader.py
georgeberry: Change committed, thanks!

stevenbird:
Thanks @dimazest. @fievelk: would you please also review?

fievelk:
@georgeberry Thank you for your PR. Could you please provide a reproducible benchmark test to compare your code to the previous version?

Edit: `nltk.download('vader_lexicon')` worked for me. In any case, I think we have to fix the vader lexicon import problem.

georgeberry:
@fiveelk there is a unit test for the Vader sentiment module at [nltk/test/sentiment.doctest](https://github.com/nltk/nltk/blob/develop/nltk/test/sentiment.doctest) beginning on line 69. On my local machine, my changes passed this test. I did this with the [test/run_test.py](https://github.com/nltk/nltk/blob/develop/nltk/test/runtests.py) script.

Let me know if you can't reproduce this. I'd consider this test sufficient validation, but let me know if you'd like something else.

stevenbird:
Thanks @georgeberry – I'm happy to go with this.

fievelk:
@georgeberry Sorry, I meant that I would like to have a reproducible test to evaluate the performance improvements of your code with regards to the old version.

Issue #1410, opened 2016-06-07
CorpusReader chokes on empty files


alexisdimi:
If there is a completely empty file (size 0) among the files of a corpus, the reader chokes with a thoroughly uninformative `AssertionError`.

```
>>> reader = nltk.corpus.reader.PlaintextCorpusReader("bad-corpus", r".*\.txt")
>>> reader.words()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../python3.4/site-packages/nltk/util.py", line 716, in __repr__
    for elt in self:
  File ".../python3.4/site-packages/nltk/corpus/reader/util.py", line 394, in iterate_from
    for tok in piece.iterate_from(max(0, start_tok-offset)):
  File ".../python3.4/site-packages/nltk/corpus/reader/util.py", line 336, in iterate_from
    assert self._len is not None
AssertionError
```

I found a [pull request](https://github.com/nltk/nltk/pull/1408) for this issue, but no associated bug report. I don't believe the work-around in the pull request is a good solution. Submitting my own pull request.

PR #1411, opened 2016-06-07
Handle empty files in a corpus

alexisdimi:
This is a solution to issue https://github.com/nltk/nltk/issues/1410. 

I have tested it with various requests on a mini-corpus containing empty files, files with just one space, just one bracket, etc. It seems to behave correctly.

[This pull request](https://github.com/nltk/nltk/pull/1408) addresses the same issue, but is not a good fix in my opinion.


stevenbird:
Good idea, thanks @alexisdimi

Issue #1421, opened 2016-06-21
Issues with UnicharsCorpusReader and NonbreakingPrefixesCorpusReader


alvations:
For `UnicharsCorpusReader` in Python 2.7, the doctests needs handle the unicode characters properly, see https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py27-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/nltk.corpus.reader.wordlist/UnicharsCorpusReader/chars/

For `NonbreakingPrefixesCorpusReader` in Python 3.5, something funky is going on with how the paths are read and NLTK is looking for `'corpora/nonbreaking_prefixes.zip/nonbreaking_prefixes/'` instead of `'corpora/nonbreaking_prefixes.zip` or `'corpora/nonbreaking_prefixes/'`, see https://nltk.ci.cloudbees.com/job/nltk/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/lastCompletedBuild/testReport/nltk.corpus.reader.wordlist/NonbreakingPrefixesCorpusReader/words/

Same for `UnicharsCorpusReader`  in Python 3.5.

@alvations, please take a look at this.

alvations:
I am not able to replicate the errors in Python 3.5.1 on my machine:

```
Python 3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import nltk
>>> nltk.download('perluniprops')
[nltk_data] Downloading package perluniprops to
[nltk_data]     /home/alvas/nltk_data...
[nltk_data]   Unzipping misc/perluniprops.zip.
True
>>> nltk.download('nonbreaking_prefixes')
[nltk_data] Downloading package nonbreaking_prefixes to
[nltk_data]     /home/alvas/nltk_data...
[nltk_data]   Unzipping corpora/nonbreaking_prefixes.zip.
True
>>> from nltk.corpus import nonbreaking_prefixes as nbp
>>> nbp.words('ta')
[...]

PR #1422, opened 2016-06-22
Fixed doctest for UnicharsCorpusReader and NonbreakingPrefixesCorpusReader

alvations:
Fixing the doctest issue in #1421. Ensures that we check for equality in list of strings rather than using the console outputs.


stevenbird:
Thanks @alvations

alvations:
@stevenbird Thanks for the merge! It's puzzling why would the doctest resolution make the nosetest problem with the .zip file loading go away from #1421. I'll close that and reopen it when it comes back on the   problem with nltk.data loading the .zip file directly without extracting it first.

PR #1423, opened 2016-06-22
Adding Basic Ngram Models

iliakur:
I've gotten some feedback from folks trying out the `model` branch and they all say it would be nice to have at least some basic ngram model classes already present in the module. I've gotten around to adding a class for MLE estimation as well as Lidstone (and as a consequence Laplace).
I really hope someone finds the time to handle the more "interesting" stuff like SGT and Kneser-Ney.

### DIFF
--- a/nltk/model/__init__.py
+++ b/nltk/model/__init__.py
@@ -6,5 +6,5 @@
 # URL: <http://nltk.org/
 # For license information, see LICENSE.TXT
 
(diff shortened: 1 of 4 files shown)


stevenbird:
Thanks @Copper-Head – sorry for the delay

iliakur:
No worries!

PR #1429, opened 2016-07-08
Removes Python 2.6 and Python 3.1-3.3 code, Improves Python 3.5 syntax

adamn:
- Changed shebang to be more platform-neutral.
- Warnings encountered with Python 3.5 were fixed.  
- Old Python 2.6 and Python 3.1-3.2 code was removed.

I also took the liberty of removing Python 3.3 but am happy to put that back in if anybody is actually using nltk on Python 3.3.  I suspect that the numbers are very small and don't reflect anything more than people playing around since most real work is probably happening on 2.7 or 3.5.

PIP is also discussing the removal of [Python 3.3](https://github.com/pypa/pip/issues/3796).


alvations:
Removing dead code from Python2.6 looks good but we might have to run some tests to check that it doesn't break something else. 

Could you check with the doctest and unittest to see the PR didn't break anything? E.g.

```
alvas@ubi:~/git/nltk$ python3 -m doctest nltk/test/*.doctest
alvas@ubi:~/git/nltk$ python2 -m doctest nltk/test/*.doctest
alvas@ubi:~/git/nltk$ python -m unittest discover
alvas@ubi:~/git/nltk$ python3 -m unittest discover
```

stevenbird:
Thanks @adamn, @alvations. I'm going to merge this and then try to deal with whatever our CI server shows up.

stevenbird:
I've dropped 3.3 from our [CI server](https://nltk.ci.cloudbees.com/job/nltk/) config.

alvations:
Maybe it's out of scope for this PR, but it might be time to consider 3.6 soon.

adamn:
Thanks for the merge.  I'm happy to look into Python 3.6 when it's in beta.  Since 3.6 is still alpha and the changes will be minimal there's no point to do it too soon.

PR #1436, opened 2016-07-14
Patch rte

ClaudeCoulombe:
I've found a small problem with the nltk.RTEFeatureExtractor, Chapter 6, example 2.2, on Recognizing Textual Entailment : RTE

RTEFeatureExtractor is not working with the actual tokenizer, created from RegexpTokenizer.

In fact if we use the same RegexpTokenizer with the exact text from the extractor, it will produce only empty strings. The goal as commented was probably to keep 'U.S.' and '$23.00' in one token. 

The proposed solution is to replace the regular expression used by RegexpTokenizer for '[A-Za-z.]+|\w+|\$[\d.]+'. It preserves 'U.S.' (i.e. included dot) in one token and '$23.00'

I could eventually patch the code add euro € or english pounds £ ou yuan?

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -39,6 +39,7 @@
 - David Coles
 - Lucas Cooper
 - Robin Cooper
+- Claude Coulombe
 - Chris Crowner
 - James Curran
 - Arthur Darcet
--- a/nltk/classify/rte_classify.py
+++ b/nltk/classify/rte_classify.py
@@ -58,10 +58,10 @@ class RTEFeatureExtractor(object):
 
         self.negwords = set(['no', 'not', 'never', 'failed', 'rejected',
                              'denied'])
-        # Try to tokenize so that abbreviations like U.S.and monetary amounts
-        # like "$23.00" are kept as tokens.
+        # Try to tokenize so that abbreviations like U.S., numbers like 3.1416, email like 
+        # name.family@mail.com, web sites, and monetary amounts like "$23.00" are kept as tokens.
         from nltk.tokenize import RegexpTokenizer
-        tokenizer = RegexpTokenizer('[A-Za-z.]+|\w+|\$[\d\.]+')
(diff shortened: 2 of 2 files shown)


stevenbird:
@ClaudeCoulombe – it looks like the net change is to replace this:

```
[A-Za-z.]+
```

with this:

```
[A-Za-z\d\.@:\/]+
```

I'll need to squash the extra commits.

Also, note that a `.` within `[]` is a literal not a wildcard so it shouldn't be escaped.

ClaudeCoulombe:
Yes, please go ahead! Sorry for all the extra commits!

stevenbird:
Thanks @ClaudeCoulombe

aetilley:
@ClaudeCoulombe, @stevenbird 

Have you found that the change made in 96f3e8f accomplishes what you want?  For me it still gives incorrect tokens (this time just the last letter of each word).

I believe the cause of the empty tokens before the change, and the cause of the new incorrect tokens after the change is in the first internal set of parentheses in either

`tokenizer = RegexpTokenizer('([A-Z]\.)+|\w+|\$[\d\.]+')`

or in 

`tokenizer = RegexpTokenizer('([\w.@:/])+|\w+|\$[\d.]+')`

The RegexpTokenizer documentation says:

:param pattern: The pattern used to build this tokenizer.                                    
    (This pattern must not contain capturing parentheses;                                    
    Use non-capturing parentheses, e.g. (?:...), instead)   

I have been able to fix this problem by either changing

`tokenizer = RegexpTokenizer('([\w.@:/])+|\w+|\$[\d.]+')`

to 

`tokenizer = RegexpTokenizer('(?:[\w.@:/])+|\w+|\$[\d.]+')`

or by simply getting rid of the first set of parentheses entirely, since they are redundant in the new regex:

`tokenizer = RegexpTokenizer('[\w.@:/]+|\w+|\$[\d.]+')`

Please let me know if this does anything for you.  If so I can submit a pull request.

yours,
Arthur Tilley

Issue #1434, opened 2016-07-13
word_tokenize doesn't tokenize "»"


mwess:
Hi all,
we tried to use nltk for the part-of-speech tagging of german texts and in some of them the french quotation marks "»" and "«" are used for quotation instead of " " ", the normal (or standard) quotation mark.

As an example consider the following sentence:

»Now that I can do.«

This is tokenized:
nltk.word_tokenize("»Now that I can do.«")

and yields:

['»Now', 'that', 'I', 'can', 'do.«']

instead of:

['»' , 'Now', 'that', 'I', 'can', 'do', '.', '«']

Can this be fixed?

Best regards.

Edit: Minor fix.

alvations:
c.f. #1682

PR #1437, opened 2016-07-21
Change treebank tokenizer to handle chevron quote brackets.

matthewmc:
Chevron brackets such as the Guillemet in French weren't being
tokenized.  Tokenizer regexs were changed to recognize the brackets.

See: #1434

However, it seems that the brackets point outward according to other examples,
so I changed the regexs with that in mind, i.e. «text», not »text«

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -39,6 +39,9 @@ Some test strings.
     >>> s10 = "There were 300,000, but that wasn't enough."
     >>> word_tokenize(s10)
     ['There', 'were', '300,000', ',', 'but', 'that', 'was', "n't", 'enough', '.']
+    >>> s11 = "«Now that I can do.»"
+    >>> word_tokenize(s11)
+    ['«', 'Now', 'that', 'I', 'can', 'do', '.', '»']
 
 Sentence tokenization in word_tokenize:
 
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -49,6 +49,7 @@ class TreebankWordTokenizer(TokenizerI):
     STARTING_QUOTES = [
         (re.compile(r'^\"'), r'``'),
         (re.compile(r'(``)'), r' \1 '),
+        (re.compile(r'(«)'), r' \1 '),
         (re.compile(r'([ (\[{<])"'), r'\1 `` '),
     ]
 
@@ -58,7 +59,7 @@ class TreebankWordTokenizer(TokenizerI):
         (re.compile(r'([:,])$'), r' \1 '),
         (re.compile(r'\.\.\.'), r' ... '),
         (re.compile(r'[;@#$%&]'), r' \g<0> '),
-        (re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'), r'\1 \2\3 '),
+        (re.compile(r'([^\.])(\.)([\]\)}>"\'»]*)\s*$'), r'\1 \2\3 '),
         (re.compile(r'[?!]'), r' \g<0> '),
 
         (re.compile(r"([^'])' "), r"\1 ' "),
@@ -73,6 +74,7 @@ class TreebankWordTokenizer(TokenizerI):
     #ending quotes
     ENDING_QUOTES = [
         (re.compile(r'"'), " '' "),
+        (re.compile(r'(»)'), r' \1 '),
         (re.compile(r'(\S)(\'\')'), r'\1 \2 '),
 
         (re.compile(r"([^' ])('[sS]|'[mM]|'[dD]|') "), r"\1 \2 "),
@@ -96,7 +98,7 @@ class TreebankWordTokenizer(TokenizerI):
     def tokenize(self, text):
         for regexp, substitution in self.STARTING_QUOTES:
             text = regexp.sub(substitution, text)
-
+            
         for regexp, substitution in self.PUNCTUATION:
             text = regexp.sub(substitution, text)
 


nltk/test/tokenize.doctest
```diff
@@ -39,6 +39,9 @@ Some test strings.
     >>> s10 = "There were 300,000, but that wasn't enough."
     >>> word_tokenize(s10)
     ['There', 'were', '300,000', ',', 'but', 'that', 'was', "n't", 'enough', '.']
+    >>> s11 = "«Now that I can do.»"
```
alvations, line 42: Possibly this will break the doctest in Python2. You might need to explicitly encode it as this instead:

``` python
s11 = u'\xabNow that I can do.\xbb'
```

nltk/test/tokenize.doctest
```diff
@@ -39,6 +39,9 @@ Some test strings.
...
+    >>> word_tokenize(s11)
+    ['«', 'Now', 'that', 'I', 'can', 'do', '.', '»']
```
alvations, line 44: Similarly the doctest should expect: 

```
[u'\xab', u'Now', u'that', u'I', u'can', u'do', u'.', u'\xbb']
```

But since Python doctest doesn't use the `u'somestring'` when returning `str` type. You might need to do the test like this:

``` python
>>> s11 = u'\xabNow that I can do.\xbb'
>>> expected = [u'\xab', u'Now', u'that', u'I', u'can', u'do', u'.', u'\xbb']
>>> word_tokenize(s11) == expected
True
```

Possibly this example might help you if you encounter problems regarding doctest on both Python2 and Python3: https://github.com/nltk/nltk/blob/develop/nltk/tokenize/toktok.py#L34

alvations:
I think it's good to fix the lack in the PTB tokenizer, but this might break the comparison between NLTK version of the PTB tokenizer vs Robert McIntyre's `sed` version from `http://www.cis.upenn.edu/~treebank/tokenizer.sed`.

Possibly this fix shouldn't be directly on the `TreebankWordTokenizer` but to a duck-type of it, so that we distinguish between the port of the PTB tokenizer vs NLTK default's `word_tokenize`. 

E.g. from https://github.com/matthewmc/nltk/blob/develop/nltk/tokenize/treebank.py#L96, add a `return_str` option:

``` python
    def tokenize(self, text, return_str=False):
        for regexp, substitution in self.STARTING_QUOTES:
            text = regexp.sub(substitution, text)

        for regexp, substitution in self.PUNCTUATION:
            text = regexp.sub(substitution, text)

        for regexp, substitution in self.PARENS_BRACKETS:
            text = regexp.sub(substitution, text)

        #add extra space to make things easier
        text = " " + text + " "

        for regexp, substitution in self.ENDING_QUOTES:
            text = regexp.sub(substitution, text)

        for regexp in self.CONTRACTIONS2:
            text = regexp.sub(r' \1 \2 ', text)
        for regexp in self.CONTRACTIONS3:
            text = regexp.sub(r' \1 \2 ', text)

        # We are not using CONTRACTIONS4 since
        # they are also commented out in the SED scripts
        # for regexp in self.CONTRACTIONS4:
        #     text = regexp.sub(r' \1 \2 \3 ', text)

        return text if return_str else text.split() 
```

Then in the https://github.com/nltk/nltk/blob/develop/nltk/tokenize/__init__.py#L96 , we do the monkey patching here:

``` python

CHERVON_REGEXES = [(re.compile(r'(\xab)'), r' \1 '), (re.compile(r'(\xbb)'), r' \1 ')]
NLTK_REGEXES = itertools.chain(*[ CHERVON_REGEXES, ])

_treebank_word_tokenize = TreebankWordTokenizer().tokenize
def nltk_word_tokenize(sent):
    tokenized_string =  _treebank_word_tokenize(sent, return_str=True)
    # Patching flaws in PTB tokenizer.
    for pattern, replacement in NLTK_REGEXES:
        tokenized_string = re.sub(pattern, replacement, tokenized_string):
    return tokenized_string.split()

def word_tokenize(sent, language='english'):
    return [token for sent in sent_tokenize(text, language)
            for token in nltk_word_tokenize(sent)]
```

nltk/tokenize/treebank.py
```diff
@@ -58,7 +59,7 @@ class TreebankWordTokenizer(TokenizerI):
         (re.compile(r'([:,])$'), r' \1 '),
         (re.compile(r'\.\.\.'), r' ... '),
         (re.compile(r'[;@#$%&]'), r' \g<0> '),
-        (re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'), r'\1 \2\3 '),
+        (re.compile(r'([^\.])(\.)([\]\)}>"\'»]*)\s*$'), r'\1 \2\3 '),
```
alvations, line 62: This line needs to be to be Python2 compatible:

``` python
(re.compile(r'([^\.])(\.)([\]\)}>"\'\xbb]*)\s*$'), r'\1 \2\3 '),
```

Could you give en example of when this would be used? Currently, I'm not getting splits from this example =)

``` python
>>> import re
>>> sent = u'\xabNow that I can do.\xbb'
>>> pattern, replacement = re.compile(r'([^\.])(\.)([\]\)}>"\'\xbb]*)\s*$'), r'\1 \2\3 '
>>> re.sub(pattern, replacement, sent).split()
[u'\xabNow', u'that', u'I', u'can', u'do', u'.\xbb']
>>> sent = u'\xabNow that I can do.\xbb      '
>>> re.sub(pattern, replacement, sent).split()
[u'\xabNow', u'that', u'I', u'can', u'do', u'.\xbb']
```

Intuitively, if the start and end quote rules fire first, you'll get them tokenized without this pattern, e.g.:

``` python
>>> sent = u'\xabNow that I can do.\xbb'
>>> pattern, replacement = re.compile(r'(\xab)'), r' \1 '
>>> re.sub(pattern, replacement, sent).split()
[u'\xab', u'Now', u'that', u'I', u'can', u'do.\xbb']
>>> pattern, replacement = re.compile(r'(\xbb)'), r' \1 '
>>> re.sub(pattern, replacement, sent).split()
[u'\xabNow', u'that', u'I', u'can', u'do.', u'\xbb']
```

nltk/tokenize/treebank.py
```diff
@@ -49,6 +49,7 @@ class TreebankWordTokenizer(TokenizerI):
     STARTING_QUOTES = [
         (re.compile(r'^\"'), r'``'),
         (re.compile(r'(``)'), r' \1 '),
+        (re.compile(r'(«)'), r' \1 '),
```
alvations, line 52: This line needs to be to be Python2 compatible:

``` python
(re.compile(r'(\xab)'), r' \1 '),
```

nltk/tokenize/treebank.py
```diff
@@ -73,6 +74,7 @@ class TreebankWordTokenizer(TokenizerI):
     #ending quotes
     ENDING_QUOTES = [
         (re.compile(r'"'), " '' "),
+        (re.compile(r'(»)'), r' \1 '),
```
alvations, line 77: This line needs to be to be Python2 compatible:

``` python
(re.compile(r'(\xbb)'), r'  \1 '),
```

nltk/tokenize/treebank.py
```diff
@@ -96,7 +98,7 @@ class TreebankWordTokenizer(TokenizerI):
     def tokenize(self, text):
         for regexp, substitution in self.STARTING_QUOTES:
             text = regexp.sub(substitution, text)
-
+
```
alvations, line 101: These extra spaces shouldn't be introduced.

nltk/tokenize/treebank.py
matthewmc, line 62: This is to tokenize the period.  Instead of u'do.', you should get [..., u'do', u'.', ...].

matthewmc:
Okay, your duck-typed method seems to be a better way to make the change.  When you make the commit though, I do still think that you should have a punctuation regex to catch punctuation (see: my line note above).

alvations:
Would it be alright if you make the changes to your branch with the duck-type instead? Otherwise it might be that I am only free on the weekends to work on this.

matthewmc:
Yeah, okay.

stevenbird:
@matthewmc: do you have time to wrap this up in the coming days please so we can include it in the next release?

stevenbird:
@alvations: I think it is fine to add new quote marks even though this breaks the comparison with the original (created before the unicode era).

stevenbird:
Superseded by https://github.com/nltk/nltk/pull/1682

PR #1444, opened 2016-08-03
Update stanford_segmenter.py

janissl:
Fixed a separator for the classpath. On Windows, a semi-colon must be used instead of the colon. Otherwise, calling the java process results in _**Could not find or load main class 'edu.stanford.nlp.ie.crf.CRFClassifier'**_.


stevenbird:
Thanks @janissl

PR #1447, opened 2016-08-05
Use a more efficient idiom for flattening nested lists

rmalouf:
A few spots (most notably `TaggerI.evaluate`) use this idiom for flattening nested lists:

```
sum(L, [])
```

I've replaced that with:

```
list(chain(*L))
```

which is more transparent (I think) and much faster.


stevenbird:
Thanks @rmalouf

PR #1450, opened 2016-08-09
explained data format for constructor

drevicko:
Added an explanation of the data and distance parameters for the AnnotationTask constructor.


stevenbird:
Thanks @drevicko

PR #1452, opened 2016-08-10
improved documentation of load_array()

drevicko:
Noted that loaded data is appended to any existing data in the AnnotationTask instance.

Also fixed a typo in the doctring for the constructor.


stevenbird:
@drevicko – thanks for these contributions. Note that the list vs iter thing is pervasive in NLTK. Sometimes I think we say "sequence" to cover both cases. Changing from "list" to "list or iterator" in two instances is probably increasing the diversity of cases that we need to find and change at some point.

PR #1454, opened 2016-08-15
Deal gracefully with degenerate cases.

drevicko:
This edit deals with a degenerate case in Krippendorff's alpha - if there is only one class.

current behaviour throws a ZeroDivisionError, this edit changes that to return complete agreement (since there is only one class)

Happy to discuss alternative behaviours...

How should we deal with no data (ie: len(self.K)==0)? Currently it'll raise a zero division error also, which isn't too informative...

### DIFF
--- a/nltk/metrics/agreement.py
+++ b/nltk/metrics/agreement.py
@@ -311,6 +311,10 @@ class AnnotationTask(object):
         """Krippendorff 1980
 
         """
+        # check for degenerate cases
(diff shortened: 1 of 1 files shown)


stevenbird:
Nice, thanks +1

stevenbird:
@drevicko: I suggest catching the zero division error and raising a ValueError with a more informative message

PR #1455, opened 2016-08-16
deal with (hopefully) all degenerate cases

drevicko:
These are all the cases I can see where the calculations can cause an exception.

### DIFF
--- a/nltk/metrics/agreement.py
+++ b/nltk/metrics/agreement.py
@@ -312,8 +312,13 @@ class AnnotationTask(object):
 
         """
         # check for degenerate cases
+        if len(self.K)==0:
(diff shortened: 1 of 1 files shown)


drevicko:
Ok, this parses and runs with my non-degenerate data.
I also did a couple of tests with dummy data and its behaving as expected.
(does nltk have unit tests?)

dimazest:
it does! you can use tox: `tox -epy35,py27`. Alternatively you can run runtests.py inside of /nltk/test, e.g.: `python runtests.py  nltk.parse.stanford nltk.tokenize.stanford parse.doctest` (you can specify what modules and doctests to run.

stevenbird:
Thanks @drevicko!

Issue #1445, opened 2016-08-04
nltk.cluster.gaac demo error


yzmyyff:
I ran this method under python3, but it throws following exception:

``` python
>>> import nltk.cluster.gaac
>>> nltk.cluster.gaac.demo()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/gaac.py", line 150, in demo
    clusters = clusterer.cluster(vectors, True)
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/gaac.py", line 41, in cluster
    return VectorSpaceClusterer.cluster(self, vectors, assign_clusters, trace)
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/util.py", line 57, in cluster
    self.cluster_vectorspace(vectors, trace)
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/gaac.py", line 79, in cluster_vectorspace
    self.update_clusters(self._num_clusters)
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/gaac.py", line 99, in update_clusters
    clusters = self._dendrogram.groups(num_clusters)
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/util.py", line 212, in groups
    return root.groups(n)
  File "/usr/local/lib/python3.5/dist-packages/nltk/cluster/util.py", line 160, in groups
    queue.sort()
TypeError: unorderable types: _DendrogramNode() < _DendrogramNode()
```

rsnape:
I believe @alvations is looking at a very similar issue based on a stackoverflow answer I wrote and some discussion in the comments there - see http://stackoverflow.com/a/28327086/838992 and the discussion on an issue I raised at the time https://github.com/nltk/nltk/issues/873.  Hopefully this helps avoid duplication if the issues have the same root cause (they seem to at a glance).

Issue #873, opened 2015-02-04
nltk.gaac.demo() incompatible with python 3


rsnape:
I found this in answering a [StackOverflow question](http://stackoverflow.com/a/28327086/838992).  The gaac.demo() does not work as in util.py, _DendrogramNode does not implement the __lt__() function.  There is an additional small bug in the format() method due to the different way the division operator works in Python 3 vs 2. 

I've hacked a solution in the linked answer which I'm quite happy for the project to adopt should they wish, obviously.

stevenbird:
Thanks @rsnape. Given that we're clustering vectors (1D arrays), I don't see how `<` is meant to work. I'd be grateful for any clarification, thanks.

```
def __lt__(self, comparator):
    return self._value < comparator._value
```

rsnape:
whoops - my mistake - that didn't work properly (I was rushing to try and help out the OP and let them try the tool - they found the same problem!).  On my machine, I'd actually used `value.any()` -  gist of the exact hack I used on util.py [available here](https://gist.github.com/rsnape/0d9c7a1ac62dd90dffa0).  Your query raises a deeper question, though - how should one compare the two vectors.  In this simple case, the `.any()` trick worked to get the OP going with the same result as in Python 2.x.  I guess one could compare the Euclidean distance from the origin.  I'm a novice with your tool, although very intrigued by it, so not sure of all the possibilities that the value of the _DendrogramNode could take.  I'm a coder rather than a Natural Language expert.

rsnape:
Musing on this and realised the other thing I haven't done is check to see whether that hack either:
1. Breaks the tool in Python 2.x (can't see why it should, but should have tested it)
2. Breaks some other functionality in the clustering toolbox (ditto)

alvations:
This is an interest problem that involves porting py2.x -> py3. And this is highlighted on http://python3porting.com/problems.html#unorderable-types-cmp-and-cmp.

Actually, the `nltk.cluster.gaac.demo()` code works in py2.x because there's an in-built `cmp()` function in Python2 that tries to compare arrays and when it tries the following compares the array using `cmp(x.any(), y.any())`:

``` python
>>> import numpy as np
>>> x, y = np.array([4, 0]), np.array([3, 3])
>>> cmp(x.any(), y.any())
0
>>> cmp(x.all(), y.all())
-1
```

---

With the current code in NLTK in py2.x, we get:

``` python
>>> import nltk
>>> nltk.cluster.gaac.demo()
[...]

PR #1456, opened 2016-08-25
Enable unordered comparison for gaac clustering

alvations:
Fixing #873 and #1445.

Instead of comparing any element between 2 vectors in `nltk.cluster.util._DendrogramNode.__lt__`, we will use the cosine distance between the vectors which should yield a (-1, +1) range and we treat any distance below zero as the case that the dendrogram node is lesser than the node it is comparing against.


stevenbird:
Thanks @alvations!

PR #1460, opened 2016-08-29
Language Model Improvements: Negative Infinity, Unit Test Names

iliakur:
The most significant change here is that `NEG_INF` which previously was set to an arbitrary "small" number is now using Python's `float("-inf")`. I tried to add some thorough tests for the effect this would have on the different models, including unit tests of entropy/perplexity calculations, please let me know what you think.

Oh and while I was at it, I did some house-keeping in the unit tests :)

### DIFF
--- a/nltk/model/ngram.py
+++ b/nltk/model/ngram.py
@@ -11,7 +11,7 @@ from math import log
 from nltk import compat
 
 
-NEG_INF = -1e6
+NEG_INF = float("-inf")
 
 
 @compat.python_2_unicode_compatible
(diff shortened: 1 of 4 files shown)


nltk/model/ngram.py
```diff
@@ -11,7 +11,7 @@
 from nltk import compat
 
 
-NEG_INF = -1e6
+NEG_INF = float("-inf")
```
stevenbird, line 14: I seem to recall that float("-inf") didn't work on some platforms (depends on the underlying version of C). But let's hope that problem is in the past.

stevenbird:
Thanks @Copper-Head – nice to have more work on this

nltk/model/ngram.py
iliakur: I'll investigate this. Wanna be sure we at least know about these issues.

PR #1462, opened 2016-08-31
Model: Important Tests and Fixes

iliakur:
We finally have unit tests that make sure our scores/probabilities sum to 1 for a given ngram context. This was a pesky issue with the old implementation, so I'm very happy to say we now have proof this issue has been addressed. Moreover, the tests should warn us if we ever regress on this front :)

While working on this I noticed that I wasn't testing the edge case where an `NgramCounter` is initialized with an empty `NgramVocabulary`, so I fixed that.

I also noticed that when counting ngrams we were not including unigram counts of items in the context of the first ngram of a sentence.
## Update:

In addition to one minor housekeeping change, I've introduced logic that checks some properties of the `context` argument of a model's `score` method. I'd really like feedback on that!

### DIFF
--- a/nltk/model/counter.py
+++ b/nltk/model/counter.py
@@ -132,8 +132,15 @@ class NgramCounter(object):
 
         for sent in training_text:
             checked_sent = (self.check_against_vocab(word) for word in sent)
+            sent_start = True
             for ngram in self.to_ngrams(checked_sent):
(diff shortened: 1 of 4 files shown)


stevenbird:
Looks good to me, thanks @Copper-Head

PR #1464, opened 2016-09-02
Remove duplicates from samples in RandomProbDist

ghost:
Currently, if there are duplicates in the samples list provided to RandomProbDist's unirand method, it produces a probability distribution where the sample do not sum to one. This problem was raised [here](https://github.com/nltk/nltk/pull/375). The simplest solution is to remove duplicate from the samples list.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -202,3 +202,4 @@
 - George Berry
 - Adam Nelson
 - J Richard Snape
+- Alexander Shtyrov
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @ashtyrov... glad to have this implementation of @kmike's [suggestion](https://github.com/nltk/nltk/pull/375#issuecomment-15539332).

PR #1468, opened 2016-09-12
Fixes bug where download status not returned.

nsfabina:
Status would return None rather than True/False under certain conditions.


stevenbird:
Thanks for catching this @nsfabina.

PR #1469, opened 2016-09-13
Allow StanfordSegmenter to work for Arabic as well

aconstantin2:
- simple extension of StanfordSegmenter definition to work for Arabic as well 
- conditional unit tests in test_tokenize.py on STANFORD_SEGMENTER and STANFORD_MODELS (usually 'STANFORD_SEGMENTER + ./data/') env variables being set)


nltk/tokenize/stanford_segmenter.py
```diff
@@ -82,7 +83,8 @@ def __init__(self,
         self._encoding = encoding
         self.java_options = java_options
         options = {} if options is None else options
-        self._options_cmd = ','.join('{0}={1}'.format(key, json.dumps(val)) for key, val in options.items())
+        self._options_cmd = ','.join('{0}={1}'.format(
```
iliakur: This actually looks more ugly. I'm guessing it was autoformatted. Mind making it look nicer while being pep8-compliant?

aconstantin2: Sure thing. I didn't know how stringent pep8 compliance should be here. I can relax rules a bit.

iliakur: Oh, I have to confess I dunno if there's a linter check as part of the CI tests, but personally I am really not that big on enforcing specifically the line length limit. In this particular case, you could split the line right before `for key, val ...` and I think you might still be within 80 characters ;)

dimazest:
It's not directly relevant to your patch, but it would be nice to fix. When I run tests with 

```
tox -e py35,py27 -- nltk.tokenize.stanford_segmenter
```

I get the following error:

```
py27 runtests: commands[1] | python runtests.py nltk.tokenize.stanford_segmenter
E
ERROR: Failure: ImportError (No module named six)
  File "/Users/dimazest/Documents/qmul/tools/src/nltk/.tox/py27/lib/python2.7/site-packages/nose/loader.py", line 407, in loadTestsFromName
    module = resolve_name(addr.module)
  File "/Users/dimazest/Documents/qmul/tools/src/nltk/.tox/py27/lib/python2.7/site-packages/nose/util.py", line 312, in resolve_name
    module = __import__('.'.join(parts_copy))
  File "/Users/dimazest/Documents/qmul/tools/src/nltk/nltk/__init__.py", line 134, in <module>
    from nltk.tokenize import *
  File "/Users/dimazest/Documents/qmul/tools/src/nltk/nltk/tokenize/__init__.py", line 70, in <module>
    from nltk.tokenize.repp     import ReppTokenizer
  File "/Users/dimazest/Documents/qmul/tools/src/nltk/nltk/tokenize/repp.py", line 12, in <module>
    from six import text_type
ImportError: No module named six
```

Could you please replace `from six import text_type` with `from nltk.six import text_type` in nltk/nltk/tokenize/repp.py

nltk/test/unit/test_tokenize.py
```diff
@@ -21,3 +24,65 @@ def test_tweet_tokenizer(self):
         expected = [':', "Let's", 'test', 'these', 'words', ':', 'resumé',
                     'España', 'München', 'français']
         self.assertEqual(tokens, expected)
+
+    @skipIf(not os.environ.get('STANFORD_MODELS') or
+            not os.environ.get('STANFORD_SEGMENTER'),
+            "At least one of STANFORD_MODELS or STANFORD_SEGMENTER "
```
dimazest: do you need both or only one variable? The code requires both.

nltk/test/unit/test_tokenize.py
```diff
@@ -21,3 +24,65 @@ def test_tweet_tokenizer(self):
...
+        The Stanford Word Segmenter for Arabic and Chinese
+        """
+        from nltk.tokenize.stanford_segmenter import StanfordSegmenter
+        from nltk.internals import find_file, find_jar, find_dir
+
+        _stanford_url = 'http://nlp.stanford.edu/software'
+        _JAR = 'stanford-segmenter.jar'
+        _SLF4J = 'slf4j-api.jar'
+
+        path_to_jar = find_jar(_JAR, None, env_vars=('STANFORD_SEGMENTER',),
```
dimazest: this will fail if STANFORD_SEGMENTER is not set.

dimazest: I would try to find the jar and catch all the Lookup exceptions, if nothing can be found then I would skip the test.

nltk/test/unit/test_tokenize.py
```diff
@@ -21,3 +24,65 @@ def test_tweet_tokenizer(self):
...
+            "At least one of STANFORD_MODELS or STANFORD_SEGMENTER "
+            "environment variables is not set")
+    def test_stanford_segmenter(self):
+        """
+        The Stanford Word Segmenter for Arabic and Chinese
+        """
+        from nltk.tokenize.stanford_segmenter import StanfordSegmenter
+        from nltk.internals import find_file, find_jar, find_dir
+
+        _stanford_url = 'http://nlp.stanford.edu/software'
```
dimazest: http://nlp.stanford.edu/software seems to be down, is there another way to get models?

nltk/tokenize/stanford_segmenter.py
```diff
@@ -43,11 +46,17 @@ class StanfordSegmenter(TokenizerI):
...
+                 path_to_jar=None, path_to_slf4j=None,
+                 java_class='edu.stanford.nlp.ie.crf.CRFClassifier',
+                 path_to_model=None,
+                 path_to_sihan_corpora_dict=None,
+                 sihan_post_processing='true',
+                 keep_whitespaces='false',
+                 path_to_dict=None,
+                 encoding='UTF-8', options=None,
+                 verbose=False, java_options='-mx2g'):
+
```
dimazest: Could you please add docstring saying what `path_to_sihan_corpora_dict`, `sihan_post_processing` and `keep_whitespaces` are.

nltk/test/unit/test_tokenize.py
aconstantin2: `nlp.stanford.edu` is not responding currently. I guess it's safe to assume temporary downtime? -- http://stanfordnlp.github.io/CoreNLP/ is the only other official site I know of, but it also links back to the same domain.

nltk/test/unit/test_tokenize.py
aconstantin2: I'm a bit concerned of it starting not to look like an NLTK unit test any more. Do you know if there's a place for something like `get_stanford_segmenter(lang)` ? Maybe `unit/utils.py` ?

nltk/test/unit/test_tokenize.py
dimazest: These are the websites i've checked. Let's wait a couple of days, maybe the website wile come back.

nltk/test/unit/test_tokenize.py
dimazest: I don't know such a function.

You could move the segmenter code to the setup method to make the test code more readable.

aconstantin2: Tried to clean things up in latest commit with `StanfordSegmenter.default_config()`. The setup method & the unit tests will now call it & skip in case of LookupErrors.

nltk/test/unit/test_tokenize.py
aconstantin2: Site's back up.

nltk/tokenize/stanford_segmenter.py
aconstantin2: I'm not entirely sure of these, as they are for Chinese and I got into this because of Arabic.

From what I can see here:
https://github.com/stanfordnlp/CoreNLP/blob/master/doc/segmenter/README-Chinese.txt
https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/wordseg/ChineseStringUtils.java#L155

the `corpora_dict` is needed for the Chinese model, and `sihan_post_processing` + `keep_whitespaces` have to do with further segmenting numbers & punctuation marks, while `keeping` or collapsing whitespace.

stevenbird:
Note that the local copy of `six` is now dropped from nltk, cf https://github.com/nltk/nltk/commit/90ac1bf13f8ecd7842d9cc39970631dcbba97ea8

aconstantin2:
Please let me know this needs anything else from my side

fievelk:
@aconstantin2 I had to stop python 2.7 tests related to your PR on our CI. They have been running for 21 hours and could not complete. Did you successfully run tests on your local machine using `tox`?

image

fievelk:
I think that the problem was related to the fact that your previous commit was pushed before you got up to date with the develop branch.

alvations:
[CI: retest]

aconstantin2:
Build times out for py27 after 30 minutes and is marked as Failed. Locally it finishes just fine with `tox`. I got the up-to-date develop branch beforehand. @fievelk or anyone, please advise.

alvations:
[CI: retest]

stevenbird:
[CI: retest]

stevenbird:
[CI: retest]

stevenbird:
@aconstantin2 – tests are clean now, for reasons that are not clear. Sorry this missed 3.2.3. I hope we get another release done within a couple of months.

stevenbird:
Released!

PR #1470, opened 2016-09-15
modified pos_tag and pos_tag_sents for russian pos-tagging

tsolakghukasyan:
pos_tag and pos_tag_sents methods from tag module altered so that they work not only for English texts but also for Russian.

I have trained an averaged perceptron model for part-of-speech tagging of Russian texts. For training I used Russian National Corpus (consists of 94240 tagged sentences with over 1 million tagged tokens): http://ruscorpora.ru/en/index.html
I used methods of PerceptronTagger class from tag module for training and storing the model file. 

The average accuracy of the model measured by 10-fold cross-validation is 99%.


stevenbird:
Thanks @tsolakghukasyan, but would you please rebase this on the head revision and don't add redundant files?

tsolakghukasyan:
I have updated it.

stevenbird:
Sorry @tsolakghukasyan this still has problems with spurious commits. I suggest you do a fresh clone of the repository, copy just your modified files across, and then a fresh commit with a detailed commit message. Thanks.

KostyaEsmukov:
Why was this closed? Looks like a very good feature to me!

KostyaEsmukov:
Oh, I'm sorry, it's in the #1480 now.

PR #1473, opened 2016-09-21
Fix undefined variable bugs

timleslie:
This pull request identifies and fixes some undefined variable bugs.

It also (as a side effect) makes trailing whitespace in these files pep8 compliant, mostly because my editor does that automatically. If you prefer I can remove all the whitespace changes and just leave the variable changes.

### DIFF
--- a/nltk/collections.py
+++ b/nltk/collections.py
@@ -584,11 +584,11 @@ class LazyIteratorList(AbstractLazySequence):
 
     def __add__(self, other):
         """Return a list concatenating self with other."""
-        return type(self)(itertools.chain(self, other))
+        return type(self)(chain(self, other))
 
     def __radd__(self, other):
         """Return a list concatenating other with self."""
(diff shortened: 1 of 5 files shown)


iliakur:
@timleslie this one commit contains a lot of unrelated changes (variable name changes, whitespace changes, the use of imported modules).
Would you mind splitting all of them up a bit, so that in case a regression is introduced it's easier to identify which commit it comes from?

stevenbird:
I agree. And a more detailed commit message too please! Thanks @timleslie – much appreciated.

timleslie:
No worries, thanks for the feedback, I will make the suggested changes later tonight and let you know when it's ready for another look.

timleslie:
I've removed the whitespace changes from the PR, so hopefully now all the changes are more obvious.

stevenbird:
Thanks @timleslie – much appreciated!

Issue #1474, opened 2016-09-22
Unused argument in nltk.sem.relextract.conllned


timleslie:
The changes in the following commit introduced a bug into `conllned()`.

https://github.com/nltk/nltk/commit/3e73834a2f18f9956c3e11466a49a9354c3924c7#diff-96e974114c1e7b1efc5976f2c387deadR409

The arguments to `show_raw_rtuple` are hardcoded to `True`, which makes `lcon` and `rcon` unused, which in turn makes the argument `trace` effectively unused.

It's not clear whether the hardcoded arguments are the intended behaviour (in which case the correct solution would be to remove the unused variables and function argument) or whether that change is a bug and the line should read `print show_raw_rtuple(rel, lcon=lcon, rcon=rcon)` as it was originally written.

PR #1475, opened 2016-09-22
Use lcon and rcon variables in conllned() when trace is False - Issue #1474

timleslie:
Provides a possible solution to https://github.com/nltk/nltk/issues/1474


stevenbird:
Looks good, thanks @timleslie

PR #1477, opened 2016-09-30
Fixing Chen&Cherry method 6 smoothing in BLEU score 

alvations:
Catching up on #1330, smoothing method 6 was returning erroneous values. After a closer look, there were 2 things that went wrong:
- `enumerate(p_n)` starts from 0 not 1, so the for loop should skip `[0,1]` and not `[1,2]`. 
- the formula ([equation 3](http://www.aclweb.org/anthology/W14-3346)) should use the `p_n.numerator` as `m` for interpolation ([equation 10](http://www.aclweb.org/anthology/W14-3346)) not the floating point precision that has been normalized by the `l`

After fixing the formula and the iterator, the bleu score with method 6 smoothing is returning reasonable values and more importantly giving more weight to the higher order n-grams as compared to method 1-5 =)


alvations:
Now we're left with the wrapper to `mteval.pl` on the mini-roadmap for BLEU.

stevenbird:
Nice work @alvations, thanks!

PR #1478, opened 2016-10-05
Use shutil.copyfileobj for unzipping

ikorolev93:
Reading into variable and writing does consume all memory in case file is big (like panlex_lite)


stevenbird:
Good idea, thanks @deadNightTiger

PR #1480, opened 2016-10-13
modified pos_tag and pos_tag_sents for russian pos-tagging

tsolakghukasyan:
I have modified pos_tag and pos_tag_sents methods from tag module so that they work not only for English texts but also for Russian. An additional parameter 'language' was added. The parameter's default value is 'english', thus the methods behave the same way as before when 'language' is not specified. For Russian pos-tagging, the parameter needs to be set to 'russian'.

I trained an averaged perceptron model for part-of-speech tagging of Russian texts. For training I used Russian National Corpus (consists of 94240 tagged sentences with over 1 million tagged tokens): http://ruscorpora.ru/en/index.html
Methods of PerceptronTagger class from tag module were used for training and storing the model file.

The average accuracy of the model is 99% (measured by 10-fold cross-validation).

### DIFF
--- a/nltk/tag/perceptron.py
+++ b/nltk/tag/perceptron.py
@@ -197,7 +197,7 @@ class PerceptronTagger(TaggerI):
         # Pickle as a binary file
         if save_loc is not None:
             with open(save_loc, 'wb') as fout:
-                pickle.dump((self.model.weights, self.tagdict, self.classes), fout, -1)
+                pickle.dump((self.model.weights, self.tagdict, self.classes), fout, 2)
         
 
     def load(self, loc):
--- a/nltk/tag/perceptron.py
+++ b/nltk/tag/perceptron.py
@@ -197,6 +197,7 @@ class PerceptronTagger(TaggerI):
         # Pickle as a binary file
         if save_loc is not None:
             with open(save_loc, 'wb') as fout:
+                # changed protocol from -1 to 2 to make pickling Python 2 compatible
                 pickle.dump((self.model.weights, self.tagdict, self.classes), fout, 2)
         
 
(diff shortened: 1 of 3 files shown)


KostyaEsmukov:
Looking forward to this one. Good work!

nltk/tag/perceptron.py
```diff
@@ -197,7 +197,7 @@ def train(self, sentences, save_loc=None, nr_iter=5):
         # Pickle as a binary file
         if save_loc is not None:
             with open(save_loc, 'wb') as fout:
-                pickle.dump((self.model.weights, self.tagdict, self.classes), fout, -1)
+                pickle.dump((self.model.weights, self.tagdict, self.classes), fout, 2)
```
KostyaEsmukov, line 201: I think this change deserves a comment, something like

```
# Make pickle dump compatible with Python 2
# https://docs.python.org/3/library/pickle.html#data-stream-format
```

stevenbird:
@tsolakghukasyan – thanks for the contribution. Can I ask you to change the optional argument and value to `lang = "rus"` for consistency with the rest of NLTK, since we use the `lang` attribute and ISO 639-3 codes for identifying languages. The default value of `lang` would be `"eng"` (e.g. https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L435). Thanks.

tsolakghukasyan:
@stevenbird thanks for the reply. I have made the necessary changes.

nltk/tag/perceptron.py
tsolakghukasyan: @KostyaEsmukov thanks for the feedback. I have added a comment.

stevenbird:
Thanks @tsolakghukasyan

PR #1488, opened 2016-10-26
Final tests and tweaks to fix BLEU

alvations:
**Wrote tests to validate NLTK BLEU with mteval-13a.pl**:

Currently, it's checking that NLTK implementations don't stray too far away from mteval's by 0.05 BLEU, when it ranges between [0.0, 1.0]; i.e. 0.5 BLEU if score range between [0.0, 100.0]. Details on #1330 

**Tweak in BLEU sum(log(p_i)**:

One thing that cannot be controlled is the data structures that the different smoothing methods can return. As much as possible, the code will try to keep the `Fraction` object, but when it's float, then checking the `p_i.numerator` wouldn't be possible. Since smoothen values are always floatable and non-zero and unsmoothen values are always `Fractions`, we can tease apart the sum log calculations and that would look a little ugly but that'll avoid creation of a new polymorphic float-fraction object that allow floats to hallucinate a numerator/denominator.

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -188,24 +188,3 @@ class TestBLEUvsMteval13a(unittest.TestCase):
                     #       +/- 1.0 BLEU might be "statistically significant",
                     #       the actual translation quality might not be.
                     assert abs(mteval_bleu - nltk_bleu) < 0.5
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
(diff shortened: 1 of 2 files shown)


stevenbird:
Sorry for the delay @alvations – I'll review this as soon as possible

stevenbird:
Thanks @alvations!

alvations:
Thanks @stevenbird for reviewing this and all the previous related PRs!

PR #1496, opened 2016-11-10
Added ChrF and GLEU MT evaluation metrics

alvations:
ChrF is basically a character level BLEU without brevity penalty but it that takes an interpolated score between ngram precision and recall. There's no sentence-level version of this. 

GLEU is the Google Neural Machine Translation version for sentence BLEU where it takes the `min(precision, recall)` as the score. There's no corpus-level version of this either. 

Do we raise an error if users try to do sentence-level chrf like `sentence_chrf` or corpus-level gleu like `corpus_gleu`?

### DIFF
--- a/nltk/translate/chrf_score.py
+++ b/nltk/translate/chrf_score.py
@@ -1,0 +1,137 @@
+# -*- coding: utf-8 -*-
+# Natural Language Toolkit: ChrF score
+#
+# Copyright (C) 2001-2016 NLTK Project
+# Authors: Maja Popovic
+# Contributors: Liling Tan
+# URL: <http://nltk.org/>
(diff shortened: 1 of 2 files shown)


stevenbird:
Great to have, thanks @alvations!

alvations:
> Do we raise an error if users try to do sentence-level chrf like sentence_chrf or corpus-level gleu like corpus_gleu?

After some thought, I think we should have a `sentence_chrf` function that throws a `NotImplementedError` and a `corpus_gleu` function that throws a `NotImplementedError` too. We can also raised a meaning message with the error. That way, the user won't be caught off-guard.

Penny for a thought, anyone?

PR #1499, opened 2016-11-12
Fix SENNA tests

fievelk:
This PR replaces Senna doctests with unit tests.

Doctests were failing on systems that did not have a `Senna` executable in `/usr/share/senna-v3.0`. Since SENNA is optional, this behaviour is not optimal. New tests check if Senna is present on the system and, if the executable is not found, tests are skipped.

Note: this PR is related to #1500 (`SENNA` environment variable would not work otherwise).

Related issue: #1466


stevenbird:
Thanks @fievelk

PR #1501, opened 2016-11-12
Fix TransitionParser tests

fievelk:
This PR fixes `TransitionParser` tests. 

These tests are currently producing an extremely verbose SVM output that messes up the normal `tox` output. Since we do not care about visualizing SVM output values during automated tests, I just added a `verbose` flag to the `train` method. Tests are now conducted using `verbose=False`.

`TransitionParser` tests were also generating temporary files that were not removed after testing:
```
temp.arceager.model
temp.arcstd.model
```
I removed the related entries from `.gitignore` (one of these entries was mistyped, so it was not working anyway) and made sure that these files are deleted after tests.

### DIFF
--- a/nltk/parse/transitionparser.py
+++ b/nltk/parse/transitionparser.py
@@ -328,8 +328,8 @@ class TransitionParser(ParserI):
     def _is_projective(self, depgraph):
         arc_list = []
         for key in depgraph.nodes:
-            node = depgraph.nodes[key]           
-            
+            node = depgraph.nodes[key]
+
             if 'head' in node:
                 childIdx = node['address']
                 parentIdx = node['head']
@@ -490,7 +490,7 @@ class TransitionParser(ParserI):
         print(" Number of valid (projective) examples : " + str(countProj))
         return training_seq
 
-    def train(self, depgraphs, modelfile):
+    def train(self, depgraphs, modelfile, verbose=True):
         """
         :param depgraphs : list of DependencyGraph as the training data
         :type depgraphs : DependencyGraph
@@ -522,7 +522,7 @@ class TransitionParser(ParserI):
                 coef0=0,
                 gamma=0.2,
                 C=0.5,
-                verbose=True,
+                verbose=verbose,
                 probability=True)
 
             model.fit(x_train, y_train)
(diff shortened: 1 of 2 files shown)


nltk/parse/transitionparser.py
```diff
@@ -490,7 +490,7 @@ def _create_training_examples_arc_eager(self, depgraphs, input_file):
         print(" Number of valid (projective) examples : " + str(countProj))
         return training_seq
 
-    def train(self, depgraphs, modelfile):
+    def train(self, depgraphs, modelfile, verbose=True):
```
stevenbird, line 493: Is verbose=True the correct default setting?

fievelk: `TransitionParser` used `verbose=True` as default value to train the SVM model (see [line 525][1] before my PR). Conversely, `sklearn.svm.SVC` uses `verbose=False` as default (see [sklearn.svm.SVC][2]).

I made this variable dynamic so that it can be changed by users (and it's useful to clean tests output), but I did not change the behaviour that was originally implemented by @longdt219. If you think this has to be fixed, I can work on it.

[1]: https://github.com/nltk/nltk/pull/1501/files#diff-ac01d9460dcfd9e576978fbf2885bb87L525
[2]: http://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html

fievelk: @stevenbird `TransitionParser` used `verbose=True` as default value to train the SVM model (see [line 525](https://github.com/nltk/nltk/pull/1501/files#diff-ac01d9460dcfd9e576978fbf2885bb87L525) before my PR). 
Conversely, `sklearn.svm.SVC` uses `verbose=False` as default (see [sklearn.svm.SVC](http://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html)).

I made this variable dynamic so that it can be changed by users (and it's useful to clean tests output), but I did not change the behaviour that was originally implemented by @longdt219. If you think this has to be fixed, I can work on it.

stevenbird: Thanks for your response @fievelk. I notice that there are about 120 `verbose=False` in our codebase, and about 10 `verbose=True`. But since you're preserving existing behaviour, I'm happy to let it stand and merge this, and deal with this inconsistency in a separate issue: https://github.com/nltk/nltk/issues/1509

Issue #1418, opened 2016-06-21
aline breaks in Python 2.7


stevenbird:
There's a character encoding issue for `metrics.aline` in Python 2.7, as reported by our [CI server](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/testReport/).

stevenbird:
@geoffbacon – would you please take a look at this?

fievelk:
Please refer to the [NLTK Testing](http://www.nltk.org/dev/local_testing.html#writing-tests) guide:

> - Do not write doctests that has non-ascii output (they are not supported in
>   Python 2.x). Incorrect::
> 
> ```
> >>> greeting
> u'Привет'
> ```
> 
>  The proper way is to rewrite such doctest as unittest.

PR #1502, opened 2016-11-12
Fix Aline tests

fievelk:
This PR fixes `aline.py` tests (see #1418) and adds more test cases.

Doctests were failing with `python2.7` because of encoding problems. We now use unittest and add `from __future__ import unicode_literals` in the module. 

The `numpy` import strategy as also been fixed: this import used to break NLTK since `numpy` is considered to be optional.

Thanks @geoffbacon for your help.

### DIFF
--- a/nltk/metrics/aline.py
+++ b/nltk/metrics/aline.py
@@ -516,7 +516,10 @@ def demo():
     """
     data = [pair.split(',') for pair in cognate_data.split('\n')]
     for pair in data:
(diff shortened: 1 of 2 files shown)


stevenbird:
@geoffbacon: would you please review this

geoffbacon:
Everything works great, thanks @fievelk!

stevenbird:
Thanks @fievelk and @geoffbacon

Issue #815, opened 2014-12-16
Floating point division


stevenbird:
Audit the codebase and replace old idioms for getting floating point division, e.g.:

```
z = x / float(y)
z = x / (1.0 * y)
```

with:

```
from __future__ import division
...
z = x / y
```

erip:
Note: anywhere you use `__future__.division`, you'll need to convert the existing integer division to `//`. Fret not, this also works in python 2. This should be the first step in the refactor.

Alternatively you could cast the quotient to `int`, but I think this is less idiomatic.

stevenbird:
Good point, thanks @erip.

PR #1508, opened 2016-11-15
Replace old floating point division

prasastoadi:
Replace old floating point division as stated here https://github.com/nltk/nltk/issues/815


stevenbird:
Thanks @prasastoadi

PR #1511, opened 2016-11-18
Fixed a PATH separator in stanford_segmenter.py

janissl:
Just improved my own commit submitted previously (on 03 Aug 2016).


stevenbird:
@janissl - would you please rebase this commit on the current head revision so it can be merged easily?

stevenbird:
Thanks @janissl

PR #1513, opened 2016-11-19
Fix for def dispersion_plot(self, words): error

SaintNazaire:
Changed import to fix import error:
from nltk.draw.dispersion import dispersion_plot


stevenbird:
@SaintNazaire: the init.py file already includes this line: `from nltk.draw.dispersion import dispersion_plot`. So user code with a package level import statement `from nltk.draw import dispersion_plot` ought to work fine.

What error are you trying to fix?

Issue #1334, opened 2016-03-14
'Text' object has no attribute 'generate'


SeanPlusPlus:
On page 7 of the hard copy of the book "Natural Language Processing with Python", there is an example of the command:

`>>> text3.generate()`

Using nltk version 3.2 I get the following error:

`AttributeError: 'Text' object has no attribute 'generate'`

The online book [states that this feature will be reinstated in a subsequent version](http://www.nltk.org/book/ch01.html#searching-text).

Which version will this feature be in?

Thanks!

ghost:
I also have this issue and have found nothing to say when it will be reinstated. Very frustrating.

alvations:
`Text.generate()` has been deprecated.

Issue #1180, opened 2015-10-27
Is generate() coming back? 


JonathanReeve:
Would really love to see this function back again. The book says it will be reinstated in a later version, but when?

alvations:
+1 .  Given the recent RNN generation from @karpathy (https://github.com/karpathy/char-rnn), it might be very easy to create a `generate()` algorithm but support of any neural net implementation is all very unstable, API changes almost every 2-3 weeks =(

JonathanReeve:
Is there a reason why this function was removed? Just curious.

Issue #736, opened 2014-08-20
text.generate() does not exist, but is still referenced


Kasran:
I've been trying to follow along with the Natural Language Processing book, but right in the first chapter I'm coming across some issues. After `import`ing everything from `nltk.book`, my first thought was to try `text3.generate()` as was demonstrated in one of the examples. Of course, I got a lovely `AttributeError` because the Text class apparently doesn't have that method in the NLTK that I installed.
Furthermore, even running `nltk.text.demo()` tries to do generated text - and returns the same error.
Of course, I couldn't find any documentation for the `generate()` method, so I'm assuming it was removed; if that's the case, you should remove references to it from `nltk.text.demo()` and from the textbook.

I'm using Python 2.7.8 with NLTK 3.0.0b1 (which was the version available via Windows installer package from PyPI at the time of this writing). The `text3.generate()` example is in both the old and current versions of the textbook.

stevenbird:
Sorry for the confusion. There were problems with NLTK's language modelling
class and so we removed it until the problems are resolved. This broke the
text generation functionality. I've updated the online version of the book
with a note about this, and updated the text demo (to be included in
release 3.0.0b2 soon).

-Steven Bird

Kasran:
Alright! Totally understandable. I guess I'll have to write my own chatterbots with applied Markov chains, the old crude way, until it gets fixed.

Thanks for your prompt response.

speedmancs:
I also met with this same issue which puzzled me a little bit. Thanks for this thread which google search had taken me to.

alvations:
hmmm. does this help (https://github.com/alvations/nltk/blob/develop/nltk/translate/decoder.py#L33)? The language model seems to work fine when decoding. I'll be pushing up an updated and properly documented code for the `translate` library soon. 

Sorry I was out for hackathon, workshops and conferences back to back. I'll try to push it by the end of this week, i hope.

Ah but my model loads a precacluated model, not building the model from scratch. Is anyone building a language model module?

Gallaecio:
I see two issues with the current code (besides the fact that it does not work), user-experience-wise:

1.  In the book, the method does not take any parameters (at least when it first appears in the book), but the current signature requires a `words` parameter. So when you follow the book, what you get is the following:

        >>> text1.generate()
        Traceback (most recent call last):
          File "<stdin>", line 1, in <module>
        TypeError: generate() missing 1 required positional argument: 'words'

[...]

PR #1514, opened 2016-11-19
Issues warning upon calling deprecated generate() function

SaintNazaire:


fievelk:
@SaintNazaire Even if this code is self explanatory, I would suggest you to provide a description for your future pull requests, in order to make it easier to review them.
Referenced issues: #1180, #736, #1334. 

(I think we could close some of those issues since they all refer to the same problem).

Regarding the content of the PR: I would probably avoid specifying the NLTK version in the warning, or make it dynamic so that it refers the current NLTK version. We do not know if this will be fixed in the next release; if not, we would have to change the warning every time NLTK is updated and I think this is not desirable.

fievelk (approved):

stevenbird:
I'll merge this and fix the version issue.

PR #1527, opened 2016-11-27
A text frequency counter with an option to ignore non-alphanumeric characters

sr-murthy:


stevenbird:
@sr-murthy: thanks for your contribution. However, I think the following is sufficient and that we don't need to make a special case in the code: `FreqDist(tok for tok in text if tok.isalnum())`

PR #1536, opened 2016-12-04
Further improve tests output

fievelk:
This PR aims to further improve test output readability using [`rednose`][1] plugin for `nose`.

Here is a comparison between test outputs before and after:
- Before
tox_before_rednose1

- After
tox_after_rednose1

(Note: `gluesemantics` tests are still broken, so the output is slightly different. However the difference is still very clear).

I do not think that `rednose` should be added to `setup.py` as a required NLTK dependency. Therefore, I modified `runtests.py` so that it can be run (without `tox`) using `rednose` just as an optional package.

[1]:https://github.com/JBKahn/rednose


stevenbird:
Thanks for this @fievelk and for making it optional

Issue #1531, opened 2016-12-02
Dependencies on PyPI needs to be updated. 


alvations:
Most users installs `nltk` through:

```
$ pip install -U nltk
```

or 

```
$ conda install nltk
```

but somehow the PyPI requirements seems to be far lacking behind the ones in `pip-req.txt`.

fievelk:
I think this is related to the fact that [`nltk/setup.py`][1] only lists `six` as required dependency:
```python
setup(
    ...
    install_requires = ['six>=1.10.0'],
    ...
)
```

From [`pip` docs][2]:
> **`Install_requires`**
> `install_requires` is a setuptools `setup.py` keyword that should be used to specify what a project minimally needs to run correctly. When the project is installed by `pip`, this is the specification that is used to install its dependencies.

Conversely, [`nltk/pip-req.txt`][3] is a [requirement file][4]:
> Whereas `install_requires` metadata is automatically analyzed by `pip` during an install, requirements files are not, and only are used when a user specifically installs them using `pip install -r`.

(See [docs][5]).

Please note that [`nltk/jenkins.sh`][6] is actually using `pip-req.txt`:
```bash
#download nltk python dependencies
pip install --upgrade -r pip-req.txt --allow-external matplotlib --allow-unverified matplotlib
```

So I think we should add our minimal requirements to `install_requires` in `setup.py`. Moreover, in order to avoid redundancy, we could probably think about completely removing `pip-req.txt`.

@stevenbird @kmike what do you think about this?

[1]: https://github.com/nltk/nltk/blob/4a4a5031f5c0f99b1ca13079518c221b338ef413/setup.py
[2]: https://packaging.python.org/requirements/#install-requires
[3]: https://github.com/nltk/nltk/blob/4a4a5031f5c0f99b1ca13079518c221b338ef413/pip-req.txt
[4]: https://pip.pypa.io/en/latest/user_guide/#requirements-files
[5]: https://packaging.python.org/requirements/#requirements-files
[6]: https://github.com/nltk/nltk/blob/4a4a5031f5c0f99b1ca13079518c221b338ef413/jenkins.sh

kmike:
I think if NLTK is usable without matplotlib then we shouldn't add matplotlib to install_requires. It is a huge dependency, and there is no reason to force users to install it. Creating extra_requires makes more sense - this way user may e.g. run `pip install nltk[plotting]` and get all dependencies required for nltk plotting support.

fievelk:
[...]

PR #1538, opened 2016-12-07
Add optional dependencies for NLTK installation

fievelk:
Reference: #1531 

---

This PR aims to include all optional dependencies to `setup.py` using [`extras_require`][1].

This will allow users to install NLTK with specific sets of additional (non-required) dependencies using commands like to the following:
```bash
pip install nltk[all]    # Install all optional Python dependencies
pip install nltk[plot]   # Install optional dependencies for plots
# etc.
```

For the moment, the `'all'` requirements list contains all the dependencies that were already written in `pip-req.txt`.

I would like to have your opinion about the best strategy to create a list of these requirements for other subsets (e.g. `'plot'` or `'machine-learning'`, as suggested by @alvations). 

1. Do you think we should do it in `setup.py`, or should we use a different module instead? I am thinking about something like a module called `optional_dependencies.py` that could define several arrays of dependencies (one for each `extras_require` key).

2. As for the content, what do you think would be the best way to group these packages? Could you please provide some kind of classification to create the groups?

[1]: http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ natural language processing.  NLTK requires Python 2.7, or 3.4+.""",
     'Topic :: Text Processing :: Linguistic',
     ],
     package_data = {'nltk': ['test/*.doctest', 'VERSION']},
-    install_requires = ['six>=1.10.0'],
+    install_requires = ['six'],
     extras_require = extras_require,
     packages = find_packages(),
     zip_safe=False, # since normal files will be present too?
--- a/setup.py
+++ b/setup.py
@@ -69,6 +69,22 @@ natural language processing.  NLTK requires Python 2.7, or 3.4+.""",
     ],
     package_data = {'nltk': ['test/*.doctest', 'VERSION']},
     install_requires = ['six>=1.10.0'],
+    extras_require = {
+        'all': [
+            'nose>=1.3.0',
+            'tox>=1.6.1',
+            'coverage>=3.7.1',
+            'pylint>=1.1.0',
+            'numpy>=1.8.0',
+            'scipy>=0.13.2',
+            'matplotlib>=1.3.1',
+            'scikit-learn>=0.14.1',
+            'python-crfsuite>=0.8.2',
+            'gensim>=0.11.1',
+            'pyparsing>=2.0.3',
+            'twython>=3.2.0'
+        ]
+    },
     packages = find_packages(),
     zip_safe=False, # since normal files will be present too?
     )
--- a/setup.py
+++ b/setup.py
@@ -41,12 +41,6 @@ extras_require = {
     'plot': [
         'matplotlib',
     ],
-    'tests': [
-        'coverage',
-        'nose',
-        'pylint',
-        'tox'
-    ],
     'tgrep': [
         'pyparsing',
     ],
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,6 @@ extras_require = {
     'machine_learning': [
         'gensim',
         'numpy',
-        'pyparsing',
         'python-crfsuite',
         'scikit-learn',
         'scipy'
@@ -47,6 +46,9 @@ extras_require = {
         'nose',
         'pylint',
         'tox'
+    ],
+    'tgrep': [
+        'pyparsing',
     ],
     'twitter': [
         'twython'
--- a/setup.py
+++ b/setup.py
@@ -71,38 +71,38 @@ natural language processing.  NLTK requires Python 2.7, or 3.4+.""",
     install_requires = ['six>=1.10.0'],
     extras_require = {
         'all': [
-            'nose>=1.3.0',
-            'tox>=1.6.1',
-            'coverage>=3.7.1',
-            'pylint>=1.1.0',
-            'numpy>=1.8.0',
-            'scipy>=0.13.2',
-            'matplotlib>=1.3.1',
-            'scikit-learn>=0.14.1',
-            'python-crfsuite>=0.8.2',
-            'gensim>=0.11.1',
-            'pyparsing>=2.0.3',
-            'twython>=3.2.0'
+            'nose',
+            'tox',
+            'coverage',
+            'pylint',
+            'numpy',
+            'scipy',
+            'matplotlib',
+            'scikit-learn',
+            'python-crfsuite',
+            'gensim',
+            'pyparsing',
+            'twython'
         ],
         'machine_learning': [
-            'gensim>=0.11.1',
-            'numpy>=1.8.0',
-            'pyparsing>=2.0.3',
-            'python-crfsuite>=0.8.2',
-            'scikit-learn>=0.14.1',
-            'scipy>=0.13.2'
+            'gensim',
+            'numpy',
+            'pyparsing',
+            'python-crfsuite',
+            'scikit-learn',
+            'scipy'
         ],
         'plot': [
-            'matplotlib>=1.3.1',
+            'matplotlib',
         ],
         'tests': [
-            'coverage>=3.7.1',
-            'nose>=1.3.0',
-            'pylint>=1.1.0',
-            'tox>=1.6.1'
+            'coverage',
+            'nose',
+            'pylint',
+            'tox'
         ],
         'twitter': [
-            'twython>=3.2.0'
+            'twython'
         ]
     },
     packages = find_packages(),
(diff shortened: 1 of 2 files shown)


fievelk:
Since importing an external module would have been tricky and would have added some complexity without a good reason, I wrote the dependencies directly within `setup.py` (thank you @kmike).

I'm wondering if we actually need to provide a `test` group. In any case this won't hurt. 

If you do not have suggestions (especially about the groups) I think this is ready to be merged.

setup.py
```diff
@@ -69,6 +69,42 @@
...
+            'scikit-learn>=0.14.1',
+            'python-crfsuite>=0.8.2',
+            'gensim>=0.11.1',
+            'pyparsing>=2.0.3',
+            'twython>=3.2.0'
+        ],
+        'machine_learning': [
+            'gensim>=0.11.1',
+            'numpy>=1.8.0',
+            'pyparsing>=2.0.3',
```
kmike: hm, what is pyparsing used for by NLTK? It looks like a transitive dependency for some other package

fievelk: I can see an import in [`tgrep.py`](https://github.com/nltk/nltk/blob/develop/nltk/tgrep.py#L118). It is then used in [`_build_tgrep_parser()`](https://github.com/nltk/nltk/blob/develop/nltk/tgrep.py#L759).

kmike: aha, makes sense!

setup.py
```diff
@@ -69,6 +69,42 @@
     ],
     package_data = {'nltk': ['test/*.doctest', 'VERSION']},
     install_requires = ['six>=1.10.0'],
+    extras_require = {
```
dimazest: that's great! I have two comments though.
 
1. Unless it's really known that nltk doesn't work with a lower version of a package, I would not set the minimum version.  For example, NLTK doesn't make use of numpy's functionality that is available only from version 1.8.0, I guess.

2. The list should be easy maintainable. Right now, gensim is listed under ``all`` and ``machine_learning``. Maybe, it worth constructing the dictionary of extras outside of the setup() call. Then ``all`` is the union of all extras, something like:
```python
extras_require = {
    'machine_learning': [...],
    'plot ': [...],
    ...
}

extras_require['all'] = set(itertools.chain.from_iterable(extras_require.values()))

```

dimazest:
Great effort. 

I suggest to keep everything inside of setup.py, because importing might become too difficult. For example, you can't import nltk because nltk/__init__.py might import something that requires a package that is not installed yet.

There are ways to overcome this, but I would keep things simple.

setup.py
fievelk: 1. It seems that these package versions have been specified almost 3 years ago (see https://github.com/nltk/nltk/commit/2423dc79152dab2de13bfa9a7672e1bc1d0e1ca2). So I think we can safely remove the minimum version requirements if there are no objections.

2. I like your solution. I was trying to have something like this importing a separate module, but - as you said in the other comment - this was getting overly complex without any good reason.

Thank you!

fievelk:
@dimazest Yes, I tried several solutions but the easiest ones were failing (especially during tests with `tox`). Editing the manifest file could have also solved the problem, but I wasn't very convinced by the idea.

I removed the packages versions from every optional dependency. Do you think it would be better to remove the minimal requirements from `six` as well? I don't think it should be a problem.

dimazest:
I would delete it. Maybe try `1.9.0`, it was released in January 2015. `1.8.0` was released in 2014 according to https://pypi.python.org/pypi/six/json

Ubuntu Trusty (14.04LTS) comes with 1.5.2 http://packages.ubuntu.com/trusty/python-six

I would test with 1.5.2 and if it doesn't then set the minimal working version.

fievelk:
I ran tests with `six>=1.5.2` and `six` (without minimal version). Tests are succesful in any case, so I'm removing the minimum version requirement.

dimazest:
could you do `six==1.5.2` to check that it works with 1.5.2, not just with a newer version.

setup.py
```diff
@@ -29,6 +29,33 @@
 # setuptools
 from setuptools import setup, find_packages
 
+# Specify groups of optional dependencies
+extras_require = {
+    'machine_learning': [
+        'gensim',
+        'numpy',
+        'pyparsing',
```
kmike: If we want to be pedantic then pyparsing should be moved to some other section ("tgrep"?) because neither pyparsing nor tgrep are related to machine learning.

setup.py
```diff
@@ -29,6 +29,33 @@
...
+        'scipy'
+    ],
+    'plot': [
+        'matplotlib',
+    ],
+    'tests': [
+        'coverage',
+        'nose',
+        'pylint',
+        'tox'
```
kmike: To run tests you need either tox or coverage+nose+pylint, not both of them. I think removing tox from here is fine.

fievelk:
@dimazest Sorry for the typo. Tests are successful using `six==1.5.2`. So I think we can safely remove the minimum version requirement.

stevenbird:
@dimazest, @kmike: thanks for the feedback!
@fievelk: what still needs to happen before I merge this?

fievelk:
@stevenbird I want to create a `tgrep` group as suggested by @kmike. I also want to check his suggestions about the `test` group: would it be better to only keep `tox`, or keep `coverage`, `nose` and `pylint` instead?

stevenbird:
@fievelk: It looks like @kmike thinks it would be better to remove tox. I don't have any advice here.

fievelk:
If we remove `nose`, `pylint` and `coverage` users would not be able to run
```bash
python nltk/test/runtests.py
```
At the same time, if we remove `tox` users would not be able to use it to run tests.

I think that running `tox` is the easiest way to run tests in NLTK: it performs tests in several virtual environments, with and without optional dependencies. 
In any case, I think that it's pretty uncommon for users to install NLTK using `pip` and then use test functionalities. I suppose that, most of the time, tests will be performed by developers that cloned our GitHub repo. So they will probably decide which package to use for testing.

@kmike do you have some reason to prefer `nose`, `coverage` and `pylint` over `tox`?

[1]: https://github.com/nltk/nltk/blob/develop/web/dev/local_testing.rst

kmike:
@fievelk for me adding a requirements-text.txt file makes more sense than a testing section in extra_requires, because of the reason you described - tests are usually executed on source checkout, not on NLTK from pypi. If you run `pip install nltk[tests]` then testing dependencies for released NLTK will be installed, not dependencies needed to run tests for NLTK master. 

You can also run `pip install .[tests]` at source checkout; it makes sense if the goal is to install nose, coverage and pylint. Installing tox doesn't make much sense to me, because by running `pip install .[tests]` you're also installing NLTK itself, which is not required to run tox tests.

`pip install tox` and `pip install -r requirements-test.txt` look like more straightforward solutions. As you said, there is usually no reason for users to run `pip install nltk[tests]`.

fievelk:
@kmike It would still be possible to run `pip install -e .[tests]` in order to install test dependencies for a local NLTK repo (up-to-date with master) under development mode. However I don't know if this is a common practice.

So I agree with you: it would probably be better to remove `tests` from `extras_require` and provide a `requirements-test.txt` file. However, should we populate this file with all the dependencies that are needed to execute (almost) all tests? E.g. `twython`, `pyparsing`, etc.? Or should we only provide the basic `nose`, `pylint` and `coverage` instead?

dimazest:
I probably makes sense to list in `requirements-test.txt` only the tools needed to run tests.

If I refactor a particular module of NLTK and want to run its tests, then I probably don't want to install unrelated dependencies.

Then a possible development deployment could be
```
git clone
cd nltk
pip install -e .  # or .[whatever-is-needed] or .[all]
pip install -r requirements-test.txt
# hack
# run specific tests with runtests.py
# finish hacking
# run all tests with tox
# commit and push
```

I use `tox` to test several python versions to make sure that I didn't break anything.

fievelk:
@dimazest I usually use `tox` even to run specific tests. For example `tox -e py27 <filename.py>`.
I still think it would not be too bad to include `tox` together with `nose`, `coverage` and `pylint` :)

dimazest:
right, if tox is used then the `pip install` steps are unnecessary. Well, tox has to be installed, and that's it.

stevenbird:
A big Grazie and Спасибо to @fievelk, @kmike, and @dimazest!

Issue #1539, opened 2016-12-09
nltk.translate.bleu_score gives false result when ngram larger than maximum ngrams of given sentence


StarWang:
Given weight = [0.25, 0.25, 0.25, 0.25] (default value),
sentence_bleu([['a', 'b', 'c']], ['a', 'b', 'c']) = 0 
While sentence_bleu([['a', 'b', 'c']], ['a', 'b', 'd']) = 0.7598
Obviously the previous score should be larger than the latter, or both scores should be 0

alvations:
Which version of the code are you using?

```
$ python
>>> import nltk
>>> nltk.__version__
'3.2.1'
```

The BLEU implementation has been just recently fixed with #1330 resolved. If you're using the `develop` branch of `nltk`, this should be the output:

```python
>>> import nltk
>>> from nltk import bleu
>>> ref = hyp = 'abc'
>>> bleu([ref], hyp)
1.0
>>> from nltk import bleu
>>> ref, hyp = 'abc', 'abd'
>>> bleu([ref], hyp)
0.7598356856515925
```

Since a string is a list of chars and `nltk` imports the `sentence_bleu()` to the top-level imports,  the code above is the same as:


```python
>>> from nltk.translate.bleu_score import sentence_bleu
>>> sentence_bleu([['a', 'b', 'c']], ['a', 'b', 'c'])
1.0
>>> sentence_bleu([['a', 'b', 'c']], ['a', 'b', 'd'])
0.7598356856515925
```

To install the latest `develop` branch, try:

```
pip install https://github.com/nltk/nltk/archive/develop.zip
```

(**Do note that the develop branch is subjected to _more unexpected bugs_ and it is recommended that users install the `master` or official release**)

----

On a related note but not directly involved with the current `nltk` implementation of `bleu`, the previous implementation without the #1330 fix is subjected to the same flaws of the popular `multi-bleu.perl`. Maybe you might find it interesting to know why it returned 0 without the recent fix: https://gist.github.com/alvations/e5922afa8c91472d25c58b2d712a93e7

StarWang:
Thanks @alvations . The original version of nltk I used was 3.2. I have updated it to 3.2.1 now and it's now raising ZeroDivisionError. And I used Python 3.5.2

alvations:
The only stable version of BLEU is in the `develop` branch. Please wait for it to be release in NLTK 3.2.2 or install the `develop` branch (but do note that the development branch might be subjected to untested bugs).

[...]

PR #1545, opened 2016-12-14
Hotfix/bug on bleu score

liuq901:
The total bleu score is the geometric mean of n-gram count. If one n-gram count equals zero, the bleu score should be zero.
The original code will ignore the zero value to avoid log(0.0), but it will return a non-zero value, I think it is wrong.

The uni-gram check is also wrong. Even with no unigrams, the fraction can be non-zero due to the smoothing function.

Sorry for my poor English

### DIFF
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -166,19 +166,18 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     if p_numerators[1] == 0:
         return 0
 
-    if not smoothing_function: # No smoothing, values remain as Fractions.
-        s = (w * math.log(p_i) for i, (w, p_i) in enumerate(zip(weights, p_n))
-             if p_i.numerator != 0)
-    else: # Smoothen the modified precision.
-        # Note: smoothing_function() may convert values into floats;
-        #       it tries to retain the Fraction object as much as the
-        #       smoothing method allows.
+    if smoothing_function:
+        # Smoothen the modified precision
         p_n = smoothing_function(p_n, references=references,
                                  hypothesis=hypothesis, hyp_len=hyp_len)
-        s = (w * math.log(p_i) for i, (w, p_i) in enumerate(zip(weights, p_n)))
 
-    return bp * math.exp(math.fsum(s))
+    s = (w * math.log(p_i) for i, (w, p_i) in enumerate(zip(weights, p_n)))
 
+    try:
+        return bp * math.exp(math.fsum(s))
+    except ValueError, e:
+        # ValueError means log(0.0). Return 0
+        return 0.0
 
 def modified_precision(references, hypothesis, n):
     """
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -205,6 +205,7 @@
 - J Richard Snape
 - Tsolak Ghukasyan
 - Prasasto Adi
+- Qi Liu
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer


alvations:
Could you explain the problem further with some test cases / examples and why the changes you've made would fix that?

~~From a quick look, your changes will revert the fixes from #1330 that tries to emulate ['`mteval-v13a.pl`](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl) to emulate the original BLEU scores from [the original paper](www.aclweb.org/anthology/P02-1040.pdf). I think your commits would also break fringe cases test in `nltk.test.unit.translate.test_bleu.py`.~~

----

> The total bleu score is the geometric mean of n-gram count. If one n-gram count equals zero, the bleu score should be zero.

Yes, the code point at https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L163 , checks that. Although this is not described in the original BLEU paper, this is a logical thing to do. There's really no need to check for other order of ngrams and end up in math domain error later on when summing the geometric mean of the modified ngram precision.

> The original code will ignore the zero value to avoid log(0.0), but it will return a non-zero value, I think it is wrong.

Is it wrong because it's not emulating the original BLEU paper? We're not following the original BLEU paper because it didn't cover many fringe cases, which the MT community has found to be perplexing, esp. when using BLEU at sentence level, so the historical account of how BLEU has changed can be found on https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L17

> The uni-gram check is also wrong. Even with no unigrams, the fraction can be non-zero due to the smoothing function.

Please explain further on this claim. Without the unigram counts, there should be 0 counts on all higher n-gram orders. E.g. if the hypothesis sentences is `this is a foo bar sentence' and the reference sentence is 'the cats are on the mat', there should be no n-gram overlaps which should naturally return n-gram of zero:


```python
>>> from nltk import everygrams, ngrams

# 1 grams.
>>> hyp = list(ngrams('this is a foo bar sentence'.split(), n=1))
>>> ref = list(ngrams('the cats are on the mat'.split(), n=1))
>>> set(hyp).intersection(ref)
set([])

# 4 grams.
>>> hyp = list(everygrams('this is a foo bar sentence'.split(), 1, 4))
>>> ref = list(everygrams('the cats are on the mat'.split(), 1, 4))
>>> hyp
[('this',), ('is',), ('a',), ('foo',), ('bar',), ('sentence',), ('this', 'is'), ('is', 'a'), ('a', 'foo'), ('foo', 'bar'), ('bar', 'sentence'), ('this', 'is', 'a'), ('is', 'a', 'foo'), ('a', 'foo', 'bar'), ('foo', 'bar', 'sentence'), ('this', 'is', 'a', 'foo'), ('is', 'a', 'foo', 'bar'), ('a', 'foo', 'bar', 'sentence')]
>>> ref
[('the',), ('cats',), ('are',), ('on',), ('the',), ('mat',), ('the', 'cats'), ('cats', 'are'), ('are', 'on'), ('on', 'the'), ('the', 'mat'), ('the', 'cats', 'are'), ('cats', 'are', 'on'), ('are', 'on', 'the'), ('on', 'the', 'mat'), ('the', 'cats', 'are', 'on'), ('cats', 'are', 'on', 'the'), ('are', 'on', 'the', 'mat')]
>>> set(hyp).intersection(ref)
set([])
```

There would be cases where there are 0 unigram matches but still an okay translation, and I think those cases are the ones that you wanted to cover in your PR:

```python
>>> ref = list(ngrams('the cats are on the mat'.split(), n=1))
>>> hyp = list(ngrams('those feline at carpet'.split(), n=1))
>>> set(hyp).intersection(ref)
set([])
```

I suppose the hypothesis above should get some credit although there's zero matches. But I think METEOR would be a better metric for such cases. It's because if we allow this case of unigram match to slip though and get a smoothed score, non-related sentence would also be smoothed. But I'm not sure which is a better approach, allow all hypothesis to have a non-zero score (i.e. smooth every possible sentence) or be tough and enforce zero scores for no unigram matches (i.e. allow zero BLEU score). 

> Even with no unigrams, the fraction can be non-zero due to the smoothing function.

In the case that there are no unigram matches, I don't think we should smooth the BLEU score at all. This is partly because there is no reason to do at a corpus level. At a sentence level, when there are no unigram matches, assigning a smoothed BLEU score would be too permissive.

Furthermore, some of the smooth methods summarized in Chen and Cherry (2014), requires the previous order of ngrams to have non-zero counts, e.g. [method 3](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L475) or [method 6](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L529). 

Possibly the hard constraint for skipping the smoothing if there's no unigram counts is wrong. I may not be the best person to vouch for this, perhaps someone more experienced with MT research should advise us whether the fixes you suggested are more appropriate. 

----

You might find the discussion on #1539 related to this issue.

nltk/translate/bleu_score.py
```diff
@@ -166,18 +166,18 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
...
-        s = (w * math.log(p_i) for i, (w, p_i) in enumerate(zip(weights, p_n)))
 
-    return bp * math.exp(math.fsum(s))
+    s = (w * math.log(p_i) for i, (w, p_i) in enumerate(zip(weights, p_n)))
+
+    try:
+        return bp * math.exp(math.fsum(s))
+    except ValueError, e:
+        # ValueError means log(0.0). Return 0
+        return 0.0
```
alvations, line 181: This try-except with `ValueError` should not happen. The purpose of the smoothing was such that the summation over the logs don't contain any 0 values.  Although the default smoothing is no-smoothing, and to avoid the `log(0.0)`, the if condition at https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L170 would skip that order of ngram with 0 counts. So that should still not result in `ValueError`.

Could you write some test cases or examples that cause the `ValueError`? 

Also, with your PR could you help to run the unittest to check that it doesn't break other parts of the BLEU:

```bash
~/git-stuff/nltk$ ls
AUTHORS.md              LICENSE.txt             RELEASE-HOWTO.txt       nltk/                   pip-req.txt             tox.ini
CONTRIBUTING.md         MANIFEST.in             coverage.xml            nosetests.xml           setup.cfg               unittest.out
ChangeLog               Makefile                jenkins-job-config.xml  nosetests_scrubbed.xml  setup.py                web/
INSTALL.txt             README.md               jenkins.sh*             output                  tools/
~/git-stuff/nltk$ python -c 'import nltk; nltk.download('wmt15_eval')'
~/git-stuff/nltk$ python -m unittest nltk.test.unit.translate.test_bleu
..........
Ran 10 tests in 0.631s

OK
```

liuq901:
I have some commits to delete the change about unigram check.

The problem is that, for example, we have

image

The hyp1 is better than hyp2, but if we calculate the bleu score with weights = (0.5, 0.5), the original code gives 

image

the hyp2 get a higher score.

This because the 2-gram counts of hyp2 is 0, the original code will ignore it only calculate the 1-gram score, it is unfair to other sentence which count 1-gram and 2-gram score together.

If all the n-gram counts in [p_n](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L160) are non-zero, my code and the original code are the same. If some values in p_n equals zero, the original code will **IGNORE** it, this will give a very high bleu score just like my example. I think the right way is return 0. So my code doesn't have the zero value check at [https://github.com/liuq901/nltk/blob/hotfix/bug-on-bleu_score/nltk/translate/bleu_score.py#L174](https://github.com/liuq901/nltk/blob/hotfix/bug-on-bleu_score/nltk/translate/bleu_score.py#L174). Instead I catch the exception to return 0.

And my code passed the unit test.
image

alvations:
I don't think returning zero in this example is the right thing to do. `hyp2` does have a unigram match so the score must be non-zero. But You're right that `hyp2` should not a higher score than `hyp1` with uniform weights. 

Imagine you have a reference and hypothesis as follows, with the fixes you suggested, `bleu([ref], hyp)` would have returned 0 but that shouldn't be the case:

```
>>> from random import shuffle
>>> ref = range(10)
>>> hyp = range(10)
>>> shuffle(hyp)
>>> ref
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> hyp
[1, 0, 6, 3, 2, 7, 9, 5, 4, 8]
```

Maybe someone else more experienced in MT should advise on this. I would just enforce smoothing and possibly setting default to always smooth might be the right thing to do when there is a unigram count but no higher ngram counts. 

----

About the commits to undo your deletes of the unigram counts. Please rebase your git branch to the main develop branch and then recommit your changes without the deletes and re-adding of the code you've deleted.

liuq901:
I am also not familiar with MT. But some people says one disadvantage of bleu score is that if the sentence is short, maybe 3-gram and 4-gram will have zero counts and will make the final bleu score to 0. This means even hyp2 have a unigram match, the total score is still zero due to it is a geometric means. I think the right way to avoid this is use appropriate weights and smoothing function. 

And I will rebase my commits. Thanks

alvations:
> But some people says one disadvantage of bleu score is that if the sentence is short, maybe 3-gram and 4-gram will have zero counts and will make the final bleu score to 0. This means even hyp2 have a unigram match, the total score is still zero due to it is a geometric means. 

This is not logical, especially when distortion and reordering is taken into account. Within MT applications or any other sequential tasks that needs to use BLEU, assigning zero for hypothesis that comprises n-grams order below the n set by the user is not rational. E.g., imagine the sentence:

```python
>>> from nltk import bleu
>>> src = "eat the sushi".split()
>>> ref = ["寿司", "を", "食べろ"] # "the", "object-marker", "eat"
>>> hyp = ["食べろ" ,"寿司"] # "eat", "sushi"
>>> bleu([ref], hyp)
0.6065306597126334
```

In this case the unigram chain is totally adequate though not fluent. And giving it a zero BLEU should be wrong.

P/S: Don't just look at the geometric mean formula in the original BLEU paper, even the `mteval-v13a.perl` heavily edited their implementation. The users of the functions will end up raising more issues if the use cases of our BLEU implementations don't fit the typical usage =)

liuq901:
But according to the original formula from [paper](http://www.aclweb.org/anthology/P02-1040.pdf), 
image
when some p_i = 0, the sum in the parenthese is -inf, then exp(-inf) = 0.

I think for short sentence, calculate high order n-gram is not suit. I think give 0 bleu score have no problem, because it can be non-zero by adjust the weights and smoothing function parameters. I think we should not accept a wrong bleu score even if it make sense.

alvations:
> P/S: Don't just look at the geometric mean formula in the original BLEU paper, even the mteval-v13a.perl heavily edited their implementation. The users of the functions will end up raising more issues if the use cases of our BLEU implementations don't fit the typical usage =)

It's strict implementation that follows the paper vs functional implementation that suits typical usage. We need to arbitrate between these. I will go for the latter.

Please take a look at https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L17

liuq901:
Sorry I cannot understand perl script. I think the problem is that the default setting of BLEU count up to 4-gram and have no smoothing function. 

In your case, the value makes sense because it ignore the value of 2-gram to 4-gram and only count the 1-gram score. But it is not known by the user that the code **ONLY** calculate the 1-gram value, by the documents it should calculate the value form 1-gram to 4-gram.

In my case, one sentence calculate 1-gram and 2-gram, another only calculate the 1-gram. It is unfair. And 
I believe this is a bigger problem. Because the original code can give a higher bleu score on bad sentence. And my code may make some good sentence to 0, but bad one is also 0 in this case.

I understand your issue. Some good examples may get 0 value due to it is too short. But the key problem is using the same weights with no smoothing function, how to avoid the unfair behavior. Could you give me any other comments?

alvations:
When using BLEU, I think it's best to smooth by default with [method4](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L497), esp. for short translations and esp. for the examples raised in this issue:

```python
    def method4(self, p_n, references, hypothesis, hyp_len):
        """
        Smoothing method 4:
        Shorter translations may have inflated precision values due to having
        smaller denominators; therefore, we give them proportionally
        smaller smoothed counts. Instead of scaling to 1/(2^k), Chen and Cherry
        suggests dividing by 1/ln(len(T)), where T is the length of the translation.
        """
        incvnt = 1
        for i, p_i in enumerate(p_n):
            if p_i.numerator == 0 and hyp_len != 0:
                p_n[i] = incvnt * self.k / math.log(hyp_len) # Note that this K is different from the K from NIST.
                incvnt+=1
        return p_n
```

I'm not sure what should be the default we should set from NLTK though.

liuq901:
All the problems are because of 0 n-gram counts. In the original code, ignore it will lead to wrong result. In my code, return 0.0 will lead to score which not make sense.

I think using a smoothing function is the best way.
But change the default value may cause other problems.

What about throwing an exception when we found 0 n-gram counts? This can help users to know what happened and suggest them to use some smoothing function in the error message

alvations:
After some more testing, here's a clear example of why this bug is quite serious. It seems like discarding higher ngram when it don't exist leads to some problem with regards to distortion (i.e. word ordering issues). 

With current BLEU implementation

```python
>>> from nltk import bleu
>>> ref = 'let it go'.split()
>>> hyp = 'let go it'.split()
>>> bleu([ref], hyp)
1.0
```

But the fix in this PR wouldn't really solve the problem since it will return:


```python
>>> from nltk import bleu
>>> ref = 'let it go'.split()
>>> hyp = 'let go it'.split()
>>> bleu([ref], hyp)
0.0
```

Distorted hypothesis should have a non-zero score but surely not a 1.0 score. We'll have to look at how `mteval-v13a.pl` and Moses' BLEU handles such cases and get back to this PR / issue.

stevenbird:
Thanks @liuq901  and @alvations – I agree #1554 is a good solution.

PR #1549, opened 2016-12-19
fix bug about utf-8 space

xinyuan-liu:
replace b'\xa0' with b'\x00\xa0'.
'\x00\xa0' is a white space in utf-8 but not 'x\a0'.
The old version may trigger a 'utf-8 decode error' in python when the string contains a utf-8 character which has a byte a0, such as '传' in chinese.

eg:
parser=StanfordParser(model_path='PATH_TO_chineseFactored.ser.gz')
parser.parse('射雕英雄传 是 谁 的 作品')

### DIFF
--- a/nltk/parse/stanford.py
+++ b/nltk/parse/stanford.py
@@ -216,7 +216,7 @@ class GenericStanfordParser(ParserI):
                                       stdout=PIPE, stderr=PIPE)
                 
             stdout = stdout.replace(b'\xc2\xa0',b' ')
-            stdout = stdout.replace(b'\xa0',b' ')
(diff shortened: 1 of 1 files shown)


dimazest:
Looks good! Could you please write a test for this. Tests will make sure that the issue will not be reintroduced in the future.

fievelk:
@xinyuan-liu Thank you for your contribution. You can find some guidelines about tests here: [NLTK Testing][1].

[1]: https://github.com/nltk/nltk/blob/develop/web/dev/local_testing.rst

stevenbird:
Merging without the test, sorry :)

PR #1550, opened 2016-12-20
Fix typo in `quote_counts` variable

alvations:
The variable name should be `quote_counts`, as declared/initialized in https://github.com/nltk/nltk/blob/develop/nltk/tokenize/moses.py#L472. 

The "s" is missing the variable following the initialization.

----

Also the quotation counts are missing the grave accent quotes and the double single quotes.  These are added to the `quote_chars` variable.

### DIFF
--- a/nltk/tokenize/moses.py
+++ b/nltk/tokenize/moses.py
@@ -540,15 +540,15 @@ class MosesDetokenizer(TokenizerI):
                 normalized_quo = token
                 if re.match(r'^[„“”]+$', token):
                     normalized_quo = '"'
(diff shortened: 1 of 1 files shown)


fievelk:
@alvations I am thinking that these features are not currently tested then, since we didn't have related failures. It would be great if you could create some tests for them, if you have time.

alvations:
Yes, I'm creating the tests and also more fixes. I should have added WIP in the previous post =)

Just realized that there are several "mistakes" in the implementation that is specific to how Python treats strings (very much different from Perl).

PR #1555, opened 2016-12-23
LazyIteratorList bugfix/simplification

nschneid:
Bugs in the `iterate_from()` method meant it was possible to get spurious `IndexError`s when accessing list elements. Removed the `cache_limit` option, which greatly simplifies the implementation.


stevenbird:
Nice clean up, thanks @nschneid

Issue #1554, opened 2016-12-23
How to handle BLEU scores for ngrams where n<4 in NLTK? 


alvations:
From #1545, BLEU is buggy for ngrams where the n<4. 

One simple way is to clip the weights distribution such that the `weights=(0.25, 0.25, 0.25, 0.25)` redistributes uniformly with respect to the highest ngram order from the hypothesis or reference. 

But we're not sure whether that's the most appropriate solution to fix that. 

Meanwhile, while smoothing functions work fine when reference length is n>=4, it goes haywire when n<4 too. 

----

Without smoothing, NLTK's BLEU is overly-optimistic with the condition on https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L170:

```
>>> from nltk import bleu
>>> ref = 'let it go'.split()
>>> hyp = 'let go it'.split()
>>> bleu([ref], hyp)
1.0
```

With smoothing, the BLEU goes haywire:

```python
>>> from nltk.translate.bleu_score import SmoothingFunction
>>> chencherry = SmoothingFunction()
>>> bleu([ref], hyp, smoothing_function=chencherry.method0) # No smoothing
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "nltk/translate/bleu_score.py", line 79, in sentence_bleu
    return corpus_bleu([references], [hypothesis], weights, smoothing_function)
  File "nltk/translate/bleu_score.py", line 180, in corpus_bleu
    return bp * math.exp(math.fsum(s))
  File "nltk/translate/bleu_score.py", line 178, in <genexpr>
    s = (w * math.log(p_i) for i, (w, p_i) in enumerate(zip(weights, p_n)))
ValueError: math domain error
>>> bleu([ref], hyp, smoothing_function=chencherry.method1) 
0.14953487812212207
>>> bleu([ref], hyp, smoothing_function=chencherry.method2) 
0.537284965911771
>>> bleu([ref], hyp, smoothing_function=chencherry.method3) 
0.2973017787506803
>>> bleu([ref], hyp, smoothing_function=chencherry.method4)
4.876766661577289
>>> bleu([ref], hyp, smoothing_function=chencherry.method5) 
0.19245008972987526
>>> bleu([ref], hyp, smoothing_function=chencherry.method6) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "nltk/translate/bleu_score.py", line 79, in sentence_bleu
[...]

PR #1557, opened 2016-12-27
Fixing SmoothingFunction() in BLEU

alvations:
Beyond #1330, there are also issues with the smoothing functions implemented in `nltk.translate.bleu_score.py`.

In this PR, it is fixing:
 - **Method 4 smoothing function**: corrected the implementation such that it binds the outputs to the probability space [0,1] and not logspace, esp. when max n-grams order is less than 4.
 - **Use method 0 (no smoothing) by default**: This is to minimize maintaining 2 sets of code to handle fringe cases of BLEU when users don't use the smoothing function. 
 - **Added non code-breaking warnings** when users use the BLEU score implementation in NLTK without smoothing. 
 - **Added an automatic reweighing feature** to cover the cases discussed in #1554 
 - **Added info for docstring in method 6**
 - **Use ellipses for neurotically long floats outputs in doctest**

Please discuss whether these (warnings, no smoothing by default, etc.) features are the ones we want to keep in NLTK's implementation of BLEU before merging. Also what should be NLTK's default setting for BLEU?

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -123,7 +123,11 @@ class TestBLEU(unittest.TestCase):
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary who loves Mike'.split()
         self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.4729, places=4)
-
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @alvations

PR #1562, opened 2016-12-29
Use abstract base class syntax

naoyak:
Ref: #1130 

This updates most of the abstract base classes with the `@six.add_metaclass` and `@abc.abstractmethod` decorators.

TODO:
- [x] Update doctests that check for improperly instantiated ABCs


nltk/classify/api.py
```diff
@@ -38,12 +43,13 @@ class ClassifierI(object):
...
       - either ``prob_classify()`` or ``prob_classify_many()`` (or both)
     """
+    @abstractmethod
     def labels(self):
         """
         :return: the list of category labels used by this classifier.
         :rtype: list of (immutable)
         """
-        raise NotImplementedError()
+        pass
```
naoyak, line 52: `WekaClassifier` doesn't implement `labels()` - this would throw an error?

dimazest: Actually, `pass` is not needed here, the docstring is enough.

naoyak: Oh yeah, thanks. Actually I was just figuring out how to reconcile the "hard" enforcement of `abc.abstractmethod` with how how the `ClassifierI` ABC is implemented - see 121a24ca66d2913d1567e7c9e7e26af6a78c3afc

naoyak: @dimazest so is it better to remove the `pass` statements? Most examples of abstract methods I've seen elsewhere leave it in. I think it's a bit more explicit and clear.

dimazest: Personally I would not put the ``pass`` if there is a docstring. It looks redundant, in the sense that there is a code that does nothing.

fievelk:
[CI: retest]

naoyak:
Looks like some of the abstract classes are tricky - moved the simpler ones to #1639.

fievelk:
[CI: retest]

naoyak:
The straightforward classes were taken care of with #1639 #1707 - closing.

Issue #1577, opened 2017-01-04
Tokenize `xml_escape` tests are failing


fievelk:
Tests for `tokenize.util.xml_escape` are currently failing, see [Jenkins CI test](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py27-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.tokenize/util/xml_escape).

stevenbird:
@alvations – would you please look into this?

stevenbird:
@alvations – looks like we still have this error on the [CI server](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.tokenize/util/xml_escape/)

stevenbird:
Thanks @alvations

PR #1578, opened 2017-01-05
Checks for string equality to fix nltk.util.escape_xml doctest

alvations:
Fixing #1577 by checking using string literals to check for equality.


stevenbird:
Thanks @alvations.
What do you think about also demonstrating the use of `xml_escape` itself, in the docstring for `xml_escape`?

I think the current documentation for `escape` is there to justify the need for `xml_escape` (and this justification could go into a Python comment before the function). Users expect to see usage examples in docstrings.

alvations:
@stevenbird sorry I was a little busy for the past few days. I missed out the doctest for xml_escape. Sorry about it. Added in the PR #1592 and also corrected one of the escape symbol.

PR #1583, opened 2017-01-07
Fix unclosed files in porter test causing ResourceWarnings

ExplodingCabbage:
Without this, I get:

    test_stem.py:74: ResourceWarning: unclosed file <_io.BufferedReader name='/home/mark/nltk_data/stemmers/porter_test/porter_martin_output.txt'>
    .open(encoding='utf-8')


ExplodingCabbage:
(Not sure why I never saw these warnings when I first authored these tests.)

stevenbird:
Thanks @explodingcabbage!

Issue #1584, opened 2017-01-07
StanfordTokenizer failure


fievelk:
We are experiencing [`StanfordTokenizer` failures on our CI][1]:
```bash
Failed example:
    StanfordTokenizer().tokenize(s)
Exception raised:
    Traceback (most recent call last):
      File "/scratch/jenkins/python/python-3.4.2-x86_64/lib/python3.4/doctest.py", line 1324, in __run
        compileflags, 1), test.globs)
      File "<doctest nltk.tokenize.stanford.StanfordTokenizer[2]>", line 1, in <module>
        StanfordTokenizer().tokenize(s)
      File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py34-jenkins/jdk/jdk8latestOnlineInstall/nltk/tokenize/stanford.py", line 68, in tokenize
        return self._parse_tokenized_output(self._execute(cmd, s))
      File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py34-jenkins/jdk/jdk8latestOnlineInstall/nltk/tokenize/stanford.py", line 94, in _execute
        stdout=PIPE, stderr=PIPE)
      File "/scratch/jenkins/workspace/pull_request_tests/TOXENV/py34-jenkins/jdk/jdk8latestOnlineInstall/nltk/internals.py", line 134, in java
        raise OSError('Java command failed : ' + str(cmd))
    OSError: Java command failed : ['/opt/jdk/jdk8.latest/bin/java', '-mx1000m', '-cp', '/home/jenkins/third/stanford-postagger/stanford-postagger-3.7.0-sources.jar:/home/jenkins/third/stanford-postagger/stanford-postagger-3.7.0.jar:/home/jenkins/third/stanford-postagger/stanford-postagger-3.7.0-javadoc.jar:/home/jenkins/third/stanford-postagger/stanford-postagger.jar:/home/jenkins/third/stanford-postagger/lib/slf4j-simple.jar:/home/jenkins/third/stanford-postagger/lib/slf4j-api.jar', 'edu.stanford.nlp.process.PTBTokenizer', '-charset', 'utf8', '/tmp/tmptx_7hdjm']
```

@alvations I think that you have had some experience with this kind of problems previously. Could you please verify? I have the feeling that this could be related to some changes in the Stanford jar, but I could be wrong.

[...]

Issue #1509, opened 2016-11-18
default verbosity


stevenbird:
90+% of verbose flags in the codebase default to `False` while less than 10% default to `True`. I think they should all default to `False` except in demonstration code.

PR #1585, opened 2017-01-08
Fix Stanford interfaces

fievelk:
Related issue: #1584

---

This PR aims to solve the problems we currently experience with Stanford tokenizers and taggers.
We use too many arguments in our classpath, and some of them (i.e. the `slf4j.jar` paths) are breaking current functionalities. This new problem is also related to some changes recently made to the Stanford CoreNLP codebase.

Note: I also took the chance to change some verbosity levels to `False`, as requested in #1509.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -54,14 +54,22 @@ if [[ ! -d $senna_folder_name ]]; then
 fi
 
 # Setup the Enviroment variable
+export CLASSPATH=$(pwd)"/${stanford_corenlp_package_name}"
+export CLASSPATH=${CLASSPATH}:$(pwd)"/${stanford_parser_package_name}"
+export CLASSPATH=${CLASSPATH}:$(pwd)"/${stanford_tagger_package_name}"
 export STANFORD_CORENLP=$(pwd)'/stanford-corenlp'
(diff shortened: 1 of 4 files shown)


fievelk:
The last changes solved the failures we were experiencing on Jenkins CI. I specified a `CLASSPATH` environment variable for Stanford jars, and I also fixed the `STANFORD_MODELS` variable which was apparently incorrect.
Please let me know if there were some reasons behind the choice of not setting these two variables.

Problems related to the `StanfordNeuralDependencyParser` are currently ignored by this PR. I will open a new issue for them.

I am removing the "work in progress" status.

P.S. As stated in stanfordnlp/CoreNLP#334:
>  In v3.6 [of the Stanford Tagger] you had to have the slf4j logging jars, but we decided that was a big mistake.

I have updated our interfaces so that they work with the latest Stanford release (`v3.7`). This means that they will presumably not work anymore for `v3.6`, since `slf4j.jar` won't be included in the classpath anymore. I think that this is a good solution for the moment, at least until a new and compatible version of `slf4j.jar` is shipped with the Stanford software.

stevenbird:
Thanks @fievelk - this is super helpful

Issue #1586, opened 2017-01-10
StanfordNeuralDependencyParser memory issues


fievelk:
**In brief:** should we raise the amount of memory used by default by `StanfordNeuralDependencyParser`?

---

I am experiencing some problems with `StanfordNeuralDependencyParser`. Steps to reproduce the issue:
```python
from nltk.parse.stanford import StanfordNeuralDependencyParser
dep_parser=StanfordNeuralDependencyParser()
dep_parser.raw_parse("The quick brown fox jumps over the lazy dog.")
```

First I get this error:
> Exception in thread "main" java.lang.OutOfMemoryError: Java heap space

Then, if I allocate more heap space using `java_options='-mx2000m'` and try again:
> java.lang.OutOfMemoryError: GC overhead limit exceeded

If I increase it again (e.g. to `java_options='-mx4g'`) I finally get a result:
```python
from nltk.parse.stanford import StanfordNeuralDependencyParser
dep_parser=StanfordNeuralDependencyParser()
result = dep_parser.raw_parse("The quick brown fox jumps over the lazy dog.")
next(result)
<DependencyGraph with 11 nodes>
```

In fact, [Stanford Parser FAQ][1] contains the following useful advice:
> The parser uses considerable amounts of memory. If you see a java.lang.OutOfMemoryError, you either need to give the parser more memory or to take steps to reduce the memory needed. (You give java more memory at the command line by using the -mx flag, for example -mx500m.

They also provide several examples using `-mx4g` or `-mx12g`, so I suppose that we actually need to expect a lot of memory usage.

I think we therefore need to change our code in order to expect a higher memory usage for `StanfordNeuralDependencyParser`.

[1]: http://nlp.stanford.edu/software/parser-faq.shtml#k

fievelk:
Fixed by #1588

PR #1588, opened 2017-01-12
Fix StanfordNeuralDependencyParser tests

fievelk:
# WIP

Reference: #1586 

---

This PR aims to solve `StanfordNeuralDependencyParser` test failures. It is supposed to be structured in two parts:

- [x] Fix java memory allocation issues in doctests

The `StanfordNeuralDependencyParser` needs a lot of memory in order to work. However, changing the default value of an argument in the constructor of a subclass is not very straightforward, especially when using `*args` and `*kwargs` (our case). I therefore decided to keep the current constructors and improve the documentation, fixing tests at the same time. This is done specifying the argument `java_options='-mx3g'` when instantiating the `StanfordNeuralDependencyParser`:
```python
dep_parser=StanfordNeuralDependencyParser(java_options='-mx3g')
```

### DIFF
--- a/nltk/parse/stanford.py
+++ b/nltk/parse/stanford.py
@@ -62,8 +62,8 @@ class GenericStanfordParser(ParserI):
         )
 
         #self._classpath = (stanford_jar, model_jar)
-        
-        # Adding logging jar files to classpath 
+
+        # Adding logging jar files to classpath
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @fievelk

PR #1590, opened 2017-01-14
Fix failing doctests in Python 2.7

safwank:
I've modified the code to use `try-finally` instead. 

What I don't fully understand is why other calls to `with` aren't failing. Care to explain?


stevenbird:
Thanks @safwank. I don't understand what the problem was here!

PR #1592, opened 2017-01-16
Added doctest for xml_escape and corrected escape from ampersand symbol

alvations:
Spill over from #1577 and #1578


stevenbird:
Thanks @alvations!

Issue #1489, opened 2016-10-26
Reimplementing Monolingual Word Aligner in NLTK


alvations:
The [Monolingual Word Aligner](https://github.com/ma-sultan/monolingual-word-aligner) is an important tool in the [Semantic Textual Similarity](http://alt.qcri.org/semeval2017/task1/) task.

@amalinovskiy created a fork at https://github.com/amalinovskiy/upf-cobalt but the Stanford parser and JSON RPC is still based on outdated code.

**May I suggest that we try to re-implment it within NLTK?** The version on @ma-sultan repo is rather old and it's based on an older implementation some other Stanford CoreNLP wrapper outside of NLTK and also older NLTK versions for tokenization, etc. 

Should this be under `nltk.sem`?

alvations:
Similar to previous feature sprint, here's mini-roadmap that we can follow and anyone can join to help out:

 - #1593: Write a corpus reader for The Paraphrase Database (PPDB), more specifically [PPDB-1.0-XXXL]( https://raw.githubusercontent.com/ma-sultan/monolingual-word-aligner/master/Resources/ppdb-1.0-xxxl-lexical.extended.synonyms.uniquepairs)
 - Write a corpus reader for Semantic Textual Similarity (STS) English task 2012 - 2016 datasets
 - Interface Stanford CoreNLP tools with #1249
 - Alignment at nouns, adjectives, adverbs, entities (i.e. https://github.com/ma-sultan/monolingual-word-aligner/blob/master/aligner.py)
   - Algorithm 1: *depContext()*
   - Algorithm 2: *textContext()*
   - Algorithm 3: *cwDepAlign()*
   - Algorithm 4: *cwTextAlign()*
 - High level interface that combines the alignments (i.e. Algorithm 5: *align()*)
 - Write test to verify that the outputs are the same as Sultan et al. (2014) implementation. Possibly just check against outputs from [Stasis repo](https://github.com/alvations/stasis)

rameshjes:
![image](https://user-images.githubusercontent.com/25647309/32292715-4025829c-bf41-11e7-9eba-eb00b5012c17.png)

Hi guys, 
I am Ramesh Kumar, currently doing my research and development project on automatic short answer grading. For that, I am using Sultan et al.(2014) word aligner and implementing in NLTK. So far, I  have completed till Named Entities module as shown in block diagram. I have also compared my outputs with word aligner by Sultan, so far everything is same.   For other modules, I will keep you guys updating.   

[...]

PR #1593, opened 2017-01-16
Add a corpus reader to the MWA subset of PPDB 

alvations:
Moving on with #1489 , the corpus reader from the Sultan et al. (2014) subset of the PPDB is ready.

The original data source is on https://raw.githubusercontent.com/ma-sultan/monolingual-word-aligner/master/Resources/ppdb-1.0-xxxl-lexical.extended.synonyms.uniquepairs and it's a post-processed dataset from the full [PPDB](http://www.cis.upenn.edu/~ccb/ppdb/)

@stevenbird can you help to upload [this zipball](https://drive.google.com/file/d/0Bzz3wLacJ7WoN2pIOFBHanVCbk0/view?usp=sharing), so that it's accessible on `nltk_data`. I'm not sure where it should go to so now the code is putting it at `nltk_data/misc/mwa_ppdb`. 

I'll move on the other steps in the mini-roadmap while reviewing is done on this PR.

### DIFF
--- a/nltk/corpus/__init__.py
+++ b/nltk/corpus/__init__.py
@@ -285,6 +285,9 @@ nonbreaking_prefixes = LazyCorpusLoader(
 perluniprops = LazyCorpusLoader(
     'perluniprops', UnicharsCorpusReader, r'(?!README|\.).*', nltk_data_subdir='misc', encoding='utf8')
 
+mwa_ppdb = LazyCorpusLoader(
+    'mwa_ppdb', MWAPPDBCorpusReader, r'(?!README|\.).*', nltk_data_subdir='misc', encoding='utf8')
+
 
 def demo():
     # This is out-of-date:
--- a/nltk/corpus/reader/__init__.py
+++ b/nltk/corpus/reader/__init__.py
@@ -142,5 +142,6 @@ __all__ = [
     'MTECorpusReader', 'ReviewsCorpusReader', 'OpinionLexiconCorpusReader',
     'ProsConsCorpusReader', 'CategorizedSentencesCorpusReader',
     'ComparativeSentencesCorpusReader', 'PanLexLiteCorpusReader',
-    'NonbreakingPrefixesCorpusReader', 'UnicharsCorpusReader'
+    'NonbreakingPrefixesCorpusReader', 'UnicharsCorpusReader',
+    'MWAPPDBCorpusReader',
 ]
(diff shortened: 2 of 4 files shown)


stevenbird:
@alvations: I see you use the unicode string notation `u""`. I think our use of the python_2_unicode_compatible decorator makes this unnecessary.

alvations:
Would it be simpler to use `from __future__ import unicode_literals` instead?

That way we don't have to manually decorate every class/function with the `python_2_unicode_compatible`, i.e.

```python
$ python 
Python 2.7.10 (default, Jul 30 2016, 19:40:32) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
>>> print '\u0f3a'
\u0f3a
>>> from __future__ import unicode_literals
>>> print '\u0f3a'
༺
```

Related to `nltk.compat` usage in `nltk/corpus/reader/wordlist.py`, I think the `nltk.compat.string_type` could also be replaced by `six.text_type`.

nschneid:
I have been burned by `unicode_literals` in the past, and would advise against adopting it lightly. Some commentary here: http://python-future.org/unicode_literals.html

alvations:
`nltk/test/unit/test_corpora.py` is already using it, I have changed them to `nltk.compat. python_2_unicode_compatible` decorators instead.

stevenbird:
Thanks @alvations, @nschneid

rameshjes:
Hi @alvations,
 
Is paraphrase database used by Sultan et al.(2014) added in nltk? If yes, how can I access it?

Thanks

alvations:
@rameshjesswani 

There's quite a lot of work that needs to be done for #1489 and hopefully, someday someone will pick it up again but for now, you can access the PPDB subset used in Sultan et al. (2014) with:


```python
>>> from nltk.corpus import LazyCorpusLoader
>>> from nltk.corpus.reader.wordlist import MWAPPDBCorpusReader
>>> mwa_ppdb = LazyCorpusLoader( 'mwa_ppdb', MWAPPDBCorpusReader, r'(?!README\|\.).*', nltk_data_subdir='misc', encoding='utf8')
>>> mwa.ppdb.words()


>>> for pairs in mwa_ppdb.words():
...     word1, word2 = pairs.split('\t')
...     print(word1, '->', word2)
```

rameshjes:
thanks @alvations.  I am currently re-implementing Sultan et al.(2014) word aligner in NLTK, so that is why I want to use paraphrase database. I already posted my half-completed diagram here [#1489](https://github.com/nltk/nltk/issues/1489)

PR #1598, opened 2017-01-19
Fix/chunk tag pattern

chmeyer:
Fix 1597. Allow for curly bracket quantifiers in nltk.chunk.regexp.CHUNK_TAG_PATTERN.

### DIFF
--- a/nltk/chunk/regexp.py
+++ b/nltk/chunk/regexp.py
@@ -829,7 +829,7 @@ class ChunkRuleWithContext(RegexpChunkRule):
 # this should probably be made more strict than it is -- e.g., it
 # currently accepts 'foo'.
 CHUNK_TAG_PATTERN = re.compile(r'^((%s|<%s>)*)$' %
-                                ('[^\{\}<>]+',
+                                ('[^<>]+',
                                  '[^\{\}<>]+'))
 
 def tag_pattern2re_pattern(tag_pattern):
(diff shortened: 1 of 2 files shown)


fievelk:
retest this please

nltk/chunk/regexp.py
```diff
@@ -829,7 +829,7 @@ def __repr__(self):
 # this should probably be made more strict than it is -- e.g., it
 # currently accepts 'foo'.
 CHUNK_TAG_PATTERN = re.compile(r'^((%s|<%s>)*)$' %
-                                ('[^<>]+',
+                                ('([^\{\}<>]|\{\d+,?\d*\}|\{\d*,?\d+\})+',
```
stevenbird: @chmeyer: I think there's a readability issue here. The case where a comma is omitted requires us to realise the `\d+\d*` is harmless. I would rather see the cases `{m}`, `{m,}`, `{,n}`, and `{m,n}` laid out more distinctly. E.g. as `{\d+,?}|{\d*,\d+}`. What do you think?

(NB I didn't escape the braces here, for readability)

stevenbird:
@chmeyer  do you have any response to my feedback?

chmeyer:
@stevenbird fully agree to your suggestion (should yield equivalent results). Was planning to test and integrate it, but am currently away from the workspace. Feel free to do the change which might speed up things...

stevenbird:
OK @chmeyer – I'll take a look. Can you tell me, is there a reason for modifying the first regexp (line 832) but not the second (line 833)?

stevenbird:
I'm going to merge then make the stylistic change on the develop branch.

Issue #1486, opened 2016-10-24
Training on PerceptronTagger with alpino throws TypeError


alvations:
With `BigramTagger`:

``` python
>>> from nltk.corpus import alpino as alp
>>> from nltk.tag import UnigramTagger, BigramTagger
>>> training_corpus = alp.tagged_sents()
>>> unitagger = UnigramTagger(training_corpus)
>>> bitagger = BigramTagger(training_corpus, backoff=unitagger)
>>> pos_tag = bitagger.tag
>>> sent = 'NLTK is een goeda taal voor NLP'.split()
>>> pos_tag(sent)
[('NLTK', None), ('is', u'verb'), ('een', u'det'), ('goeda', None), ('taal', u'noun'), ('voor', u'prep'), ('NLP', None)]
```

But with `PerceptronTagger`:

``` python
>>> from nltk.tag import PerceptronTagger
>>> from nltk.corpus import alpino as alp
>>> training_corpus = alp.tagged_sents()
>>> tagger = PerceptronTagger(load=False)
>>> tagger.train(training_corpus)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Python/2.7/site-packages/nltk/tag/perceptron.py", line 194, in train
    random.shuffle(sentences)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/random.py", line 291, in shuffle
    x[i], x[j] = x[j], x[i]
TypeError: 'StreamBackedCorpusView' object does not support item assignment
```

alvations:
It seems like there's a need to materialize the corpus before feeding it to the `PerceptronTagger`, so that the [`shuffle` operation can be possible when using the `train()` function](https://github.com/nltk/nltk/blob/develop/nltk/tag/perceptron.py#L194):

``` python
>>> from nltk.tag import PerceptronTagger
>>> from nltk.corpus import alpino as alp
>>> training_corpus = list(alp.tagged_sents()) # Materialize the corpus.
>>> tagger = PerceptronTagger(load=True)
>>> tagger.train(training_corpus)
>>> sent = 'NLTK is een goeda taal voor het leren over NLP'.split()
>>> tagger.tag(sent)
[...]

PR #1604, opened 2017-01-26
Allow iterator in perceptron tagger training

jacksonllee:
This pull request resolves issue #1486.

Goal: To allow the training ``sentences`` for the `PerceptronTagger` class to be a *generator/iterator* of sentences as well (in addition to a list).

Because ``self._make_tagdict(sentences)`` runs at the beginning of the `train` method and iterates through the sentences, we first initialize an empty list of `self._sentences`  and make `self._make_tagdict` populate ``self._sentences`` with all the sentences. This saves the overheard of just iterating through ``sentences`` to get the list by ``sentences = list(sentences)``. Now `random.shuffle` can happily take `self._sentences`. 

With this update/fix, the following doesn't throw an error anymore:

```python
>>> from nltk.tag import PerceptronTagger
>>> from nltk.corpus import alpino as alp
>>> training_corpus = alp.tagged_sents()
>>> tagger = PerceptronTagger(load=False)
>>> tagger.train(training_corpus)
```

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -98,6 +98,7 @@
 - Pierre-François Laquerre
 - Stefano Lattarini
 - Haejoong Lee
+- Jackson Lee
 - Max Leonov
 - Chris Liechti
 - Tom Lippincott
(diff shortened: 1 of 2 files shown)


stevenbird:
thanks @jacksonllee – looks good

Issue #1623, opened 2017-02-14
nltk.set_proxy not wroks


eromoe:
Hi, I am using a proxy named shadowsocks(very very slow in China, have to use proxy), but unable to download assert through the proxy.

I am in windows 10 X64, python 2.7 latest.

I have tried either `nltk.set_proxy('http://127.0.0.1:1080')` or `nltk.set_proxy('http://127.0.0.1:1080', '', '')`, both not work.

PR #1624, opened 2017-02-14
fix set_proxy for https

eromoe:
`set_proxy` not working due to current github is over https. 
Add support.

relate to https://github.com/nltk/nltk/issues/1623.


stevenbird:
Thanks @eromoe

PR #1608, opened 2017-02-03
Dropping mwa_ppdb temporarily

alvations:

Issue #1606, opened 2017-02-03
MWAPPDB corpus fails build tests


stevenbird:
The CI server reports an [error](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.test.unit.test_corpora/TestMWAPPDB/test_entries/) with the `mwa_ppdb` corpus.

alvations:
@stevenbird sorry about the test failures on `mwa_ppdb`. It seems like it's not reading the `.zip` file correctly. Maybe we should remove all the `mwa_ppdb` code for now.

PR #1627, opened 2017-02-15
Reviving mwa_ppdb - WIP (Do not merge)

alvations:
From #1606 and #1608, it seems like `nltk` is not downloading the `mwa_ppdb`. 

The code works fine locally in Python 3.5.2:

```python
>>> from nltk.corpus.reader import *
>>> from nltk.corpus import LazyCorpusLoader
>>> import nltk
>>> nltk.download('mwa_ppdb')
[nltk_data] Downloading package mwa_ppdb to
[nltk_data]     /Users/liling.tan/nltk_data...
[nltk_data]   Unzipping misc/mwa_ppdb.zip.
True
>>> mwa_ppdb = LazyCorpusLoader(
...     'mwa_ppdb', MWAPPDBCorpusReader, r'(?!README|\.).*', nltk_data_subdir='misc', encoding='utf8')
>>> mwa_ppdb.entries()
```

But it's breaking on CI for Python3.5 on https://nltk.ci.cloudbees.com/job/nltk/642/testReport/ leading to https://nltk.ci.cloudbees.com/job/nltk/645/

I'm debugging what's wrong. I suspect that when jenkin runs the test, it's not unzipping the corpus but nevertheless at the first use of `WordListCorpusReader`, it should automatically extract the content.


alvations:
This is interesting, when looking for `mwa_ppdb` it looks for `misc/mwa_ppdb` in Python2.7 and 3.4 but in Python3.5 it was looking for `corpora/mwa_ppdb` which wouldn't be there, see https://nltk.ci.cloudbees.com/job/nltk/642/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/testReport/nltk.test.unit.test_corpora/TestMWAPPDB/test_entries/

It seems like the culprit is https://github.com/nltk/nltk_data/blob/gh-pages/collections/all-corpora.xml#L75

It was added to the `all-corpora.xml`!! But it's still strange that only Python3.5 has this issue. Let's first try removing the entry on `nltk_data`.

alvations:
This is really odd, Python 3.5 test fails while 2.7 and 3.4 does alright and it's still trying to access `corpora/mwa_ppdb` rather than `misc/mwa_ppdb`. 

@fievelk any clue(s) on why this would happen?

alvations:
There seem to be an issue with `**kwargs` behavior in Python 3.5 : https://nltk.ci.cloudbees.com/job/pull_request_tests/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/lastBuild/console

Seems like the `nltk_data_subdir` isn't recognized in Python 3.5 https://github.com/nltk/nltk/blob/develop/nltk/corpus/util.py#L57

alvations:
From https://nltk.ci.cloudbees.com/job/pull_request_tests/83/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/console , it seems like the `*args` and `**kwargs` is causing the problem because `nltk_data_subdir` was dropped when loading the `mwa_ppdb` a second time. 

It seems like it's something to do with python syntax internally, http://stackoverflow.com/questions/32521140/generator-as-function-argument when `fileids` receives a regex with `r'(?!README|\.).*'`. Possibly changing to all `kwargs`.

fievelk:
[CI: retest]

PR #1636, opened 2017-02-21
Reviving mwa_ppdb - WIP (Do not merge)

alvations:
From #1606,  #1608 and #1627, something is wrong when corpus is loading from other subdirectory in `nltk_data` other than `nltk_data/corpora`.


alvations:
I'm out of wits, why would Python 3.5 test fail while Python 3.4 and Python 2.7 works. 

The only clue is that the `nltk_sub_dir` kwargs disappeared at the 2nd import in Python3.5.

PR #1637, opened 2017-02-22
Fix #1387: clean up Tk's postscript output

### LINKED ISSUE #1387: GhostScript error when displaying inline Tree structure for Jupyter notebook
I just reinstalled my system including Jupyter notebook and nltk, and found out that the inline display of tree as png file which used to work like a charm is now broken.

First, relevant installation info
nltk (3.2.1)
jupyter (1.0.0)
jupyter-client (4.2.2)
jupyter-console (4.1.1)
jupyter-core (4.1.0)

Also
GPL Ghostscript 9.18 (2015-10-05)

I experimented on both python3.5 and python2.7, but this shouldn't be relevant.

The error message is as follows

```
IOError                                   Traceback (most recent call last)
/home/nykh/Documents/cs224u/program/local/lib/python2.7/site-packages/IPython/core/formatters.pyc in __call__(self, obj)
    341             method = _safe_get_formatter_method(obj, self.print_method)
    342             if method is not None:
--> 343                 return method()
    344             return None
    345         else:

/home/nykh/Documents/cs224u/program/local/lib/python2.7/site-packages/nltk/tree.pyc in _repr_png_(self)
    727                             '-q -dEPSCrop -sDEVICE=png16m -r90 -dTextAlphaBits=4 -dGraphicsAlphaBits=4 -dSAFER -dBATCH -dNOPAUSE -sOutputFile={0:} {1:}'
    728                             .format(out_path, in_path).split())
--> 729             with open(out_path, 'rb') as sr:
    730                 res = sr.read()
    731             os.remove(in_path)

IOError: [Errno 2] No such file or directory: u'/tmp/tmppncGPb.png'
```

Apparently, the temp png file is not emitted by GhostScript, which complains

```
Error: /undefinedresult in --stringwidth--
Operand stack:
   --nostringval--   0   12   67   0   0   (test)
Execution stack:
   %interp_exit   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   false   1   %stopped_push   1977   1   3   %oparray_pop   1976   1   3   %oparray_pop   --nostringval--   1960   1   3   %oparray_pop   1852   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   --nostringval--   %array_continue   --nostringval--   --nostringval--   --nostringval--   %array_continue   --nostringval--
Dictionary stack:
   --dict:1201/1684(ro)(G)--   --dict:0/20(G)--   --dict:83/200(L)--   --dict:23/50(L)--
Current allocation mode is local
Current file position is 7312
GPL Ghostscript 9.18: Unrecoverable error, exit code 1
```

I wonder whether it's the way that `_repr_png` calls GS that causes the GhostScript error

``` python
subprocess.call([find_binary('gs', binary_names=['gswin32c.exe', 'gswin64c.exe'], env_vars=['PATH'], verbose=False)] +
                            '-q -dEPSCrop -sDEVICE=png16m -r90 -dTextAlphaBits=4 -dGraphicsAlphaBits=4 -dSAFER -dBATCH -dNOPAUSE -sOutputFile={0:} {1:}'
                            .format(out_path, in_path).split())
```

Yorwba:
There is a bug in Tk on Linux which causes font sizes to be reported as 0 occasionally. That doesn't seem to impact display, but when saving to postscript, the output contains `0 scalefont` for each text item. When nltk calls Ghostscript to convert the postscript file to an image in `Tree._repr_png_`, it rightfully complains about divide-by-zero errors caused by this.

A workaround is to replace each instance of `0 scalefont` by `9 scalefont`, using the default size Tk would generate if it weren't for the bug. (I don't think nltk sets the font size anywhere?)

### DIFF
--- a/nltk/draw/util.py
+++ b/nltk/draw/util.py
@@ -1712,11 +1712,15 @@ class CanvasFrame(object):
                                          defaultextension='.ps')
             if not filename: return
         (x0, y0, w, h) = self.scrollregion()
-        self._canvas.postscript(file=filename, x=x0, y=y0,
+        postscript = self._canvas.postscript(x=x0, y=y0,
                                 width=w+2, height=h+2,
                                 pagewidth=w+2, # points = 1/72 inch
                                 pageheight=h+2, # points = 1/72 inch
                                 pagex=0, pagey=0)
+        # workaround for bug in Tk font handling
+        postscript = postscript.replace(' 0 scalefont ', ' 9 scalefont ')
+        with open(filename, 'w') as f:
+            f.write(postscript.encode('utf8'))
 
     def scrollregion(self):
         """


stevenbird:
Thanks @Yorwba.

(I don't see why this could have caused a regression in nltk.tag, so I'm merging anyway.)

PR #1638, opened 2017-02-26
Fix Gensim deprecated method call in tests

fievelk:
As noticed by @naoyak, a new release of Gensim is causing tests to break due to a deprecated method being used.
In particular, `gensim.models.Word2Vec.load_word2vec_format` should now be abandoned in favor of `gensim.models.KeyedVectors.load_word2vec_format`.

This PR fixes the issue calling the new suggested method.


stevenbird:
Thanks @fievelk

PR #1639, opened 2017-02-27
Use abstract classes

naoyak:
Ref: #1130

Cherry-picking some commits from #1562.

### DIFF
--- a/nltk/ccg/api.py
+++ b/nltk/ccg/api.py
@@ -7,42 +7,57 @@
 from __future__ import unicode_literals
 from functools import total_ordering
 
+from abc import ABCMeta, abstractmethod
+from six import add_metaclass
+
 from nltk.internals import raise_unorderable_types
 from nltk.compat import (python_2_unicode_compatible, unicode_repr)
 
+
+@add_metaclass(ABCMeta)
 @total_ordering
 class AbstractCCGCategory(object):
     '''
     Interface for categories in combinatory grammars.
     '''
 
-    # Returns true if the category is primitive
+    @abstractmethod
     def is_primitive(self):
-        raise NotImplementedError()
+        """
+        Returns true if the category is primitive.
+        """
 
-    # Returns true if the category is a function application
+    @abstractmethod
     def is_function(self):
-        raise NotImplementedError()
+        """
+        Returns true if the category is a function application.
+        """
 
-    # Returns true if the category is a variable
+    @abstractmethod
     def is_var(self):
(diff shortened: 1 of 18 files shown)


nltk/ccg/api.py
```diff
@@ -5,43 +5,58 @@
...
+         - Returns None if they cannot be unified
+         - Returns a list of necessary substitutions if they can.
+        """
 
     # Utility functions: comparison, strings and hashing.
-
+    @abstractmethod
     def __str__(self):
-        raise NotImplementedError()
+        pass
```
alvations, line 60: The `pass` here is inconsistent with the rest of the `@abstractmethod`. Is it necessary?

alvations: Ah my mistake there's no docstring for this, that's why =)

nltk/metrics/association.py
```diff
@@ -61,16 +64,14 @@ class NgramAssocMeasures(object):
     _n = 0
 
     @staticmethod
+    @abstractmethod
     def _contingency(*marginals):
         """Calculates values of a contingency table from marginal values."""
-        raise NotImplementedError("The contingency table is not available"
-                                    "in the general ngram case")
```
alvations: For these cases I think the error message is important. Is there a way to use `@abstractmethod` and still raise the error message?

nltk/metrics/association.py
```diff
@@ -61,16 +64,14 @@ class NgramAssocMeasures(object):
...
 
     @staticmethod
+    @abstractmethod
     def _marginals(*contingency):
         """Calculates values of contingency table marginals from its values."""
-        raise NotImplementedError("The contingency table is not available"
-                                    "in the general ngram case")
```
alvations: Same as above. 

For these cases I think the error message is important. Is there a way to use `@abstractmethod` and still raise the error message?

nltk/metrics/association.py
naoyak: good idea, added it back!

nltk/metrics/association.py
naoyak: good idea, added it back!

naoyak:
@alvations thanks for the feedback - reflected

naoyak:
Rebased and updated.

naoyak:
@alvations @stevenbird - does this look better?

stevenbird:
@naoyak – yes, looking good!

PR #1643, opened 2017-03-01
Rte minor bug fix

aetilley:
@ClaudeCoulombe, @stevenbird

Have you found that the change made in 96f3e8f accomplishes what you want? For me it still gives incorrect tokens (this time just the last letter of each word).

I believe the cause of the empty tokens before the change, and the cause of the new incorrect tokens after the change is in the first internal set of parentheses in either

`tokenizer = RegexpTokenizer('([A-Z]\.)+|\w+|\$[\d\.]+')`

or in

`tokenizer = RegexpTokenizer('([\w.@:/])+|\w+|\$[\d.]+')`

The RegexpTokenizer documentation says:

:param pattern: The pattern used to build this tokenizer.
(This pattern must not contain capturing parentheses;
Use non-capturing parentheses, e.g. (?:...), instead)

I have been able to fix this problem by either changing

`tokenizer = RegexpTokenizer('([\w.@:/])+|\w+|\$[\d.]+')`

to

`tokenizer = RegexpTokenizer('(?:[\w.@:/])+|\w+|\$[\d.]+')`

or by simply getting rid of the first set of parentheses entirely, since they are redundant in the new regex:

`tokenizer = RegexpTokenizer('[\w.@:/]+|\w+|\$[\d.]+')`

This last change is what is new in this pull request.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -208,6 +208,7 @@
 - Tsolak Ghukasyan
 - Prasasto Adi
 - Safwan Kamarrudin
+- Arthur Tilley
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer
--- a/nltk/classify/rte_classify.py
+++ b/nltk/classify/rte_classify.py
@@ -61,7 +61,7 @@ class RTEFeatureExtractor(object):
         # Try to tokenize so that abbreviations, monetary amounts, email
         # addresses, URLs are single tokens.
         from nltk.tokenize import RegexpTokenizer
-        tokenizer = RegexpTokenizer('([\w.@:/])+|\w+|\$[\d.]+')
+        tokenizer = RegexpTokenizer('[\w.@:/]+|\w+|\$[\d.]+')
 
         #Get the set of word types for text and hypothesis
         self.text_tokens = tokenizer.tokenize(rtepair.text)


aetilley:
(I'm not sure if I should have edited the AUTHORS.md file.  I was following your instructions in CONTRIBUTING.md, but regardless it's a small change, so feel free to remove my name if you see fit.)

ClaudeCoulombe:
I agree. I was using tokenizer = RegexpTokenizer('[A-Za-z.]+|\w+|\$[\d\.]+') which is similar with no parenthesis. 

Thank you!

aetilley:
@ClaudeCoulombe, @stevenbird 

Perhaps a deeper question is why we are using the RegexpTokenizer here instead of the much more robust TreebankWordTokenizer.

The documentation in RTEFeatureExtractor says:

"Try to tokenize so that abbreviations, monetary amounts, email                                             
addresses, URLs are single tokens."

For instance:

In []: tt = nltk.TreebankWordTokenizer()                                                                                                                      
In []: rt = nltk.RegexpTokenizer('[\w.@:/]+|\w+|\$[\d.]+')                                                                               
In []: sent = "I paid $3.50 for an old U.S. coin at name@domain.com yesterday."
In []: print tt.tokenize(sent)                                                                                     
['I', 'paid', '$', '3.50', 'for', 'an', 'old', 'U.S.', 'coin', 'at', 'name', '@', 'domain.com', 'yesterday', '.']                                                                         
In []: print rt.tokenize(sent)                                                                                     
['I', 'paid', '$3.50', 'for', 'an', 'old', 'U.S.', 'coin', 'at', 'name@domain.com', 'yesterday.'] 

(Notice that both agree on the abbreviation "U.S.", but notice the difference in the monetary amount, the email address, and the final period.)

I can understand that one might want to treat URLs/email addresses as a single token (although I'm not sure I fully understand why $3.50 should be one token).  Regardless, given that this RegexpTokenizer will have poor performance in other areas, perhaps the best strategy would be to teach the TreebankWordTokenizer to recognize URLs/email addresses and use this in RTEFeatureExtractor?

I can move this discussion elsewhere if you think this is not the place for it.  Thanks.

Arthur Tilley

stevenbird:
Thanks @aetilley, @ClaudeCoulombe.

@aetilley: can I suggest that you copy your most recent comment into a new issue please?

aetilley:
@stevenbird 

Done:  https://github.com/nltk/nltk/issues/1650

PR #1645, opened 2017-03-05
Fix a bug on Python 3.3 and earlier versions

abougouffa:
The `reload` function in python version 3.0 to 3.3 is located in the `imp` module and after 3.4 the `imp` module gets depracated and the `reload` function moved to `importlib` module

You can see:

https://docs.python.org/3/whatsnew/3.0.html?highlight=reload#builtins

https://docs.python.org/3/whatsnew/3.4.html?highlight=reload#importlib


stevenbird:
Thanks @abougouffa!

PR #1651, opened 2017-03-12
fixed nltk.grammar.cfg_demo() Nonterminal division error

aetilley:
`from __future__ import division`

 makes slash ('/') division of non-terminals (and hence cfg_demo()) break.

### DIFF
--- a/nltk/grammar.py
+++ b/nltk/grammar.py
@@ -1333,9 +1333,9 @@ def cfg_demo():
     # Create some nonterminals
     S, NP, VP, PP = nonterminals('S, NP, VP, PP')
     N, V, P, Det = nonterminals('N, V, P, Det')
-    VP_slash_NP = VP/NP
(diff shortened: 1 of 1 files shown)


stevenbird:
@aetilley: thanks for the fix. Unfortunately, we need to find another way to solve this, to preserve the slash notation of categorial grammar. I wonder if we can drop the `from future import __division__` line instead. Would you mind updating your PR, and also including a regression test, and we can see if this works?

aetilley:
@stevenbird 

I think removing the future import of division would be risky since it might break other numerical operations for python 2 users.

I have found that including a `__truediv__` operation which uses `__div__` will allow the slash operator to work correctly.  I will submit a pull request.

aetilley:
@stevenbird:  Wrt above, see https://github.com/nltk/nltk/pull/1660

stevenbird:
Thanks @aetilley – much appreciated!

Issue #1648, opened 2017-03-10
Porting several stemming algorithms to NLTK


alvations:
`whoosh` has a couple of great things that we can port to NLTK, other than the porter stemmer, there are:

 - [Paice-Husk stemming ](https://bitbucket.org/mchaput/whoosh/src/e344fb64067e45d47ec62dc65a75a50be51264a7/src/whoosh/lang/paicehusk.py?at=default&fileviewer=file-view-default) 
 - [Lovin stemmer](https://bitbucket.org/mchaput/whoosh/src/e344fb64067e45d47ec62dc65a75a50be51264a7/src/whoosh/lang/lovins.py?at=default&fileviewer=file-view-default)

This will be an easy port since the original implementation is in Python. Simply port the code, write some tests and document the functions appropriately, put the module in `nltk/nltk/stem/` and do a pull-request =)

Any takers?

stevenbird:
+1

bora-pajo:
+1

jaehoonhwang:
Hi, can I try this one?

alvations:
@jaehoonhwang Sure! Try porting the stemmers one at a time and then do a PR and if you need help, we can guide you through it =)

Start with https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md

ayushmandey97:
Hey, are there any more enhancements required for this? Would be happy to help.

PR #1654, opened 2017-03-17
Improving Lancaster stemmer with strip prefix function and customizable rules from Whoosh

jaehoonhwang:
Issue: #1648 

- Clean up documentation
- Added doctests (copied from `lancaster.py`)
- Added AUTHORS.md
- Fix the code where if there is no rule table present in __init__, the stemmer uses default rule.

Code is from [Whoosh](https://bitbucket.org/mchaput/whoosh/src/e344fb64067e45d47ec62dc65a75a50be51264a7/src/whoosh/lang/paicehusk.py?at=default&fileviewer=file-view-default)

If I could get more detail on `nltk` testing, that would be great.
Because, I'm a bit confused with doctest and how to run the test.


alvations:
@jaehoonhwang the tests are failing because the stemming function is returning empty strings instead of the expected stems. https://nltk.ci.cloudbees.com/job/pull_request_tests/118/TOXENV=py35-jenkins,jdk=jdk8latestOnlineInstall/testReport/junit/nltk.stem/paicehusk/PaiceHuskStemmer/

Could you take a look at what's wrong? Or would you need some guidance on debugging it?

jaehoonhwang:
Oh no. I already verified it.
When there is no stem, such as word `multiply` the function returns an empty string.
I was wondering:
1. How can I run doctest locally? Should I use following line?

```py
if __name__ == "__main__":
    import doctest
    doctest.mod()
```

2. Is doctest enough for this module or do I need to write `unittest`?

alvations:
When no pattern matches, possibly the normal way to handling that is to return the original string instead of an empty string. E.g.

```
>>> from nltk.stem import PorterStemmer
>>> porter = PorterStemmer()
>>> porter.stem('dog')
'dog'
```

Could you make changes to the code such that it returns the original string when there are no stem pattern match?

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,274 @@
...
+        'say'
+        >>> st.stem('crying')      #     letters and one of those letters must be a vowel
+        'cry'
+        >>> st.stem('string')      # ditto
+        'string'
+        >>> st.stem('meant')       # ditto
+        'meant'
+        >>> st.stem('cement')      # ditto
+        'cem'
+    """
```
alvations: The comments right for the lancaster stemmer but is Paice-Husk stemmer exhibiting the same behavior?

jaehoonhwang:
Yes.
I was working on it, just in case you wanted that way.

nltk/stem/paicehusk.py
jaehoonhwang: Yes. When I tested them locally, the doctest didnt give me any error except for empty string ones. But I copied lancaster doctest for testing purpose. I will checl them in the morning

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        :param ruletable: a string containing the rule data, separated
+            by newlines.
+        """
+        self.rules = defaultdict(list)
+        if ruletable:
+            self.read_rules(ruletable)
+        else:
+            self.read_rules(self.defaultrules)
+
```
alvations: It'll be good to add the ruletable variable to the stemmer object that way, it can be modified later on (if needed), e.g. 

```python
self._ruletable = ruletable if ruletable else self.defaultrules
self.read_rules(self._ruletable)
```

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,274 @@
...
+            match = rule_expr.match(line)
+            if match:
+                ending = match.group("ending")[::-1]
+                lastchar = ending[-1]
+                intact = match.group("intact") == "*"
+                num = int(match.group("num"))
+                append = match.group("append")
+                cont = match.group("cont") == ">"
+
+                rules[lastchar].append((ending, intact, num, append, cont))
```
alvations: I think this `append` here don't change the `self.rules` and it isn't used anywhere else in this function, right? Would it break anything if we remove this?

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,274 @@
...
+            self.read_rules(ruletable)
+        else:
+            self.read_rules(self.defaultrules)
+
+    def read_rules(self, ruletable):
+        """Verify integrity of ruletable.
+        """
+        rule_expr = self.rule_expr
+        rules = self.rules
+
```
alvations: I think if we are not modifying `self.rule_expr` and `self.rules`, we might not need to have a local copy of this =)

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,274 @@
...
+    def strip_prefix(self, word):
+        """Remove prefix from a word.
+        """
+        for prefix in ("kilo", "micro", "milli", "intra", "ultra", "mega",
+                       "nano", "pico", "pseudo"):
+            if word.startswith(prefix):
+                return word[len(prefix):]
+        return word
+
```
alvations: Interesting, any idea why `strip_prefix` is necessary? @ExplodingCabbage any tips?

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,274 @@
...
+            pfv = self.first_vowel(stem)
+            rulelist = rules.get(stem[-1])
+            if not rulelist:
+                break
+
+            continuing = False
+            for ending, intact, num, append, cont in rulelist:
+                if stem.endswith(ending):
+                    if intact and not is_intact:
+                        continue
```
alvations: This if-condition `if intac and not is_intact` seems a little funny. Maybe some comments on this would help users and us better maintain the code

nltk/stem/paicehusk.py
alvations: Cool! When we have more of them maybe we can add a wiki on different effects of different stemmers =)

alvations:
Thanks @jaehoonhwang! I've finished the review, I hope the line-by-line review helps the revision =)

ExplodingCabbage:
Hmm. Licensing aside (though it would be good if someone could confirm there are no issues there, too), how do Whoosh feel about NLTK copying their code? To avoid code duplication and allow any future fixes and features to benefit both userbases, would it good for them to make NLTK a dependency and import the stemmer from it, or perhaps for NLTK to extract out this particular stemmer into a standalone package that both Whoosh and NLTK can depend upon? I believe @mchaput is the maintainer of Whoosh; I suggest that it would be both useful and courteous to give him a chance to weigh in here and at https://github.com/nltk/nltk/issues/1648 before merging copies of any of his stuff.

Also, if we're going to copy people's work, I think they should be credited in AUTHORS.md. You may want to emulate the approach I took with the Porter Stemmer.

I'll take a further look at this PR when I have the chance, probably at the weekend.

alvations:
I agree with @ExplodingCabbage on asking the Whoosh maintainers before we merge the code.  Started a thread on https://groups.google.com/forum/#!topic/whoosh/8o_iYCn1Sao

jaehoonhwang:
@alvations Is there any more changes you want me to do?

ExplodingCabbage:
- [x] The stemmer should be imported in `nltk/stem/__init__.py` like all the other stemmers are.

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,315 @@
...
+# policies, either expressed or implied, of Matt Chaput.
+
+"""
+Paice-Husk Stemmer
+
+If you just want to use the standard Paice-Husk stemming rules, use the
+module's stem()` function::
+
+    stemmed_word = stem(word)
+
```
alvations: This would be a more explicit example:

```python
paicehusk = PaiceHuskStemmer()
stemmed_word = paicehusk.stem(word)
```

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,315 @@
...
+If you want to use a custom rule set, read the rules into a string where the
+rules are separated by newlines, and instantiate the object with the string,
+then use the object's stem method to stem words::
+
+    stemmer = PaiceHuskStemmer(my_rules_string)
+    stemmed_word = stemmer.stem(word)
+
```
alvations: It'll be good to give an example of how to instantiate `my_rules_string`, esp. the syntax of how `my_rules_string` should be written.

jaehoonhwang: Ok, I'm on it.

nltk/stem/paicehusk.py
```diff
@@ -0,0 +1,315 @@
...
+The base implementation was done by
+
```
alvations: Missing words?

alvations:
For now there's no preference for doctest or unittest but writing doctest for the stemmers will be a good start.

BTW, after reading the literature and the code, the Paice-Husk stemmer has the same reference as the Lancaster stemmer and the rules and algorithm looks really similar. **Does anyone know whether they're the same thing?** 

If they are the same, then merging some code to allow custom rules from here to the `nltk.stem.lancaster.py` would be more appropriate than keeping two copies of the same algorithm.

Sorry I didn't realize that they might be the same earlier.

ExplodingCabbage:
> BTW, after reading the literature and the code, the Paice-Husk stemmer has the same reference as the Lancaster stemmer and the rules and algorithm looks really similar. Does anyone know whether they're the same thing?

I've also been trying to figure that out, and I think the answer is "yes". In particular, NLTK's `lancaster.py` cites

> Paice, Chris D. "Another Stemmer." ACM SIGIR Forum 24.3 (1990): 56-61.

which is *also* the paper that many sources on the internet cite as the original definition of Paice/Husk. Note also that Chris Paice worked at Lancaster University before he died (per http://wp.lancs.ac.uk/chrispaice/).

I have no idea why the word "Husk" appears in the name - that bit of history yet eludes me.

Also, unfortunately the paper only seems to exist on the internet behind a paywall, SciHub seems unable to pirate it for me, and I can't ask Chris directly for it because he's dead. Perhaps there are some academics here who have access via their institutions?

I can, however, test purported implementations of Paice/Husk on the internet against this one. For instance, there's one at http://paicehusk.appspot.com/. It doesn't have the prefix-removal stuff that the one in this PR does (e.g. this PR stems `'kilogram'` to `'gram'`; that implementation stems it to `'kilogram'`). `lancaster.py` also doesn't do that prefix removal. That strongly hints that `lancaster.py` is the more faithful implementation.

ExplodingCabbage:
Aha, "Husk" seems to be the surname of another researcher that Chris Paice worked worth - this article is apparently by *C.D. Paice* and *G.D. Husk*: http://www.sciencedirect.com/science/article/pii/0885230887900039

I'm guessing that when someone gets their hands on a PDF of "Another Stemmer" we'll learn that Husk was a co-author, and that through some quirk of history his name never gets included when academics cite the paper and yet made it into the popular name of the stemmer?

alvations:
Ah, Paice wrote and published the paper and Husk wrote the software, and they're both from lancaster: https://pdfs.semanticscholar.org/3b8b/099f62391ff24e94b10ebafbb0488be83632.pdf and https://web.archive.org/web/20060501021257/http://www.comp.lancs.ac.uk/computing/research/stemming/Links/implementations.htm

It looks like the prefix stripping function is whoosh's addition, the [original pascal code](https://web.archive.org/web/20060501021257/http://www.comp.lancs.ac.uk/computing/research/stemming/Files/Pascal.zip) didn't have that. 

@jaehoonhwang I think it'll be better if we merge the custom rules code from here to lancaster and allow an option to allow users to  remove prefix or keep the original implementation.

ExplodingCabbage:
Yeah, he's from Lancaster, per his Facebook: https://www.facebook.com/ghusk

jaehoonhwang:
I can remove prefix function, if you prefer.

alvations:
@jaehoonhwang we should take what whoosh has (in `nltk/stem/paicehusk.py`) and then add then to the `nltk/stem/lancaster.py` code. That's so that we don't maintain 2 variants of the same algorithm at separate places. E.g. for Porter stemmer, if i'm not mistaken, @ExplodingCabbage allowed users to choose the variants, see https://github.com/nltk/nltk/blob/develop/nltk/stem/porter.py#L49

There're some good things that whoosh has that the current lancaster stemmer in NLTK does, more specifically, the:

 - prefix stripping function
 - custom rule

jaehoonhwang:
Those functionalities are now added to Lancaster and I removed `paicehusk.py`.
~~Can I get a rule for testing custom rule tuples?~~
~~I've tried different things, but can't come up with anything that works.~~
I got it.

nltk/stem/lancaster.py
```diff
@@ -304,7 +319,14 @@ def __applyRule(self, word, remove_total, append_string):
             word += append_string
         return word
 
+    def strip_prefix(self, word):
+        """Remove prefix from a word.
+        """
```
alvations: It'll be good to acknowledge the original src of this function from whoosh here by adding a docstring that includes the link: https://bitbucket.org/mchaput/whoosh/src/e344fb64067e45d47ec62dc65a75a50be51264a7/src/whoosh/lang/paicehusk.py#paicehusk.py-73

ExplodingCabbage: Better yet, let's also try to track down the history of how it got added to Whoosh via their source control. For all we know it may have been a copy-and-paste to begin with from some other source that we're failing to credit! Knowing the history will also let us credit individual authors in CONTRIBUTORS.md.

None of this is vital, but would be nice (and also potentially *practically* useful in helping understand the alterations that were made to the original algorithm and what motivated them).

alvations: This is interesting, it seems like many versions of Porter stemmer (except NLTK's) contain that function ;P

https://www.google.com.sg/search?client=safari&rls=en&q=for+prefix+in+(%22kilo%22,+%22micro%22,+%22milli%22,+%22intra%22,+%22ultra%22,+%22mega%22

nltk/stem/lancaster.py
```diff
@@ -192,14 +201,20 @@ def stem(self, word):
         """Stem a word using the Lancaster stemmer.
         """
         # Lower-case the word, since all the rules are lower-cased
-        word = word.lower()
+        word = self.strip_prefix(word).lower() if self._strip_prefix else word.lower()
```
ExplodingCabbage: - [x] Surely we should be coercing the word to lowercase *before* the `strip_prefix()` call, not after? Won't this implementation map `KILOGRAM` to `kilogram` but `kilogram` to `gram`?

nltk/stem/lancaster.py
jaehoonhwang: For now I'll add Whoosh as a source.

nltk/stem/lancaster.py
jaehoonhwang: Confirmed and fixed.

jaehoonhwang:
Should I add Whoosh on `contributing.md`?
Also, do you guys think having `_strip_prefix` and `__strip_prefix` is too confusing?
One is class member and one is class function.
Also, i have an argument in Lancaster init as `strip_prefix`

- `strip_prefix`: Argument variable
- `_strip_prefix`: Class variable (bool)
- `__strip_prefix`: Class member function.

stevenbird:
@jaehoonhwang: I agree that the variable names are too similar, given the different purposes.
I agree about acknowledging indirect contributions.

jaehoonhwang:
Any other thing I should do? @alvations

alvations:
@jaehoonhwang sorry for the late reply. It seems like the Whoosh maintainers didn't reply on https://groups.google.com/forum/#!topic/whoosh/8o_iYCn1Sao 

Let's wait till the weekend and if there's still no response, it looks good to merge after a last round of quick review @ExplodingCabbage and me. 

@ExplodingCabbage @mchaput anything to add?

jaehoonhwang:
Thanks for the reply!

alvations:
Good news, we got the go ahead to merge from @mchaput =) 

@jaehoonhwang just one small request, do a `git pull` and `git merge` to include the docstring addition that @ExplodingCabbage wrote. 

@ExplodingCabbage any thing to add to this PR?

jaehoonhwang:
Ok. I'll do it in few hours and do all those. Thanks for getting back.

jaehoonhwang:
@alvations I think this is it. Thank you and @ExplodingCabbage for the help!

ExplodingCabbage:
A few things:

1. There's some commented-out code that this version currently adds that shouldn't be there
2. We've added an argument called `rule_tuples` (plural), but doesn't it actually take (at least in our test case) a *single* tuple? And shouldn't it therefore be named `rule_tuple` (singular)? Am I missing something?
3. The (pre-existing) comment here is interesting:

    ```py
        # If the user hasn't supplied any rules, setup the default rules
        if len(self.rule_dictionary) == 0:
            self.parseRules(LancasterStemmer.rule_tuple)
    ```

    It looks like the possibility of users passing in custom rules was at least something that the authors of the NLTK version of the stemmer had conceived of, despite not taking them as an argument to the constructor. I guess this is why `parseRules()` is public and takes a `rule_tuple` argument; using the *old* stemmer, I can do this (similar to the doctest added in this PR):

    ```
    >>> st_custom = LancasterStemmer()
    >>> st_custom.parseRules(rule_tuple=("ssen4>", "s1t."))
    >>> st_custom.stem("ness") # Change s to t
    'nest'
    ```

    As such, one of the two pieces of functionality this PR adds was actually already supported, just without us noticing!

    I guess I don't object to adding the `rule_tuple` argument to the constructor, though, since:
    - it doesn't break the old custom rules API
    - the old custom rules API was completely undocumented and non-obvious; the new way is significantly more discoverable

    However, we *should* delete that pre-existing comment since this PR renders it completely wrong (since the block being commented on can now run even if the user has supplied rules, and the rules it sets up are not necessarily default ones).

jaehoonhwang:
That makes sense.
What should I do with this PR?
This seems like redundant PR that supports already existing function in Lancasterstemmer.

jaehoonhwang:
@ExplodingCabbage 
I've fixed/added documentation on parseRules  and __stripPrefix, removed commented out codes, fixed comment in stem.
This fix still supports old API for parseRule.

alvations:
It looks good to merge now =)

@ExplodingCabbage anything to add? BTW, just like to check, do you have merging rights?

alvations (approved):

ExplodingCabbage:
> @ExplodingCabbage anything to add?

Nope. :)

> do you have merging rights?

Nope.

jaehoonhwang:
Thank you guys for all the help!

stevenbird:
@jaehoonhwang, @ExplodingCabbage, @alvations: thanks for wrapping this up. I'm thinking the name of this PR is out of date – perhaps one of you could fix that please, before I merge?

stevenbird:
Thanks @alvations

PR #1655, opened 2017-03-17
fix two typo-level errors in docstrings of PerceptronTagger

cfbolz:
In both cases the docstring and the code disagreed. I made the docstrings match the code.


stevenbird:
Thanks @cfolz

Issue #1634, opened 2017-02-19
corpus_bleu does not match multi-bleu.perl for very poor translations


AndreasMadsen:
example:

**tokenized reference:**
```
Their tasks include changing a pump on the faulty stokehold .
Likewise , two species that are very similar in morphology were distinguished using genetics .
```

**tokenized hypothesis:**
```
Teo S yb , oe uNb , R , T t , , t
Tue Ar saln S , , 5istsi l , 5oe R ulO sae oR R
```

with the `multi-bleu.perl` script the BLEU score is `0`:

```
BLEU = 0.00, 3.4/0.0/0.0/0.0 (BP=1.000, ratio=1.115, hyp_len=29, ref_len=26)
```

with `corpus_bleu` the BLUE score is `43.092382`, my understanding is that default `corpus_bleu` settings corresponds to the `multi-bleu.perl` script.

---

```py
from nltk.translate.bleu_score import corpus_bleu

references = [
    'Their tasks include changing a pump on the faulty stokehold .',
    'Likewise , two species that are very similar in morphology were distinguished using genetics .'

]

hypothesis = [
    'Teo S yb , oe uNb , R , T t , , t',
    'Tue Ar saln S , , 5istsi l , 5oe R ulO sae oR R'
]

hypothesis_tokens = [line.split(' ') for line in hypothesis]
references_tokens = [[line.split(' ')] for line in references]

# calculate corpus-bleu score
bleu = corpus_bleu(
    references_tokens, hypothesis_tokens
)

print('BLEU: %f' % (bleu * 100)) # 43.092382
```

alvations:
If we look at the numbers from `multi-bleu.perl`, we see unigram matches of `3.4`. 

```
BLEU = 0.00, 3.4/0.0/0.0/0.0 (BP=1.000, ratio=1.115, hyp_len=29, ref_len=26)
```

And the bad thing about `multi-bleu.perl` is that it throws away this 3.4 and assumes 0 score once it doesn't have a higher order ngram match. This is not an ideal situation.

So to resolve the isse in #1554,  the `develop` branch of the `nltk` after #1330 and #1554 produces:

```python
>>> from nltk.translate.bleu_score import corpus_bleu
>>> hyp = "Teo S yb , oe uNb , R , T t , , t Tue Ar saln S , , 5istsi l , 5oe R ulO sae oR R"
[...]

PR #1657, opened 2017-03-19
Added parameter to emulate multi-bleu.pl 

alvations:
Added a function to emulate multi-bleu for fringe cases of BLEU usage, c.f. #1634


stevenbird:
Thanks @alvations

stevenbird:
@alvations: the CI server reports this error: `TypeError: method4() got an unexpected keyword argument 'emulate_multibleu'`

alvations:
@stevenbird missed that out in the when I only checked the unittests. Sorry about it. Patched that in #1662. Maybe I should add the doctests for BLEU scores in the unittests too. I'll leave that for another PR.

PR #1661, opened 2017-03-20
Include "Paice/Husk" in lancaster.py's docstring for Googleability

ExplodingCabbage:
Should help users who are looking for an NLTK implementation of Paice/Husk but don't realise that Lancaster == Paice/Husk.

https://github.com/nltk/nltk/pull/1654 was the (somewhat comical) inspiration for this.


stevenbird:
Thanks @ExplodingCabbage

PR #1670, opened 2017-03-26
Ignore pyenv python version files

fievelk:
The python version management [`pyenv`][1] uses a file called `.python-version` to determine which Python version to use in the application folder. This PR ignores `.python-version` files.

[1]: https://github.com/pyenv/pyenv


fievelk:
[CI: retest]

PR #1680, opened 2017-04-10
Speed up TnT tagger; improve FreqDist and ConditionalFreqDist

vthorsteinsson:
The current implementations of FreqDist and ConditionalFreqDist in probability.py have member functions called N() that return a count of the samples in the distribution. These functions are quite inefficient as they perform a sum each time over the contained sample buckets. In typical cases, for instance during POS tagging, these sums can loop over tens or hundreds of thousands of buckets for every call to N(). To add insult to injury, the FreqDist.freq() function unnecessarily calls N() twice.

This PR adds a function freeze_N() to FreqDist and ConditionalFreqDist. The function evaluates N() and then caches or memoizes it by replacing the self.N() function with a lambda that simply returns the memoized value, making subsequent calls to N() very fast. After freeze_N() has been called on a frequency distribution object, no further samples should be added to that object.

The PR modifies the nltk.tag.tnt.TnT class to call freeze_N() after training and before POS tagging commences, and serves as an example of how this functionality can be used.

With this PR applied, the **execution time of the nltk.tag.tnt.demo() function drops from 183 seconds to 107 seconds** on my Debian box, a speedup of ~70%. I have verified by profiling that the time saved was almost all previously spent in the N() function of FreqDist and ConditionalFreqDist. Test results are otherwise unaffected.

The PR also includes minor clean up of the main tagging loop of the TnT tagger.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -209,6 +209,7 @@
 - Prasasto Adi
 - Safwan Kamarrudin
 - Arthur Tilley
+- Vilhjalmur Thorsteinsson
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -116,6 +116,17 @@ class FreqDist(Counter):
         """
         return sum(self.values())
 
+    def freeze_N(self):
+        """
+        Set N permanently to its current value, making subsequent
+        calls to ``FreqDist.N()`` much faster. Use this for instance
+        after training a tagger, but before running it. After calling
+        ``FreqDist.freeze_N()``, no more samples should be added to the
+        FreqDist object.
+        """
+        n = self.N()
+        self.N = lambda: n
+
     def B(self):
         """
         Return the total number of sample values (or "bins") that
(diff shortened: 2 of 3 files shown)


stevenbird:
Thanks for this @vthorsteinsson – great idea. I would like to discuss an alternative implementation where the value is cached, but this cached value is flagged as stale by any operations that modify the number of samples in the frequency distribution. Then, when `N()` is called, the flag is checked, and only if it is stale is the sum performed, and then the flag is unset.

Then we can avoid (a) changing the API, and (b) requiring users to follow the discipline of not modifying a distribution once `freeze_N()` is called.

vthorsteinsson:
Here is an updated and improved PR per your suggestion @stevenbird. It adds a cached _N member of FreqDist which is invalidated (set to `None`) whenever the contents of the FreqDist may have been changed, i.e. in `__setitem__`, `__delitem__`, `update()` and `setdefault()` (these are the cases you need to cover for updating a dict or defaultdict). The _N value is cached whenever `N()` is called and reused on the next call unless invalidated. With this, `freeze_N()` is no longer needed and the external API of FreqDist is kept unchanged. Also, ConditionalFreqDist has been reverted to its original form as calculating `N()` there is much less expensive and gains implicitly from the caching in FreqDist. The calls to `freeze_N()` have likewise been removed from the TnT tagger as they are no longer necessary.

With this PR applied as it is now, the `nltk.tag.tnt.demo2()` function goes from **112.8** seconds down to **65.8** on my computer. All other clients of FreqDist should automatically benefit from this. The drawback is slightly more overhead on each FreqDist update, but in typical use cases much faster frequency queries far outweigh the slightly slower updates.

stevenbird:
Thanks @vthorsteinsson!

Issue #1667, opened 2017-03-24
nltk.translate.gleu_score problems


goodmami:
I see a few possible issues with the gleu_score implementation.

1. It doesn't allow multiple references, despite the [docstring saying it does](https://github.com/nltk/nltk/blob/develop/nltk/translate/gleu_score.py#L72). This is either a code or a doc issue. Can anyone clarify which?

2. I don't see why there is no `corpus_gleu()` function. The Wu et al. 2016 paper states (emphasis added):

   >According to our experiments, GLEU score correlates quite well with the BLEU metric **on a corpus level** but does not have its drawbacks for our per sentence reward objective

   So I would think there's a way to calculate this for the corpus. Maybe just `min(corpus_precision, corpus_recall)`, or alternatively the average of sentence-level GLEU scores?

3. The name GLEU is unfortunate; there are at least 2 other GLEUs that predate the 2016 paper:
   * Mutton et al. (2007) "GLEU: Automatic Evaluation of Sentence-Level Fluency" http://anthology.aclweb.org/P/P07/P07-1044.pdf
   * Napoles et al. (2015) "Ground Truth for Grammatical Error Correction Metrics" http://www.aclweb.org/anthology/P/P15/P15-2097.pdf

   Maybe the Wu et al. 2016 paper will choose a different name if they publish outside of arXiv? But for NLTK, would `google_bleu` or `gnmt_bleu` be less likely to step on toes?

I'd be happy to submit a PR when the above issues get resolved.

alvations:
@goodmami I'll try to answer some of the questions:

> It doesn't allow multiple references, despite the docstring saying it does. This is either a code or a doc issue. Can anyone clarify which?

It's a docstring problem. It doesn't take multiple references because the GMNT paper didn't account for multiple references. 

> I don't see why there is no corpus_gleu() function. The Wu et al. 2016 paper states (emphasis added):
> 
> According to our experiments, GLEU score correlates quite well with the BLEU metric on a corpus level but does not have its drawbacks for our per sentence reward objective
>
> So I would think there's a way to calculate this for the corpus. Maybe just min(corpus_precision, corpus_recall), or alternatively the average of sentence-level GLEU scores?

Yes, this was unclear because they were using GLEU to do some sort of reinforcement learning (I would call it nbest reranking) at the decoder, so it must have been something at sentence level. 

[...]

PR #1683, opened 2017-04-12
Addresses #1667; add gleu_score.corpus_gleu()

goodmami:
From an email thread on nltk-dev, we got some more information about
the corpus-level GLEU calculation. I used that to create a
corpus_gleu() function in nltk.translate.gleu_score. I did not change
the functions to accept multiple references per hypothesis because
that would change the API, and it doesn't appear that the authors of
GLEU implemented it (though they suggested it could just be the max
score among the references). If we want to allow it, we could do
something like:

 1. Inspect if the reference argument is `list(str)` or
    `list(list(str))` automatically, and silently adapt to the input.
 2. Allow an optional `multiple_reference` parameter which, when True,
    expects the reference argument to be multiple references.
 3. Wait for the next major version of NLTK and change the API.

### DIFF
--- a/nltk/translate/gleu_score.py
+++ b/nltk/translate/gleu_score.py
@@ -3,7 +3,7 @@
 #
 # Copyright (C) 2001-2017 NLTK Project
 # Authors:
-# Contributors:
+# Contributors: Mike Schuster, Michael Wayne Goodman, Liling Tan
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 
--- a/nltk/translate/gleu_score.py
+++ b/nltk/translate/gleu_score.py
@@ -46,9 +46,9 @@ def sentence_gleu(references, hypothesis, min_len=1, max_len=4):
          metric on a corpus level but does not have its drawbacks for our per
          sentence reward objective."
 
-    Note: The previous implementation only allowed a single reference; in order
-          to maintain backward compatibility, the first argument may be given
-          as a single reference or a list of references.
+    Note: The initial implementation only allowed a single reference, but now
+          a list of references is required (which is consistent with
+          bleu_score.sentence_bleu()).
 
     The infamous "the the the ... " example
 
@@ -81,11 +81,6 @@ def sentence_gleu(references, hypothesis, min_len=1, max_len=4):
     :return: the sentence level GLEU score.
     :rtype: float
     """
-
-    # compatibility with previous single-reference API:
-    if not references or isinstance(references[0], string_types):
-        references = [references]
-
     return corpus_gleu(
         [references],
         [hypothesis],
--- a/nltk/translate/gleu_score.py
+++ b/nltk/translate/gleu_score.py
@@ -12,7 +12,7 @@ from __future__ import division
 from collections import Counter
 
 from nltk.util import ngrams, everygrams
-from nltk.compat import string_types
+
 
 def sentence_gleu(references, hypothesis, min_len=1, max_len=4):
     """
@@ -158,7 +158,7 @@ def corpus_gleu(list_of_references, hypotheses, min_len=1, max_len=4):
         hyp_ngrams = Counter(everygrams(hypothesis, min_len, max_len))
         tpfp = sum(hyp_ngrams.values())  # True positives + False positives.
         
-        scores_by_ref = []
+        hyp_counts = []
         for reference in references:
             ref_ngrams = Counter(everygrams(reference, min_len, max_len))
             tpfn = sum(ref_ngrams.values())  # True positives + False negatives.
@@ -176,11 +176,11 @@ def corpus_gleu(list_of_references, hypotheses, min_len=1, max_len=4):
             n_all = max(tpfp, tpfn)
 
             if n_all > 0:
-                scores_by_ref.append((tp/n_all, tp, n_all))
+                hyp_counts.append((tp, n_all))
 
         # use the reference yielding the highest score
-        if scores_by_ref:
-            _, n_match, n_all = max(scores_by_ref)
+        if hyp_counts:
+            n_match, n_all = max(hyp_counts, key=lambda hc: hc[0]/hc[1])
             corpus_n_match += n_match
             corpus_n_all += n_all
 
--- a/nltk/translate/gleu_score.py
+++ b/nltk/translate/gleu_score.py
@@ -45,9 +45,8 @@ def sentence_gleu(reference, hypothesis, min_len=1, max_len=4):
          metric on a corpus level but does not have its drawbacks for our per
          sentence reward objective."
 
-    Note: The GLEU score is designed for sentence based evaluation thus there is
-          no corpus based scores implemented in NLTK. Also, unlike
-          multi-reference BLEU, GLEU only supports a single reference.
+    Note: Unlike multi-reference BLEU, this implementation of GLEU only
+          supports a single reference.
 
     The infamous "the the the ... " example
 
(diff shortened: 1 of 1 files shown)


alvations:
Thanks @goodmami for following up with the Google guys and the PR. Great discussion!

1. Actually with the corpus level score, we can make the sentence level a duck-type of the corpus level, something like: https://github.com/alvations/nltk/blob/develop/nltk/translate/bleu_score.py#L87

But since Google BLEU is sort of created in the reverse way, sentence-level first then corpus-level, I'm not sure whether doing it the duck-type way is okay.

2. Acutally empty reference must be allowed. Imagine that there are no reference translations given for the sentence just because the MT system returns empty line and passes and empty string as the reference to the metric. In the case, the traditional BLEU has put in quite a lot of hacks to handle the fringe cases: https://github.com/alvations/nltk/blob/develop/nltk/test/unit/translate/test_bleu.py

We can leave that for the next PR though =)

3. Function annotations are great to handle the multiple vs single reference situation, [PEP3107](https://www.python.org/dev/peps/pep-3107/), sadly I think it's Py3.x only. 
----

(Not exactly related)

About the multiple reference API. I think it's just my personal view but I hope it isn't just me. I think the multiple reference part of MT metrics is overlooked in MT. Although it's sort of "out-dated" view of MT, I think it's best to keep the multiple reference as the default because translations are naturally idiosyncratic. I think one main reason why the  H* metrics have quietly fade away because it's hard to produce annotations for them.

goodmami:
Thanks @alvations (and thanks for initiating that discussion),

* RE your (1): i'll see about updating the PR to do that
* RE your (2): that check was to avoid division-by-zero errors, which happens with `bleu_score.sentence_bleu()` if you use `auto_reweigh`:
    ```python
    >>> bleu_score.sentence_bleu(['bla', 'blah'], [], auto_reweigh=True)
      File "/home/goodmami/repos/xmt/env/lib/python3.4/site-packages/nltk/translate/bleu_score.py", line 81, in sentence_bleu
        weights, smoothing_function, auto_reweigh)
      File "/home/goodmami/repos/xmt/env/lib/python3.4/site-packages/nltk/translate/bleu_score.py", line 165, in corpus_bleu
        weights = ( 1 / hyp_lengths ,) * hyp_lengths
    ZeroDivisionError: division by zero
    ```
  So should I just return a score of 0 if hyp or ref is empty?
* RE your (3): using function annotations seems, perhaps, too clever, and probably Python 3 only as you say.

I agree we should make an effort to allow multiple references. Do you like one of the strategies I listed in the first post?

alvations:
@goodmami sorry for the delayed reply. The new push looks good!

On the type checks before for the `list(list(str))`, allowing multiple input types might be problematic. I'll review the code more thoroughly after work hours =)

On allowing `multiple_reference` parameter. I think this should be discouraged since the default `multiple_references` per hypothesis should be norm to be consistent with other metrics. But this could be change in another PR but it should be kept consistent for all metrics. 

Let's keep the fringe cases for another PR and write comprehensive unittest for multiple scores and use the same regression test for all new metrics. 

@stevenbird would it be alright if we sub-path the  `nltk.translate.*_score` modules to `nltk.translate.metrics`? Maybe this should be another PR too. 

----

About (3): Hmmm, let's wait for [2020](https://pythonclock.org)

stevenbird:
@alvations fine with me if you'd like to move `nltk.translate.*_score` down inside a sub-package `nltk.translate.metrics`.

goodmami:
@alvations, the type check only occurs in `sentence_gleu()`, since the `corpus_gleu()` function wasn't there before (so I don't need to maintain any backward compatibility), but let me know if the asymmetry is a problem. We could also type-check the argument to `corpus_gleu()`.

As for the type check, it is currently:

```python
if not references or isinstance(references[0], string_types):
    references = [references]
```

That is if you pass an empty token list for a reference (`[]`) or the fist element on a non-empty list is a string, then it is assumed you were trying to pass a single reference, and it is wrapped in an outer list. It occurs to me that the first case may not always be a correct assumption, e.g. if `[]` is not an empty single reference, but an empty list of references. I'm not sure there's any practical difference, though. Thoughts? Also, the boolean test (`if not references`) would also fire for other False values, like an empty string, `None`, or `False`, but I don't know how much type checking we should really do.

stevenbird:
@alvations: I think this PR is waiting for you to move `nltk.translate.*_score`, and then for @goodmami to update this PR. Please advise if I'm wrong on either count.

alvations:
@stevenbird I think we can leave the `nltk.translate.*_score` for another PR. I'll be a little busy this week.

goodmami:
I'm waiting on a decision regarding multiple references with `sentence_gleu()`. To rephrase, should I:

1. Break the existing (but still relatively new) API and only allow multiple references
2. Do a type-check to allow either single or multiple references
3. Only allow single references by default (for `sentence_gleu()`, not `corpus_gleu()`) to maintain the existing API?

Otherwise I'm not sure what I need to update. I could rebase my branch to the current upstream/develop, but I don't think it's necessary for a clean merge. The comprehensive unit tests should probably wait until the other `nltk.translate.metrics` PR, I think?

alvations:
@goodmami I'll vote for 1. Break the existing `gleu_score` api but at least it's consistent with the rest of the metrics. And to clean up and standardize the MT metric api, we can do it in another PR. 

Any objections?

goodmami:
No objections. See the latest commit.

nltk/translate/gleu_score.py
```diff
@@ -80,16 +81,113 @@ def sentence_gleu(reference, hypothesis, min_len=1, max_len=4):
...
+
+        # use the reference yielding the highest score
+        if scores_by_ref:
+            _, n_match, n_all = max(scores_by_ref)
+            corpus_n_match += n_match
+            corpus_n_all += n_all
+
+    # corner case: empty corpus or empty references---don't divide by zero!
+    if corpus_n_all == 0:
+        gleu_score = 0.0
```
alvations, line 189: In follow-up PRs from #1683, unittest for fringe cases would be good. (Noting this down, so that we keep track of things)

nltk/translate/gleu_score.py
```diff
@@ -80,16 +81,113 @@ def sentence_gleu(reference, hypothesis, min_len=1, max_len=4):
...
+            # recall, we can reduce the number of division operations by one by
+            # instead finding the maximum of the denominators for the precision
+            # and recall formulae, since the numerators are the same:
+            #     precision = tp / tpfp
+            #     recall = tp / tpfn
+            #     gleu_score = min(precision, recall) == tp / max(tpfp, tpfn)
+            n_all = max(tpfp, tpfn)
+
+            if n_all > 0:
+                scores_by_ref.append((tp/n_all, tp, n_all))
```
alvations: A little obtuse to read `tp/n_all` but I understand that's as per the discussion from the google group thread =)

alvations: I'm not good with variable names but maybe `this_gleu` or something along that line.

nltk/translate/gleu_score.py
```diff
@@ -80,16 +81,113 @@ def sentence_gleu(reference, hypothesis, min_len=1, max_len=4):
...
+
+        # use the reference yielding the highest score
+        if scores_by_ref:
```
alvations: Since we're already optimizing the F1, precision, recall computation. We can also optimize the max computation `scores_by_ref`.

alvations: `Fraction` would be a nice thing to use here.

nltk/translate/gleu_score.py
goodmami: I changed some variable names to match the discussion from the Google folk (part of which is repeated in the docstring). I guess `tp/n_all` is more PEP-8 compliant if it's `tp / n_all` (with spaces). Also, initially I thought I could save more calculations by just finding the minimum `tp`, but that wouldn't be correct because `n_all` isn't constant for all `tp` values.

nltk/translate/gleu_score.py
alvations: Something like:

```python
for references, hypothesis in zip(list_of_references, hypotheses):
        hyp_ngrams = Counter(everygrams(hypothesis, min_len, max_len))
        tpfp = sum(hyp_ngrams.values())  # True positives + False positives.

        highest_gleu_score = Fraction(0, 10000, _normalize=False)
        for reference in references:
            ref_ngrams = Counter(everygrams(reference, min_len, max_len))
            tpfn = sum(ref_ngrams.values())  # True positives + False negatives.

            overlap_ngrams = ref_ngrams & hyp_ngrams
            tp = sum(overlap_ngrams.values())  # True positives, i.e. numerator.

            # While GLEU is defined as the minimum of precision and
            # recall, we can reduce the number of division operations by one by
            # instead finding the maximum of the denominators for the precision
            # and recall formulae, since the numerators are the same:
            #     precision = tp / tpfp
            #     recall = tp / tpfn
            #     gleu_score = min(precision, recall) == tp / max(tpfp, tpfn)
            n_all = max(tpfp, tpfn) # denominator.

            this_gleu = Fraction(tp, n_all, _normalize=False)

            if this_gleu > highest_gleu_score:
                highest_gleu_score = this_gleu


        # use the reference yielding the highest score
        if highest_gleu_score:
            corpus_n_match += highest_gleu_score.numerator
            corpus_n_all += highest_gleu_score.denominator
```

But you do have to use a py2.x and py3.x compatible Fraction at the imports section of the code: 

```python
try:
    fractions.Fraction(0, 1000, _normalize=False)
    from fractions import Fraction
except TypeError:
    from nltk.compat import Fraction
```

nltk/translate/gleu_score.py
```diff
@@ -80,16 +81,113 @@ def sentence_gleu(reference, hypothesis, min_len=1, max_len=4):
...
+            _, n_match, n_all = max(scores_by_ref)
+            corpus_n_match += n_match
+            corpus_n_all += n_all
+
+    # corner case: empty corpus or empty references---don't divide by zero!
+    if corpus_n_all == 0:
+        gleu_score = 0.0
+    else:
+        gleu_score = corpus_n_match / corpus_n_all
+
```
alvations, line 192: Actually, if you like the idea of `Fraction`, we should do it all the way and the ultimate gleu_score should use `Fraction` too.

alvations: After some thought, this it the final score that users are getting float here should be better.

nltk/translate/gleu_score.py
alvations: Ah, I think Fraction gets kind of trippy on the fringe cases where numerator is 0 and denominator cannot be fixed this way.

alvations: How about this solution: https://gist.github.com/alvations/e48fdc0a46013d0d85cab88111c13b86 ?

alvations (commented):
@goodmami Depending on whether you like the Fraction or keeping separate floats. If it's the latter, it should be good to merge =)

alvations:
@goodmami some strange stuff is happening on the Jenkins server for Py2.7 and Py3.5 . I'm not sure what it is though. Perhaps it's coming from https://github.com/nltk/nltk/pull/1691 

Maybe @ExplodingCabbage @fievelk might have a better idea.

goodmami:
Yes, I noticed the failing tests looked odd, but I'm not sure what's causing it. Maybe the other people do?

Your Fraction-based solution looks good in general. I've never used the Fraction class, so I don't know how well it performs. It's getting late for me, so if the tests pass for you and performance looks good, feel free to add it in, otherwise I can take a closer look tomorrow.

alvations:
[CI: retest]

alvations:
@goodmami No worries, it's a timeout issue. On different CI retest runs, it's timing out on different python versions, my guess is third-party downloads are taking way longer than they should.

fievelk:
@goodmami @alvations In theory, a timeout is triggered on CI builds when they take more than 3 times the time needed (on average) by previous successful builds. I can set a fixed timeout (something like 60 minutes) to see if this solves the issue. Unfortunately I don't think I'll be able to do it until tomorrow.

goodmami:
Thanks for explaining the Fraction class and giving an example. I hadn't used that before.

However, after experimenting, I don't think I'll use it, because while it is a convenient mutable tuple for storing the numerator and denominator with comparison based on the computed fraction, it:

1. It relies on the undocumented (not part of Python's public API) and relatively new `_normalize` parameter, which you can't understand without, e.g., reading the source code to the fractions module,
2. It brings in another dependency, and
3. It is ~12% slower (in my local benchmarks) than just storing a tuple and comparing on the computed float

I did, however, change the variable name `scores_by_ref` to `hyp_counts` and I only store the numerator and denominator, then I compute the float when I do the `max()` operation. Performance is equivalent, but I hope this is less "obtuse" :)

I also removed an unused import.

I hope this is adequate

alvations:
Great! It looks good to merge now =)
Just one minor thing, add yourself and Mike Schuster to the author/contributor at the top of the code =) 

----

Actually, [Fraction](https://docs.python.org/3.5/library/fractions.html) is an official Python object, the NLTK version was there to patch the Py2to3 differences, https://docs.python.org/3.5/library/fractions.html =)

I agree that Fraction is rather slow, possibly because of [`gcd`](https://docs.python.org/3.5/library/fractions.html#fractions.gcd) and that `_normalize` is not documented. We should report this upstream to CPython but the PEP process can be painful =)

The plus point is that when catching fringe cases, we can use some Fraction tricks but that can be easily done with checking numerator and denominator too.

goodmami:
RE Fraction, yes I saw that it was an official Python module. So "dependency" may be too strong a word, but it is a non-builtin that has different behavior across Python versions, so there's some nonzero amount of headache associated with keeping the code updated.

Also, I'd rather be explicit about catching fringe cases than relying on Fraction tricks.

I'll add myself and Mike Schuster to the contributor list. And I think you originally wrote the code? But Schuster apparently designed the metric (claimed via off-list email), so... should it be:

* Authors: Liling Tan
* Contributors: Mike Schuster, Michael Wayne Goodman

alvations:
Hmmm, I'll prefer this though =):

 - Contributors: Mike Schuster, Michael Wayne Goodman, Liling Tan 

I guess Mike Schuster or the GNMT guys would be the true authors.

goodmami:
> Hmmm, I'll prefer this though =):
>
> * Contributors: Mike Schuster, Michael Wayne Goodman, Liling Tan

Done

alvations (approved):

stevenbird:
Thanks @goodmami, @alvations – nice work

PR #1684, opened 2017-04-13
Fix typo in `tokenize.word_tokenize()` docstring

gmarmstrong:
Unless I'm misunderstanding something, `nltk.tokenize.word_tokenize()` splits text into words, not into sentences (which `nltk.tokenize.sent_tokenize()` does).


stevenbird:
Um, yes, indeed! Thanks @gmarmstrong

PR #1691, opened 2017-04-16
.gitignore some non-ignored stuff that running `tox` creates

ExplodingCabbage:
Just now after I ran `tox` from the root of the repo I saw all this untracked stuff:

```
mark@lunchbox:~/nltk$ git status
On branch develop
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	brown.embedding
	model.crf.tagger
	pylintoutput
```

I'm guessing it should all be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,14 +10,16 @@ dist/
 nltk.egg-info/
 web/_build
 
-# Unit test / coverage reports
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @ExplodingCabbage

PR #1692, opened 2017-04-16
Make synset_from_pos_and_offset officially a public method

### LINKED ISSUE #1618: Expose _synset_from_pos_and_offset as a public method
Two reasons for this:

1. Given that http://stackoverflow.com/a/27145655/1709587 has thousands of views and 6 upvotes, it seems that people are already using it. Pragmatism suggests that it should therefore be exposed publicly; it's not good to have a notionally-private-but-public-in-practice part of an API.
2. The fact that NLTK chooses to expose the `offset()` of synsets via a public method means that there's currently no philosophy of hiding the (ugly! awful! fragile!) offset-based data structure of WordNet, and given that offset information is being exposed, looking up a synset by offset (perhaps the single most basic offset-related task you'd ever want to do, more fundamental than wanting to know a given synset's offset) seems like something that NLTK should support and is an odd gap in the current API.

Precisely because of point 1, the old `_`-prefixed method should remain too to preserve backwards compatibility, at least until the next major release.

ExplodingCabbage:
It was previously being used as one by many users of the library, despite
being underscored-prefixed, so we might as well mark it as being public.

Resolves https://github.com/nltk/nltk/issues/1618

- [x] We should comment on http://stackoverflow.com/a/27145655/1709587 to note this change after merging.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -293,7 +293,7 @@ class Lemma(_WordNetObject):
         return "%s('%s.%s')" % tup
 
     def _related(self, relation_symbol):
-        get_synset = self._wordnet_corpus_reader._synset_from_pos_and_offset
+        get_synset = self._wordnet_corpus_reader.synset_from_pos_and_offset
         return sorted([
             get_synset(pos, offset)._lemmas[lemma_index]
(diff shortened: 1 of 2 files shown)


stevenbird:
I'll adopt this, but add a deprecation warning to the old method

PR #1698, opened 2017-04-23
Fixing a syntax error in the wordnet module.

dimazest:


stevenbird:
Oops, thanks @dimazest!

PR #1709, opened 2017-05-04
Float checking in doctest should use round/ellipsis

alvations:
c.f. #1333 , non-ellipsis doctest and floating point is causing uncertain checks. Preferred float checking methods should be explicitly using  (i) `# doctest: +ELLIPSIS` or (ii) `round(float)`.

### DIFF
--- a/nltk/tag/__init__.py
+++ b/nltk/tag/__init__.py
@@ -51,8 +51,8 @@ of ``None``.
 
 We evaluate a tagger on data that was not seen during training:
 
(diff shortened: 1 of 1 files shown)


alvations:
Closing this PR. The issue with float checking should consider [PEP485](https://www.python.org/dev/peps/pep-0485/) and a large problem than it seems. 

But this should be high priority because doctest is failing and breaking tests on CI.

Issue #1338, opened 2016-03-22
panlex_lite installation via nltk.download() appears to fail


grayben:
Platform: Python 3.5 on Mac OS X 10.11.2
Steps to reproduce:
1. $ python3
2. >>> import nltk; nltk.download('all', halt_on_error=False)

Symptoms:
 # Partial console write:
[nltk_data]    | Downloading package panlex_lite to
[nltk_data]    |     /Users/beng/nltk_data...
[nltk_data]    |   Unzipping corpora/panlex_lite.zip.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 664, in download
    for msg in self.incr_download(info_or_id, download_dir, force):
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 543, in incr_download
    for msg in self.incr_download(info.children, download_dir, force):
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 529, in incr_download
    for msg in self._download_list(info_or_id, download_dir, force):
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 572, in _download_list
    for msg in self.incr_download(item, download_dir, force):
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 549, in incr_download
    for msg in self._download_package(info, download_dir, force):
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 638, in _download_package
    for msg in _unzip_iter(filepath, zipdir, verbose=False):
  File "/usr/local/lib/python3.5/site-packages/nltk/downloader.py", line 2039, in _unzip_iter
    outfile.write(contents)
OSError: [Errno 22] Invalid argument

stevenbird:
@grayben – would you please install the current version of NLTK and report if you still have this issue?

grayben:
@stevenbird sorry for my delay in replying - you know how uni assignments can be!
I was experiencing the issue on v3.2. I just upgraded to v3.2.1 and am having the same issue.

alvations:
@grayben How did you install NLTK? Do you have an error when downloading a single corpus, e.g. `nltk.download('brown')`? Do you have an error when using Python2.7?

grayben:
@alvations 
1. I installed NLTK for python2 and python3 via pip and pip3, respectively.
[...]

PR #1714, opened 2017-05-10
[FIX] Protection against invalid data collections added to nltk

AvnerCohen:
https://github.com/nltk/nltk/issues/1338
https://github.com/nltk/nltk_data/issues/35
https://github.com/nltk/nltk/issues/882

These are all cases of (what I understand) bad packages added to the nltk index.

This causes pain :), the idea here is simply to protect the code from such cases.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -859,8 +859,10 @@ class Downloader(object):
             for child in queue:
                 if isinstance(child, Collection):
                     queue.extend(child.children)
-                else:
+                elif isinstance(child, Package):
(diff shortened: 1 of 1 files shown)


stevenbird:
Applied in https://github.com/nltk/nltk/commit/60258a9fb61f3f5c8d9043356ab80f34959ef0c5. Thanks @AvnerCohen

AvnerCohen:
@stevenbird Cool!
Don't you think the `pass` on the else might hide issues that will be very hard to track down?

PR #1716, opened 2017-05-12
(Do not merge) Testing CI

alvations:
Possibly, CI is breaking for Python3.5 because of some outdated index on other PR. This PR is for testing purposes only. Please do not merge.

----

First test failing due to rapid push, [a standard CI failure; no one wants to fix this queueing mechanism](https://github.com/BitFunnel/BitFunnel/issues/280)

~~It looks like if the `downloader.py` isn't changed, the xml remains stale long after timeout. I've no idea why that happens only for Python3.5 though.~~

LazyCorpusLoader materializes in Python3.5 but not 3.4. Something is different in how Zipfile / getattrs works. Any idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,3 +47,6 @@ For license information, see [LICENSE.txt](LICENSE.txt).
 - NLTK corpora are provided under the terms given in the README file for each
   corpus; all are redistributable, and available for non-commercial use.
 - NLTK may be freely redistributed, subject to the provisions of these licenses.
+
+
+<!-- testing CI -->
--- a/README.md
+++ b/README.md
@@ -47,6 +47,3 @@ For license information, see [LICENSE.txt](LICENSE.txt).
 - NLTK corpora are provided under the terms given in the README file for each
   corpus; all are redistributable, and available for non-commercial use.
 - NLTK may be freely redistributed, subject to the provisions of these licenses.
-
-
-<!-- testing CI -->
--- a/nltk/corpus/util.py
+++ b/nltk/corpus/util.py
@@ -9,7 +9,7 @@
 #{ Lazy Corpus Loader
 ######################################################################
 
-from __future__ import unicode_literals
+from __future__ import unicode_literals, print_function
 import re
 import gc
 import nltk
@@ -22,7 +22,7 @@ class LazyCorpusLoader(object):
     """
     To see the API documentation for this lazily loaded corpus, first
     run corpus.ensure_loaded(), and then run help(this_corpus).
-    
+
     LazyCorpusLoader is a proxy object which is used to stand in for a
     corpus object before the corpus is loaded.  This allows NLTK to
     create an object for each corpus, but defer the costs associated
@@ -38,7 +38,7 @@ class LazyCorpusLoader(object):
     NLTK data package.  Once they've properly installed the data
     package (or modified ``nltk.data.path`` to point to its location),
     they can then use the corpus object without restarting python.
-    
+
     :param name: The name of the corpus
     :type name: str
     :param reader_cls: The specific CorpusReader class, e.g. PlaintextCorpusReader, WordListCorpusReader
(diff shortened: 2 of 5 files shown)


alvations:
The problem seems to be that the way Zipfile works in Python3.5 and how LazyCorpusReader works is incompatible when the data does exist:

```python
$ python
Python 2.7.13 (default, Dec 18 2016, 07:03:39) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> from nltk.corpus import panlex_lite
>>> panlex_lite
<PanLexLiteCorpusReader in u'.../corpora/panlex_lite' (not loaded yet)>
>>> exit()

liling.tan@ASRHQ289:~/git-stuff/nltk$ python3
Python 3.5.2 (default, Oct 11 2016, 04:59:56) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
>>> from nltk.corpus import panlex
ImportError: cannot import name 'panlex'
>>> exit()
```

alvations:
So it seems like the source of the errors come from the difference in `getattrs` behavior in Python3.5. There's no documentation on the https://docs.python.org/3/whatsnew/3.5.html so we've to trawl the CPython commits or ask around about it =(


This hack in the PR with the try-except works to satisfy the test but it's **shouldn't be merge**, it will fail to raise the exception when user doesn't have the data.

----

From the looks, it seems like the inspect function has some changes that checks for a `__wrapped___` attribute from  https://bugs.python.org/issue20691 

```
    /Users/liling.tan/git-stuff/nltk/.tox/py35/lib/python3.5/site-packages/nose/failure.py line 39 in runTest
      raise self.exc_val.with_traceback(self.tb)
    /Users/liling.tan/git-stuff/nltk/.tox/py35/lib/python3.5/site-packages/nose/plugins/manager.py line 154 in generate
      for r in result:
    doctest_nose_plugin.py line 106 in loadTestsFromModule
      for suite in super(DoctestPluginHelper, self).loadTestsFromModule(module):
    /Users/liling.tan/git-stuff/nltk/.tox/py35/lib/python3.5/site-packages/nose/plugins/doctests.py line 228 in loadTestsFromModule
      tests = self.finder.find(module)
    /usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/doctest.py line 924 in find
      self._find(tests, obj, name, module, source_lines, globs, {})
    /usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/doctest.py line 983 in _find
      if ((inspect.isroutine(inspect.unwrap(val))
    /usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/inspect.py line 471 in unwrap
      while _is_wrapper(func):
    /usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/inspect.py line 465 in _is_wrapper
      return hasattr(f, '__wrapped__')
    /Users/liling.tan/git-stuff/nltk/nltk/corpus/util.py line 129 in __getattr__
      raise e
    /Users/liling.tan/git-stuff/nltk/nltk/corpus/util.py line 122 in __getattr__
      self.__find() # First, find whether it exist.
    /Users/liling.tan/git-stuff/nltk/nltk/corpus/util.py line 82 in __find
      except LookupError: raise e
    /Users/liling.tan/git-stuff/nltk/nltk/corpus/util.py line 79 in __find
      root = nltk.data.find('{}/{}'.format(self.subdir, self.__name))
    /Users/liling.tan/git-stuff/nltk/nltk/data.py line 648 in find
      raise LookupError(resource_not_found)
   LookupError: 
   **********************************************************************
     Resource 'corpora/panlex_lite' not found.  Please use the NLTK
     Downloader to obtain the resource:  >>> nltk.download()
     Searched in:
       - '/Users/liling.tan/nltk_data'
       - '/usr/share/nltk_data'
       - '/usr/local/share/nltk_data'
       - '/usr/lib/nltk_data'
       - '/usr/local/lib/nltk_data'
   **********************************************************************


```

alvations:
The cleanest way to handle this is to disable the LazyCorpusLoader interface for packages not distributed through `nltk_data`. 

The other way is to hallucinate a `__wrapped` in the `LazyCorpusLoader.__getattr__` like this before the `self.__load()`:

```
    if attr == '__wrapped__':
            return False
```

But I'm not sure enough whether this will cause problems later on as more python functions checks for the `__wrapped__` attribute.

Issue #1725, opened 2017-05-18
NLTK depends on corenlp, should not be optional sub dependency


pykler:
Here is the exception with NLTK 3.2.3

```
.pyenv/lib/python2.7/site-packages/nltk/__init__.py:128: in <module>
    from nltk.chunk import *
.pyenv/lib/python2.7/site-packages/nltk/chunk/__init__.py:157: in <module>
    from nltk.chunk.api import ChunkParserI
.pyenv/lib/python2.7/site-packages/nltk/chunk/api.py:13: in <module>
    from nltk.parse import ParserI
.pyenv/lib/python2.7/site-packages/nltk/parse/__init__.py:81: in <module>
    from nltk.parse.corenlp import CoreNLPParser, CoreNLPDependencyParser
.pyenv/lib/python2.7/site-packages/nltk/parse/corenlp.py:17: in <module>
    import requests
E   ImportError: No module named requests
```

alvations:
To use the `corenlp` module, install with

```
pip install -U nltk[corenlp]
```

And for fuzz-free installation:

```
pip install -U nltk[all]
```

Sorry about the lack of documentation for `corenlp` api, @dimazest will be preparing it in the coming days.  I think the documentation for the optional `corenlp` installation would be documented by then too.

pykler:
@alvations all i did was import nltk, and it crashed. Are you not supposed to install nltk without any optional dependencies?

alvations:
Ah I see. Yes then adding `requests` would be necessary because of it was exposed at https://github.com/nltk/nltk/blob/develop/nltk/parse/__init__.py#L81 

I think commenting out the import there would be more appropriate than adding additional dependencies that only one module uses.

stevenbird:
I think this is resolved.

stevenbird:
Can anyone replicate this problem?

alvations:
@stevenbird The `requests` package needs to be installed if `corenlp` is to be used. 

#1725 will make the `requests` optional but users have to install with `pip install -U nltk[corenlp]` but that's on the develop branch.

Could the master branch be repackaged and an update on PyPI be made? `pip install nltk` from the current tarball on PyPI still throws the error:

```
alvas@ubi:~$ mkvirtualenv testnltk
Using base prefix '/usr'
New python executable in /home/alvas/.virtualenvs/testnltk/bin/python3
Also creating executable in /home/alvas/.virtualenvs/testnltk/bin/python
[...]

PR #1727, opened 2017-05-18
Don’t import requests during the module import. #1725

dimazest:


pykler:
Corenlp is still being imported upon `import nltk` this means that even if now import nltk works some of the generic nltk features might still be broken. The right approach "if corenlp" is not part of the required packages, is to not import it at all rather than move the import for the requests library.

alvations:
@pykler is right. If the `corenlp` interface is optional, it shouldn't be expose to the top level imports at https://github.com/nltk/nltk/blob/develop/nltk/parse/__init__.py#L81 , commenting out L81 on the `nltk.parse.__init__.py` is needed. 

@dimazest could you help to check whether commenting out `nltk.parse.__init__.py` causes any test to fail since the namespace use in the unittest or doctest might be different? Thanks in advance!

stevenbird:
Here's an alternative approach, to convert the load-time error to a runtime error. If we did it this way, we avoid the inconvenience of a longer import statement when using corenlp (and needing to remember which modules are included in package imports).

https://github.com/nltk/nltk/blob/develop/nltk/classify/maxent.py#L56

dimazest:
Does it still requires a change?

I would still import corenlp classes in the .parse module, even if requests are not installed. In this way the corenlp classes are not different from others.

I would not silence the import error, because when the corenlp interface is used, but requests are not installed, then seeing an import error is more informative than seeing a name error.

pykler:
Again my comment is, if corenlp is required, then it should not be an
optional dependency. Otherwise everyone who wants to install nltk and have
the default set of use cases installed, would always need to do, `pip
install nltk[corenlp]`. Moreover breaking much already existing code bases
and install_requires. For me, I pinned to 3.2.2 until this is resolved.

Hatem Nassrat

dimazest:
There are two conflicting things. First is to be able to use NLTK without any dependencies installed. For example, someone just extracted nltk distribution and wants to use the minimal functionality.

Second, NLTK has dependencies, and even if setup.py is used, requiring a user to install numpy among other dependencies is too much.

NLTK should be impoetable without dependencies installed (I hope that is fixed, but we should have a test for that, so the situation doesn't repeat).

Now the question is what to do if you want to use a part of NLTK that has a dependency? For example, you've been using the NLTK blue score function, and now got into parsing and want to use corenlp​. You already have NLTK installed, but without the extra. When you use the code that has an uninstalled dependency, you get an import error and know exactly what should be installed (unless you've read the to be written documentation and installed nltk[corenlp]).

Why is it a good idea import everything? I guess when ipython is used, so people could inspect provided classes and always see the same list.

Why do we have extras in setup.py? Because there should be a complete list of dependencies.

-- I'm sorry if it's not clear, typing on the phone is harder than I expected.

pykler:
Sounds good, if thats the design then the exceptions raised should indicate
potentially that sub dependencies are needed. This is more of a feature
request at this point.

Hatem Nassrat

On Fri, May 19, 2017 at 1:00 PM, Dmitrijs Milajevs <notifications@github.com

stevenbird:
@dimazest, I agree. But @pykler said that [Corenlp is still being imported upon import nltk](https://github.com/nltk/nltk/pull/1727#issuecomment-302577992). That would seem to be a showstopper. I haven't been able to replicate this problem. If there is no such problem, then I think we don't need to do anything further, right?

dimazest:
I'll give it a closer look.

pykler:
Corenlp is being imported in the ntlk.parse which is being imported in
nltk.__init__  ... See the stack trace shared on #1725

stevenbird:
I'll package a new release as soon as the CI test has completed.

stevenbird:
I think that @dimazest's solution of embedding the imports is sufficient.

PR #1734, opened 2017-05-24
add bigrams function import to book.py

dan-sf:
Add missing bigrams function import to book.py. Adding this import because the bigrams function is used in chapter 1 section 3.3 of the nltk book (http://www.nltk.org/book/ch01.html).


stevenbird:
Thanks @dsfcode. That must be a regression since we originally had automated testing of all book examples.

dan-sf:
@stevenbird no problem, maybe I'll take a quick look at the tests as well since I'm currently working my way though the book anyways.

Issue #1632, opened 2017-02-15
StanfordPnOSTagger failed to tag the underscore


dongfang91:
Please see the python code below:

from nltk.tag.stanford import StanfordPOSTagger

english_postagger = StanfordPOSTagger('./models/english-left3words-distsim.tagger', './stanford-postagger.jar')
english_postagger.java_options='-mx4096m'

print english_postagger.tag(['_'])

Here is my result:
[(u'', u'SYM')]

The underscore just disappeared. I checked the code, that's because in function "parse_output(self, text, sentences = None)", the code "word_tags = tagged_word.strip().split(self._SEPARATOR)" would delete all "_SEPARATOR" which is underscore based on the result of StandfordTagger.

One suggestion, probably we can use this line of code instead:  
word_tags = tagged_word.strip().rsplit(self._SEPARATOR,1)

ludens11:
where should i put those code to fix underscore issue?

dimazest:
Please create a pull request. https://help.github.com/articles/creating-a-pull-request/

stevenbird:
This should be resolved in our new interface to the Stanford Tools. If not, please reopen.

Issue #1510, opened 2016-11-18
DependencyGraph or Stanford Parser API issues with sentences with "/"


alvations:
A user has reported that this sentence throws and `AssertionError` when using Stanford's `DependencyParser` API in NLTK for this sentence:

> for all of its insights into the dream world of teen life , and its electronic expression through cyber culture , the film gives no quarter to anyone seeking to pull a cohesive story out of its 2 1/2-hour running time . 

Code:

```python
>>> from nltk.parse.stanford import StanfordDependencyParser                                                                                       >>> dep_parser=StanfordDependencyParser(model_path="edu/stanford/nlp/models/lexparser/englishPCFG.ser.gz")                                        
>>> sent = 'for all of its insights into the dream world of teen life , and its electronic expression through cyber culture , the film gives no quarter to anyone seeking to pull a cohesive story out of its 2 1/2-hour running time . '
>>> dep_parser.raw_parse(sent)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Python/2.7/site-packages/nltk/parse/stanford.py", line 132, in raw_parse
    return next(self.raw_parse_sents([sentence], verbose))
  File "/Library/Python/2.7/site-packages/nltk/parse/stanford.py", line 150, in raw_parse_sents
    return self._parse_trees_output(self._execute(cmd, '\n'.join(sentences), verbose))
  File "/Library/Python/2.7/site-packages/nltk/parse/stanford.py", line 91, in _parse_trees_output
    res.append(iter([self._make_tree('\n'.join(cur_lines))]))
  File "/Library/Python/2.7/site-packages/nltk/parse/stanford.py", line 339, in _make_tree
    return DependencyGraph(result, top_relation_label='root')
  File "/Library/Python/2.7/site-packages/nltk/parse/dependencygraph.py", line 84, in __init__
    top_relation_label=top_relation_label,
  File "/Library/Python/2.7/site-packages/nltk/parse/dependencygraph.py", line 328, in _parse
    assert cell_number == len(cells)
AssertionError
```

Possibly, it might be how `DependencyGraph` is reading the output or that the Stanford output is inconsistent.  

More details on the setup for NLTK + Stanford tools is on https://gist.github.com/alvations/e1df0ba227e542955a8a#stanford-parser

hoavt-54:
Hi @alvations Any updates on this?
Thanks

alvations:
[...]

PR #1737, opened 2017-05-28
More tests to CoreNLP.

dimazest:
These are tests cases that are reported not to work with the Stanford interface #1510 and #1632.

Please check that the output actually makes sense.


stevenbird:
Looks good to me, thanks @dimazest

PR #1738, opened 2017-05-29
Build DecisionTree description str only if verbose is True

### LINKED ISSUE #1723: Unnecessary slow down on DecisionTreeClassifier
Just a minor unnecessary calculation that might become huge on large trees.

https://github.com/nltk/nltk/blob/develop/nltk/classify/decisiontree.py#L269

Even if verbose is false you are building that description string. Wouldn't it be a good idea to have that string building inside the if condition of verbose?

nelson-liu:
Fixes #1723 by moving the `descr` generation into the check for `if verbose` (so we don't have to build this string if `verbose` is `False`)


stevenbird:
Thanks @nelson-liu

PR #1739, opened 2017-05-29
Make vader negation case insensitive

sbagan:
From my point of view negated function of vader sentiment intensity analyzer is not working correctly now, as it looks for NEGATE parts in case-sensitive words, while NEGATE contains case-insensitive ones. This can be ilustrated by a simle example:

`
In [1]: from nltk.sentiment.vader import SentimentIntensityAnalyzer
`
`
In [2]: sid = SentimentIntensityAnalyzer()
`
`
In [3]: print(sid.polarity_scores('Not as good as previous'))
`
`
{'neu': 0.58, 'pos': 0.42, 'compound': 0.4404, 'neg': 0.0}
`
`
In [4]: print(sid.polarity_scores('not as good as previous'))
`
`
{'neu': 0.624, 'pos': 0.0, 'compound': -0.3412, 'neg': 0.376}
`

I suppose check in negated function should be case-insensitive. I tried to find similar issues or pull-requests, but haven't found any, which is surprising, as from my point of view this thing is quite huge. So please correct me if I'm wrong.

### DIFF
--- a/nltk/sentiment/vader.py
+++ b/nltk/sentiment/vader.py
@@ -27,6 +27,7 @@ import re
 import string
 from itertools import product
 import nltk.data
+from .util import pairwise
 
 ##Constants##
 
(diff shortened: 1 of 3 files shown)


nltk/sentiment/vader.py
```diff
@@ -88,18 +87,16 @@ def negated(input_words, include_nt=True):
...
     if include_nt:
         for word in input_words:
-            if "n't" in word:
+            if "n't" in word.lower():
                 return True
-    if "least" in input_words:
-        i = input_words.index("least")
-        if i > 0 and input_words[i-1] != "at":
+    for i, word in enumerate(input_words):
+        if i > 0 and word.lower() == "least" and input_words[i-1].lower() != 'at':
```
sbagan: Honestly I thinkg something like pairwise from https://docs.python.org/3/library/itertools.html#itertools-recipes should be used here, but I'm not sure what code convetion is used in the library, so I made it similar to what was before
`
def pairwise(iterable):
    "s -> (s0,s1), (s1,s2), (s2, s3), ..."
    a, b = tee(iterable)
    next(b, None)
    return zip(a, b)
`

stevenbird:
Nice work, thanks @sbagan!

PR #1742, opened 2017-05-31
Adds possibility to return most_informative_features list

jabou:
Same feature as in naivebayes.


stevenbird:
Thanks @jabou. I think some refactoring is required. Now that you've added `most_informative_features()`, `show_most_informative_features()` ought to use it, rather than duplicate the same code.

jabou:
Hi @stevenbird ,

while on first this is the same thing, `show_most_informative_features()` prints **all information** about most informative features. On the other hand, `most_informative_features()` returns dictionary where key is word and value is position (so it returns _some data_, not all).
I'm not sure how can I reuse one of this methods as `most_informative_features()` returns less data than `show_most_informative_features()` prints.

As I said, this is the same feature as in `naivebayes.py` with `most_informative_features()` and `show_most_informative_features()`.

One of the possibilities is to create private method i.e. `__generate_most_informative_features()` that returnes list of data, and in `show_most_informative_features()` It can just print that data out, and inside `most_informative_features()` it can iterate throught list and remove unnecessarily information and than return that clenaed list. But as I wasn't seen that `maxent.py` uses private methods, I was copied and changed existing code.

What is your suggestion?

alvations:
@jabou @stevenbird would something like this work:

```python
    def most_informative_features(self, n=10):
        """
        Generates the ranked list of informative features from most to least.
        """
        if hasattr(self, '_most_informative_features'):
            return self._most_informative_features[:n]
        else:
            self._most_informative_features = sorted(list(range(len(self._weights))),
                                                key=lambda fid: abs(self._weights[fid]),
                                                reverse=True)
            return self._most_informative_features[:n]

    def show_most_informative_features(self, n=10, show='all'):
        """
        :param show: all, neg, or pos (for negative-only or positive-only)
        :type show: str
        :param n: The no. of top features
        :type n: int
        """
        # Use None the full list of ranked features.
        fids = self.most_informative_features(None)
        if show == 'pos':
            fids = [fid for fid in fids if self._weights[fid] > 0]
        elif show == 'neg':
            fids = [fid for fid in fids if self._weights[fid] < 0]
        for fid in fids[:n]
            print('%8.3f %s' % (self._weights[fid],
                                self._encoding.describe(fid)))
```

stevenbird:
@jabou: would you please let us know what you think?

jabou:
Hi @stevenbird and @alvations ,

I have finally found some time to look at this.

Unfortunately, this is not what I meant with my implementation. @alvations I saw merging where you have made some changes with bayes as well. The problem with your implementation is that by definition most informative feature should be a tuple of name + value. What you are suggesting is to return an array of integers (I assume weight/values only) without words and this is not good for later usage.

I.e. with my solution for `most_informative_feature` for positive polarity will return:
```
{('restore', 2), ('improvement', 2), ('successful', 3), ('top', 4), ('joke', 4), ('smart', 2), ('god', 6), ('allow', 2), ('happy', 9), ('cheerful', 2), ('glorious', 2), ('improved', 2), ('excellence', 3), ('jesus', 3), ('funny', 8)}
```
and `show_most_informative_features` for same thing will return:
```
 10.134 happy==9 and label is 'n'
   9.115 top==4 and label is 'p'
   8.846 allow==2 and label is 'n'
   8.739 glorious==2 and label is 'n'
   8.186 joke==4 and label is 'n'
   7.730 jesus==3 and label is 'p'
   7.478 improved==2 and label is 'n'
   7.441 smart==2 and label is 'n'
   7.394 cheerful==2 and label is 'p'
   7.095 funny==8 and label is 'p'
   6.790 restore==2 and label is 'n'
   6.644 excellence==3 and label is 'n'
   6.644 god==6 and label is 'p'
   6.644 improvement==2 and label is 'p'
   6.644 successful==3 and label is 'n'
```
Main difference is that `show_most_informative_feature` will only **print** data with _full description_ (descriptive text, just for printing) and `most_informative_feature` will return stripped data for later usage.

What would be great is that `show_most_informative_feature` returns (not prints) and than we can reuse it in `most_informative_feature` to strip and return data, but this is not what we want.

My second proposal (mini refactoring would be):
```python
    def __calculate_most_informative_features(self, show='all'):
        fids = sorted(list(range(len(self._weights))),
                      key=lambda fid: abs(self._weights[fid]),
                      reverse=True)
        if show == 'pos':
            fids = [fid for fid in fids if self._weights[fid] > 0]
        elif show == 'neg':
            fids = [fid for fid in fids if self._weights[fid] < 0]

        return fids

    def show_most_informative_features(self, n=10, show='all'):
        """
        Prints all informations about most informative features
        :param show: all, neg, or pos (for negative-only or positive-only)
        """
        fids = self.__calculate_most_informative_features(show)
        for fid in fids[:n]:
            print('%8.3f %s' % (self._weights[fid],
                                self._encoding.describe(fid)))

    def most_informative_features(self, n=10, show='all'):
        """
        Return a list of the 'most informative' features used by this
        classifier.  For the purpose of this function, the
        informativeness of a feature ``(fname,fval)`` is equal to the
        highest value of P(fname=fval|label), for any label, divided by
        the lowest value of P(fname=fval|label), for any label:

        |  max[ P(fname=fval|label1) / P(fname=fval|label2) ]

        :param show: all, neg, or pos (for negative-only or positive-only)

        """
        # The set of (fname, fval) pairs used by this classifier.
        features = set()

        fids = self.__calculate_most_informative_features(show)
        for fid in fids[:n]:
            name, _, value = self._encoding.describe(fid).partition('==')
            value = re.sub('[^\d]', '', value)
            features.add((name, int(value)))

        return features
```

alvations:
@jabou I'm bad at function naming but I think given that the classifier object has `self._most_informative_features` attributes, it's up to the user how he/she wants to prints out the information. 

I think the `__calculate_most_informative_features()` function you're suggesting is not really calculating but sorting and this can be done by users who wants to have customized information from the `self._most_informative_features`. 

In any case, the `most_informative_features` will return all the information that's needed to get subset(s) of score/probability information about the features.

jabou:
Hi @alvations ,

I agree that naming of `__calculate_most_informative_features` should be better, this was just proposal how this could be done. I can name it just` __most_informative_features`.

I don't know if I agree with you that user should always process the data. My proposal was to add additional feature in maxent as we have in naivebayes. Sometimes you just want to compare which classifier is better for usecase that you have, and when comparing 2 or 3 classifiers you want to print out most informative features that that classifier have take.

In your PR I saw that you have removed that option in naive bayes and you have left all work to the user.

@stevenbird what is your proposal in which way this should go? Should print of most informative features be returned ti the naive bayes and added to the maxent or should this be leave out from the core and left to the user to do whatever he/she want with raw data?

alvations:
I find the following interface intuitive and more consistent with other NLTK parts, e.g. how we access attributes and functions in `nltk.corpus.wordnet`:

 - `self._most_informative_features` :  the actual **object** attribute that holds the sorted list of tuple consisting of (feature_name, feature_value). 

- `self.most_informative_features()`: the lazy loading **function** that instantiate the `self._most_informative_features` for the first time. No printing. 

- `self.show_most_informative_feature()`: the function that prints out the information with in the format that follows the NLTK book. Prints. 

@jabou do you agree with the motivations attribute and function above? 

----

For comparison across  classifiers, I would expect the users to do:

```python
clf1 = X()
clf1.fit()

cf2  = Y()
clf2.fit()

clf1_best_features, clf1_best_values = zip(*clf1.most_informative_features())
clf2_best_features, clf2_best_values = zip(*clf2.most_informative_features())

# If they want to compare, just the features. (I suspect the desired use)
# But still, if classifiers are in different spaces, (e.g. [-1, 1] vs [0, 1], this is still not comparable
print(clf1_best_features)
print(clf2_best_features)

# Do some magic (e.g. normalization) to make sure that features are the same.
do_magic(clf1_best_values, clf2_best_values)
```

Technically, the weights for 2 classifiers cannot be compared directly, esp. they're in different space. So I think it won't be right if we pre-supposed some techniques to compare classifier weights, and it's out-of-scope for NLTK. 

It would be up to users to choose which/what process they would like to do to make the feature space similar across classifiers. 

Individually, within each classifier, we could still figure out which feature is more salient and that we have the `show_most_informative_features`.

alvations:
@jabou we're trying to clean up the PR and issues a little before a release. 

Just checking in to see whether you agree with the changes made to the latest version of the `most_informative_features()` functions?

alvations:
Closing the PR to clean up the issue tracker. Please re-open if there's more questions on the `most_informative_features()` function.

PR #1743, opened 2017-06-01
Hotfix/bug on stanford segmenter

xu2333:
That’s exciting.

### DIFF
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -50,7 +50,7 @@ class StanfordSegmenter(TokenizerI):
 
     def __init__(self,
                  path_to_jar=None,
-                 path_to_slf4j='PLACEHOLDER',
+                 path_to_slf4j=None,
                  java_class=None,
                  path_to_model=None,
                  path_to_dict=None,
@@ -66,8 +66,6 @@ class StanfordSegmenter(TokenizerI):
                 searchpath=(), url=_stanford_url,
                 verbose=verbose)
         if path_to_slf4j is not None:
-            if path_to_slf4j == 'PLACEHOLDER':
-                path_to_slf4j = None
             slf4j = find_jar(
                 self._SLF4J, path_to_slf4j,
                 env_vars=('SLF4J', 'STANFORD_SEGMENTER',),
@@ -78,7 +76,7 @@ class StanfordSegmenter(TokenizerI):
 
         # This is passed to java as the -cp option, the old version of segmenter needs slf4j.
         # The new version of stanford-segmenter-2016-10-31 doesn't need slf4j
-        # When using new version, set 'path_to_slf4j = None' in the initilization of StanfordSegmenter
+        # When using old version, set 'path_to_slf4j=YOUR_PATH' in the initilization of StanfordSegmenter
         self._stanford_jar = os.pathsep.join(
             [_ for _ in [stanford_segmenter, slf4j] if not _ is None])
 
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -46,7 +46,6 @@ class StanfordSegmenter(TokenizerI):
     """
 
     _JAR = 'stanford-segmenter.jar'
-    _SLF4J = 'slf4j-api.jar'
 
     def __init__(self,
                  path_to_jar=None,
@@ -67,7 +66,7 @@ class StanfordSegmenter(TokenizerI):
                 verbose=verbose)
         if path_to_slf4j is not None:
             slf4j = find_jar(
-                self._SLF4J, path_to_slf4j,
+                'slf4j-api.jar', path_to_slf4j,
                 env_vars=('SLF4J', 'STANFORD_SEGMENTER',),
                 searchpath=(), url=_stanford_url,
                 verbose=verbose)
@@ -76,7 +75,7 @@ class StanfordSegmenter(TokenizerI):
 
         # This is passed to java as the -cp option, the old version of segmenter needs slf4j.
         # The new version of stanford-segmenter-2016-10-31 doesn't need slf4j
-        # When using old version, set 'path_to_slf4j=YOUR_PATH' in the initilization of StanfordSegmenter
+        # When using old version, set path_to_slf4j='/YOUR_PATH/slf4j-api.jar' in the initilization of StanfordSegmenter
         self._stanford_jar = os.pathsep.join(
             [_ for _ in [stanford_segmenter, slf4j] if not _ is None])
 
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -31,6 +31,8 @@ _stanford_url = 'https://nlp.stanford.edu/software'
 class StanfordSegmenter(TokenizerI):
     """
     Interface to the Stanford Segmenter
+    If stanford-segmenter version is older than 2016-10-31, then path_to_slf4j should explicitly be set to '/YOUR_PATH/slf4j-api.jar'::
+        seg = StanfordSegmenter(path_to_slf4j='/YOUR_PATH/slf4j-api.jar')
     >>> from nltk.tokenize.stanford_segmenter import StanfordSegmenter
     >>> seg = StanfordSegmenter()
     >>> seg.default_config('zh')
@@ -66,7 +68,7 @@ class StanfordSegmenter(TokenizerI):
                 verbose=verbose)
         if path_to_slf4j is not None:
             slf4j = find_jar(
-                'slf4j-api.jar', None,
+                'slf4j-api.jar', path_to_slf4j,
                 env_vars=('SLF4J', 'STANFORD_SEGMENTER',),
                 searchpath=(), url=_stanford_url,
                 verbose=verbose)
@@ -75,8 +77,6 @@ class StanfordSegmenter(TokenizerI):
 
         # This is passed to java as the -cp option, the old version of segmenter needs slf4j.
         # The new version of stanford-segmenter-2016-10-31 doesn't need slf4j
-        # When using old version, set path_to_slf4j='slf4j-api.jar' in the initilization of StanfordSegmenter
-        # The initilization program will search for your slf4j-api.jar according to the environment parameters you set
         self._stanford_jar = os.pathsep.join(
             [_ for _ in [stanford_segmenter, slf4j] if not _ is None])
 
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -70,9 +70,8 @@ class StanfordSegmenter(TokenizerI):
                 searchpath=(), url=_stanford_url,
                 verbose=verbose)
 
-        # This is passed to java as the -cp option, the segmenter needs slf4j.
-        self._stanford_jar = os.pathsep.join(
-            [_ for _ in [stanford_segmenter, slf4j] if not _ is None])
+        # Update: The new version of stanford-segmenter-2016-10-31 don't need the path of slf4j
+        self._stanford_jar = stanford_segmenter
 
         self._java_class = java_class
         self._model = path_to_model
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -29,10 +29,13 @@ _stanford_url = 'https://nlp.stanford.edu/software'
 
 
 class StanfordSegmenter(TokenizerI):
-    """
-    Interface to the Stanford Segmenter
-    If stanford-segmenter version is older than 2016-10-31, then path_to_slf4j should explicitly be set to '/YOUR_PATH/slf4j-api.jar'::
+    """Interface to the Stanford Segmenter
+    
+    If stanford-segmenter version is older than 2016-10-31, then path_to_slf4j
+    should be provieded, for example::
+
         seg = StanfordSegmenter(path_to_slf4j='/YOUR_PATH/slf4j-api.jar')
+
     >>> from nltk.tokenize.stanford_segmenter import StanfordSegmenter
     >>> seg = StanfordSegmenter()
     >>> seg.default_config('zh')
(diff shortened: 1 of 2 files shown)


xu2333:
That's exactly the issue #1652

stevenbird:
@dimazest would you please review

nltk/tokenize/stanford_segmenter.py
```diff
@@ -70,9 +70,8 @@ def __init__(self,
                 searchpath=(), url=_stanford_url,
```
dimazest, line 76: if `slf4j` is not used, then the line 67 can be removed. Also, `self._SLF4J` is not needed.

dimazest: `path_to_slf4j` also should be removed, line 52.

nltk/tokenize/stanford_segmenter.py
```diff
@@ -70,9 +70,8 @@ def __init__(self,
...
                 verbose=verbose)
 
-        # This is passed to java as the -cp option, the segmenter needs slf4j.
-        self._stanford_jar = os.pathsep.join(
-            [_ for _ in [stanford_segmenter, slf4j] if not _ is None])
+        # Update: The new version of stanford-segmenter-2016-10-31 don't need the path of slf4j
+        self._stanford_jar = stanford_segmenter
```
dimazest: should it be a list?

something like `self._stanford_jar = [stanford_segmenter]`

nltk/tokenize/stanford_segmenter.py
dimazest: and line 49 is redundant.

dimazest (changes requested):
I would keep the current code, because an old version of stanford-segmenter might be used, then slf4 is needed.

dimazest:
[CI: retest]

nltk/tokenize/stanford_segmenter.py
xu2333: No, the previous one is also a string because it's using "join", which is concatenating the two element.

nltk/tokenize/stanford_segmenter.py
xu2333: I could modify these! And I think we could write a version checker or something to help.

xu2333: I have removed the redundant parts and finished the tests locally. In addition, I think since the stanford-segmenter-2016-10-31 is released 7 months ago, it's acceptable to move forward.

dimazest:
On the other side, one might need to be able to use an older version, for example, to replicate an experiment. I'm not sure whether this change is necessary.

xu2333:
Actually, the update of stanford-segmenter.jar only need to modify path parameter point to the .jar file. 
And the stanfordNLP's Named Entity Recognizer of Chinese is relied on the output of Segmenter of Chinese. I read the recommendation for nltk from the homepage of stanfordNLP's NER and I think other students would do the same. I think this modification would benefit them a lot from those cmd bugs.

dimazest:
Do you mean that without this patch the code doesn't work?

nltk/tokenize/stanford_segmenter.py
```diff
@@ -46,10 +46,9 @@ class StanfordSegmenter(TokenizerI):
     """
 
     _JAR = 'stanford-segmenter.jar'
-    _SLF4J = 'slf4j-api.jar'
 
     def __init__(self,
-                 path_to_jar=None, path_to_slf4j=None,
```
dimazest, line 52: keep `path_to_slf4j`, but make it default to `'slf4j-api.jar'`. Also move it to the new line.

nltk/tokenize/stanford_segmenter.py
```diff
@@ -64,15 +63,9 @@ def __init__(self,
                 env_vars=('STANFORD_SEGMENTER',),
                 searchpath=(), url=_stanford_url,
                 verbose=verbose)
-        slf4j = find_jar(
```
dimazest, line 67: now here do something like
```python
if path_to_slf4j is not None:
    slf4j = find_jar(...)
else:
    slf4j = None

**Review by dimazest — CHANGES_REQUESTED** [review]
Once these changes are applied, by setting `path_to_slf4j` to `None` should not append the slf4j.

**Comment by dimazest** [comment]
By the way, have you considered using the CoreNLP instead? https://github.com/nltk/nltk/blob/develop/nltk/parse/corenlp.py

It should be faster, but the segmenter might need to be implemented.

**Comment by alvations** [comment]
The segmenter might need some additional work to the interface `nltk.parse.corenlp` because it needs to use different properties for different languages when firing the corenlp server from java.

File: nltk/tokenize/stanford_segmenter.py (hunk 4 above)
Comment (xu2333) [reply nltk/tokenize/stanford_segmenter.py:67]: I have finished the tests locally. The reason for using 'PLACEHOLDER' as default value is that  if path_to_slf4j is not None (like 'slf4j-api.jar'), the find_jar_iter in internals.py would execute:

    if path_to_jar is not None:
        if os.path.isfile(path_to_jar):
            yielded = True
            yield path_to_jar
        else:
            raise LookupError('Could not find %s jar file at %s' %
                            (name_pattern, path_to_jar))

But we should search the file using environment variables. Hence in default situation of using old version, we need to set path_to_slf4j to None.

**Comment by dimazest** [comment]
[unrelated to this PR]

@alvations does the returned json depend on the language? We need to think how to support multiple languages with the CoreNLP interface.

File: nltk/tokenize/stanford_segmenter.py (hunk 4 above)
Comment (dimazest) [reply nltk/tokenize/stanford_segmenter.py:67]: but in my example `find_jar` will not be called if `path_to_slf4j` is `None`.

File: nltk/tokenize/stanford_segmenter.py [hunk 5]
```diff
@@ -64,13 +65,20 @@ def __init__(self,
...
+        if path_to_slf4j is not None:
+            if path_to_slf4j == 'PLACEHOLDER':
```
dimazest: this will lead to a failure if ``path_to_slf4j is None`.

nltk/tokenize/stanford_segmenter.py
dimazest: Did you try my proposal? Something like:
```python
StanfordSegmenter(path_to_slf4j=None)
```

you have set `path_to_slf4j` to force exclusion of the `slf4j`. This is explicit and compatible with the current behavior.

nltk/tokenize/stanford_segmenter.py
```diff
@@ -64,13 +65,20 @@ def __init__(self,
...
                 self._SLF4J, path_to_slf4j,
                 env_vars=('SLF4J', 'STANFORD_SEGMENTER',),
                 searchpath=(), url=_stanford_url,
                 verbose=verbose)
+        else:
+            slf4j = None
 
-        # This is passed to java as the -cp option, the segmenter needs slf4j.
+        # This is passed to java as the -cp option, the old version of segmenter needs slf4j.
+        # The new version of stanford-segmenter-2016-10-31 doesn't need slf4j
```
dimazest, line 82: I would put this text to the docstring. Something like "if stanford-segmenter-2016-10-31 or newer is used, then path_to_slf4j should explicitly be set to None".

dimazest: Unless, NLTK detects the segmenter's version and makes the decision.

dimazest (changes requested):

nltk/tokenize/stanford_segmenter.py
xu2333: The 'PLACEHOLDER' is to enable the old version to execute.
When using old version, and write

segmenter = StanfordSegmenter()

We need to find the correct slf4j-api.jar, and the find_jar_iter in internals.py would execute the following:

    if path_to_jar is not None:
        if os.path.isfile(path_to_jar):
            yielded = True
            yield path_to_jar
        else:
            raise LookupError('Could not find %s jar file at %s' %
                            (name_pattern, path_to_jar))

and ignore the environment variables. We can only find correct slf4j-api.jar using the environment variables. So in this default situation, we need to set path_to_jar, which is the second parameter transferred to find_jar(), to None.

xu2333: For new version, you are right, we need to write:

StanfordSegmenter(path_to_slf4j=None)

I have tested it locally and it works.

nltk/tokenize/stanford_segmenter.py
stevenbird: How about the converse of the first proposal, where using the more recent versions is the default. Users will get a runtime error when they upgrade NLTK, but the docstring (and the release notes) can cover for this.

xu2333: I think take the recent version as the default is workable. The question is how to set the parameter of old version. I am waiting for dimazest's opinion.

dimazest: You can set the default value of `path_to_slf4j` to `None`. And then have an if like
```python
if path_to_slf4j is not None:
    slf4j = find_jar(path_to_slf4j=path_to_slf4j, ...)
else:
    slf4j = None
```

Then the only missing bit is the documentation, which needs to tell that if an old version if used, then a path to `path_to_slf4j` has to be set (most probably, to `'slf4j-api.jar'`.

xu2333: OK, I think the current code meets the requirement you mentioned.

dimazest: Yes, it looks so. Could you please document this behavior and mention that if an old version of the segmenter is used, then the `path_to_slf4j` has to be set and a good value for it is `'slf4j-api.jar'`.

Also line 49 `_SLF4J = 'slf4j-api.jar'` needs to be cleaned up.

xu2333: OK. I have deleted _SLF4J. 
As I have recorded, as long as the second parameter transmitted to find_jar is not None, the find_jar_iter would execute without environment parameters. So If the user only input 'slf4j-api.jar' instead of the path to slf4j-api.jar, I need to set this parameter to None.

nltk/tokenize/stanford_segmenter.py
```diff
@@ -76,7 +75,7 @@ def __init__(self,
 
         # This is passed to java as the -cp option, the old version of segmenter needs slf4j.
         # The new version of stanford-segmenter-2016-10-31 doesn't need slf4j
-        # When using old version, set 'path_to_slf4j=YOUR_PATH' in the initilization of StanfordSegmenter
+        # When using old version, set path_to_slf4j='/YOUR_PATH/slf4j-api.jar' in the initilization of StanfordSegmenter
```
dimazest: Could you please move this note into the docstring that starts on the line 33. It's also better to specify concretely that versions before stanford-segmenter-2016-10-31 need to be used like this.

dimazest: Something like
```
If stanford-segmenter version is older than 2016-10-31, then path_to_slf4j should explicitly be set to '/YOUR_PATH/slf4j-api.jar'::

   segmenter = StanfordSegmenter(path_to_slf4j='/YOUR_PATH/slf4j-api.jar')
```

xu2333: I think you have decided that we would choose the plan A, which is letting user input the full path of slf4j-api.jar. Hence I change the parameter back to path_to_slf4j.

nltk/tokenize/stanford_segmenter.py
dimazest: does it require a full path, i guess, a relative path (just 'slf4j-api.jar') should also work.

nltk/tokenize/stanford_segmenter.py
```diff
@@ -31,6 +31,8 @@
 class StanfordSegmenter(TokenizerI):
     """
     Interface to the Stanford Segmenter
```
dimazest: The final minor thing, could you format this nicely:
```python
"""Interface to the Stanford Segmenter.

If stanford-segmenter version is older than 2016-10-31, then path_to_slf4j
should be provieded, for example::

    seg = StanfordSegmenter(path_to_slf4j='/YOUR_PATH/slf4j-api.jar')

>>> from nltk.tokenize.stanford_segmenter import StanfordSegmenter
>>> seg = StanfordSegmenter()
>>> seg.default_config('zh')
...

"""
```

nltk/tokenize/stanford_segmenter.py
```diff
@@ -75,8 +77,6 @@ def __init__(self,
 
         # This is passed to java as the -cp option, the old version of segmenter needs slf4j.
         # The new version of stanford-segmenter-2016-10-31 doesn't need slf4j
-        # When using old version, set path_to_slf4j='slf4j-api.jar' in the initilization of StanfordSegmenter
-        # The initilization program will search for your slf4j-api.jar according to the environment parameters you set
         self._stanford_jar = os.pathsep.join(
             [_ for _ in [stanford_segmenter, slf4j] if not _ is None])
```
dimazest: I'm not sure that a list needs to be built here, try
```python
self._stanford_jar = os.pathsep.join(
    _ for _ in [stanford_segmenter, slf4j] if _ is not None
)
```

Also `if not _ is None` is strange `_ is not None` is much better.

xu2333: Yes, you are right, join() could take a sequence as input. And of course, _ is not None is more natural.

xu2333: And I am sure that a relative path would not work. When you execute find_jar(), whose second parameter is path_to_slf4j, it would directly search according to the path without environment parameters and then terminate. So we have 3 choices:
A. let user input full path
B. set the second parameter of find_jar() to None when path_to_slf4j is not None, which means that the path user input is just a signal to make sure that it's old version
C. revise the find_jar_iter() function in internals.py

dimazest:
👍 Looks good.

dimazest (approved):

stevenbird:
@xu2333: thanks for your contribution
@dimazest: thanks for your review. Can you confirm whether [this comment of yours](https://github.com/nltk/nltk/pull/1743#issuecomment-305541086) is still relevant?

dimazest:
@stevenbird that was addressed:  it's still possible to use an older version of Stanford tools.

stevenbird:
Thanks @dimazest!

PR #1749, opened 2017-06-10
Added functions to convert word array into numerical array

therealproprogrammer:
Added functions to convert word array into numerical array, based on the word's position in YourListOfWords.txt. YourListOfWords.txt is a text file with a word at each line. Users should feel free to add words to the text file manually (through Notepad) or through the function "AddWordToDictionary." Numerical arrays are important for machine learning as natural language neural networks can only understand numerical values.


stevenbird:
@therealproprogrammer: sorry, this is not a suitable submission to NLTK.

Issue #1750, opened 2017-06-12
span_tokenize failed when sentence contains double quotation


albertauyeung:
If we feed in a sentence with double quotation into TreebankWordTokenizer's span_tokenize function, there will be errors. Probably this is because the function sends the raw string input along with the tokenized string to the align_tokens function, without considering that the tokenize function would replace double quotation marks with something else.

alvations:
Thanks @albertauyeung for reporting the issue. Do you have an example where you met an error with the `TreebankWordTokenizer.span_tokenize()`?

Do you mean something like this?

```python
>>> from nltk.tokenize.treebank import TreebankWordTokenizer
>>> tbw = TreebankWordTokenizer
>>> tbw = TreebankWordTokenizer()
>>> s = '''This is a sentence with "quotes inside" and alsom some 'single quotes', etc.'''
>>> print(s)
This is a sentence with "quotes inside" and alsom some 'single quotes', etc.
>>> tbw.span_tokenize(s)
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/nltk/tokenize/util.py", line 230, in align_tokens
    start = sentence.index(token, point)
ValueError: substring not found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/nltk/tokenize/treebank.py", line 167, in span_tokenize
    return align_tokens(tokens, text)
  File "/usr/local/lib/python3.5/site-packages/nltk/tokenize/util.py", line 232, in align_tokens
    raise ValueError('substring "{}" not found in "{}"'.format(token, sentence))
ValueError: substring "``" not found in "This is a sentence with "quotes inside" and alsom some 'single quotes', etc."
```

**Suboptimal solution**:

```python
>>> s = '''This is a sentence with `` quotes inside '' and alsom some 'single quotes', etc.''' 
>>> tbw.span_tokenize(s)
[(0, 4), (5, 7), (8, 9), (10, 18), (19, 23), (24, 26), (27, 33), (34, 40), (41, 43), (44, 47), (48, 53), (54, 58), (59, 66), (67, 73), (73, 74), (74, 75), (76, 79), (79, 80)]
```

albertauyeung:
@alvations Yes. That's the exact error I got. Right now it seems we have to preprocess the sentence before submitting to span_tokenize.

alvations:
[...]

PR #1751, opened 2017-06-13
Patch for span_tokenize in TreebankWordTokenizer

albertauyeung:
Ref: #1750 The original code in `span_tokenize `will fail when the input sentence contains double quotation marks. This patch adds additional logic to convert "converted quotes (e.g. `` and '') back to double quotation before calling `align_tokens`.

### DIFF
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -163,7 +163,22 @@ class TreebankWordTokenizer(TokenizerI):
             True
 
         """
-        tokens = self.tokenize(text)
+        raw_tokens = self.tokenize(text)
+
+        # Find double quotes and converted quotes
+        matched = [m.group() for m in re.finditer(r'[(``)('')(")]+', text)]
+
+        # Convert converted quotes back to original double quotes
+        index = 0
+        tokens = []
+        for tok in raw_tokens:
+            if tok in ["\"", "``", "''"]:
+                if matched[index] == "\"":
+                    tokens.append("\"")
+                    index += 1
+                    continue
+            tokens.append(tok)
+
         return align_tokens(tokens, text)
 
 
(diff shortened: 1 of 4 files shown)


nltk/tokenize/treebank.py
```diff
@@ -163,7 +163,26 @@ def span_tokenize(self, text):
...
 
         """
-        tokens = self.tokenize(text)
+        raw_tokens = self.tokenize(text)
+
+        # Convert converted quotes back to original double quotes
+        # Do this only if original text contains double quote(s)
+        if '"' in text:
+            # Find double quotes and converted quotes
+            matched = [m.group() for m in re.finditer(r'[(``)(\'\')(")]+', text)]
```
alvations, line 172: 👍

nltk/tokenize/treebank.py
```diff
@@ -163,7 +163,26 @@ def span_tokenize(self, text):
...
+            
+            # Replace converted quotes back to double quotes
+            index = 0
+            tokens = []
+            for tok in raw_tokens:
+                if tok in ['"', "``", "''"]:
+                    tokens.append(matched[index])
+                    index += 1
+                else:
+                    tokens.append(tok)
```
alvations: If we use `matched.pop(0)`, we can avoid the index counting =)

How about this?

```python
tokens = [matched.pop(0) if tok in ['"', "``", "''"] else tok for tok in raw_tokens]
```

Essentially:

```python
tokens = []
for tok in raw_tokens:
    if tok in ['"', "``", "''"]:
        tokens.append(matched.pop(0))
    else:
        tokens.append(tok)
```

albertauyeung: Nice idea. Have made the change.

alvations:
Thank you, nice job @albertauyeung!! It looks good to merge now =)

We just have to wait for @stevenbird to merge this.

alvations (approved):

stevenbird:
Thanks for your contribution @albertauyeung and for the review @alvations

PR #1760, opened 2017-06-27
Corenlp typeerror

f0lie:
If I change the doctest from 9 to 10, the test goes through just fine. I don't know if this is just my setup or not, so I am avoiding changing the doctest.
```
Failed example:
    next(
        parser.raw_parse(
            "The broader Standard & Poor's 500 Index <.SPX> was 0.46 points lower, or "
            '0.05 percent, at 997.02.'
        )
    ).height()
Expected:
    9
Got:
    10
```

Anyway, originally I got TypeErrors when I run CoreNLPServer and I fixed it.

### DIFF
--- a/nltk/parse/corenlp.py
+++ b/nltk/parse/corenlp.py
@@ -66,7 +66,7 @@ class CoreNLPServer(object):
         # find the most recent code and model jar
         stanford_jar = max(
             jars,
-            key=lambda model_name: re.match(self._JAR, model_name)
+            key=lambda model_name: re.search(self._JAR, model_name).group()
         )
 
         if port is None:
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -215,6 +215,7 @@
 - sbagan
 - Zicheng Xu
 - Albert Au Yeung <https://github.com/albertauyeung>
+- Alex Diep
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer


f0lie:
So, it seems like it was on my end.

Gotta love mystery environment issues. 

Edit: I thought close meant closing the comment box.

nltk/parse/corenlp.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         # find the most recent code and model jar
         stanford_jar = max(
             jars,
-            key=lambda model_name: re.match(self._JAR, model_name)
+            key=lambda model_name: re.search(self._JAR, model_name).group()
```
dimazest, line 69: `.search()` might return `None` it might be an issue.

(Ha, `.match()` is looking at the beginning of a string.)

Also, it looks for the 'largest' string, while the comment says about the most recent version (a number).

I think---since I strongly believe that the usage of `CoreNLPServer` should be limited to unit tests and maybe to classroom demos where the environment is controlled---the code should be as simple as possible. Probably just getting the first element from the list of `jars` is enough.

The CoreNLP java server should be started outside of a Python process.

dimazest (commented):
Maybe it worth not looking for the most recent version, but to take the first available jar.

dimazest:
Regarding the doctest, could you see what the parse tree is. It might be because of a newer corenlp model.

f0lie:
After trying to use the thing, I agree people shouldn't be using it in any practical matter.

Maybe __init__ should print an warning against using corenlpserver outside of isolated cases. An written docstring is definitely needed.

PR #1768, opened 2017-07-05
fix "list index out of range" in handles_nonbreaking_prefixes

SwordYork:
The following code would raise an Error ("IndexError: list index out of range"), because the index has not been checked.
```
from nltk.tokenize.moses import MosesTokenizer
a = 'No doubt the given aid will be popular with those who aim to unearth the mysteries of the world of Art.'
tokenizer = MosesTokenizer()
e = tokenizer.tokenize(a)
```

PR #1769, opened 2017-07-05
add index check for handles_nonbreaking_prefixes

lmatt-bit:
token[i] could be the last element, add check to prevent exception.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -215,6 +215,8 @@
 - sbagan
 - Zicheng Xu
 - Albert Au Yeung <https://github.com/albertauyeung>
+- Shenjian Zhao 
+- Deng Wang <https://github.com/lmatt-bit>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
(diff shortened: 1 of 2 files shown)


alvations:
@lmatt-bit Would you be able illustrate what the patch fixes in 1-2 examples and add it to the doctest? That'll be helpful for users and learners to know why the patch is added =)

Thanks in advance!

alvations:
Also it seems like #1768 is trying to fix the same issue =)

lmatt-bit:
@alvations, i just found that #1768 is fixing the same issue.

alvations:
@lmatt-bit Thanks for the doctest example! If you don't mind, could you add your name and "Shenjian Zhao" (@SwordYork) to the [`AUTHORS.md`](https://github.com/nltk/nltk/blob/develop/AUTHORS.md))? 

Thereafter this PR looks good to merge.

lmatt-bit:
No problem. Updated.

alvations (approved):

stevenbird:
Thanks @lmatt-bit

PR #1773, opened 2017-07-07
[WIP] Added the NIST Tokenizer

alvations:
Added the NIST tokenizer based on the `mteval-v14.pl`  and `nmtpy`:

  - https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v14.pl#L926
  - https://github.com/lium-lst/nmtpy/blob/master/nmtpy/metrics/mtevalbleu.py#L162


stevenbird:
Good to have this @alvations

PR #1775, opened 2017-07-10
Fix some lgtm alerts

jhelie:
This PR fixes some alerts flagged by lgtm.com code queries. Most are minor improvements suggestions but a few some should fix unintentional code behaviour (e.g. wrong signatures in calls). 

The standard lgtm engineering analytics highlights dependencies, vulnerabilities and tracks [introduced and fixed alerts as the code changes](https://lgtm.com/projects/g/nltk/nltk/). There are currently [188 alerts](https://lgtm.com/projects/g/nltk/nltk/alerts/) flagged, probably better investigated by someone more familiar with the code base than me, but you can also investigate further by writing [your own custom queries](https://lgtm.com/docs/ql/about-ql) to explore your code base.

I personally find lgtm alerts most useful when using [PR integration](https://lgtm.com/docs/lgtm/config/pr-integration) as it allows to check and deal with potential issues before they find their way into master. 

Full disclosure: I work on lgtm.com, hence my interest! I'm not very familiar with `nltk` but I hope you find this useful, any question please ask.

### DIFF
--- a/nltk/classify/rte_classify.py
+++ b/nltk/classify/rte_classify.py
@@ -32,7 +32,7 @@ def ne(token):
         return True
     return False
 
-def lemmatize(word):
+def _lemmatize(word):
     """
     Use morphy from WordNet to find the base form of verbs.
     """
@@ -70,8 +70,8 @@ class RTEFeatureExtractor(object):
         self.hyp_words = set(self.hyp_tokens)
 
         if lemmatize:
-            self.text_words = set(lemmatize(token) for token in self.text_tokens)
-            self.hyp_words = set(lemmatize(token) for token in self.hyp_tokens)
+            self.text_words = set(_lemmatize(token) for token in self.text_tokens)
+            self.hyp_words = set(_lemmatize(token) for token in self.hyp_tokens)
 
         if self.stop:
             self.text_words = self.text_words - self.stopwords
--- a/nltk/collections.py
+++ b/nltk/collections.py
@@ -228,8 +228,7 @@ class AbstractLazySequence(object):
             length += len(pieces[-1]) + 2
             if length > self._MAX_REPR_SIZE and len(pieces) > 2:
                 return '[%s, ...]' % text_type(', ').join(pieces[:-1])
-        else:
-            return '[%s]' % text_type(', ').join(pieces)
+        return '[%s]' % text_type(', ').join(pieces)
 
     def __eq__(self, other):
         return (type(self) == type(other) and list(self) == list(other))
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -877,8 +877,7 @@ class PrettyLazyMap(LazyMap):
             length += len(pieces[-1]) + 2
             if length > self._MAX_REPR_SIZE and len(pieces) > 2:
                 return "[%s, ...]" % text_type(', ').join(pieces[:-1])
-        else:
-            return "[%s]" % text_type(', ').join(pieces)
+        return "[%s]" % text_type(', ').join(pieces)
 
 @python_2_unicode_compatible
 class PrettyLazyIteratorList(LazyIteratorList):
@@ -900,8 +899,7 @@ class PrettyLazyIteratorList(LazyIteratorList):
             length += len(pieces[-1]) + 2
             if length > self._MAX_REPR_SIZE and len(pieces) > 2:
                 return "[%s, ...]" % text_type(', ').join(pieces[:-1])
-        else:
-            return "[%s]" % text_type(', ').join(pieces)
+        return "[%s]" % text_type(', ').join(pieces)
 
 @python_2_unicode_compatible
 class PrettyLazyConcatenation(LazyConcatenation):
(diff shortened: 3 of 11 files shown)


jnothman:
fwiw, we fixed several small bugs over at scikit-learn thanks to lgtm's
alerts, and fixed more after suggesting a bespoke query that corresponds to
a frequent source of bugs. This is not mere issue tracker spam.

jhelie:
Thanks @jnothman ! very glad to read that :)

nltk/classify/rte_classify.py
```diff
@@ -32,7 +32,7 @@ def ne(token):
         return True
     return False
 
-def lemmatize(word):
+def _lemmatize(word):
```
stevenbird, line 35: @jhelie: would you please comment on this change? How do we know `lemmatize` isn't being used externally?

nltk/classify/rte_classify.py
```diff
@@ -70,8 +70,8 @@ def __init__(self, rtepair, stop=True, lemmatize=False):
         self.hyp_words = set(self.hyp_tokens)
 
         if lemmatize:
-            self.text_words = set(lemmatize(token) for token in self.text_tokens)
```
jnothman, line 73: It looks like the wrong fix, but it's clear that this has never been used or tested!

nltk/classify/rte_classify.py
jhelie: @stevenbird ah yes I did mean to comment on that. As I understand it the `__init__` method of the `RTEFeatureExtractor` class is broken due to a conflict between its `lemmatize` argument and the `lemmatize` function defined here. To fix this we need to change the name of one or the other.

To get an idea of the impact I ran 2 quick QL queries to check how many functions called `lemmatize` exist in the code base as well as how many calls to them there are:
- https://lgtm.com/query/2024710808/project:1870011/lang:python/
- https://lgtm.com/query/2021851733/project:1870011/lang:python/

Note that the first one wast just out of curiosity/sanity check, I could have modified the second one to only show calls to the lemmatize function defined in `rte_classify`.

According to these queries the lemmatize() function is not used anywhere outside `RTEFeatureExtractor.__init__` (and the one defined in `wordnet.py` is never used) so the modification here shouldn't have undesired impact.

jhelie: Out of curiosity I also wrote a query looking for all instantiation of the `RTEFeatureExtractor` class and again it appears it only occurs in `rte_classify.py` so modifying the `__init__` method would also be fine.
https://lgtm.com/query/2025790705/project:1870011/lang:python/

jhelie: (in any case this`_lemmatize()` change is merely a suggestion, based on the docstring it seems this file is work in progress and I don't know about naming convention or indeed possibly existing documentation referring to `lemmatize`)

nltk/classify/rte_classify.py
jhelie: hi @jnothman ! I agree with the second part of your statement but I'm not sure about the first one: choice of names aside this is essentially `set(list generating expression)` which I think is correct and was the original intention?
I'm not very familiar with that code and I might be totally off though!

jnothman: What I mean is that:
* the parameter to this class is broken
* you renamed a public function
* so you broke something that worked for the sake of fixing something that no one uses

So the right fix, assuming someone *does* want to use the parameter, is to rename the parameter or give the external function two names.

jhelie: ah I see. based on the queries results I mentioned above I'd say that `RTEFeatureExtractor` constructor and `lemmatize()` are both equally public and unused - but I think you're right that it'd probably make more sense to rename the parameter.

Maybe something like `use_lemmatize`? I'll wait to hear from @stevenbird on what he'd prefer to call it

stevenbird: Sorry for delay... I'm on a field trip and will take a few days

jhelie: @stevenbird no worries - thanks for the update! enjoy the field trip

jhelie: @stevenbird have you had a chance to take a look at these?

stevenbird: @jhelie: hopefully soon now

stevenbird: @jhelie: sorry for the slow response. I like @jnothman's suggestion to rename the parameter. The name `use_lemmatize` looks good to me. Thanks.

stevenbird:
@jhelie I'm merging this and will fix the names afterwards. Thanks for your contribution. We'd welcome more PRs like this.

alvations:
There's still more work on the LGTM and the spillover issues would be tracked on #1776.

jhelie:
@stevenbird thanks for your message and for merging this. I'm sorry I didn't get to update the parameters names myself last week - I see you just did so in 68390bdb53576a5c2a6a48b76d6195dc59e381fe :)

+@alvations: thanks both for the positive feedback, you might also be interested in turning on [PR integration](https://lgtm.com/docs/lgtm/using-lgtm-analysis-continuous-integration) to get the chance to act on possible issues at review time (in case you're wondering very very soon you'll also be able to configure which type of alerts you want to hear about in the PR).

We're always looking for feedback so do feel free to get in touch with any comment you may have. In the meantime I'll try to have a look at current alerts and submit another PR if I get the chance.

stevenbird:
:+1: to LGTM PR integration

PR #1779, opened 2017-07-14
Added NIST MT evaluation metric and added NIST international_tokenize() 

alvations:
Added the NIST score, based on http://dl.acm.org/citation.cfm?id=1289273 and [mteval-14.pl](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v14.pl#L889) implementation

I'm unsure of whether the final arithmetic mean computation is correct, i.e. https://github.com/alvations/nltk/blob/develop/nltk/translate/nist_score.py#L121 . 

~Should it be `return bp * sum(p_n)` or `return bp * sum(map(math.log, p_n))`? If anyone knows the right answer, please help to clarify. Thanks in advance!~

### DIFF
--- a/nltk/tokenize/nist.py
+++ b/nltk/tokenize/nist.py
@@ -49,7 +49,7 @@ class NISTTokenizer(TokenizerI):
     >>> from nltk.tokenize.nist import NISTTokenizer
     >>> nist = NISTTokenizer()
 
-    # Intput strings.
+    # Input strings.
     >>> albb = u'Alibaba Group Holding Limited (Chinese: 阿里巴巴集团控股 有限公司) us a Chinese e-commerce company...'
     >>> amz = u'Amazon.com, Inc. (/ˈæməzɒn/) is an American electronic commerce...'
     >>> rkt = u'Rakuten, Inc. (楽天株式会社 Rakuten Kabushiki-gaisha) is a Japanese electronic commerce and Internet company based in Tokyo.'
--- a/nltk/tokenize/nist.py
+++ b/nltk/tokenize/nist.py
@@ -3,7 +3,7 @@
 #
 # Copyright (C) 2001-2015 NLTK Project
 # Author:
-# Contributors: Ozan Caglayan
+# Contributors: Ozan Caglayan, Wiktor Stribizew
 #
 # URL: <http://nltk.sourceforge.net>
 # For license information, see LICENSE.TXT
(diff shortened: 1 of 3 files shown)


alvations:
I've squashed the commits and corrected the arithmetic mean implementation. 

This PR is ready for review =)

alvations:
[CI: retest]

nltk/tokenize/nist.py
```diff
@@ -34,9 +38,32 @@ class NISTTokenizer(TokenizerI):
...
+    >>> nist.tokenize(s, lowercase=True) == expected_lower  # Lowercased.
+    True
+
+    The international_tokenize() is the preferred function when tokenizing
+    non-european text, e.g.
+
+    >>> from nltk.tokenize.nist import NISTTokenizer
+    >>> nist = NISTTokenizer()
+
+    # Intput strings.
```
stevenbird: Input strings

nltk/tokenize/nist.py
```diff
@@ -87,8 +143,25 @@ def tokenize(self, text, preserve_case=True,
...
+    def international_tokenize(self, text, lowercase=False,
+                               split_non_ascii=True,
+                               return_str=False):
         text = text_type(text)
-        # Language independent regex.
-        text = self.lang_independent_sub(text)
-        #
+        # Different from the 'normal' tokenize(), STRIP_EOL_HYPHEN is applied
+        # first before unescaping.
+        regexp, subsitution = self.STRIP_SKIP
```
stevenbird: substitution

nltk/tokenize/nist.py
```diff
@@ -55,6 +82,35 @@ class NISTTokenizer(TokenizerI):
     LANG_DEPENDENT_REGEXES = [PUNCT, PERIOD_COMMA_PRECEED,
                               PERIOD_COMMA_FOLLOW, DASH_PRECEED_DIGIT]
 
+    # Perluniprops characters used in NIST tokenizer.
+    Number = text_type(''.join(set(perluniprops.chars('Number')))) # i.e. \p{N}
```
stevenbird: Unusual to have uppercase variable name?

stevenbird:
@alvations: you could identify yourself in the code as author (then after your name, say porting from Perl or whatever; also name the Perl authors?)

nltk/tokenize/nist.py
alvations: Using lowercase, those were bad habits ported from the perl script ;P

alvations:
Actually, I've no idea who wrote the NIST tokenizer, I would assume it's Doddington himself years ago but it's unclear. 

But it's so tied up with original BLEU from Papineni 16 years ago that they cite him as the original author of the nist mteval script.

See ftp://jaguar.ncsl.nist.gov/mt/resources/mteval-v14.pl

PR #1781, opened 2017-07-20
Python Implementation of CoreNLP Neural Network Dependency Parser

tsolakghukasyan:
Python implementation of CoreNLP's neural network dependency parser. I tried to keep the implementation close to the original one from CoreNLP, particularly main classes and method names.

Parser's features:
- uses arc-standard parsing system
- uses neural network for classifying transitions
- uses pre-computation trick for speeding up parsing and training
- uses cube activation in network's hidden layer
- uses/learns dense representations for words, part-of-speech tags and dependency labels.

This implementation provides the following methods:
- parsing raw tokenized sentences,
- parsing part-of-speech tagged sentences (parse, parse_sents),
- training new models (train),
- evaluating parser's performance on CoNLL-U data (test_conll),
- can read pretrained models from CoreNLP (load_from_model_file).

For parsing raw sentences, parser requires a part-of-speech tagger that uses the same tagset as the parser. When training a parser model, 'train' can also train a model for averaged perceptron tagger.

On a computer with 8GB RAM and Intel i5-6200U CPU (2.30GHz), a parser model demonstrated 75 sents/second, 1320 words/second parsing speed. 

That model was trained on SynTagRus Russian treebank (almost 60 000 sentences) and demonstrated 84.7% UAS and 78.7% LAS on test sentences. Along with the parser model, I also trained an averaged perceptron tagger model for the Russian language with 97.5% accuracy, which along with the parser model can be added to NLTK Data. I suggest creating a new module in NLTK Data for storing such parser models.

Thanks,
Tsolak Ghukasyan

### DIFF
--- a/nltk/parse/stanfordnndep.py
+++ b/nltk/parse/stanfordnndep.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """
 This module implements the dependency parsing algorithm presented by Danqi Chen and
 Christopher Manning at EMNLP 2014.
@@ -720,9 +721,9 @@ class DependencyParser(object):
 
 class DependencyTree(object):
     """ 
-	Represents a dependency parse tree. Stores a sentence's syntactic structure, but not its words.
-	Dependency relations and their types are stored in attributes 'head' and 'label' respectively.
-	"""
+    Represents a dependency parse tree. Stores a sentence's syntactic structure, but not its words.
+    Dependency relations and their types are stored in attributes 'head' and 'label' respectively.
+    """
 
     def __init__(self, tree=None):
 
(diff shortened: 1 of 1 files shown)


stevenbird:
@tsolakghukasyan: Thanks for your work; I hope we can include this in the toolkit. Note that a port of GPL'ed code is GPL, so we would need permission from the copyright holder.

@manning: would you please advise us?

stevenbird:
@tsolakghukasyan: I don't think we can go ahead without explicit permission from the copyright holder.

stevenbird:
@tsolakghukasyan: sadly, I don't think we can proceed with this.

alvations:
@tsolakghukasyan It will be a waste if the code isn't used somehow. IMHO, I suggest that you should create a new repository and packaged the code and license it under GPL. That way people can still use it although it's not in NLTK. 

Afterwards then getting a stamp of approval on the code from the Stanford NLP group and @manning would be a good thing too =)

PR #1784, opened 2017-07-24
specify tagset for russian tagger in tag init documentation

tsolakghukasyan:
Specifies the tagset for the russian tagger and adds an example in tag/init's documentation.

### DIFF
--- a/nltk/tag/__init__.py
+++ b/nltk/tag/__init__.py
@@ -19,12 +19,16 @@ the word ``'fly'`` with a noun part of speech tag (``'NN'``):
 
     >>> tagged_tok = ('fly', 'NN')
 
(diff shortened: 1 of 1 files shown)


tsolakghukasyan:
the code example in docs caused build errors (probably because the russian model was not installed), so I removed it.

stevenbird:
Hi @tsolakghukasyan, sorry for the delay. How about you say something about the tagger in the documentation still, so that people know it is available. E.g. mention that a russian tagger is available if you specify `lang="rus"` etc.

tsolakghukasyan:
Hi @stevenbird,

Edited the documentation according to your suggestion. I also added the doc example back but with a "# doctest: +SKIP" to avoid build errors.

stevenbird:
Thanks @tsolakghukasyan

Issue #1787, opened 2017-07-26
Failed to download NLTK data: HTTP ERROR 405 / 403


matthew-z:
```
>>> nltk.download("all")
[nltk_data] Error loading all: HTTP Error 405: Not allowed.

>>> nltk.version_info
sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)

```

Also, I tried to visit [https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/packages/corpora/cmudict.zip](https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/packages/corpora/cmudict.zip). Got the same HTTP 405 ERROR.

Find the same problem on stackoverflow: [https://stackoverflow.com/questions/45318066/getting-405-while-trying-to-download-nltk-dta](https://stackoverflow.com/questions/45318066/getting-405-while-trying-to-download-nltk-dta)

Any comments would be appreciated.

alvations:
It seems like the Github is down/blocking access to the raw content on the repo. 

Meanwhile the temporary solution is something like this:

```
PATH_TO_NLTK_DATA=/home/username/nltk_data/
wget https://github.com/nltk/nltk_data/archive/gh-pages.zip
unzip gh-pages.zip
mv nltk_data-gh-pages/ $PATH_TO_NLTK_DATA
```

**Currently downloading the `gh-pages.zip` and replacing the `nltk_data` directory is the working solution for now.**

Before we find on another channel to distribute `nltk_data`, please use the above solution. 

----

~Strangely, it only seems to affect the `nltk` user account. It works fine on the fork: https://raw.githubusercontent.com/alvations/nltk_data/gh-pages/index.xml~

~Doing this would work too:~

matthew-z:
@alvations Thank you very much!

plaihonen:
Is there any alternative for the command line downloads such as this?
`python -m nltk.downloader -d ./nltk_data punkt`

rvause:
@plaihonen you should be able to use this alternative index by doing something like `python -m nltk.downloader -u https://pastebin.com/raw/D3TBY4Mj punkt`

plaihonen:
@rvause Works perfectly. Thank you!

owaaa:
+1. this was a several hour surprise this morning. Went with completely bypassing nltk download for now

ewan-klein:
[...]

PR #1795, opened 2017-07-27
Add environment variable option for DL_URL

zxiiro:
With GH-1787 we have come to realize that we need a method to locally
cache NLTK data so that we are not affected by outages caused by
upstream infrastructure. The proposal here is to allow an environment
variable NLTK_DOWNLOAD_URL to be set which will allow us to configure an
override on CI systems to force all of our projects that call
ntlk.downloader to download from a local copy of nltk data instead of
from upstream default URL.

Issue: GH-1787
Signed-off-by: Thanh Ha <thanh.ha@linuxfoundation.org>

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -2258,7 +2258,8 @@ if __name__ == '__main__':
     parser.add_option("-e", "--exit-on-error", dest="halt_on_error", action="store_true",
         default=False, help="exit if an error occurs")
     parser.add_option("-u", "--url", dest="server_index_url",
-        default=None, help="download server index url")
+        default=os.environ.get('NLTK_DOWNLOAD_URL'),
+        help="download server index url")
 
     (options, args) = parser.parse_args()
 


zxiiro:
Example steps to configure a local cache:

1) Download ntlk_data.zip
2) Run: `sed -i 's#https://raw.githubusercontent.com/nltk/nltk_data/gh-pages#http://localhost:8000#' /tmp/nltk_data/index.xml`
3) Setup webserver: python -m http.server

alvations:
This is a good idea to start the http server on localhost but maybe there's a better way than overwriting the `index.xml` manually.

zxiiro:
Indeed. This was just an example on one way to use the NLTK_DOWNLOAD_URL parameter and I think another patch to provide a way to rewrite the index.html should be created.

What is needed to get this particular patch merged?

alvations:
@zxiiro Agreed on another way to correct the `index.xml` file instead of asking users to do `sed -i`. Let's add another issue to keep track this. 

The code looks good to merge. But setting an additional environment should be tested either locally on at least Linux+Mac before merging. 

Is it possible to set up additional tox test just to test whether the `nltk_data` packages statically linked works?

zxiiro:
I am not familiar with nltk tests. How would I go about doing that?

alvations:
NLTK runs [`tox` test](http://tox.readthedocs.io/en/latest/example/general.html). 

The different environments for NLTK tests can be found on https://github.com/nltk/nltk/blob/develop/tox.ini 

But I'm not exactly sure how to test this out because it also involve changing how NLTK is downloaded at https://github.com/nltk/nltk/blob/develop/jenkins.sh#L6 

@fievelk would be a better person to advise on how to test the new `NLTK_DOWNLOAD_URL` that you've added.

zxiiro:
All we're doing here is adding an environment variable to an already existing function. If there was already a test for "-u" then it would be covered by the environment variable I think. Can we just get this patch merged and released?

I'd like to move forward with some ability to setup a backup nltk server for our CI systems soon.

alvations:
I think it looks okay to merge. 
But it's up to @stevenbird and/or @fievelk to merge it.

stevenbird:
@fievelk would you mind reviewing this please?

zxiiro:
Any chance someone can review this?

fievelk:
I'm sorry but I can't find enough time to accurately review this at the moment. I don't spot any major problem, but It would be good if someone else could take a look. 

Just some remarks: @zxiiro is there a reason why you're explicitly providing `None` as default fallback? It should be the default value for the `get` method. Besides that, out of curiosity, how are you calling `downloader.py`? Could you provide an example?

zxiiro:
@fievelk the original method had "default=None" all I did was add support to read an environment variable and continuing to default to None if it does not exist. So I believe no change in behaviour is made here other than the introduction of the NLTK_DOWNLOAD_URL environment variable which is optional.

I am running the code in this way:
```
export NLTK_DOWNLOAD_URL="http://localhost:8000/index.xml"
python -m nltk.downloader all
```

zxiiro:
In case it helps here is how you can test this without installing nltk and running directly from the repo:

Test case 1: WITHOUT NLTK_DOWNLOAD_URL configured
```
virtualenv /tmp/nltk
pip install -r pip-req.txt
python -m nltk.downloader.py all
```
In this case it should download from the default github url as configured by nltk.

Test case 2: WITH NTLK_DOWNLOAD_URL configured
```
virtualenv /tmp/nltk
pip install -r pip-req.txt
export NLTK_DOWNLOAD_URL="http://localhost:8000/index.xml"
python -m nltk.downloader all
```
In this case it should download from your locally configured NLTK_DOWNLOAD_URL. I setup a web server using python http.server locally and copied the data from ntlk_data github repo, unzipped it and modified index.html to replace the URLs to my localhost one.

Optional Test Case 3 can be done by setting up a NLTK_DOWNLOAD_URL pointing to an invalid URL. This will prove that it fails to download when it's not pointed to the right place that is configured by the environment variable.

nltk/downloader.py
```diff
@@ -2258,7 +2258,8 @@ def update():
     parser.add_option("-e", "--exit-on-error", dest="halt_on_error", action="store_true",
         default=False, help="exit if an error occurs")
     parser.add_option("-u", "--url", dest="server_index_url",
-        default=None, help="download server index url")
+        default=os.environ.get('NLTK_DOWNLOAD_URL', None),
```
fievelk: @zxiiro I mean that you are using `os.environ.get('NLTK_DOWNLOAD_URL', None)`, but I think that `None` is not needed when using `get()`, since it's already the default fallback value.

fievelk:
@zxiiro In general, the PR looks good to me. I would maybe change the `get()` call as I wrote in the review, but I agree that this PR can be useful.

zxiiro:
@fievelk I had not realized that. Okay I'll update the patch shortly.

zxiiro:
Is this ready to go? or do I need to get more reviews from other folks?

fievelk:
@stevenbird I think this can be safely merged.

stevenbird:
Thanks @fievelk! And thanks @zxiiro for your work (and your patience while we checked it out).

zxiiro:
Thanks everyone.

PR #1797, opened 2017-07-28
add no_escaping option

Henry-E:
Adding option from the perl version of moses tokenizer to not perform html escaping https://github.com/moses-smt/mosesdecoder/blob/master/scripts/tokenizer/tokenizer.perl#L83

### DIFF
--- a/nltk/tokenize/moses.py
+++ b/nltk/tokenize/moses.py
@@ -43,6 +43,12 @@ class MosesTokenizer(TokenizerI):
     >>> m = MosesTokenizer()
     >>> m.tokenize('2016, pp.')
     [u'2016', u',', u'pp', u'.']
+    
+    >>> sent = "This ain't funny. It's actually hillarious, yet double Ls. | [] < > [ ] & You're gonna shake it off? Don't?"
+    >>> m.tokenize(sent, escape=True)
+    ['This', 'ain', '&apos;t', 'funny', '.', 'It', '&apos;s', 'actually', 'hillarious', ',', 'yet', 'double', 'Ls', '.', '&#124;', '&#91;', '&#93;', '&lt;', '&gt;', '&#91;', '&#93;', '&amp;', 'You', '&apos;re', 'gonna', 'shake', 'it', 'off', '?', 'Don', '&apos;t', '?']
+    >>> m.tokenize(sent, escape=False)
+    ['This', 'ain', "'t", 'funny', '.', 'It', "'s", 'actually', 'hillarious', ',', 'yet', 'double', 'Ls', '.', '|', '[', ']', '<', '>', '[', ']', '&', 'You', "'re", 'gonna', 'shake', 'it', 'off', '?', 'Don', "'t", '?']
     """
 
     # Perl Unicode Properties character sets.
@@ -417,6 +423,11 @@ class MosesDetokenizer(TokenizerI):
     >>> detokens = d.detokenize(tokens)
     >>> " ".join(detokens) == expected_detokens
     True
+    
+    >>> d.detokenize(expected_tokens, unescape=True)
+    ['This', "ain't", 'funny.', "It's", 'actually', 'hillarious,', 'yet', 'double', 'Ls.', '|', '[]', '<', '>', '[]', '&', "You're", 'gonna', 'shake', 'it', 'off?', "Don't?"]
+    >>> d.detokenize(expected_tokens, unescape=False)
+    ['This', 'ain', '&apos;t', 'funny.', 'It', '&apos;s', 'actually', 'hillarious,', 'yet', 'double', 'Ls.', '&#124;', '&#91;', '&#93;', '&lt;', '&gt;', '&#91;', '&#93;', '&amp;', 'You', '&apos;re', 'gonna', 'shake', 'it', 'off?', 'Don', '&apos;t?']
     """
     # Currency Symbols.
     IsAlnum = text_type(''.join(perluniprops.chars('IsAlnum')))
@@ -618,6 +629,6 @@ class MosesDetokenizer(TokenizerI):
 
         return detokenized_text if return_str else detokenized_text.split()
 
-    def detokenize(self, tokens, return_str=False):
+    def detokenize(self, tokens, return_str=False, unescape=True):
         """ Duck-typing the abstract *tokenize()*."""
-        return self.tokenize(tokens, return_str)
+        return self.tokenize(tokens, return_str, unescape)
--- a/nltk/tokenize/moses.py
+++ b/nltk/tokenize/moses.py
@@ -236,7 +236,7 @@ class MosesTokenizer(TokenizerI):
                                 ESCAPE_LEFT_SQUARE_BRACKET,
                                 ESCAPE_RIGHT_SQUARE_BRACKET]
 
-    def __init__(self, lang='en', no_escaping=False):
+    def __init__(self, lang='en'):
         # Initialize the object.
         super(MosesTokenizer, self).__init__()
         self.lang = lang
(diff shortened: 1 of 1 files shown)


alvations:
Thank you for the PR. 

I think it's to avoid "double negative" in code, i.e. `escape=True` instead of `no_escape=False`. 
Anyone else has strong opinions on this? 

Also, if the tokenizer gets the parameter, the same parameter should be available on the detokenizer =)

nltk/tokenize/moses.py
```diff
@@ -236,7 +236,7 @@ class MosesTokenizer(TokenizerI):
                                 ESCAPE_LEFT_SQUARE_BRACKET,
                                 ESCAPE_RIGHT_SQUARE_BRACKET]
 
-    def __init__(self, lang='en'):
+    def __init__(self, lang='en', no_escaping=False):
```
alvations: `escape=True` here would be preferred to avoid "double negative" ;P

alvations: Actually, I think the `escape` parameter should go to the `tokenize()` function instead of the initialization.

nltk/tokenize/moses.py
```diff
@@ -245,6 +245,7 @@ def __init__(self, lang='en'):
         self.NUMERIC_ONLY_PREFIXES = [w.rpartition(' ')[0] for w in
                                       self.NONBREAKING_PREFIXES if
                                       self.has_numeric_only(w)]
+        self.no_escaping = no_escaping
```
alvations: If the parameter controlling the escaping is at the function then we wouldn't need the self.escape parameter too.

nltk/tokenize/moses.py
```diff
@@ -382,8 +383,9 @@ def tokenize(self, text, agressive_dash_splits=False, return_str=False):
         text = re.sub(regexp,substitution, text).strip()
         # Restore multidots.
         text = self.restore_multidots(text)
-        # Escape XML symbols.
-        text = self.escape_xml(text)
+        if not self.no_escaping:
```
alvations: If we were to allow the function instead of the class object to take the escape parameter, then here we would just need:
```
if escape:
    text = self.escape_xml(text)
```

alvations: It's also important for the Detokenizer to get the same escape parameter too.

alvations (changes requested):
First, let's avoid the "double negation" parameters ;P change `no_escaping` to `escape` or `escape_xml_chars` and set the default to True

Next, it's better to move the `escape` parameter to the `tokenize()` function instead of the initialization. 

Lastly, the same parameter is needed for the detokenize function in the MosesDetokenizer object.

Henry-E:
> I think it's to avoid "double negative" in code, i.e. escape=True instead of no_escape=False.
> Anyone else has strong opinions on this?

I was just following the example of the perl version.

> Also, if the tokenizer gets the parameter, the same parameter should be available on the detokenizer =)

I thought about this when submitting the pull request and checked the perl detokenizer but it's missing the no_escaping option

As for making it function option rather than a class option, I guess it doesn't matter too much. I would prefer to initialise the class in the first place with a no-escape option rather than having to reuse the option every time the function is called. But if that's the more standard way to do things in python I'm happy to have it go that way.

alvations:
I try to avoid perl bad habits when porting Moses code over ;P

Raymond Hettinger had a [nice video](https://www.youtube.com/watch?v=wf-BqAjZb8M) on how we often takes the bad habits from another non-Pythonic language when porting code over. (I make the same mistakes many times too)

Let's go with `escape` and check for True-tiness instead of `no_esape` and checking for False-ness. 

```
if escape:
    text = escape_xml(text)

```

vs

```
if not no_escape:
    text = escape_xml(text)
```
----

I think it's safe to assume that the default is to escape=True. At least for Moses SMT, it's tightly dependent on some XML-formatted file (aka SGM). 

I am seeing it from a user-perspective:

```
# Initializing one object and allow alternative function usage.
from nltk.tokenize.moses import MoseTokenizer

momo = MosesTokenizer()
momo.tokenize('some string; foo bar') # by defaul escape.
momo.tokenize('some string; foo bar', escape=False) # allow for the same object class to skip escaping
```

vs

```
# Initializing two objects and restrict function usage dependent on the object initialization.

from nltk.tokenize.moses import MoseTokenizer

momo_escape = MosesTokenizer(escape=True)
momo_escape.tokenize('some string; foo bar') 

momo_dont_escape = MosesTokenizer(escape=False)
momo_dont_escape.tokenize('some string; foo bar')
```

Anyone feel strongly for either? Both I fine by me, but I prefer the first instance.

Henry-E:
Thanks for all the recommendations. I made the necessary changes.

alvations:
@Henry-E Thanks for making the edits! 

One small additional request, can you add examples in the doctest for the 

 - MosesTokenizer's `tokenize()` with the `escape=True` and `escape=False` 
 - MosesDetokenizer's `detokenize()` with the `escape=True` and `escape=False` 

After that, we can do a last round of review =)

Henry-E:
Sorry I'm not familiar with the doctest. Which file do I need add examples to? The required formatting will be self evident from the other examples I'm guessing?

alvations:
Let's walk through the testing together =)

There are 2 parts to testing. 

 - Writing the test
 - Running the test


Writing the test
====

There are 2 tests that `nltk` relies on, [doctest](https://docs.python.org/3.5/library/doctest.html) and [unittest](https://docs.python.org/3.5/library/unittest.html). 

Walk through the tutorial when you're free but specific to this PR, all that's needed in writing the test is to add the doctest to the docstring you find at 

 - https://github.com/nltk/nltk/blob/develop/nltk/tokenize/moses.py#L21 for `MosesTokenizer`
 - https://github.com/nltk/nltk/blob/develop/nltk/tokenize/moses.py#L392 for `MosesDetokenizer`


For `MosesTokenizer`, simply at the end of the docstring add the

```python
>>> tokens_escape = t.tokenize(sent, escape=True)
>>> tokens_escape == expected_tokens
True
>>> t.tokenize(sent, escape=False)
[u'This', u'ain', u"'t", u'funny', u'.', u'It', u"'s", u'actually', u'hillarious', u',', u'yet', u'double', u'Ls', u'.', u'|', u'[', u']', u'<', u'>', u'[', u']', u'&', u'You', u"'re", u'gonna', u'shake', u'it', u'off', u'?', u'Don', u"'t", u'?']
```

And do the same for the `MosesDetokenizer` with respective input and output. Then when you do the PR it'll run the test for the new feature you've added.


Running the test locally
====

After you commit the code with the doctest you've written, if you had some error in the code/test, the automatic tests setup through Github will turned red and it will show **"All checks have failed"** instead of **"All checks have passed"**.

To run test locally:
====

First install `tox` on your machine, on Ubuntu/Debian/MacOSX use `pip install tox`.

Then go to the directory where you've the `nltk` develop branch saved. 

Lastly, you'll see the `tox.ini` file inside the directory and you see the different modes in square brackets `[mode]`, let's try to test `[py27-nodeps]` and `[py35-nodeps]`.

To test, use the `tox` command on the terminal:

```
tox -e py27
```

You will see the tests running. And at the end you should see an encouraging message: 

```
  py27: commands succeeded
  congratulations :)
```

Henry-E:
Wow, thanks so much for writing all that. I will try get it done as soon as I can.

alvations:
@Henry-E would you be able to write the tests for the new feature in the next couple of days? We're planning for a minor release and it'll be good for this feature to be added =)

Henry-E:
Hi @alvations sorry about the delay. Been travelling a lot lately and gotten bogged down with other work. I tried to spend a bit of time on it last week but I couldn't get the test package working locally. I'm still not entirely sure if I'm doing the tests correctly but I'll commit them and let me know how they look.

alvations:
@Henry-E Thanks for adding the tests! That's the correct way to add the doctest. 

Don't worry too much about the local testing. But in the future if you add more complex features, local testing would help you more, esp. when you don't want to wait for the automated checks =)

This looks good to merge now. Thank you again for the PR!

alvations (approved):

stevenbird:
Thanks @Henry-E!

Issue #1724, opened 2017-05-18
Better WordNet Reader


alvations:
This is a proposal to improve the WordNet interface in terms of code quality, functionalities and speed. Please feel free to add to this issues about current wordnet reader in `nltk` and how and what should be improved. 

-----

Suggestions / Issues
====

- Are there still issues with the [`_morphy()` function](https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L68)?  c.f. http://stackoverflow.com/questions/33594721/why-nltk-lemmatization-has-wrong-output-even-if-verb-exc-has-added-right-value

- What is the best params settings for `lowest_common_subsumers` in `wup_similarity`? See https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L899 

- Should the list of exceptions in the Princeton WordNet be extended? c.f. http://stackoverflow.com/questions/22999273/python-nltk-lemmatization-of-the-word-further-with-wordnet 

- [Add in the option to manually add a new root node; this will be useful for verb similarity as there exist multiple verb taxonomies](https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L1946)

- Allow direct offset + pos query, i.e. `wn.synset('dog.n.01') == wn.synset('2084071-n')`.

- Pre-compute similarities for all Synsets. 
  - We know that the Information Content (IC) based similarities scores should have been frozen by now since any new WordNet updates/extensions won't affect it so it's possible to precompute these and save them. That'll speed up the similarity matching quite drastically. 
  - As for the path related similarities, they would change with WordNet versions but its nice to still have them pre-computed respective to the WordNet versions from 3.0 onwards.

Code 
====

- Would `float('inf')` be a better approach than `_INF = 1e300` at https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L68 ? 

- Should encoding be a "set-able" parameter for the WordNetCorpusReader? https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L1052  

- Should imports be at the top of the script unless it's an `@abstractmethod` accessible from outside of the `CorpusReader` object? e.g.   
  - At `langs()` https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L1145
  - At `closure()`  https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L537

[...]

Issue #1030, opened 2015-06-29
Bug in WordNet wup_similarity for same-synset comparisons


davidjurgens:
The implementation of the Wu & Palmer similarity function WordNet seems to report an incorrect value when comparing a synset to itself.

``` python
from nltk.corpus import wordnet as wn
syn1 = wn.synset('bread.n.1')
syn1.wup_similarity(syn1) # yields 1.0 as expected
syn1 = wn.synset('dog.n.1')
syn1.wup_similarity(syn1) # yields 0.9285714285714286
```

For the second dog.n.1 comparison, [WordNet::Similarity](http://maraca.d.umn.edu/cgi-bin/similarity/similarity.cgi?word1=dog%23n%231&senses1=synset&word2=dog%23n%231&senses2=synset&measure=wup&rootnode=yes) gives the correct similarity of 1.0, as does [WS4J](http://ws4jdemo.appspot.com/?mode=w&s1=&w1=dog%23n%231&s2=&w2=dog%23n%231) (though the latter has its own bugs too)

I see the above results when using python 2.7.5,  NLTK 3.0.3, and WordNet 3.0

stevenbird:
Thanks @davidjurgens

alvations:
@stevenbird @davidjurgens I would think it's a feature and not a bug but i agree, it should be fixed. =)

This issue was also raised here: http://stackoverflow.com/questions/32333996/python-nltk-wup-similarity-score-not-unity-for-exact-same-word/32347253#32347253

The simple fix could have been at https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L861 but I'm not sure whether I can do the if statement to self. 

```
subsumer = self if self in subsumers else subsumers[0]
```

or maybe this is computationally cheaper at https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L855:

```
if self == other:
    subsumers = [self]
else:
    subsumers = self.lowest_common_hypernyms(other, simulate_root=simulate_root and need_root, use_min_depth=True)
```

PR #1813, opened 2017-08-15
Ensure self is its own lowest subsumer when needed

alvations:
Patching #1030 

Related to #1724 

Before the patch:

```python
>>> orange = wn.synsets('orange')[0]
>>> orange.lowest_common_hypernyms(orange, simulate_root=(False and orange._needs_root()), use_min_depth=True)
[Synset('fruit.n.01'), Synset('orange.n.01')]
>>> orange.wup_similarity(orange)
0.75
```

After the patch:

```python
>>> orange = wn.synsets('orange')[0]
>>> orange.lowest_common_hypernyms(orange, simulate_root=(False and orange._needs_root()), use_min_depth=True)
[Synset('fruit.n.01'), Synset('orange.n.01')]
>>> orange.wup_similarity(orange)
1.0
```

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -905,7 +905,7 @@ class Synset(_WordNetObject):
         if len(subsumers) == 0:
             return None
 
-        subsumer = subsumers[0]
+        subsumer = self if self in subsumers else subsumers[0]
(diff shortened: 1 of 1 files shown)


alvations:
[CI: retest]

PR #1816, opened 2017-08-18
data.py: Add sys.prefix in the paths for unix

AbdealiLoKo:
This is already present in Windows but not for Unix. Hence
Anaconda in Unix doesn't work easily.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -105,7 +105,9 @@ else:
         str('/usr/share/nltk_data'),
         str('/usr/local/share/nltk_data'),
         str('/usr/lib/nltk_data'),
-        str('/usr/local/lib/nltk_data')
+        str('/usr/local/lib/nltk_data'),
+        os.path.join(sys.prefix, str('nltk_data')),
+        os.path.join(sys.prefix, str('lib'), str('nltk_data'))
     ]
 
 


alvations:
@AbdealiJK thank you for proposing this PR. 

But adding the paths added in this PR might not be desirable, esp. if users are not using `nltk` inside a virtual environment, it might be better to advise users to optionally set the data path with `nltk.data.append()` if needed. 

In Unix, 

 - ` os.path.join(sys.prefix, str('nltk_data'))` will usually point to `/usr/nltk_data` 
   -  I don't think `nltk_data` should go into that `/usr/`, it might cause permission problems and might possibly allow malicious stuff

 - ` os.path.join(sys.prefix, str('lib'), str('nltk_data'))` will usually point to `/usr/lib/nltk_data`
   - which is already on [L107](https://github.com/AbdealiJK/nltk/blob/acbebb7dd06a1d146e26a63392fa271e5019329e/nltk/data.py#L107)

AbdealiLoKo:
@alvations consider the case when nltk is not installed to the system python but in an anaconda or pyenv python. 
In anaconda python the /usr is not sys.prefix

alvations:
But this should not affect where the `nltk_data` directory is stored. The `nltk` core code would be stored as Python packages under the virtual environment that conda creates but the `nltk_data` directory should be independent of the environment. 

Perhaps I'm mistaken. Maybe someone with more experience with Anaconda should take a look at this PR =)

AbdealiLoKo:
sys.prefix for anaconda is: `~/anaconda` (the anaconda location)
sys.prefix for system python is: `/usr`

When using anaconda you want all your files to be modularly kept inside the anaconda folder and not in the system folder. The anaconda folder mimics some of the paths available in the system root: `etc`, `lib`, `bin`, `share`, `var`, etc.

Hence, if i am using nltk in a anaconda environment, I'd want to keep my data files in one of the data folders given by anaconda so that data/configs in my anaconda are not affected by data/configs outside it.

stevenbird:
@AbdealiJK this is delayed, pending review by someone who has experience with Anaconda. Perhaps you can post a message to nltk-dev to request a review.

AbdealiLoKo:
CC @danielfrg - perhaps you'd like to review ?

danielfrg:
This is great! I had planned to do this PR but forgot. I have a conda recipe for a `nltk-with-data` conda package and it required exactly this change. The patch and recipe can be found here: https://github.com/danielfrg/conda-recipes/blob/master/nltk-with-data/data.py.patch

The only thing I would add is to add `os.path.join(sys.prefix, str('share'), str('nltk_data'))` to the list.

Other than that I can confirm that this works perfectly and allows you to download the nltk data to inside an anaconda install and keep everything self contained into those environments.

AbdealiLoKo:
I do not have any objection to add it as it would be the equivalent of `/usr/share/nltk_data` in the anaconda environment.
@stevenbird  @alvations - any opinions ?

EDIT: Note that share/nltk_data is not present i the windows section - so an anaconda on windows would behave differently from anaconda on linux. Either we add it to both or we don't add it because anaconda tries to be consistent between platforms so we should try to do the same within anaconda environments.

alvations:
Adding more paths to automatically search the `nltk_data` should be okay and I agree that it has to be consistent for both windows/mac/linux for conda. 

But I have one more question regarding how `nltk_data` are used by conda users. Is it common for conda users to put `nltk_data` with the site-packages? This not exactly related to this PR but possibly if conda users are installing `nltk_data` like packages, we should move forward with an optional "pip-able" data downloading system (related to #1787  and nltk/nltk_data#84)

AbdealiLoKo:
The reason a lot of conda users prefer all data, configs, code within the
environment is because they won't have write access to /usr and other
locations...


Now conda does have a local/ and share/ making it easy to save data there.
A pip or conda based data package would be great for some users
specifically those that have proxy constraints and so on.


That's where this discussion came from :
https://github.com/conda-forge/staged-recipes/issues/3615


Having a pypi package for data and hence a conda package for data would be
a +1 from me. Essentially data installable with package managers which are
trackable is a plus from an Operations standpoint.

danielfrg:
Yes like @AbdealiJK , its quite common for anaconda users to put everything into the anaconda installation or even just conda environments (basically the same as virtualenv but for conda). The main reason I ended up doing a conda pkg was to ship NLTK to spark jobs, and other type of jobs that required the nltk data to be present, this distributed jobs can be tricky.

That way I was able to have a single pkg with everything I needed for NLTK without having to actually download the nltk data in multiple nodes and manage that which can be a pain.

I would be +1 for a pip-able nltk data package similar to what spacy does.

stevenbird:
Yes, +1 for a pip-able data package!

stevenbird:
Thank you @AbdealiJK, @danielfrg, @alvations

AbdealiLoKo:
Any idea when the next release with this is planned?

PR #1826, opened 2017-09-06
Fix IBM model 1 documentation

ZiyaoWei:
IBM model 1 documentation seems to be getting the alignment wrong. This fixes that (assuming I didn't misunderstand how the model works).

### DIFF
--- a/nltk/translate/ibm1.py
+++ b/nltk/translate/ibm1.py
@@ -16,7 +16,9 @@
 Lexical translation model that ignores word order.
 
 In IBM Model 1, word order is ignored for simplicity. Thus, the
-following two alignments are equally likely.
+following three alignments are equally likely. As long as the word
+alignments are equivalent, it doesn't matter where the word
+occurs in the source or target sentence.
 
 Source: je mange du jambon
 Target: i eat some ham
@@ -24,7 +26,11 @@ Alignment: (1,1) (2,2) (3,3) (4,4)
 
 Source: je mange du jambon
 Target: some ham eat i
-Alignment: (1,4) (2,3) (3,2) (4,1)
+Alignment: (1,4) (2,3) (3,1) (4,2)
+
+Source: du jambon je mange
+Target: eat i some ham
+Alignment: (1,3) (2,4) (3,2) (4,1)
 
 The EM algorithm used in Model 1 is:
 E step - In the training data, count how many times a source language


alvations:
Thanks @ZiyaoWei for highlight the mistake in the IBM Model 1 algorithm documentation.

Just to validate that the fix is right. The original documentation states:

```
In IBM Model 1, word order is ignored for simplicity. Thus, the
following two alignments are equally likely.

Source: je mange du jambon
Target: i eat some ham
Alignment: (1,1) (2,2) (3,3) (4,4)

Source: je mange du jambon
Target: some ham eat i
Alignment: (1,4) (2,3) (3,2) (4,1)
```

By saying:

> In IBM Model 1, word order is ignored for simplicity. Thus, the
> following two alignments are equally likely.

It's trying to say that if we jumble up the word order given the same alignments of words, the overall probabilities of the alignments should be the same. In another words, as long as the word alignments are equivalent, it doesn't matter where the word occurs in the source or target sentence. 

We can walk through this step by step, in the first example:

```
Source: je mange du jambon
Target: i eat some ham
Alignment: (1,1) (2,2) (3,3) (4,4)
```

We are saying that the word alignments are:

```
je -> i
mange -> eat
du -> some
jambon -> ham
```

And thus the one to one alignment that are sequential (following the order of the source and target sentences).

The **fixed** second example in this PR corrected the alignments to `(1,4) (2,3) (3,1) (4,2)` which is right:

```
Source: je mange du jambon
Target: some ham eat i
Alignment: (1,4) (2,3) (3,1) (4,2)
```

We see that the alignment isn't sequential because the word order of the target sentence is jumbled up but IBM model 1 doesn't care for the lexical ordering, so it only considers 

 - `(1,4)` means the 1st word in source aligns to 4th word in target, i.e. 
   - je -> i
 - `(2,3)` means the 2nd word in source aligns to 3rd word in target, i.e. 
   - mange -> eat 
 - `(3,1)` means the 3nd word in source aligns to 2rd word in target, i.e. 
   - du -> ham
 - `(4,2)` means the 4nd word in source aligns to 2rd word in target, i.e. 
   - jambon -> some

And that yield the alignments which is the same as the first example:

```
je -> i
mange -> eat 
du -> some
jambon -> ham
```

The original example in the docstring would have yielded 

```
je -> i
mange -> eat
du -> ham 
jambon -> some
```

which is shouldn't give the same probability as the first example.

alvations (approved):

alvations:
It would be even clearer if you added another example and more explicit explanation ;P

```

In IBM Model 1, word order is ignored for simplicity. Thus, the
following three alignments are equally likely. As long as the word 
alignments are equivalent, it doesn't matter where the word 
occurs in the source or target sentence.

Source: je mange du jambon
Target: i eat some ham
Alignment: (1,1) (2,2) (3,3) (4,4)

Source: je mange du jambon
Target: some ham eat i
Alignment: (1,4) (2,3) (3,1) (4,2)

Source: du jambon je mange
Target: eat i some ham 
Alignment: (1,3) (2,4) (3,2) (4,1)

```

stevenbird:
@ZiyaoWei would you please follow @alvations recommendation

ZiyaoWei:
Oops - sure, will do (apologies to @alvations for dropping the ball)!

stevenbird:
Thanks @ZiyaoWei, @alvations

Issue #1836, opened 2017-09-24
Clean up the RTE classify code


alvations:
The `RTEFeatureExtractor` object in `nltk.classify.rte_classify.py` has functions outside the class object, it would be good clean up the code by: 

 - having RTE relation extracting  function as staticmethod within the object
 - porting the demo code into proper unit test in `nltk.test.unit`

alvations:
Patched in #1837

Issue #1776, opened 2017-07-12
Fixing LGTM alerts 


alvations:
#1775 raised a PR from [188 alerts from lgtm.com](https://lgtm.com/projects/g/nltk/nltk/alerts/)

Possibly, it'll be a good thing to look into each one of them.

PR #1837, opened 2017-09-25
Cleaning up RTE classification code

alvations:
Attending to #1836

- Moved `lemmatize` and `ne` tagging functions to a class function inside the `RTEFeatureExtractor` object

 - Moved the `demo*()` functions to proper unittest in `nltk/test/unit/test_rte_classify.py`.

I'm not sure how `megam` on CI but in the case that the testing fails, we can skip the `megam` and `bfgs` tests.


stevenbird:
@ewan-klein alerting you to this work on your code in case you can review

ewan-klein:
@alvations @stevenbird Not all of the affected code is mine. But the proposed changes look sensible and I don't foresee any problems with merging them in. I haven't tried to chase down the failing check, but I assume it is innocuous or fixable.

alvations:
@ewan-klein Thanks for looking through the proposed changes! The main purpose was so that the changes made to resolve the LGTM issues (from #1776) can be tracked with proper unittest by moving the demo code there.

The latest commit should resolve the string checking fails on the unittest.

stevenbird:
Thanks @alvations and @ewan-klein

PR #1840, opened 2017-09-27
Cleaned up NIST mean aggregation

alvations:
Cleaned up NIST mean aggregation


alvations:
The NLTK implementation is as close as it gets to the paper but it's not getting the same score as the `mteval-13a.pl`

```python
import io
from nltk.data import find

from nltk.translate.nist_score import corpus_nist

ref_file = find('models/wmt15_eval/ref.ru')
hyp_file = find('models/wmt15_eval/google.ru')
mteval_output_file = find('models/wmt15_eval/mteval-13a.output')


# Reads the BLEU scores from the `mteval-13a.output` file.
# The order of the list corresponds to the order of the ngrams.
with open(mteval_output_file, 'r') as mteval_fin:
    # The numbers are located in the last 2nd line of the file.
    # The first and 2nd item in the list are the score and system names.
    mteval_nist_scores = list(map(float, mteval_fin.readlines()[-4].split()[1:-1]))

with io.open(ref_file, 'r', encoding='utf8') as ref_fin:
    with io.open(hyp_file, 'r', encoding='utf8') as hyp_fin:
        # Whitespace tokenize the file.
        # Note: split() automatically strip().
        hypothesis = list(map(lambda x: x.split(), hyp_fin))
        # Note that the corpus_bleu input is list of list of references.
        references = list(map(lambda x: [x.split()], ref_fin))
        # Without smoothing.
        for i, mteval_nist in zip(range(1,6), mteval_nist_scores):
            nltk_nist = corpus_nist(references, hypothesis, n=i)
            print(nltk_nist, mteval_nist)
```

[out]

```
0.0003245040705132866 4.4488
0.0006275925534519339 5.0043
0.0009551587456855594 5.052
0.0012806946888239113 5.0564
0.0016773154748957778 5.0564
```

nltk/translate/nist_score.py
```diff
@@ -118,17 +115,15 @@ def corpus_nist(list_of_references, hypotheses, n=5):
...
-
     # Eqn 2 in Doddington (2002):
     # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
-    info = [0 if p_n[i].numerator == 0 or p_n[i+1].numerator == 0 # Handles math domain and zero division errors.
-            else math.log(p_n[i].numerator / p_n[i+1].numerator)
-            for i in range(len(p_n)-1)]
-    return sum(info_i/sysoutput_lengths[i] for i, info_i in enumerate(info)) * bp
+    info = [p_numerators[i] / p_numerators[i+1] for i in range(1, len(p_numerators))]
+    # Eqn 3 in Doddington (2002)
+    score = sum(info_i/p_denominators[i] for i, info_i in enumerate(info, start=1)) * bp
```
alvations: From the paper, in Eqn (3), I'm assuming the `sum(l)` (for all w_1 ... w_n) in sysoutput) to be the hypothesis outputs. But when we do the division, the number of n-grams goes to a really low number. 

The `Info(w_1...w_n)` computation looks correct though.

alvations: Note that in MT eval, they took the max sysoutput_length instead of the sum: https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L888

nltk/translate/nist_score.py
```diff
@@ -118,17 +115,12 @@ def corpus_nist(list_of_references, hypotheses, n=5):
...
     # Eqn 2 in Doddington (2002):
     # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
-    info = [0 if p_n[i].numerator == 0 or p_n[i+1].numerator == 0 # Handles math domain and zero division errors.
-            else math.log(p_n[i].numerator / p_n[i+1].numerator)
-            for i in range(len(p_n)-1)]
-    return sum(info_i/sysoutput_lengths[i] for i, info_i in enumerate(info)) * bp
-
+    info = [math.log(p_numerators[i] / p_numerators[i+1]) for i in range(1, len(p_numerators))]
+    # Eqn 3 in Doddington (2002)
+    score = sum(info_i/p_denominators[i] for i, info_i in enumerate(info, start=1)) * bp
```
alvations: After verifying with the Perl version, our final score function for Eqn 3 is the same as https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L617

nltk/translate/nist_score.py
```diff
@@ -118,17 +115,12 @@ def corpus_nist(list_of_references, hypotheses, n=5):
...
-           for i, _ in enumerate(range(1,n+1))]
-
     # Eqn 2 in Doddington (2002):
     # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
-    info = [0 if p_n[i].numerator == 0 or p_n[i+1].numerator == 0 # Handles math domain and zero division errors.
-            else math.log(p_n[i].numerator / p_n[i+1].numerator)
-            for i in range(len(p_n)-1)]
-    return sum(info_i/sysoutput_lengths[i] for i, info_i in enumerate(info)) * bp
-
+    info = [math.log(p_numerators[i] / p_numerators[i+1]) for i in range(1, len(p_numerators))]
```
alvations: The info calculation is different from the perl version, we're reporting:

```
[0.6501985972926496, 0.5767564872468458, 0.5903376866937243, 0.5538851132264376]
```

But the perl version is reporting:

```
[8888.76188473492 1054.22451254869 85.7306840538572 7.58496250072116]
```

Hmmm... 🤔

alvations: Yes, it looks like our implementation of Eqn 2 is different from MT eval.

It's something a little more terse there:
https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L796

alvations: From http://doras.dcu.ie/484/1/owczarzak_karolina_2008.pdf page 17:

> Here the precision is modified by the information weights Info(w1 . . . wn), computed on the basis of the frequency of a given n-gram in relation to the frequency of a ‘parent’ (n-1)-gram, in accordance with the equation (2.6)

alvations:
Now this is interesting, may how the information is computed is though  local information micro-averaged per ngram instead of how BLEU computes macro-averaged globally across all ngrams:

```python
import io, math
from collections import Counter, defaultdict
from nltk.data import find
from nltk import ngrams

from nltk.tokenize.nist import NISTTokenizer

ntok = NISTTokenizer()

def ngram_overlaps(references, hypothesis, n):
    counts = Counter(ngrams(hypothesis, n)) if len(hypothesis) >= n else Counter()
    # Extract a union of references' counts.
    ## max_counts = reduce(or_, [Counter(ngrams(ref, n)) for ref in references])
    max_counts = {}
    for reference in references:
        reference_counts = Counter(ngrams(reference, n)) if len(reference) >= n else Counter()
        for ngram in counts:
            max_counts[ngram] = max(max_counts.get(ngram, 0),
                                    reference_counts[ngram])
    # Assigns the intersection between hypothesis and references' counts.
    clipped_counts = {ngram: min(count, max_counts[ngram])
                      for ngram, count in counts.items()}
    return clipped_counts

ref_file = find('models/wmt15_eval/ref.ru')
hyp_file = find('models/wmt15_eval/google.ru')
mteval_output_file = find('models/wmt15_eval/mteval-13a.output')

with io.open(ref_file, 'r', encoding='utf8') as ref_fin:
    with io.open(hyp_file, 'r', encoding='utf8') as hyp_fin:
        # Whitespace tokenize the file.
        # Note: split() automatically strip().
        hypotheses = list(map(lambda x: ntok.tokenize(x), hyp_fin))
        # Note that the corpus_bleu input is list of list of references.
        references = list(map(lambda x: [ntok.tokenize(x)], ref_fin))

overlap_grams = defaultdict(Counter)

for reference, hypothesis in zip(references, hypotheses):
    for i in range(1,6):
        overlap_grams[i] += ngram_overlaps(reference, hypothesis, i)

for i in range(2,6):
    info = sum([overlap_grams[i-1][ng[:-1]]/float(count) 
                for ng, count in overlap_grams[i].items()])
    print(str(i-1)+'-gram info. weights =', info)
```

[out]:

```
1-gram info. weights = 6839.0666666666675
2-gram info. weights = 411.0
3-gram info. weights = 178.0
4-gram info. weights = 100.0
```

----
Using the above information weights computation, the final NIST score compared to the Perl version:

```python
import io, math
from collections import Counter, defaultdict
from nltk.data import find
from nltk import ngrams

from nltk.translate.nist_score import corpus_nist
from nltk.tokenize.nist import NISTTokenizer

ntok = NISTTokenizer()

ref_file = find('models/wmt15_eval/ref.ru')
hyp_file = find('models/wmt15_eval/google.ru')
mteval_output_file = find('models/wmt15_eval/mteval-13a.output')

# Reads the BLEU scores from the `mteval-13a.output` file.
# The order of the list corresponds to the order of the ngrams.
with open(mteval_output_file, 'r') as mteval_fin:
    # The numbers are located in the last 2nd line of the file.
    # The first and 2nd item in the list are the score and system names.
    mteval_nist_scores = list(map(float, mteval_fin.readlines()[-4].split()[1:-1]))

with io.open(ref_file, 'r', encoding='utf8') as ref_fin:
    with io.open(hyp_file, 'r', encoding='utf8') as hyp_fin:
        # Whitespace tokenize the file.
        # Note: split() automatically strip().
        hypotheses = list(map(lambda x: ntok.tokenize(x), hyp_fin))
        # Note that the corpus_bleu input is list of list of references.
        references = list(map(lambda x: ntok.tokenize(x), ref_fin))
        # Without smoothing.
        for i, mteval_nist in zip(range(1,5), mteval_nist_scores):
            nltk_nist = corpus_nist(references, hypotheses, n=i)
            print(nltk_nist, mteval_nist)
```

[out]:

```
3.41327246949777 4.4488
3.629255070894102 5.0043
3.728023595302803 5.052
3.786796784067901 5.0564
```

It gets close to what mteval-v13a.pl is reporting but still pretty far from the numbers that the Perl version gives. The trend is the same though. 

But note that there's no `math.log` anywhere in the information computation which is different from the paper =(

It'll be great if someone who is sure verify this.

jnothman:
If the Perl implementation has a permissive licence, one solution is try to port the Perl from scratch then integrate it into the current NLTK implementation.

jnothman:
Any idea what the licence is?

alvations:
The mteval-v13.pl script doesn't have a specified license and has passed through many hands so it's unclear who really "owns" the code other than https://www.nist.gov/multimodal-information-group/metrics-machine-translation-evaluation 

Amazon ported the BLEU and tokenization part of the perl script to `sockeye` and gave it an Apache license: https://github.com/awslabs/sockeye/blob/master/contrib/sacrebleu/sacrebleu.py

I've been trying to breakdown the code and port from scratch but it'll take a little more time and if someone could help to explain what the NIST is exactly doing (local or global information weights computation) that'll make it a little easier =) https://gist.github.com/alvations/9f98f0dc56d04da198edf0ca15ff4470 

P/S: We have the tokenizer part reimplemented from scratch and the BLEU part modified from legacy NLTK to match mteval-v13.pl and the NIST part is the unclear one.

jnothman:
So I suppose the issue is in `score_segment`'s construction of `tst_info` and `ref_info`?

alvations:
Yes, `tst_info` and `ref_info` computes the information weights and score_segment does the aggregation with arithmetic mean. 

On the paper, it says it's summing log weights. Maybe I've missed it but I can't seem to find the log in the perl code.

jnothman:
Is `compute_ngram_info` what you're missing? It's called on all the ref texts before starting, and populates the global `%ngram_info`. It calculates the total ngram frequency, across the corpus, then for each ngram calculates the `-log2(freq[ngram] / freq[ngram[:-1]])` for non-unigrams, and `-log2(freq[ngram] / N)` where `ngram[:-1]` is the ngram missing its final word.

In `score_segment` it sums these scores for each ref ngram to make ref_info, then for each test ngram (capping a repeated ngram's contributions to `ref_ngrams_max` for that ngram) for tst_info.

`tst_count` and `ref_count` took me a long time to work out. They're simply counting how many n-grams of a particular length exist in each corpus, which can be determined analytically from the number of words and the number of responses in it. :\

alvations:
@jnothman Thank you for the explanation! It's getting clearer now. 

For `-log2(freq[ngram] / N)`, does `N` refer to the ngram order?  

As for sentences with multiple reference sentences, does it all count towards the information weights or only the closest reference sentences is computed?

alvations:
@jnothman I've modified the code to the explanation you've made. 

I think I might not be getting it right. The scores are further away from the perl implementation =(

```python
import io, math
from collections import Counter, defaultdict
from nltk.data import find
from nltk import ngrams

from nltk.translate.nist_score import corpus_nist
from nltk.tokenize.nist import NISTTokenizer

ntok = NISTTokenizer()

ref_file = find('models/wmt15_eval/ref.ru')
hyp_file = find('models/wmt15_eval/google.ru')
mteval_output_file = find('models/wmt15_eval/mteval-13a.output')

# Reads the BLEU scores from the `mteval-13a.output` file.
# The order of the list corresponds to the order of the ngrams.
with open(mteval_output_file, 'r') as mteval_fin:
    # The numbers are located in the last 2nd line of the file.
    # The first and 2nd item in the list are the score and system names.
    mteval_nist_scores = list(map(float, mteval_fin.readlines()[-4].split()[1:-1]))

with io.open(ref_file, 'r', encoding='utf8') as ref_fin:
    with io.open(hyp_file, 'r', encoding='utf8') as hyp_fin:
        # Whitespace tokenize the file.
        # Note: split() automatically strip().
        hypotheses = list(map(lambda x: ntok.tokenize(x), hyp_fin))
        # Note that the corpus_bleu input is list of list of references.
        references = list(map(lambda x: ntok.tokenize(x), ref_fin))
        # Without smoothing.
        for i, mteval_nist in zip(range(1,5), mteval_nist_scores):
            nltk_nist = corpus_nist(references, hypotheses, n=i)
            print(nltk_nist, mteval_nist)
```

[out]:

```
-0.9090950588099418 4.4488
-2.069103318789428 5.0043
-2.0678637475694184 5.052
-2.0678637475694184 5.0564

```

jnothman:
Sorry by N I mean the number of words. These are maximum likelihood estimates of the probability of seeing a word given its preceding (n-1)gram context.

jnothman:
You say you've modified the code, but I don't see any new commit pushed

alvations:
Sorry pushed it up to the wrong branch. 
It's here now =)

nltk/translate/nist_score.py
```diff
@@ -93,34 +93,47 @@ def corpus_nist(list_of_references, hypotheses, n=5):
...
+            # Counter of ngrams in hypothesis.
+            hyp_ngrams = Counter(ngrams(hypothesis, i)) if len(hypothesis) >= i else Counter()
+            # Adds the no. of ngrams in the hypothesis.
+            sysoutput_lengths[i] += len(hyp_ngrams)
+            # Compute the information weights per overlapped ngram.
+            for ng in hyp_ngrams:
+                # Eqn 2 in Doddington (2002):
+                # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
+                numerator = ngram_freq[ng]
+                denominator = ngram_freq[ng[:-1]] if i > 1 else len(ng)
```
jnothman: Yes, the len(ng) is incorrect. Sorry for misleading you

alvations: The latest code using `hyp_len` instead of `len(ng)` outputs this when comparing against perl outputs:

```
-0.35709729625648523 4.4488
-0.9651077936825132 5.0043
-0.9638682224625039 5.052
-0.9638682224625039 5.0564
```

alvations:
The strange part is that with the `-log2()`, we're in a different space for the information weights:

```
[-1265.4593805409081, -1265.4593805409081, 3.3890422907458992, 0]
```

When I print out the `tst_info` from https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L796 , it's returning:

```
[8888.76188473492 1054.22451254869 85.7306840538572 7.58496250072116]
```

jnothman:
No, you don't want `hyp_len`. Stop and think about what this is calculating. It's trying to calculate the log probability of each ngram in the hypothesis under a simple MLE (i.e. relative frequencies from a reference corpus) markov language model (setting the log probability of an unseen ngram to 0, which is a bit weird in the world of language models). The reference corpus could be some external corpus (Google Web1T or books ngrams or Gigaword, perhaps), but in the NIST scorer it is just the concatenation of all the reference texts provided for evaluation.

These probabilities are and must be static for a given reference corpus, and the hypothesis text only affects which probabilities are multiplied (or summed in log space) to get the overall measure. The fact that these are static means that NIST just calculated all these log probabilities, for every ngram, before even looking at the hypothesis, avoiding a little bit of repeated work.

So for a unigram the probability is the relative frequency of that unigram in the entire reference corpus. The denominator is the sum of the lengths of all reference texts. It's that simple.

nltk/translate/nist_score.py
```diff
@@ -93,41 +93,46 @@ def corpus_nist(list_of_references, hypotheses, n=5):
...
+            # Compute the information weights per overlapped ngram.
+            for ng in hyp_ngrams:
+                # Eqn 2 in Doddington (2002):
+                # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
+                numerator = ngram_freq[ng]
+                denominator = ngram_freq[ng[:-1]] if i > 1 else hyp_len
+                if numerator == 0 or denominator == 0:
+                    information_weights[i] += 0
+                else:
+                    information_weights[i] += -1 * math.log(numerator/ denominator, 2)
```
jnothman: this should have `-hyp_ngrams[ng]` instead of -1. In fact, not `-hyp_ngrams[ng]` but `-min(hyp_ngrams[ng], max(number of times ng occurs in reference_text for reference_text in reference_texts))`

alvations: Doubling checking my understanding.

Since we're going to need ` max(number of times ng occurs in reference_text for reference_text in reference_texts))`, we can pre-compute these before going through the references while counting the ngrams in the reference texts, i.e. `max_ngram_freq`:

```python
# Compute the information weights based on the reference sentences.
ngram_freq = Counter()
max_ngram_freq = Counter()
for references in list_of_references: # For each source sent, there's a list of reference sents.
   for reference in references:
       # For each order of ngram, count the ngram occurrences.
       for i, _ in enumerate(range(1,n+1)):
           ng_i = ngrams(reference, i)
           ngram_freq.update(ng_i)
           for ng, count in Counter(ng_i).items():
               if count > max_ngram_freq[ng]:
                   max_ngram_freq[ng] = count
```

And afterwards, the "backoff" denominator for the unigram probability under the MLE would be:

```python
if i > 1:
    denominator = ngram_freq[ng[:-1]] 
else:
    denominator = min(hyp_ngrams[ng], max_ngram_freq[ng])
```

Is that right?

Something like:

```
p(abc | ab) = freq(abc) / f(ab)  # Ignoring f(ab) == 0, just return 0 

p(a | null) = freq(a) / min(a, max_ngram_freq(a))  
```

where max_ngram_freq is `max(number of times ng occurs in reference_text for reference_text in reference_texts))`?

alvations: And to compute the `information_weights[i]`, it would be:

```python
if numerator == 0 or denominator == 0:
    information_weights[i] += 0
else:
    information_weights[i] += numerator/ denominator
```

Then the `-log2()` can be done at the final return sum?

```python
return sum(-math.log2(info_i)/sysoutput_lengths[i] for i, info_i in information_weights.items()) * bp
```

jnothman:
you certainly cannot do the log at the end. A sum of logs is almost never
equal to a log of sums. And the 0 there has to be in log space: taking a
log of 0 will get you negative infinity

alvations:
So, the log has to be done at each ngram check?

And it looks something like this?

```python
def corpus_nist(list_of_references, hypotheses, n=5):
    """
    Calculate a single corpus-level NIST score (aka. system-level BLEU) for all
    the hypotheses and their respective references.

    :param references: a corpus of lists of reference sentences, w.r.t. hypotheses
    :type references: list(list(list(str)))
    :param hypotheses: a list of hypothesis sentences
    :type hypotheses: list(list(str))
    :param n: highest n-gram order
    :type n: int
    """
    # Before proceeding to compute NIST, perform sanity checks.
    assert len(list_of_references) == len(hypotheses), "The number of hypotheses and their reference(s) should be the same"

    # Compute the information weights based on the reference sentences.
    ngram_freq = Counter()
    max_ngram_freq = Counter()
    for references in list_of_references: # For each source sent, there's a list of reference sents.
        for reference in references:
            # For each order of ngram, count the ngram occurrences.
            for i, _ in enumerate(range(1,n+1)):
                ng_i = ngrams(reference, i)
                ngram_freq.update(ng_i)
                for ng, count in Counter(ng_i).items():
                    if count > max_ngram_freq[ng]:
                        max_ngram_freq[ng] = count


    hyp_lengths, ref_lengths = 0, 0
    sysoutput_lengths = Counter()    # Key = ngram order, and value = no. of ngram in hyp.
    information_weights = Counter()  # Key = ngram order, and value = sum of info weights per nth order.
    # Iterate through each hypothesis and their corresponding references.
    for references, hypothesis in zip(list_of_references, hypotheses):
        # Calculate the hypothesis length and the closest reference length.
        # Adds them to the corpus-level hypothesis and reference counts.
        hyp_len =  len(hypothesis)
        hyp_lengths += hyp_len
        ref_lengths += closest_ref_length(references, hyp_len)

        # For each order of ngram.
        for i in range(1,n+1):
            # Counter of ngrams in hypothesis.
            hyp_ngrams = Counter(ngrams(hypothesis, i)) if len(hypothesis) >= i else Counter()
            # Adds the no. of ngrams in the hypothesis.
            sysoutput_lengths[i] += len(hyp_ngrams)
            # Compute the information weights per overlapped ngram.
            for ng in hyp_ngrams:
                # Eqn 2 in Doddington (2002):
                # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
                numerator = ngram_freq[ng]
                denominator = ngram_freq[ng[:-1]] if i > 1 else min(hyp_ngrams[ng], max_ngram_freq[ng])
                if numerator == 0 or denominator == 0:
                    information_weights[i] += 0
                else:
                    information_weights[i] += math.log(numerator/denominator, 2)

    # Calculate corpus-level brevity penalty.
    bp = nist_length_penalty(ref_lengths, hyp_lengths)

    return sum(info_i/sysoutput_lengths[i] for i, info_i in information_weights.items()) * bp
```

alvations:
@jnothman looking at Eqn 2 from the Doddington (2002) paper:

> Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]

Should it be the numerator that is `ngram_freq[ng[:-1]]` instead of the denominator?

jnothman:
log(a/b) = -log(b/a), so whichever. I've not looked at your whole
implementation and probably can't for  next while.

alvations:
Thanks @jnothman for all the tips! No worries on reading the full code =)

I'll waddle a little while longer on it.

jnothman:
I think the max_ngram_freq should not be calculated per sentence... should
check if it should be calculated over the document or the corpus...

jnothman:
How is it going, @alvations?

alvations:
I'm still struggling with the NLTK implementation, I'm not sure where the problem is and I think reimplementing it might be easier.

But I'm moving forward to completing the walkthrough and hopefully get better insight. 
  https://www.kaggle.com/alvations/rantings-on-mteval

alvations:
At last, sneaked out some free time to clean up our implementation of NIST score. 

Here's the results from the original `mteval-v13a.pl` vs the current implementation.

```
Original:  [4.4488, 5.0043, 5.0520, 5.0564]

Ours:      [4.2879, 4.8274, 4.8704, 4.8767]
```

Now it's much much like the original NIST, only difference is that we're taking macro-average across of sum of precision per reference, instead of the original NIST that takes the absolute count and compute the micro-averaged precision. 

The choice to choose macro-average over micro is because I'm not sure of what's the strategy when there's multiple references. I think the original perl script chooses the hypothesis with the closest length but in the NLTK implementation, it maximizes the sentence level NIST score. 

Thanks @jnothman for the tips previously! It helped a lot to cool down re-read the paper and literally reimplement bits and pieces of the perl script and then come back to NLTK implementation =)

jnothman:
so are you able to change that averaging and replicate that reference
result, regardless of the justification for the averaging approach?

alvations:
Yes, it looks like so. I've managed to replicated the result with macro average instead of micro.  And the micro vs marcro average has minor differences.

nltk/translate/nist_score.py
```diff
@@ -93,44 +93,59 @@ def corpus_nist(list_of_references, hypotheses, n=5):
...
+    information_weights = {}
+    for _ngram in ngram_freq: # w_1 ... w_n
+        _mgram = _ngram[:-1] #  w_1 ... w_n-1
+        if _mgram and _mgram in ngram_freq:
+            denominator = ngram_freq[_mgram]
+        else:
+            denominator = total_reference_words
+        # Eqn 2 in Doddington (2002):
+        # Info(w_1 ... w_n) = log_2 [ (# of occurrences of w_1 ... w_n-1) / (# of occurrences of w_1 ... w_n) ]
+        information_weights[_ngram] = -1 * math.log(ngram_freq[_ngram]/denominator) / math.log(2)
```
alvations: Mathematically, this is the same formula as Eqn2 but I've no idea why did the script https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl#L546 

Is it because of preventing underflow from log of really low fraction value? 

The line could have been this and yield the same result:

```
information_weights[_ngram] = math.log(denominator/ngram_freq[_ngram], 2)
```

jnothman: Does Perl's log function allow you to specify the base?

jnothman: Oh you're asking about the -, not the base. I am not sure about the relative numerical stability, but negating should be cheap

alvations: Don't think so, everything in Perl is usually base 10. If I'm wrong, I had to do something like http://perldoc.perl.org/functions/log.html

alvations:
Now this is very interesting!

```
# Original output from mteval-v13a

Original:  [4.4488, 5.0043, 5.0520, 5.0564]

# Macro-precision across NIST scores per reference

Ours:      [4.2879, 4.8274, 4.8704, 4.8767]

# Micro-precision, summing precision's numerator and denominators 
# across all references before one single score computation at the end

Ours:      [4.2232, 4.7779, 4.8256, 4.8301]
```

Code is a little dirty now with long variable names but the difference should be clear =)

nltk/translate/nist_score.py
alvations: I think we can stick with the version closer to paper's interpretation without the perl's limitation =)

alvations:
After some thoughts, the NIST score was never meant for macro-averaging because that'll bring the precision denominator out of the global sum and create cumulative sum in the inner loop across each reference-hypothesis pair. 

Lets keep only the micro-average version =)

Also, just found the last "bug" that we had to consider repeated ngrams. Now, we've (at last):

```
Original:  [4.4488, 5.0043, 5.0520, 5.0564]
Ours: [4.4496, 5.0051, 5.0527, 5.0572]
```

 Thanks @jnothman for the tips again!

jnothman:
yes, that looks like a much more believable reproduction result. Well done!

The lack of support for base arg to log is hardly a limitation. Internally
numpy is very likely to just log and divide.

alvations:
@jnothman regarding license. I think it's unlicensed but ownership belongs originally to Kishore Papineni (BLEU author) and I think it's maintained by NIST. 

If they do license this under some non-permissive license (at any point), then in this case, we need to fight for it to remain permissive. Because (i) it's an evaluation script that a huge chunk of literature relies on, (ii) the Perl script has became a monster of adding on more stuff with a handful of people knowing what's inside.

alvations:
[CI: retest]

jnothman:
It should be public domain in the USA and permissive elsewhere as per
http://www.nist.gov/director/licensing

stevenbird:
Thanks @alvations and @jnothman

Issue #1839, opened 2017-09-26
Deprecating the old Stanford Parser


alvations:
We have deprecated the `StanfordTokenizer`/`StanfordSegmenter`, `StanfordPOSTagger` and `StanfordNERTagger`. 

It would be good to also deprecate the old `StanfordParser`, `StanfordDependencyParser` and `StanfordNeuralDependencyParser` by 

1. Adding the appropriate warnings to the old interface

2a. Wrap the duck-types for `CoreNLPParser` that emulates the functions of the old interface

2b.  Write up documentations of how to use the `CoreNLPParser` to use dependency and neural dependency parsing

3. Write tests for the new CoreNLP parser interfaces

Both (2a) and (2b) methods should only affect the `properties` argument of the `api_call`

The current interface for `CoreNLPParser`:

```
>>> from nltk.parse.corenlp import CoreNLPParser
>>> sent = 'The quick brown fox jumps over the lazy dog.'
>>> next(parser.raw_parse(sent)).pretty_print()  # doctest: +NORMALIZE_WHITESPACE
                         ROOT
                          |
                          S
           _______________|__________________________
          |                         VP               |
          |                _________|___             |
          |               |             PP           |
          |               |     ________|___         |
          NP              |    |            NP       |
      ____|__________     |    |     _______|____    |
     DT   JJ    JJ   NN  VBZ   IN   DT      JJ   NN  .
     |    |     |    |    |    |    |       |    |   |
    The quick brown fox jumps over the     lazy dog  .
```

The desired interface might look like this:

```
# Using Duck-types
>>> from nltk.parse.stanford import CoreNLPParser
>>> depparser = CoreNLPDependencyParser('http://localhost:9000')
>>> depparser.parse(sent)
>>> ndepparser = CoreNLPNeuralDependencyParser('http://localhost:9000')
>>> ndepparser.parse(sent)
```

```
# Using arguments to control `properties` for `api_call()` 
>>> from nltk.parse.stanford import CoreNLPParser

>>> depparser = CoreNLPParser('http://localhost:9000', parser_type="dependency")
>>> depparser.parse(sent)

>>> ndepparser = CoreNLPNeuralDependencyParser('http://localhost:9000', parser_type="neural_dependency")
>>> ndepparser.parse(sent)
```

This would make a good class project or good first challenge ;P

artiemq:
[...]

PR #1843, opened 2017-10-02
Deprecating the old Stanford Parser

artiemq:
Worked on #1839 issue

Also refactored CoreNLPTokenizer and CoreNLPTagger tests


nltk/parse/stanford.py
```diff
@@ -281,6 +282,15 @@ class StanfordParser(GenericStanfordParser):
 
     _OUTPUT_FORMAT = 'penn'
 
+    def __init__():
+        warnings.simplefilter('always', DeprecationWarning)
```
dimazest: I'm not sure whether `warnings.simplefilter` should be called here. It should be up to the user to decide what warnings are shown.

nltk/parse/stanford.py
```diff
@@ -281,6 +282,15 @@ class StanfordParser(GenericStanfordParser):
...
+        warnings.warn(str("\nThe StanfordParser will "
+                          "be deprecated\n"
+                          "Please use \033[91mnltk.parse.corenlp.CoreNLPParser\033[0m instead.'"),
+                      DeprecationWarning, stacklevel=2)
+        warnings.simplefilter('ignore', DeprecationWarning)
```
dimazest: Also here, what if the user had some custom filter.

nltk/parse/stanford.py
```diff
@@ -23,6 +23,7 @@
 from nltk.parse.api import ParserI
 from nltk.parse.dependencygraph import DependencyGraph
 from nltk.tree import Tree
+from nltk.parse.corenlp import CoreNLPParser, CoreNLPDependencyParser
```
dimazest: I would do this:
```python
from nltk.parse import corenlp
```

and later refer to `CoreNLPParser` and `CoreNLPDependencyParser` as `corenlp.CoreNLPParser` and `corenlp.CoreNLPDependencyParser`.

nltk/parse/stanford.py
```diff
@@ -400,6 +426,76 @@ def _make_tree(self, result):
         return DependencyGraph(result, top_relation_label='ROOT')
 
 
+class CoreNLPParser(CoreNLPParser):
```
dimazest: With a new import this becomes:
```python
class CoreNLPParser(corenlp.CoreNLPParser):
```

I suggest not to use `CoreNLPParser` as a name, as it's confusing.

nltk/parse/stanford.py
```diff
@@ -228,6 +228,7 @@ def _execute(self, cmd, input_, verbose=False):
 
         return stdout
 
+
 class StanfordParser(GenericStanfordParser):
```
dimazest, line 232: Actually, why not to rewrite the implementation of `StanfordParser` so that it uses `CoreNLP parser`, but then one needs to be careful with the CoreNLPServer.

nltk/parse/stanford.py
dimazest: is `str("...")` really needed? would `"..."` be enough?

dimazest: It looks like there is an unmatched `'`.

dimazest (changes requested):
I'm not sure it worth creating a drop-in replacement for the Stanford parser, mostly because the API is incompatible: CoreNLPParser requires CoreNLPServer.

Deprecation warnings are useful!

artiemq:
[StanfordTagger](https://github.com/nltk/nltk/blob/develop/nltk/tag/stanford.py#L130) and [StanfordTokenizer](https://github.com/nltk/nltk/blob/develop/nltk/tokenize/stanford.py#L110)  have been deprecated in the same way. If you think it really not worth to replace the `StanfordParser` then should i rewrite tagger and tokenizer too?

dimazest:
Right, these are a tagger and a tokenizer, not a parser. I would deprecate the Stanford parser and suggest to use corenlp.CorreNLPParser.

@alvations what do you think?

alvations:
Deprecating the `StanfordParser` and leaving the old interface there while advising users to use `CoreNLPParser` is a good idea. That way the code on the users' side don't break if they decide to still use the old Stanford tools without CoreNLP.

But we have to ensure that it behaves as how a `ParserI` would function; it should since the inheritance tree is =)

```
> ParserI, TokenizerI
    > GenericCoreNLPParser
        > CoreNLPParser
        > CoreNLPDependencyParser
```

Maybe something like this:

```
/nltk
    /parse
        /corenlp
            GenericCoreNLPParser                            
            CoreNLPParser                                        
            CoreNLPDependencyParser                     
            CoreNLPNeuralDependencyParser 
        /stanford
            GenericStanfordParser                           
            StanfordParser                                    
            StanfordDependencyParser               
            StanfordNeuralDependencyParser       
     /tag
        /stanford
            StanfordTagger               
            StanfordPOSTagger        
            StanfordNERTagger        
            CoreNLPTagger               
            CoreNLPPOSTagger        
            CoreNLPNERTagger        
     /tokenize
        /stanford
            StanfordTokenizer           
            CoreNLPTokenizer          
        /stanford_segmenter
            StanfordSegmenter        
```

dimazest:
Another way is to make CoreNLPParser to implement all those interfaces and get rid of CoreNLPTokenizer and CoreNLP*Tagger. The reasoning being is that we provide an interface to a tool which is a tagger, parser and tokenizer.

artiemq:
What should i do with different return types? The original `CoreNLPParser.tokenize()` returns a generator of string, but `StanfordTokenizer.tokenize()` returns a list, should i create something like `CoreNLPParser.stanford_tokenize()` that returns a list or just leave tokenizer part as is?

Also `StanfordParser` and `CoreNLPParser` have different method name for parsing. `StanfordParser` has `parse()` but `CoreNLPParser` has `raw_parse()`.

dimazest:
@artiemq a list and a generator are both iterables, so I would not pay attention to the difference. Personally, I prefer a generator, because then it up to the user to decide what container to use (a list, a set or none and just iterate over the generator). I would not add `.stanford_tokenize()` as it will clutter the already quite mysterious API.

`.raw_parse()` is defined only on StanfordParser, and consequently in CoreNLP, because in the beginning it was meant to be a drop in replacement. Since CoreNLP is not a drop in replacement, it worth removing it and follow the interface defined in https://github.com/nltk/nltk/blob/develop/nltk/parse/api.py#L14

It seems that the difference between `.raw_parse()` and `.parse()` is that `.raw_parse()` takes a sentence as a string, while `.parse()` expects a list of tokens.

nltk/parse/corenlp.py
```diff
@@ -176,15 +177,18 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     """Interface to the CoreNLP Parser."""
 
-    def __init__(self, url='http://localhost:9000', encoding='utf8'):
+    def __init__(self, url='http://localhost:9000', encoding='utf8', tagtype=None):
         import requests
 
         self.url = url
         self.encoding = encoding
 
+        assert tagtype in ['pos', 'ner', None]
```
dimazest: Throw an exception here saying that `tagtype` must be either `'pos'`, `'ner'` or `None`.

nltk/parse/corenlp.py
```diff
@@ -339,6 +343,30 @@ def tokenize(self, text, properties=None):
...
+        default_properties = {'ssplit.isOneSentence': 'true',
+                              'annotators': 'tokenize,ssplit,' }
+        # Supports only 'pos' or 'ner' tags.
+        assert self.tagtype in ['pos', 'ner']
+        default_properties['annotators'] += self.tagtype
+        for sentence in sentences:
+            tagged_data = self.api_call(sentence, properties=default_properties)
+            assert len(tagged_data['sentences']) == 1
+            # Taggers only need to return 1-best sentence.
+            yield [(token['word'], token[self.tagtype]) for token in tagged_data['sentences'][0]['tokens']]
```
dimazest: But they can return several, according to the API (I guess so). I would not limit the functionality here and if CoreNLP returns several parses, pass all of them.

nltk/parse/stanford.py
```diff
@@ -23,9 +23,11 @@
 from nltk.parse.api import ParserI
 from nltk.parse.dependencygraph import DependencyGraph
 from nltk.tree import Tree
+from nltk.parse import corenlp
```
dimazest: This looks redundant.

nltk/corpus/reader/plaintext.py
```diff
@@ -14,7 +14,7 @@
 from six import string_types
 import codecs
 
-import nltk.data
+from nltk.data import LazyLoader
```
dimazest: It might be a good idea to leave this code as it was.

nltk/downloader.py
```diff
@@ -181,6 +181,7 @@
 from six.moves.urllib.error import HTTPError, URLError
 
 import nltk
+from nltk import data, internals
```
dimazest: Same here, it might not worth to play with imports

artiemq:
I've updated `raw_tag_sents()` method, now it returns list of parses for each sentence, but `tag()` still returns `list(tuple(str, str))` with only one parse, is it fine?

dimazest:
Looks good, thanks!

dimazest (approved):

alvations:
Lets get this merged and prepare for #1892 =)

stevenbird:
Thanks @artiemq, @dimazest, @alvations

hexingren:
Hello,

I was trying to move to CoreNLPPOSTagger and CoreNLPNERTagger from StanfordNERTagger with NLTK v3.2.5. The CoreNLPPOSTagger worked as expected but CoreNLPNERTagger threw an HTTPError: 500 Server Error. I tried both CoreNLP v3.9.1 (the latest version) and v3.8.0, and they both threw the same error. I posted more details at [#2010](https://github.com/nltk/nltk/issues/2010)

Does anyone have some idea on this problem? Or I should be safe if I stay with StanfordNERTagger using v3.2.5? Thanks.

Issue #1838, opened 2017-09-25
A weird edge case for bleu scoring. 


benleetownsend:
There is an unintuitive functionality of the bleu_score module. 

```
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,8,1], weights=(0.5, 0.5)) 
0.5
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,1,2], weights=(0.5, 0.5))
0.408248290463863
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,2,3], weights=(0, 0.5))
0.5773502691896257
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,7,3], weights=(0, 0.5))
1.0
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,8,7], weights=(0, 0.5))
0

```
The default `SmoothingFunction.method0` skips any fractions that are equal to zero making bleu scores in these cases quite unintuitive. In my opinion this would be more intuitive to have the log(0) evaluate to -inf resulting in the overall bleu score in these cases evaluating to 0. This along with a clearer warning would make this much more intuitive. 

To clarify, as seen below it gives a higher bleu for a worse match when i believe it should receive a bleu score of 0.
```
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,8,1], weights=(0.5, 0.5))
0.5
>>> bleu_score.sentence_bleu([[1,2,3,4]],[5,6,1,2], weights=(0.5, 0.5))
0.408248290463863
```

alvations:
Just a clarification before diving in. Do you get a warning when you try these fringe cases?

benleetownsend:
Yes, I receive the warning.

alvations:
Good. Let's go through the rational of why the outputs are as such:

Firstly, note that un-smoothed BLEU has never been meant for sentence level evaluation, it is advisable to find the optimal smoothing method that your data require, please look at https://github.com/alvations/nltk/blob/develop/nltk/translate/bleu_score.py#L425

If user wants to skip the step to find the optimal smoothing, let's look at the examples you've listed.

First scenario, when your n-gram order is larger than the overlap that exists between the reference and hypothesis, corpus/sentence BLEU will not work in this case. So the results of this cannot be taken into account:

```
>>> from nltk.translate.bleu_score import sentence_bleu
[...]

PR #1844, opened 2017-10-02
Hotfix/bug in bleu score

bamattsson:
Fixing the bug discussed in #1838

I opted for using `p_n_new.append(sys.float_info.min)` instead of `return [sys.float_info.min]` which also would have worked. I made this decision since it makes the code cleaner and more easy to understand. The computational bottle-neck will never be in this part of the code anyway so optimising it to avoid one or two extra passes through this loop is irrelevant. Especially since it made the code a bit cryptic.

### DIFF
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -122,8 +122,10 @@ class TestBLEU(unittest.TestCase):
     def test_partial_matches_hypothesis_longer_than_reference(self):
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary who loves Mike'.split()
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.4729, places=4)
-        # Checks that the warning has been raised because len(reference) < 4.
+        # Since no 4-grams matches were found the result should be zero
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
+        # Checks that the warning has been raised.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
         except AttributeError:
@@ -138,8 +140,10 @@ class TestBLEUFringeCases(unittest.TestCase):
         hypothesis = 'John loves Mary'.split()
         n = len(hypothesis) + 1 #
         weights = [1.0/n] * n # Uniform weights.
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.7165, places=4)
-        # Checks that the warning has been raised because len(hypothesis) < 4.
+        # Since no n-grams matches were found the result should be zero
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
+        # Checks that the warning has been raised.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
         except AttributeError:
@@ -149,7 +153,9 @@ class TestBLEUFringeCases(unittest.TestCase):
         # it's a special case where reference == hypothesis.
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary'.split()
-        assert(sentence_bleu(references, hypothesis, weights) == 1.0)
+        # Since no 4-grams matches were found the result should be zero
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
 
     def test_empty_hypothesis(self):
         # Test case where there's hypothesis is empty.
@@ -174,8 +180,9 @@ class TestBLEUFringeCases(unittest.TestCase):
         # is shorter than 4.
         references = ['let it go'.split()]
         hypothesis = 'let go it'.split()
-        # Checks that the value the hypothesis and reference returns is 1.0
-        assert(sentence_bleu(references, hypothesis) == 1.0)
+        # Checks that the value the hypothesis and reference returns is 0.0
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
         # Checks that the warning has been raised.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -125,12 +125,13 @@ class TestBLEU(unittest.TestCase):
         self.assertAlmostEqual(sentence_bleu(references, hypothesis, backoff=True), 0.4729, places=4)
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, enable_multibleu=True), 0.0, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, emulate_multibleu=True), 0.0, places=4)
         # Checks that the warning has been raised because len(reference) < 4.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
         except AttributeError:
             pass # unittest.TestCase.assertWarns is only supported in Python >= 3.2.
+
 
 #@unittest.skip("Skipping fringe cases for BLEU.")
 class TestBLEUFringeCases(unittest.TestCase):
@@ -144,7 +145,7 @@ class TestBLEUFringeCases(unittest.TestCase):
         self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, backoff=True), 0.7165, places=4)
         # Since no n-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, enable_multibleu=True), 0.0, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, emulate_multibleu=True), 0.0, places=4)
         # Checks that the warning has been raised because len(hypothesis) < 4.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
@@ -158,7 +159,7 @@ class TestBLEUFringeCases(unittest.TestCase):
         assert(sentence_bleu(references, hypothesis, weights, backoff=True) == 1.0)
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, enable_multibleu=True), 0.0, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, emulate_multibleu=True), 0.0, places=4)
 
     def test_empty_hypothesis(self):
         # Test case where there's hypothesis is empty.
@@ -187,7 +188,7 @@ class TestBLEUFringeCases(unittest.TestCase):
         assert(sentence_bleu(references, hypothesis, backoff=True) == 1.0)
         # Checks that the value the hypothesis and reference returns is 0.0
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, enable_multibleu=True), 0.0, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, emulate_multibleu=True), 0.0, places=4)
         # Checks that the warning has been raised.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -228,20 +228,17 @@ class TestBLEUvsMteval13a(unittest.TestCase):
                                             smoothing_function=chencherry.method3)
                     assert abs(mteval_bleu - nltk_bleu) < 0.005
 
-class TestEmulateMultiBLEU(unittest.TestCase):
+class TestBLEUWithBadSentence(unittest.TestCase):
     def test_corpus_bleu_with_bad_sentence(self):
         hyp = "Teo S yb , oe uNb , R , T t , , t Tue Ar saln S , , 5istsi l , 5oe R ulO sae oR R"
         ref = str("Their tasks include changing a pump on the faulty stokehold ."
                   "Likewise , two species that are very similar in morphology "
                   "were distinguished using genetics .")
         references = [[ref.split()]]
-        hypothese = [hyp.split()]
+        hypotheses = [hyp.split()]
         try: # Check that the warning is raised since no. of 2-grams < 0.
             with self.assertWarns(UserWarning):
                 # Verify that the BLEU output is undesired since no. of 2-grams < 0.
-                self.assertAlmostEqual(corpus_bleu(references, hypothese), 0.4309, places=4)
-        except AttributeError:
-            pass # unittest.TestCase.assertWarns is only supported in Python >= 3.2.
-        desired_output = corpus_bleu(references, hypothese)
-        #assert
-        assert desired_output == 0.0
+                self.assertAlmostEqual(corpus_bleu(references, hypotheses), 0.0, places=4)
+        except AttributeError: # unittest.TestCase.assertWarns is only supported in Python >= 3.2.
+            self.assertAlmostEqual(corpus_bleu(references, hypotheses), 0.0, places=4)
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -122,6 +122,9 @@ class TestBLEU(unittest.TestCase):
     def test_partial_matches_hypothesis_longer_than_reference(self):
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary who loves Mike'.split()
+        # Test with 3-gram BLEU scores
+        weights = [1.0/3] * 3  # Uniform weights
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.36840, places=4)
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
         self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
@@ -156,6 +159,11 @@ class TestBLEUFringeCases(unittest.TestCase):
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
         self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
+        # Checks that the warning has been raised because len(hypothesis)
+        try:
+            self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
+        except AttributeError:
+            pass # unittest.TestCase.assertWarns is only supported in Python >= 3.2.
 
     def test_empty_hypothesis(self):
         # Test case where there's hypothesis is empty.
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -125,7 +125,7 @@ class TestBLEU(unittest.TestCase):
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
         self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
-        # Checks that the warning has been raised.
+        # Checks that the warning has been raised because len(reference) < 4.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
         except AttributeError:
@@ -143,7 +143,7 @@ class TestBLEUFringeCases(unittest.TestCase):
         # Since no n-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
         self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
-        # Checks that the warning has been raised.
+        # Checks that the warning has been raised because len(hypothesis) < 4.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
         except AttributeError:
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -122,12 +122,10 @@ class TestBLEU(unittest.TestCase):
     def test_partial_matches_hypothesis_longer_than_reference(self):
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary who loves Mike'.split()
-        # Test with 3-gram BLEU scores
-        weights = [1.0/3] * 3  # Uniform weights
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.36840, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, backoff=True), 0.4729, places=4)
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, enable_multibleu=True), 0.0, places=4)
         # Checks that the warning has been raised because len(reference) < 4.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
@@ -143,9 +141,10 @@ class TestBLEUFringeCases(unittest.TestCase):
         hypothesis = 'John loves Mary'.split()
         n = len(hypothesis) + 1 #
         weights = [1.0/n] * n # Uniform weights.
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, backoff=True), 0.7165, places=4)
         # Since no n-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, enable_multibleu=True), 0.0, places=4)
         # Checks that the warning has been raised because len(hypothesis) < 4.
         try:
             self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
@@ -156,14 +155,10 @@ class TestBLEUFringeCases(unittest.TestCase):
         # it's a special case where reference == hypothesis.
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary'.split()
+        assert(sentence_bleu(references, hypothesis, weights, backoff=True) == 1.0)
         # Since no 4-grams matches were found the result should be zero
         # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
-        # Checks that the warning has been raised because len(hypothesis)
-        try:
-            self.assertWarns(UserWarning, sentence_bleu, references, hypothesis)
-        except AttributeError:
-            pass # unittest.TestCase.assertWarns is only supported in Python >= 3.2.
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights, enable_multibleu=True), 0.0, places=4)
 
     def test_empty_hypothesis(self):
         # Test case where there's hypothesis is empty.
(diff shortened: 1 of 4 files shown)


nltk/test/unit/translate/test_bleu.py
```diff
@@ -122,7 +122,9 @@ def test_full_matches(self):
     def test_partial_matches_hypothesis_longer_than_reference(self):
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary who loves Mike'.split()
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.4729, places=4)
+        # Since no 4-grams matches were found the result should be zero
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
```
alvations, line 127: This makes little sense to give 0 credits to this hypothesis given the references. 

The sentence is half correct, saying that it's 0.0 BLEU is really not sound at all.

alvations (dismissed):

nltk/test/unit/translate/test_bleu.py
```diff
@@ -138,7 +140,9 @@ def test_case_where_n_is_bigger_than_hypothesis_length(self):
         hypothesis = 'John loves Mary'.split()
         n = len(hypothesis) + 1 #
         weights = [1.0/n] * n # Uniform weights.
-        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.7165, places=4)
+        # Since no n-grams matches were found the result should be zero
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
```
alvations, line 146: This example only has a punctuation difference, it should not be evaluated as 0 BLEU too =(

nltk/test/unit/translate/test_bleu.py
```diff
@@ -149,7 +153,9 @@ def test_case_where_n_is_bigger_than_hypothesis_length(self):
         # it's a special case where reference == hypothesis.
         references = ['John loves Mary'.split()]
         hypothesis = 'John loves Mary'.split()
-        assert(sentence_bleu(references, hypothesis, weights) == 1.0)
+        # Since no 4-grams matches were found the result should be zero
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis, weights), 0.0, places=4)
```
alvations, line 159: Please read the hypothesis and reference carefully and think about the values you're changing in the unittest.  The hypothesis is a **perfect translation**.

nltk/test/unit/translate/test_bleu.py
```diff
@@ -174,8 +180,9 @@ def test_reference_or_hypothesis_shorter_than_fourgrams(self):
         # is shorter than 4.
         references = ['let it go'.split()]
         hypothesis = 'let go it'.split()
-        # Checks that the value the hypothesis and reference returns is 1.0
-        assert(sentence_bleu(references, hypothesis) == 1.0)
+        # Checks that the value the hypothesis and reference returns is 0.0
+        # exp(w_1 * 1 * w_2 * 1 * w_3 * 1 * w_4 * -inf) = 0
+        self.assertAlmostEqual(sentence_bleu(references, hypothesis), 0.0, places=4)
```
alvations, line 186: This is the **only** test case that your hot fix should be solving. 

I agree that the original 1.0 is not perfect but changing it to 0.0 isn't acceptable too =(

nltk/translate/bleu_score.py
```diff
@@ -481,16 +481,14 @@ def method0(self, p_n, *args, **kwargs):
...
                 _msg = str("\nCorpus/Sentence contains 0 counts of {}-gram overlaps.\n"
-                           "BLEU scores might be undesirable; "
-                           "use SmoothingFunction().").format(i+1)
+                           "BLEU score will be zero; consider "
+                           "using SmoothingFunction().").format(i+1)
                 warnings.warn(_msg)
-                # If this order of n-gram returns 0 counts, the higher order
-                # n-gram would also return 0, thus breaking the loop here.
-                break
+                p_n_new.append(sys.float_info.min)
```
alvations: Scores **shouldn't** be zero if they are perfect or almost perfect match. 
See the above comments for the detailed review.

We're in a glass is half empty, half full situation and more often than not in machine translation, it's half full. Otherwise the `emulate_multibleu` option should kick in and it should reflect the canonical `multi-bleu.pl`

Please rework the PR.

alvations:
@bamattsson Thank you for the PR attempt. 

I think it's safe to say that the unit test cases are generally sound for [`TestBLEU`](https://github.com/nltk/nltk/blob/develop/nltk/test/unit/translate/test_bleu.py#L15) and they shouldn't be changed unless there's a very good reason or at least the outputs shouldn't be changed that drastically from 1.0 -> 0.0 or 0.7 -> 0.0.

Focus the PR on proposing the fix to the BLEU hotfix such that it keeps and passes the test on `TestBLEU` but most probably, only the value unit test cases for [test_reference_or_hypothesis_shorter_than_fourgrams()](https://github.com/nltk/nltk/blob/develop/nltk/test/unit/translate/test_bleu.py#L172) should be changed. 

Similarly for the doctest, the values on https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L37 shouldn't change more than 2 decimal places. These are example sentences from the original BLEU paper: http://www.aclweb.org/anthology/P02-1040.pdf 

Please rework the PR.

nltk/translate/bleu_score.py
alvations: I think reading #1330 and all the attached issues to that thread would better clarify the scope of BLEU.

bamattsson:
@alvations 

1. If you follow the equations for calculating BLEU that are presented in http://www.aclweb.org/anthology/P02-1040.pdf and http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf to calculate the BLEU scores for the tests they are not what you have in the tests currently. That is why I changed some of them.
In the 2nd example with `'John loves Mary ?'` and `'John loves Mary'` we get that p_4 = 0 (since we have no 4-gram overlaps). If p_4=0 we will (according to both equations that you refer to) get a BLEU-4 score of 0. I strongly urge you to go through this with pen and paper by hand, and see how the math works out. And yes, log(0) should in this case evaluate to -inf, and not simply be ignored.

2. The 2nd doctest (I missed it sorry) should also evaluate to BLEU-4=0 due to the same reason. There is no 3- or 4-grams overlap, and thus, p_3=0 and p_4=0. Please read through the paper where you've taken the examples from, nowhere does it state that the BLEU-4 value of that hypothesis given those references should be equal to  what you have in the doctest. I can change this.

3. There are two edge case though. One is when both the reference and hypothesis are `'let it go'`. In this case there are no 4-grams overlap, but there are also no 4-grams at all, so here p_4 = m_4/l_4 = 0/0 which is undefined. And here it might very well make more sense to let p_4 = 1 than to let p_4 = 0. And BLEU-4 would in this case evaluate to 1. The same edge case happens in the first test. Sorry for overlooking this. I can change this.

4. Yes, I had a look at the conversations in #1330 before making this PR, and saw that you had done a big overhaul of the BLEU score. However, that does not change the fact that the code at it's current state is incorrect.

5. Here's one example that might be a bit easier to understand if the math is tricky. Try to run:
```
nltk.translate.bleu_score.sentence_bleu([['one', 'example', 'here', '.']], ['one', 'example', 'here', '.'])
```
and
```
nltk.translate.bleu_score.sentence_bleu([['one', 'example', 'here', '.']], ['.', 'here', 'example', 'one'])
```
both gives BLEU-4 = 1. That's not good. One is a perfect translation, whereas the other is the sequence reversed, and thus a very poor translation.

alvations:
I understand that if there's no 4grams the decision in this PR is to return closer to zero due to the multiplication of the very small numbers. And in fact enable_multibleu did exactly that but in a different way.  



We'll post this PR in a highlight to the dev mailing list to get more input to review this PR.

alvations:
I'm agnostic as to how we should handle #1554 and the proposal here is one option. 

If the majority of NLTK devs agrees to go ahead with this issue, my suggestion is to either 

 - Go ahead with the changes made to the test examples in `TestBLEU`. 
    - Please ensure that warnings are placed properly and advise users to do smoothing.
    - Think of something to check for perfect translations given no. of 4-grams < 1, most probably a `ref == hyp` condition
    - Make smoothing method other than method0 as the default case so that we avoid users' gotcha moments and reduce maintenance of answering users' question. 

 - Allow an NLTK BLEU vs Multi-bleu vs Original BLEU option 
   - Make it clear that there are many many BLEU versions out in the wild and original BLEU is original but might not suit users' data / use-case. (like what we did with @ExplodingCabbage's porter stemmers)

If the decision is to merge this PR, please ensure that we are available to handle questions on the mailing list, the github issues and stackoverflow on this.

bamattsson:
@alvations 

Okey that sounds reasonable. How to handle it is up to you, and those involved in NLTK.

- I will make changes to fix 1a and 1b in this PR. So that you can choose this option if you'd like. I can try to write more in the docs to make what happens clearer.
- 2a is also an option, but in that case you should not in the docs write that you (if method0 is used) are following the equations in http://www.aclweb.org/anthology/P02-1040.pdf and http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf. Because as it's currently implemented, you are not.

alvations:
Lets wait for others to vote on this and if we had a no votes. Perhaps we'll reach out to ask @rsennrich and @mjpost or marian/moses devs.

Meanwhile, please go ahead and fix the issues to be consistent with the PR you're suggesting:

 - Please ensure that warnings are placed properly and advise users to do smoothing.
 - Think of something to check for perfect translations given no. of 4-grams < 1, most probably a ref == hyp condition or if we keep the 0 BLEU, then make the
 

If you would like to, please also consider the interaction of the `enable_multibleu` parameter, maybe it's no longer necessary. Or perhaps we need another parameter to `enable_backoff` or something.

bamattsson:
@alvations 

- Alright so now I fixed the docstrings to make it clearer for the user.
- I tried to make the 0/0 case evaluate to = 1, and thus return a higher value when the example is identical strings but of shorter lengths than the n-grams we are considering. However it was not possible and would have required a deeper refactoring. The usage of `Fraction` and how it is handled on this [line](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L309) makes that impossible. I want to alert you that this leads to another bug on this [line](https://github.com/nltk/nltk/blob/develop/nltk/translate/bleu_score.py#L164), the denominator is here incremented by one even if the hypothesis was shorter than the corresponding n-gram.
- Do note though that the 0/0 edge case is very uncommon. Whereas the 0/den where den != 0 is common.

alvations:
If you would like to, remove all the usage of Fraction, use this as a reference implementation: https://github.com/mjpost/sacreBLEU and change the whole nltk.translate.bleu module and make it look at mteval-v13.pl 

Perhaps that's easier than changing the current code.

Just make sure that the unittest and doctest are edited appropriately.

nltk/translate/bleu_score.py
```diff
@@ -479,18 +490,17 @@ def method0(self, p_n, *args, **kwargs):
...
-                _msg = str("\nCorpus/Sentence contains 0 counts of {}-gram overlaps.\n"
-                           "BLEU scores might be undesirable; "
-                           "use SmoothingFunction().").format(i+1)
+            if p_i.numerator == 0:
+                # We have 0/den where den might be 0 or !=0. The result should equal 0, or undefined
+                # However, math.log(0) gives math error. Therefore we replace 0 with sys.float_info.min
+                p_n_new.append(sys.float_info.min)
+                _msg = str("\nThe hypothesis contains 0 counts of {}-gram overlaps. Therefore\n"
+                           "the BLEU score evaluates to 0, independently of how many N-gram\n"
+                           "overlaps of lower order it contains. Consider using BLEU score\n"
```
alvations: Lowering order of n-grams will require users to get into a wave of other problems that concerns weights hacking. It think it's better to ask users to simply smooth their n-gram precision.

alvations:
@bamattsson I've added more documentation to the method0 as well as the motivation and enforce  the `p_n_new.append(sys.float_info.min)` when `emulate_multibleu` is kicked in. That should follow closer to the BLEU implementations (`multi-bleu.pl` and `mteval-v13a.pl`) out there. With the updated code:

```python
from functools import partial

from nltk.translate.bleu_score import sentence_bleu

# User defined their favorite variant of BLEU with functools.partial once, e.g.
mybleu = partial(sentence_bleu, emulate_multibleu=True)
mybleu([['one', 'example', 'here', '.']], ['.', 'here', 'example', 'one'])
mybleu([['one', 'example', 'here', '.']], ['.', 'here', 'john', 'one'])

# User define their variant of BLEU each time they use:
sentence_bleu([['one', 'example', 'here', '.']], ['.', 'here', 'example', 'one'], emulate_bleu=True)
sentence_bleu([['one', 'example', 'here', '.']], ['.', 'here', 'john', 'one'], emulate_bleu=True)
```

That should enable the users to choose whether they want to use the `backoff` strategy or the traditional BLEU to 0 strategy when handling fringe cases. In any case, users that asks how they should handle it should be advised to use method1 to method6 from `SmoothingFunction()` whenever possible.

I've also changed the unittest so that it should pass our CI server and get this PR merge after you've okay the changes made. 

Thank you for your contribution to NLTK and the insightful discussion on BLEU!

alvations (dismissed):
Agreed. Strategy to handle fringe cases should be closer to how `mteval-v13a.pl` is appending the -inf instead of directly returning a -inf precision score. 

But keeping the `emulate_multibleu` as default False because we previous had many complains about these wack-a-mole issues when just returning 0.0 whenever something goes out of BLEU scope. 

See the many related issues linked to #1330

nltk/translate/bleu_score.py
```diff
@@ -140,10 +140,14 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     :type weights: list(float)
     :param smoothing_function:
     :type smoothing_function: SmoothingFunction
-    :param auto_reweigh:
+    :param auto_reweigh: Option to re-normalize the weights uniformly.
     :type auto_reweigh: bool
-    :param emulate_multibleu: bool
+    :param emulate_multibleu: Option to emulate multi-bleu.pl behavior
```
bamattsson: Can you add here. "If set to `True` the original equation proposed by K Papineni et. al". 

That is much clearer than "Option to emulate multi-bleu.pl behavior", which might be useful for you who have read through their code, but is not useful for external users who haven't

nltk/translate/bleu_score.py
```diff
@@ -140,10 +140,14 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
...
+    :type emulate_multibleu: bool
+    :param backoff: Option to backoff to the largest n-gram overlap if there's less than one 4-gram overlap.
```
bamattsson: Can you change this to "if set to `True` the precision for n-gram overlap with no overlap will evaluate to 1".

bamattsson (commented):
Hey,

Sorry for dropping out for a while. I've been busy with a lot of other things.

To be honest I think this is a very weird way of solving the problem. It is like designing a physics simulation software, and after having some users complaining about that you cannot make things go faster than the speed of light, as a solution you decide remove the that boundary. Even though you know that it breaks all fundamental maths etc.

Two proposals:
1. I've been busy lately, but will have more time in the beginning of November. If you want I can pick this up then and spend more time on it. I can change it so that two identical sentences shorter than n evaluates to BLEU=1.0.
2. If you decide that it is better to get this merged now I would suggest the changes I've proposed here. And that you add the following to the docstring (following line 35): "As default (when `backoff=True`) uses a modified version of the original BLEU score where the precision for n-gram overlap with no overlap will evaluate to 1. If `emulate_multibleu=True` the original equation proposed by K Papineni et. al is used, and the precision for n-gram overlap with no overlap will evaluate to 0.

nltk/translate/bleu_score.py
alvations: Actually, it's not exactly the original equation proposed by Papineni et al. because on paper it didn't really account for the case where no. of 4-grams overlap is < 0. There's no resolution until mteval.pl was released openly.

But I understand the docstring for the parameter needs to be more obvious. I'll make the changes.

nltk/translate/bleu_score.py
alvations: The backoff doesn't always evaluate 1.0, (i) it evaluates to 1.0 only if the unigrams matches totally and there's no bigram matches or (ii) when the largest order of n-th ngram overlap is 1.0 and n < 4. 

Actually, we could also swap it to be `emulate_multibleu=True` and `backoff=False` by default but I'm afraid we have to handle the complaints we had before using the backoff strategy. 

I would really like to hear more people's view on which to set as default because there wasn't any response when the `emulate_multibleu` was introduced in #1554.

alvations:
No worries about coming in to contribute on this later in November but lets get this merged first so that users have a closer implementation of BLEU like in `multi-bleu.pl` =)

I think it's a good idea to enforce a real smoothing method whenever the `sentence_bleu` is used but it would be better if other users/devs give us feedbacks on how and what they would like to see as the default BLEU setup in NLTK. 

Because `0/denominator` should be a very very rare case in `corpus_bleu` and since it is a common case in `sentence_bleu`, enforcing a smoothing method that has been proven to be effective (in various papers) is more sound than  following a bad tradition of falling back to 0 unsmoothed BLEU (though widely used).

nltk/translate/bleu_score.py
bamattsson: I assume you mean "didn't really account for the case where no. of 4-grams overlap is < 0"?

In that case their paper did account for that case as well. If you look at the equation for p_n in section 2.1 you can see that if we have zero 4-grams overlap the numerator will evaluate to 0 which means that p_n = 0. (And log(0) should evaluate to -inf in this case!) If you don't believe me ask somebody else or email the authors for example. 

What they didn't specify is when the reference sentence is shorter than N, where N is the order of N-grams used. I.e. when we have a 0/0 or x/0 situation. This is the thing I intend to solve later so that we can let 0/0 evaluate to 1 in that case.

alvations: I believe you with the -inf case =)

I meant to say:

 - 0/1 = 0  and math.log(0) = -inf is a python problem
 - 1/0 = undefined, maths error 
 - and that Papineni et al. never accounted for sentence BLEU 

How about just changing the parameter name to `original_bleu=True` by default?

alvations: Also, see the non-inline comment. We can avoid all this is we make one of the smooothing method the default for sentence BLEU. 

The argument is that there is BLEU vs sentence BLEU and sentence BLEU should never exist as unsmoothed. Does that make sense? Or am I miscommunicating again? I hope that's clear.

nltk/translate/bleu_score.py
bamattsson: I did not mention the BLEU score here, I mentioned the precision for the n-gram (or p_n in Papineni et. al.). The effect of using `backoff=True` is essentially matehmatically equivalent to letting p_n=1 in these case. Look through the equations carefully. 

It might help to use SmoothingFunction.method1, when you have set epsilon=1. Which (if we have `ref = [['hello', 'how', 'are', 'you']]` and `sent = ['hello', 'how', 'are']`) returns the same results as if SmoothingFunction.method0 was used. And if you have a look at the equation for smoothing 1 in http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf I think you'll understand =)

In my mind either you use `emulate_multibleu=True` by default, or you specify clearly in the docs that the function intentionally does not follow the equations in the papers.

bamattsson:
I don't think it's a good idea to enforce smoothing when sentence_bleu is used. We do use sentence_bleu when evaluating our chat-bots, and we intentionally do not use any smoothing function. We have carefully analysed the different smoothing functions, and decided that they are not good for us.

BTW no smoothing can currently be achieved by using smoothing function 1 with `epsiolon = sys.float_info.min`. We first used smoothing function 0, but saw some unexpected behaviour, we then read through the code, and found the bug. Which is the reason why I've tried to fix it here since :)

nltk/translate/bleu_score.py
bamattsson: I think changing the parameter name to `original_bleu` would be good, and very explanatory for the user yes :)

What should be default is up to you. However if you ask me I think `original_bleu = True` as default makes sense. Typically when you start using some new software tool you expect the default parameters to be "bare-bones", without any extra stuff etc. By the same reason I also think that the current implementation of not using any smoothing function by default, but raising a warning to urge the user to do so is a nice approach.

And yes, exactly that is what I mean: when there's zero 4-gram overlaps Papini et. al. specifies what to do, but when there isn't even 4 tokens in the reference sentence (and thus 0 4-grams) they do not specify what to do. 

You're right, they never mention sentence BLEU. And as you say corpus BLEU is probably mostly what is of interest rather than sentence BLEU. We actually use sentence BLEU and average afterwards, I don't know if that's the best, but it works well.

nltk/translate/bleu_score.py
alvations: Possibly, we need more inputs from users and other devs to know what should be the default `emulate_multibleu` or `backoff. 

Regarding the documentation, the new extended docstring  in method0 is already quite verbose =(

----

`backoff=True` effects don't always results `p_n=1`

In the simplest use case that most chatbot system might see, yes it does. Largely because the input is already `p_1=1`, `p_2=1` and `p_3=1`:

```
>>> from nltk.translate.bleu_score import SmoothingFunction
>>> from nltk.translate.bleu_score import modified_precision
>>> ref = ['hello', 'how', 'are', 'you']
>>> hyp = ['hello', 'how', 'are']
>>> zero = SmoothingFunction().method0

>>> [modified_precision([ref], hyp, n=i) for i in range(1,5)]
[Fraction(3, 3), Fraction(2, 2), Fraction(1, 1), Fraction(0, 1)]

>>> p_n = [modified_precision([ref], hyp, n=i) for i in range(1,5)]

>>> zero(p_n, emulate_multibleu=False, backoff=True)
[Fraction(3, 3), Fraction(2, 2), Fraction(1, 1)]
```

But in the case that the input with non-perfect lower order of ngram precision, that's the case for most MT outputs, the removal of the final order of n-gram doesn't result in `p_n=1` e.g.

```
>>> ref = ['hello', 'how', 'are', 'you', 'you']
>>> hyp = ['you', 'you', 'hello', 'how', 'are']
>>> p_n = [modified_precision([ref], hyp, n=i) for i in range(1,5)]
>>> p_n
[Fraction(5, 5), Fraction(3, 4), Fraction(1, 3), Fraction(0, 2)]
>>> zero(p_n, emulate_multibleu=False, backoff=True)
[Fraction(5, 5), Fraction(3, 4), Fraction(1, 3)]
```

Although to make it a real backoff function, there's more work to be done with regards to conditional probabilities but for now I think this is adequate.

alvations: This would be off-topic, I think:

Regarding `epsilon` and lapacian smoothing, it's always weird to use laplace ;P 
It's added so that it follows Chen and Cherry (2014) method 1 to 6. The usual MT strategy is to go with `method3`.

And on the averaging sentence BLEU -> corpus metric, I think that's not the norm in MT but I see how it can be useful for localized scores per chat utterance as oppose to global score.

bamattsson: @alvations 

Let's go through your last example in detail. For easier referencing later we can define the geometric mean equation: `BLEU = BP * exp(\sum_{n-1}^N w_n log(p_n))` as (1), for simplicity I'll assume w_n = 1/4 and I will disregard BP. 

Originally following their equations you have: `p_n = [5/5, 3/4, 1/3, 0/2]`, if you stick that into (1) you get: `exp(1/4 * log(1) + 1/4 * log(3/4) + 1/4 * log(1/3) + 1/4 * log(0/2))`. -> `exp(1/4 * 0 + 1/4 *(-0.288) + 1/4 * (-1.099) + 1/4 * -\infty)` -> `exp(-\infty)` -> `0`

With your backoff strategy you have: `p_n = [5/5, 3/4, 1/3, 0/2]` -> `[5/5, 3/4, 1/3]`, if you stick that into (1) you get: `exp(1/4 * log(1) + 1/4 * log(3/4) + 1/4 * log(1/3))`. -> `exp(1/4 * 0 + 1/4 *(-0.288) + 1/4 * (-1.099))` -> `exp( 1/4 *(-0.288) + 1/4 * (-1.099))`-> `exp(-0.346)` -> `0.7070`

I claim that this is equivalent of setting p_n=1 for the n:s that had no overlap (4 in this case). Let's try that: `p_n = [5/5, 3/4, 1/3, 0/2]` -> `[5/5, 3/4, 1/3, 1]`, if you stick that into (1) you get: `exp(1/4 * log(1) + 1/4 * log(3/4) + 1/4 * log(1/3) + 1/4 * log(1))`. -> `exp(1/4 * 0 + 1/4 *(-0.288) + 1/4 * (-1.099) + 1/4 * 0)` -> `exp( 1/4 *(-0.288) + 1/4 * (-1.099))` -> `exp(-0.346)` -> `0.7070`.

Do you follow? The main point is that when you set p_n=1 the log(p_n) will evaluate to 0. And you have a method that is equivalent to your backoff strategy.

So what you have done with your backoff strategy is to introduce a new type of smoothing. Which if following Chen and Cherry's notation could be described as `m_n' = l_n if m_n = 0`. 

And what buggs (sorry if I'm sounding condescending) me is that this smoothing technique is the worst I possibly can imagine. What you want from a smoothing technique (and what is the common theme through all smoothing techniques in presented by Chen and Cherry) is to if `m_n = 0` replace it with some constant value that is **small** compared to `l_n`. What you mathematically is doing is replacing it with `l_n` **itself** (which of course not is small compared to `l_n`).

alvations:
Okay, I've removed the `emulate_multibleu` and `backoff` parameters; the default is to return 0 BLEU when fringe cases appears.

Please make the changes in the doctests and unittests accordingly and perhaps someone else should review this PR.

nltk/translate/bleu_score.py
alvations: Thank you for the clarification. Yes, essentially the backoff gives full credits for the ngrams that didn't exist:

```
>>> math.exp(sum([math.log(1.0), math.log(0.8), math.log(0.5)]))
0.4

>>> math.exp(sum([math.log(1.0), math.log(0.8), math.log(0.5), math.log(1.0)]))
0.4
```

I've removed it, please change the code for the doctest and unittest accordingly.

Pardon my bad math.

bamattsson:
@alvations Thank you for understanding. I will change the docstrings, tests and doctests that needs changing, hopefully this week. I can tell you when that is done so that you can assign somebody else as reviewer, if you did not want to review it.

However as I mentioned earlier I still believe there will be a bug left in there (due to x/0 -> x/1) in the fraction calculation. This is however minor in comparison. But I will clearly point it out in the code so that it is known, and I might return back to this in November to do a larger refactoring so that it is possible to fix that bug.

bamattsson:
I see that you're making the changes now. Tell me if there is anything else you need help with.

alvations:
I'm done with the tests cleanups. The CI automation should pass.

We still have to wait for someone else to review this PR before it can get merged though =(

alvations:
~When `denominator == 0`, maybe doing `p_n_new.append(1)` should fix it.~

Actually, there is never a case of `x>0` when `denominator ==0` since there can be 0 n-grams overlaps from the hypothesis when there is 0 n-grams to overlap with in the reference.  

So the code shouldn't break unless a user artificially change the numerator =)

I think the potential bigger problem will come when users multiply the non-rounded super small values, so we might need to enforce a strict decimal place parameter to prevent underflow.

alvations:
[CI: retest]

alvations:
@bamattsson Thank you for the contribution and discussion, any more suggestions?

This PR is should be good for review. Any takers?

bamattsson:
Thanks for making the changes @alvations! And yep you're right, when denominator==0, we can never have that numerator > 0.

I'll try to have a look through the PR myself this evening :)

nltk/translate/bleu_score.py
```diff
@@ -58,8 +58,8 @@ def sentence_bleu(references, hypothesis, weights=(0.25, 0.25, 0.25, 0.25),
     >>> sentence_bleu([reference1, reference2, reference3], hypothesis1) # doctest: +ELLIPSIS
     0.5045...
 
-    >>> sentence_bleu([reference1, reference2, reference3], hypothesis2) # doctest: +ELLIPSIS
-    0.3969...
+    >>> round(sentence_bleu([reference1, reference2, reference3], hypothesis2),4) # doctest: +ELLIPSIS
+    0.0
```
dimazest: Is the change in the value justified? The previous value is taken from the paper (if I remember correctly).

If it's really 0 now, the reason should be given.

nltk/translate/bleu_score.py
```diff
@@ -140,10 +137,10 @@ def corpus_bleu(list_of_references, hypotheses, weights=(0.25, 0.25, 0.25, 0.25)
     :type weights: list(float)
     :param smoothing_function:
     :type smoothing_function: SmoothingFunction
-    :param auto_reweigh:
+    :param auto_reweigh: Option to re-normalize the weights uniformly.
     :type auto_reweigh: bool
-    :param emulate_multibleu: bool
     :return: The corpus-level BLEU score.
+    :
```
dimazest: this looks strange.

dimazest (commented):
Some minor comments.

nltk/translate/bleu_score.py
bamattsson: I now added a more in-depth description here why 0 is the expected value here.

The value was not taken from the paper. The thing that was taken from the paper were the reference and hypothesis examples. But not the values.

nltk/translate/bleu_score.py
bamattsson: removed

bamattsson:
I did some updates now. I cannot run more than the python 3.5 tests on my machine. So fingers crossed that the tests still are passing 🤞

nltk/translate/bleu_score.py
```diff
@@ -58,12 +58,23 @@ def sentence_bleu(references, hypothesis, weights=(0.25, 0.25, 0.25, 0.25),
     >>> sentence_bleu([reference1, reference2, reference3], hypothesis1) # doctest: +ELLIPSIS
     0.5045...
 
+    If no ngram overlap exists for any of the orders that are used to evaluate BLEU
+    it will return the value zero. This since BLEU is a geometric mean of the precisions
+    for the different ngram orders (no overlap means a precision of 0).
+
```
alvations: Maybe being a little verbose here would help ease the users' pain when this is merged:

> If there is no ngrams overlap for any order of n-grams, BLEU would return the value 0. This is because the precision for the order of n-grams without overlap returns a 0 and the geometric mean in the final BLEU score computation multiplies the 0 with the precision of of other n-grams, which results in 0.

bamattsson: I changed to what you proposed now.

I also expanded the description of the test below. And updated it (origianlly it said 6-grams but had 5 weights). Now it says 5grams and have 5 weights.

bamattsson:
@alvations @dimazest What's happening with this? I see that there are merge conflicts now. Tell me if you're planning to look at it, then I'll fix them

alvations:
@bamattsson the conflict is caused by PEP-zation from #1870. 

Could you help to clean up the git branch? If not, I can do it but it will have to wait for a weekend later this month when I'm free-er.

bamattsson:
Merge conflicts are now fixed @alvations @dimazest

stevenbird:
Thanks @bamattsson, @alvations, @dimazest!

PR #1845, opened 2017-10-03
add step support to Text splicing

clugg:
The current behaviour of splicing a Text instance is unexpected:

```python
>>> import nltk
>>> text = nltk.Text(nltk.word_tokenize("okay when lisa and raymond got home from school"))
>>> text
<Text: okay when lisa and raymond got home from...>
>>> text[:2]
['okay', 'when']
>>> text[:2:2]
['okay', 'when']
```

This change simply takes into consideration the `step` attribute of a `slice` when one is provided.

### DIFF
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -311,7 +311,7 @@ class Text(object):
 
     def __getitem__(self, i):
         if isinstance(i, slice):
-            return self.tokens[i.start:i.stop]
+            return self.tokens[i.start:i.stop:i.step]
         else:
             return self.tokens[i]
 


alvations:
Interesting! Thanks @clugg for catching this =)

The function can be even simpler since the `self.tokens` is a `list` type and lists takes `slice` types as input for their index, e.g.

```
>>> import nltk
>>> text = nltk.Text(nltk.word_tokenize("okay when lisa and raymond got home from school"))

>>> type(text)
<class 'nltk.text.Text'>
>>> type(text.tokens)
<class 'list'>

>>> text[slice(0,2,2)]
['okay']
```

so this should work too:

```
    def __getitem__(self, i):
            return self.tokens[i]
```

Maybe the `isinstance(i, slice)` check was an artifact of pre-[PEP 357](https://www.python.org/dev/peps/pep-0357/)?

stevenbird:
Thanks for this @clugg. How about submitting a PR based on @alvations's suggestion

stevenbird:
@alvations, thanks for pointing out the more general solution.

PR #1848, opened 2017-10-05
Add docs for local development environment set-up

jacksonllee:
This PR provides more explicit notes for how to set up a local development environment. Hopefully the notes will be helpful for getting more people started with contributions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,8 +51,6 @@ repository [nltk/nltk](https://github.com/nltk/nltk/):
   (`git clone https://github.com/<your-github-username>/nltk.git`);
 - Run `cd nltk` to get to the root directory of the `nltk` code base;
 - Install the dependencies (`pip install -r pip-req.txt`);
-- Install the `nltk` library in development mode
-  (`python setup.py develop`);
 - Download the datasets for running tests
   (`python -m nltk.downloader tests`);
 - Create a remote link from your local repository to the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,8 +53,8 @@ repository [nltk/nltk](https://github.com/nltk/nltk/):
 - Install the dependencies (`pip install -r pip-req.txt`);
 - Install the `nltk` library in development mode
   (`python setup.py develop`);
-- Download the datasets (`python -c "import nltk; nltk.download('all')"`) --
-  they are needed for running tests locally;
+- Download the datasets for running tests
+  (`python -c "import nltk; nltk.download('tests')"`);
 - Create a remote link from your local repository to the
   upstream `nltk/nltk` on GitHub
   (`git remote add upstream https://github.com/nltk/nltk.git`) --
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,23 +39,47 @@ contribute is to learn how to use it and put your changes on a Git repository.
 There's a plenty of documentation about Git -- you can start with the [Pro Git
 book](http://git-scm.com/book/).
 
-### Forks + GitHub Pull requests
+
+### Setting up a Development Environment
+
+To set up your local development environment for contributing to the main
+repository [nltk/nltk](https://github.com/nltk/nltk/):
+
+- Fork the [nltk/nltk](https://github.com/nltk/nltk/) repository on GitHub
+  to your account;
+- Clone your forked repository locally
+  (`git clone https://github.com/<your-github-username>/nltk.git`);
+- Run `cd nltk` to get to the root directory of the `nltk` code base;
+- Install the dependencies (`pip install -r pip-req.txt`);
+- Install the `nltk` library in development mode
+  (`python setup.py develop`);
+- Download the datasets (`python -c "import nltk; nltk.download('all')"`) --
+  they are needed for running tests locally;
+- Create a remote link from your local repository to the
+  upstream `nltk/nltk` on GitHub
+  (`git remote add upstream https://github.com/nltk/nltk.git`) --
+  you will need to use this `upstream` link when updating your local repository
+  with all the latest contributions.
+
+### GitHub Pull requests
 
 We use the famous
 [gitflow](http://nvie.com/posts/a-successful-git-branching-model/) to manage our
 branches.
 
 Summary of our git branching model:
-- Fork the desired repository on GitHub to your account;
-- Clone your forked repository locally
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -39,23 +39,47 @@ contribute is to learn how to use it and put your changes on a Git repository.
...
+
+- Fork the [nltk/nltk](https://github.com/nltk/nltk/) repository on GitHub
+  to your account;
+- Clone your forked repository locally
+  (`git clone https://github.com/<your-github-username>/nltk.git`);
+- Run `cd nltk` to get to the root directory of the `nltk` code base;
+- Install the dependencies (`pip install -r pip-req.txt`);
+- Install the `nltk` library in development mode
+  (`python setup.py develop`);
+- Download the datasets (`python -c "import nltk; nltk.download('all')"`) --
```
jacksonllee: Added this point about getting the NLTK datasets, as a few tests failed because of the lack of several datasets when I tried to run the tests. I put down `'all'` for convenience, but felt slightly uncomfortable because of wasting unnecessary resources (cf. #1787). If it'd be more desirable to download just those datasets one by one for the sake of running the tests (this would seem a little weird? I don't have strong feelings though), I'd be happy to single those out and update this PR.

stevenbird:
@jacksonllee, thanks for this. Yes I would prefer not to suggest `all`, given the demands we already place on GitHub.

CONTRIBUTING.md
```diff
@@ -39,23 +39,47 @@ contribute is to learn how to use it and put your changes on a Git repository.
...
+- Fork the [nltk/nltk](https://github.com/nltk/nltk/) repository on GitHub
+  to your account;
+- Clone your forked repository locally
+  (`git clone https://github.com/<your-github-username>/nltk.git`);
+- Run `cd nltk` to get to the root directory of the `nltk` code base;
+- Install the dependencies (`pip install -r pip-req.txt`);
+- Install the `nltk` library in development mode
+  (`python setup.py develop`);
+- Download the datasets for running tests
+  (`python -c "import nltk; nltk.download('tests')"`);
```
jacksonllee: I've created the label `'tests'` to cover all and just the resources needed for running tests. The list of these resources are defined in the `downloader.py` module.

stevenbird:
@jacksonllee: here's where we list collections of corpora. Would you mind submitting a PR there for `tests.xml` please?

https://github.com/nltk/nltk_data/tree/gh-pages/collections

jacksonllee:
@stevenbird Done, `tests.xml` added at https://github.com/nltk/nltk_data/pull/93

CONTRIBUTING.md
jacksonllee: Update: Instead of pinning down the list of "tests" packages in `downloader.py`, the list is now defined in the `nltk_data` repo. See https://github.com/nltk/nltk_data/pull/93

CONTRIBUTING.md
```diff
@@ -39,23 +39,47 @@ contribute is to learn how to use it and put your changes on a Git repository.
...
+  (`git pull upstream develop`);
 - Create a new branch off of `develop` with a descriptive name (for example:
   `feature/portuguese-sentiment-analysis`, `hotfix/bug-on-downloader`). You can
   do it switching to `develop` branch (`git checkout develop`) and then
   creating a new branch (`git checkout -b name-of-the-new-branch`);
 - Do many small commits on that branch locally (`git add files-changed`,
   `git commit -m "Add some change"`);
-- Add your name to the `AUTHORS.markdown` file as a contributor;
+- Run the tests to make sure nothing breaks
+  (`tox -e py35` if you are on Python 3.5);
```
alvations, line 79: 👍

CONTRIBUTING.md
alvations: Use `python -m nltk.downloader tests` instead of `python -c ...` =)

CONTRIBUTING.md
```diff
@@ -39,23 +39,47 @@ contribute is to learn how to use it and put your changes on a Git repository.
...
+repository [nltk/nltk](https://github.com/nltk/nltk/):
+
+- Fork the [nltk/nltk](https://github.com/nltk/nltk/) repository on GitHub
+  to your account;
+- Clone your forked repository locally
+  (`git clone https://github.com/<your-github-username>/nltk.git`);
+- Run `cd nltk` to get to the root directory of the `nltk` code base;
+- Install the dependencies (`pip install -r pip-req.txt`);
+- Install the `nltk` library in development mode
+  (`python setup.py develop`);
```
alvations: I don't think there's a need to install the `nltk` in development mode but I might be wrong. I've been testing with `tox` without installing and not letting `setup.py` override my environment's `nltk`.

CONTRIBUTING.md
jacksonllee: I followed the CI build instructions where `python -c` is used: 

https://github.com/nltk/nltk/blob/0c7e43f0bd43ed8f028a13053a6a3edf1749e10d/jenkins.sh#L9

but if `nltk.downloader` is already set up as a script, I'm switching to `python -m` now.

CONTRIBUTING.md
jacksonllee: tl;dr -- I'm removing this `python setup.py develop` line. Simpler is better. :-)

I'm not entirely sure either, but I have the opposite situation (and see it with co-workers as well): We install a library in the dev mode so that we can use it not necessarily from the repo root directory. If we want the upstream official release version of the library, we check out the `develop` branch (or the master branch for other projects). If we want to use or test whatever new feature being developed, we check out the relevant branch instead. Now that I've spelled this out, I see the advantage of this workflow is to avoid potential confusion by having only one installation of `nltk` (this is the part that's opposite of what you achieve for not overriding the installed, official version by not installing a dev version), with the trade-off of slightly more git work for always committing work in progress to allow the back-and-forth branch switching. That said, I think it'd be better to keep things simple for the purpose of getting more people started with contributing to NLTK. And I think you're right that `tox` should work without installing `nltk`, if one is at the repo root directory where the directory `nltk/` is available as the local relative-path package.

jacksonllee:
@stevenbird @alvations Requested changes made -- ready for another pass.

alvations:
Thanks @jacksonllee! hopefully the updated `contribute.md` ease first time contributors pain =)

I think it LGTM.

alvations (approved):

stevenbird:
Thanks @jacksonllee @alvations!

PR #1857, opened 2017-10-13
Fix deprecation message for Tagger classes

yannick-couzinie:
Small update to the deprecation message, which was wrongly pointing to `nltk.tag.corenlp.CoreNLP---Tagger` when it should have been `nltk.tag.stanford.CoreNLP---Tagger`.

I don't know if this is intentional to change the name of `stanford.py` to `corenlp.py` in the next release. But in my opinion the deprecation message should correctly point to the currently correct location (if I haven't missed anything here, this should be correct).


stevenbird:
:+1:

stevenbird:
Thanks @yannick-c

PR #1864, opened 2017-10-17
Fix TreebankWordTokenizer.span_tokenize() Exceptions for Corner-case Input

yibinlin:
Please see the doctest str as corner-case test string. The existing develop branch implementation will throw out an error against the str.

Also line 172 regex in existing implementation seems to be wrong - it has duplicate characters in the group. Even if we remove the enclosing character group, it will throw error when testing against my doctest str, because the tokenizer will view "''good" as one token (there are two single quotes ', instead of a double quote ", before the word "good". This is a real-life example I saw in my work).

This solution will work if TreebankWordTokenizer.tokenize() only modifies double quote character and nothing else.


yibinlin:
create [issue 1865](https://github.com/nltk/nltk/issues/1865) for this.

nltk/tokenize/treebank.py
```diff
@@ -147,36 +147,46 @@ def tokenize(self, text, convert_parentheses=False, return_str=False):
...
+            ... (69, 71), (72, 75), (76, 77), (77, 81), (81, 82), (82, 83),
+            ... (83, 84)]
             >>> TreebankWordTokenizer().span_tokenize(s) == expected
             True
-            >>> expected = ['Good', 'muffins', 'cost', '$', '3.88', 'in',
-            ... 'New', '(', 'York', ')', '.', 'Please', '(', 'buy', ')',
-            ... 'me', 'two', 'of', 'them.', '(', 'Thanks', ')', '.']
+            >>> expected = ['I', 'said', ',', '"', 'I', "'d", 'like', 'to',
+            ... 'buy', 'some', "''good", 'muffins', '"', 'which', 'cost', '$',
+            ... '3.88', 'each', 'in', 'New', '(', 'York', ')', '.', '"']
```
alvations: Instead of changing the example, could you add this as an additional example?

nltk/tokenize/treebank.py
```diff
@@ -147,36 +147,46 @@ def tokenize(self, text, convert_parentheses=False, return_str=False):
...
+                else:
+                    real_token = '"'
+            else:
+                real_token = word_token
+            ix = text.find(real_token, ix)
+            end = ix + len(real_token)
+            spans.append((ix, end))
+            ix = end
+
+        return spans
```
alvations: Maybe it's better for this code to improve  `align_tokens()` at https://github.com/nltk/nltk/blob/develop/nltk/tokenize/util.py#L220 so that its more of a generic function and not just specific to the `TreebankTokenizer`

yibinlin: Thanks. I think it is a good idea. However, I am not sure if the align_tokens() function will be used in other tokenizers' span_tokenize() function in the future? If other tokenizers have other unique behaviors that will modify "tokens" of the original string (similar to modifying " to ``. Is it possible to modify . . . (as one token) to ... in another tokenizer? This is just a made-up example). In that case align_tokens() will still break if such other tokenizer uses align_tokens().

I think the current align_tokens() should work well if a tokenizer does not modify any "tokens". I can find examples for such tokenizer, e.g. tokenize.util.string_span_tokenize(), tokenize.util.regexp_span_tokenize() - they are very simple tokenizers tho.

Adding my code to align_tokens() will not do harm in terms of quality, but it would add some computational cost for tokenizers which doesn't modify " to `` or ''.

Another possibility is to extend align_tokens() to take one additional optional parameter "modified_tokens", e.g. {'"': ['``', "''"]} to specify that some tokens may be modified to other forms.

I would like to hear your thought. Thank you!

alvations: I was thinking more of replacing the whole `align_tokens()` with a better code like yours.

But I also think there're better string search and offset indexing strategies that should be considered if possible. 

The ideal solution would have to be that the offsets are kept and updated during tokenization =(

Someone else with a better understanding of regex and strings should review this PR.

yibinlin: Thank you @alvations ! I agree with you about the ideal solution where the offset should be kept during the tokenize() logic. 

Here are just my two cents:
I am not very familiar with the tokenizer regex code. But from what I can see, there are quite a few modifications (some of them are transient modifications that will be modified back to original state) in the tokenize() function of the  TreebankWordTokenizer. Therefore keeping the original offsets might require quite a lot of thought and new code - and it could even be computationally more expensive than my current method. But once it's done it should guarantee correctness.

If it sounds OK we can start with my current method as a workaround solution - I can also add some note in tokenize() method comment, to update the span_tokenize() method when some other special tokens will be modified (like " => ``) in the future?

alvations:
Sorry for taking longer to get back on this. 

The conversion loop you've wrote works but I'm not sure whether that's the best way to resolve the problem ~~because there could also be instances of single-quotes that converts single quote to backtick.~~

There's just something fundamentally wrong when the offsets are not kept at tokenization but that's a long-shot issue. Let me take a closer look at it.

alvations:
After checking the brown corpus and PTB, it seems like the problem happens not at the `realign_token` but at the tokenization time =)

This is because:

```python
>>> from nltk.tokenize import TreebankWordTokenizer
>>> s = '''I said, "I'd like to buy some ''good muffins" which cost $3.88\\n each in New (York)."'''
>>> TreebankWordTokenizer().tokenize(s)
['I', 'said', ',', '``', 'I', "'d", 'like', 'to', 'buy', 'some', "''good", 'muffins', "''", 'which', 'cost', '$', '3.88\\n', 'each', 'in', 'New', '(', 'York', ')', '.', "''"]
```

Note that it's happening because `"''good"` is being tokenized as a single token. So naturally the  `align_token` would fail because it's trying to file `''` as a single token. 

That makes it a little easier to resolve! 

We just got to find the suitable regexes to append to the `TreebankWordTokenizer`:

```
    #starting quotes
    STARTING_QUOTES = [
        (re.compile(r'^\"'), r'``'),
        (re.compile(r'([ (\[{<])"'), r'\1 `` '),
    ]
```

alvations:
This is the regex `re.compile(r'([ (\[{<])"')` that's responsible to post-padding the opening brackets with space

To handle the double single quotes and treat it the same, we can use  [`([ (\[{<])(\"|\'{2})`](https://regex101.com/r/7cTspL/3)

Now, we also have to find the regex fix to the ending quotes: 

```
    #ending quotes
    ENDING_QUOTES = [
        (re.compile(r'"'), " '' "),
        (re.compile(r'(\S)(\'\')'), r'\1 \2 '),
        (re.compile(r"([^' ])('[sS]|'[mM]|'[dD]|') "), r"\1 \2 "),
        (re.compile(r"([^' ])('ll|'LL|'re|'RE|'ve|'VE|n't|N'T) "), r"\1 \2 "),
    ]
```

alvations:
Awesome, the original tokenizer regex from Robert McIntyre already covered the closing quotes from [`(\S)(\'\')`](https://regex101.com/r/fAKKXI/1):

```
# Using the v3.2.5 version.
>>> s= '''I said, "I'd like to buy some ''good muffins'' which cost $3.88\\n each in New (York)."'''
>>> TreebankWordTokenizer().tokenize(s)
['I', 'said', ',', '``', 'I', "'d", 'like', 'to', 'buy', 'some', "''good", 'muffins', "''", 'which', 'cost', '$', '3.88\\n', 'each', 'in', 'New', '(', 'York', ')', '.', "''"]
```

We're good to go, with the `([ (\[{<])("|'{2})` patch in the TreebankWordTokenizer `STARTING_QUOTES` regex, that resolve the tokenization part:

```
>>> from nltk.tokenize import TreebankWordTokenizer
>>> s = '''I said, "I'd like to buy some ''good muffins" which cost $3.88\\n each in New (York)."'''
>>> TreebankWordTokenizer().tokenize(s)
['I', 'said', ',', '``', 'I', "'d", 'like', 'to', 'buy', 'some', '``', 'good', 'muffins', "''", 'which', 'cost', '$', '3.88\\n', 'each', 'in', 'New', '(', 'York', ')', '.', "''"]
```

Now we can go back and look at the `span_tokenize()` and it has became a simpler problem.

alvations:
Ah! There's a problem with the current regex when  `span_tokenize` regex to tries tocatch the `matches` at https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L172. 

The regex there seems to be a PHP regex format ;P

If we change that regex to this: `r"``|'{2}|\""` and also the changes to the `STARTING_QUOTES` regexes during tokenization, we get the output:

```
>>> from nltk.tokenize import TreebankWordTokenizer
>>> s = '''I said, "I'd like to buy some ''good muffins'' which cost $3.88\\n each in New (York)."'''
>>> TreebankWordTokenizer().span_tokenize(s)
[(0, 1), (2, 6), (6, 7), (8, 9), (9, 10), (10, 12), (13, 17), (18, 20), (21, 24), (25, 29), (30, 32), (32, 36), (37, 44), (44, 46), (47, 52), (53, 57), (58, 59), (59, 65), (66, 70), (71, 73), (74, 77), (78, 79), (79, 83), (83, 84), (84, 85), (85, 86)]

>>> TreebankWordTokenizer().tokenize(s)
['I', 'said', ',', '``', 'I', "'d", 'like', 'to', 'buy', 'some', '``', 'good', 'muffins', "''", 'which', 'cost', '$', '3.88\\n', 'each', 'in', 'New', '(', 'York', ')', '.', "''"]

>>> s = '''I said, "I'd like to buy some ''good muffins" which cost $3.88\\n each in New (York)."'''
>>> TreebankWordTokenizer().span_tokenize(s)
[(0, 1), (2, 6), (6, 7), (8, 9), (9, 10), (10, 12), (13, 17), (18, 20), (21, 24), (25, 29), (30, 32), (32, 36), (37, 44), (44, 45), (46, 51), (52, 56), (57, 58), (58, 64), (65, 69), (70, 72), (73, 76), (77, 78), (78, 82), (82, 83), (83, 84), (84, 85)]
>>> TreebankWordTokenizer().tokenize(s)
['I', 'said', ',', '``', 'I', "'d", 'like', 'to', 'buy', 'some', '``', 'good', 'muffins', "''", 'which', 'cost', '$', '3.88\\n', 'each', 'in', 'New', '(', 'York', ')', '.', "''"]
```

alvations:
@yibinlin please revert the changes you've made to the `treebank.py` and then do these changes instead:

 - This line https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L70 to `(re.compile(r"([ (\[{<])(\"|\'{2})"), r'\1 `` ')` and

 - this line https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L172 to `matched = [m.group() for m in re.finditer(r"``|'{2}|\"", text)]`

And that'll resolve the issue =)

Please make the above changes and we'll do another round of light review before merging. Thank you for catching this problem!

alvations (changes requested):
Please make the suggested changes in the conversation of the PR.

stevenbird:
@yibinlin - ping

yibinlin:
@alvations I think your changes make sense. Thanks! 

I am currently traveling. I will send out a new version for review in a day or two.

yibinlin:
Sorry for the late update - now I made the changes:

1. Changed STARTING_QUOTE and span_tokenize as @alvations  suggested.
2. Added a test case in doctest
3. Added some minor in-line comments.

Please review when you have time @alvations  Thank you!

stevenbird:
Thanks @yibinlin!

alvations (approved):

alvations:
Thanks @yibinlin for the changes! 

LGTM now =)

yibinlin:
Thank you @alvations ! @stevenbird  Do I need to do anything to merge it, or someone with write access will do the merging?

alvations:
@yibinlin nope, we're just waiting for @stevenbird to do a quick last check and merge =)
Thank you again!

stevenbird:
LGTM - thanks @yibinlin and @alvations

gkucsko:
There seem to be some more corner cases that break during span_tokenize, eg:
```
>>> from nltk.tokenize import TreebankWordTokenizer
>>> s = u"\'\'good muffins\""
>>> list(TreebankWordTokenizer().span_tokenize(s))
my_venv/lib/python2.7/site-packages/nltk/tokenize/util.pyc in align_tokens(tokens, sentence)
    256             start = sentence.index(token, point)
    257         except ValueError:
--> 258             raise ValueError('substring "{}" not found in "{}"'.format(token, sentence))
    259         point = start + len(token)
    260         offsets.append((start, point))

ValueError: substring "''" not found in "''good muffins""
```

gkucsko:
@stevenbird the problem arises in the case of `''` at the beginning of a line. this is due to the fact that treebank tokenizes the following cases differently: `[" \'\'good muffins\"", "\'\'good muffins\"", "\'\' good muffins\""]`, which in turn causes incorrect match groups for keeping track of the substituted quotation characters.

a fix using the current implementation would be in line 190 of tokenize/treebank.py:
```
 matched = [m.group() for m in re.finditer(r"\`\`|(?<!^)\'\'|^\'\'(?=\s)|\"", text)]
```
this is starting to look pretty ugly, but at least wouldn't change the current implementation at all. happy to submit merge req if this makes sense.

yibinlin:
Sorry for the late reply - I can confirm the issue locally on my laptop too. @alvations FYI.

gkucsko:
thanks, opened as new bug in #2076

alvations:
I think the treebank tokenizer should be left without a span tokenization because the regex replacement without keeping an offset is not sustainable. We can patch one regex and something else will cause it to break.

slbayer:
@alvations is right. The problem is that the regex-based, non-string-preserving tokenization is impossible to reverse engineer reliably using these simple techniques. I'm trying to write a reliable alternative, and so far I've tried using difflib, which mostly works but doesn't all the time because it produces non-minimal alignments, and then throwing the Hirschberg algorithm at the remainder, but that STILL breaks because of some perverse cases I'm still tracking down.

My recommendation, if you care, is that the regex-based tokenization approach here is untenable if you want to retrieve spans, unless you want to throw a lot of additional machinery at reverse engineering the output. I'd recommend replacing it with an explicit state machine.

PR #1867, opened 2017-10-20
Renovations to CorpusView

rmalouf:
The CorpusReader/CorpusView mechanism is starting to show its age.  It was designed to facilitate random access to tokens in a corpus stored with a 1 byte per character encoding while using a very small, predictable amount of memory.  These days a more common use is likely sequential access to tokens stored in a multi-byte encoding on a machine with tons of excess RAM.  

This PR is a first step towards updating the corpus reader mechanism.  Besides some general code cleanups, it replaces SeekableUnicodeStream with io.TextIOWrapper from the standard library (which python2 has had since 2.6).  Random access to files at the character level is reasonably fast in python3 but unusably slow in python2.  I've submitted PR that fixes it in python2, but until/unless that gets accepted, we can read each file into a string buffer and work with that.  This potentially increases the memory footprint, since we'll need to be able read in each corpus file as a unit. There will still only be one piece loaded at once, though, so this shouldn't cause a problem for anyone 

There's a lot more that could be done: I bet most corpora used by most NLTK users would easily fit in memory, and the XML handling is still kind of a disaster, but this is a start.


stevenbird:
Thanks @rmalouf. I've created a new branch `new-corpus-view` and would be grateful if you would submit the PR there.

rmalouf:
Done!

stevenbird:
Thanks @rmalouf

PR #1868, opened 2017-10-21
Typo in tokenizer docstring

zuik:
I think this is suppose to be the type of the `text` param instead of the param itself right?


alvations (approved):

stevenbird:
Thanks @zuik

PR #1881, opened 2017-11-07
Fixed parentheses in MASI distance

aliiae:
This fixes MASI distance so it is calculated without parentheses as in the original paper (see https://github.com/nltk/nltk/pull/295#commitcomment-25409252 and below).
A counterexample provided by @drevicko against the current implementation: if two sets are disjoint, the distance between them will be 0 instead of 1.

Doctest related to MASI is also modified accordingly.

### DIFF
--- a/nltk/metrics/distance.py
+++ b/nltk/metrics/distance.py
@@ -121,7 +121,7 @@ def masi_distance(label1, label2):
 
     >>> from nltk.metrics import masi_distance
     >>> masi_distance(set([1, 2]), set([1, 2, 3, 4]))
-    0.335
+    0.665
 
     Passonneau 2006, Measuring Agreement on Set-Valued Items (MASI)
     for Semantic and Pragmatic Annotation.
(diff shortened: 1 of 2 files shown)


drevicko:
looks good to me

nltk/metrics/distance.py
```diff
@@ -140,7 +140,7 @@ def masi_distance(label1, label2):
     else:
         m = 0
 
-    return (1 - (len_intersection / float(len_union))) * m
+    return 1 - len_intersection / float(len_union) * m
```
alvations: `return 1 - len_intersection / len_union * m` can avoid type casting.

stevenbird:
Thanks all

drevicko:
Are we intending for this to work in Python 2?
If so, the the type cast to float removed in the last commit is necessary...

aliiae:
We have `from __future__ import division`, so I guess it should work fine in Python 2.

PR #1898, opened 2017-11-27
Fixed pickle bug for MWETokenizer #1761

### LINKED ISSUE #1761: MWE tokenizer doesnt support pickling
You can reproduce this as follows on python 2.7.11

````
from nltk.tokenize import MWETokenizer
import pickle

tokenizer = MWETokenizer([('a', 'little'), ('a', 'little', 'bit'), ('a', 'lot')])

with open('tokenizer.pkl', 'wb') as f:
    pickle.dump(tokenizer, f)

with open('tokenizer.pkl', 'rb') as f:
    t = pickle.load(f)

````

Error that i am getting is

````
TypeError                                 Traceback (most recent call last)
<ipython-input-47-205693e12b16> in <module>()
      1 with open('tokenizer.pkl', 'rb') as f:
----> 2     t = pickle.load(f)

/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.pyc in load(file)
   1382 
   1383 def load(file):
-> 1384     return Unpickler(file).load()
   1385 
   1386 def loads(str):

/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.pyc in load(self)
    862             while 1:
    863                 key = read(1)
--> 864                 dispatch[key](self)
    865         except _Stop, stopinst:
    866             return stopinst.value

/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.pyc in load_reduce(self)
   1137         args = stack.pop()
   1138         func = stack[-1]
-> 1139         value = func(*args)
   1140         stack[-1] = value
   1141     dispatch[REDUCE] = load_reduce

/Users/vinay/Sites/nltk_mwe/.env/lib/python2.7/site-packages/nltk/collections.pyc in __init__(self, strings)
    611         defaultdict.__init__(self, Trie)
    612         if strings:
--> 613             for string in strings:
    614                 self.insert(string)
    615 

TypeError: 'type' object is not iterable
````

ethnhll:
To support pickling for MWETokenizer and fix #1761, I used a similar approach to the solution for #393.
The issue, as a commenter (Dobatymo) already pointed out, was that `MWETokenizer` makes use of `Trie` in collections. Since `Trie` is a `defaultdict`, pickling is difficult without heavy modification. The path of least resistance seemed to be to make `Trie` inherit from a plain dictionary instead and simply override `__missing__`. This allowed me to remove the `as_dict()` from `Trie`, which converted the underlying `defaultdict` to `dict`
I'll leave it up to the reviewers to decide whether `as_dict` should be put back in order to avoid breaking code that might make use of it, since it was a part of the public contract of `Trie`

Thanks for taking the time to review my PR, hope this helps!

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -226,6 +226,7 @@
 - Björn Mattsson
 - Oleg Chislov
 - Pavan Gururaj Joshi <https://github.com/PavanGJ>
+- Ethan Hill <https://github.com/hill1303>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer
--- a/nltk/collections.py
+++ b/nltk/collections.py
@@ -591,7 +591,7 @@ class LazyIteratorList(AbstractLazySequence):
 ######################################################################
 # Trie Implementation
 ######################################################################
-class Trie(defaultdict):
+class Trie(dict):
     """A Trie implementation for strings"""
     LEAF = True
 
@@ -607,7 +607,7 @@ class Trie(defaultdict):
         :type strings: list(str)
 
         """
-        defaultdict.__init__(self, Trie)
+        super(Trie, self).__init__()
         if strings:
             for string in strings:
                 self.insert(string)
@@ -621,9 +621,11 @@ class Trie(defaultdict):
         :Example:
 
         >>> from nltk.collections import Trie
-        >>> trie = Trie(["ab"])
-        >>> trie
-        defaultdict(<class 'nltk.collections.Trie'>, {'a': defaultdict(<class 'nltk.collections.Trie'>, {'b': defaultdict(<class 'nltk.collections.Trie'>, {True: None})})})
+        >>> trie = Trie(["abc", "def"])
+        >>> expected = {'a': {'b': {'c': {True: None}}}, \
+                        'd': {'e': {'f': {True: None}}}}
+        >>> trie == expected
+        True
 
         """
         if len(string):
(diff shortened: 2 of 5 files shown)


alvations:
@rmalouf Could you help to take a look at this PR? 
Is removing Trie for the `MWETokenizer` a good idea?

Thanks in advance!

ethnhll:
Hi @alvations (and @rmalouf) thanks for taking a look! Just to be clear with my change here, there wouldn't be a need for removing Trie from MWETokenizer, just a change of the parent class of Trie from a defaultdict to a dict. The only things that have been removed are methods converting the internals of Trie from a defaultdict to a dict for printing purposes. 
Thanks again for taking a look at this!

rmalouf:
It looks fine to me, but that version of Trie was actually written by @nschneid

nschneid:
> It looks fine to me, but that version of Trie was actually written by @nschneid

Actually, I think I just moved it from util.py to collections.py: #1382

rmalouf:
After a little more git excavating: maybe it was @mrecachinas ?

mrecachinas:
@rmalouf is partially correct. There were originally two trie implementations: one a `dict` in mwe.py and another a full class in sem/logic.py called `StringTrie` which subclassed `defaultdict`. I synthesized the two into util.py. I’m not the original author of either though. 

Moving the base class from `defaultdict` to `dict` seems totally reasonable to me.

stevenbird:
I think I was the author of the original trie implementation in util.py.
I'm also fine with the change from `defaultdict` to `dict`.

stevenbird:
Thanks @hill1303 @nschneid @rmalouf @alvations

ethnhll:
Thanks @stevenbird, @nschneid, @alvations, @mrecachinas, and @rmalouf for taking the time to review and discuss my changes :)

PR #1899, opened 2017-11-27
Make align functions public to be consistent with other functions

### LINKED ISSUE #1886: __align and __align_all in IBMModel should be public
The `__align` and `__align_all` functions in `IBMModel`s should should be public. What use is an alignment model if it can only be run on the train set?

hoontw:
Fixes #1886

### DIFF
--- a/nltk/translate/ibm1.py
+++ b/nltk/translate/ibm1.py
@@ -139,7 +139,7 @@ class IBMModel1(IBMModel):
         for n in range(0, iterations):
             self.train(sentence_aligned_corpus)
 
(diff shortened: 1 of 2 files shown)


alvations:
LTGM, it's just a change of function names from `__*` to `*` =)

alvations (approved):
Still, I don't think there are reasons to use the pre-trained IBM models and it shouldn't be hard to continue re-adjusting the probabilities on new datasets but this is a harmless improvement to standardize the code =)

stevenbird:
Thanks @hoontw

Issue #1901, opened 2017-11-27
Differences between FrameNet in Python3.6 and Python2.7


alvations:
Framenet in Python3.6 and Python2.7 seems to behave differently:

From https://github.com/nltk/nltk/blob/develop/nltk/test/framenet.doctest#L119

**Python3.6**

```python
~/git-stuff/nltk$ python3
Python 3.6.1 (default, Apr  4 2017, 09:40:21) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from nltk.corpus import framenet as fn
>>> fn.frames_by_lemma(r'(?i)a little') 
[<frame ID=2001 name=Degree>, <frame ID=189 name=Quantified_mass>]
>>> exit()
```

**Python2.7**

```python
~/git-stuff/nltk$ python
Python 2.7.13 (default, Dec 18 2016, 07:03:39) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from nltk.corpus import framenet as fn
>>> fn.frames_by_lemma(r'(?i)a little')
[<frame ID=189 name=Quantified_mass>, <frame ID=2001 name=Degree>]
```

Is the order of the frames important?

@nschneid any hints on why this is happening?

----

Another instance is during the doctest at https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/framenet.py#L2250

**Python 3.6**

```
~/git-stuff/nltk$ python3
Python 3.6.1 (default, Apr  4 2017, 09:40:21) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from nltk.corpus import framenet as fn
>>> from nltk.corpus.reader.framenet import PrettyDict
>>> frts = list(fn.frame_relation_types())
>>> PrettyDict(frts[0], breakLines=True)
{'ID': 10,
 '_type': 'framerelationtype',
 'frameRelations': [<Causative=Cause_to_move_in_place -- Causative_of -> Inchoative/state=Moving_in_place>, <Causative=Apply_heat -- Causative_of -> Inchoative/state=Absorb_heat>, ...],
 'name': 'Causative_of',
 'subFrameName': 'Inchoative/state',
[...]

Issue #1659, opened 2017-03-19
NLTK support for Python 3.6


alvations:
Would NLTK be supporting Python3.6? 

Python3.6 had been shipped in Dec 2016 as scheduled on  [PEP 494](https://www.python.org/dev/peps/pep-0494/) . There's also an increase in Py3.6 uptake and the shelf-life should be around 2 years before [Py3.7 reaches the final dev stage](https://www.python.org/dev/peps/pep-0537/).

This might help in the port to Python 3.6 while keeping 2.7 and 3.4/3,5 functionalities: https://docs.python.org/3/whatsnew/3.6.html

stevenbird:
:+1:

nschneid:
When I import the framenet module in Python 3.6.0 with warnings enabled via the `-Wd` flag, I get:

```
/Users/ns1057/anaconda/lib/python3.6/site-packages/nltk/decorators.py:59: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  regargs, varargs, varkwargs, defaults = inspect.getargspec(func)
```

among other non-deprecation warnings. Then, the following code

```py
from nltk.corpus import framenet as fn
sents = fn.ft_sents('ANC__HistoryOfJerusalem')
for i,sent in enumerate(sents):
    pass
```

causes this warning when the loop is exhausted:

```
__main__:1: DeprecationWarning: generator 'LazyIteratorList.iterate_from' raised StopIteration
```

which I believe is due to [PEP 479](https://www.python.org/dev/peps/pep-0479/), and can be avoided by modifying the `LazyIteratorList.iterate_from()` method (which is a generator) to catch `StopIteration` and simply return. Perhaps other collections methods suffer from this as well—e.g. `LazyZip.iterate_from()`, which is another generator that calls `next()` on an iterator. Someone who understands PEP 479 better than I do should probably take a look. :)

alvations:
Bumping issue. Any brave soul(s)?

natgillin:
Porting from py35 to py36 might not be that hard with only 4 failed test =)

```
~/git-stuff/nltk$ tox -e py36

GLOB sdist-make: /Users/minion/git-stuff/nltk/setup.py
py36 create: /Users/minion/git-stuff/nltk/.tox/py36
py36 installdeps: numpy, nose >= 1.2.1, coverage, text-unidecode, twython, pyparsing, python-crfsuite, rednose
[...]

PR #1902, opened 2017-11-28
Framenet: Casting list and dict into sorted for doctests

alvations:
Added a `__lt__` function so that frames (i.e. `AttrDict`) can be sorted by their IDs. 

Making sure that doctests for Framenet are consistent across Python versions. #1901 

Preparing for Py3.6 support #1659

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -22,6 +22,7 @@ from six import string_types, text_type
 from six.moves import zip_longest
 
 from collections import defaultdict, OrderedDict
+from operator import itemgetter
 from pprint import pprint, pformat
 from nltk.internals import ElementWrapper
 from nltk.corpus.reader import XMLCorpusReader, XMLCorpusView
@@ -685,10 +686,6 @@ class AttrDict(dict):
         if isinstance(v,Future):
             return v._data()
         return v
-
-    def __lt__(self, other):
-        """ Comparison operator used mainly to sort a list of AttrDict by ID. """
-        return self['ID'] < other['ID']
 
     def _short_repr(self):
         if '_type' in self:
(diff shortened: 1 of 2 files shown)


alvations:
[CI: restest]

alvations:
[CI: retest]

stevenbird:
Thanks in advance @nschneid :)

nltk/corpus/reader/framenet.py
```diff
@@ -686,6 +686,10 @@ def __getitem__(self, name):
             return v._data()
         return v
 
+    def __lt__(self, other):
+        """ Comparison operator used mainly to sort a list of AttrDict by ID. """
+        return self['ID'] < other['ID']
```
nschneid: I don't think it's guaranteed that `AttrDict` instances have an ID. But that's only an issue if there is code that attempts to compare or sort them, so this is probably fine.

nltk/corpus/reader/framenet.py
```diff
@@ -2242,7 +2247,7 @@ def frame_relation_types(self):
         Obtain a list of frame relation types.
 
         >>> from nltk.corpus import framenet as fn
-        >>> frts = list(fn.frame_relation_types())
+        >>> frts = sorted(list(fn.frame_relation_types()))
```
nschneid: I think the `list()` is redundant if it's wrapped in `sorted()`.

nschneid (changes requested):
Looks good, just a minor simplification

stevenbird:
Thanks very much @nschneid

nltk/corpus/reader/framenet.py
alvations: Ah now I understand, the `AttrDict` is a generic class. I thought it was a specific object made for Frames.

alvations:
Thank you @nschneid for the review!

Specifying a `self["ID"]` in the generic `AttrDict` less than operator would have resulted in KeyError. I've changed the doctest to sort on key using the `operator.itemgetter`. 

I hope that's a better solution for standardizing the output of the doctests for framenet across Python versions.

nschneid (approved):

nschneid:
Makes sense to me. Thanks @alvations

stevenbird:
Thanks @alvations, @nschneid!

PR #1904, opened 2017-11-30
look for data in sys.prefix/share/nltk_data too

djsutherland:
It seems silly to look for `nltk_data` in `/usr/local/share/nltk_data` and in `os.path.join(sys.prefix, 'lib/nltk_data')` but not in `os.path.join(sys.prefix, 'share/nltk_data')`; imo `share` is the right place to put it. This change would look for data there too.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -95,6 +95,7 @@ if sys.platform.startswith('win'):
     path += [
         str(r'C:\nltk_data'), str(r'D:\nltk_data'), str(r'E:\nltk_data'),
         os.path.join(sys.prefix, str('nltk_data')),
+        os.path.join(sys.prefix, str('share'), str('nltk_data')),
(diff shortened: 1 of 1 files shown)


jakirkham:
Interestingly this came up [before]( https://github.com/nltk/nltk/pull/1816#issuecomment-327075327 ) and was decided against because Unix and Windows would differ. However it sounds like our intent in conda-forge is to actually use `share` for both. :)

Edit: IOW this should be a non-issue now.

jakirkham:
cc @danielfrg ;)

djsutherland:
Oh, interesting. It should definitely be consistent so the conda package can be `noarch`, but I didn't think about `share` being a weird directory on Windows. Still, conda paths are unixy enough that I think it makes sense.

stevenbird:
Good idea, thanks @dougalsutherland

Issue #1877, opened 2017-11-01
UnicodeEncodeError when chunking non-ASCII text


jgarbers:
I've encountered a UnicodeEncodeError when chunking text that includes non-ASCII characters, using Python 2.7.12 and NLTK 3.2.5. As a new user of NLTK, there may be some subtleties when working with Unicode that I'm not sensitive to yet, but here's a sample bit of code that demonstrates the problem:

```
import nltk

text = u"You may have heard of Ren\xe9 Descartes."
tagged = nltk.pos_tag(nltk.word_tokenize(text))
entities = nltk.chunk.ne_chunk(tagged)
print entities
```

When run, this code reports an error at the chunking line above, ending with the report
```
 File "/Users/jeffgarbers/.virtualenvs/nlplay/lib/python2.7/site-packages/nltk/chunk/named_entity.py", line 104, in _feature_detector   
  'word+nextpos': '{0}+{1}'.format(word.lower(), nextpos),
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 3: ordinal not in range(128)
```

It seems that the error goes away if the format strings in lines 104-106 of _named_entity.py_ are changed to Unicode (_u'word+nextpos'_, for example), although I don't know if such a fix has other implications. 

I'd much appreciate any suggestions anyone might offer. Thanks!

nschneid:
In Python 2.7, I think the issue is that `format()` is being called with non-ASCII Unicode arguments on a non-Unicode string. Should be `u'{0}+{1}'.format(...)`. (The current code is not a problem for Python 3.)

alvations:
@jgarbers Thank you for catching that!

@nschneid yes, `str.format()` is problematic for unicode in Python2

From https://docs.python.org/2/library/string.html#format-string-syntax , it seems like `format()` is implicitly having the `!s` [argument which should have been optional](https://github.com/python/cpython/blob/2.7/Lib/string.py#L616)

But in any case, `u'{}'.format()` resolves the problem =)

```python
>>> u'电脑'
u'\u7535\u8111'

>>> x = u'电脑'

>>> '{}'.format(x)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)

>>> u'{}'.format(x)
u'\u7535\u8111'
```

alvations:
[...]

PR #1906, opened 2017-12-03
Adds unicode support via __future__

Ch0ronomato:
Address the small issue in #1877 

Anything else I should do here? I may need some help with the unit testing framework going further.

I choose just to import __future__ as it seems the repo is moving towards support python 3. Of course, I can go through and change the `.format` calls to be `u'...'.format` if that's the preferred method.

### DIFF
--- a/nltk/chunk/named_entity.py
+++ b/nltk/chunk/named_entity.py
@@ -9,6 +9,7 @@
 Named entity chunker
 """
 from __future__ import print_function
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @IanSchweer. No need for unit testing for this.

ECR-FTC:
Thanks for the fix, everyone! How can I determine when a release including the fix is available through PyPI?

PR #1907, opened 2017-12-05
Adding better documentation for HMM training

### LINKED ISSUE #1277: Better documentation for HMM unsupervised training
I had some problems using unsupervised training for hmms.
The docs say:

```
        :param unlabeled_sequences: the unsupervised training data, a set of
            sequences of observations
        :type unlabeled_sequences: list
```

When i read this i thought about a List [ word_1, ..., word_n ] < a unlabeled sequence > but the correct way is [ (word_1, tag_1),...,(word_n,tag_n) ] < a labeled sequence imo > .
I think we should explain this better to avoid confusion.

accraze:
Adding basic examples for both parameters in the `train` method.

fixes #1277


stevenbird:
Thanks @accraze

PR #1909, opened 2017-12-07
Prevents zero divisions

gogochi:
empty text will crash the code.


stevenbird:
Thanks @gogochi. I think empty text should raise a ValueError.

PR #1910, opened 2017-12-08
New concordance_list() function for Text

alvations:
Following up on #1333


alvations:
This would provide a cleaner interface to saving concordance lines and printing concordance. 

```python
>>> from nltk.corpus import gutenberg
>>> corpus = gutenberg.words('melville-moby_dick.txt')
>>> text = Text(corpus)
>>> text.concordance('monstrous')
Displaying 11 of 11 matches:
ong the former , one was of a most monstrous size . ... This came towards us , 
ON OF THE PSALMS . " Touching that monstrous bulk of the whale or ork we have r
ll over with a heathenish array of monstrous clubs and spears . Some were thick
d as you gazed , and wondered what monstrous cannibal and savage could ever hav
that has survived the flood ; most monstrous and most mountainous ! That Himmal
they might scout at Moby Dick as a monstrous fable , or still worse and more de
th of Radney .'" CHAPTER 55 Of the Monstrous Pictures of Whales . I shall ere l
ing Scenes . In connexion with the monstrous pictures of whales , I am strongly
ere to enter upon those still more monstrous stories of them which are to be fo
ght have been rummaged out of this monstrous cabinet there is no telling . But 
of Whale - Bones ; for Whales of a monstrous size are oftentimes cast up dead u
>>> conc_list = text.concordance('monstrous', save=True)
>>> conc_list[0].line
'ong the former , one was of a most monstrous size . ... This came towards us , '
>>> conc_list[1].line
'ON OF THE PSALMS . " Touching that monstrous bulk of the whale or ork we have r'
```

There are also additional attributes to the `ConcordanceLine` namedtuple:


```python
>>> type(conc_list[0])
<class 'nltk.text.ConcordanceLine'>
>>> dir(conc_list[0])
['__add__', '__class__', '__contains__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getitem__', '__getnewargs__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__iter__', '__le__', '__len__', '__lt__', '__module__', '__mul__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__rmul__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '_asdict', '_fields', '_make', '_replace', '_source', 'count', 'index', 'left', 'left_print', 'line', 'offset', 'query', 'right', 'right_print']
>>> conc_list[0].left_print
'ong the former , one was of a most'
>>> conc_list[0].right_print
'size . ... This came towards us , '
>>> conc_list[0].left
['Whales', 'and', 'other', 'monsters', 'of', 'the', 'sea', ',', 'appeared', '.', 'Among', 'the', 'former', ',', 'one', 'was', 'of', 'a', 'most']
>>> conc_list[0].right
['size', '.', '...', 'This', 'came', 'towards', 'us', ',', 'open', '-', 'mouthed', ',', 'raising', 'the', 'waves', 'on', 'all', 'sides']
>>> conc_list[0].offset
899
```

nltk/text.py
```diff
@@ -30,6 +30,7 @@
 from nltk.collocations import BigramCollocationFinder
 from nltk.compat import python_2_unicode_compatible
 
+ConcordanceLine = namedtuple('ConcordanceLine', 'left query right offset left_print right_print line')
```
story645: Please make this a list:
```python
['left', 'query', 'right', 'offset', 'left_print', 'right_print line']
```

nltk/text.py
```diff
@@ -173,36 +173,52 @@ def __repr__(self):
...
+        :type save: bool
         """
         half_width = (width - len(word) - 2) // 2
-        context = width // 4 # approx number of words of context
+        context = width // 4  # approx number of words of context
 
+        # Find the instances of the word to create the ConcordanceLine
+        concordance_list = []
         offsets = self.offsets(word)
         if offsets:
```
story645: probably don't need the `if offsets` cause the loop just won't run if there are no offsets.

story645:
So this would supersede #1333 or does that need to go in first? (also, I'm assuming this PR means I no longer have to merge anything?)

nltk/text.py
alvations: I think list is not a good idea here. Following the "explicit is better than implicit" rule,  namedtuple are more appropriate, because ConcordanceLine is sort of a proper class object.

It also provides a more readable interface when users calls `ConcordanceLine.left` instead of `concordance_line[0]`. 

Also, a namedtuple can become a list by simple type casting in Python. We can leave it to the user to cast it into a list if they would like to, e.g.


```python
>>> from collections import namedtuple
>>> StarWars = namedtuple('StarWars', 'skywalker, lightsaber, pewpewpew')
>>> StarWars(skywalker='luke', lightsaber='blue', pewpewpew='noisy')
StarWars(skywalker='luke', lightsaber='blue', pewpewpew='noisy')
>>> x = StarWars(skywalker='luke', lightsaber='blue', pewpewpew='noisy')
>>> x.skywalker
'luke'
>>> list(x)
['luke', 'blue', 'noisy']
>>> list(x)[0]
'luke'
```

alvations:
@story645 if it's okay, lets continue the issue here since it's easier to read through the commits. We can also go back to #1333 if you prefer. As long as @stevenbird agrees with the updates and merges, I've no preference =)

nltk/text.py
story645: Sorry I wasn't clear here, I meant instantiate the namedtuple using a list:

```Python
namedtuple('ConcordanceLine', ['left', 'query', 'right', 'offset', 'left_print', 'right_print line'])
```
Because I think the spaces make it harder to distinguish the individual elements of the named tuple.

stevenbird:
Thanks for diving into this @alvations... I'm glad we can resolve it soon. My main concern is to not introduce the practice of changing the return type based on the value of a function parameter. I would prefer:

* retain the existing interface of `Text.concordance` (for backwards compatibility with the published book)
* add a new method that returns instead of prints the concordance data
* modify the implementation of `Text.concordance` to use this new method

What do you think?

alvations:
@stevenbird 

Internally, `Text.concordance`, uses the `ConcordanceIndex.print_concordance`, I think we can split the finding and printing of the concordance in `ConcordanceIndex`. 

On the user side, `Text.concordance()` would remain the same. And the new `Text.save_concordance()`  with the interface would save the concordance into a list. 

E.g.



```python
>>> from nltk import Text
>>> from nltk.corpus import gutenberg
>>> corpus = gutenberg.words('melville-moby_dick.txt')
>>> text = Text(corpus)
>>> text.concordance('monstrous')
Displaying 11 of 11 matches:
ong the former , one was of a most monstrous size . ... This came towards us , 
ON OF THE PSALMS . " Touching that monstrous bulk of the whale or ork we have r
ll over with a heathenish array of monstrous clubs and spears . Some were thick
d as you gazed , and wondered what monstrous cannibal and savage could ever hav
that has survived the flood ; most monstrous and most mountainous ! That Himmal
they might scout at Moby Dick as a monstrous fable , or still worse and more de
th of Radney .'" CHAPTER 55 Of the Monstrous Pictures of Whales . I shall ere l
ing Scenes . In connexion with the monstrous pictures of whales , I am strongly
ere to enter upon those still more monstrous stories of them which are to be fo
ght have been rummaged out of this monstrous cabinet there is no telling . But 
of Whale - Bones ; for Whales of a monstrous size are oftentimes cast up dead u
>>> conc_list = text.save_concordance('monstrous')
>>> conc_list[0].line
'ong the former , one was of a most monstrous size . ... This came towards us , '
>>> conc_list[1].line
'ON OF THE PSALMS . " Touching that monstrous bulk of the whale or ork we have r'
```

nltk/text.py
alvations: Ah yes, a list would be easier to read and maintain the code. Pardon the laziness to type the attributes in strings =)

I've made the changes accordingly.

stevenbird:
Thanks @alvations, this looks good.

story645:
Can the function be called `concordance_list()` instead since it's not really saving the concordance since there's no output?

alvations:
I've no preference for what it's called =)

But it does output a list of ConcordanceLine namedtuple.

Any preference @stevenbird?

story645:
Sorry, I mean there's no writing out to a file/buffer/etc.

stevenbird:
Right, there's no "saving" going on.

alvations:
I'm bad with names =) 
I've changed it to `concordance_list()`.

alvations:
[CI: retest]

stevenbird:
Thanks @alvations!

Issue #1805, opened 2017-08-12
Order of lemmas returned by derivationally_related_forms() of WordNet?


ymfa:
I use `derivationally_related_forms()` to get the derivationally related lemmas of a WordNet lemma. When I upgraded nltk from version 2.0.4 to 3.2.3, I realized that the lemmas returned by this function are no longer sorted as they were. Instead, the order could change from one Python session to another. Of course, I could sort the list afterwards to get a deterministic result, but I want to replicate the same order as I found in nltk 2 (which doesn't work on Python 3).

The following code is an example. Please save it as a script, and run it a few times to see if the output stays the same for each run. (Remember that for nltk 2, you need to change `synset.lemmas()` to `synset.lemmas` on the third line.)

    from nltk.corpus import wordnet as wn
    synset = wn.synsets('study', 'n')[1]
    lemma = synset.lemmas()[0]
    print(lemma.derivationally_related_forms())

The output from nltk 2 is always:
`[Lemma('study.v.02.study'), Lemma('study.v.05.study'), Lemma('bookish.s.01.studious'), Lemma('learn.v.04.study')]`

However, according to my observation, the output from nltk 3 is either:
`[Lemma('bookish.s.01.studious'), Lemma('study.v.05.study'), Lemma('study.v.02.study'), Lemma('learn.v.04.study')]`
Or:
`[Lemma('bookish.s.01.studious'), Lemma('study.v.02.study'), Lemma('learn.v.04.study'), Lemma('study.v.05.study')]`

Does anyone know how was the list ordered in nltk 2? Does the order have any linguistic significance? A temporary workaround that allows me to replicate the same order in nltk 3 would be appreciated.

alvations:
@ymfa Which version of Python3 and Python2 are you using? 

    python3 -c "import sys; print(sys.version)"
    python2 -c "import sys; print(sys.version)"

Try upgrading your Python version to 3.4 or 3.5.

Currently, NLTK supports Python2.7 and Python 3.4 and Python3.5. 
Python 3.6 should work (most of the time) too but regression tests might fail. 

On my Ubuntu machine, they return the same list (Python 2.7.12 and Python 3.5.2):

```
$ python2

Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from nltk.corpus import wordnet as wn

[...]

PR #1918, opened 2017-12-17
fix the ordering of derivationally_related_forms() of WordNet

ymfa:
This fixes issue https://github.com/nltk/nltk/issues/1805. The only change is that, I assume there is no particular reason to use a set to represent `_lemma_pointers`, and therefore changed it to a list. The original method of sorting the list before returning it didn't ensure the same ordering, as described in the issue.

The ordering of lemmas returned by derivationally_related_forms() is now consistent with the WordNet official website.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -298,11 +298,11 @@ class Lemma(_WordNetObject):
 
     def _related(self, relation_symbol):
         get_synset = self._wordnet_corpus_reader.synset_from_pos_and_offset
-        return sorted([
+        return [
             get_synset(pos, offset)._lemmas[lemma_index]
             for pos, offset, lemma_index
             in self._synset._lemma_pointers[self._name, relation_symbol]
-        ])
+        ]
 
     def count(self):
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @ymfa!
@alvations: suggesting you to review given your analysis of #1805, thanks.

alvations:
@ymfa Thank you for the PR!

This change will fix the issue of `_related()` lemmas being jumbled up. 
It *looks good to merge*. 

--- 

@stevenbird @fcbond Ideally, under the hood, I would love the wordnet interface to be a little more graphical since it's a graph like nature and less of a mix of various `defaultdict()` of things. But I think that's for another PR =)

alvations (approved):
The modified function checks out in the doctest https://github.com/nltk/nltk/blob/develop/nltk/test/wordnet.doctest#L158 

Partly because that lemma has only one derivationally related form =)

stevenbird:
@alvations: I agree with the sentiment, only so many methods return lists of vertices not individual vertices.
@ymfa: I'd welcome an extra doctest with a meaningful test please, if you can manage it.

PR #1919, opened 2017-12-18
Split CoNLL lines at single tabs

MrEbbinghaus:
CoNLL should be separated by a single tab.
Some CoNLL files (like the Hamburg Dependency Treebank) are treating words like "New York" as a single form with a space!
The CoNLLReader couldn't read those files.

CoNLL-X Specification defines TAB as the separator:
http://anthology.aclweb.org/W/W06/W06-2920.pdf
References:
HDT: https://corpora.uni-hamburg.de/hzsk/de/islandora/object/treebank:hdt

### DIFF
--- a/nltk/corpus/reader/conll.py
+++ b/nltk/corpus/reader/conll.py
@@ -70,7 +70,7 @@ class ConllCorpusReader(CorpusReader):
     def __init__(self, root, fileids, columntypes,
                  chunk_types=None, root_label='S', pos_in_tree=False,
                  srl_includes_roleset=True, encoding='utf8',
-                 tree_class=Tree, tagset=None):
+                 tree_class=Tree, tagset=None, separator=' '):
         for columntype in columntypes:
             if columntype not in self.COLUMN_TYPES:
                 raise ValueError('Bad column type %r' % columntype)
@@ -84,6 +84,7 @@ class ConllCorpusReader(CorpusReader):
         self._tree_class = tree_class
         CorpusReader.__init__(self, root, fileids, encoding)
         self._tagset = tagset
+        self.sep = separator
 
     #/////////////////////////////////////////////////////////////////
     # Data Access Methods
@@ -194,7 +195,7 @@ class ConllCorpusReader(CorpusReader):
             block = block.strip()
             if not block: continue
 
-            grid = [line.split('\t') for line in block.split('\n')]
+            grid = [line.split(self.sep) for line in block.split('\n')]
 
             # If there's a docstart row, then discard. ([xx] eventually it
             # would be good to actually use it)
--- a/nltk/corpus/reader/conll.py
+++ b/nltk/corpus/reader/conll.py
@@ -194,7 +194,7 @@ class ConllCorpusReader(CorpusReader):
             block = block.strip()
             if not block: continue
 
-            grid = [line.split() for line in block.split('\n')]
+            grid = [line.split('\t') for line in block.split('\n')]
 
             # If there's a docstart row, then discard. ([xx] eventually it
             # would be good to actually use it)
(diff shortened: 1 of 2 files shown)


dimazest:
Looks good. Could you add a test case and add references to the documentation.

alvations:
@MrOerni The automated are failing because not all corpora are split on `\t`. It might be better to either allow a `separator` argument when initializing the `ConllCorpusReader` object instead of hardcoding it at https://github.com/MrOerni/nltk/blob/develop/nltk/corpus/reader/conll.py#L197

alvations:
[CI: retest]

nltk/corpus/reader/conll.py
```diff
@@ -70,7 +70,7 @@ class ConllCorpusReader(CorpusReader):
     def __init__(self, root, fileids, columntypes,
                  chunk_types=None, root_label='S', pos_in_tree=False,
                  srl_includes_roleset=True, encoding='utf8',
-                 tree_class=Tree, tagset=None):
+                 tree_class=Tree, tagset=None, separator=' '):
```
jnothman: please use `'\t'` for readability

jnothman (commented):
It would be good, I think, if this had a non-regression test.

nltk/corpus/reader/conll.py
alvations: The `' '` here is a space =)

alvations: I'll change it to `\s` for readibility.

jnothman:
But aren't you meant to be splitting on tabs? Have I misunderstood?

jnothman:
Oh I see the discussion now. For backwards compatibility you should allow
for regex split.

MrEbbinghaus:
@alvations I changed the default value to `None`. 
With `.split('  ')`, the string is splited by the actual symbol `space`.

With `.split(None)`  the behavior is more like a regex `\s+`.

#### EDIT:
The first version `''` is wrong. `None` should it be. My mistake.

jnothman:
None would be clearer than ''

MrEbbinghaus:
@jnothman You are right.
`None` is the actual [default value](https://docs.python.org/3/library/stdtypes.html#str.split).

The `''` is actually wrong. I followed https://www.tutorialspoint.com/python/string_split.htm 😞

alvations:
Yes `None` is a good default

alvations:
[CI: retest]

alvations:
[CI: retest]

alvations:
@jnothman Thanks for the reviews and the suggestions! 

@MrOerni we had some hiccups on the automated testing. Now it's all cleared up. The tests should pass now =)

MrEbbinghaus:
@alvations I was starting to worry... 👍

alvations:
[CI: retest]

alvations:
[CI: retest]

nltk/test/corpus.doctest
```diff
@@ -1472,9 +1472,12 @@ constructor will first call its base class's constructor, and then
 store the customizable parameters.  For example, the
 `ConllChunkCorpusReader`\ 's constructor is defined as follows:
 
-    >>> def __init__(self, root, files, chunk_types):
-    ...     CorpusReader.__init__(self, root, files)
-    ...     self.chunk_types = tuple(chunk_types)
+    >>> def __init__(self, root, fileids, chunk_types, encoding='utf8',
+    ...          tagset=None, separator=None):
+    ... ConllCorpusReader.__init__(
+    ...     self, root, fileids, ('words', 'pos', 'chunk'),
```
alvations: The indentation is misaligned here.

alvations: Also I think creating a class object in the doctests isn't necessary to test this feature.

alvations:
The latest changes removed the need for regression test on the constructor in the `corpus.doctest`. 
That should clear the automated test.

alvations:
@MrOerni Sorry the automated test took some time to fix. Now it's all cleared up. 

It LGTM now.

alvations (approved):

stevenbird:
Thanks @MrOerni, @alvations, @jnothman, @dimazest

PR #1924, opened 2017-12-26
Make `CFG.fromstring` easy to inherit

iliakur:
A while back I wanted to subclass `nltk.grammar.CFG` and found out that I had to copy/paste the `fromstring` method in order to instantiate my custom subclass from a string even though my initialization arguments didn't differ from `CFG` at all.

This was annoying, so I've added a little tweak to all the grammars that allows one to inherit `fromstring` more easily.

### DIFF
--- a/nltk/grammar.py
+++ b/nltk/grammar.py
@@ -540,13 +540,13 @@ class CFG(object):
     @classmethod
     def fromstring(cls, input, encoding=None):
         """
-        Return the ``CFG`` corresponding to the input string(s).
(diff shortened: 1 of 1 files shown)


alvations:
@stevenbird I'm curious why the CI was not taking the normalize whitespace and ellipsis flags. 

Not sure why is this not kicking in =( https://github.com/nltk/nltk/blob/7e83998135a1931c01af743592487dae26100cb0/nltk/test/doctest_nose_plugin.py#L133

PR #1925, opened 2017-12-27
Prevents zero divisions by raising ValueError

gogochi:
empty text will crash the code.
fix the error by raising ValueError
if len(text) == 0: raise ValueError('Text is empty')


alvations:
@gogochi Thank you for suggesting the changes! But the PR is a little awkward. 

The LBYL idiom is something like this:
```python
try:
    # do something
except SomeError:
    # do something else
```

And the EAFP idiom is something like:

```python
if condition:
    # do something
else:
    # do something else
```

Mixing them like the code changes is a little strange. @brettcannon wrote a nice article on the idioms: https://blogs.msdn.microsoft.com/pythonengineering/2016/06/29/idiomatic-python-eafp-versus-lbyl/

@gogochi Could you change it to use either of the idioms and not a mix of both? 

Also, could you clean up the git tree and remove the un-necessary commits? E.g. 

```shell
git rebase -i HEAD~5
# Then drop the un-used commits 
```

Issue #1926, opened 2017-12-27
nist.py regexp breaks in py27


stevenbird:
Our CI server [reports](https://nltk.ci.cloudbees.com/job/nltk/lastCompletedBuild/TOXENV=py27-jenkins,jdk=jdk8latestOnlineInstall/testReport/nose.failure/Failure/runTest/) error `nothing to repeat` on line https://github.com/nltk/nltk/blob/develop/nltk/tokenize/nist.py#L110

@alvations: would you mind having a look at this please?

alvations:
Hmmm... It's a little strange that the error doesn't happen locally:

```python
>>> import re
>>> from nltk.corpus import perluniprops
>>> from six import text_type
>>> pup_symbol = text_type(''.join(set(perluniprops.chars('Symbol')))) 
>>> symbol_regex = re.sub(r'[]^\\-]', r'\\\g<0>', pup_symbol)
>>> re.compile(u"({s})".format(s=symbol_regex))
<_sre.SRE_Pattern object at 0x35d9930>
```

Let me take a closer look.

alvations:
Looks like there's a `[]` missing in the compiled regex but it still doesn't throw the error. 

```python
>>> sent = 'this is a foo' + pup_symbol[1] +'sentence.'
>>> patt =  re.compile(u"({s})".format(s=symbol_regex))
>>> patt.sub(' \\1 ', sent)
u'this is a foo\u2604sentence.'
>>> patt =  re.compile(u"([{s}])".format(s=symbol_regex))
>>> patt.sub(' \\1 ', sent)
u'this is a foo \u2604 sentence.'

```

alvations:
Argh, github.com is doing strange thing with bad wifi... Reverting the latest https://github.com/nltk/nltk/commit/052cc88295d122834b7a21e1c73fdceb23130d8f and doing a fresh PR.

PR #1927, opened 2017-12-27
Patching regex in #1926

alvations:


alvations:
@stevenbird This should patch the #1926 =)

stevenbird:
Great, thanks @alvations!

PR #1935, opened 2018-01-12
update entire library to pep8

wardbradt:


alvations:
@wardbradt thank you for the pull request. 

Just like to check so that we understand how the pep8-ify changes were made.
Are they detected automatically and then manually edited? Or automatically edited?

wardbradt:
They are automatically edited using [autopep8](https://pypi.python.org/pypi/autopep8).

nltk/app/collocations_app.py
```diff
@@ -29,44 +29,46 @@
...
+    'Portuguese: Floresta Corpus (Portugal)':
+    lambda: floresta.words(),
+    'Portuguese: MAC-MORPHO Corpus (Brazil)':
+    lambda: mac_morpho.words(),
+    'Portuguese: Machado Corpus (Brazil)':
+    lambda: machado.words(),
+    'Spanish: CESS-ESP Corpus':
+    lambda: cess_esp.words()
+}
+
```
stevenbird, line 68: I'm finding the above less readable than the original. So PEP-8 doesn't permit dictionary literals to have their values nested relative to their keys?

nltk/app/collocations_app.py
```diff
@@ -324,23 +340,27 @@ def run(self):
...
+                           (vocab[w1] * vocab[w2])) for w1, w2 in fd]
                 scored.sort(key=itemgetter(1), reverse=True)
                 self.model.collocations = list(map(itemgetter(0), scored))
                 self.model.queue.put(CORPUS_LOADED_EVENT)
             except Exception as e:
                 print(e)
                 self.model.queue.put(ERROR_LOADING_CORPUS_EVENT)
 
-#def collocations():
+# def collocations():
```
stevenbird, line 355: The reason for no space is that a commented-out code block can be easily commented-in.

nltk/cluster/kmeans.py
```diff
@@ -198,7 +202,8 @@ def demo():
     print('Means:', clusterer.means())
     print()
 
-    vectors = [numpy.array(f) for f in [[3, 3], [1, 2], [4, 2], [4, 0], [2, 3], [3, 1]]]
+    vectors = [numpy.array(f) for f in [[3, 3], [1, 2], [
+        4, 2], [4, 0], [2, 3], [3, 1]]]
```
stevenbird, line 206: Seriously?

nltk/corpus/reader/conll.py
```diff
@@ -58,7 +59,7 @@ class ConllCorpusReader(CorpusReader):
     CHUNK = 'chunk'   #: column type for chunk structures
     NE = 'ne'         #: column type for named entities
     SRL = 'srl'       #: column type for semantic role labels
-    IGNORE = 'ignore' #: column type for column that should be ignored
+    IGNORE = 'ignore'  # : column type for column that should be ignored
```
stevenbird, line 62: It's a shame the pep8 processor isn't able to detect a block of comments and line them up.

nltk/metrics/segmentation.py
```diff
@@ -82,18 +81,19 @@ def windowdiff(seg1, seg2, k, boundary="1", weighted=False):
     if len(seg1) != len(seg2):
         raise ValueError("Segmentations have unequal length")
     if k > len(seg1):
-        raise ValueError("Window width k should be smaller or equal than segmentation lengths")
+        raise ValueError(
+            "Window width k should be smaller or equal than segmentation lengths")
     wd = 0
     for i in range(len(seg1) - k + 1):
-        ndiff = abs(seg1[i:i+k].count(boundary) - seg2[i:i+k].count(boundary))
+        ndiff = abs(seg1[i:i + k].count(boundary) -
```
stevenbird, line 88: This is unfortunate; the grouping is `i:(i+k)` which means `i : i+k` is more readable than `i:i + k`

alvations:
@wardbradt do note these tips from too when PEP8-ing https://www.youtube.com/watch?v=wf-BqAjZb8M

nltk/cluster/kmeans.py
alvations: Beyond 80 chars is very subjective. But yeah, the auto-pep8 tool is making a bad change for this line ;P

psukys:
Agree  with @alvations . pep8 is a style guide for readability after all, and if we try to adhere to default pep8 standards to sacrifice current codebase's readability, then it beats the purpose.

stevenbird:
@psukys, @alvations: I think that ideally means we/someone goes through the recommendations and accepts/rejects them on a case-by-case basis.

psukys:
@stevenbird yes, I doubt that such PR is maintanable due to it's sheer size and variety of issues tackled. I would propose @wardbradt to PR [single autopep8 errors](https://pypi.python.org/pypi/autopep8) by issuing an argument `--select` with specific PEP8 rule defined.

frankier:
Note there's now [black](https://github.com/ambv/black) which explicitly has no options on the basis of preferring slightly suboptimal formatting which is totally regular on the basis that you can totally automate it -- inspired by gofmt and the like.

stevenbird:
Closing due to inactivity.

PR #1938, opened 2018-01-16
Adding new prefixes to be removed for arlstem.py

anati89:
Adding new prefixes to be removed for ArabicStemmer


anati89:
Adding new prefixes to be removed.

stevenbird:
Thanks @anati89

PR #1940, opened 2018-01-17
[WIP] Added a version agnostic python testenv

alvations:
In cloudbees, the config matrix would have different python version in each environment. And we only need one test per python version environment.


alvations:
[CI: retest]

alvations:
[CI: retest]

PR #1951, opened 2018-02-08
Evaluate type before comparing equality of WordNet and other objects

### LINKED ISSUE #1944: Wordnet: Comparison of  a _WordNetObject fails if other object is not a _WordNetObject.

See output log:
```bash 
    107 d = 0
    110         d += 1

    196 
    199 

```

vivekscl:
Fixes #1944 

As explained in the commit message,

When comparing the equality of WordNet objects with other objects, it is assumed that the other object is a WordNet object and hence the name attribute of the other object is compared with the name attribute of the WordNet object even though the other object might not be a WordNet object.

This will result in an AttributeError exception to be thrown in the case where the other object is not a WordNet object. For instance, when comparing a WordNet object with a string, since the string does not contain a name attribute, an exception is thrown.

To prevent this from happening, the type of the objects should be evaluated before checking for equality like so:
```python
def __eq__(self, other):
        return type(self) == type(other) and self._name == other._name

    def __ne__(self, other):
        return not self.__eq__(other)

    def __lt__(self, other):
        return type(self) == type(other) and self._name < other._name
```

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -195,13 +195,20 @@ class _WordNetObject(object):
         return hash(self._name)
 
     def __eq__(self, other):
-        return self._name == other._name
+        try:
+            return self.name() == other.name()
+        except (AttributeError, TypeError):
+            return False
+
 
     def __ne__(self, other):
-        return self._name != other._name
+        return not self.__eq__(other)
 
     def __lt__(self, other):
-        return self._name < other._name
+        try:
+            return self.name() < other.name()
+        except (AttributeError, TypeError):
+            return False
 
 
 @python_2_unicode_compatible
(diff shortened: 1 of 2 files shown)


iliakur:
imho, this goes against the duck-typing philosophy that Python embraces.

vivekscl:
>imho, this goes against the duck-typing philosophy that Python embraces.

Correct me if I'm wrong, but by duck-typing, you mean that we should not check the whether the object is of a certain type but whether the **object behaves like the object that has that specific attribute or method** right?  For instance, using the duck example, `don’t check whether it IS-a duck: check whether it QUACKS-like-a duck and WALKS-like-a duck`.

In this case, this code would violate duck-typing.
```python
type(self) == type(other)
```

So we should find a workaround such as:
```python
def __eq__(self, other):
        try:
            self._name == other._name
        except TypeError:
            # deal with error
```

iliakur:
Yup! Also bear in mind that `other._name` starts with an underscore, so it's not a "public" attribute, i.e. I wouldn't in the first place even rely on it being present. Are there attributes of the WordNet object that don't start with an underscore?

vivekscl:
>Are there attributes of the WordNet object that don't start with an underscore?

In the WordNet object, all fields start with an underscore, but there are methods that return these fields. So we should use them instead?

vivekscl:
@stevenbird  I made new commits for this pull request. But since this was closed, may I know if there were any errors in the pull request? :smile:

Jay-Patankar:
@stevenbird can I take this issue if its still open.

Issue #1954, opened 2018-02-12
Casual tokeniser allows newline in elipses and phone number tokens


drevicko:
This may be intended behaviour, but if so, it would be good to point it out in the docs as it is ONLY possible in these two cases and it is a common assumption that tokens do not contain newline characters.

It may be appropriate to disallow newlines in these tokens. This can be done by, [for example](https://stackoverflow.com/a/3469155/420867), using `[^\S\r\n]` in place of `\s` in the relevant regexes.

It is due to `\s` appearing in the regexes for [elipses](https://github.com/nltk/nltk/blob/develop/nltk/tokenize/casual.py#L168) and [phone numbers](https://github.com/nltk/nltk/blob/develop/nltk/tokenize/casual.py#L130).

alvations:
I think something like there's more things that `\s` represent. To capture a single space, it would be `[^\S\t\n\r\f\v]` or simply space ` ` =)

We could simply use `(?:\.(?: *\.){1,})`  for elipses instead of `(?:\.(?:[^\S\t\n\r\f\v]`*\.){1,})`. 

But I think of why `\s` is helpful in tweets, e.g. people who use "vertical elipsis"

```
. 
. 
. 
```

I'm not sure which is more appropriate since I don't work with NUTs (user generated text) as much.

BatMrE:
Hi @alvations seems like its open for a quite while, if its fixed can you close it else is there anything I can get involved with contributing ?

PR #1960, opened 2018-02-17
Small fix to tokenizer

anlandu:
Per [issue #1954](https://github.com/nltk/nltk/issues/1954), fixed treatment of strings that resemble phone numbers but contain newlines so that they're no longer recognized as phone numbers. Added unit test to ensure correct functionality. Left ellipses alone, as the comments seemed ambivalent about whether to change that.
(I'm fairly new to contributing, so any and all feedback is welcome!)


nltk/test/unit/test_tokenize.py
```diff
@@ -54,6 +54,25 @@ def test_stanford_segmenter_chinese(self):
...
+        expected = ['(393)  928 -3010']
+        result = tokenizer.tokenize(test1)
+        self.assertEqual(result, expected)
+
+        # Due to newline, first three elements aren't part of a phone number;
+        # fourth is
+        test2= "(393)\n928 -3010"
+        expected = ['(', '393', ')', "928 -3010"]
+        result = tokenizer.tokenize(test2)
+        self.assertEqual(result, expected)
```
alvations, line 74: :+1:

stevenbird:
Thanks @anlandu

Issue #1915, opened 2017-12-13
eval() in DependencyEvaluator returns LAS and UAS in wrong order


nputikhin:
According to documentation, return order is Labeled Attachment Score (LAS) and then Unlabeled Attachment Score (UAS). 
UAS is computed based on correct number of HEAD fields only, while LAS is based both on HEAD and DEPREL fields.
Actual code returns UAS as the first item and LAS as the second:
https://github.com/nltk/nltk/blob/8490d9d8ef58482d4c3d17826b4b5224e421714e/nltk/parse/evaluate.py#L121-L127

PR #1965, opened 2018-02-26
Fixed issue #1915, where UAS and LAS were swapped

anlandu:
Per [this issue](https://github.com/nltk/nltk/issues/1915), swapped UAS and LAS output. Am fairly new to contributing, so any and all feedback welcome (let me know if you'd like a unit test!)


stevenbird:
[CI: retest]

alvations:
I'm not exactly sure whether it's the computing of the labelled vs unlabelled is correctly swapped here but it makes some sense that the unlabelled score will be higher or equal than the labelled one (in this PR) and not the other way (before the patch).

stevenbird:
Thanks @anlandu

Issue #1962, opened 2018-02-20
Better CONLLCorpusReader


alvations:
Alexis has an improved version of the CONLL corpus reader on https://stackoverflow.com/a/46875652/610569

It'll be good if we incorporate these changes by either:

 - retaining the name `BetterConllCorpusReader` and inherit from `ConllCorpusReader`
 - refactoring the `ConllCorpusReader` and remove the hard-coded columns and a more flexible `self._get_iob_words`

jaehoonhwang:
Hi, can I take this one?

alvations:
@jaehoonhwang sure! Create a pull-request once it's ready and there'll be some review and it'll be merge once the test pass and it's good to go =)

jaehoonhwang:
I have implemented using the first route: creating another class named `BetterConllCCorpusReader`.
Is there a way that I can get this tested?

jaehoonhwang:
@alvations Just waiting for suggestions.

alvations:
@jaehoonhwang Sorry been a little busy for a while. I've tried to ask alexis from SO to help look at the PR, lets wait a little =)

PR #1966, opened 2018-02-26
BetterConllCorpusReader

jaehoonhwang:
Issue: #1962 

I've implemented according to https://stackoverflow.com/a/46875652/610569


alvations:
[CI: retest]

alexisdimi:
Sorry to put it this way but this is a really naive implementation and should be immediately rejected. The SO answer you link to (written by myself) shows how to get the job done without modifying the nltk codebase. In a patch request you need to modify `CONLLCorpusReader` itself, not derive a subclass!

Also, as my answer points out it is trivial to make the code accept multiple columns, and this should be done as well. (I left it out of the answer because the question did not need it.) This is how:

```
def _get_iob_words(self, grid, tagset=None, columns="chunk"):
        if isinstance(columns, str):
            columns = [ columns]
        pos_tags = self._get_column(grid, self._colmap['pos'])
        if tagset and tagset != self._tagset:
            pos_tags = [map_tag(self._tagset, tagset, t) for t in pos_tags]
        return list(zip(self._get_column(grid, self._colmap['words']), pos_tags,
                   *list(self._get_column(grid, self._colmap[c]) for c in columns)))
```
I.e. I changed `column` to `columns`, and modified the rest of the code to expect a list but accept a single argument as a special case. This matches the standard nltk handling, e.g., of `fileids` or `categories`. I haven't tested the code-- you can do that and submit a new pull request.

`iob_words()` only needs to rename `column` to `columns`, since all it does is pass this argument to `_get_iob_words()`.

**Correction:** The syntax `*list(...)` is not backward-compatible (I think it is first supported in Python 3.6), and should be replaced by an explicit way of collecting the arguments to return.

alexisdimi:
PS @alvations, what were you thinking? You didn't need me to spell all this out.

alvations:
@alexisdimi sorry, I didn't look at the PR carefully. I think it's still worth extending the current CoNLL reader to support user defined columns built upon the solution in SO. But yes, the code in this PR needs more work.

jaehoonhwang:
Hi, I have updated the code with Alexis input.
I am getting errors on my Windows machine when I try to use `iob_words()` 
I will test on a different machine and see if I get the same error.
Also, I will redo this push, since I deleted the wrong line.
Thank you for the help!

jaehoonhwang:
I did some test with the original question using this corpus.

```txt
WORD      POS   CHUNK NE
U.N.      NNP   I-NP  I-ORG
official  NN    I-NP  O
Ekeus     NNP   I-NP  I-PER
heads     VBZ   I-VP  O
for       IN    I-PP  O
Baghdad   NNP   I-NP  I-LOC
.         .     O     O
```

Result:
```py
>>> from nltk.corpus.reader import ConllCorpusReader
>>> a = ConllCorpusReader(r'/Users/jae/Documents/nltk_test/', r'test.txt', ('words', 'pos', 'chunk', 'ne'))
>>> a.iob_words()
[(u'WORD', u'POS', u'CHUNK'), (u'U.N.', u'NNP', u'I-NP'), ...]
>>> a._get_iob_words(a._grids())
[([u'WORD', u'POS', u'CHUNK', u'NE'], [u'U.N.', u'NNP', u'I-NP', u'I-ORG'], [u'official', u'NN', u'I-NP', u'O'])]
>>> a._get_iob_words(a._grids(), columns=['ne'])
[([u'WORD', u'POS', u'CHUNK', u'NE'], [u'U.N.', u'NNP', u'I-NP', u'I-ORG'], [u'Ekeus', u'NNP', u'I-NP', u'I-PER'])]
>>> a._get_iob_words(a._grids(), columns=['chunk', 'ne'])
[([u'WORD', u'POS', u'CHUNK', u'NE'], [u'U.N.', u'NNP', u'I-NP', u'I-ORG'], [u'official', u'NN', u'I-NP', u'O'], [u'Ekeus', u'NNP', u'I-NP', u'I-PER'])]
```

Is this expected behavior for this?
Thank you.

nltk/corpus/reader/conll.py
```diff
@@ -224,12 +224,12 @@ def _get_tagged_words(self, grid, tagset=None):
             pos_tags = [map_tag(self._tagset, tagset, t) for t in pos_tags]
         return list(zip(self._get_column(grid, self._colmap['words']), pos_tags))
 
-    def _get_iob_words(self, grid, tagset=None):
+    def _get_iob_words(self, grid, tagset=None, columns=['chunk']):
```
dimazest: It's considered a bad practice to use mutable objects as function defaults (it usually leads to nasty bugs that are difficult to fix.

Use either a tuple (a sequence that can't be changed):
```python
def _get_iob_words(self, grid, tagset=None, columns=('chunk', )):
    ...
```
or
```python
def _get_iob_words(self, grid, tagset=None, columns=None):

    if columns is None:
        columns = ['chunk']

    ...
```

The reason mutable defaults lead to bugs is that they are created when the function is created, and persist between function calls. If you happen to modify the list of `columns`, for example, by adding an element to it, then the next time the function is called `columns` will include that element.

dimazest:
`*list(...)` should work on Python 2.7:
```
Python 2.7.13 (default, Dec 18 2016, 05:36:03)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
>>> def f(*args):
...     print args
...
...
>>> f(*list('abcdefg'))
>>> ('a', 'b', 'c', 'd', 'e', 'f', 'g')
```

nltk/corpus/reader/conll.py
jaehoonhwang: I used something similar to the second method.
Thank you for your comment.
If you have time, check the new commit.

jaehoonhwang:
Apparently, string literals are `unicode` type. 
```py
def _get_iob_words(self, grid, tagset=None, columns='chunk'):
    ...
    print(type(columns)) # => <type 'unicode'>
```
So, rather than checking if `column` argument is `unicode`, I did check on `list` instead.

Thank you for feedback @dimazest

nltk/corpus/reader/conll.py
```diff
@@ -224,12 +224,13 @@ def _get_tagged_words(self, grid, tagset=None):
             pos_tags = [map_tag(self._tagset, tagset, t) for t in pos_tags]
         return list(zip(self._get_column(grid, self._colmap['words']), pos_tags))
 
-    def _get_iob_words(self, grid, tagset=None):
+    def _get_iob_words(self, grid, tagset=None, columns='chunk'):
+        columns = columns if isinstance(columns, list) else [columns]
```
dimazest: I would suggest to set `columns` to `None` rather then checking whether a list is passed, what if a `tuple` is passed or a `set` or any other `iterable`?

Also, this is a private method, so taking care that an argument could be an item or a list is not necessary. (It might be a good idea for a public API.)

I would prefer a clear solution that `columns` is an iterable.

jaehoonhwang: Updated!

jaehoonhwang:
No need to be sorry.
Fixed.

alexisdimi:
@dimazest wrote:

> I would suggest to set columns to None rather then checking whether a list is passed, what if a tuple is passed or a set or any other iterable?
>
> Also, this is a private method, so taking care that an argument could be an item or a list is not necessary. (It might be a good idea for a public API.)

Many nltk methods accept a single string or a list of them, e.g. the `fileids` argument of corpus readers. This was all handled properly until Jaehoon got rid of my code, which set `columns="chunk"` as default and then checked for `str`. But better yet: use `six.string_types` to catch str & unicode on Python 2:

    from six import string_types

    if isinstance(columns, string_types):
        columns = [ columns ]

alexisdimi:
Anyway I take it back: Handling more than one extra column correctly is evidently not trivial. Maybe just integrate a (working) patch for one extra column, and the general case can be dealt with later.

jaehoonhwang:
So something along the lines of: 

```py
def _get_iob_words(self, grid, tagset=None, column='chunk'):
          pos_tags = self._get_column(grid, self._colmap['pos'])
          if tagset and tagset != self._tagset:
              pos_tags = [map_tag(self._tagset, tagset, t) for t in pos_tags]
          return list(zip(self._get_column(grid, self._colmap['words']), pos_tags,
               self._get_column(grid, self._colmap[column]))
```

This would allow the single custom column, but following your suggestion for checking types.

```py
from six import string_types

...

def _get_iob_words(self, grid, tagset=None, columns='chunk'):
          columns = [columns] if isinstance(columns, string_types) else columns
          pos_tags = self._get_column(grid, self._colmap['pos'])
          if tagset and tagset != self._tagset:
              pos_tags = [map_tag(self._tagset, tagset, t) for t in pos_tags]
          return list(zip(self._get_column(grid, self._colmap['words']), pos_tags,
                *list(self._get_column(grid, self._colmap[c]) for c in columns)))
```

jaehoonhwang:
The reason why I got rid of following code line, `if isinstance(columns, str):` is that, in Python 2.7.13, the code will return `False` for unicode. But, in Python 3.x, they would evaluate `unicode` as a `str` instance which the code would return as `True`.

**Python 2** 
```py
Python 2.7.13 |Anaconda custom (x86_64)| (default, Dec 20 2016, 23:05:08)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> a = "abc"
>>> b = u"abc"
>>> isinstance(a, str)
True
>>> isinstance(b, str)
False
```

**Python 3**
```py
Python 3.6.4 (default, Mar  5 2018, 21:43:20)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)] on darwin
>>> a = 'abc'
>>> b = u'abc'
>>> isinstance(a, str)
True
>>> isinstance(b, str)
True
```

alexisdimi:
I understand about `str` vs. `unicode`. The solution is in my previous comment, and uses `six` -- like all nltk code does for this purpose.

jaehoonhwang:
Ah, okay. So, do I one with this string check or just one custom column?

purificant:
https://github.com/nltk/nltk/pull/2163 reformatted the return statement and expanded it into multiple lines. To resolve the conflict in this PR please take the code from `develop` and replace `self._get_column(grid, self._colmap['chunk']),` with `*list(self._get_column(grid, self._colmap[c]) for c in columns)`

jaehoonhwang:
I'll just close this request, since it seems like this is dead.

Issue #1756, opened 2017-06-20
Fulfilling the requirements of the Treebank Tokenizer License


oxinabox:
I am concerned about the Treebank Tokenizer's use, by its License terms.

The source for the [tokeniser (via waybackmachine)](https://web.archive.org/web/20130804202913/http://www.cis.upenn.edu/~treebank/tokenizer.sed) includes the text:

> by Robert MacIntyre, University of Pennsylvania, late 1995.  
> If this wasn't such a trivial program, I'd include all that stuff about
 no warrantee, free use, etc. from the GNU General Public License.  If you
 want to be picky, assume that all of its terms apply.  Okay?

This is really problematic, AFAICT.
It is an awful crayon license, 
neither its meaning nor its precise intent is clear.
Under a literal interpretation, one is most restricted when not being picky,
and still very restricted if not being picky

I asked about it on [OpenSource StackExchange](https://opensource.stackexchange.com/q/5612/934)

> As I see it, one has two choices:
>  - **Do not be picky:** then this has no license and so can not be incorporated into any open source project. All rights remain with Robert MacIntyre.
>  - **Be picky:** this is *effectively* under *some* version of the GPL, it can be incorporated only into things that can incorporate GPL works. Since this work is from 1995, it means it is either GPL 1, or GPL 2, is intended.

As far as I can tell, that reasoning is correct, and applies to NLTK's usage.
As NLTK is Apache 2 Licensed, incorporating GPL works is not (normally considered) possible. 

I can't find any further reference to the tokenizer or to Robert MacIntyre. [LICENSE](https://github.com/nltk/nltk/blob/develop/LICENSE.txt)  nor in [AUTHORS.md](https://github.com/nltk/nltk/blob/develop/AUTHORS.md).

nschneid:
As a layperson not extensively familiar with licensing, it sounds like he's saying this program is so simple that he doesn't feel the need to bother with legal language, and anyone is free to use the software so long as they don't sue him.

But perhaps contacting him would clarify things: https://www.linkedin.com/in/robert-macintyre-07330280

stevenbird:
Thanks @oxinabox and @nschneid. I'm trying to contact Robert MacIntyre.

oxinabox:
@stevenbird  any update on this?
I'ld like to use the code in my own projects, but not while the licensing is up in the air.

dseddah:
seems to me that there're two possible licenses here

# If this wasn't such a trivial program, I'd include all that stuff about
[...]

PR #1999, opened 2018-04-10
Clarify comment

oxinabox:
List of tokens is not adapted from Robert MacIntyre's tokenizer. It *is* from then tokeniser -- it is identical AFAICT.
Compare to https://web.archive.org/web/20130804202913/http://www.cis.upenn.edu/%7Etreebank/tokenizer.sed

### DIFF
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -23,7 +23,7 @@ from nltk.tokenize.util import align_tokens
 
 class MacIntyreContractions:
     """
-    List of contractions adapted from Robert MacIntyre's tokenizer.
+    List of contractions from Robert MacIntyre's tokenizer.
     """
     CONTRACTIONS2 = [r"(?i)\b(can)(?#X)(not)\b",
                      r"(?i)\b(d)(?#X)('ye)\b",


alvations:
@oxinabox just to clarify the regexes are adapted the list of contractions isn't thus the "adapated from ..." =)

oxinabox:
They are equivalent regexes though, right? I am trying to determine the changes from the original. And afaict  the contractions are the same. The differences are in the punctuation and openning/ending handling.

I ported the NLTK changes back to sed, as part of porting it to Julia in github.com/JuliaText/WordTokenizers.jl
And want to be confidant I spotted them all.

alvations:
I think they're not exactly the same regexes, in NLTK there's the `(?#X)` comment part of the regexes that was added so that the detokenization can work at https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L262 

(Note: the original sed treebank tokenizer doesn't have any detokenization functionalities)

The contractions was taken out of the `TreebankWordTokenizer()` object because it can be reused by the `TreebankWordDetokenizer()` so it's "adapted" =)

But it's just documentation semantics that has little effect on the in/output. 

As for licensing matters, we still can't find Robert MacIntyre for #1756

----

In terms of punctuation, it's a different issue from the contractions, it's abstracted out of the `TreebankTokenizer` object and injected at https://github.com/nltk/nltk/blob/develop/nltk/tokenize/__init__.py#L100 

So if the Julia port is a Treebank tokenizer, it should be a port without the hacks in the `word_tokenize()` function.

oxinabox:
There are other changes beyond the ones in https://github.com/nltk/nltk/blob/develop/nltk/tokenize/__init__.py#L100
to punctuation handling.
The original basically does not preserve any punctuation -- it is all mapped to `&`.
Edit: no I am wrong about that sed handles & as a special character, so the original does preserve punctuation

Still there are changes to punctuation handling I think (though I don't have a test case infront of me).

oxinabox:
The contraction  regexs do indeed function differently to the orginals.
Eg orginal fails to split the second cannot in 

"I cannot cannot work under these conditions!"

But I feel like this PR isn't on the right path, really a more extensive PR to document the changes is what is wanted. and idk if it is worth the effort

PR #2002, opened 2018-04-11
Fix nonescaped openning bracket in  stating quote

oxinabox:
The regex had an unbalanced opening quote,
because (I believe) it meant the literal openning quote.

I am surprised `re.compile` don't error at that.
Maybe I am missing something, though.

This has been in the code in one form or another since
https://github.com/nltk/nltk/commit/e660a6827d44e242b4114abba7fc30fef6b0125c

### DIFF
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -67,7 +67,7 @@ class TreebankWordTokenizer(TokenizerI):
     STARTING_QUOTES = [
         (re.compile(r'^\"'), r'``'),
         (re.compile(r'(``)'), r' \1 '),
-        (re.compile(r"([ (\[{<])(\"|\'{2})"), r'\1 `` '),
+        (re.compile(r"([ \(\[{<])(\"|\'{2})"), r'\1 `` '),
     ]
 
     # punctuation


alvations:
Actually, the regex would work without escaping the characters in the 1st capturing group, https://regex101.com/r/hw2ipk/1. I think the python regex interpreter isn't wrong in allow unescaped characters inside the single character set matches, i.e. `[...]`, thus ignoring the opening bracket as a regex catching group. 

So the regex would still work correctly without the changes but the changes made in this PR would make the round opening bracket consistent with the rest of the escaped characters before the double opening quote.

nltk/tokenize/treebank.py
```diff
@@ -67,7 +67,7 @@ class TreebankWordTokenizer(TokenizerI):
     STARTING_QUOTES = [
         (re.compile(r'^\"'), r'``'),
         (re.compile(r'(``)'), r' \1 '),
-        (re.compile(r"([ (\[{<])(\"|\'{2})"), r'\1 `` '),
+        (re.compile(r"([ \(\[{<])(\"|\'{2})"), r'\1 `` '),
```
alvations, line 70: The changes made in this regex would make the round opening bracket consistent with the rest of the escaped characters before the double opening quote.

alvations (approved):

oxinabox:
> I think the python regex interpreter isn't wrong in allow unescaped characters inside the single character set matches, i.e. [...], thus ignoring the opening bracket as a regex catching group.

Interesting, it seems like all common regex engines except GNU sed  with `-r` can handle this.
Still  indeed I find escaping it more readable

sed without `-r` can handle it. It is in the original even
`s=\([ ([{<]\)"=\1 `` =g`

stevenbird:
I agree that the escaped version is more readable. Thanks @oxinabox

PR #2007, opened 2018-04-23
Install matplotlib explicitly after pip install -r

alvations:
Removed the usage of `--allow-external`, instead just install in a separate line in the shell script.

```
pip install -U -r pip-req.txt
pip install -U matplotlib
```

I think putting it in a separate line is clearer than the alternative to repeat the dependency of the library as itself, something like:

```
pip install -U -r pip-req.txt —allow-external matplotlib matplotlib
```

That should clear up the problem on CI https://nltk.ci.cloudbees.com/job/pull_request_tests/PYV=2.7.14,jdk=jdk8latestOnlineInstall/669/console

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -3,7 +3,8 @@
 cd `dirname $0`
 
 #download nltk python dependencies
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @alvations

Issue #2024, opened 2018-05-21
Text.concordance does not take into account "width" argument and shows maximum of 25 lines (no matter "lines" argument value)


Coykto:
Just started learning nlp with nltk book.
I don't know why, but i'm having very apparent problems nobody seems to talk about (google search gave me nothing).

From the beginning of the book (1.3 Searching Text):
`text1.concordance("monstrous")`
Gives me 
```
Displaying 11 of 11 matches:
ong the former , one was of a most monstrous size . ... This came towards us ,
ON OF THE PSALMS . " Touching that monstrous bulk of the whale or ork we have r
ll over with a heathenish array of monstrous clubs and spears . Some were thick
d as you gazed , and wondered what monstrous cannibal and savage could ever hav
that has survived the flood ; most monstrous and most mountainous ! That Himmal
they might scout at Moby Dick as a monstrous fable , or still worse and more de
th of Radney .'" CHAPTER 55 Of the Monstrous Pictures of Whales . I shall ere l
ing Scenes . In connexion with the monstrous pictures of whales , I am strongly
ere to enter upon those still more monstrous stories of them which are to be fo
ght have been rummaged out of this monstrous cabinet there is no telling . But
of Whale - Bones ; for Whales of a monstrous size are oftentimes cast up dead u
```

As it should. But
`text1.concordance("whale")`
Gives me:
```
Displaying 25 of 25 matches:
s , and to teach them by what name a whale - fish is to be called in our tongue
t which is not true ." -- HACKLUYT " WHALE . ... Sw . and Dan . HVAL . This ani
ulted ." -- WEBSTER ' S DICTIONARY " WHALE . ... It is more immediately from th
ISH . WAL , DUTCH . HWAL , SWEDISH . WHALE , ICELANDIC . WHALE , ENGLISH . BALE
HWAL , SWEDISH . WHALE , ICELANDIC . WHALE , ENGLISH . BALEINE , FRENCH . BALLE
least , take the higgledy - piggledy whale statements , however authentic , in
 dreadful gulf of this monster ' s ( whale ' s ) mouth , are immediately lost a
 patient Job ." -- RABELAIS . " This whale ' s liver was two cartloads ." -- ST
 Touching that monstrous bulk of the whale or ork we have received nothing cert
 of oil will be extracted out of one whale ." -- IBID . " HISTORY OF LIFE AND D
ise ." -- KING HENRY . " Very like a whale ." -- HAMLET . " Which to secure , n
[...]

PR #2026, opened 2018-05-21
Fixed print_concordance()

Coykto:
by changing return of the find_concordance() function. It now returns full (not trimmed by "lines" argument value) concordance_list. Number of lines printed by print_concordance() is now determined inside that function.

### DIFF
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -376,7 +376,7 @@ class Text(object):
         """
         if '_concordance_index' not in self.__dict__:
             self._concordance_index = ConcordanceIndex(self.tokens, key=lambda s:s.lower())
-        return self._concordance_index.find_concordance(word, width)
+        return self._concordance_index.find_concordance(word, width)[:lines]
 
     def collocations(self, num=20, window_size=2):
         """
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -201,7 +201,7 @@ class ConcordanceIndex(object):
                                                     right_context, i,
                                                     left_print, right_print, line_print)
                 concordance_list.append(concordance_line)
-        return concordance_list[:lines]
+        return concordance_list
 
     def print_concordance(self, word, width=80, lines=25):
         """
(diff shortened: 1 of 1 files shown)


Coykto:
Fix for #2024 issue.

nltk/text.py
```diff
@@ -215,7 +215,7 @@ def print_concordance(self, word, width=80, lines=25):
         :param save: The option to save the concordance.
         :type save: bool
         """
-        concordance_list = self.find_concordance(word, width=80, lines=25)
+        concordance_list = self.find_concordance(word, width=width, lines=lines)
```
alvations: :+1:

nltk/text.py
```diff
@@ -201,7 +201,7 @@ def find_concordance(self, word, width=80, lines=25):
                                                     right_context, i,
                                                     left_print, right_print, line_print)
                 concordance_list.append(concordance_line)
-        return concordance_list[:lines]
+        return concordance_list
```
alvations, line 204: The expected behavior should be `concordance_list[:lines]`

alvations: Ah, my bad, it's a bad function design. We should let this function return all possible concordances so that the print_concordance knows the number.

alvations:
Thanks Coykto for catch the bug! 

There was some mistake in the function design to accommodate the `concordance_list()` feature. 

On the user-end, nothing changed but internally in `text.py`:

 - `ConcordanceIndex.find_concordance` finds all possible concordance lines
 - `ConcordanceIndex.print_concordance` calls `ConcordanceIndex.find_concordance` and then restricts the lines and print out the requested no. of lines
 - `text.concordance_list` calls `ConcordanceIndex.find_concordance`  and then restricts the lines and keeps only the requested no. of lines.
 - `text.concordance` wraps around the `ConcordanceIndex.print_concordance` and prints out the requested no. of lines

alvations (approved):

stevenbird:
Thanks for sorting this out, @Coykto, @alvations!

PR #2031, opened 2018-05-30
Optional argument to pipe error message for nltk.download()

alvations:
Added argument to specify where the error messages for download() should print to.


stevenbird:
Thanks @alvations

PR #2037, opened 2018-06-07
add span tokenizers to standard api

gituser768:
In a project of mine I had to duplicate this file to get convenient access to the span tokenizers. I figured it made sense to make them available at the same level as the regular tokenizers.


alvations:
`span_tokenize` is actually quite buggy so I think it's not the best to exposure it like `word_tokenize`; at least not until the kinks with spruious spaces and imbalance quotes are worked out for `span_tokenize`

gituser768:
Are you aware of any issues other than https://github.com/nltk/nltk/issues/1750 ?

alvations:
Additional spaces and quotes are the main problems but I'm sure there's still other issues since the original during gets `re.sub` without recording offset.

PR #2039, opened 2018-06-08
Added Cistem Stemmer for German

LeonieWeissweiler:
Adds a new stemmer for German, introduced in  _Leonie Weissweiler, Alexander Fraser (2017). Developing a Stemmer for German Based on a Comparative Analysis of Publicly Available Stemmers. In Proceedings of the German Society for Computational Linguistics and Language Technology (GSCL)_
to be found [here](https://link.springer.com/content/pdf/10.1007%2F978-3-319-73706-5_8.pdf ).

In the paper, we found the CISTEM stemmer to be much faster and better performing than the Snowball stemmer, currently the only German stemmer made available here. This is why we think CISTEM would be a valuable addition to nltk/stem.


nltk/stem/cistem.py
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/python3
```
stevenbird: NB NLTK still supports Python 2.7. Please omit this line.

nltk/stem/cistem.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def __init__(self, case_insensitive=False):
+        self._case_insensitive = case_insensitive
+        self._strip_ge = re.compile(r"^ge(.{4,})")
+        self._repl_xx = re.compile(r"(.)\1")
+        self._repl_xx_back = re.compile(r"(.)\*")
+        self._strip_emr = re.compile(r"e[mr]$")
+        self._strip_nd = re.compile(r"nd$")
+        self._strip_t = re.compile(r"t$")
+        self._strip_esn = re.compile(r"[esn]$")
```
stevenbird: I think these should be class attributes, constants defined at the level of the class, not the instance

nltk/stem/cistem.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def stem(self, word):
+        """
+        This method takes the word to be stemmed and a boolean specifiying if
+        case-insensitive stemming should be used and returns the stemmed word. If only
```
stevenbird: what boolean?

nltk/stem/cistem.py
```diff
@@ -0,0 +1,216 @@
...
+
+        word = self._strip_ge.sub(r"\1", word)
+        word = word.replace("sch", "$")
+        word = word.replace("ei", "%")
+        word = word.replace("ie", "&")
+        word = self._repl_xx.sub(r"\1*", word)
+
+        while len(word) > 3:
+            if len(word) > 5:
+                (word, success) = self._strip_emr.subn("", word)
```
stevenbird: It would be more pythonic to raise an error inside the function and catch it in the outer loop here

nltk/stem/cistem.py
```diff
@@ -0,0 +1,216 @@
...
+        word = word.replace("%", "ei")
+        word = word.replace("&", "ie")
+        word = word.replace("$", "sch")
+
+        return word
+
+
+    def segment(self, word):
+        """
+        This method works very similarly to stem. The difference is that in
```
stevenbird: NB. Not essential but note that Sphinx lets you refer to other methods in a way that generates a link in the HTML docs (plenty of examples in NLTK code)

nltk/stem/cistem.py
```diff
@@ -0,0 +1,216 @@
...
+        word = word.lower()
+
+        original = word[:]
+
+        word = word.replace("sch", "$")
+        word = word.replace("ei", "%")
+        word = word.replace("ie", "&")
+        word = self._repl_xx.sub(r"\1*", word)
+
+        while len(word) > 3:
```
stevenbird, line 188: Looks like the code at line 102... any scope for reducing the amount of code?

stevenbird:
Thanks for the contribution @LeonieWeissweiler

nltk/stem/cistem.py
LeonieWeissweiler: While I agree that the continue statements are not elegant, I think it would give the wrong idea to raise errors here because nothing has actually gone wrong. The continue statements essentially form a glorified if-statement that goes through the regexes and stops as soon as one of them has matched.

LeonieWeissweiler:
I addressed your comments, please take a look.

LeonieWeissweiler:
I think there's a problem with the doctest system @stevenbird

LeonieWeissweiler:
I completely removed and then re-added the code to force the doctest system to re-evaluate it, works now @stevenbird

stevenbird:
@LeonieWeissweiler: sorry for the long delay. This looks good, thank you.

PR #2047, opened 2018-06-21
WordNet: Fix memory leak when unknown lemmas are looked up

frankier:
Currently, calling synsets (and therefore by extension, lemmas and possibly some other methods) with a lemma not known to omw will result in the value of self._lang_data[lang][1][lemma] being read. Since this is a read on a defaultdict([]), this will create an entry with the lemma as a key and the empty list as a value. A long running program which looks up many different unknown lemmas will eventually run out of memory since it will be used to store these unknown lemmas and empty lists. The solution is to check if the entry exists beforehand.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1545,10 +1545,11 @@ class WordNetCorpusReader(CorpusReader):
         else:
             self._load_lang_data(lang)
             synset_list = []
-            for l in self._lang_data[lang][1][lemma]:
-                if pos is not None and l[-1] != pos:
-                    continue
-                synset_list.append(self.of2ss(l))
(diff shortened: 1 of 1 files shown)


jnothman (commented):
FWIW, if the defaultdict is static, you can freeze that automatic entry functionality and avoid such surprises with:
```
self._lang_data.default_factory = None
```

frankier:
Seems like a reasonable belt & braces precaution.

frankier:
Added some more protection on things that seem to grow when using omw

frankier:
Not sure why the tests appear to be failing. When I click the link says they all pass. Is there something I should fix?

alvations:
[CI: retest]

stevenbird:
Thanks @frankier @jnothman

PR #2054, opened 2018-07-01
Map satellite pos to adjective pos for omw, fixing problems with wordnet.lemmmas(...)

frankier:
Before:
```
>>> from nltk.corpus import wordnet
>>> wordnet.lemmas('Ｔ形+的', lang='cmn')
[]
```
After
```:
>>> from nltk.corpus import wordnet
>>> wordnet.lemmas('Ｔ形+的', lang='cmn')
[Lemma('t-shaped.s.01.Ｔ形+的')]
```

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1128,7 +1128,10 @@ class WordNetCorpusReader(CorpusReader):
 
     def ss2of(self, ss):
         ''' return the ID of the synset '''
-        return ("{:08d}-{}".format(ss.offset(), ss.pos()))
+        pos = ss.pos()
+        if pos == 's':
+            pos = 'a'
+        return ("{:08d}-{}".format(ss.offset(), pos))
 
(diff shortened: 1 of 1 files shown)


stevenbird:
@frankier is this justified as a global change?
@fcbond would you please confirm?

frankier:
It's justified because:
1. ss2of is only called in one place, lemma_names(...) (which is in turn called by lemmas(...) )
2. Open Multilingual Wordnet is meant to have scrapped the satellite pos tag. I can't find a reference for this any more, but a cursory look at the data will confirm.

Unfortunately, in practice omw does sometimes use the satellite pos tag. The languages that do are: nld, lit and slk. I have been in contact with @fcbond about this already. Given a new release is not forthcoming immediately, this should perhaps be taken into account.

frankier:
Okay. Now I've added an exception which can be removed if/when these languages also remove the satellite tag.

stevenbird:
That looks good, thanks @frankier

PR #2060, opened 2018-07-11
Fix span tokenizing when sentence starts with two single quotes

dennisverspuij:
In followup to #1751 for bug #1750, with this small regexp fix span word tokening a sentence starting with two single quotes ('') yields correct results, otherwise it would raise a ValueError: substring not found

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -166,3 +166,14 @@ class TestTokenize(unittest.TestCase):
         ]
         result = list(tokenizer.span_tokenize(test3))
         self.assertEqual(result, expected)
+
+        # Test case with double quotes at start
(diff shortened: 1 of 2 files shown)


dennisverspuij:
No, sorry, don't merge! The problem is not with the regexp I adjusted, I now understand it is for splitting double quotes after opening brackets while converting the quotes to `` to mark them as "starting quotes". The problem is with the attempt to revert these changes in: https://github.com/nltk/nltk/blob/ecced11335a8942beeab4ed9020eb93f18457d10/nltk/tokenize/treebank.py#L188-L196
This code should only revert the double quotes in the situations prescribed by the regexp's, not just hardcoded always! And since these regexp's are mutable (like done in the unicode quote tweaks of word_tokenize() documented nowhere) this can never be done correctly this way. The only way is the keeping track of splits while tokening. I hope I can find some time to rewrite the tokenizer to do so efficiently.

PR #2065, opened 2018-07-20
Added Two functions to that are missing in the _WordNetObject class.

afshinsadeghi:
Added Two functions to that was missing in the _WordNetObject class.
They read in_topic_domains and in_region_domains relations from WordNet dataset.  
These relations are already in WordNet. 
Cheers,
Afshin Sadeghi (University of Bonn)

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -167,8 +167,14 @@ class _WordNetObject(object):
     def topic_domains(self):
         return self._related(';c')
 
+    def in_topic_domains(self):
+        return self._related('-c')
+    
     def region_domains(self):
         return self._related(';r')
+    
+    def in_region_domains(self):
+        return self._related('-r')
 
     def usage_domains(self):
         return self._related(';u')


stevenbird:
@bethard: do you have an opinion on this? Do you know if these methods were omitted for a reason?

bethard:
I don't think there's a reason. They were probably just missed. Or maybe they're from a newer version of WordNet, and they didn't exist before? Either way, it makes sense to include them, though the commit should probably add a test case to ensure the implementations do the right thing.

afshinsadeghi:
Nice,  Bethard can you lead me to the folder to push the test cases?

alvations:
@afshinsadeghi you can add the tests for the functions of the 2 relations in https://github.com/nltk/nltk/blob/develop/nltk/test/wordnet.doctest and/or https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_wordnet.py

stevenbird:
@afshinsadeghi sorry for the delay...

Issue #1785, opened 2017-07-26
Fringe cases in MT eval metrics


alvations:
There are several fringe that still bugs the MT evaluation metrics in `nltk.translate`. 

The BLEU related issues are mostly resolved in #1330. But similar issues happens in RIBES and CHRF too:

 - `ribes_score.py`
   - https://github.com/nltk/nltk/blob/develop/nltk/translate/ribes_score.py#L290 and https://github.com/nltk/nltk/blob/develop/nltk/translate/ribes_score.py#L320 are subjected to ZeroDivisionError when the no. of possible ngram pairs is 0

 - `chrf_score.py`
   - The interface for the references for other scores supports multi-reference by default while ChrF score supports single reference. It should be standardize to accommodate multi-reference
      - But in the case of multi-reference score, there's no indication of which reference to choose in ChrF, we might need to contact the author to understand how to handle this.

ales-t:
In the Maja Popovic's implementation, when multiple references are provided, the one which leads to the highest f-score is used, see:

https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L155

danielgildea:
There seems to be a bug in computing bleu when one but not all of the sentences are shorter than the max n-gram length.  For example, the following test case should get bleu of 1.0 but does not:
```
references = [['John loves Mary'.split()], ['John still loves Mary'.split()]]
hypothesis = ['John loves Mary'.split(), 'John still loves Mary'.split()]
n = 4  #
weights = [1.0 / n] * n  # Uniform weights.
print    (corpus_bleu(references, hypothesis, weights))
```
A sentence of length 3 that is identical to the reference is scored as having 0 out of 1
correct 4-grams, instead of 0 out of 0 correct 4-grams.

Suggested patch:
```

--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -183,6 +183,8 @@ def corpus_bleu(
         # denominator for the corpus-level modified precision.
         for i, _ in enumerate(weights, start=1):
             p_i = modified_precision(references, hypothesis, i)
+            if (p_i == None):
+                continue       # no ngrams because ref was shorter than i
             p_numerators[i] += p_i.numerator
             p_denominators[i] += p_i.denominator
 
@@ -240,6 +242,7 @@ def modified_precision(references, hypothesis, n):
     and denominator necessary to calculate the corpus-level precision.
     To calculate the modified precision for a single pair of hypothesis and
[...]

PR #2069, opened 2018-07-27
Chrf bug fixes

ales-t:
Fix several problems in the implementation of CHRF metric.

### Missing check for division by zero

When nothing matches between the reference and hypothesis, unhandled `ZeroDivisionError` was raised.

### Unsafe type conversion

Lists of tokens were automatically converted to strings but type checking was only done on the hypothesis. Consequently, when there was a type mismatch between the hypothesis and reference, the results were incorrect.

### Differences from the original implementation

The papers do not specify all details needed to fully implement the metric. However, now there is code available from the original author:

https://github.com/m-popovic/chrF

It differs from the NLTK implementation in several aspects:

* Micro vs. macro average over n-gram orders. M. Popovic uses macro average over n-gram orders.
* Micro vs. macro average over sentences. Again, M. Popovic uses macro average here.
* Treatment of white space. By default, white spaces are ignored in CHRF++.

I changed the implementation, now it outputs identical results for all the test examples in the source file.

This can be verified by creating files with hypotheses and references and running the script with the following parameters (sets beta=3 and disables word-level F-score):

```
python chrF++.py -R tests.references.txt -H tests.hypotheses.txt -nw 0 -b 3
```

(Tested with commit ID: 6d3c384c8327a35c937ae231437cdb33ad4cd42d)

This PR deliberately does not include support for multiple references as that would change the interface of the functions.


nltk/translate/chrf_score.py
```diff
@@ -75,18 +84,31 @@ def sentence_chrf(reference, hypothesis, min_len=1, max_len=6, beta=3.0):
...
 
-def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
+def _preprocess(sent, ignore_whitespace):
+    if type(sent) != str:
+        # turn list of tokens into a string
+        sent = ' '.join(sent)
+
+    if ignore_whitespace:
+        sent = re.sub(r'\s+', '', sent)
+    return sent
```
alvations, line 103: Hmmm, this is the awkward part of chrf where the string is taken as an input.

nltk/translate/chrf_score.py
```diff
@@ -75,18 +84,31 @@ def sentence_chrf(reference, hypothesis, min_len=1, max_len=6, beta=3.0):
...
+
+def corpus_chrf(references, hypotheses, min_len=1, max_len=6, beta=3.0,
+                ignore_whitespace=True):
     """
     Calculates the corpus level CHRF (Character n-gram F-score), it is the
-    micro-averaged value of the sentence/segment level CHRF score.
+    macro-averaged value of the sentence/segment level CHRF score.
 
-    CHRF only supports a single reference.
+    This implementation of CHRF only supports a single reference at the moment.
```
alvations, line 154: I think it's possible to support multiple references. From the paper, it's unclear but from the code, it looks like the default is to use the best score https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L155

nltk/translate/chrf_score.py
```diff
@@ -110,28 +132,59 @@ def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
     :type max_len: int
     :param beta: the parameter to assign more importance to recall over precision
     :type beta: float
+    :param ignore_whitespace: ignore whitespace characters in scoring
+    :type ignore_whitespace: bool
```
alvations, line 179: +1

nltk/translate/chrf_score.py
```diff
@@ -110,28 +132,59 @@ def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
...
     :return: the sentence level CHRF score.
     :rtype: float
     """
 
-    assert len(list_of_references) == len(hypotheses), "The number of hypotheses and their references should be the same"
+    assert len(references) == len(hypotheses), (
+        "The number of hypotheses and their references should be the same")
+
+    # Keep f-scores for each n-gram order separate
+    n_gram_fscores = {k: [] for k in range(min_len, max_len + 1)}
```
alvations: It's not necessary to pre-populate, hashmaps with empty array as values in Python. If the aim is for optimization, then an fixed size array should be used as the values not the default Python list that is dynamic.

nltk/translate/chrf_score.py
```diff
@@ -110,28 +132,59 @@ def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
...
-        tpfp = sum(hyp_ngrams.values()) # True positives + False positives.
-        tffn = sum(ref_ngrams.values()) # True posities + False negatives.
-
-    precision = tp / tpfp
-    recall = tp / tffn
-    factor = beta**2
-    score = (1+ factor ) * (precision * recall) / ( factor * precision + recall)
-    return score
+    for reference, hypothesis in zip(references, hypotheses):
+        num_sents += 1
```
alvations: This is not necessary, we can use the `enumerate()`.

nltk/translate/chrf_score.py
```diff
@@ -110,28 +132,59 @@ def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
...
+
+        # preprocess both reference and hypothesis
+        reference = _preprocess(reference, ignore_whitespace)
+        hypothesis = _preprocess(hypothesis, ignore_whitespace)
```
alvations, line 196: I still find this weird. I think it's inconsistent with the rest of the scores in NLTK where the default is to assume that the inputs are preprocess and the scoring functions should focus on the numerical computation as much as possible.

alvations: Maybe it's time to really refactor the `nltk.translate.*_score.py` to a proper module that handles the preprocessing then the computation.

nltk/translate/chrf_score.py
```diff
@@ -110,28 +132,59 @@ def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
...
+            # calculate the number of ngram matches
+            overlap_ngrams = ref_ngrams & hyp_ngrams
+            tp = sum(overlap_ngrams.values())  # True positives.
+            tpfp = sum(hyp_ngrams.values())  # True positives + False positives.
+            tffn = sum(ref_ngrams.values())  # True posities + False negatives.
+
+            # Avoid division by zero errors. We also return 0 when both
+            # reference and hypothesis are empty.
+            if 0 in (tp, tpfp, tffn):
+                n_gram_fscores[order].append(0.0)
```
alvations: Hmmm, I think there's a more elegant way to handle this with try-except to be more explicit about the `ZeroDivisionError`. Also checking for `tp == 0` is unnecessary because it's in the numerator and it won't get into the math domain error.

nltk/translate/chrf_score.py
```diff
@@ -110,28 +132,59 @@ def corpus_chrf(list_of_references, hypotheses, min_len=1, max_len=6, beta=3.0):
...
+    # computes macro-averages both over n-gram lengths and sentences.
+
+    # how many n-gram sizes
+    num_ngram_sizes = max_len + 1 - min_len
+
+    # sum of f-scores over all sentences for each n-gram order
+    total_scores = [sum(n_gram_fscores[o]) for o in range(min_len, max_len + 1)]
+
+    # macro-average over n-gram orders and over all sentences
+    return (sum(total_scores) / num_ngram_sizes) / num_sents
```
alvations, line 213: Regarding macro vs micro average, we can do better than the original implementation by allowing a parameter with `average="macro"` as default. That'll follow the API design of `sklearn` metrics.

alvations (changes requested):
@ales-t Thanks for the contribution! It's good but requires some minor improvements.

alvations:
This is tricky, the implementation from https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L151 is totally different from the paper http://www.statmt.org/wmt15/pdf/WMT49.pdf

It included token n-gram scores on top of character ngram scores. That makes the interface quite different.

----

And in http://www.statmt.org/wmt16/pdf/W16-2341.pdf , although there's word score, it's separate from the character score.

alvations:
@ales-t  Regarding, this:

> Micro vs. macro average over sentences. Again, M. Popovic uses macro average here.

I think it's clear that Maja was using the *micro*-average here https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L186

nltk/translate/chrf_score.py
alvations: Hmmm, looking at https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L186 

For each hypothesis, per n-gram it's not taking average across all references but the max F-score. https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L155

In the final score computation, for each ngram, It's actually taking the micro average not macro-average across all hypotheses. https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L186 

Then across the n-grams, it's taking the macro-average at https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L189

nltk/translate/chrf_score.py
ales-t: I'm fairly sure I would get a `KeyError` if I try to append to a list corresponding to a nonexistent key. One option would be to use `collections.defaultdict` but this seems a bit like overkill here.

nltk/translate/chrf_score.py
ales-t: Isn't that out of the scope of this PR?

nltk/translate/chrf_score.py
ales-t: The value `num_sents` is used after the `for` loop finishes. Just to be sure, is this what you have in mind?

```
for num_sents, (reference, hypothesis) in enumerate(zip(references, hypotheses)):
    ...

num_sents += 1 # num_sents counts from zero
```

nltk/translate/chrf_score.py
ales-t: Right: across **references**, we take the **maximum** F-score.

Across **n-gram orders**, we always do **macro** average.

Now across **sentences**, the code outputs both macro and micro average, I missed that. This is the output for the test corpus of `corpus_chrf`:

```
$ python chrF++.py -R tests.references.txt -H tests.hypotheses.txt -nw 0 -b 3
start_time:	1534343254
c6+w0-F3	38.1372
c6+w0-avgF3	39.1009
end_time:	1534343254
```

So, `avgF3` is the macro average (what I implemented) and `F3` is the micro average. So I think it would make sense to add the parameter and perhaps make "micro" the default value.

I assume it should only apply to averaging over sentences in the corpus, do you agree?

ales-t:
Just to clarify, Maja Popovic's code is the implementation of her latest version of the metric, called chrf++: 

http://www.statmt.org/wmt17/pdf/WMT70.pdf

I tried to replicate its outputs in the setting where word (token) matches are disabled (`-nw 0`). I assume that the "character part" of the metric is the same as in the original chrf(3).

nltk/translate/chrf_score.py
ales-t: Checking for `tp == 0` was needed because if it's zero, then both precision and recall are zero, which means the denominator of f-score formula is also zero.

But I changed the code to explicitly handle `ZeroDivisionError` so it doesn't matter now :)

nltk/translate/chrf_score.py
ales-t: Yes, I suggested it in the related issue as well: https://github.com/nltk/nltk/issues/1785

Would it make sense to implement this behavior as part of the current PR? That means that `chrf3` would have the same interface as e.g. BLEU, is that correct?

I didn't want to break the current interface so I left this unimplemented but I'm happy to do it this way.

nltk/translate/chrf_score.py
alvations: Not an overkill. `defaultdict` seems like the perfect usage here =)

nltk/translate/chrf_score.py
alvations: `enumerate(..., start=1)` is an option. https://docs.python.org/3/library/functions.html#enumerate ;P

nltk/translate/chrf_score.py
alvations: Hmmm, let sees what we can do. I'll relook at the changes and then refactor the code again. It's easier to do it right for each metric and keeping in mind that someday we'll have this unified across all metrics.

nltk/translate/chrf_score.py
ales-t: Ok, done.

nltk/translate/chrf_score.py
alvations: Yes, I think `micro` and `marco` needs to be a parameter. 
I think Maja's code follows the old perl style BLEU and NIST code that's why she was keeping hashmaps and then using different averaging techniques based on what's easiest to access from the hashmap.

I'll take a look at this again tomorrow morning (~10-12 hours) later and see how we can work this out elegantly.

alvations:
It sounds crazy but if you don't mind lets write the code such that we can replicate the chrf across the years . 

I think I've an idea for an elegant API design for that since everything is dependent of precision, recall and fscore and the macro/micro average. Let me sleep on it and think through again and if it's a good idea, I'll try to incorporate the design into the existing code you've written =)

nltk/translate/chrf_score.py
ales-t: I would argue that's less readable than the current version but I'm fine doing it this way.

ales-t: For now, I calculate `num_sents` afterwards based on the size of f-score lists.

nltk/translate/chrf_score.py
```diff
@@ -176,6 +175,8 @@ def corpus_chrf(references, hypotheses, min_len=1, max_len=6, beta=3.0,
 
             n_gram_fscores[order].append(f_score)
 
+    num_sents = len(n_gram_fscores[min_len])
```
alvations: +1 Much better, but there's this might be even better! `num_sents = len(hypotheses)`

Simple is better than complex =)

nltk/translate/chrf_score.py
alvations: How about `num_sents = len(hypotheses)` =)

nltk/translate/chrf_score.py
ales-t: Your proposed solution won't work for generators. If you run `len(hypotheses)`, you'll exhaust the generator. This would break chrf3 usage in stream-based scenarios.

nltk/translate/chrf_score.py
ales-t: Again, as I wrote already: your proposed solution won't work for generators. If you run `len(hypotheses)`, you'll exhaust the generator. This would break chrf3 usage in stream-based scenarios.

nltk/translate/chrf_score.py
alvations: Yeah, that's true. Hmmm, maybe using the original `count=0; count+=1` code is more readable and easier to maintain. Sorry for the confusion. 

Let me think about it again, there must be a better way =)

ales-t: One way would be to simply not care about the streaming case. Maybe I'm being too "defensive" here...

alvations:
If you don't mind, let me try to add to your code step-by-step.

First we take out the ngram precision, recall, fscore computation and make it a function since, we would have to repeat the code if we do character and token ngrams.

alvations:
Okay in addition to the code/algorithm changes, I think now Maja explicitly uses GPL 
https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L7

I've asked Maja about relicesning https://github.com/m-popovic/chrF/issues/1 we'll wait and see but @stevenbird should we just drop this metric if she's sticking with GPL.

alvations:
[CI: retest]

nltk/translate/chrf_score.py
alvations: I think we can avoid doing this. By duck-typing 2 separate function and expose the `corpus_chrf_char` and `corpus_chrf_token` functions. And underspecify what the default `corpus_chrf` takes. 

Explicit is better than implicit?

alvations (approved):
Now this should be okay for this PR. 

But there are still more work on multi-reference, multiple points of micro vs macro average.

nltk/translate/chrf_score.py
```diff
@@ -75,18 +84,73 @@ def sentence_chrf(reference, hypothesis, min_len=1, max_len=6, beta=3.0):
...
+        prec = rec = fscore = epsilon
+    return prec, rec, fscore, tp
+
+
+def corpus_chrf(references, hypotheses, min_len=1, max_len=6, beta=3.0,
+                ignore_whitespace=True):
     """
     Calculates the corpus level CHRF (Character n-gram F-score), it is the
-    micro-averaged value of the sentence/segment level CHRF score.
+    macro-averaged value of the sentence/segment level CHRF score.
```
ales-t, line 152: Based on your comments and the code, it now seems to me that micro average would make more sense as the default. I'm happy to make the changes if you approve.

nltk/translate/chrf_score.py
ales-t: I don't like this type checking either, I was just trying to maintain the same interface. Two separate function names make sense, although I'm not sure I would be able to guess the difference between `*_char` and `*_token` from the name. (I assume the first would take sentences as strings and tokenize them whereas the other would take already tokenized lists?)

nltk/translate/chrf_score.py
alvations: +1

nltk/translate/chrf_score.py
alvations: This is really interesting problem of API design. I thought of doing


```
def _corpus_chrf(...):
    # generic function that is agnostic to input. 
    # When given list(str) -> character chrf
    # When given list(list(str)) -> token chrf

def corpus_chrf_char(...):
    _corpus_chrf(...)

def corpus_chrf_token(...):
    _corpus_chrf(...)

def corpus_chrf_token_char(...):
    corpus_chrf_token(...)
    corpus_chrf_char(...)

def sent_chrf_char(...):
    corpus_chr_char(...)

def sent_chrf_token(...):
    corpus_chr_char(...)
```

But that creates many level of spaghetti code =(

alvations: This is cleaner but harder to implement.

```
def _corpus_chrf(...):
    # generic function that is agnostic to input. 
    # When given list(str) -> character chrf
    # When given list(list(str)) -> token chrf

def corpus_chrf(references, hypotheses, min_char_n, max_char_n, min_tok_n, max_tok_n, ...):
    _corpus_chrf(references, hypotheses, min_char_n, max_char_n, ...)
    references_char, hypotheses_char = preprocess(references), preprocess(hypotheses)
    _corpus_chrf(references_char, hypotheses_char, min_tok_n, max_tok_n, ...)

def sent_chrf(min_char_n, max_char_n, min_tok_n, max_tok_n, ...):
    corpus_char(min_char_n, max_char_n, min_tok_n, max_tok_n, ...)

```

alvations: I was trying to reimplement the 2nd design but there are a couple averaging decisions invovled for `corpus_chrf` , looking at Maja's ChrF++ code.

 - average between token and char scores (looks like pseudo-micro/macro https://github.com/m-popovic/chrF/blob/master/chrF%2B%2B.py#L151)
 - average between ngrams orders (default: micro)
 - average between sentences (default: macro and micro)

nltk/translate/chrf_score.py
ales-t: I was being silly, `len(hypotheses)` is already used in the function. I'm using it to get `num_sents` as well. Let's not worry about generators for now?

nltk/translate/chrf_score.py
ales-t: I was hoping to get it done today but I didn't have enough time to write a clean implementation. I'm going on vacation for a week, I'll get back to this when I return. Thanks for the comments so far and for the patience ;)

alvations:
[CI: retest]

alvations:
[CI: retest]

stevenbird:
Thanks @ales-t, @alvations!

PR #2080, opened 2018-08-10
Change binary_search_file to avoid decoding part of a code point

frankier:
This shows up when trying to: e.g. repurpose the Wordnet reading code to deal with reading other WordNets that contain characters outside the single byte UTF-8 character range. (I know about OMW -- but it doesn't suit my current use-case at the moment.) The problem is that when binary_search_file seeks to the middle of a line, it might seek to the middle of a code point, causing a UnicodeDecodingError. This patch makes to avoid decoding in this case.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -1139,6 +1139,13 @@ class SeekableUnicodeStreamReader(object):
 
         return chars
 
+    def discard_line(self):
+        if self.linebuffer and len(self.linebuffer) > 1:
+            line = self.linebuffer.pop(0)
+            self._rewind_numchars += len(line)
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @frankier. I'm concerned about the performance implications of this patch. How about making it conditional on a flag being set?

frankier:
I don't think there should be any (negative) performance implications. `SeekableUnicodeStreamReader.discard_line` does less work than `SeekableUnicodeStreamReader.readline`. In particular it does no decoding.

stevenbird:
Ok, I see, thanks.

PR #2081, opened 2018-08-11
Updated sent_tokenize to seperate word.word (No space after fullstop)

py-ranoid:
Solves #2082 for rudimentary cases.
Problem : sent_tokenize fails to split when there's no space after '.'
Naive Solution : Used Regex to replace `.` with `.<space>`
```
    if seperate:
        matches = re.findall(r'[a-z]+[.][A-Z][a-z]+',text)
        for m in matches:
            text = text.replace(m,m.replace('.','. '))        
```
I presumed 
- the previous sentence ends with a series of small letters
- the next sentence starts with a capital letter followed by small letters

### DIFF
--- a/nltk/tokenize/__init__.py
+++ b/nltk/tokenize/__init__.py
@@ -81,7 +81,7 @@ from nltk.tokenize.util     import string_span_tokenize, regexp_span_tokenize
 from nltk.tokenize.stanford_segmenter import StanfordSegmenter
 
 # Standard sentence tokenizer.
-def sent_tokenize(text, language='english'):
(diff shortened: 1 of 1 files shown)


alvations:
This regex should not be made there are many cases of `X.Y` where `X.Y` is a token by itself. 

IMHO, since the Punkt sentence tokenizer is statistical, there shouldn't be regex hacks put in to determine how sentences should break.

alvations:
@py-ranoid A possible fix would be to retrain the punkt models on wikipedia and then re-submitting a PR with the models to https://github.com/nltk/nltk_data.

Otherwise, I think it's a `wontfix` for now =(

PR #2094, opened 2018-08-24
Compile some modules with Cython and add a manylinux wheel build target.

scoder:
Following #2093, this pull request adds optional compilation support for Python modules in nltk. I also added a Makefile target to build Linux wheels.

For now, only two modules are compiled: `nltk.grammar` and `nltk.parse.chart`. They proved to be the main bottlenecks in our application, and compiling them brought us a speedup of 30%. It is easy to add more modules if benchmarks prove their compilation helpful, but any additional compiled module also increases the size of the binary distributions, so this is a trade-off.

I set the CFLAGS so that the extension modules should work on any x86-ish system (`-mtune=generic`) and include basic debugging symbols (`-g1`) to get usable C stack traces should they crash.

Note that the compilation is not enabled by default but only when the envvar `CYTHONIZE_NLTK=true` is set. I added tox build targets to include this setup.


scoder:
The tests passed in Jenkins apparently, but it doesn't look like they tested the cythonized version. They don't even seem to be driven by tox, so I don't know how to make them use the compiled version.

alvations:
I've added an ad-hoc CI task on https://nltk.ci.cloudbees.com/job/cythonize/

Lets see how it fairs.

alvations:
[CI: cythonize-retest]

alvations:
[CI: cythonize-retest]

alvations:
[CI: retest]

alvations:
[CI: retest]

alvations:
[CI: retest]

setup.py
```diff
@@ -55,6 +55,30 @@
 # Add a group made up of all optional dependencies
 extras_require['all'] = set(package for group in extras_require.values() for package in group)
 
+
+MODULES_TO_COMPILE = [
+    'nltk.grammar',
+    'nltk.parse.chart',
+]
```
alvations: The CI passed! Lets see how far we can push this. Try adding these first: `nltk.classify`, `nltk.tag`,  `nltk.tokenize` and `nltk.translate`.

alvations:
Lets see how far we can push this, i.e. how many modules we can add.

scoder:
There's still something wrong with the Py3.6 build. It creates Py3.5 binaries, meaning, it's actually using Py3.5 and not Py3.6 (at least for the build):
https://nltk.ci.cloudbees.com/job/cythonize/PYV=3.6.4,jdk=jdk8latestOnlineInstall/ws/.tox/py3.6.4-cythonized/lib64/python3.5/site-packages/nltk/tokenize/
My guess is that the `basepython = python3` makes it use the system Python.

scoder:
Also, I checked the coverage analysis and the line coverage is still the same as the non-compiled case. That suggests that the binaries might actually not be used, because profiling/coverage support is not enabled by default for Cython compiled modules for performance reasons. Profiling of inner-Cython code requires the directive `profile=True`, and coverage analysis requires `linetrace=True` to generate the tracing code _and_ a C compile time macro to enable it in the extension modules.

In the workspace, the nltk source directory in the project root does not have the extension modules, only the `.tox/` subdirectory has them. If the tests are executed inside of tox from the subdirectory and not from the project root, that should be ok, but it should also impact the reported coverage results.

A simple way of debugging this is to print the `module.__file__` of modules that you expect to be compiled. It should be the path to the shared library and not the .py file.

alvations:
@scoder sorry for the delayed response. We realized that the old CI had issues and was sunsetting so we moved to Travis and it took a while for us to make it work and clean up while migrating to Travis. 

Currently the new CI can be fully controlled by the [`.travis.yml`](https://github.com/nltk/nltk/blob/develop/.travis.yml) and [`tox.ini`](https://github.com/nltk/nltk/blob/develop/tox.ini). 

Let me try and see whether I can enable a `cython` build on the new CI.

alvations:
Seems like the `cythonize` build is failing https://travis-ci.org/nltk/nltk/jobs/430825857#L1406 

@scoder any ideas how can this be fixed?

scoder:
I don't have any experience with `nose`, but the setup seems to be different. It runs the tests from the `nltk/test/` directory instead of the project root, and passes the `tox.ini` file as argument to nose instead of test runner arguments. The `py-travis` setup seems to use coverage analysis and runs the tests from the project root. Probably worth making both setups behave the same.

You can also enable coverage analysis for Cython modules, but since it slows them down (uses tracing), it's off by default, so it would require two compilation steps, one for the production wheels and one for the test runs. So I'd rather leave coverage analysis out here.

purificant:
While this PR does not target `develop` branch, there were some reformatting changes in `develop` due to https://github.com/nltk/nltk/pull/2163 which will eventually affect this PR as it makes it's way to `develop`. In order to resolve the formatting conflict please take formatting from `setup.py` in `develop` and insert line `ext_modules=ext_modules,` after `packages = find_packages(),` at the end of the file.

alvations:
@scoder If you don't mind, lets try this again in #2289 with the fast-forwarded branch.

PR #2114, opened 2018-09-12
Fix concordance print lines

cpbird:


alvations:
I presume the issue that you faced is that the concordance isn't printing the desired number of lines. That has been fixed on #2026 in the bleeding edge branch.

If you want it to be resolved now, you can try the bleeding edge version with the following:

```
pip3 install -U https://github.com/nltk/nltk/archive/develop.zip 
```

Otherwise waiting for the next version is the other option =)

cpbird:
Yea, sorry about that. Thanks.

PR #2125, opened 2018-09-19
Return probability-based measure of confidence of best tag from AveragedPerceptron

free-variation:
A small modification to the AveragedPerceptron which adds an option to the `tag` method to return, along with the best tag, its probability based on a multinomial distribution, computed as the softmax of the scores produced by the model.

This edit is to support an experiment in semi-supervised training of taggers for minor and/or dead languages, which often only have small annotated corpora.


free-variation:
How could *all* checks fail?  Is the build env sound?

nltk/tag/perceptron.py
```diff
@@ -45,7 +47,14 @@ def __init__(self):
         # Number of instances seen
         self.i = 0
 
-    def predict(self, features):
+
+    def _softmax(self, scores):
+        s = np.fromiter(scores.values(), dtype=float)
+        exps = np.exp(s)
+        return exps / np.sum(exps)
```
alvations, line 55: Understandably, numpy is used here for speed. In that case only the exponential needs to be kept. 

```python
def _softmax(self, scores):
    exps = np.exp(scores.values()) 
    return exps / np.sum(exps)
```

`np.exp()` should automatically infer the dtype to `numpy.float64`, e.g. 

```python
>>> import numpy as np
>>> type(np.exp([0.1, 0.2, 0.5])[0])
<class 'numpy.float64'>
```

nltk/tag/perceptron.py
```diff
@@ -45,7 +47,14 @@ def __init__(self):
...
+
+
+    def predict(self, features, conf=False):
```
alvations: Explicit is better than implicit, esp. when it comes to arguments that users' can access. 

I would recommend, `def predict(self, features, confidence=False)` or `def predict(self, features, return_conf=False)`

nltk/tag/perceptron.py
```diff
@@ -154,10 +169,14 @@ def tag(self, tokens):
             tag = self.tagdict.get(word)
             if not tag:
                 features = self._get_features(i, word, context, prev, prev2)
-                tag = self.model.predict(features)
+                tag = self.model.predict(features, conf)
```
alvations: Actually this is not a good way to handle the function. 
Because now `tag` type is left unknown, between container or basic type.

nltk/tag/perceptron.py
```diff
@@ -140,7 +155,7 @@ def __init__(self, load=True):
             AP_MODEL_LOC = 'file:'+str(find('taggers/averaged_perceptron_tagger/'+PICKLE))
             self.load(AP_MODEL_LOC)
 
-    def tag(self, tokens):
+    def tag(self, tokens, conf=False):
```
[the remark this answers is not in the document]
alvations: Lets use default `confidence=True` or `return_conf=True` instead.

nltk/tag/perceptron.py
alvations: Lets use the default `return_conf=True` and `confidence=True`, so that return types can be explicit when used by `tag()` or `train()`.

alvations: Hmmm, maybe an argument here is not the best, esp. when this is an intermediate step that most users won't use. Moreover multiple return type is nasty.

How about making `predict()` always return confidence and do away with the mess of allowing multiple return types =)

So how about:

```python
def predict(self, features):
```

and then

```python
# Do a secondary alphabetic sort, for stability.
best_label = max(self.classes, key=lambda label: (scores[label], label))
# Compute the confidence.
conf = max(self._softmax(scores))

return best_label, conf
```

nltk/tag/perceptron.py
alvations: If we make `predict()` always return confidence then we don't have to propagate the `conf` kwarg =)

nltk/tag/perceptron.py
alvations: If we make `predict` always return confidence, then we will do:

```python
tag, conf = self.model.predict(features)
```

nltk/tag/perceptron.py
```diff
@@ -154,10 +169,14 @@ def tag(self, tokens):
...
-                tag = self.model.predict(features)
+                tag = self.model.predict(features, conf)
             output.append((word, tag))
+            
             prev2 = prev
-            prev = tag
+            if conf == True:
+                prev = tag[0]
+            else:
+                prev = tag
```
alvations: If we make predict always return confidence, then we can avoid the if-else and return to the original:

```python
prev2 = prev
return tag
```

nltk/tag/perceptron.py
alvations: We have to do the same for `train()` too at https://github.com/nltk/nltk/blob/develop/nltk/tag/perceptron.py#L196 

```python
guess, conf = self.model.predict(feats)
```

free-variation: > Actually this is not a good way to handle the function.
> Because now `tag` type is left unknown, between container or basic type.

Agreed, though this is where I really dislike untyped languages...

free-variation:
All great suggestions @alvations , so thank you.  I'll implement shortly.

nltk/tag/perceptron.py
free-variation: scores.values() here returns a `dict_values` object which must be iterated.

nltk/tag/perceptron.py
free-variation: Hang on though -- calling the softmax has a computational cost.  Shouldn't the client be able to decide to not incur that extra cost?

nltk/tag/perceptron.py
free-variation: Again, do we want to force all clients to compute the probabilities?  I'm assuming that in the vast majority of cases clients just want the argmax.

nltk/tag/perceptron.py
free-variation: It's true, but as indicated above, we may want to allow the client to decide, since there's cost.

nltk/tag/perceptron.py
alvations: Hmmm... Multiple return types is really nasty though. Let me read up on some API design. 

Current ideas:
 1. Multiple return type
 2. Spaghetti function, e.g. `predict_with_conf()` -> `predict()` and add softmax
 3. Create a return object (maybe namedtuple or something) and when `conf=False`, the `object.conf` is None.

free-variation: We want to not break current client code, right?

Perhaps a separate function (your #2) is the least disruptive solution.  Not beautiful, but functional?

alvations: ~Let sit on this for a day.~

I'll look through some C++ design and other Python code design to see whether there's a better solution. It's because this is not the only instance of this so it's better to find a convention that we're are comfortable and let newer code from henceforth follow it =)

alvations: How about we still return the tuple of (best_label, conf) but use default `conf=False` and put an if-else, e.g.


```python
def predict(self, features, return_conf=False):
    ...
    # Do a secondary alphabetic sort, for stability.
    best_label = max(self.classes, key=lambda label: (scores[label], label))
    # Compute the confidence.
    conf = max(self._softmax(scores)) if return_conf else None

    return best_label, conf
```

Then for `tag()` and `train()`, read the conf accordingly and propagate the `**kwargs` to `predict()`.

free-variation:
@alvations any further wisdom about how we might set up the API?  I suspect that the tension between not breaking current clients and putting in place a general method for the future might not be easily resolved.  Locally I'm returning either a pair or a triple depending on the `confidence` parameter.

alvations:
@free-variation if there's no objections, lets go with the following?

```python
def predict(self, features, return_conf=False):
    ...
    # Do a secondary alphabetic sort, for stability.
    best_label = max(self.classes, key=lambda label: (scores[label], label))
    # Compute the confidence.
    conf = max(self._softmax(scores)) if return_conf else None

    return best_label, conf
```

And then in `tag()` and `train()`:

```python
tag, conf = self.model.predict(features)
```

That'll change the return type of `predict()` but at least we keep it consistent =)

And most of the time, users are exposed to `tag()` and `train()` which won't be affected if we change the functions this way.

free-variation:
Thanks @alvations.  What does `tag()` return, in that case?  The original motivation here is to have `tag()` optionally return the multinomial probs -- without breaking existing clients, naturally.

free-variation:
@alvations see updated version.  There are a couple of differences:

- Depending on the `return_conf` parameter, `tag()` returns either a list of pairs or a list of triples.  This way clients won't break.
- For my particular use case, I need to turn *off* the `tagdict` optimization in `tag()`, which looks up category-"unambiguous" words for their POS value in a table, stored with the model at training time.  This is an obvious and necessary optimization, but for the sort of semi-supervised training I'm working on it needs to be turned off.

Thanks.

alvations:
[CI: retest]

stevenbird:
@alvations or @purificant: Are you happy for me to merge this?

free-variation:
I've long given up on this -- the pace of response is glacial.  Switching to spaCy.

stevenbird:
Sorry about that @free-variation

stevenbird:
[CI: retest]

alvations:
Sorry about the slow replies @free-variation =)

@stevenbird LGTM.

Issue #2102, opened 2018-08-31
TreePrettyPrinter throws TypeError when Tree.leaves() are list of tuples


alvations:
When the `Tree.leaves()` are made up of list of tuples, e.g. `[('the', 'DT'), ('cat', 'NN')])]`, `Tree.pretty_print()` that calls the `TreePrettyPrinter` throws a `TypeError`, e.g.

```python
import nltk 
sentence = [("the", "DT"), ("little", "JJ"), ("yellow", "JJ"), ("dog", "NN"), ("barked","VBD"), ("at", "IN"), ("the", "DT"), ("cat", "NN")]

pattern = """NP: {<DT>?<JJ>*<NN>}
VBD: {<VBD>}
IN: {<IN>}"""
NPChunker = nltk.RegexpParser(pattern) 
result = NPChunker.parse(sentence)
result.pretty_print()
```

[out]:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-32-306ca23c9095> in <module>()
      7 NPChunker = nltk.RegexpParser(pattern)
      8 result = NPChunker.parse(sentence)
----> 9 result.pretty_print()

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nltk/tree.py in pretty_print(self, sentence, highlight, stream, **kwargs)
    697         """
    698         from nltk.treeprettyprinter import TreePrettyPrinter
--> 699         print(TreePrettyPrinter(self, sentence, highlight).text(**kwargs),
    700               file=stream)
    701 

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nltk/treeprettyprinter.py in __init__(self, tree, sentence, highlight)
     95                             if not isinstance(b, Tree):
     96                                 a[n] = len(sentence)
---> 97                                 sentence.append('%s' % b)
     98         self.nodes, self.coords, self.edges, self.highlight = self.nodecoords(
     99                 tree, sentence, highlight)

TypeError: not all arguments converted during string formatting
```

More details on https://stackoverflow.com/questions/52107352/how-can-i-pretty-print-a-nltk-tree-object

aquatiko:
Fixed in PR #2129 , have a look

stevenbird:
Thanks @aquatiko

PR #2129, opened 2018-09-22
Issue #2102 solved i.e preety_print for tuples

aquatiko:


nltk/treeprettyprinter.py
```diff
@@ -93,6 +93,8 @@ def __init__(self, tree, sentence=None, highlight=()):
                         for n, b in enumerate(a):
                             if not isinstance(b, Tree):
                                 a[n] = len(sentence)
+                                if type(b) == tuple:
+                                    b = '/'.join(b)
```
alvations, line 97: These type checking is awkward but I guess it's necessary. 

Maybe @dimazest has better suggestions since he worked most on the `nltk.parse` modules.

aquatiko: @alvations would ```isinstance``` be better than ```type```?

alvations:
[CI: retest]

stevenbird:
@dimazest: do you have an opinion about this?

PR #2130, opened 2018-09-24
Added GoodTuring Smoothing, fixed few errs

### LINKED ISSUE #2124: Adding Good turing and Add -n Smoothing
Apart from the existing 2 smoothing algorithms already present, I think the add -n smoothing and the Good turing smoothing can be added

sahitpj:
The following Pull Request resolves #2124 

Fixed a few ambiguous stuff in smoothing. Do let me know of any other changes


iliakur:
@sahitpj thank you for taking the time to look at this. A couple of comments.

1. I noticed you haven't included any examples of how this Good-Turing estimate would be used. If you look under `nltk/test/unit/lm` you can see a whole bunch of examples for how models are expected to perform. It would be great to see some tests for Good-Turing in there as well.
1. You set `counter` and `vocabulary` attributes on the smoothing objects, but if you look at the `__init__` of their parent class, `Smoothing`, you'll see that this already happens. Yes, I'm not sure about the names to be honest, but let's discuss that in a separate thread, and keep this changeset focused on Good-Turing.
2. From what I remember of reading about Good-Turing estimates, we'd need to fit some kind of linear regression to handle missing datapoints for the higher counts. Feel free to take a look at [something like this](http://u.cs.biu.ac.il/%7Eyogo/courses/mt2013/papers/chen-goodman-99.pdf) as a reference. I don't see this in the current draft.

I'm very grateful that you're tackling this issue, but I'm afraid I cannot in good conscience sign off on this PR until these points have been adequately addressed.

sahitpj:
Yeah definitely @Copper-Head, will do the rest and come back to you

purificant:
https://github.com/nltk/nltk/pull/2163 reformatted code in `develop`, because of this a one character change is required in order to resolve a conflict in this PR.
Please change `return 1. / len(self.vocabulary)` to `return 1.0 / len(self.vocabulary)` in order to resolve the conflict - the change is to add a `0`.

stevenbird:
Thanks @purificant

iliakur:
@stevenbird Sorry to be a stickler, but could you undo this merge please? The code Sahit had in this PR had some issues that were not addressed (see discussion). In its current state this code only reduces the quality of the LM module.

We can address the black formatting issues separately.

Thanks!

Issue #2134, opened 2018-10-01
WordNetError when loading lemmas with numeric values >= 10


fabianhoward:
Since at least NLTK 3.3 running the below code  results in the error shown below.

```python
>>> from nltk.corpus import wordnet as wn
>>> wn.lemma('jump.v.11.jump')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fabian/venv/lib/python3.6/site-packages/nltk/corpus/reader/wordnet.py", line 1272, in lemma
    raise WordNetError('no lemma %r in %r' % (lemma_name, synset_name))
nltk.corpus.reader.wordnet.WordNetError: no lemma '.jump' in 'jump.v.1'
```

I have reproduced it inside a clean venv with the latest nltk commit (https://github.com/nltk/nltk/commit/b991f244558154d09041d5a62b0c0e55faaab802 at the time of writing) on ubuntu 16.04 with python 3.6.5 and verified this bug does not happen with NLTK version 3.2.2

Based on experimentation I believe this bug affects all lemmas where the number XX (as in  jump.v.XX.jump) is 10 or above.

fabianhoward:
I have added the failing test in https://github.com/nltk/nltk/pull/2135

fabianhoward:
I think the assertion @campionfellin made in
https://github.com/nltk/nltk/issues/1789#issuecomment-320452986 is incorrect since the `jump.v.11.jump` is a valid lemma. I will try and put up a fix.

PR #2135, opened 2018-10-01
Fix https://github.com/nltk/nltk/issues/2134

fabianhoward:
Close https://github.com/nltk/nltk/issues/2134


stevenbird:
Nice fix, thanks @fabianhoward

PR #2142, opened 2018-10-13
Run tests for Python 3.7 in Travis

purificant:
Python 3.7 was released on 2018-06-27. Running automated tests for this version of Python will help future proof existing and new contributions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,14 @@ matrix:
     - python: '3.6' # Only runs checks on 3rd party for latest Python distribution
       env: TOXENV="py-travis-third-party"
 
+    - python: 3.7
+      env: TOXENV="py-travis"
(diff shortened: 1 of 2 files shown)


purificant:
Looking at the build logs, there are issues with python 3.7, closing this PR as those need to be resolved first.

alvations (commented):
Still sudo is required. For now it's okay because there's no auto deployment code but if we forget this later on, it'll come back to bite us.

purificant:
@alvations What's the issue with sudo?

alvations:
I'm not sure what does `sudo` actually and our test environments don't affect anything so I think that's okay. 

I think it's okay to use it as long as we remember that at some point when Travis/Docker resolve the issue, we should disable it.

Issue #2145, opened 2018-10-14
python 3.7 compatibility: re.sub() behavior


purificant:
[re.sub()](https://docs.python.org/3/library/re.html#re.sub) in python 3.7 returns different result to previous versions for some regular expressions. This is due to a fix for a bug where previously re.sub() didn't replace empty matches adjacent to a previous non-empty match.

The rationale for the fix is here: https://bugs.python.org/issue32308
The code change is here: https://github.com/python/cpython/commit/fbb490fd2f38bd817d99c20c05121ad0168a38ee
The change is reflected in the documentation through the following paragraph [here](https://docs.python.org/3.7/library/re.html#re.sub):
`Empty matches for the pattern are replaced when adjacent to a previous non-empty match.`

For NLTK this means that the following line produces different result in 3.7 compared to <= 3.6:
https://github.com/nltk/nltk/blob/ad415c94b177b43561fc8e48bf1cda24bb66afc7/nltk/corpus/util.py#L69
As a result, many unit tests fail when running python 3.7+ due to corpus loader not finding the data files.

Example of behaviour in python <= 3.6
```
$ python
Python 3.6.6 (default, Oct 13 2018, 12:50:59) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> re.sub(r'(([^/]*)(/.*)?)', r'\2.zip/\1/', 'test')
'test.zip/test/'
```

and in python 3.7+
```
$ python
Python 3.7.0 (default, Oct 13 2018, 17:53:23) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> re.sub(r'(([^/]*)(/.*)?)', r'\2.zip/\1/', 'test')
'test.zip/test/.zip//'
>>>
```

purificant:
Proposed fix is here: https://github.com/nltk/nltk/pull/2146
Thoughts / suggestions are welcome.

Issue #2144, opened 2018-10-14
build failing on latest develop


purificant:
The latest [build](https://travis-ci.org/nltk/nltk) is failing for `develop` branch due to a unit test `test_in_topic_domains`. Any new pull requests that do no include a fix for this will also have failing builds.

purificant:
Proposed fix: https://github.com/nltk/nltk/pull/2141

purificant:
Resolved

PR #2146, opened 2018-10-14
fix LazyCorpusLoader for python 3.7+ 

purificant:
This pull request addresses Python 3.7 compatibility issue outlined in: https://github.com/nltk/nltk/issues/2145

### DIFF
--- a/nltk/corpus/util.py
+++ b/nltk/corpus/util.py
@@ -66,7 +66,7 @@ class LazyCorpusLoader(object):
 
     def __load(self):
         # Find the corpus root directory.
-        zip_name = re.sub(r'(([^/]*)(/.*)?)', r'\2.zip/\1/', self.__name)
+        zip_name = re.sub(r'(([^/]+)(/.*)?)', r'\2.zip/\1/', self.__name)
         if TRY_ZIPFILE_FIRST:
             try:
                 root = nltk.data.find('{}/{}'.format(self.subdir, zip_name))
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{27,35,36}
+    py{27,35,36,37}
     pypy
     py{27,35,36}-nodeps
     py{27,35,36}-jenkins


purificant:
Travis build will fail due to https://github.com/nltk/nltk/issues/2144. Once resolved, the build should be green.

alvations:
BTW, travis doesn't support Python 3.7 for their default Ubuntu distributions without `sudo:yes`, see https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905

purificant:
For python 3.7 on travis we can do something like this: https://github.com/nltk/nltk/pull/2142
once the compatibility fixes are in develop in order to produce a green build.

stevenbird:
So if I merge this, are we ready to deal with any fallout on Travis for python 3.7?

alvations:
I'm not sure. But perhaps creating a branch and merging it there to check would be better than merging directly to `develop`.

purificant:
The proposed code changes here only do two things:
- address the python 3.7 compatibility issues outlined in https://github.com/nltk/nltk/issues/2145
- add py37 to tox configuration so that you can run `tox` and verify test execution under python 3.7 in a local dev environment

This PR does not change Travis configuration, so we will not see anything different on Travis, the build will only run for <= 3.6 as before. For example, there is a successful build for this PR here: https://travis-ci.org/nltk/nltk/builds/444056973

That being said, I think this might be the last fix need to get a green Travis build for Python 3.7 and I'd like to propose a new PR to start testing on 3.7 in Travis along the lines of https://github.com/nltk/nltk/pull/2142 once all the fixes are in develop.

I would hold of on merging this at the moment though due to an ongoing GitHub outage, best to wait until the status page confirms that all service is restored: https://status.github.com/messages
For example, webhooks are down at the moment, which means externals builds like Travis won't get triggered and not all comments are visible on the PR as sometimes you are shown an old version.

alvations:
Thanks @purificant  for the detailed explanation!

It passes on all other python versions from the last time it tested so it shouldn't be a problem. But yes, we should wait for github to be back up.

alvations:
Github has resolved their issues, lets try merging this.

We do still have to remember to disable the `sudo` so when travis support it by default.

Issue #2148, opened 2018-10-17
python 3.7 compatibility PEP 479


purificant:
There are several places in the code base that result in `RuntimeError: generator raised StopIteration` when executed with python 3.7+.

This happens because [PEP 479](https://www.python.org/dev/peps/pep-0479/) made a change to generators: when [StopIteration](https://docs.python.org/3/library/exceptions.html#StopIteration) is raised inside a generator, it is replaced with [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) when the exception is about to bubble out of the generator.
This behaviour is not backwards compatible and was slowly introduced with this [transition plan](https://www.python.org/dev/peps/pep-0479/#transition-plan).

In python 3.6 the code emits a non-silent deprecation warning, but starting with python 3.7 it becomes a runtime error.

Code changes are necessary to enable python 3.7 compatibility for PEP 479.

purificant:
Several amends merged as part of https://github.com/nltk/nltk/pull/2153

alvations:
All patched up for Py3.7? The CI tests pass =) 

Thanks @purificant for all the work!

purificant:
Thanks @alvations for your help. I think we can close this for now as no further known changes are needed. In the future if there are any python 3.7 issues with `RuntimeError: generator raised StopIteration` we can reference this issues.

ChrisIsKing:
I'm facing this issue when using `nltk.AnnotationTask().kappa()`

GavinAbercrombie:
Same issue with AnnotationTask

PR #2149, opened 2018-10-17
fix RuntimeError in doctest as per PEP 479

purificant:
Fix failing `toolbox.doctest` when executing with python 3.7 as per https://github.com/nltk/nltk/issues/2148

How to reproduce:
```
>>> from nltk import toolbox
>>> f = toolbox.StandardFormat()
>>> f.open_string('')
>>> list(f.raw_fields())
Traceback (most recent call last):
  File "/home/foo/projects/nltk/nltk/toolbox.py", line 72, in raw_fields
    line = next(file_iter)
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: generator raised StopIteration
```

PR #2150, opened 2018-10-17
fix RuntimeError in unit test as per PEP 479

purificant:
Fixes `test_sentence_nist` unit test when executed with python 3.7 as per https://github.com/nltk/nltk/issues/2148

Example stack trace:
```
1) ERROR: test_sentence_nist (nltk.test.unit.translate.test_nist.TestNIST)
    unit/translate/test_nist.py line 35 in test_sentence_nist
      nltk_nist = corpus_nist(references, hypotheses, i)
    /home/foo/projects/nltk/nltk/translate/nist_score.py line 97 in corpus_nist
      ngram_freq.update(ngrams(reference, i))
    /home/foo/projects/nltk/.tox/py37/lib/python3.7/collections/__init__.py line 653 in update
      _count_elements(self, iterable)
   RuntimeError: generator raised StopIteration
```

Can also be reproduced with:
```
>>> from nltk.util import ngrams
>>> list(ngrams([], 3))
  File "/home/foo/projects/nltk/nltk/util.py", line 467, in ngrams
    history.append(next(sequence))
StopIteration

RuntimeError: generator raised StopIteration
```

### DIFF
--- a/nltk/util.py
+++ b/nltk/util.py
@@ -464,7 +464,13 @@ def ngrams(sequence, n, pad_left=False, pad_right=False,
 
     history = []
     while n > 1:
-        history.append(next(sequence))
+        # PEP 479, prevent RuntimeError from being raised when StopIteration bubbles out of generator
+        try:
+            next_item = next(sequence)
+        except StopIteration:
+            # no more data, terminate the generator
+            return
+        history.append(next_item)
         n -= 1
     for item in sequence:
         history.append(item)


alvations:
Thanks @purificant for the proposal to catch up with PEP 479. 

Could you merge the two PRs into one? 
 - #2150
 - #2149

Are there more instances of `StopIteration` from generators in the codebase?

Also, I think it's might not be a good idea to share a `return` in a generator `yield`-ing function like `ngrams`. I understand that the PEP canonically suggests:

```python
if context is None:
    while True:
        yield next(line_pair_iterator)
```

Becomes:

```python
if context is None:
    yield from line_pair_iterator
    return
```

But let me check if there's a better solution.

purificant:
@alvations Combined PR for the PEP 479 changes is here: https://github.com/nltk/nltk/pull/2153

purificant:
I can't rule out that there aren't any more instances of `StopIteration` being thrown inside of a generator like this. The two places that I've found were highlighted through code coverage by doc tests and unit tests. Theoretically, any generator that can raise a `StopIteration` from any source other than a `return` statement could result in `RuntimeError`. Easiest example of this is any `def` function with a `yield` and a `next()` call.

The goal of fixes like this is not necessarily to catch up with PEP syntax, but rather to make NLTK work on python 3.7+ 

Now that PEP 479 semantics are rigorously enforced in python 3.7, the only way to terminate a generator gracefully is through a `return`, by either explicitly having this statement or reaching an end of a function where `return None` is implicit, which will also raise a `StopIteration` exception, but it will not be converted into a `RuntimeError` as it bubbles out.

purificant:
Closing as commit was merged as part of https://github.com/nltk/nltk/pull/2153

PR #2157, opened 2018-10-23
Testing 3.7 with third party tools on CI

alvations:


.travis.yml
```diff
@@ -32,13 +32,16 @@ matrix:
     - python: '3.6'
       env: TOXENV="py-travis"
 
-    - python: '3.6' # Only runs checks on 3rd party for latest Python distribution
+    - python: '3.6' # To be removed when 3.7 doesn't need sudo.
```
purificant, line 35: Is this comment intended?

purificant (approved):

.travis.yml
alvations: Whoops nope.

alvations: Oh wait, yes, that's intended. Because eventually, we only want to test 1 environment for third party. Now it's doing both py 3.6 and 3.7 for third party tools.

purificant: Thanks for clarifying @alvations, I got a bit confused as not sure how third party tests on py3.6 relate to travis sudo for py3.7

So the intention is to only run third party tests on the latest version?
And once sudo is not required for 3.7 we can remove third party tests for 3.6?

alvations: Yes, that should be the goal. The third party tests puts extra loads on the academic sites hosting the data/models and it takes quite some time to complete so I think it's best to minimize the tests on the third party tools build.

Issue #2115, opened 2018-09-13
cls vs self in classmethods


alvations:
There are several instances of `@classmethod` that doesn't have `cls` as the first argument:

```
nltk/test/unit/lm/test_vocabulary.py:20: [C0202(bad-classmethod-argument), NgramModelVocabularyTests.setUpClass] Class method setUpClass should have 'cls' as first argument
nltk/corpus/reader/mte.py:47: [C0202(bad-classmethod-argument), MTEFileReader._word_elt] Class method _word_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:51: [C0202(bad-classmethod-argument), MTEFileReader._sent_elt] Class method _sent_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:55: [C0202(bad-classmethod-argument), MTEFileReader._para_elt] Class method _para_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:59: [C0202(bad-classmethod-argument), MTEFileReader._tagged_word_elt] Class method _tagged_word_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:78: [C0202(bad-classmethod-argument), MTEFileReader._tagged_sent_elt] Class method _tagged_sent_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:82: [C0202(bad-classmethod-argument), MTEFileReader._tagged_para_elt] Class method _tagged_para_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:86: [C0202(bad-classmethod-argument), MTEFileReader._lemma_word_elt] Class method _lemma_word_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:93: [C0202(bad-classmethod-argument), MTEFileReader._lemma_sent_elt] Class method _lemma_sent_elt should have 'cls' as first argument
nltk/corpus/reader/mte.py:97: [C0202(bad-classmethod-argument), MTEFileReader._lemma_para_elt] Class method _lemma_para_elt should have 'cls' as first argument
nltk/parse/bllip.py:203: [C0202(bad-classmethod-argument), BllipParser.from_unified_model_dir] Class method from_unified_model_dir should have 'cls' as first argument
```

c.f. [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments) and [What is the 'cls' variable used for in Python classes?](https://stackoverflow.com/questions/4613000/what-is-the-cls-variable-used-for-in-python-classes)

[...]

PR #2161, opened 2018-10-28
pylint cleanup: cls vs self

purificant:
Cleanup name of first parameter in class method: https://github.com/nltk/nltk/issues/2115


stevenbird:
Thanks @purificant!

PR #2109, opened 2018-09-07
Feature/spanish corpus reader

RaulAranovich:
Hello. I developed a reader for the Spanish books in Project Gutenberg. It has some advantages over generic readers. I thought I would contribute it. Hopefully you will find it useful. This is my first commit to the project, I would like to receive comments in case I made any mistakes. Thanks!

PR #2043, opened 2018-06-12
K alpha

myproblemchild:
Rewrite K-Alpha calculation method.

The previous method gives incorrect results on the data I've tested.
The simplest one being: [('c1', 'item1', 'yes'), ('c2', 'item1', 'no'), ('c1', 'item2', 'yes'), ('c2', 'item2', 'no'), ('c3', 'item2', 'no'), ('c3', 'item3', 'yes')].
My implementation gives correct results on all the data I've tried it on.

For one, the item3 should not be taken into account since it has only 1 rating present.
But even ignoring that, the existing method gives incorrect results.

By the way, I'd be happy to add tests if you could explain how to run it in a simple command :)

PR #2018, opened 2018-05-14
Adding improved regexes to handle clitics c.f. #1995

djinn-anthrope:
The resolution to issue #1995 has been attempted. Please do take a look.

PR #2163, opened 2018-10-28
code formatting with black

purificant:
As a followup from https://github.com/nltk/nltk/pull/1935

This is a result of running [black](https://github.com/ambv/black) code formatter on the entire codebase.

How to reproduce:
```bash
$ pip install black
$ black -S .
```

## Motivation:
- Running linting tools on the codebase produces many noisy warnings / errors related to code formatting / pep8, this makes it harder to focus on the important issues without getting lost in the unnecessary detail.
- Adopt a single automated tool for all code formatting needs. Less manual formatting, more consistent code style across the library.
- Black was chosen as a tool due to it's focus on simplicity, practicality and ease of use. Other tools such as autopep8 were considered but their use produces worse diffs with more superficial changes.

## Configuration:
- Black aims to be a tool that ships with sensible defaults and requires little to no configuration. In this instance, -S was chosen in order to minimise changes where historically entire files were written using a particular string quoting style, omitting this flag will reformat all strings to use double quotes.


alvations:
@purificant Black formatter is nice but most of the time it's really okay to exceed the no. of chars length per line, esp. for some specific algorithms implemented in NLTK. https://www.youtube.com/watch?v=wf-BqAjZb8M has some good motivation as to why it's okay to compromise sometimes.

Was the formatting for these changes in the PR checked manually or was it automatically created?

purificant:
> @purificant Black formatter is nice but most of the time it's really okay to exceed the no. of chars length per line, esp. for some specific algorithms implemented in NLTK. https://www.youtube.com/watch?v=wf-BqAjZb8M has some good motivation as to why it's okay to compromise sometimes.

@alvations 
This is exactly the philosophy on line length that black is following, more here: https://github.com/ambv/black#line-length and it quotes https://youtu.be/wf-BqAjZb8M?t=260 the same video for reference ( with direct link into the line length comments )

purificant:
The PR is entirely automatically generated. Instructions to reproduce this PR are supplied at the top so that anyone can replicate it to confirm that no other changes were made. Manual review was undertaken with spot checks covering majority of the changes to determine what classes of formatting changes were done. Seeing the changes on the command line with `git show <commit>` is easier than clicking on every file in the GitHub UI.

stevenbird:
I'm ok with this, except that it's likely to cause every other PR to break

purificant:
> I'm ok with this, except that it's likely to cause every other PR to break

@stevenbird 
That's a very good point. As demonstrated when merging https://github.com/nltk/nltk/pull/2161 created conflicts in this PR, which I have now resolved.

There are currently 12 outstanding PRs that could be affected by code reformatting. I've compared the code in the PRs against the reformatting changes to determine the impact.

:warning: https://github.com/nltk/nltk/pull/2130 affected, one character change, should be trivial to rectify when addressing outstanding comments on the PR
:heavy_check_mark:  https://github.com/nltk/nltk/pull/2129 not affected
:heavy_check_mark: https://github.com/nltk/nltk/pull/2125 existing conflicts, not affected
:heavy_check_mark: https://github.com/nltk/nltk/pull/2109 existing conflicts, not affected because contains only code additions
:warning: https://github.com/nltk/nltk/pull/2094 affected, one line change, should be trivial to resolve when addressing the failing build
:warning: https://github.com/nltk/nltk/pull/2043 affected, reformatted code is entirely replaced by a mostly new method, should be low to moderate amount of effort to resolve by taking the rewritten method from the PR
:heavy_check_mark: https://github.com/nltk/nltk/pull/2039 not affected because contains only code additions
:heavy_check_mark: https://github.com/nltk/nltk/pull/2018 not affected because contains mostly code additions
:question: https://github.com/nltk/nltk/pull/2009 multiple existing conflicts in a lot of code, hard to tell what the impact will be here
:heavy_check_mark: https://github.com/nltk/nltk/pull/1976 not affected
:heavy_check_mark: https://github.com/nltk/nltk/pull/1966 not affected
:heavy_check_mark: https://github.com/nltk/nltk/pull/1965 existing conflict, not affected

In short, out of 12 pull requests, 8 will not be affected, 3 will be affected in a way that is straightforward to resolve. Impact on 1 PR is hard to measure and remains unknown.

I will help mitigate any impact from this PR on other open PRs.

nltk/app/collocations_app.py
```diff
@@ -13,12 +13,35 @@
...
+    LEFT,
+    Label,
+    Menu,
+    OptionMenu,
+    SUNKEN,
+    Scrollbar,
+    StringVar,
+    Text,
+    Tk,
+)
```
alvations, line 30: I'm not sure whether it's okay but I don't necessary agree that listing one submodule per line here is useful.

nltk/book.py
```diff
@@ -58,25 +64,142 @@ def texts():
...
+    "of",
+    "Dashwood",
+    "had",
+    "long",
+    "been",
+    "settled",
+    "in",
+    "Sussex",
+    ".",
+]
```
alvations, line 81: This is not necessary esp. when it is clearer when the sentence tokens line max out the 80 chars as much as possible and then move to the next line.

nltk/app/collocations_app.py
purificant: The whole point of [https://github.com/ambv/black](url) is to avoid having conversations about preferred formatting style. From their readme:

_Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters._

More discussion on [reddit](https://www.reddit.com/r/Python/comments/8lmnzr/thoughts_on_black/).

If you prefer, whole files (`--exclude`) or blocks of text can be excluded from formatting with `# fmt: off` and `# fmt: on`

nltk/book.py
purificant: See my other comment about the philosophy of [black](https://www.reddit.com/r/Python/comments/8lmnzr/thoughts_on_black/).

If you prefer, existing formatting can be preserved with `# fmt: off` and `# fmt: on`

alvations: I understand the non-compromising part but is that really what we want is a question to NLTK dev and community rather than just pushing the PR through.

Personally, going through each line to preserve the format with `# fmt: off` might be as good as enforcing the style before merging. But that's just my personal view. Maybe other users and devs have thoughts that differs.

purificant: That's fair.

This PR is intended as a continuation of https://github.com/nltk/nltk/pull/1935, I would be equally happy with the original PR as well. Key issue for me personally is having consistency and significantly less noise from pep8 / linting violations because this results in higher productivity and easier maintainability for me.
[black](https://github.com/ambv/black) is chosen because of practicality, simplicity, ease of use and the fact that it produces a better diff than alternatives in my view. I would be equally happy with alternative tooling that achieve similar results.

Given the significant length of time that has passed since https://github.com/nltk/nltk/pull/1935 was first proposed in Jan 2018, I see this as a sign that community is broadly happy with reformatting the code base. If that is not the case, feel free to reject the proposed changes in whole or in part.

stevenbird:
I'm happy for us to proceed, but I don't have time to process the merge conflicts. Can we please pick a time to do this merge when @purificant has time to sort out the conflicts?

purificant:
@stevenbird Happy to help any time. Later today and on the weekend, evenings next week too, this should not be an issue.

The changes are quite simple, I can supply the exact code change for the 3 affected PRs as comments on the PR. But to remove the conflict directly in the PR would need the[ maintainer](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) permissions in order to be able to edit the PR by adding another commit to it.

stevenbird:
Ok, let's do this.

purificant:
Thank you for the merge @stevenbird. I left comments on all affected PRs.

stevenbird:
Thanks @purificant!

PR #2165, opened 2018-10-29
fix ResourceWarning: unclosed file

purificant:
Fix for issue https://github.com/nltk/nltk/issues/1928

This approach closes opened resources during garbage collection.

How to reproduce:

```bash
python -Wd -c "import nltk; print(nltk.corpus.stopwords.words('english'))"
```

### DIFF
--- a/nltk/test/unit/test_seekable_unicode_stream_reader.py
+++ b/nltk/test/unit/test_seekable_unicode_stream_reader.py
@@ -127,6 +127,13 @@ def test_reader_on_large_string():
             pass
 
 
+def test_reader_stream_is_closed():
+    reader = SeekableUnicodeStreamReader(BytesIO(b''), 'ascii')
+    assert reader.stream.closed is False
+    reader.__del__()
+    assert reader.stream.closed is True
+
+
 def teardown_module(module=None):
     import gc
 
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -1270,6 +1270,11 @@ class SeekableUnicodeStreamReader(object):
         """Return self"""
         return self
 
+    def __del__(self):
+        # let garbage collector deal with still opened streams
+        if not self.closed:
+            self.close()
+
     def xreadlines(self):
         """Return self"""
         return self


alvations (approved):

alvations:
@purificant the GC solution looks like a good alternative to the BufferReader proposed in #1928. 

Assuming that we always read the file sequentially within a single stream reading function, I don't see any foreseeable problems where the `__del__` closes abruptly, but I'm not exactly sure the behavior of the `__del__`. Do you anticipate any issues?

Lets also test this in the CI =)

purificant:
Hey @alvations, thanks for taking a look at the PR.

[`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__) will be called by GC when instance of the object is about to be destroyed. For example, for CPython this happens when there are no more references to the object left and code can no longer interact with the object because of this.

Scenarios where stream is partially read, then read some more at a later point in time are fine because instance will not be destroyed and `__del__` is never called. I don't anticipate any issues as this is performing cleanup on objects that are no longer needed.

While this code path is already implicitly covered through other tests, I've also added an explicit unit test for the new method.

nltk/test/unit/test_seekable_unicode_stream_reader.py
```diff
@@ -119,6 +119,14 @@ def _check(encoding, n=1000):
...
             pass
 
+
+def test_reader_stream_is_closed():
+    reader = SeekableUnicodeStreamReader(BytesIO(b''), 'ascii')
+    assert reader.stream.closed is False
+    reader.__del__()
+    assert reader.stream.closed is True
+
+
```
alvations, line 136: Awesome! This test should be sufficient =)

alvations (approved):
@stevenbird LGTM =) This should resolve #1928 and #1945.

purificant:
Conflict resolved and this is once again ready for merge.

stevenbird:
Thanks @purificant

apusarla:
Thanks @purificant

PR #2169, opened 2018-11-06
Update decorators.py

KrupaH:
More robust in case you somehow end up with "None" in your sys.path list somewhere


alvations (approved):

stevenbird:
Thanks @KrupaH

PR #2170, opened 2018-11-06
using sudo is dangerous

ppartarr:
`pip` can install packages to your home directory directly instead


alvations:
Agreed, but we should have some sort of reference also to virtual environment on top of using `--user`. 

----

This is a good time to update the docs on the installation,  I think we should have multiple installation instructions:

 1. Creating a virtual environment before using `pip install nltk`
 2. Installing native using `pip install --user nltk`
 3. Installing using conda, `conda install nltk` (esp. for Windows users)

alvations (approved):

stevenbird:
Thanks @ppartarr

Issue #2053, opened 2018-06-28
Remove the "twython library has not been installed" warning?


ataylor32:
I'm using nltk for sentiment analysis in a Django project. Every time I run a command with Django's `manage.py`, I get this output:

```
/var/www/example.com/venv-example/lib/python3.5/site-packages/nltk/twitter/__init__.py:20: UserWarning: The twython library has not been installed. Some functionality from the twitter package will not be available.
  warnings.warn("The twython library has not been installed. "
```

I also get emailed by cron since cron emails whenever output is generated. Could this warning just be in the nltk documentation or is it important for it to be in the code too?

alvations:
Try installing with the `twitter` option, e.g.

```
pip3 install -U nltk[twitter] 
```

Or if you like it fuss-free, all batteries-included:

```
pip3 install -U nltk[all]
```

mjalkio:
@alvations are you recommending that everyone install with Twitter?

For example, if I just do `from nltk.sentiment.vader import SentimentIntensityAnalyzer`, I'll get the error. I don't need any twitter functionality, I'm just getting the warning due to the way things are being imported (some of the Twitter functions seem to be used for `demo_` functions in `nltk.sentiment.util`).

alvations:
It's a warning so it's not breaking anything critical. I think it makes sense to install with twitter if you're using `nltk.sentiment`, at some point you might need the twython functionalities.

One could also do something like:

```
export PYTHONWARNINGS="ignore"
$ python
$ from nltk.sentiment.vader import SentimentIntensityAnalyzer
```

xulaus:
I am also finding this issue quite annoying. We don't want to install with the twitter option as our application will fairly likely never need it and we like to keep our docker images as slim as possible. We also don't want to turn off warnings as that might hide genuine issues from other modules.

It seems like the cause of this is `json2csv_preprocess` in `nltk/sentiment/util.py`, which in turn seems to only be used in the demos in that file. Perhaps a new `demos.py` file could be created, localising the dependency on the twitter option to that one file? That might represent a breaking API change depending on how you define the public API (I couldn't find information on this or where to find information in the CONTRIBUTING.md).

I'm happy to do the work if that solution is acceptable?

ithomas-glint:
> I am also finding this issue quite annoying. We don't want to install with the twitter option as our application will fairly likely never need it and we like to keep our docker images as slim as possible. We also don't want to turn off warnings as that might hide genuine issues from other modules.

[...]

PR #2175, opened 2018-11-12
decouple sentiment and twitter packages

purificant:
This PR addresses https://github.com/nltk/nltk/issues/2053

### How to reproduce:
```
>>> import twython
ModuleNotFoundError: No module named 'twython'
>>> import nltk
>>> from nltk.sentiment import SentimentIntensityAnalyzer
/home/foo/projects/nltk/nltk/twitter/__init__.py:22: UserWarning: The twython library has not been installed. Some functionality from the twitter package will not be available.
  "The twython library has not been installed. "
```

1 Ensure that `twython` is not installed.
2 Attempt to import a class from `sentiment` module.
3 See warning from `twitter` module.

### Current situation
`sentiment` package includes demo code that that makes use of utility functions from `twitter` package. This demo code is currently exposed through a user facing public api in `sentiment\util.py` module. Because of this, any `sentiment` package class that attempts to make use of functions in `util.py` will also cause the interpreter to load `twitter` package at runtime, which causes `twitter` specific code to run and generates the warning.

### Proposed change
`sentiment\util.py` is refactored to only depend on `twitter` when executing demo code.
Utility functions from `util.py` are still accessible to `sentiment` package without importing `twitter` at runtime while still preserving the existing public api for demo code.

### DIFF
--- a/nltk/sentiment/util.py
+++ b/nltk/sentiment/util.py
@@ -27,7 +27,6 @@ import nltk
 from nltk.corpus import CategorizedPlaintextCorpusReader
 from nltk.data import load
 from nltk.tokenize.casual import EMOTICON_RE
-from nltk.twitter.common import outf_writer_compat, extract_fields
 
 # ////////////////////////////////////////////////////////////
 # { Regular expressions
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @purificant

Issue #2090, opened 2018-08-21
Zipfile as corpus fails unless a subdirectory is appended to the filename


meldavis:
I am adding another issue because I could not reopen #1986.

Issue #1986 is still broken.   The fix did remove the exception; however, now the corpus fails to return any data.    Note that this issue applies to all corpus; I am using the Twitter Corpus as an example.

The issue is using a zip file as a corpus fails, unless a subdirectory is appended to the zip's filename.
Example: ```twitter_samples.zip``` will fail; but ```twitter_samples.zip/twitter_samples/``` will work.

Python: 3.7.0
NLTK: 3.3.0
OS: Windows 10 x64

### Appending a subdirectory to the zip filename
``` python
from nltk.corpus.reader import TwitterCorpusReader
c = TwitterCorpusReader(root='twitter_samples.zip/twitter_samples/', fileids="negative_tweets.json")
print(c.abspaths())
print(c.fileids())
```
This correctly reports
 ```
[ZipFilePathPointer('D:\\Projects\\corpus\\twitter_samples.zip', 'twitter_samples/negative_tweets.json')]
['negative_tweets.json']
```

### Using a zipfile as a corpus root produces no data
```
c = TwitterCorpusReader(root='twitter_samples.zip', fileids="twitter_samples/negative_tweets.json")
print(c.abspaths())
print(c.fileids())
```
This incorrectly reports an empty set: 
```
[]
[]
```

purificant:
@meldavis Proposed fix is here: https://github.com/nltk/nltk/pull/2176

stevenbird:
Resolved

PR #2013, opened 2018-04-29
Fix #1986 TwitterCorpusReader read zipfile error

fanjin-z:
#1986

Issue #1986, opened 2018-03-20
Zipfile as corpus raises exception unless subdirectory is appended to filename


meldavis:
Unable to use a zipfile as a corpus unless a subdirectory is appended to the zipfile name.  A root name like 'twitter_samples.zip' will raise an exception.  But 'twitter_samples.zip/twitter_samples/' will work.

Using python 3.6.4, NLTK 3.2.5, Windows.

#### Fails with exception
```
from nltk.corpus.reader import TwitterCorpusReader
TwitterCorpusReader(root='twitter_samples.zip', fileids="twitter_samples/negative_tweets.json")
```
#### This works
```
TwitterCorpusReader(root='twitter_samples.zip/twitter_samples/', fileids="negative_tweets.json")
```

## Traceback
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Apps\Tools\python\conda3\envs\nlp\lib\site-packages\nltk\corpus\reader\twitter.py", line 74, in __init__
    CorpusReader.__init__(self, root, fileids, encoding)
  File "C:\Apps\Tools\python\conda3\envs\nlp\lib\site-packages\nltk\corpus\reader\api.py", line 82, in __init__
    root = ZipFilePathPointer(zipfile, zipentry)
  File "C:\Apps\Tools\python\conda3\envs\nlp\lib\site-packages\nltk\compat.py", line 221, in _decorator
    return init_func(*args, **kwargs)
  File "C:\Apps\Tools\python\conda3\envs\nlp\lib\site-packages\nltk\data.py", line 506, in __init__
    (zipfile.filename, entry))
OSError: Zipfile 'C:\\Projects\\mfd\\nlp\\twitter_samples.zip' does not contain '.'
```

fanjin-z:
@meldavis @alvations I think the problem lies in here:
https://github.com/nltk/nltk/blob/a8446adace93893332d42ff0f564343748b7f83c/nltk/data.py#L491

If entry is empty(which it is in our case), it add a dot to entry, causing unintended error later. Here is an error showcase:
```
import os
import sys
entry = ''
entry = normalize_resource_name(entry, True, '/').lstrip('/')
entry
```
out
```
'.'
```

A simple fix could be changing this line:
https://github.com/nltk/nltk/blob/a8446adace93893332d42ff0f564343748b7f83c/nltk/data.py#L494
to
```
if entry != '.'
```

Shall I create a PR to fix it?

[...]

PR #2176, opened 2018-11-13
allow CorpusReader to load zip files without requiring a subfolder inside

purificant:
This PR addresses https://github.com/nltk/nltk/issues/2090

While the fix for issue https://github.com/nltk/nltk/issues/1986 did suppress the exception, it did not go far enough in enabling corpus reader to work with zip files without a folder inside - https://github.com/nltk/nltk/pull/2013


stevenbird:
Thanks @purificant

PR #2180, opened 2018-11-15
Cleanup

purificant:
Collection of various cleanups: redundant code, number of arguments and their use, legacy python support. Details are in individual commit messages.


stevenbird:
Thanks @purificant

PR #2187, opened 2018-11-22
add guards to next() calls in punkt.py

purificant:
This PR addresses another instance of https://github.com/nltk/nltk/issues/2148, except that due to the nature of the code, `RuntimeError` should only occur on empty / specially crafted input, which is why no errors have showed up in tests.

The PR adds guards around next() calls in generators as suggested in [PEP 479 section on backwards and forwards compatible code](https://www.python.org/dev/peps/pep-0479/#writing-backwards-and-forwards-compatible-code).


stevenbird:
Nice work, thanks @purificant

PR #2190, opened 2018-11-27
Reverting some changes from #2130

alvations:
There are some changes that are superfluous. Which makes the updates for `self.vocab` vs `self.vocabulary` and 'self.counts` vs `self.counter` confusing. c.f. #2130 

It's best to stick with the current default in `nltk.lm.api` for now and then re-open a PR for refactoring. 

Regarding `GoodTuring` addition in #2130 by @sahitpj , there might be implementation missing as reported  @Copper-Head 

We should either:

 1. Correct the implementation and add more tests 
 2. Remove the incomplete implementation
 3. Write more tests to check that the expected output from GoodTuring is correct

@Copper-Head @sahitpj , please advise.

### DIFF
--- a/nltk/lm/api.py
+++ b/nltk/lm/api.py
@@ -45,11 +45,15 @@ class Smoothing(object):
     Implements Chen & Goodman 1995's idea that all smoothing algorithms have
     certain features in common. This should ideally allow smoothing algoritms to
     work both with Backoff and Interpolation.
-
-    counter represents the number of counts for ngrams
     """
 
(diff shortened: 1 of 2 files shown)


iliakur:
@alvations Thanks for this!

I would recommend also removing the SGT class. It's not finished yet, but it shows up as part of the API and people will be tempted to use it, then discover that they can't, which isn't a great experience imho.

stevenbird:
Looks like spurious failures. Merging.

Issue #2195, opened 2018-11-30
Dozens of test failures from unchanged develop branch


NQNStudios:
I followed the instructions in `CONTRIBUTING.md` to setup up a development environment, and run the tests with `tox -e py36 > test_output.txt`. (Running python 3.6.6.)

Here are the full contents of test_output.txt:

[test_output.txt](https://github.com/nltk/nltk/files/2634537/test_output.txt)

Observations:
* Several deprecation warnings.
* Some resources, i.e. omw, must not have been installed when I called `python3 -m nltk.downloader tests`

Is this possibly related to the documentation being outdated as mentioned in #1853? Did I miss a step in the setup?

purificant:
@NQNStudios Having had a quick look at your test output log, most of the failures that I can see are due to missing nltk data. Try `python -m nltk.downloader all`, that should download the missing data and make all tests green.

NQNStudios:
That fixed most of the problems.
```
GLOB sdist-make: /home/nat/nltk/setup.py
py36 inst-nodeps: /home/nat/nltk/.tox/.tmp/package/1/nltk-3.4.zip
py36 installed: certifi==2018.11.29,chardet==3.0.4,colorama==0.4.1,coverage==4.5.2,idna==2.7,nltk==3.4,nose==1.3.7,numpy==1.15.4,oauthlib==2.1.0,pyparsing==2.3.0,python-crfsuite==0.9.6,rednose==1.3.0,requests==2.20.1,requests-oauthlib==1.0.0,scikit-learn==0.20.1,scipy==1.1.0,singledispatch==3.4.0.3,six==1.11.0,termstyle==0.1.11,text-unidecode==1.2,twython==3.7.0,urllib3==1.24.1
py36 run-test-pre: PYTHONHASHSEED='2805722241'
py36 runtests: commands[0] | pip install scipy scikit-learn
Requirement already satisfied: scipy in /home/nat/nltk/.tox/py36/lib/python3.6/site-packages (1.1.0)
Requirement already satisfied: scikit-learn in /home/nat/nltk/.tox/py36/lib/python3.6/site-packages (0.20.1)
Requirement already satisfied: numpy>=1.8.2 in /home/nat/nltk/.tox/py36/lib/python3.6/site-packages (from scipy) (1.15.4)
py36 runtests: commands[1] | python runtests.py
-........................................................-...................
.........-.....-.................-........-..................................
[...]

PR #2198, opened 2018-11-30
Fix CONTRIBUTING.md

NQNStudios:
This simply changes the documentation so other contributors won't run into my problem with #2195.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -240,6 +240,7 @@
 - Tim Gianitsos <https://github.com/timgianitsos>
 - Andrew Owen Martin
 - Adrian Ellis <https://github.com/adrianjellis>
+- Nat Quayle Nelson <https://github.com/nqnstudios>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer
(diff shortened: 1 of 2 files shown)


purificant:
Should the docs specify `all` for data downloads, or should we amend `tests` to actually include all the necessary data for tests?

Any thoughts @alvations?

NQNStudios:
The test_output.txt I uploaded to my issue report should contain all the
information needed to make that fix. If we do, we should change
CONTRIBUTING.md to note that when writing tests, the necessary files should
be added to the `tests` download list to prevent this from becoming an
evergreen bug.

purificant:
This is a valid issue, as any new contributor is likely to run into it by following the docs. Hat off to @NQNStudios for identifying the issue and proposing a fix.

If `python -m nltk.downloader tests` is replaced with `python -m nltk.downloader all` in the docs, will that essentially make `tests` obsolete? Why was only a subset of data installed for tests previously and not all of the available data?

NQNStudios:
Would you like me to measure how much file size is added by using `all` instead of `tests`? In any case, I'm also happy to update the `tests` resource list to add what's needed, or remove it if `all` is the way to go.

alvations:
@purificant @NQNStudios @Copper-Head yes, since the CI is already using `nltk.download('all')` https://github.com/nltk/nltk/blob/develop/tools/travis/install.sh#L10 , it makes sense that contributors should start testing with a similar environment as the CI. 

Until we find a better way to handle and separate test-needed data from all, downloading `all` nltk_data seems to be necessary for contributors to get through the testing with ease.

alvations (approved):

alvations:
Thanks @NQNStudios for raising this! Thanks @purificant for following up!

PR #2199, opened 2018-12-02
Added .editorconfig

KOLANICH:


rth:
I think there is a consensus against this https://github.com/pandas-dev/pandas/pull/24042 https://github.com/matplotlib/matplotlib/pull/12922

purificant:
Agreed with @rth and @Copper-Head, I don't see why this is needed. Writing code and then running it through `black -S .` before commit/push works better for me personally.

Issue #2203, opened 2018-12-11
Include/port SyllabiPy (which includes LegaliPy) for Syllable Tokenization


no-identd:
See here for the project by @henchc:

https://github.com/henchc/syllabipy

Here for the project homepage:

http://syllabipy.com/

And here for the research paper both originate from:
https://escholarship.org/uc/item/13c6h2z2

See PDF page 44 (document page 31) for a schematic of how it works.

Incorporating this would definitely require some refactoring (throwing out util.py to make the tokenizer freely chosable, for example) and properly abstracting the current hardcoding. @Yomguithereal seems to have already done this for their port of it into their Talisman library:

https://github.com/Yomguithereal/talisman/tree/master/src/tokenizers/syllables

However, Talisman uses Javascript - otherwise I'd just suggest including most of Talisman (albeit not all, since Talisman itself uses NLTK, so that'd become a tad bit too recursive.😉) in NLTK. 😉

Both Talisman & SyllabiPy/LegaliPy use the MIT license, which should have compatibility.

henchc:
Thanks, @no-identd , I'm happy to work on this over the next week or two. I'll refactor it best I can and make a PR, if you'd help me refine it further beyond that I'd appreciate it.

no-identd:
Huh. Didn't expect that! Very nice! Thanks! I'll try, but unfortunately, I lack the time to definitely commit to that - I've already barely had the time to open this issue. 

Also, and I only bring up since you replied here, on a somewhat off-topic, somewhat on-topic notion, which, a suggestion:
You might want to take a look at #2202. I suspect the tokenizer I suggested there there could improve your scansion project a lot. :)

alvations:
Should be resolving soon when #2205 is merged =)

Thanks @henchc and @no-identd !

juliangilbey:
Hi @henchc and @alvations and @no-identd,

Is there any intention to one day incorporate LegaliPy as well?  I've just come across another library (a distant dependency of the current Spyder) which is still depending on syllabipy because of LegaliPy.

Best wishes, Julian

henchc:
Hi @juliangilbey , I wanted to make a PR for this but never got around to it. Coincidentally, I actually have some time this week so let me see if I can get one started in the next day or so. I'll @ you on the thread.

henchc:
Made PR here: https://github.com/nltk/nltk/pull/2652

juliangilbey:
Thanks @henchc !

[...]

PR #2205, opened 2018-12-16
add SSP to tokenizers

henchc:
Addresses one part of [#2203](https://github.com/nltk/nltk/issues/2203). If this can get refined and merged I can work on the Legality Principle tokenizer.


alvations:
Thank you for the suggestion @no-identd. Thank you in advance for contributing too, @henchc 

Since it looks like it can be useful but we lack knowledge of how it works,  what the limitations are and how to maintain it if this is merged into NLTK.

So I've just a few question to ask before we proceed to review. 

   - Am I right to say that the current implementation in the PR only supports English syllabification of word?

   - What is the intended input to the `tokenize()` function? A sentence string? Or a single word. If it's a sentence string, how does it differentiate between the word and the syllable boundaries? 

   - If we want to support more languages, where do we find a list of `sonority_hierarchy` for other languages? 

   - What are the languages that the principle covers? Only languages with romanizable alphabets? Or would it also work on something like Chinese pinyin? 

   - Is space necessary for the implementation to work? For example, the algorithm is given romanized Japanese without spaces would it still work? 

  - Is there a canonical reference publication that shows that the  `Sonority Sequencing Principle` works for languages outside of German? Also, is there a specific chapter from ["Lehrbuch der Phonetik."](https://ia902609.us.archive.org/23/items/lehrbuchderphone00jespuoft/lehrbuchderphone00jespuoft_bw.pdf) that users/readers of the code should refer to to understand how it works? Citing the whole 200+ pages textbook might not be optimal for something that isn't mainstream. 

Sorry in advance for the multiple question. They are asked so that we better understand what we merge to the codebase and document it in the code docstring with more specificity. And we know how to maintain or extend it, if needed.

henchc:
Thanks @alvations ! And no worries, I'm happy to document whatever is necessary.

- Am I right to say that the current implementation in the PR only supports English syllabification of word?

> The default hierarchy in the PR will only support English, and not as fine-grained as an IPA translation would. The SSP is a universal syllabification algorithm, but that doesn't mean it performs equally across languages. [Wikipedia](https://en.wikipedia.org/wiki/Sonority_Sequencing_Principle) actually has a really good summary. Relatedly, [Bartlett et al.](http://www.anthology.aclweb.org/N/N09/N09-1035.pdf) is a good benchmark for English accuracy if utilizing IPA (see their hierarchy on pg. 311). The idea with this PR is to allow users to provide the hierarchy for whatever language their working with. The implementation @no-identd first found was my implementation for medieval German. This is definitely more targeted for NLP researchers as opposed to an out-of-the-box solution for production syllabification. Though as the paper shows, the accuracy is quite high from this algorithm alone provided an accurate hierarchy.

- What is the intended input to the tokenize() function? A sentence string? Or a single word. If it's a sentence string, how does it differentiate between the word and the syllable boundaries?

> Intended input is any string the user wishes to be broken into syllables. For word-based languages this would be a word. For other languages it could be any sequence of phonemes. So it does not differentiate between word boundaries, it simply takes a string *without* spaces and will syllabify it. For logographic/ideographic languages this would need to be translated out of the script first. 

- If we want to support more languages, where do we find a list of `sonority_hierarchy` for other languages?

> Good question! I'm not sure of a resource with all language hierarchies. My suggested approach would be to go to the wikipedia page for the phonology of a language, e.g. [Turkish](https://en.wikipedia.org/wiki/Turkish_phonology) and then create the hierarchy from the consonant and vowel tables. As the [Bartlett et al.](http://www.anthology.aclweb.org/N/N09/N09-1035.pdf) paper discovers (as I did in my research), a strict and large hierarchy is not always most accurate. It's sometimes better to collapse certain levels.

- What are the languages that the principle covers? Only languages with romanizable alphabets? Or would it also work on something like Chinese pinyin?

> The accuracy varies by language. Generally, if the script is converted to phonemes in IPA and an IPA hierarchy is used for the language it can be quite accurate for most languages. As stated above, logographic/ideographic scripts would have to be converted as they are not phoneme based to any degree (English, e.g., is also not a 1-1 orthography to phonology, but the given hierarchy approximates it).

- Is space necessary for the implementation to work? For example, the algorithm is given romanized Japanese without spaces would it still work?

 > See input response above, but no space should be included in the input to `.tokenize()`

- Is there a canonical reference publication that shows that the Sonority Sequencing Principle works for languages outside of German? Also, is there a specific chapter from "Lehrbuch der Phonetik." that users/readers of the code should refer to to understand how it works? Citing the whole 200+ pages textbook might not be optimal for something that isn't mainstream.

> I tried citing the earliest reference I am aware of, but the principle is widely discussed in the literature. A common citation for its explanation is Selkirk: *Selkirk, E. (1984). On the major class features and syllable theory. In Aronoff & Oehrle (eds.) Language Sound Structure: Studies in Phonology. Cambridge: MIT Press. 107-136.* As for Jesperson, it would be [Chapter 13, Silbe, pp. 185-203](https://archive.org/details/lehrbuchderphone00jespuoft/page/184)

nltk/test/unit/test_tokenize.py
```diff
@@ -17,6 +17,7 @@
     TweetTokenizer,
     StanfordSegmenter,
     TreebankWordTokenizer,
+    SonoritySequencingPrinciple,
```
alvations: Lets call it `SyllableTokenizer` instead that'll be clearer.

nltk/tokenize/__init__.py
```diff
@@ -88,6 +88,7 @@
 from nltk.tokenize.treebank import TreebankWordTokenizer
 from nltk.tokenize.util import string_span_tokenize, regexp_span_tokenize
 from nltk.tokenize.stanford_segmenter import StanfordSegmenter
+from nltk.tokenize.ssp import SonoritySequencingPrinciple
```
alvations: How about `from nltk.tokenize.sonority_sequencing import SyllableTokenizer` ?

nltk/tokenize/ssp.py
```diff
@@ -0,0 +1,142 @@
...
+
+Importantly, if a custom hiearchy is supplied and vowels span across more than
+one level, they should be given separately to the `vowels` attribute.
+"""
+
+from __future__ import unicode_literals
+from nltk.tokenize.api import TokenizerI
+
+
+class SonoritySequencingPrinciple(TokenizerI):
```
alvations: Rename this to `SyllableTokenizer`.

nltk/tokenize/ssp.py
```diff
@@ -0,0 +1,142 @@
...
+    '''
+    Syllabifies words based on the Sonority Sequencing Principle (SSP)
+
+    >>> SSP = SonoritySequencingPrinciple()
+    >>> SSP.tokenize('justification')
+    ['jus', 'ti', 'fi', 'ca', 'tion']
+    '''
+
+    def __init__(self, sonority_hierarchy=False):
```
alvations: Lets specify the language here `def __init__(self, lang='en', snority_hierarchy=False):`

nltk/tokenize/ssp.py
```diff
@@ -0,0 +1,142 @@
...
+        '''
+        Sonority hierarchy should be provided in descending order.
+        If vowels are spread across multiple levels, they should be
+        passed assigned self.vowels var together.
+        '''
+
+        if not sonority_hierarchy:
```
alvations: `if not sonority_hierarchy and lang='en':`

nltk/tokenize/ssp.py
```diff
@@ -0,0 +1,142 @@
...
+        '''
+        Ensures each syllable has at least one vowel.
+        If the following syllable doesn't have vowel,
+        add it to the current one.
+        '''
+        valid_syllables = []
+        front = ""
+        for i, syllable in enumerate(syllabified):
+
+            if not any(char in self.vowels for char in syllable):
```
alvations: This might be slow if the list of syllable is long. Maybe a regex here would be faster.

nltk/tokenize/ssp.py
```diff
@@ -0,0 +1,142 @@
...
+                if len(valid_syllables) == 0:
+                    front += syllable
+                else:
+                    valid_syllables = valid_syllables[
+                        :-1] + [valid_syllables[-1] + syllable]
```
alvations: Don't break lines in this manner, esp. when it's indicating string slice indices.

nltk/tokenize/ssp.py
```diff
@@ -0,0 +1,142 @@
...
+            if i == 0:  # if it's the first char, append automatically
+                syllable += c
+
+            else:
+                # main algorithm
+                if (i < len(self.syllables_values) - 1):
+
+                    # these cases trigger syllable break
+                    if v == self.syllables_values[i + 1][1] and \
+                            v == self.syllables_values[i - 1][1]:
```
alvations: This is a good indication that looping through a trigrams of `self.syllables_values` would be good here.

alvations (changes requested):
Happy New Year! 

Sorry for the delayed review. The class needs some refactoring and more importantly we might need to do some profiling and check whether the loops and if-else can be replace with more terse regex or ngrams operations to speed things up.

henchc:
@alvations  I think I've addressed your comments, also added some more details to the information in the header. I'm not sure if there's a particular n-gram function you wanted me to use, so let me know if that section should be changed. Thanks!

nltk/tokenize/sonority_sequencing.py
```diff
@@ -0,0 +1,145 @@
...
+    def assign_values(self, token):
+        '''
+        Assigns each phoneme its value from the sonority hierarchy
+        '''
+        self.syllables_values = []
+        for c in token:
+            try:
+                self.syllables_values.append((c, self.phoneme_map[c]))
+            except KeyError:
+                print("Warning: '{}' not defined in hierarchy".format(c))
```
alvations: What is the expect behavior for this exception? 

Raise and break the function?  Or silently let it go pass and perhaps raise a warning? 

If raising a warning, should the syllable values for the character that's not in the hierarchy be still added? Or leave it and skip the character (i.e. current behavior in the code)?

nltk/tokenize/sonority_sequencing.py
```diff
@@ -0,0 +1,145 @@
...
+        self.phoneme_map = {}
+        for i, level in enumerate(sonority_hierarchy):
+            for c in level:
+                self.phoneme_map[c] = len(sonority_hierarchy) - i
+
+    def assign_values(self, token):
+        '''
+        Assigns each phoneme its value from the sonority hierarchy
+        '''
+        self.syllables_values = []
```
alvations: Hmmm, this isn't a good use of `self`. Since for every token the `syllables_values` resets and the content changes, it doesn't belong to the `SyllableTokenizer` variable. Try returning `syllable_values` instead.

```python
def assign_values(self, token):
    syllable_values = []
    ...
    return syllable_values
```

It should just be `syllables_values` here without the self. And then in the `tokenize()` function, it should be:

```python
syllables_values = self.assign_values(token)
```

alvations:
@henchc I hope you don't mind that I took the liberty to change the `trigrams` iterations. 

I think now we just need to clarify the `try-except` in `assign_values()` and we should be clear to merge =)

nltk/tokenize/sonority_sequencing.py
```diff
@@ -0,0 +1,141 @@
...
+                                  'zvsf',  # fricatives.
+                                  'bcdgtkpqxhj'  # stops.
+                                  ]
+
+        self.vowels = sonority_hierarchy[0]
+        self.phoneme_map = {}
+        for i, level in enumerate(sonority_hierarchy):
+            for c in level:
+                self.phoneme_map[c] = len(sonority_hierarchy) - i
+
```
henchc, line 83: add
```suggestion
self.phoneme_map[c.upper()] = len(sonority_hierarchy) - i
```
(see thought in PR thread)

henchc:
Thanks, @alvations ! And thanks for cleaning up the n-gram function, looks much better :) .

Your question about the `try:` `except:` is a good one. I'm trying to think from the perspective of the most common use-case. I would think this is a user who just used `nltk.word_tokenize` and is sending the words through the syllabifier. 

1. First note in this thought exercise is that we should add the `.upper()` character set to each string in the current hierarchy so that upper case characters won't raise an error/warning. I'm thinking this could be added at [line 68](https://github.com/nltk/nltk/pull/2205/files#r246221923) since capitalization should not affect the sonority. The one issue I see is that the German `'ß'.upper()` goes to `"SS"` for some reason, but the single `"s"` should be in the original anyway so that will still cover `"S"` and the extra key shouldn't cause a problem.

2. Second note is that for an English use case directly after tokenization I imagine the main undefined characters being punctuation or foreign words with accents. For the most accurate return, I believe we should stop the function if it encounters an unknown character that is not punctuation (e.g. "déjà-vu"). Otherwise, I believe punctuation is safe to skip.

Does that make sense? I'm happy to make the changes if you agree, but I think we could go either way here and I'd defer to your sense for the library's standard practice.

alvations:
Hmmm, lets try something like this before the loop in the `assign_value()`:

```python
import unicodedata 

token = 'déjà-vu'
unicodedata.normalize('NFKD', token.upper().lower()).encode('ascii', 'ignore').decode()

token='Vosßtraße'
unicodedata.normalize('NFKD', token.upper().lower()).encode('ascii', 'ignore').decode()
```

I wonder whether `str.encode('ascii', 'replace')` would be a better option. I'm not sure how NLTK users would use this so I think we should choose `'replace'` or `'ignore'` and just see how users react to the feature =)

-----

Meanwhile, please add the param documentation in the `___init__()` docstring:

```python
def __init__(...):
        """
        :param lang: Language parameter, default is English, 'en'
        :type lang: str
        :param sonority_hierarchy: Sonority hierarchy according to the Sonority Sequencing Principle.
        :type sonority_hierarchy: list(str)
        """
```

nltk/tokenize/sonority_sequencing.py
alvations: 👍 this should be done to handle the `ß` issue. Would it be better to:

```python
self.phoneme_map[c.upper().lower()] = len(sonority_hierarchy) - i
```

alvations: Hmmm but actually that complicates the access, you can loop through character by character after `ß -> ss` =)

henchc:
I'm not sure `normalize` or `lower` is what we want, as that would transform the input. Why not let the input remain in its case with the special characters? I think we can just extend the hierarchy to include the `upper` form of the hierarchy, and warn about missing characters except punctuation? It could assign a default value of vowel and warn, so the user could add it to the hierarchy if they wanted.

henchc:
I think my latest commit would be my suggested functionality, let me know what you think!

no-identd:
German native speaker question:
Why remap `ß` to `ss` instead of `sz`?

henchc:
Well we're not proposing the mapping it's just what python does with `"ß".upper()` . I was just calling that out as an edge case. I'm not sure why they would change it at all, especially since it messes with the `len`, but the "ß" as far as I know is to follow long vowels while "ss" follow short, though I think "ss" has begun to dominate beyond this context. "ẞ" was also used beyond that context before ("ich muss" now instead of "ich muß" was common before the Rechtschreibreform 1996). 

As for "sz", I think that was just orthographical when the "ß" first surfaced and the "sz" and "ss" sounds were converging in the language.

henchc:
@alvations  any updates here?

stevenbird:
@alvations are there any outstanding changes you have requested?

alvations (approved):

alvations:
@henchc sorry for the really long delay, was busy with some other stuff. 

I've made some changes w.r.t. handling punctuations because previously the implementation chomps up the punctuations within the token and when the punctuation is a token itself, in the example:

```python
from nltk import word_tokenize
SSP = SyllableTokenizer()
text = "This is a foobar-like sentence."
print([SSP.tokenize(token) for token in word_tokenize(text)])
```

@henchc maybe we can handle the edge cases after the merge and see what users want to use the syllable tokenizer on =)

@stevenbird Once the test passes, it should be good to merge =)

henchc:
No worries, and thanks, @alvations ! That makes sense to me, and yes, I'm happy to monitor and address edge cases, :)

stevenbird:
Thanks @henchc and @alvations !

PR #2207, opened 2018-12-17
Reverting #2130

alvations:
Removing the GoodTuring implementation since we hasn't gotten a correction in #2190 from @sahitpj and from #2130 it lacked proper testing. It's best to remove it to prevent users from unwanted gotchas.


iliakur:
@alvations thanks a ton for this!

@sahitpj I'm sorry about all the confusion. When you get a chance, please do open a new PR :)

PR #2208, opened 2018-12-20
remove duplicate/redundant imports

purificant:
This PR is small cleanup of duplicate or unused imports.

+ `import nltk` in `find_deprecated.py` is unused.
+ All other imports already exist at the top of respective files and are therefore duplicates.


stevenbird:
Thanks @purificant

PR #2211, opened 2019-01-06
Update Travis CI matrix

kevin-brown:
- [x] Remove sudo from Python 3.7 builds
- [ ] Test on Python 3.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,16 +32,11 @@ matrix:
     - python: '3.6'
       env: TOXENV="py-travis"
 
-    - python: '3.6' # To be removed when 3.7 doesn't need sudo.
-      env: TOXENV="py-travis-third-party"
-      
     - python: '3.7'
       env: TOXENV="py-travis"
-      sudo: required #FIXME: To be removed once, Travis supports 3.7 properly.
-      
+
     - python: '3.7' # Only runs checks on 3rd party for latest Python distribution
       env: TOXENV="py-travis-third-party"
-      sudo: required #FIXME: To be removed once, Travis supports 3.7 properly.
 
 before_install:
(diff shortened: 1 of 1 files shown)


alvations:
@kevin-brown Thanks for the contribution! 

I think py3.4 was purposely dropped to minimize tests and also encourage people to move on to newer Python3 versions. Could you revise your commits and remove the Py3.4 tests?

kevin-brown:
I added Python 3.4 to the testing matrix because it looks like NLTK officially supports Python 3.4. While I'm in favor of removing it if it's no longer supported, we should probably also remove it from the other places where it is referenced:

* https://github.com/nltk/nltk/blob/399343024888e73e8714203738ea826e1224489d/setup.py#L54
* https://github.com/nltk/nltk/blob/399343024888e73e8714203738ea826e1224489d/setup.py#L83
* https://github.com/nltk/nltk/blob/399343024888e73e8714203738ea826e1224489d/web/install.rst#L4

Otherwise I'm hesitant to recommend that we don't test on Python 3.4 when it's a version of Python that NLTK claims to officially support.

alvations:
@stevenbird could we drop support for Python 3.4? If so, @kevin-brown do you mind removing that from the `setup.py` and `install.rst`?

stevenbird:
@kevin-brown: ready for you to drop the change concerning Python 3.4 from your PR

kevin-brown:
Sorry for the delay. I dropped Python 3.4 so now it's just the changes to the Python 3.7 builds.

alvations (approved):
Removing sudo from Travis =)

stevenbird:
Thanks @kevin-brown, @alvations

PR #2216, opened 2019-01-16
Simplify code and fix inconsistency for NomBank

zhaoyanpeng:
simplify code: `self._framefiles` is redundant because `self._fileids` can be used to refer to a list of frame files, which is similar to the use of `self._fileids` (referring to a list of parse files) in `BracketParseCorpusReader`. Besides, `[nomfile, nounsfile] + framefiles` in line 66 is obscure. It mixes up `nomfile`, `nounsfile`, and `framefiles.` Considering there are `self._nomfile` and `self._nounsfile` referring to `nomfile` and `nounsfile` respectively, it is better to remove `nomfile` and `nounsfile` from `self._fileids`.

fix inconsistency: as is described in the doc, `root` is of `str` type normally. In the old code, however, errors will be raised from `find_corpus_fileids` in line 63 when `framefiles` is as well a string. It is because this function expects `root` to be an instance of `PathPointer`. The modified code can resolve this problem.


alvations:
Thanks @zhaoyanpeng for the patches! 

Is there a way to test this? Is it right that the nombank requires access to the full PTB?

zhaoyanpeng:
@alvations  I am not sure what a test for the code simplification part looks like. The patches aim to provide a more clear logic.

For the inconsistency issue, can we add a test which uses _string_types_ `root` and `framefiles` to instantiate `NombankCorpusReader`? In this case, _string_types_ is valid according to the parameter descriptions and no errors will occur with the patches.

Yes, the nonbank is dependent on the PTB since it is a kind of annotation of the PTB. If we only want to test it, we can use the portion of the PTB available in the NLTK data.

alvations:
Did a quick check on the identity, there's two files that aren't framefiles. 

```python
>>> from nltk.corpus import nombank
>>> len(nombank._framefiles), len(nombank.fileids())
(4705, 4707)
>>> set(nombank.fileids()).difference(nombank._framefiles)
{'nombank.1.0.words', 'nombank.1.0'}
```

Yes looks like `_framefiles` is a subset of `fileids()`. Let me try to add something =)

alvations:
It looks like you've removed the regex reading into the framefiles https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/nombank.py#L61 from the `LazyCorpusLoader()` here: https://github.com/nltk/nltk/blob/develop/nltk/corpus/__init__.py#L379

I've put it back so that the `fileids()` reading exists. So that users can still specify only the subset that they want to. Otherwise the `fileids()` would include the readme files and other things that are not the `.xml`.

Also, added some tests to make sure the numbers are right as well as some template for anyone to extend the unittest.

Thanks @zhaoyanpeng for the contribution and confirming the info on fileids and framefiles!

alvations:
This should LGTM once the test passes. Thanks again @zhaoyanpeng !

nltk/corpus/reader/nombank.py
```diff
@@ -58,6 +58,11 @@ def __init__(
             corresponding to this corpus.  These parse trees are
             necessary to resolve the tree pointers used by nombank.
         """
+        
+	    # If framefiles is specified as a regexp, expand it.
+        if isinstance(framefiles, string_types):
```
zhaoyanpeng: Hi @alvations, I am not sure if we need to expand `framefiles` here when it is a regexp since `framefiles` will be expanded again from the following `CorpusReader.__init__(self, root, framefiles, encoding)`. Correct me if I am wrong.

zhaoyanpeng, line 63: @alvations, is it necessary to expand `framefiles` here when it is a regexp? I think it will be expanded again from the following `CorpusReader.__init__(self, root, framefiles, encoding)`.

stevenbird: @alvations: please let us know if you have an opinion here

stevenbird:
Thanks @zhaoyanpeng and @alvations

Issue #2196, opened 2018-11-30
Programmatic access to collocation/concordance lists


NQNStudios:
The functions `nltk.Text.collocations()` and `nltk.Text.concordances()` print their output directly to the console, but I would like to be able to access and manipulate those lists through code. I'd propose adding helper functions `get_collocations()` and `get_concordances()` that return lists of objects (`ConcordanceInfo` and `CollocationInfo`, or maybe just tuples) that contain raw information, and refactoring the existing methods to print output based on those helper results.

I'll open a PR soon when I start working on this.

alvations:
@NQNStudios There's a [`Text.concordance_list()`](https://github.com/nltk/nltk/blob/develop/nltk/text.py#L380) function from #1333 and #1910

For example:

```python
from nltk.corpus import gutenberg
from nltk.text import Text

corpus = gutenberg.words('melville-moby_dick.txt')
text = Text(corpus)
con_list = text.concordance_list("monstrous")
```

[out]:

> ConcordanceLine(left=['Whales', 'and', 'other', 'monsters', 'of', 'the', 'sea', ',', 'appeared', '.', 'Among', 'the', 'former', ',', 'one', 'was', 'of', 'a', 'most'], query='monstrous', right=['size', '.', '...', 'This', 'came', 'towards', 'us', ',', 'open', '-', 'mouthed', ',', 'raising', 'the', 'waves', 'on', 'all', 'sides'], offset=899, left_print='ong the former , one was of a most', right_print='size . ... This came towards us , ', line='ong the former , one was of a most monstrous size . ... This came towards us , ')

Each `ConcordanceLine` object contains the left, right side context of  the query word, the line where query word occurs and the offset is the position of the start of the query word. The `left_print` and `right_print` is the words that should be printed in the concordance.

----

Collocations on the other hand relies heavily on hard-coded `BigramCollocationFinder`, `TriramCollocationFinder` and `QuadgramCollocationFinder`

Example:

```python
 finder = BigramCollocationFinder.from_words(tokens, window_size)
 finder.apply_freq_filter(2)
 finder.apply_word_filter(lambda w: len(w) < 3 or w.lower() in ignored_words)
 bigram_measures = BigramAssocMeasures()
 collocations = finder.nbest(bigram_measures.likelihood_ratio, num)
```

So it might not be as easy as it seems. 

[...]

PR #2226, opened 2019-02-07
Updating collocations.py docs

NQNStudios:
I wanted to add features to collocation retrieval (per my issue #2196) and along the way, found it helpful to write some comments and update the docs in `collocations.py`. This PR is for my doc fixes only, and contains no code modifications. I've annotated two lines with `FIXME` comments because I couldn't figure out if they were bugs or not. If the code is meant to work the way I think it should, I can make the suggested code changes as well. Otherwise, I can update the docs further to indicate why things are the way they are.

### DIFF
--- a/nltk/collocations.py
+++ b/nltk/collocations.py
@@ -18,10 +18,12 @@ will then requiring filtering to only retain useful content terms. Each ngram
 of words may then be scored according to some association measure, in order
 to determine the relative likelihood of each ngram being a collocation.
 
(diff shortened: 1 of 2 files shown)


stevenbird:
@NQNStudios thanks for your contribution. The inline comments such as ``count this word`` run counter to our practice across the toolkit of writing literate code. We mainly use inline comments to document anything that is not fairly evident from the code.

PR #2233, opened 2019-02-12
Feature/corenlp server options

matanrak:
Due to the stdout and stderr params being 'pipe' when creating a CoreNLP server, the server would become unresponsive and each request to it will timeout once the pipe has been filled (As was stated in a #TODO by @dimazest 2 years ago).

So I have added made those params configureable :D

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -113,7 +113,7 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
 
     stdin = subprocess_output_dict.get(stdin, stdin)
     stdout = subprocess_output_dict.get(stdout, stdout)
-    stderr = subprocess_output_dict.get(stderr, stdout)
+    stderr = subprocess_output_dict.get(stderr, stderr)
 
     if isinstance(cmd, string_types):
         raise TypeError('cmd should be a list of strings')
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -88,7 +88,7 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
         standard input, standard output and standard error file
         handles, respectively.  Valid values are ``subprocess.PIPE``,
         an existing file descriptor (a positive integer), an existing
-        file object, and None.  ``subprocess.PIPE`` indicates that a
+        file object, 'pipe', 'stdout', 'devnull' and None.  ``subprocess.PIPE`` indicates that a
         new pipe to the child should be created.  With None, no
         redirection will occur; the child's file handles will be
         inherited from the parent.  Additionally, stderr can be
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -108,12 +108,13 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
 
     :raise OSError: If the java command returns a nonzero return code.
     """
-    if stdin == 'pipe':
-        stdin = subprocess.PIPE
-    if stdout == 'pipe':
-        stdout = subprocess.PIPE
-    if stderr == 'pipe':
-        stderr = subprocess.PIPE
+
+    subprocess_output_dict = {'pipe': subprocess.PIPE, 'stdout': subprocess.STDOUT, 'devnull': subprocess.DEVNULL}
+
+    stdin = subprocess_output_dict.get(stdin)
+    stdout = subprocess_output_dict.get(stdout)
+    stderr = subprocess_output_dict.get(stderr)
+
     if isinstance(cmd, string_types):
         raise TypeError('cmd should be a list of strings')
 
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -111,9 +111,9 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
 
     subprocess_output_dict = {'pipe': subprocess.PIPE, 'stdout': subprocess.STDOUT, 'devnull': subprocess.DEVNULL}
 
-    stdin = subprocess_output_dict.get(stdin)
-    stdout = subprocess_output_dict.get(stdout)
-    stderr = subprocess_output_dict.get(stderr)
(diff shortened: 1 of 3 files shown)


matanrak:
Does anyone know why the tests failed on 3.7?

purificant:
```No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.```

It's worth re-running the build to see if it will pass. The issue could be on Travis end, but it could also be code related.

File: nltk/parse/corenlp.py [hunk 1]
```diff
@@ -105,7 +105,12 @@ def __init__(
         self.corenlp_options = corenlp_options
         self.java_options = java_options or ['-mx2g']
 
-    def start(self):
+    def start(self, stdout='devnull', stderr='devnull'):
```
purificant, line 108: Will this change start suppressing previously visible output?

matanrak:
> `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`
> 
> It's worth re-running the build to see if it will pass. The issue could be on Travis end, but it could also be code related.

How do I re-run it? its my first nltk contribution

nltk/parse/corenlp.py
matanrak: So you would prefer having 'pipe' as the default?  the problem with it is that the pipe will get filled quickly (for me it was 882 requests with random test until it was full)

purificant:
> > `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`
> > It's worth re-running the build to see if it will pass. The issue could be on Travis end, but it could also be code related.
> 
> How do I re-run it? its my first nltk contribution

I would suggest running the build on your fork of nltk through Travis. Make an account on Travis and it should pick up your fork and allow running this build for your branch. Alternatively an admin can comment on this PR with `[CI:retest]` which will trigger a rebuild, this is a custom feature of nltk repo.

nltk/parse/corenlp.py
matanrak: Ohhh I just now got what you were asking. No, the output was not visible through python before and it has not changed.

matanrak:
> I would suggest running the build on your fork of nltk through Travis. Make an account on Travis and it should pick up your fork and allow running this build for your branch. Alternatively an admin can comment on this PR with `[CI:retest]` which will trigger a rebuild, this is a custom feature of nltk repo.

Running it now :D 
https://travis-ci.org/matanrak/nltk/builds/492575604

matanrak:
> Running it now :D
> https://travis-ci.org/matanrak/nltk/builds/492575604

It all passed 😃 
https://travis-ci.org/matanrak/nltk/builds/492575604

nltk/internals.py
```diff
@@ -108,12 +108,13 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
 
     :raise OSError: If the java command returns a nonzero return code.
     """
-    if stdin == 'pipe':
```
dimazest, line 111: It might worth updating the docstring, as it seems to me that the code and docs are not in sync. Specifically, the docs above are saying:
```
Valid values are ``subprocess.PIPE``,
an existing file descriptor (a positive integer), an existing
file object, and None.
```

But the function is happy to take strings in the form of `"pipe"`, which is not mentioned there.

nltk/internals.py
```diff
@@ -108,12 +108,13 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
...
-        stdin = subprocess.PIPE
-    if stdout == 'pipe':
-        stdout = subprocess.PIPE
-    if stderr == 'pipe':
-        stderr = subprocess.PIPE
+
+    subprocess_output_dict = {'pipe': subprocess.PIPE, 'stdout': subprocess.STDOUT, 'devnull': subprocess.DEVNULL}
+
+    stdin = subprocess_output_dict.get(stdin)
```
dimazest: I would replace this (and the rest accordingly) with
```python
stdin = subprocess_output_dict.get(stdin, stdin)
```
so if an integer or something else is passed, it is used.

nltk/parse/corenlp.py
```diff
@@ -105,7 +105,12 @@ def __init__(
...
+        """ Starts the CoreNLP server
+
+        :param stdout: Specifies the subprocess's loc Either 'devnull', 'stdout' or 'pipe'
```
dimazest: These might need to be rewritten, to something like:
```
Specifies where CoreNLP output is redirected. Valid values are 'devnull', 'stdout', 'pipe'.
```

Maybe, it worth explaining what the intention of the  `devnull` value is, as it might not be clear.

dimazest (changes requested):
Good work!

But be careful with starting CoreNLP server within Python. Starting a standalone CoreNLP process in a shell is way more flexible.

nltk/internals.py
matanrak: Agreed.

nltk/parse/corenlp.py
matanrak: I also agree with this! 😄

nltk/internals.py
matanrak: `    :param stdin, stdout, stderr: Specify the executed programs'
        standard input, standard output and standard error file
        handles, respectively.  Valid values are ``subprocess.PIPE``,
        an existing file descriptor (a positive integer), an existing
        file object, 'pipe', 'stdout', 'devnull' and None.  ``subprocess.PIPE`` indicates that a
        new pipe to the child should be created.  With None, no
        redirection will occur; the child's file handles will be
        inherited from the parent.  Additionally, stderr can be
        ``subprocess.STDOUT``, which indicates that the stderr data
        from the applications should be captured into the same file
        handle as for stdout.`

Is this what you had in mind?

dimazest: 👍

dimazest (approved):

purificant (approved):

stevenbird:
Thanks @dimazest, @purificant, @matanrak

Issue #2239, opened 2019-02-23
Plot methods don't return Axes object


nick-ulle:
The `.plot()` methods on `FreqDist` and `ConditionalFreqDist` currently return `None`. This makes it difficult to customize the resulting plot.

In addition, returning `None` here is counterintuitive to users coming from other scientific computing packages. As an example, Matplotlib, Pandas, and Seaborn, all typically return an `Axes` object from their plotting methods.

I propose changing the `.plot()` methods to return an `Axes` object.

alvations:
#2240 should have fixed the plots from `nltk.probability`

But there's a few more in NLTK, for reference:

```
$ grep matplotlib nltk/**/*.py
nltk/app/__init__.py:        from matplotlib import pylab
nltk/app/__init__.py:            "nltk.app.wordfreq not loaded " "(requires the matplotlib library)."
nltk/app/wordfreq_app.py:from matplotlib import pylab
nltk/draw/dispersion.py:        from matplotlib import pylab
nltk/draw/dispersion.py:            'The plot function requires matplotlib to be installed.'
nltk/draw/dispersion.py:            'See http://matplotlib.org/'
nltk/parse/dependencygraph.py:        from matplotlib import pylab
nltk/sentiment/util.py:        import matplotlib.pyplot as plt
nltk/sentiment/util.py:            'The plot function requires matplotlib to be installed.'
nltk/sentiment/util.py:            'See http://matplotlib.org/'
nltk/tbl/demo.py:    Note: requires matplotlib
nltk/tbl/demo.py:    import matplotlib.pyplot as plt
nltk/tokenize/texttiling.py:    from matplotlib import pylab
```

tomaarsen:
8c759729d3a1d03fd724c96c6d82c38382ef2e82 should have solved the last of these. (At least, for commonly accessed plots)

PR #2240, opened 2019-02-23
Make plot methods return Axes objects

nick-ulle:
This pull request addresses issue #2239.

So far I've only modified the code for `FreqDist.plot`. If the maintainers agree with the proposal in #2239, I will also modify the code for `ConditionalFreqDist.plot`.

Feedback is appreciated.


alvations:
Thanks @nick-ulle for the proposal! The change looks good, please proceed with other plots that you think should use the same `ax.set_title()` idiom.

stevenbird:
Hi @nick-ulle -- please go ahead with more of this, if you can, thanks.

alvations:
There's still a few points where `pylab` is used, it can be improved but this PR looks good to merge since we have cleaned up the `probability.py` and the test passes

----

For reference:

```
$ grep matplotlib nltk/**/*.py
nltk/app/__init__.py:        from matplotlib import pylab
nltk/app/__init__.py:            "nltk.app.wordfreq not loaded " "(requires the matplotlib library)."
nltk/app/wordfreq_app.py:from matplotlib import pylab
nltk/draw/dispersion.py:        from matplotlib import pylab
nltk/draw/dispersion.py:            'The plot function requires matplotlib to be installed.'
nltk/draw/dispersion.py:            'See http://matplotlib.org/'
nltk/parse/dependencygraph.py:        from matplotlib import pylab
nltk/sentiment/util.py:        import matplotlib.pyplot as plt
nltk/sentiment/util.py:            'The plot function requires matplotlib to be installed.'
nltk/sentiment/util.py:            'See http://matplotlib.org/'
nltk/tbl/demo.py:    Note: requires matplotlib
nltk/tbl/demo.py:    import matplotlib.pyplot as plt
nltk/tokenize/texttiling.py:    from matplotlib import pylab
```

alvations (approved):
LGTM now with the improvements to `nltk.probability`

stevenbird:
Thanks @nick-ulle... more welcome!

PR #2246, opened 2019-03-06
Fix ignoring non-Cyrillic in RussianStemmer

KMiNT21:
This fixes bug with processing English words in Russian (Cyrillic) text.

For example, if we use word "life" in text, the result now will be "lif" (wrong).

So the best is just to ignore all non-Cyrillic word (i.e. it can be brand names).


stevenbird:
Thanks @KMiNT21

PR #2253, opened 2019-03-18
fix infinite-loop of extract when fe not in f_aligned or fe is over t…

kobikun:
I found the infinite-loop when I extract phrase using alignments.
Main reason is caused by `fe`. 
when fe is equal `trglen`,  fe(`fe +=1`) is not in f_aligned and fe is over trglen. then the loop goes into infite-loop.


stevenbird:
Thanks @kobikun!

Issue #2076, opened 2018-08-05
TreebankWordTokenizer span_tokenize still throwing exception for some corner-cases


gkucsko:
very similar to #1865, there are still unhandled inputs to TreebankWordTokenizer that will cause span_tokenize to throw:
```
>>> from nltk.tokenize import TreebankWordTokenizer
>>> s = u"\'\'good muffins\""
>>> list(TreebankWordTokenizer().span_tokenize(s))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "nltk/tokenize/treebank.py", line 198, in span_tokenize
    for tok in align_tokens(tokens, text):
  File "nltk/tokenize/util.py", line 258, in align_tokens
    raise ValueError('substring "{}" not found in "{}"'.format(token, sentence))
ValueError: substring "''" not found in "''good muffins""
```
a working fix could be:
```
line 190 in tokenize.treebank.py:
matched = [m.group() for m in re.finditer(r"\`\`|(?<!^)\'\'|^\'\'(?=\s)|\"", text)]
```
i don't like the extra look arounds, but at least this wouldn't change anything about the current implementation. happy to submit a merge req if it's helpful!

gkucsko:
@alvations (in regards to https://github.com/nltk/nltk/pull/1864#issuecomment-411657125)
thanks, agreed. due to the substitutions this will probably never work reliably. close this as won't fix?

alvations:
If the current `span_tokenize()` is really buggy, we might want to deprecate it back to the `NotImplementedError` until a better solution. Meanwhile, we can leave this open because someone might be able to find some way to fix this smartly =)

PR #2254, opened 2019-03-19
Make word_tokenize treat two single quotes ('') consistently.

piero:
Example:

    "Here is a ''double quoted'' text"

Tokenization before this commit:

    ['Here', 'is', 'a', '``', 'double', 'quoted', "''", 'text']

Tokenization after this commit:

    ['Here', 'is', 'a', '``', 'double', 'quoted', '``', 'text']

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -40,6 +40,7 @@
 - Claude Coulombe
 - Lucas Cooper
 - Robin Cooper
+- Piero Cornice
 - Chris Crowner
 - James Curran
 - Arthur Darcet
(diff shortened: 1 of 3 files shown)


alvations:
Thank you for the suggestion. But this should **not** be merged. 

The second set of single quotes should be kept as two single quotes. This '' is to encode end of quotation and `` is to encode start of quotation.

piero:
Thank you for your observation.
Another reason for this change was to address problems with the span tokenizer down the line (see https://github.com/nltk/nltk/issues/1750 and https://github.com/nltk/nltk/issues/2076) but I see the reason why you may not want to merge it.

stevenbird:
Thanks @alvations and @piero

Issue #2221, opened 2019-01-29
Minimum Edit Distance Word Alignment


okz12:
I've been using NLTK for work and one of the features that would be quite useful is aligning words / characters between two strings by using a minimum edit distance. The alignment tools I came across in NLTK were looking at translations between two different languages. In my case, I was looking at the quality of OCR'd documents and needed to compare the output with a reference and highlight the changes.

If accepted, I would like to work on this issue. It would involve using wagner-fischer algorithm to create the optimal alignment followed by a backtrace to retrieve it.

PR #2257, opened 2019-03-25
Added minimum edit/Levenshtein distance based alignment function with doctest

okz12:
Created this for feature/issue #2221, a minimum edit-distance based alignment of two strings. The implementation reuses edit_distance functions in nltk.metrics but adds backtrace to be able to recover the original strings with minimum distance mappings.

E.g., 'f', 'bar' will align in 'fubar' and 'foobar'. 

`print(edit_distance_align('fubar', 'foobar'))`
> [(0, 0), (1, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)]
NB: 0 is a null/stop index

The mapping from first string to second printed out:
```
s1 = '_fubar'
s2 = '_foobar'
print([(s1[x[0]], s2[x[1]]) for x in edit_distance_align('fubar','foobar')])
```
> [('_', '_'), ('f', 'f'), ('f', 'o'), ('u', 'o'), ('b', 'b'), ('a', 'a'), ('r', 'r')]

### DIFF
--- a/nltk/metrics/distance.py
+++ b/nltk/metrics/distance.py
@@ -103,6 +103,76 @@ def edit_distance(s1, s2, substitution_cost=1, transpositions=False):
     return lev[len1][len2]
 
 
+def _edit_dist_backtrace(lev_map):
+    i, j = len(lev_map) - 1, len(lev_map[1]) - 1
+    alignment = [(i, j)]
+
+    while (i, j) != (0, 0):
+        # find direction of minimum cost
+        s1_skip_cost = s2_skip_cost = sub_cost = float('inf')
+        if i - 1 >= 0:
+            s1_skip_cost = lev_map[i - 1][j]
+        if j - 1 >= 0:
+            s2_skip_cost = lev_map[i][j - 1]
+        if i - 1 >= 0 and j - 1 >= 0:
+            sub_cost = lev_map[i - 1][j - 1]
+
+        # move in direction of minimum cost
+        if sub_cost <= s1_skip_cost and sub_cost <= s2_skip_cost:
+            i, j = i - 1, j - 1
+        elif s2_skip_cost <= s1_skip_cost:
+            i, j = i, j - 1
+        else:
+            i, j = i - 1, j
+
+        alignment.append((i, j))
+    alignment.reverse()
+    return alignment
+
+
+def edit_distance_align(s1, s2, substitution_cost=1):
+    """
+    Calculate the minimum Levenshtein edit-distance based alignment
+    mapping between two strings. The alignment finds the mapping
+    from string s1 to s2 that minimizes the edit distance cost.
+    For example, mapping "rain" to "shine" would involve 2
+    substitutions, 2 matches and an insertion resulting in
+    the following mapping:
+    [(0, 0), (1, 1), (2, 2), (3, 3), (4, 4), (4, 5)]
+    NB: (0, 0) is the start state without any letters associated
+    See more: https://web.stanford.edu/class/cs124/lec/med.pdf
+
+    This function does not support transposition.
+
+    :param s1, s2: The strings to be aligned
+    :type s1: str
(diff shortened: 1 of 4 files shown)


nltk/metrics/distance.py
```diff
@@ -103,6 +103,76 @@ def edit_distance(s1, s2, substitution_cost=1, transpositions=False):
     return lev[len1][len2]
 
 
+def _edit_dist_backtrace(lev_map):
+    i, j = len(lev_map) - 1, len(lev_map[1]) - 1
+    alignment = [(i, j)]
+
+    while (i, j) != (0, 0):
+        # find direction of minimum cost
+        s1_skip_cost = s2_skip_cost = sub_cost = float('inf')
```
iliakur: May I suggest a more declarative take on the inside of the while-loop? It takes advantage of Python's built-in `min` function and the `operator` module.

```python
directions = [
    (i - 1, j),      # skip s1
    (i, j - 1),      # skip s2
    (i - 1, j - 1),  # substitution
]
direction_costs = (
   (lev_map[i][j], (i, j)) for i, j in directions
)
_, (i, j) = min(direction_costs, key=operator.itemgetter(0))
```

Feel free to ask if anything in the snippet above is unclear!

nltk/metrics/distance.py
```diff
@@ -103,6 +103,76 @@ def edit_distance(s1, s2, substitution_cost=1, transpositions=False):
...
+        # move in direction of minimum cost
+        if sub_cost <= s1_skip_cost and sub_cost <= s2_skip_cost:
+            i, j = i - 1, j - 1
+        elif s2_skip_cost <= s1_skip_cost:
+            i, j = i, j - 1
+        else:
+            i, j = i - 1, j
+
+        alignment.append((i, j))
+    alignment.reverse()
```
iliakur: You can combine this line with the next one by doing this:
```python
return reversed(alignment)
```

nltk/metrics/distance.py
```diff
@@ -103,6 +103,76 @@ def edit_distance(s1, s2, substitution_cost=1, transpositions=False):
...
+    :param s1, s2: The strings to be aligned
+    :type s1: str
+    :type s2: str
+    :type substitution_cost: int
+    :rtype List[Tuple(int, int)]
+    """
+    # set up a 2-D array
+    len1 = len(s1)
+    len2 = len(s2)
+    lev = _edit_dist_init(len1 + 1, len2 + 1)
```
iliakur, line 156: To me the role of this variable would be clearer if it were called `lev_map`, just like you call it in the backtrace function.

iliakur:
@okz12 thanks, this looks good! I left a few minor comments. It would be cool if you could add some more *strange* cases for the tests. For instance, what happens if one of the strings is empty? It may also be worthwhile to mention in the docstring for the function what happens if multiple alignments are possible for the same minimal edit distance.

nltk/metrics/distance.py
okz12: I've made this change.

nltk/metrics/distance.py
okz12: Made the change and added list() around it to return a list rather than an iterator.

nltk/metrics/distance.py
okz12: Good spot. I changed lev_map to lev in the backtrace function because the other edit distance functions (init / step etc.) all just use lev as the variable name. Backtrace was the odd one out.

okz12:
Hey, thanks for the review, that was very helpful. I've incorporated the 3 code comments, added some more test cases and tried to explain the order of backtrace precedence in docstrings for multiple alignments (still not sure if it's clear enough, might need to think of a good example to bring it to life).

stevenbird:
Thanks @okz12 and @Copper-Head, this looks good now.

PR #2268, opened 2019-04-17
Fix random seeding in LanguageModel.generate

cifkao:
In the current implementation of `LanguageModel.generate`, the random number generator is re-seeded (in fact re-instantiated) before every word:
https://github.com/nltk/nltk/blob/66f56ead003e64c5bd215427c8d09b4a5fd7fd78/nltk/lm/api.py#L89-L92
This is especially unfortunate when `random_seed` is not `None`, because it makes the model always predict the same token for a given context.

This PR fixes this issue by seeding the generator only once.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -246,6 +246,7 @@
 - Uday Krishna <https://github.com/udaykrishna>
 - Osman Zubair <https://github.com/okz12>
 - Viresh Gupta <https://github.com/virresh>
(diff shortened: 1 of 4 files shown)


iliakur:
@cifkao thanks for finding and fixing this. A neat little bug :)

:+1: from me!

stevenbird:
Thanks @cifkao

Issue #2063, opened 2018-07-12
ConditionalFreqDist keys are mutated on call to plot


rmNULL:
### Brief
`ConditionalFreqDist` modifies the conditions(keys) when a non-existing key 
is passed as a `conditions` keyword argument to `tabulate()`/`plot()`.

Look at an
### Example  
![nltk-condist](https://user-images.githubusercontent.com/8173131/42654522-7dd6451a-8636-11e8-993a-fd10889dc211.png)

This undocumented behaviour appears a bug to me.

---

### Why is this happening to me ? 
**Note**,  please do correct(my blabbering).
The culprits behind this behaviour hide in [plot](https://github.com/nltk/nltk/blob/ecced11335a8942beeab4ed9020eb93f18457d10/nltk/probability.py#L1859) and [tabulate](https://github.com/nltk/nltk/blob/ecced11335a8942beeab4ed9020eb93f18457d10/nltk/probability.py#L1903) (forgive for stating the obvious). 

As `ConditionalFreqDist` is inherited from `defaultdict`, when we do a `self[c]`(links above), 
new key is created as a side-effect. This can be seen in an example below.

![default-dict](https://user-images.githubusercontent.com/8173131/42655543-1b815644-863a-11e8-89c2-01ee3e29eb8f.png)

__Thank You__

irisxzhou:
Hello, I'm a new contributor to open source projects and I would like to work on this bug. :)

rmNULL:
This was fixed and merged, closing.   Thanks @irisxzhou

PR #2285, opened 2019-05-05
Fix ConditionalFreqDist mutation on tabulate/plot call

irisxzhou:
This refers to issue #2063.
 
Since self[c] adds c to the ConditionalFreqDist regardless of
whether c is a valid condition or not, calling tabulate or plot
with condtiions not in the cfd puts them in conditions, causing
errors when it's called again. Checking to see whether c is in
self before calling self[c] prevents this problem.


nltk/probability.py
```diff
@@ -1790,7 +1790,7 @@ def entropy(pdist):
 
 ##//////////////////////////////////////////////////////
 ##  Conditional Distributions
-##//////////////////////////////////////////////////////
+##//////////////////////////////////////////////////////https://github.com/nltk/nltk/blob/develop/nltk/probability.py
```
alvations: Please remove this typo =)

alvations:
@irisxzhou Thanks for the contribution! It looks right to check `if c in self` before doing `self[c]`.

Is there some way to test an example of how the KeyError will happen?

nltk/probability.py
```diff
@@ -1954,8 +1956,9 @@ def tabulate(self, *args, **kwargs):
         cumulative = _get_kwarg(kwargs, 'cumulative', False)
         conditions = _get_kwarg(kwargs, 'conditions', sorted(self.conditions()))
         samples = _get_kwarg(
-            kwargs, 'samples', sorted(set(v for c in conditions for v in self[c]))
-        )  # this computation could be wasted
+            kwargs, 'samples', sorted(set(v for c in conditions
+                                          if c in self
+                                          for v in self[c])))  # this computation could be wasted
```
alvations, line 1961: Looks right but we've to test cases where this will happen.

alvations (changes requested):
Please remove the typo and if there's a test written for this, it'll be easier to maintain the patch/fix.

nltk/probability.py
irisxzhou: Oops! No idea how that typo got in there. I will fix that and write a short test in the next day or two :) Should it go in its own file or should I try to fit into an existing test?

irisxzhou:
It seems like on travis one of my tests is failing (the one for `plot`); however, it passes on my machine so I'm not sure what to do here.

alvations:
I'm getting some errors just from cloning your fork and running the unittest:

```
git clone https://github.com/irisxzhou/nltk.git
cd nltk/
git checkout cfd-mutation-bug
python3 -m unittest nltk/test/unit/test_cfd_mutation.py 
```

[out]:

```
F.
FAIL: test_plot (nltk.test.unit.test_cfd_mutation.TestEmptyCondFreq)
  File "/Users/liling.tan/fixnltk/fix-cfd/nltk/nltk/test/unit/test_cfd_mutation.py", line 22, in test_plot
    self.assertEqual(empty.conditions(),[])
AssertionError: Lists differ: ['B', 'U', 'G'] != []

First list contains 3 additional elements.
First extra element 0:
'B'

- ['B', 'U', 'G']
+ []

Ran 2 tests in 5.032s

FAILED (failures=1)
```

Let me take a look.

nltk/probability.py
```diff
@@ -1909,7 +1909,9 @@ def plot(self, *args, **kwargs):
         conditions = _get_kwarg(kwargs, 'conditions', sorted(self.conditions()))
         title = _get_kwarg(kwargs, 'title', '')
         samples = _get_kwarg(
-            kwargs, 'samples', sorted(set(v for c in conditions for v in self[c]))
+            kwargs, 'samples', sorted(set(v for c in conditions
+                                          #if v in self
```
alvations: Ah, this shouldn't be commented.

alvations (changes requested):

alvations:
Ah found the bug. You've commented out the `if v in self` for the `plot()` function thus the bug in your unittest. I've pushed a patch the test should work out right now =)

nltk/probability.py
alvations: Patched.

alvations (approved):

stevenbird:
Can you confirm that `cfd[a][b] += 1` will still work when `a` is previously unseen?

irisxzhou:
@alvations Thanks! I'm debugging for a college course and was doing a demo; I must have forgotten to uncomment that line before submitting my PR. 

> Can you confirm that `cfd[a][b] += 1` will still work when `a` is previously unseen?

@stevenbird I just tried it and it does work! The changes I made in my pull request only change `tabulate` and `plot`, which shouldn't be allowed to have side effects because they're purely for display. I don't think any other functionality is affected.

irisxzhou:
Also, do I add myself to the contributors document? It's my first time contributing to open source so I'm not sure what the procedure is for this sort of thing :)

alvations:
@irisxzhou Yes if you would like to, please do add yourself to the contributor list on https://github.com/nltk/nltk/blob/develop/AUTHORS.md.

I think as long as we don't use, `cdf[a][b].default_factory = None`, it should work =)

Also, could you add the unittest for `cfd[a][b] += 1` =) Thanks in advance!

stevenbird:
Thanks @irisxzhou and @alvations

Issue #1324, opened 2016-03-09
ParentedTree breaks deepcopy


josepvalls:
Can deepcopy Tree but not ParentedTree, not sure if may be related to this: https://github.com/nltk/nltk/issues/130

> > > from nltk.tree import Tree,ParentedTree
> > > t1 = Tree.fromstring("(TOP (S (NP (NNP Bell,)) (NP (NP (DT a) (NN company)) (SBAR (WHNP (WDT which)) (S (VP (VBZ is) (VP (VBN based) (PP (IN in) (NP (NNP LA,)))))))) (VP (VBZ makes) (CC and) (VBZ distributes) (NP (NN computer))) (. products.)))")
> > > t2 = copy.deepcopy(t1)
> > > t3 = ParentedTree.fromstring("(TOP (S (NP (NNP Bell,)) (NP (NP (DT a) (NN company)) (SBAR (WHNP (WDT which)) (S (VP (VBZ is) (VP (VBN based) (PP (IN in) (NP (NNP LA,)))))))) (VP (VBZ makes) (CC and) (VBZ distributes) (NP (NN computer))) (. products.)))")
> > > import copy
> > > t4 = copy.deepcopy(t3)
> > > Traceback (most recent call last):
> > >   File "<input>", line 1, in <module>
> > >   File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 190, in deepcopy
> > >     y = _reconstruct(x, rv, 1, memo)
> > >   File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 351, in _reconstruct
> > >     item = deepcopy(item, memo)
> > >   File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 190, in deepcopy
> > >     y = _reconstruct(x, rv, 1, memo)
> > >   File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 351, in _reconstruct
> > >     item = deepcopy(item, memo)
> > >   File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 190, in deepcopy
> > >     y = _reconstruct(x, rv, 1, memo)
[...]

PR #2286, opened 2019-05-05
Patch Tree classes to work with native Python copy library

irisxzhou:
This refers to issue #1324.

Link \_\_copy\_\_ and \_\_deepcopy\_\_ to copy function defined in Tree
classes. This allows the native Python copy library to work with
Tree, ImmutableTree, and ParentedTree.

### DIFF
--- a/nltk/tree.py
+++ b/nltk/tree.py
@@ -556,11 +556,18 @@ class Tree(list):
         else:
             return tree
 
+    def __copy__(self):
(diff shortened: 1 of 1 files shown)


alvations (approved):
Thanks @irisxzhou! LGTM =)

stevenbird:
Thanks @irisxzhou

PR #2287, opened 2019-05-06
don't require lists as arguments to PositiveNaiveBayesClassifier.train

drewmiller:
Requiring a list as an input may produce significant memory overhead. The only obvious list function used is the call to len() but since we've just iterated over the iterable, counting up the values ahead of time is trivial.


alvations (approved):
Thanks @drewmiller! Yes counting and keeping the counts of the no. of labels/un-labels is more optimal.

stevenbird:
Thanks @drewmiller

PR #2289, opened 2019-05-07
Retrying Cythonize

alvations:


alvations:
@scoder Looks like the Cython build works!

Any idea how to test the Cython functionalities vs the native CPython ones?

alvations:
Continuing the conversation from #2093, @jnothman @scoder 

Any idea how to make `pip install nltk[binary]` possible?

scoder:
I'd definitely leave out the tests. :)
Also, compiling the `__init__.py` files might not be very useful, and can even be problematic in old Python versions (pre-3.5, I think).

Apart from that, I think it's a good approach to try to compile as many modules as possible, and then comment them back out (also for documentation purposes) if it doesn't work and/or does not show a sufficiently large performance improvement. That makes it clear which modules were explicitly chosen.

Some of the `undeclared name not builtin` errors may hint at actual bugs in the nltk modules, undefined global names. Probably worth fixing those. Otherwise, setting `Cython.Compiler.Options.error_on_unknown_names = False` would provide more Python-like behaviour with a `NameError` at runtime.

alvations:
[CI: retest]

alvations:
@stevenbird @scoder @jnothman Without the baggage of Python 2 code in the develop branch, it worked!!

Let me double check whether it's doing what we're expecting =)

scoder:
Are you sure that it worked? I don't see any compilation in the travis logs.

alvations:
@scoder yes, it looked like there's no compilation =( I'll take a closer look.

alvations:
Seemed like this https://github.com/alvations/nltk/blob/develop/tox.ini#L114 didn't set properly and so https://github.com/alvations/nltk/blob/develop/setup.py#L92 didn't kick in. 

I've put quotation marks around the variable names in the `tox.ini`, if that doesn't work too, I think we can force an `export CYTHONIZE_NLTK="true"` in the environment just to make sure the cythonization works.

alvations:
Good thing is the Cython is working =) https://travis-ci.org/nltk/nltk/jobs/578147778 

But the new Travis (from v2.0) seems to have more securities put in place for environmental variables so the `CYTHONIZE_NLTK` variable isn't passing through https://travis-ci.org/nltk/nltk/jobs/578147778

alvations:
@stevenbird I've cleaned up the git commits. 

@scoder @jnothman @hoefling @stevenbird  This time round, the build really compiles the Cython-able .py files and passes the CI tests. 

Hope I'm not missing anything for this PR =)

stevenbird:
Well done!!!

hoefling:
@alvations IMO in the current state the `py-cythonized` env doesn't get the cythonized installation of `nltk`. I have added a comment in the [question on SO](https://stackoverflow.com/q/57705814/2650249), but here is my reasoning again, a bit more verbose. What you have now is:

1. set `CYTHONIZE_NLTK` env var globally;
2. invoke `sdist` - because of 1., this will trigger python sources transpiling to C, but no compilation because we're in the source dist packaging now;
3. create `py-cythonized` tox env;
4. install source dist from 2. to `py-cythonized`. Now you have the correct `CYTHONIZE_NLTK` env var globally from 1. and `setenv` with the wrong (quoted) `CYTHONIZE_NLTK` env and `passenv`. `setenv` wins over `passenv` and sets the quoted `CYTHONIZE_NLTK`. Now the source dist installation doesn't invoke transpiling to C and doesn't invoke compilation of C code to extension modules - you can clearly see it b/c the installation time doesn't change.

The changes proposed are thus:

1. Unquote `CYTHONIZE_NLTK` in `tox.ini`
2. Remove `passenv` section in `[testenv:py-cythonized]`
3. Remove setting `CYTHONIZE_NLTK` env var in `.travis.yml`. It's no good anyway - `tox` is already instructed to update Cython in `py-cythonized` env, no need to do it again in `tools/travis/install.sh`.

Optional change: move cythonizing from the top level of the setup script into a custom `build` or `build_ext` command. Cythonizing sources in `sdist` is useless anyway, but the most problematic part I see is the top-level import of Cython. If I e.g. run 
```sh
$ CYTHONIZE_NLTK=true pip install nltk --no-binary=nltk
```
and Cython is not installed, the installation will raise an `ImportError`. The most appropriate solution to that is to declare a `setup_requires=['Cython']` and invoke `cythonize()` somewhere in `build.finalize_options()` where it is ensured that the build-time deps are already installed.

hoefling:
It's kinda impossible to see whether the cythonized extensions are installed without explicitly pouring the verbose log, but observe the [installation line](https://travis-ci.org/nltk/nltk/jobs/581225054#L2104) in the travis job to the  PR #2407: the installation now lasts a lot longer b/c gcc has a lot to do and `pip`s spinner pollutes the output while the platform-specific wheel is being built. The `py-cythonized` job now takes 9 minutes longer than on the current head of the `cythonize` branch.

Issue #2034, opened 2018-06-06
Better Panlex Swadesh


alvations:
From https://github.com/nltk/nltk_data/issues/117 , we can have a better Panlex swadesh list interface that also reads the language ID files (which might be useful for nlp, e.g. language ID).

Maybe an interface in https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordlist.py#L31 might be:

```python
from __future__ import print_function
from collections import namedtuple, defaultdict
import re
from six import string_types

PanlexLanguage = namedtuple('PanlexLanguage',
                          ['panlex_uid',  # (1) PanLex UID
                           'iso639',      # (2) ISO 639 language code
                           'iso639_type', # (3) ISO 639 language type, see README
                           'script',      # (4) normal scripts of expressions
                           'name',        # (5) PanLex default name
                           'langvar_uid'  # (6) UID of the language variety in which the default name is an expression
                           ])

class PanlexSwadeshCorpusReader(WordListCorpusReader):
    """
    This is a class to read the PanLex Swadesh list from

    David Kamholz, Jonathan Pool, and Susan M. Colowick (2014).
    PanLex: Building a Resource for Panlingual Lexical Translation.
    In LREC. http://www.lrec-conf.org/proceedings/lrec2014/pdf/1029_Paper.pdf

    License: CC0 1.0 Universal
    https://creativecommons.org/publicdomain/zero/1.0/legalcode
    """
    def __init__(self, *args, **kwargs):
        super(PanlexSwadeshCorpusReader, self).__init__(*args, **kwargs)
        # Find the swadesh size using the fileids' path.
        self.swadesh_size = re.match(r'swadesh([0-9].*)\/', self.fileids()[0]).group(1)
        self._languages = {lang.panlex_uid:lang for lang in self.get_languages()}
        self._macro_langauges = self.get_macrolanguages()

    def license(self):
        print('CC0 1.0 Universal')

    def readme(self):
        print(self.raw('README'))

    def language_codes(self):
        return self._languages.keys()

    def get_languages(self):
        for line in self.raw('langs{}.txt'.format(self.swadesh_size)).split('\n'):
            if not line.strip(): # Skip empty lines.
                continue
            yield PanlexLanguage(*line.strip().split('\t'))

    def get_macrolanguages(self):
        macro_langauges = defaultdict(list)
        for lang in self._languages.values():
            macro_langauges[lang.iso639].append(lang.panlex_uid)
        return macro_langauges

[...]

PR #2291, opened 2019-05-07
Added an improved panlex swadesh reader

alvations:
c.f. #2034


alvations:
@kamholz your suggestions to improve the interface for the Panlex Swadesh reader would be greatly appreciated =) 

Thank you in advance! And thank you for the sharing the panlex resource!

stevenbird:
Thanks @alvations

Issue #2275, opened 2019-04-22
OMW lemmas sometimes has additional underscore


alvations:
OMW lemmas sometimes has additional underscore, e.g. `preobrat_v_mišljenju_` in:

```python
>>> from nltk.corpus import wordnet as wn
>>> wn.synset('about-face.n.02')
Synset('about-face.n.02')
>>> wn.synset('about-face.n.02').lemma_names(lang='slv')
['popolna_sprememba_v_mišljenju', 'popoln_obrat', 'preobrat', 'preobrat_v_mišljenju_']
```

From the raw file there isn't any tailing underscore:

```
~/nltk_data/corpora/omw/slv$ grep -n "preobrat v mišljenju" wn-data-slv.tab  
2383:00163406-n	slv:lemma	preobrat v mišljenju 
```

alvations:
Patched in #2294 =)

Issue #2274, opened 2019-04-22
Wordnet's all_lemma_names returns inconsistent types


alvations:
Depending on the language, Wordnet's `all_lemma_names()` returns inconsistent types

```python
>>> from nltk.corpus import wordnet as wn
>>> type(wn.all_lemma_names(lang='eng'))
<class 'dict_keyiterator'>
>>> type(wn.all_lemma_names(lang='cat'))
<class 'list'>
```

alvations:
Patched in #2294 =)

Issue #2265, opened 2019-04-16
Wordnet gloss parsing issues


alvations:
From https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/wordnet.py#L1396, it looks like the line parsing wasn't done correctly for the example vs the definition of the Synset gloss:

```python
	     # parse out the definitions and examples from the gloss
            columns_str, gloss = data_file_line.split('|')
            gloss = gloss.strip()
            definitions = []
            for gloss_part in gloss.split(';'):
                gloss_part = gloss_part.strip()
                if gloss_part.startswith('"'):
                    synset._examples.append(gloss_part.strip('"'))
                else:
                    definitions.append(gloss_part)
            synset._definition = '; '.join(definitions)
```

As a sanity check, from

```python
from nltk.corpus import wordnet

for ss in wordnet.all_synsets():
    if '"' in ss.definition():
        print(ss, ss.definition())
```

there are these problematic ones:

```
Synset('long.s.09') having or being more than normal or necessary:"long on brains"
Synset('unobjectionable.s.03') not objectionable; it's the means that one can't accept"
Synset('faulty.s.02') characterized by errors; not agreeing with a model or not following established rules; the wrong side of the road"
Synset('dead.s.17') devoid of activity; nothing ever happens here"
Synset('active.a.10') expressing that the subject of the sentence has the semantic function of actor: "Hemingway favors active constructions"
Synset('intercalary.s.01') having a day or month inserted to make the calendar year correspond to the solar year: "Feb. 29 is an intercalary day"
Synset('aggressive.a.01') having or showing determination and energetic pursuit of your ends; positive in his convictions"
Synset('slain.s.01') killed; `slain' is formal or literary as in "slain warriors"
Synset('out_of_play.s.01') (of a ball) "a ball that is out of play is dead"
Synset('fast_asleep.s.01') sleeping deeply; he's sound asleep"
Synset('bashful.s.01') self-consciously timid; lowering my head, I looked at the wall"- Ezra Pound
Synset('gutsy.a.01') marked by courage and determination in the face of difficulties or danger; robust and uninhibited; it was a gutsy thing to do"
Synset('cautious.a.01') showing careful forethought; never making swift decisions"
Synset('certificated.s.01') furnished with or authorized by a certificate: "certificated teachers"
Synset('ineluctable.s.01') impossible to avoid or evade:"inescapable conclusion"
[...]

PR #2294, opened 2019-05-07
Patching wordnet issues

alvations:
Fixing #2265 , #2274 and #2275 

 - #2275 OMW lemmas would not have trailing undersscores.
 - #2274 The return types for english and other OMW lemmas are both now of iterable types.
 - #2265 The definitions, examples will be extracted correctly now


stevenbird:
Thanks @alvations

PR #2297, opened 2019-05-12
Fix broken links in documentation

simonepri:
I believe that there are other files that have the same problem.
This is just a starting point.


stevenbird:
Thanks @simonepri. More updates welcomed.

simonepri:
@stevenbird
I think it would have been better to wait and fix all the links with this single PR.
I'll try to find some time to check the other links as well and open another PR in the case.

Issue #2251, opened 2019-03-17
text.generate


stevenbird:
Look into reinstating ```text.generate()``` now that the language model package ```lm``` is available.

cf https://github.com/nltk/nltk/commit/73f7e7b9ec301c39dc11035f138a27feb3c436cf#diff-f5cb9102f911e93ca5fefa00b56a1e3f

stevenbird:
@Copper-Head: do you have any advice here, or perhaps time to look into this please?

iliakur:
The impression I got from skimming the old (deleted) method is that it was a wrapper around `model.generate`. Is that correct?

I'm not sure what all the trigram model checking was about, but for a starter we could maybe have `text.generate` have the same interface as `model.generate`. Does that sound reasonable? If yes, I could find time to add it.

stevenbird:
Thanks @Copper-Head, that would be great. Yes, same interface, except ``num_words=100`` to preserve the behaviour published in the book: https://www.nltk.org/book/ch01.html#searching-text

iliakur:
This can be closed.

stevenbird:
Thanks Ilya

PR #2300, opened 2019-05-15
Add generate() for Text

alvations:
C.f. #736 and #2251

The lidstone LM is a much slower than MLE so the `generate()` didn't try to replicate Lidstone with 0.2 gamma but uses the MLE model, https://github.com/nltk/nltk/commit/73f7e7b9ec301c39dc11035f138a27feb3c436cf#diff-f5cb9102f911e93ca5fefa00b56a1e3f

### DIFF
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -534,50 +534,52 @@ class Text(object):
         from nltk.draw import dispersion_plot
 
         dispersion_plot(self, words)
-        
-    def build_ngram_lm(self, tokenized_sents, n=3):
+
+    def _train_default_ngram_lm(self, tokenized_sents, n=3):
         train_data, padded_sents = padded_everygram_pipeline(n, tokenized_sents)
         model = MLE(order=n)
         model.fit(train_data, padded_sents)
         return model
-        
+
     def generate(self, length=100, text_seed=None, random_seed=42):
         """
         Print random text, generated using a trigram language model.
         See also `help(nltk.lm)`.
-        
+
         :param length: The length of text to generate (default=100)
         :type length: int
-        
+
         :param text_seed: Generation can be conditioned on preceding context.
         :type text_seed: list(str)
-        
+
         :param random_seed: A random seed or an instance of `random.Random`. If provided,
         makes the random sampling part of generation reproducible.
         :type random_seed: int
-        
+
         """
         # Create the model when using it the first time.
-        self.tokenized_sents = [sent.split(' ') for sent in 
+        self._tokenized_sents = [sent.split(' ') for sent in
                                 sent_tokenize(' '.join(self.tokens))]
         if not hasattr(self, 'trigram_model'):
             print("Building ngram index...", file=sys.stderr)
-            self.trigram_model = self.build_ngram_lm(self.tokenized_sents, n=3)
-            
+            self._trigram_model = self._train_default_ngram_lm(self._tokenized_sents, n=3)
+
         generated_tokens = []
-        
+
         assert length > 0, 'The `length` must be more than 0.'
         while len(generated_tokens) < length:
-            for idx, token in enumerate(self.trigram_model.generate(length, text_seed=text_seed, random_seed=random_seed)):
+            for idx, token in enumerate(self._trigram_model.generate(length, text_seed=text_seed, random_seed=random_seed)):
                 if token == '<s>':
                     continue
                 if token == '</s>':
                     break
                 generated_tokens.append(token)
             random_seed += 1
-        
+
         prefix = ' '.join(text_seed) + ' ' if text_seed else ''
-        print(prefix + tokenwrap(generated_tokens[:length]))
+        output_str = prefix + tokenwrap(generated_tokens[:length])
+        print(output_str)
+        return output_str
 
     def plot(self, *args):
         """
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -544,11 +544,7 @@ class Text(object):
     def generate(self, length=100, text_seed=None, random_seed=42):
         """
         Print random text, generated using a trigram language model.
-
-        See also: 
-        
-            >>> from nltk import lm
-            >>> help(lm)
+        See also `help(nltk.lm)`.
         
         :param length: The length of text to generate (default=100)
         :type length: int
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -553,7 +553,7 @@ class Text(object):
         :type text_seed: list(str)
 
         :param random_seed: A random seed or an instance of `random.Random`. If provided,
-        makes the random sampling part of generation reproducible.
+        makes the random sampling part of generation reproducible. (default=42)
         :type random_seed: int
 
         """
--- a/nltk/text.py
+++ b/nltk/text.py
@@ -34,8 +34,8 @@ from nltk.compat import python_2_unicode_compatible
 from nltk.tokenize import sent_tokenize
 
 ConcordanceLine = namedtuple(
-    'ConcordanceLine',
-    ['left', 'query', 'right', 'offset', 'left_print', 'right_print', 'line'],
+    "ConcordanceLine",
+    ["left", "query", "right", "offset", "left_print", "right_print", "line"],
 )
 
 
@@ -50,8 +50,8 @@ class ContextIndex(object):
     @staticmethod
     def _default_context(tokens, i):
         """One left token and one right token, normalized to lowercase"""
-        left = tokens[i - 1].lower() if i != 0 else '*START*'
-        right = tokens[i + 1].lower() if i != len(tokens) - 1 else '*END*'
+        left = tokens[i - 1].lower() if i != 0 else "*START*"
+        right = tokens[i + 1].lower() if i != len(tokens) - 1 else "*END*"
         return (left, right)
 
     def __init__(self, tokens, context_func=None, filter=None, key=lambda x: x):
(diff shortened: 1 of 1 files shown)


nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
-        """
-        Issues a reminder to users following the book online
-        """
-        import warnings
-
-        warnings.warn(
-            'The generate() method is no longer available.', DeprecationWarning
-        )
+        
+    def build_ngram_lm(self, tokenized_sents, n=3):
```
iliakur: Do you expect this to be part of the public API of `Text`? If not, I would say it makes sense to make it a private helper function in this module called `_train_default_ngram_lm`.

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+        See also `help(nltk.lm)`.
+        
+        :param length: The length of text to generate (default=100)
+        :type length: int
+        
+        :param text_seed: Generation can be conditioned on preceding context.
+        :type text_seed: list(str)
+        
+        :param random_seed: A random seed or an instance of `random.Random`. If provided,
+        makes the random sampling part of generation reproducible.
```
iliakur: Do you think it makes sense to set a default number here? In your experience, do the folks who use it expect to be able to reproduce the texts?

If yes, we should document the default value in the docstring, like we did for the length argument.

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+                                sent_tokenize(' '.join(self.tokens))]
+        if not hasattr(self, 'trigram_model'):
+            print("Building ngram index...", file=sys.stderr)
+            self.trigram_model = self.build_ngram_lm(self.tokenized_sents, n=3)
+            
+        generated_tokens = []
+        
+        assert length > 0, 'The `length` must be more than 0.'
+        while len(generated_tokens) < length:
+            for idx, token in enumerate(self.trigram_model.generate(length, text_seed=text_seed, random_seed=random_seed)):
```
iliakur: Is this PEP8-compliant in terms of length? I'm assuming you ran `black` on this, right?

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+        :param text_seed: Generation can be conditioned on preceding context.
+        :type text_seed: list(str)
+        
+        :param random_seed: A random seed or an instance of `random.Random`. If provided,
+        makes the random sampling part of generation reproducible.
+        :type random_seed: int
+        
+        """
+        # Create the model when using it the first time.
+        self.tokenized_sents = [sent.split(' ') for sent in
```
iliakur: Should we consider making this a private attribute? Seems a bit scary to me that by calling `generate` we change the public state of `Text` by either adding a new attribute or re-setting an existing one. Also note, we perform this operation even if the ngram model is already defined (i.e. we don't actually *need* the tokenized sentences then).

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+        
+        :param random_seed: A random seed or an instance of `random.Random`. If provided,
+        makes the random sampling part of generation reproducible.
+        :type random_seed: int
+        
+        """
+        # Create the model when using it the first time.
+        self.tokenized_sents = [sent.split(' ') for sent in 
+                                sent_tokenize(' '.join(self.tokens))]
+        if not hasattr(self, 'trigram_model'):
```
iliakur: Similar concern here: do people expect that an ngram model gets trained in the background if they want to generate from a text? This is where I'm not familiar enough with how this class gets used to judge what folks will find surprising or not.

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+            print("Building ngram index...", file=sys.stderr)
```
iliakur, line 566: Should we make this a logging call and let the caller decide where and how it gets reported? Or are folks coming into this with expectations to see printed output?

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+        
+        assert length > 0, 'The `length` must be more than 0.'
+        while len(generated_tokens) < length:
+            for idx, token in enumerate(self.trigram_model.generate(length, text_seed=text_seed, random_seed=random_seed)):
+                if token == '<s>':
+                    continue
+                if token == '</s>':
+                    break
+                generated_tokens.append(token)
+            random_seed += 1
```
iliakur, line 585: Tell me if I understood the lines from 570 down to here correctly:

Until we've generated enough tokens, keep taking tokens from the ngram model, **one sentence at a time**. Increment the seed between calls to the ngram model so as not to generate exactly the same stuff. Since we're taking one sentence at a time, the final generated text may be longer than what the caller requested. We truncate it to the required length before returning.

nltk/text.py
```diff
@@ -530,16 +534,50 @@ def dispersion_plot(self, words):
...
+        
+        prefix = ' '.join(text_seed) + ' ' if text_seed else ''
+        print(prefix + tokenwrap(generated_tokens[:length]))
```
iliakur: I'm surprised to see `print` and not `return` here. It really seems like this is intended to be kind of a "display" or "demo" method and its output not passed on to other functions. Right?

iliakur (changes requested):
Wow, that was quick! I have a lot of questions :)

nltk/text.py
alvations: 👍

nltk/text.py
alvations: I think reproduction is important for learning. But I can't seem to reproduce the original usage of `Text.generate()` from last example of Section 1.3 from https://www.nltk.org/book/ch01.html

nltk/text.py
alvations: Nope, I haven't black on this, I don't have much preference for `black` or original PEP-8, but `black` seems to be the modern standard though. Will do that.

nltk/text.py
alvations: The weird thing is Text don't keep sentence tokenization. And sentence tokenization splits on `'\n` which gutenberg (I think it's extracted from pdf) actually has sentences spanning across `\n`.

nltk/text.py
alvations: It's because most people using the `generate()` function is following the tutorial on https://www.nltk.org/book/ch01.html so I think it's okay since the `Text` object are not using in other places.

nltk/text.py
alvations: In the old code, it prints to `stdout`, 73f7e7b#diff-f5cb9102f911e93ca5fefa00b56a1e3f. I'm not sure what to do with it and getting a logger here might be overkill =)

nltk/text.py
alvations: Genau, exactly what the code is trying to do. It does get repetitive if someone loops through the random seeds in increasing order. But that's the best I can think of generating 100 words. Seemingly, the `</s>` just keeps predicting `</s>` afterwards and this looks like a problem that LM people hasn't solve, except maybe BERT like `<sep>` hacks.

alvations: Any other suggestion to the generation logic to generate 100 words?

nltk/text.py
alvations: Yeah, the original Text was presented as a demo. But I agree, maybe doing both would be good, print as well as return.

nltk/text.py
iliakur: Nah, it seems reasonable. I just wanted to make sure I understood it and then slept on it and couldn't come up with any clear improvement. Carry on :)

nltk/text.py
iliakur: yea, let's do that for posterity

nltk/text.py
iliakur: ok, fair enough.

nltk/text.py
iliakur: Alright, if it's really mostly for demo purposes (though why would you demo something that isn't used anywhere else?) then I guess we can leave this stuff...

nltk/text.py
iliakur: What do you mean by "don't keep sentence tokenization"? The attribute is being set, so that's definitely kept after the generation. If it only gets reset the next time generation happens, why don't we use a function-scoped variable for it?

nltk/text.py
iliakur: I doubt the original would let you seed the random stuff.

nltk/text.py
alvations: I mean the original input to the Text object already lost all notion of sentence because it was using the `word_tokenize` directly (which flattens the sentences).

nltk/text.py
alvations: Done.

nltk/text.py
alvations: Change it's to a private attribute convention `self._trigram_model`.

nltk/text.py
alvations: Change it's to a private attribute convention `self._tokenized_sents `.

nltk/text.py
alvations: Changed to `self. _train_default_ngram_lm()`.

nltk/text.py
alvations: Added default to docstring.

nltk/text.py
alvations: Blacken =)

alvations:
Thanks @Copper-Head for the review and all the helpful suggestions! I've committed the appropriate changes =)

iliakur:
Look at all the pesky whitespace and long lines that black deleted for us :)

Thanks for implementing the feedback, lgtm!

stevenbird:
Thanks @alvations and @Copper-Head... it's great to have this functionality back finally!

PR #2307, opened 2019-05-23
Extension of vader rules

itsmemala:
Updates to this version:
1.Added "only" (B_DECR) in the BOOSTER_DICT
2.Added contrastive conjunctions - however, except - in the function _but_check
3.Added performance optimization in the function _but_check: 
	a. Added enumerate() to retrieve index in the for loop on sentiments
	b. replaced pop & insert with single statement using indexing
4.Added new function _only_if_check: Decrease valence of words preceding "only if"
        Eg. This is a problem only if you have kids.
5.Added new function _in_spite_of_check: Increase valence of words preceding "in spite of" and decrease valence of succeeding words
       Eg. The host was very welcoming in spite of us arriving late.
Date: 23 May 2019
Author: Malavika Suresh <malavika.suresh0794@gmail.com>


alvations:
Thanks @itsmemala for the pull-request. Just like to check whether these improvements to Vader rules are documented somewhere in a publication or some blogpost with benchmarks against the original Vader?

itsmemala:
@alvations: Hi, the benchmarking against original Vader has been captured in terms of test cases in the attached documents: 
Vader_Orig_Test: Consists of sentiment scores on test sentences for original Vader code
Vader_Update_Test: Consists of the updated source code as well as sentiment scores on the same test sentences for the updated Vader code
Blogpost is coming up soon! (I'm currently busy with my post graduation thesis submission)
[Vader_Orig_Test.pdf](https://github.com/nltk/nltk/files/3263120/Vader_Orig_Test.pdf)
[Vader_Update_Test.pdf](https://github.com/nltk/nltk/files/3263121/Vader_Update_Test.pdf)

itsmemala:
Apologies, pls ignore files attached in previous comment and refer to these:
[Vader-Test.zip](https://github.com/nltk/nltk/files/3263172/Vader-Test.zip)

alvations:
@itsmemala thank you again for the proposed changes and examples of improved numbers but it's best that some more documentation and sufficient explanation is available before we are ready to merge the changes to the original Vader algorithm.

itsmemala:
@alvations - The results of the improvements benchmarked against the original Vader are now published at: https://medium.com/@malavika.suresh0794/vader-customizing-the-library-71d9e8ed6eda

alvations:
Thanks @itsmemala, the changes seems valid for the examples you've given in the blog post but there needs to be more empirical evidence to add them as the defaults in Vader algorithms. 

Unless the extension can show better scores in one of fixed benchmark sentiment analysis test set, a few examples without showing significant changes would not suffice, it'll be quite hard to overwrite the existing Vader lexicon given just the blogpost. Or if @cjhutto agrees with the changes, we can merge the changes after resolving the merge conflicts.

----

Otherwise, it might be possible to do refactor the code in a pretty massive way to make all the static lists/rules/magic numbers to be something configurable but it'll require quite some work.

alvations:
I have ~1 hour of spare time. Let me try refactoring and see how far I go.

alvations:
That should do most of the refactoring to retain the original Vader and incorporating your changes. 


To test:

```python
from nltk import sent_tokenize
from nltk.sentiment.vader import SentimentIntensityAnalyzer

sentences = ["VADER is smart, handsome, and funny.", # positive sentence example
   "VADER is smart, handsome, and funny!", # punctuation emphasis handled correctly (sentiment intensity adjusted)
   "VADER is very smart, handsome, and funny.",  # booster words handled correctly (sentiment intensity adjusted)
   "VADER is VERY SMART, handsome, and FUNNY.",  # emphasis for ALLCAPS handled
   "VADER is VERY SMART, handsome, and FUNNY!!!",# combination of signals - VADER appropriately adjusts intensity
   "VADER is VERY SMART, really handsome, and INCREDIBLY FUNNY!!!",# booster words & punctuation make this close to ceiling for score
   "The book was good.",         # positive sentence
   "The book was kind of good.", # qualified positive sentence is handled correctly (intensity adjusted)
   "The plot was good, but the characters are uncompelling and the dialog is not great.", # mixed negation sentence
   "A really bad, horrible book.",       # negative sentence with booster words
   "At least it isn't a horrible book.", # negated negative sentence with contraction
   ":) and :D",     # emoticons handled
   "",              # an empty string is correctly handled
   "Today sux",     #  negative slang handled
   "Today sux!",    #  negative slang with punctuation emphasis handled
   "Today SUX!",    #  negative slang with capitalization emphasis
   "Today kinda sux! But I'll get by, lol" # mixed sentiment example with slang and constrastive conjunction "but"
]

paragraph = "It was one of the worst movies I've seen, despite good reviews. \
Unbelievably bad acting!! Poor direction. VERY poor production. \
The movie was bad. Very bad movie. VERY bad movie. VERY BAD movie. VERY BAD movie!"


lines_list = sent_tokenize(paragraph)
sentences.extend(lines_list)


tricky_sentences = [
   "Most automated sentiment analysis tools are shit.",
   "VADER sentiment analysis is the shit.",
   "Sentiment analysis has never been good.",
   "Sentiment analysis with VADER has never been this good.",
   "Warren Beatty has never been so entertaining.",
   "I won't say that the movie is astounding and I wouldn't claim that \
   the movie is too banal either.",
   "I like to hate Michael Bay films, but I couldn't fault this one",
   "It's one thing to watch an Uwe Boll film, but another thing entirely \
   to pay for it",
   "The movie was too good",
   "This movie was actually neither that funny, nor super witty.",
   "This movie doesn't care about cleverness, wit or any other kind of \
   intelligent humor.",
   "Those who find ugly meanings in beautiful things are corrupt without \
   being charming.",
   "There are slow and repetitive parts, BUT it has just enough spice to \
   keep it interesting.",
   "The script is not fantastic, but the acting is decent and the cinematography \
   is EXCELLENT!",
   "Roger Dodger is one of the most compelling variations on this theme.",
   "Roger Dodger is one of the least compelling variations on this theme.",
   "Roger Dodger is at least compelling as a variation on the theme.",
   "they fall in love with the product",
   "but then it breaks",
   "usually around the time the 90 day warranty expires",
   "the twin towers collapsed today",
   "However, Mr. Carter solemnly argues, his client carried out the kidnapping \
   under orders and in the ''least offensive way possible.''"
]
sentences.extend(tricky_sentences)
# This should trigger the extension.
sid = SentimentIntensityAnalyzer(extended=True)
for sentence in sentences:
    print(sentence)
    ss = sid.polarity_scores(sentence)
    for k in sorted(ss):
        print('{0}: {1}, '.format(k, ss[k]), end='')
    print()
```

The default `sid = SentimentIntensityAnalyzer()` in the doctest should still keep the behavior of the original Vader.

nltk/sentiment/vader.py
```diff
@@ -433,30 +448,76 @@ def _least_check(self, valence, words_and_emoticons, i):
...
+                    bi = words_and_emoticons.index("BUT")
+                for sentiment in sentiments:
+                    si = sentiments.index(sentiment)
+                    if si < bi:
+                        sentiments.pop(si)
+                        sentiments.insert(si, sentiment * 0.5)
+                    elif si > bi:
+                        sentiments.pop(si)
+                        sentiments.insert(si, sentiment * 1.5)
+            return sentiments
```
alvations: @itsmemala I've not checked carefully but could you confirm that the original and the modified code works the same way?

nltk/sentiment/vader.py
```diff
@@ -433,30 +448,76 @@ def _least_check(self, valence, words_and_emoticons, i):
...
+
+    def _only_if_check(self, words_and_emoticons, sentiments):
+        words_and_emoticons_lower = [str(w).lower() for w in words_and_emoticons]
```
alvations: Would `words_and_emoticons` change in the process of the algorithm, if it doesn't putting it as a class variable and accessing it might be more efficient.

alvations (commented):
With the refactoring the original vader isn't overwrriten and the extension using the `extended=True` option, the code is a lot more maintainable but the lack of empirical evidence in why the extension better still lurking, so it'll be best if the extension is tested on some benchmark and the blogpost puts more rigor into proper benchmarking.

nltk/sentiment/vader.py
itsmemala: Yes, I can confirm.

nltk/sentiment/vader.py
itsmemala: I think the suggestion makes sense, I will look to incorporate it and also benchmark against a fixed test set. Really appreciate the comments and efforts !

cjhutto:
> Unless the extension can show better scores in one of fixed benchmark sentiment analysis test set, a few examples without showing significant changes would not suffice, it'll be quite hard to overwrite the existing Vader lexicon given just the blogpost. Or if @cjhutto agrees with the changes, we can merge the changes after resolving the merge conflicts.

In general, I fully support efforts to improve performance (both in terms of accuracy and efficiency). Bravo!

Intuitively, the rules for "however" and "except" (as additional contrastive conjunctions), "only if", and "in spite of" make good sense. 

On the other hand, I can easily think of examples where "only" should NOT be included in the degree modifier dict as a B_DECR (dampener) - at least not as default, generalized use.  Consider the following:
`My only love...`
`sprung from my only hate!`
I agree with @alvations -- a bit more empirical evidence might be needed to determine appropriate baseline behavior for its use as a B_DECR modifier.  Does this change result in better accuracy in one of the fixed benchmark sentiment analysis test set?

nltk/sentiment/vader.py
stevenbird: Yes, thanks @alvations

nltk/sentiment/vader.py
```diff
@@ -433,30 +448,76 @@ def _least_check(self, valence, words_and_emoticons, i):
...
+        check = 'only'
+        if check in words_and_emoticons_lower:
+            i = words_and_emoticons_lower.index(check)
+            if len(words_and_emoticons)>i+1 and 'if' == words_and_emoticons_lower[i+1]:
```
alvations: On a closer look, this is trying to find the index of the Multi-word expression "only if". Maybe using ngrams and index of ngrams would work.

nltk/sentiment/vader.py
```diff
@@ -433,30 +448,76 @@ def _least_check(self, valence, words_and_emoticons, i):
...
+                    if si < i:
+                        sentiments[si] = sentiment * 0.5
+        return sentiments
+
+    def _in_spite_of_check(self, words_and_emoticons, sentiments):
+        words_and_emoticons_lower = [str(w).lower() for w in words_and_emoticons]
+        check = 'in'
+        if check in words_and_emoticons_lower:
+            i = words_and_emoticons_lower.index(check)
+            if len(words_and_emoticons)>i+2 and 'spite' == words_and_emoticons_lower[i+1] and 'of' == words_and_emoticons_lower[i+2]:
```
alvations: Similarly, this looks like it's looking for the 3-grams "in spite of"

nltk/sentiment/vader.py
alvations: E.g. 

```python
from nltk.util import ngrams

x = "It was smooth sailing in spite of the trade tension".split()

for idx, ng in enumerate(ngrams(x, 3)):
    if ng == ('in', 'spite', 'of'):
        print(x[:idx])
```

alvations:
Maybe this is a better approach. After the refactoring, now the Vader constants and rules can easily be subclassed. So we remove all the proposed extensions but the extended algorithm can be easily created by:

```python
class ExtendedSentimentIntensityAnalyzer(SentimentIntensityAnalyzer):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.constants.BOOSTER_DICT["only"] = self.constants.B_DECR
        

    def _only_if_check(self, words_and_emoticons, sentiments):
        """
        Extension from https://github.com/nltk/nltk/pull/2307
        """
        for idx, ng in enumerate(ngrams(words_and_emoticons, 2)):
            if ng == ("only", "if"):
                for sidx, sentiment in enumerate(sentiments):
                    if idx < sidx:
                        sentiments[sidx] = sentiment * 0.5
        return sentiments

    def _in_spite_of_check(self, words_and_emoticons, sentiments):
        """
        Extension from https://github.com/nltk/nltk/pull/2307
        """
        for idx, ng in enumerate(ngrams(words_and_emoticons, 3)):
            if ng == ("in", "spite", "of"):
                for sidx, sentiment in enumerate(sentiments):
                    if sidx == idx+1:
                        sentiments[sidx] = 0
                    elif sidx < idx+1:
                        sentiments[sidx] = sentiment * 1.5
                    elif sidx > idx+1:
                        sentiments[sidx] = sentiment * 0.5
        return sentiments
    
    # Extend the default VADER scoring.
    def polarity_scores(self, text):
        """
        Return a float for sentiment strength based on the input text.
        Positive values are positive valence, negative value are negative
        valence.
        """
        # text, words_and_emoticons, is_cap_diff = self.preprocess(text)
        sentitext = SentiText(text, self.constants.PUNC_LIST,
                              self.constants.REGEX_REMOVE_PUNCTUATION)
        sentiments = []
        words_and_emoticons = sentitext.words_and_emoticons
        for item in words_and_emoticons:
            valence = 0
            i = words_and_emoticons.index(item)
            if (
                i < len(words_and_emoticons) - 1
                and item.lower() == "kind"
                and words_and_emoticons[i + 1].lower() == "of"
            ) or item.lower() in self.constants.BOOSTER_DICT:
                sentiments.append(valence)
                continue

            sentiments = self.sentiment_valence(valence, sentitext, item, i, sentiments)

        sentiments = self._but_check(words_and_emoticons, sentiments, self.extended)
        
        # Possible extension from https://github.com/nltk/nltk/pull/2307
        sentiments = self._only_if_check(words_and_emoticons_lowered, sentiments)
        sentiments = self._in_spite_of_check(words_and_emoticons_lowered, sentiments)
```

And once we have a clear and well-documented benchmark of the extension, we can easily create an `from nltk.sentiment.extended_vader import ExtendedSentimentIntensityAnalyzer`  =)

alvations:
@itsmemala would the proposed subclassing work? 

@cjhutto @stevenbird there's no extension to algorithm in the latest commit but it is worth merging the improved code.

We've managed to:

 - cleaned up Vader code a little
 - refactored Vader such that it can be easily extended with the VaderConstant class and additional rules
 - some improvements on `_but_check` function (that comes from the original PR)

nltk/sentiment/vader.py
itsmemala: Yep, you're right.

itsmemala:
@alvations : Yep the sub-classing idea sounds good ! Thanks for the time and efforts to help.

stevenbird:
Thanks @itsmemala, @alvations

cjhutto:
Thanks @itsmemala, @alvations - much appreciated.

PR #2311, opened 2019-05-29
None/boolean feature values in decision tree output: pretty_format() and pseudocode()

### LINKED ISSUE #1524: NoneType() < bool() error when doing pretty format of nltk decision tree classifier in python3
<classifier>.pretty_format() works fine with Python2. But this is the error I get with python3:
TypeError                                 Traceback (most recent call last)
<ipython-input-629-dc64618292e3> in <module>()
----> 1 classifier.pretty_format()

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/nltk/classify/decisiontree.py in pretty_format(self, width, prefix, depth)
     84             return '%s%s %s\n' % (prefix, '.'*n, self._label)
     85         s = ''
---> 86         for i, (fval, result) in enumerate(sorted(self._decisions.items())):
     87             hdr = '%s%s=%s? ' % (prefix, self._fname, fval)
     88             n = width-15-len(hdr)

TypeError: unorderable types: NoneType() < bool()

GerhardKa:
This is to fix #1524.

For classifier features having a boolean value or being of NoneType
and being mixed with features of type String,
sorted() does not work in Python3 (but it did in Python2).

This resulted in an error message and no output
when trying to pretty_format() the decision tree or printing it in pseudocode().

I added a sort key function, which creates a tuple for each decisions-item (feature value and result) to be sorted:

First element in the sort tuple is a truth value
(True if feature value is None/True/False, otherwise False),
second element ot the sort tuple is the item's feature value cast to a lowercase string.

This results in a sorting order where ...
- all items w/ feature value of non-None/True/False-type
    come first (first element of sort tuple is True and True < False)
    and will be sorted by their lowercase string version
- items w/ feature value of type None/True/False will follow
    (in the order of their lowercase string version, so: False, None, True)

This change will allow users working with feature values None/True/False in Python3,
being able to use pretty_format() and pseudocode().

Output feature values will still be sorted,
feature values None/True/False will be added at the end.


stevenbird:
@GerhardKa: would you please try to fix these merge and CI issues so we can proceed?

GerhardKa:
@stevenbird: done

alvations:
@GerhardKa Do you have an example sentence that broke the original code in Python3? We should add that in the regression test =)

nltk/classify/decisiontree.py
```diff
@@ -339,7 +346,7 @@ def demo():
     classifier = names_demo(
         f, binary_names_demo_features  # DecisionTreeClassifier.train,
     )
-    print(classifier.pp(depth=7))
+    print(classifier.pretty_format(depth=7))
```
alvations, line 349: 👍

alvations: `pp` was never even a function, not sure how it slipped through the CI at all =)

GerhardKa: It was the predecessor of `pretty_format` until 3.0.1

GerhardKa:
@alvations added a line in the `train` set in `classify.doctest` with a None-valued feature;
unfortunately, the output I produced with Python 3.7.3 doesn't fit exactly with the output of the other Python versions used for Travis CI.

Any suggestions?

alvations:
@GerhardKa now this is a little tricky with None value feature(s). 

It looks like different versions are returning different models, might be because of the feature set that's causing the differences, https://github.com/nltk/nltk/blob/develop/nltk/classify/decisiontree.py#L144

Difference on https://travis-ci.org/nltk/nltk/builds/561909117?utm_source=github_status&utm_medium=notification

I'm not sure why the non-None values doesn't have the deterministic decision trees cutting though =(

With `3.5.6`:

```
       b=0? .................................................. x
         c=0? ................................................ x
         c=1? ................................................ y
       b=1? .................................................. y
       b=None? ............................................... x
       <BLANKLINE>
```

With `3.6.5`:

```
     a=0? .................................................. x
       c=0? ................................................ x
       c=1? ................................................ x
         b=0? .............................................. y
         b=1? .............................................. y
         b=None? ........................................... x
     a=1? .................................................. y
     <BLANKLINE>
```

With `3.7.1`:

```
       c=0? .................................................. x
         a=0? ................................................ x
         a=1? ................................................ y
       c=1? .................................................. x
         b=0? ................................................ y
         b=1? ................................................ y
         b=None? ............................................. x
       <BLANKLINE>
```

nltk/classify/naivebayes.py
```diff
@@ -185,7 +185,8 @@ def most_informative_features(self, n=100):
             # Convert features to a list, & sort it by how informative
             # features are.
             self._most_informative_features = sorted(
-                features, key=lambda feature_: (minprob[feature_] / maxprob[feature_], feature_[0])
+                features, key=lambda feature_: (minprob[feature_] / maxprob[feature_], feature_[0],
+                                                feature_[1] in [None, False, True], str(feature_[1]).lower())
```
alvations, line 190: Tricky but that seems better =)

GerhardKa (commented):
@alvations: the train set line before was hard to learn from (for a deterministic decision tree)

GerhardKa (commented):
@alvations:  now sorts labels list from lowest to highest prob value
and (if they are the same) sorts label names reverse alphabetically -
output is last label in labels list first, so covers the case 
`[...]   x : y       =       1.0 : 1.0`
 ... where x is printed before y when probs have the same value (1.0).

stevenbird:
@GerhardKa: do we now get consistent behaviour across Python versions?

GerhardKa:
@stevenbird: Yes, I think so - regression test passed all Python versions in CI

stevenbird:
Thanks @GerhardKa and @alvations!

alvations:
Thank you @GerhardKa for spotting the problem and resolving the None sorts!! 

Haven't been putting out the `awesome-contribution` tag for quite some time but for the persisting the issue and resolving the checks on CI, definitely an `awesome-contribution`! So glad that the sort works right with None =)

GerhardKa:
This closes  #1524

Issue #2317, opened 2019-06-12
Import Error/Variable Error in probability.ConditionalFreqDist.plot()


dbalgley:
It seems like changes going back to 42ffffd36cec0fa10418e7d19294d5b91c2b05b6 (2/23/2019) are causing me issues.

1. On attempting to run example from Chapter 2 of the nltk book (below) I get an ImportError.  Seems to be related to the change made on Feb 23 specifically related to the imports:
`cfd = nltk.ConditionalFreqDist( ( genre, word ) for genre in brown.categories( ) for word in brown.words( categories=genre ) )
    cfd.plot()`

> Traceback (most recent call last):
  File "C:\XXX\PycharmProjects\ciExp\venv\lib\site-packages\nltk\probability.py", line 1907, in plot
    from matplotlib import plt
ImportError: cannot import name 'plt' from 'matplotlib' (C:\XXX\PycharmProjects\ciExp\venv\lib\site-packages\matplotlib\__init__.py)

2. Correcting the import statement to `import matplotlib.pyplot as plt` as it was before Feb 23 results in an UnboundedLocalError in the same plot() function.  There was a change on May 7 to that line where it removed a comment from `if v in self`: 

> File "C:\XXX\PycharmProjects\ciExp\venv\lib\site-packages\nltk\probability.py", line 1919, in plot
    kwargs, 'samples', sorted(set(v for c in conditions
  File "C:\XXX\PycharmProjects\ciExp\venv\lib\site-packages\nltk\probability.py", line 1920, in <genexpr>
    if v in self
UnboundLocalError: local variable 'v' referenced before assignment

Code snippet from probability.py:
`samples = _get_kwarg(
            kwargs, 'samples', sorted(set(v for c in conditions
                                          if v in self
                                          for v in self[c]))
        )  # this computation could be wasted`

3. Commenting out the `if v in self` line (probably ill advised but I wanted to walk this back) results in a NameError:

>   File "C:\XXX\PycharmProjects\ciExp\venv\lib\site-packages\nltk\probability.py", line 1940, in plot
    ax.plot(freqs, *args, **kwargs)
NameError: name 'ax' is not defined

[...]

PR #2318, opened 2019-06-13
Update probability.py

zhuyuanxiang:
hi, maybe I found a error in probability.py.
At Line 1907, "from matplotlib.pyplot import plt" maybe is the right code.

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -1904,7 +1904,7 @@ class ConditionalFreqDist(defaultdict):
         :type conditions: list
         """
         try:
-            from matplotlib import plt
+            from matplotlib.pyplot import plt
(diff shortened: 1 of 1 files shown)


stevenbird:
NB https://github.com/nltk/nltk/issues/2317

BLaZeKiLL:
this still causes the import error
replacing it with
`import matplotlib.pyplot as plt`
causes a different error
UnboundLocalError: local variable 'v' referenced before assignment

I have fixed it (at least it is working for me now) [#2319](https://github.com/nltk/nltk/pull/2319)

stevenbird:
Thanks @zhuyuanxiang - we'll accept a different PR for this

PR #2319, opened 2019-06-13
Plot function bugs fixed

BLaZeKiLL:
- fixed matplot lib import
- fixed samples query (v definition)
- fixed frequency and labels logic
- fixed test_plot unit test input arguments
- fixed non existent conditions


stevenbird:
Thanks @BLaZeKiLL

PR #2322, opened 2019-06-20
Corrected asterisk and dotdot* on punctuation list for word_tokenize

oya163:
I have added asterisk in punctuation list in TreebankWordTokenizer(). And, also added same in detokenizer but with unkown padding.


stevenbird:
@oya163 can you please explain why you are submitting this PR

oya163:
Because I thought **\*** is also a part of punctuation. I was working with wikipedia data and had to deal with tokenizing words like **\*Vanesa** or **\*James** etc. TreebankWordTokenizer() was not able to handle such cases. When I added **\*** in the punctuation list, it was able to tokenize into **\*** **Vanesa** or **\*** **James**

alvations:
~Asterisk is special symbol. Maybe it needs to be escaped.~

alvations:
@oya163 Sorry for the delayed response. 

My suggestion is not to change the default `TreebankWordTokenizer()` object but to "hack" the `word_tokenize` that has been modified upon the `TreebankWordTokenizer`. It's because there are quite some scripts in the wild that checks against already the "canonical" `TreebankWordTokenize` outputs. 

As for the `TreebankWordDetokenizer()`, there's some flexibility since it's something that NLTK created and it should be okay to mess around the regexes there =)

If you don't mind, I've made the changes as described above accordingly.

oya163:
Yes. Got it. Thanks @alvations

stevenbird:
[CI: retest]

alvations:
@stevenbird This should LGTM too after the test pass =)

alvations (approved):
It's a good catch!! Why haven't we realized this sooner? =)

alvations:
Since we're fixing the tokenizer, here's another one:

```
>>> from nltk import word_tokenize
>>> word_tokenize('This is a sentence with dotdot.. but the only the dotdotdot...gets tokenized correctly')
['This', 'is', 'a', 'sentence', 'with', 'dotdot..', 'but', 'the', 'only', 'the', 'dotdotdot', '...', 'gets', 'tokenized', 'correctly']
```

Note that the double dots isn't tokenized correctly. 


In that case the line at https://github.com/nltk/nltk/blob/5dd991dbad1e2fc407665f6f14c5de2768e77635/nltk/tokenize/treebank.py#L79 should be:

```
(re.compile(r"\.{2,}"), r" ... "),
```

[out]:

```
>>> import re
>>> text = 'This is a sentence with dotdot.. but the only the dotdotdot...gets tokenized correctly'
>>> replacement_dotdotdot_regex = re.compile(r"\.{2,}", re.U)
>>> re.sub(replacement_dotdotdot_regex, r" \g<0> ", text)
'This is a sentence with dotdot ..  but the only the dotdotdot ... gets tokenized correctly'
```

stevenbird:
Thanks @oya163, @alvations

PR #2323, opened 2019-06-21
Cfd plot fix

jbbe:
The plot function in ConditionalFreqDist was throwing an ImportError
because it was trying to import 'plt' from matplotlib which does not
exist. Additionally the methods called were still the methods of
they pylab submodule and I updated those calls to those of the pyplot
submodule. Additionally the `if v in self` key check would not pass
with valid input so I removed it and the function now plots correctly.


stevenbird:
Thanks @jbbe.
Resolved in #2319

Issue #2347, opened 2019-07-30
METEOR metric does not use user set stemmer 


ndarr:
The Meteor metric offers the ability to use a custom stemmer but does not pass it through to the relevant function. So in every case even if the function meteor_score is called with a custom stemmer it still uses the default PorterStemmer

PR #2349, opened 2019-07-30
Added missing pass of custom stemmer object in meteor score

ndarr:
This fix is related to issue #2347


stevenbird:
Thanks @ndarr!

PR #2353, opened 2019-08-03
Provide  actionable error message in RegexpTagger

pombredanne:
The RegexpTagger error messages are obscure when a regexp has an error.
This is because re messages may be obscure too.

To help with tracing the cause of regex errors, we now catch regex
compilation errors when a RegexpTagger is constructed and raise a new
exception that contains the required diagnostic information with the
error message, the faulty regex and its corresponding tag.

The RegexpTagger.decode_json_obj has been streamlined also to avoid code
duplication and use the default constructor instead.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/nltk/tag/sequential.py
+++ b/nltk/tag/sequential.py
@@ -544,7 +544,11 @@ class RegexpTagger(SequentialBackoffTagger):
         """
         """
         SequentialBackoffTagger.__init__(self, backoff)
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @pombredanne!

pombredanne:
@stevenbird thank you ++ for the quick review and merge!

PR #2355, opened 2019-08-05
Create NLTK own's Tokenizer class

alvations:
This should clean up the hacks we've put onto `TreebankWordTokenizer` and now any new improvement to the tokenizer should be on the `nltk.tokenize.nltk_tokenize.NLTKDestructiveWordTokenizer` class. 

With that we can "freeze" the `TreebankWordTokenizer` so that it can replicate the original sed script behavior and we can continually improve on the NLTK tokenizer, esp for `word_tokenize` function without conflicts of back-compatibility with treebank's tokenizer.

I'm bad with naming though, so the `nltk.tokenize.nltk_tokenize` sounds a little awkward. Any better suggestions?=)


stevenbird:
@alvations: good idea; still puzzling over naming...

alvations:
The hope is that someday we'll write an `NLTKWordTokenizer` that keeps the character offset so that the detokenizer can work perfectly, thus the "destructive" =)

But yeah, I'm back with the names.

alvations:
@stevenbird I hope the renaming/refactoring sounds better now.

# Functional usage 

No change from current version.

```python
>>> from nltk import word_tokenize
>>> word_tokenize("How's it going?")
['How', "'s", 'it', 'going', '?']
```

# Object usage

```python
>>> from nltk.tokenize import NLTKWordTokenizer
>>> tokenizer = NLTKWordTokenizer()
>>> tokenizer.tokenize("How's it going?")
['How', "'s", 'it', 'going', '?']
```

# Internally

The absolute path to the tokenizer object from the main library's path.

```python
from nltk.tokenize.destructive import NLTKWordTokenizer
```

Usually the absolute path won't be used since the `nltk.tokenize` module already imported the `NLTKWordTokenizer`.

alvations:
Any suggestions on the naming? 

 - `from nltk.tokenize.destructive import NLTKWordTokenizer` (current)
 - `from nltk.tokenize.default import NLTKWordTokenizer`
 - `from nltk.tokenize.defacto import NLTKWordTokenizer`
 - `from nltk.tokenize.original import NLTKWordTokenizer`

Any other suggestions?

stevenbird:
@alvations: thanks, that is better naming, and I haven't been able to think of an improvement!

Issue #2338, opened 2019-07-16
DeprecationWarnings for formatargspec and collections in python3.7


Attryy:
Hi there, I was upgrading python2.7 to python3.7. When running pytests, bunch of warnings pop up from nltk library. 

Line 68 in decorators.py
`DeprecationWarning: 'formatargspec' is deprecated since Python 3.5. Use 'signature' and the 'Signature' object directly` 

Line 15 in counter.py and line 13 in vocabulary.py
`DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`

alvations:
Patched from #2364

asevans48:
So, this doesn't seem resolved from the pypi version 1.12.0. Is this correct? I use pytest and receive the same warnings. Is this secure?

alvations:
@asevans48 This should have been patched by #2364 on the develop branch. The next version of the NLTK release should be "pip-able" soon =)

Closed issues usually means it's solved in the `develop` branch of the code and not the release. Sorry for the confusion.

bbookman:
Still getting this issue as of 3.4.5 released Aug of 2019.  Please release this fix!  Is there a method of suppressing the warning?

bbookman:
> @asevans48 This should have been patched by #2364 on the develop branch. The next version of the NLTK release should be "pip-able" soon =)
> 
> Closed issues usually means it's solved in the `develop` branch of the code and not the release. Sorry for the confusion.

Not yet released from my experience with 3.4.5

PR #2359, opened 2019-08-09
Use collections.abc when needed for Python 3.8 compatibility.

hervenicol:
Since Python version 3.3, Collections Abstract Base Classes have been moved to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.

(source: https://docs.python.org/3/library/collections.html)

Also, fixes part of #2338.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -256,6 +256,7 @@
 - Devashish Lal <https://github.com/BLaZeKiLL>
 - Gerhard Kremer <https://github.com/GerhardKa>
 - Nicolas Darr <https://github.com/ndarr>
+- Hervé Nicol <https://github.com/hervenicol>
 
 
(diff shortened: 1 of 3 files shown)


iliakur:
afaik NLTK no longer supports Python versions below 3.5, which makes the checks unnecessary.

In fact, in #2363 I've taken the plunge to `collections.abc` along with a lot of other simplifications related to dropping older version support. Unfortunately that supercedes this PR, sorry about that.

While you have this stuff in the focus of attention, you could try fixing the other Deprecation warning in decorators.py. That would be a solid contribution :)

hervenicol:
OK, I thought Python 2 compatibility was expected. Given the cleanup in your PR, I see it's not the case ;)
The decorators.py deprecation warning is a bit more touchy. But I'll give it a try.

PR #2364, opened 2019-08-11
Formatargspec deprecation warning

hervenicol:
Fixes the part of  #2338 that @Copper-Head has not fixed yet.

Also, while I was on `decorators.py`:
- removed some Python 2 support code
- removed a few pylint alerts


nltk/decorators.py
```diff
@@ -79,6 +95,7 @@ def getinfo(func):
         name=func.__name__,
         argnames=argnames,
         signature=signature,
+        fullsignature=fullsignature,
```
hervenicol, line 98: I kept signature untouched, and added a `fullsignature` for the standard Signature object.

A cleaner version would be to replace `signature` with the Signature object. But it may break some user's code.
I'm open to suggestions if you have a better solution.

stevenbird:
Let's go with this; we can always revisit this later.

stevenbird:
Thanks @hervenicol

stevenbird:
Although the CI reported this was ok, Travis reports a problem post-merge: https://travis-ci.org/nltk/nltk/builds/572171148

hervenicol:
The initial error seems to be:
```
  Resource gutenberg not found.

  Please use the NLTK Downloader to obtain the resource:

  >>> import nltk

  >>> nltk.download('gutenberg')
```
I think I've had the same issue once, that it failed at downloading required resources. At second attempt it worked.
On top of that, even @alvations attempt at reverting my changes results in this error.
Could it be that the server hosting these resources had a problem?

I just tried to `tox -e py-travis tox.ini -vv`, and it worked here, so I guess this server is back on tracks.

Could you try re-launch the Travis build now ?

hervenicol:
Actually it seems worse than a temporary server issue, and started happening before my branch's merge.
The Travis builds history (https://travis-ci.org/nltk/nltk/builds) shows no build passed since build 830 / commit c26d875.

Still, I don't catch why it works on my environment with the latest develop branch.

asevans48:
Why are both requests closed if the issue is not fully resolved?

alvations:
@asevans48 This should have been patched by #2364 on the develop branch. The next version of the NLTK release should be "pip-able" soon =)

PR #2366, opened 2019-08-15
Fix JSON serialization of some PoS taggers

alexandredias3d:
I tried to find an alternative for serializing taggers other than using Pickle. I believe this is important to provide trained models in a safe fashion using the library.

This pull request fixes the error when using ```JSONTaggedEncoder``` on a ```NgramTagger``` with ```n``` greater than 1. Without the fix, it tries to serialize tuples and raises ```TypeError```. The workaround uses ```ast.literal_eval``` to read tuples in `repr` format. Also, this pull request can encode and decode PerceptronTagger and BrillTagger correctly.

### DIFF
--- a/nltk/tag/perceptron.py
+++ b/nltk/tag/perceptron.py
@@ -27,7 +27,7 @@ except ImportError:
 PICKLE = "averaged_perceptron_tagger.pickle"
 
 @jsontags.register_tag
-class AveragedPerceptron(object):
+class AveragedPerceptron:
 
     """An averaged perceptron, as implemented by Matthew Honnibal.
 
--- a/nltk/tag/perceptron.py
+++ b/nltk/tag/perceptron.py
@@ -17,6 +17,8 @@ import logging
 from nltk.tag.api import TaggerI
 from nltk.data import find, load
 
+from nltk import jsontags
+
 try:
     import numpy as np
 except ImportError:
@@ -24,7 +26,7 @@ except ImportError:
 
 PICKLE = "averaged_perceptron_tagger.pickle"
 
-
+@jsontags.register_tag
 class AveragedPerceptron(object):
 
     """An averaged perceptron, as implemented by Matthew Honnibal.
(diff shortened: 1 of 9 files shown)


alexandredias3d:
I would like to provide tests for these features. Should I write them as unit or doc tests?

Also, shouldn't the module ```tqdm``` be included in the requirements?

alvations:
@alexandredias3d Sorry about the `tqdm`, this shouldn't affect anything other than parallizable functions, localizing the imports inside the function would make the tqdm requirement optional only to the CLI functions. #2370

nltk/tag/perceptron.py
```diff
@@ -33,9 +35,11 @@ class AveragedPerceptron(object):
         https://explosion.ai/blog/part-of-speech-pos-tagger-in-python
     """
 
-    def __init__(self):
+    json_tag = "nltk.tag.perceptron.AveragedPerceptron"
+
+    def __init__(self, weights=None):
         # Each feature gets its own weight vector, so weights is a dict-of-dicts
-        self.weights = {}
+        self.weights = weights if weights else {}
```
alvations, line 42: 👍 Much cleaner with optional weights.

alexandredias3d:
I have one last question before I close this PR: is there a reason why the ```super``` keyword is not being used in ```sequential.py```? I have seen several cases where ```SuperClass.__init__()``` is used instead of ```super().__init__()```. I am not sure whether this is related to Python 2 support or not. If it is, I can change it to the ```super().__init__()``` notation.

alvations:
Yes, it's right. The `SuperClass.__init__()` patterns comes from Python 2 code. It should be safe to change it to `super().__init__()` as long as the regression tests checks out =)

BTW, awesome tests for the serialization and loading!

alvations (approved):

alexandredias3d:
@alvations I am not used to contributing to open source projects, so I have no idea if I made the right thing closing the Pull Request :grin:. Anyway, I am mentioning you so that this PR will not be forgotten.

alvations:
@alexandredias3d Thank you again for the contribution!! It's a really well written piece of code. 

Nope, the right thing to do is to merge the pull-request (PR) not close it =)

No worries, I've reopened it. And this PR looks good to merge (LGTM). We'll wait for @stevenbird to look through a last time and them merge it so that everyone using NLTK gets the new code =)

stevenbird:
Nice work @alexandredias3d. Thanks @alvations

PR #2367, opened 2019-08-16
Update decorators.py

alvations:
Patching CI from https://github.com/nltk/nltk/commit/81490305bbbb42c6b7fa62a4694d65607b7f420f


alvations:
@stevenbird It looks like the `fullsignature` from #2364 works differently from what NLTK is expecting for the third party apps. We've to dig digger why before putting back the `fullsignature` in the `get_info()`.

iliakur:
@alvations how much of `decorators.py` can we get rid of by using the [decorator](https://pypi.org/project/decorator/) library?

alvations:
Seems like the decorators isn't the thing causing the CI issues, c.f. #2370 

Closing this PR but the idea to replace all custom decorators to the `decorator` library is a good idea, leaving it in #2371

PR #2406, opened 2019-09-04
Experimental wrapper over SpaCy

alvations:
This should automate installation of spacy.

```
pip install -U nltk[spacy] 
```

(Note: The usual `pip install -U nltk[all]` would also install SpaCy.)

And to use, the `from nltk import astronaut` line must come first:

```python
>>> from nltk import astronaut 
>>> from nltk import word_tokenize, pos_tag

>>> word_tokenize("how are you? I'm fine. Thank you")
['how', 'are', 'you', '?', 'I', "'m", 'fine', '.', 'Thank', 'you']

>>> pos_tag(word_tokenize("how are you? I'm fine. Thank you"))
[('how', 'WRB'), ('are', 'VBP'), ('you', 'PRP'), ('?', '.'), ('I', 'PRP'), ("'", "''"), ('m', 'VBZ'), ('fine', 'JJ'), ('.', '.'), ('Thank', 'VBP'), ('you', 'PRP')]

>>> word_tokenize
<function nltk.astronaut.<lambda>(input)>

>>> pos_tag
<function nltk.astronaut.<lambda>(input)>
```

### DIFF
--- a/nltk/astronaut.py
+++ b/nltk/astronaut.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+# Natural Language Toolkit: Spacy wrapper
+#
+# Copyright (C) 2001-2019 NLTK Project
+# Authors:
+# Contributors:
+# URL: <http://nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""
+This is SpaCy wrapper that overrides the existing NLTK functions:
+    - word_tokenize
+    - pos_tag
+It'll emulate the NLTK API to the NLP functions,
+but uses the `en_core_web_sm` models from SpaCy (https://spacy.io/models)
+"""
+
+import sys
+
+try: # Try importing SpaCy if it's installed.
+    import spacy
+    import en_core_web_sm
+    nlp = en_core_web_sm.load()
+
+    module = sys.modules['nltk']
+
+    def spacy_pos_tag(input):
+        if all(isinstance(item, str) for item in input):
+            input = " ".join(input)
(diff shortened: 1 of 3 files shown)


iliakur:
Sorry to be a downer, but we're adding a layer of abstraction here and there's a price for that.
Could you give a short cost-benefit summary of this change?

alvations:
It's some out of the blue idea, purely thought experiment =)

My thought was to get something for easy benchmarking for the use of something like this, esp. in a specialize CI workflows on specific branch. Same abstraction/overloading idea can be used to overwrite the default `word_tokenize` and `pos_tag` for external packages too, e.g. new languages can use similar `override_model` to globally change the commonly used NLTK's function.


**Pros:**
 - easy benchmarking
 - general workflow to extend for other external overriding

**Cons:**
 - additional dependencies
 - not even sure who would use something like this ;P

Anyone can think/come up with more pros/cons?

alvations:
After sleeping through the idea, it's a rather bizarre one, with the 10MB, 90MB and >700MB models from spacy and for multiple languages. Maybe it's a bad idea to naively overwrite with a single model. 

But something for benchmarking across tools would be nice. Maybe a better idea/proposal another time.

iliakur:
I like that you tried it out :) Sometimes that's the only way to scratch an itch and see if it's real.

Sorry for jumping right away with the criticism. Maybe we can mark these PRs with a label or a WIP tag so that we all know when we're just messing around in a sandbox?

PR #2408, opened 2019-09-05
Move source transpiling with Cython into custom build command

hoefling:
Also, adding Cython to build deps via `setup_requires`. See [this comment](https://github.com/nltk/nltk/pull/2289#issuecomment-528374614) reasoning for changes (the optional change part at the bottom).

This should allow the user to invoke

```sh
$ CYTHONIZE_NLTK=true pip install nltk --no-binary=nltk
```

without it raising an `ImportError` when Cython is not installed beforehand. Another side effect is that

```sh
$ CYTHONIZE_NLTK=true python setup.py sdist
```
will not invoke the unnecessary cythonizing (although admitted the time saved in sdist packaging will be negligible).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
   apt:
     packages:
       - expect-dev # for unbuffer, see https://github.com/travis-ci/travis-ci/issues/7967
+      - graphviz
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,6 @@ matrix:
     - python: '3.7' # Only runs checks on Cythonization for latest Python distribution.
       env:
         - TOXENV="py-cythonized"
-        - CYTHONIZE_NLTK="true"
 
 before_install:
   - source tools/travis/pre-install.sh # Checks Java and Python versions.
--- a/nltk/metrics/segmentation.py
+++ b/nltk/metrics/segmentation.py
@@ -225,7 +225,7 @@ def pk(ref, hyp, k=None, boundary="1"):
 
 
 # skip doctests if numpy is not installed
-def setup_module(module):
+def setup_module():
     from nose import SkipTest
 
     try:
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -479,7 +479,7 @@ class DependencyGraph(object):
         ... }
         >>> cyclic_dg.root = top
 
-        >>> cyclic_dg.contains_cycle()
+        >>> cyclic_dg.contains_cycle()  # doctest: +SKIP
         [3, 1, 2, 4]
 
         """
--- a/nltk/tokenize/stanford.py
+++ b/nltk/tokenize/stanford.py
@@ -120,7 +120,7 @@ class StanfordTokenizer(TokenizerI):
         return stdout
 
 
-def setup_module(module):
+def setup_module():
     from nose import SkipTest
 
     try:
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -298,7 +298,7 @@ class StanfordSegmenter(TokenizerI):
         return stdout
 
 
-def setup_module(module):
+def setup_module():
     from nose import SkipTest
 
     try:
--- a/setup.py
+++ b/setup.py
@@ -146,7 +146,12 @@ natural language processing.  NLTK requires Python 3.5, 3.6, or 3.7.""",
         "Topic :: Text Processing :: Indexing",
         "Topic :: Text Processing :: Linguistic",
     ],
-    package_data={"nltk": ["test/*.doctest", "VERSION"]},
+    package_data={
+        "nltk": ["VERSION"],
+        "nltk.test": ["*.doctest", "*.xml"],
+        "nltk.metrics": ["artstein_poesio_example.txt"],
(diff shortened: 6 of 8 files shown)


hoefling:
Ok, inspecting the travis logs more closely now and I think there is more to do:

1. Even though the cythonized wheel is installed, the tests are being executed with the sources b/c the working dir is the repo root so `nltk` is imported from source code and not from the installed package under `.tox/py-cythonized/lib/python/site-packages/nltk`.
1a. One solution to this would be changing the working dir in the `py-cythonized` job via the [`changedir`](https://tox.readthedocs.io/en/latest/config.html#conf-changedir) setting, but I'm unsure whether this will have side effects. Will try out soon.
1b. Another solution would be moving the sources to the `src` dir that hides repository code from accidental importing; this is a lot cleaner IMO, but again, I suspect there will be a lot of stuff to be adapted and people will probably have their local stuff broken after the change, so this is for package maintainers to decide.

2. ATM each python module is transpiled, including the `__init__.py` scripts - this is useless and may break some IDEs (e.g. Pycharm starts to generate weird source maps for compiled code), so I would filter them out.

~~3. ATM the python modules are installed alongside the C extensions, resulting in this layout:
    ```sh
    $ pip show -f nltk | grep nltk/translate/api
      nltk/translate/api.cpython-36m-x86_64-linux-gnu.so
      nltk/translate/api.cpython-36m-x86_64-linux-gnu.so.avx2
      nltk/translate/api.cpython-36m-x86_64-linux-gnu.so.avx512
      nltk/translate/api.py
    ```
    (don't mind the AVX stuff, it's from my local box). The problem here is that source modules will be preferred at least by CPython, so although the compiled stuff is installed, it won't be used. This means additional filtering is required (for each extension, don't include the corresponding python module into distribution).~~ Scratch this, my local issue with custom Python patches. 

4. The C extensions take a long time to build b/c only one core is allocated for compilation by default - we can improve that by using `--parallel=N` (`N` is number of processes) in the build. Best is to write smth like
    ```ini
    [build_ext]
    parallel=17
    ```
    in a `setup.cfg` or `~/.pydistutils.cfg` in the `py-cythonized` Travis job so `distutils` can speed up.

I can solve 2. and 4. on my own, but 1. will require a decision I can't make.

hoefling:
Simple test for 1.:

```python
def test_import_loc():
    import os
    import nltk

    repo_nltk_pkg = os.path.realpath(os.path.join(__file__, '..', '..', '__init__.py'))
    assert nltk.__file__ != repo_nltk_pkg
```

This shows that `nltk` is imported from repo, ignoring whatever is installed in the tox env. Unfortunately, I didn't get far in trying to change the working dir - maybe open another issue for that?

hoefling:
Ok, running the compilation in parallel saved 30 seconds; not something I was trying to achieve ([23 minutes 49 seconds](https://travis-ci.org/nltk/nltk/builds/581225049) -> [23 minutes 13 seconds](https://travis-ci.org/nltk/nltk/builds/581298942)). Maybe the last commit with the `.pydistutils.cfg` hack is not worth it?

hoefling:
I saw the tests are also packaged and installed, so a possible solution to 1. could be navigating to installation target `.tox/.../site-packages/nltk/test`. Shouldn't the tests also pass in that case? See [the latest Travis build](https://travis-ci.org/nltk/nltk/builds/581395495) for error logs. I see some test files not being packaged, so this is an easy fix, but why are some tests selected when installed, but omitted when running from the source directory?

In the `py-cythonized` job, the test run doesn't even start. Looks like cython doesn't support wrapping `abstractmethod` into `staticmethod`, see issues [#2651](https://github.com/cython/cython/issues/2651)/[#1434](https://github.com/cython/cython/issues/1434). Usages in `nltk`: `nltk/tbl/feature.py` and `nltk/metrics/association.py`. Should these modules be excluded from compilation?

hoefling:
Ok, think I got it ready for a review. @alvations @stevenbird can you take a look at the proposed changes? Summary list:

1. Moved code cythonizing and extensions setting in a custom `build` command impl.
2. Added Cython to build dependencies if `CYTHONIZE_NLTK` is set.
3. Added test data to distribution to ensure the user can run `nltk` tests after installation (e.g. to verify the installation is working).
4. Changed test execution location into installation targets in Travis jobs - this also tests whether the user can run `nltk` tests after installation. This also has a side effect of finding more tests than currently on the `cythonize` branch - 685 against 674, no idea where these additional 11 tests come from.
5. Removed `module` argument from `setup_module` functions as `nose` incorrectly processes them in compiled extensions.
6. Skipped `nltk.parse.dependencygraph.DependencyGraph.contains_cycle` doctest - it passes with Python 3.5, but fails with Python 3.6/3.7 (observe the failure e.g. [in this job](https://travis-ci.org/hoefling/nltk/jobs/582334735)).
7. Turned off coverage invocation in `py-cythonized` job as `coverage` can't record compiled objects per se. There are options to compile Cython extensions with line tracing (see e.g. [here](http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html)), but idk whether it's worth it as there are neither `.pyx` extensions nor wrapped C libraries.

alvations (approved):
The build and the CI logs look right. 

Thank you @hoefling !!

alvations:
@stevenbird I think it LGTM =)

stevenbird:
Thanks @hoefling!

PR #2413, opened 2019-09-20
Fixed typo in a docstring

xmcgcg:
The first parameter of nltk.tag.pos_tag_sents is called `sentences`, but it is named as `tokens` in the docstring.


stevenbird:
Thanks @xmcgcg

Issue #2417, opened 2019-09-25
nltk.text --> findall improviment to return hits


marcelomsousa:
To improve findall in nltk.text I suggest to change "print(tokenwrap(hits, "; "))" in def findall(self, regexp) to "return hits". In this way it would be possible to get a list of values returned by findall. 

The way it is implemented if you do: 

>>> x = text5.findall("<.*><.*><bro>") 
you rule bro; telling you bro; u twizted bro
>>> x 

This return only a print; x will be empty. 
This, will be no control of the output.

If "print(tokenwrap(hits, "; "))" would be change to "return hits", x would be:
>>> x
['you rule bro', 'telling you bro', 'u twizted bro']

Seems much better!

Thank you.

marcelomsousa:
to 3.4.5 version

stevenbird:
Text.py is just demonstration code; you can use it as a model, and change what you like.

PR #2418, opened 2019-09-25
Update text.py

marcelomsousa:
change "print(tokenwrap(hits, "; "))" in def findall(self, regexp) to "return hits" as proposed at issue #2417


marcelomsousa:
It would be to 3.4.5 version.

stevenbird:
Duplicate of #2419

PR #2419, opened 2019-09-25
Update text.py

marcelomsousa:
change "print(tokenwrap(hits, "; "))" in def findall(self, regexp) to "return hits" as proposed at issue #2417


stevenbird:
Please see my comment on #2417

Issue #2320, opened 2019-06-15
One-shot BLEU-[2, 3, 4] computation


marco-roberti:
Hello everyone,

I need to compute the BLEU score with more than one ngram length (ideally, BLEU2, BLEU3, BLEU4, and BLEU5). In my case, this is a very long task, as every hypothesis has some thousand references.

Reading the implementation of the corpus_bleu function, which takes weights:Tuple between its parameters - and thus calculating BLEU-[len(weights)] - , I found out that it gets all the information to compute BLEU-m s.t. 2 <= m < len(weights).
Wouldn't it be nice to have a more general function that can compute BLEU with different ngram lengths at the same time? A possible implementation would be the weights parameter accepting a list of weight tuples as value, computing precisions that are useful for the longest tuple and using those precisions to get all the desired BLEU scores using the weights list.

This would result in a more general implementation and it would avoid the computation waste of calculating the same values more than one time.

stevenbird:
This would be nice to have... it's a question of who would like to implement it.

agannon:
Hi, I was looking at getting involved with contributing to NLTK and saw this with the 'goodfirstbug' tag. I will take a crack at this problem if that is OK.

alvations:
Feel free to take a look at https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md and create a pull-request and someone will review it before merging.

BatMrE:
Hi @stevenbird @alvations if its currently open can I get involved with it??

tomaarsen:
@BatMrE Yes, you may. We welcome contributions!
Feel free to look at [CONTRIBUTING.md](https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md) for more information.

BatMrE:
@ tomaarsen @alvations 
From the issue raised statement:
`Wouldn't it be nice to have a more general function that can compute BLEU with different ngram lengths at the same time? `


I will ultimately be calling the Bleu funnction for each of the  2 , 3 ,4 ... len(weights) , so how will it be saving computation waste of calculating the same values more than one time??

tomaarsen:
@BatMrE 
`corpus_bleu` (and also `sentence_bleu`) are called with a `weights` parameter. This is a list or tuple of floats with a certain length, such that the weights generally sum to 1. The length of this list or tuple determines whether BLEU-2, BLEU-3, BLEU-4, etc. is used. For example, with a length of 2, BLEU-2 is used. 

[...]

PR #2424, opened 2019-10-10
Feature/multiple ngram bleu

agannon:


nltk/translate/bleu_score.py
```diff
@@ -106,6 +106,16 @@ def sentence_bleu(
     )
 
 
+def corpus_bleu_multiple_weights(
+        list_of_references,
+        hypotheses,
+        weights,
+        smoothing_function=None,
+        auto_reweigh=False
```
stevenbird: Is this a typo? Where is this parameter used?

nltk/translate/bleu_score.py
```diff
@@ -106,16 +106,17 @@ def sentence_bleu(
...
-def corpus_bleu(
-    list_of_references,
-    hypotheses,
-    weights=(0.25, 0.25, 0.25, 0.25),
-    smoothing_function=None,
-    auto_reweigh=False,
+def corpus_bleu_multiple_weights(
+        list_of_references,
+        hypotheses,
+        weights_list,
```
alvations: There's no need to rename the class here. Also why is the variable `weights_list` more desired than the defaults `weights` that most users of BLEU expect?

nltk/translate/bleu_score.py
```diff
@@ -106,16 +106,17 @@ def sentence_bleu(
...
+        smoothing_function=None,
+        auto_reweigh=False
 ):
     """
     Calculate a single corpus-level BLEU score (aka. system-level BLEU) for all
-    the hypotheses and their respective references.
+    the hypotheses and their respective references for each weighting tuple in
+    weights_list
```
alvations: This is rather non-normal use of BLEU for >4-grams BLEU, users can already add to the weights list in the original implementation and change the n order.

alvations:
Thanks @agannon for the PR. I think this PR is meant to resolve the issue where you want the output to return the fine-grained scores for BLEU. Is that right? 

~The PR's approach to nesting `bleu -> corpus_bleu -> corpus_bleu_multiple_weights` is rather nasty and spaghetti-like. Maybe there's a better approach to design the API for BLEU.~

But I can't think of a better API off my head now.

nltk/translate/bleu_score.py
alvations: Ah the normal corpus_bleu has been moved down to another function.

nltk/translate/bleu_score.py
alvations: Ah the normal corpus_bleu has been moved down to another function.

agannon:
> Thanks @agannon for the PR. I think this PR is meant to resolve the issue where you want the output to return the fine-grained scores for BLEU. Is that right?
> But I can't think of a better API off my head now.

Yes, based on this issue: https://github.com/nltk/nltk/issues/2320.

I wanted to avoid changing the type signatures to not disrupt how users already call `corpus_bleu`, but I can see how having 2 functions would be confusing down the line. In terms of a different API do you think it worthwhile to have one function `corpus_bleu` with its type signature as:
```
def corpus_bleu(list_of_references: List[List[str]],
                           hypotheses: List[str],
                           weights: Union[Tuple[float], List[Tuple[float]]],
                           smoothing_function: SmoothingFunction = None,
                           auto_reweigh: bool = False) ->
                           Union[float, List[float]]
```
where weights can either be the original tuple of values OR a list of tuples to consider, and the return is either a single float or a list of corresponding floats.
That way users have the flexibility of using either or.

agannon:
Hi @alvations, I refactored my original changes to incorporate them into one single function. Please let me know if you think this is a better API than before

Edit: I am now seeing that it failed on py3.5 so I will look into that.

agannon:
Hi @alvations , could you let me know if this API is more to your liking than the nested corpus_bleu functions from earlier? I want to make sure I'm on the right track before I rebase to add the most recent nltk/develop commits to this PR

alvations:
@agannon sorry for the delay, let me look at this in the next couple of days and get back to you.

PR #2428, opened 2019-10-16
Implemented nltk.lm.LanguageModel.suggest

antran22:
Implemented a `suggest` method for LanguageModel, which gives out a list of words paired with their score that can follow a given context.
In order for this method to work, another method `_weighted_sample` must be implemented. This method sample k elements without replacement in a given population.
Also I've rewritten `_weighted_choice` to accept another `k` parameter, which is the number of the sampled results.

### DIFF
--- a/nltk/lm/api.py
+++ b/nltk/lm/api.py
@@ -16,37 +16,15 @@ from nltk.lm.counter import NgramCounter
 from nltk.lm.util import log_base2
 from nltk.lm.vocabulary import Vocabulary
 
-try:
-    from itertools import accumulate, repeat
-except ImportError:
-    import operator
-
-    def accumulate(iterable, func=operator.add):
-        """Return running totals"""
-        # accumulate([1,2,3,4,5]) --> 1 3 6 10 15
-        # accumulate([1,2,3,4,5], operator.mul) --> 1 2 6 24 120
-        it = iter(iterable)
-        try:
-            total = next(it)
-        except StopIteration:
-            return
-        yield total
-        for element in it:
-            total = func(total, element)
-            yield total
-
-    def repeat(value, count):
-        """An iterator with the specified value repeated for count time"""
-        for _ in range(count):
-            yield (value)
+from itertools import accumulate, repeat
 
 
 @add_metaclass(ABCMeta)
-class Smoothing(object):
+class Smoothing:
     """Ngram Smoothing Interface
 
     Implements Chen & Goodman 1995's idea that all smoothing algorithms have
-    certain features in common. This should ideally allow smoothing algoritms to
+    certain features in common. This should ideally allow smoothing algorithms to
     work both with Backoff and Interpolation.
     """
 
@@ -125,7 +103,7 @@ def _weighted_sample(population, weights=None, random_generator=None, k=1):
 
 
 @add_metaclass(ABCMeta)
-class LanguageModel(object):
+class LanguageModel:
     """ABC for Language Models.
 
     Cannot be directly instantiated itself.
@@ -160,8 +138,7 @@ class LanguageModel(object):
         if not self.vocab:
             if vocabulary_text is None:
                 raise ValueError(
-                    "Cannot fit without a vocabulary or text to " "create it from."
-                )
+                    "Cannot fit without a vocabulary or text to create it from.")
             self.vocab.update(vocabulary_text)
         self.counts.update(self.vocab.lookup(sent) for sent in text)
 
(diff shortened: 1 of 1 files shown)


iliakur:
Just from skimming this PR I have a few comments, but would like to take some time to provide better feedback. Will do so within the next two weeks.

nltk/lm/api.py
```diff
@@ -16,15 +16,37 @@
...
+
+    def repeat(value, count):
+        """An iterator with the specified value repeated for count time"""
+        for _ in range(count):
+            yield (value)
 
 
 @add_metaclass(ABCMeta)
-class Smoothing:
+class Smoothing(object):
```
iliakur: Actually, here's something you can already do. Here and at several other points in the changeset it looks like you were basing your work on an older version of the `develop` branch.

Could you rebase your branch on the latest develop and re-upload that please?

Let me know if you need help rebasing.

antran22: Actually, I patched this file on my local library first and just copied all of it into my fork. My branch is already based upon the latest version of `develop`.
If needed I can refractor the code a bit, but I don't think these differences are really breaking.

iliakur: Ah, I see.

It would be nice to remove all code that enables python 2 support, since NLTK is dropping support for that version and extra code is just that... extra.

I will comment on a few lines that I think are coming from an older version.

antran22: Refractored in 4192fd7

stevenbird:
@aldermann: thanks for this contribution - nice idea.
@Copper-Head: I assume you're planning more feedback, is that right? Thanks.

iliakur:
@stevenbird yes, it's in the todo list, I'll get to it sometime this week (including weekend)

stevenbird:
@Copper-Head : thanks, no rush

Copper-Head:
Alrighty, I took a look at this and did some thinking. @aldermann thanks for being proactive and wanting to contribute back to NLTK with such a thought-provoking PR.

Before we dig into the code, I have a couple of questions about how you use this method. I'm going to number them so that it's easy to refer to them later.

1. Are p

iliakur:
Sorry about the spam, github posted an unfinished comment.

@aldermann thanks for proactively contributing back to NLTK with such an interesting PR.

I looked at it and thought some. Before we dig into the particulars of the code, I have a more high-level question: could you share how you have been using this method and how you anticipate others using it going forward?
Feel free to even post code snippets (having adjusted them for privacy reasons, if you have to)!

I have several questions that I want to discuss, questions which will be expensive to revisit once we commit to a certain interface for this method.

1. What is the best return type for this method? Right now it's a list of `<word, score>` tuples, but perhaps an iterator over these would do, or even a dictionary-like object? Does it matter that the output is sorted? How is random sampling used?
1. Is the method ever called without a context?
1. What are some other names that we can try out for this method?
1. How much of these uses can be addressed with existing functionality, like Python's random module or NLTK's own probability module?

antran22:
@Copper-Head Thank you for these questions. This method actually came from the solution to a problem in one of my project, so I just thought it would be nice to have this method implemented directly in this package. The method is just a bit in its draft form, so I will happily take any suggestions to improve it.

> 1. What is the best return type for this method? Right now it's a list of `<word, score>` tuples, but perhaps an iterator over these would do, or even a dictionary-like object? Does it matter that the output is sorted? How is random sampling used?

Of course this method can be reworked to return an iterator. I just think it's natural to see the suggestions returned in a best to worst order, therefore the sorting and the returning of a whole list. If needed I can provide a boolean `return_sorted` parameter so the user can choose whether to get this list sorted or get a iterator of results.

> 2. Is the method ever called without a context?

I haven't really thought of that. I will make necessary change so that case works in the next commit.

> 3. What are some other names that we can try out for this method?

`predict` is also a nice idea, since the method works similarly to predictive input.

> 4. How much of these uses can be addressed with existing functionality, like Python's random module or NLTK's own probability module?

I'm sorry, can you elaborate more on this question?

iliakur:
@aldermann thanks for bearing with me on this.

I don't think I expressed myself clearly, got too carried away with details, I apologize. The overall point of my questions wasn't to ask you to fix stuff, but rather to brainstorm a little bit and explore different possibilities.

Let me do that now.

Imagine our method returns the simplest possible data-structure, which would just be `nltk.FreqDist` (that's what `lm.context_counts` returns). This object inherits from `collections.Counter` that has a method called `most_common` which returns&#x2026; a list of entries paired up with their counts and sorted from most common to the least. Moreover, you can pass it an argument to specify how many items to return. This sounds a lot like our most basic use-case :)

Even if you think it still makes more sense to have `suggest` (or `predict`) return an iterator/list, I would be in favour of taking advantage of as much existing functionality as possible to implement it under the hood.

Now to the trickier use case of sampling. The changes in this PR look a lot like we're rolling our own `random.choices`. Do we want to do this or should we once again take advantage of existing code? I'm in favour of the latter.

Bear in mind that before we merge this, we'll have to add thorough unit tests for this method, which try to test all of its arguments and all their combinations! So the less code and arguments we add, the fewer we'll have to test ;)

PR #2437, opened 2019-10-25
Correctly handle Penn Treebank trees with a unlabeled branching top node

rmalouf:
Currently, BracketParseCorpusReader removes the outermost set of unlabeled parens from a tree string before parsing.  This is to get rid of the unlabeled non-branching node at the top of most Penn Treebank trees.  But, it happens that the top node isn't always non-branching, and removing parens when it isn't creates an unparseable tree.  This PR changes it so that we parses the tree as is, and then remove the top node only if it turns out to be redundant.


stevenbird:
Thanks @rmalouf

Issue #2441, opened 2019-10-27
Wordnet synsets query problem 


ndvbd:
Doing:
`print wn.synsets("tablesssssssssssssssss")`
returns results (it shouldn't, as there is no such word in English),
But if we use the Wordnet online API here:
`http://wordnetweb.princeton.edu/perl/webwn`
for the same word, we get no results (correct).
Why is this problem?

alvations:
It's because the WordNet `wn.synsets()` functions tries to lemmatize using the `moprhy` lemmatizer before checking for synsets. Using the new API, you can get the strict behavior of not  lemmatizing by default by specifying `use_morphy=False` 

```python
>>> from wn import WordNet
>>> wn = WordNet()
wn.synsets("tablessssssssssssssss">>> wn.synsets("tablessssssssssssssss")
[Synset('table.n.01'), Synset('table.n.02'), Synset('table.n.03'), Synset('mesa.n.02'), Synset('table.n.05'), Synset('board.n.04'), Synset('postpone.v.01'), Synset('table.v.02')]
>>> wn.synsets("tablessssssssssssssss", use_morphy=False)
[]
```

ndvbd:
I upgraded from nltk 3.2.5 to 3.4.5 but still don't have the use_morphy argument. 
Why?

ekaf:
Yes @ndvbd,, the "use_morphy" argument is not even in the latest NLTK version, though was proposed in [issue 18](https://github.com/nltk/wordnet/pull/18#issue-513042920l), and sounds like a good idea...

ekaf:
Actually, the problem is not whether or not to use morphy, but rather to prevent morphy  from recursively stripping the same suffix many times. PR #3225 fixes it.

ekaf:
Fixed in #3225.

Issue #2421, opened 2019-09-26
Cannot get WordNet synsets for English without lemmatization


goodmami:
Querying wordnet with `wordnet.synsets()` will lemmatize the query word, but only for English. While this is useful for many applications, sometimes I do not want such lemmatization. For instance, I have dictionary forms for multiple languages (from, e.g., Swadesh lists) and I want to detect differences in polysemy between languages, but the lemmatization inflates the apparent polysemy for English. There appears to be no way (in the public API) to do an English query without lemmatization.

For example:

```python
>>> wn.synsets('eyeglasses')
[Synset('spectacles.n.01'), Synset('monocle.n.01')]
>>> wn.synsets('eyeglasses')[0].lemma_names()
['spectacles', 'specs', 'eyeglasses', 'glasses']
>>> wn.synsets('eyeglasses')[1].lemma_names()
['monocle', 'eyeglass']
```

The second synset (`monocle.n.01`) was found because 'eyeglass' appears in its lemmas, but not 'eyeglasses', which is only in the first synset. Sometimes specifying the POS can help, as with 'scissors' and 'scissor.v.01', but not always (as with 'eyeglasses' above, both are 'n'). I end up needing to write a wrapper like this:

```python
def synsets(lemma, pos=None, lang='eng', check_exceptions=True):
    results = wn.synsets(lemma,
                         pos=pos,
                         lang=lang,
                         check_exceptions=check_exceptions)
    if lang == 'eng':
        results = [ss for ss in results if lemma in ss.lemma_names()]
    return results
```

Am I missing something or is this currently the best way around the issue?

alvations:
@goodmami please take a look at the proposal on https://github.com/nltk/wordnet/pull/18

Soon the default NLTK WordNet API will be replaced by that standalone library =)

PR #2443, opened 2019-10-27
flag to control inclusion of wordnet synsets associated with stems

srhrshr:
@stevenbird, this is in response to your [proposal](https://github.com/nltk/wordnet/issues/17) to add a `stem` flag to getting wordnet synsets corresponding to a lemma. 

This seems to be an oft-requested feature [here](https://github.com/nltk/nltk/issues/2421) and [here](https://github.com/nltk/nltk/issues/2441). 

I've set the default to `stem=True` so as to ensure consistent behaviour for existing users of this functionality. 

Feedback appreciated! Thanks.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1560,12 +1560,14 @@ class WordNetCorpusReader(CorpusReader):
     # Retrieve synsets and lemmas.
     #############################################################
 
-    def synsets(self, lemma, pos=None, lang="eng", check_exceptions=True):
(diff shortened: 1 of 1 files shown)


alvations:
Hmmm, actually that's a different repo. Soon the default NLTK WordNet API will be replaced by that standalone library. =) 

Please take a look at the proposal on nltk/wordnet#18

srhrshr:
Haha I didn't realize there was a separate repo for `wordnet` - your resolution makes sense to me. Let me decline this!

PR #2446, opened 2019-10-30
Fix #2444 by having a special case for py 2.7

### LINKED ISSUE #2444: Py2.7 compatibility: subprocess has no attribute 'DEVNULL'
Hi,

A call to tag() when using nltk.tag.stanford.StanfordPOSTagger fails on python 2.7.

```
File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nltk/tag/stanford.py", line 116, in tag_sents
    cmd, classpath=self._stanford_jar, stdout=PIPE, stderr=PIPE

File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nltk/internals.py", line 112, in java
    subprocess_output_dict = {'pipe': subprocess.PIPE, 'stdout': subprocess.STDOUT, 'devnull': subprocess.DEVNULL}

AttributeError: 'module' object has no attribute 'DEVNULL'
```

This worked well on nltk 3.4 and it seems to have been introduced with [commit 0d3d086](https://github.com/nltk/nltk/commit/0d3d086a41f7bf4fb8418820b221e44a6a7e20ee#diff-6d3d9c7d1794a239ab676fc6fce3131f).

There has been an [SO question](https://stackoverflow.com/questions/57720341/stanford-part-of-speech-tagger-gives-attribute-error) about this. The accepted answer suggests to use corenlp instead.

This could however be a simple fix (see [this other SO answer](https://stackoverflow.com/questions/11269575/how-to-hide-output-of-subprocess-in-python-2-7) about a way to fix it). I understand py2.7 is fairly close to endlife, but as python 2.7 is still listed as a compatible version, I think it would be nice to fix this error.

Let me know if you want me to make a pull request.

Edit: for a bit more context, I was incentivized to bump the nltk version due to the vulnerability fixed in 3.4.5. However, going from 3.4 to 3.4.5 lead to this error. It might discourage some users to update their dependency to nltk.

sildar:
As Python 2.7 is still marked as supported by NLTK, version 3.4.5 shouldn't break compatibility.

This fixes #2444 which is one instance of regression that I was faced with. AFAIK, there are no other regressions.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -109,7 +109,11 @@ def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=Tru
     :raise OSError: If the java command returns a nonzero return code.
     """
 
-    subprocess_output_dict = {'pipe': subprocess.PIPE, 'stdout': subprocess.STDOUT, 'devnull': subprocess.DEVNULL}
+    try:
+        from subprocess import DEVNULL
+    except ImportError:
+        DEVNULL = open(os.devnull, 'wb')
+    subprocess_output_dict = {'pipe': subprocess.PIPE, 'stdout': subprocess.STDOUT, 'devnull': DEVNULL}
 
     stdin = subprocess_output_dict.get(stdin, stdin)
     stdout = subprocess_output_dict.get(stdout, stdout)


sildar:
Failed tests are identical to tag 3.4.5 https://travis-ci.org/nltk/nltk/jobs/574281125

stevenbird:
Thanks @sildar

sildar:
@stevenbird Will there be a new release including this fix?

PR #2453, opened 2019-11-03
Fix error in grammar.CFG.eliminate_start

jacobdweightman:
Previously, CFG.eliminate_start raised a TypeError whenever a new start symbol was actually required by the grammar. This change fixes that error.


stevenbird:
Thanks @jacobdweightman

PR #2461, opened 2019-11-18
Added more checks for detecting encoding of xml

alvations:
CI seems to be failing when reading xml files for scan for the first line to define the encoding, c.f. https://travis-ci.org/nltk/nltk/jobs/613326208#L2848 

As a naive patch, possibly we can add more fallbacks to check against the reading of XML.

Possibly there's a deeper problem with whether the default behavior when downloading the `nltk_data` is the unzip the files, esp. when Python 3.7 with 3rd party tools passes the CI but Python 3.7 w/o 3rd party tools doesn't.

PR #2462, opened 2019-11-18
Force travis to always redownload and reinstall, hopefully also re-unzipping

alvations:
@stevenbird The CI passes =)

Re-forcing the update on the CI cache, during the travis installs step, seems to resolve the problem. The travis cache should cleverly minimize the no. of times the "force" download is actually used.

LTGM, but I think this is still a temporary fix. 

We really need to:

 - Update/upgrade to a better alternative in handling `nltk_data`
 - Handling ZipFile pointers reading
 

C.f. #2461 and lots of other `nltk_data` related issues https://github.com/nltk/nltk/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+nltk_data =(

### DIFF
--- a/tools/travis/install.sh
+++ b/tools/travis/install.sh
@@ -7,4 +7,4 @@ pip install --upgrade -r pip-req.txt
 pip install --upgrade https://github.com/PyCQA/pylint/archive/master.zip
 
 #download nltk data packages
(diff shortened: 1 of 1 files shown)


alvations:
Once the full Travis CI checks out, we'll can merge this and then restart all other failed CI jobs on the other PR since 18 Oct.

PR #2466, opened 2019-11-30
make FreqDist iterable

Bonifacio2:
I was kinda surprised when I tried to iterate over a `FreqDist` and what it returned wasn't ordered. So I thought this change would be useful.

Thoughts?


stevenbird:
@Bonifacio2: nice idea, thanks

PR #2469, opened 2019-12-02
Bugfix 2356

anantvir:
Bug Fix for Issue 2356, changed probability.py, updated FreqDist.tabulate() to consider samples keyword, which was ignored earlier


stevenbird:
Thanks @anantvir. Sorry for the delay!

PR #2476, opened 2019-12-13
Arabic strip and normalization

MohammadModallal:


stevenbird:
@MohammadModallal I think you're proposing improvements to the ISRI algorithm. Please begin by raising an issue that describes the problems and your proposed solution. Note that you cannot add yourself as a co-author of their 2005 paper (https://dl.acm.org/doi/10.1109/ITCC.2005.90). Also, you cannot create new top-level modules in NLTK like this.

PR #2480, opened 2019-12-18
Fixed invalid target position when extracting phrases.

SamuelLarkin:
Using `nltk.translate.phrase_based.phrase_extraction()` I've noticed a bug.  The target indices are not correctly computed for certain target phrases.  That problem is even shown in the current example.
```
...
((2, 3), (5, 6), 'that', ', dass')
((2, 3), (5, 6), 'that', 'dass')
...
```
These two phrase pairs are valid but not their target indices.  Clearly, we cannot get phrase pairs for ((2,3), (5,6)) twice and more specifically, `((2, 3), (5, 6), 'that', ', dass')` has its target indices suggest a target phrase length of 1 token where as the target phrase itself is a 2 token phrase aka `, dass`.  The correct answer should be:
```
((2, 3), (4, 6), 'that', ', dass')
((2, 3), (5, 6), 'that', 'dass')
```

### DIFF
--- a/nltk/translate/phrase_based.py
+++ b/nltk/translate/phrase_based.py
@@ -41,11 +41,11 @@ def extract(
     :type f_start: int
     :param f_start: Starting index of the possible foreign language phrases
     :type f_end: int
-    :param f_end: Starting index of the possible foreign language phrases
+    :param f_end: End index of the possible foreign language phrases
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @SamuelLarkin

PR #2482, opened 2019-12-20
simplify expressions in snowball stemmers

noviluni:
Hi there!

I simplified some expressions to avoid calculating the length and slicing twice.

This is my first PR, if you need me to do something more just let me know :smile:


stevenbird:
Thanks @noviluni

PR #2485, opened 2020-01-07
Use html.escape instead of cgi.escape that was removed in Python 3.8.

### LINKED ISSUE #2481: treeprettyprinter imports removed function cgi.escape
Python 3.8 removed the `cgi.escape` function. Since this function is still used in the module `treeprettyprinter`, it is now unusable.

```python
Python 3.8.0 (default, Oct 23 2019, 18:51:26) 
[GCC 9.2.0] on linux
>>> from nltk import treeprettyprinter
  File "/usr/lib/python3.8/site-packages/nltk/treeprettyprinter.py", line 24, in <module>
    from cgi import escape
ImportError: cannot import name 'escape' from 'cgi' (/usr/lib/python3.8/cgi.py)
```

The Python 3.7 documentation states:

> Deprecated since version 3.2: This function is unsafe because quote is false by default, and therefore deprecated. Use html.escape() instead.

tirkarthi:
Use `html.escape` instead of `cgi.escape` that was removed in Python 3.8. Since Python 3.5+ is only supported by nltk we can use `html.escape` always. Let me know if this needs to be changed.

Fixes #2481

### DIFF
--- a/nltk/treeprettyprinter.py
+++ b/nltk/treeprettyprinter.py
@@ -19,7 +19,10 @@ http://jgaa.info/accepted/2006/EschbachGuentherBecker2006.10.2.pdf
 """
 
 import re
-from cgi import escape
+try:
+    from html import escape
+except ImportError:
+    from cgi import escape
 from collections import defaultdict
 from operator import itemgetter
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @tirkarthi

PR #2486, opened 2020-01-10
`is` operator is exclusive to None

ramazanpolat:
Old code causes this warning:
SyntaxWarning: "is not" with a literal. Did you mean `!=` ?
  if ret is not 0:


stevenbird:
Thanks @ramazanpolat

PR #2487, opened 2020-01-12
Use time.perf_counter instead of time.clock for python 3.8 compatibility

### LINKED ISSUE #2484: time.clock removal in Python 3.8
`time.clock` was removed in Python 3.8 in https://github.com/python/cpython/pull/13270 . This will cause error in Python 3.8 for https://github.com/nltk/nltk/search?q=time.clock&unscoped_q=time.clock

tirkarthi:
`time.clock` was deprecated in Python 3.4 and removed in Python 3.8 . Use `time.perf_counter` on Python 3 and use `time.clock` under Python 2 for compatibility.

Fixes #2484

### DIFF
--- a/nltk/compat.py
+++ b/nltk/compat.py
@@ -28,6 +28,7 @@ if PY3:
     StringIO = io.StringIO
     BytesIO = io.BytesIO
 
+    from time import perf_counter as time_clock
     from datetime import timezone
 
     UTC = timezone.utc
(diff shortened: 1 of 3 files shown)


tirkarthi:
Verified the patch with below commands where time is printed properly and throws attribute error on develop branch for `time.clock` with Python 3.8 .

```
$ python -m nltk.parse.earleychart
$ python -m nltk.parse.featurechart
```

stevenbird:
Thanks @tirkarthi !

PR #2493, opened 2020-01-30
Fix escape sequences

ab-10:
Pull request in response to #2378


stevenbird:
Thanks @ab-10, this is a welcome contribution. Unfortunately it breaks a lot of the doctests. Would you like to have a look at the CI logs and see if you can work out what's going wrong?

ab-10:
@stevenbird I think I took the wrong approach of rewriting all escape sequences without testing along the way. I will make a more responsible escape sequence fix when I have a little bit more time.

Issue #2495, opened 2020-02-04
Redundant operations on ngram language models cause high performance impact


palasso:
### General Info
**Module:** `nltk.lm`
**Type:** Performance
**Impact:** High
### Issue
Most ngram language models perform redundant operations causing a high performance impact
#### Language models affected
`Lidstone`, `Laplace`, `KneserNeyInterpolated`, `WittenBellInterpolated`
#### Root cause
Re-calculating `len(self.vocab)` and `self.counts[len(context) + 1].N()` instead of calculating them only once
#### Performance impact
100-1000x speed difference
### Background
While I was developing a noisy channel model I used NLTK's language models for the noisy channel model's language model. I observed a big difference in performance between `MLE` and the other language models provided by NLTK. After examining the code more thoroughly I identified the root cause of that difference to be some re-calculations being done unecessary every time the language model would calculate the probability of a ngram. I made a minimal set of changes to resolve the performance issue and observed a 100-1000x times increase in performance after I run some [tests](https://github.com/palasso/nltk/blob/9a3fa65cdf3b7175e2cda1eb704b7d57de1c2003/nltk.ipynb).
### Update
I updated the [tests](https://github.com/palasso/nltk/blob/30617c100ff07a889491fdb4ab2725359f20d1a8/nltk.ipynb) to include a different implementation that modifies only `Vocabulary` (based on discussion in PR #2496 ). This new implementation has the benefit of changing smaller parts of the `lm` module with the drawback that it doesn't resolve performance issues for `WittenBellInterpolated`.

Higgs32584:
should this issue be closed?

palasso:
https://github.com/nltk/nltk/pull/2555 fixed the cases for `Lidstone` and `KneserNeyInterpolated` but there is also `Laplace` and `WittenBellInterpolated`.

PR #2496, opened 2020-02-04
Increase performance of language models

palasso:
A minimal set of changes to resolve the performance issue I observed as explained on issue #2495.
It is expected to lead to a 100-1000x speed increase.
Highlighting relevant developers: @stevenbird @iliakur

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -261,6 +261,7 @@
 - Alexandre H. T. Dias <https://github.com/alexandredias3d>
 - Jacob Weightman <https://github.com/jacobdweightman>
 - Bonifacio de Oliveira <https://github.com/Bonifacio2>
+- Vassilis Palassopoulos <https://github.com/palasso>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer
--- a/nltk/lm/api.py
+++ b/nltk/lm/api.py
@@ -28,7 +28,7 @@ class Smoothing:
     work both with Backoff and Interpolation.
     """
 
-    def __init__(self, vocabulary, counter):
+    def __init__(self, vocabulary, counter, cache):
         """
         :param vocabulary: The Ngram vocabulary object.
         :type vocabulary: nltk.lm.vocab.Vocabulary
@@ -37,6 +37,7 @@ class Smoothing:
         """
         self.vocab = vocabulary
         self.counts = counter
+        self.cache = cache
 
     @abstractmethod
     def unigram_score(self, word):
@@ -80,6 +81,15 @@ class LanguageModel:
     Cannot be directly instantiated itself.
 
     """
+    
+    class Cache:
+    
+        def __init__(self, order, vocab, counts):
+            self.update(order, vocab, counts)
+            
+        def update(self, order, vocab, counts):
+            self.vocab_len = len(vocab)
+            self.countsN = {n: counts[n].N() for n in range(1, order+1)}
 
     def __init__(self, order, vocabulary=None, counter=None):
         """Creates new LanguageModel.
@@ -99,6 +109,7 @@ class LanguageModel:
         self.order = order
         self.vocab = Vocabulary() if vocabulary is None else vocabulary
         self.counts = NgramCounter() if counter is None else counter
+        self.cache = self.Cache(self.order, self.vocab, self.counts)
 
     def fit(self, text, vocabulary_text=None):
         """Trains the model on a text.
@@ -113,6 +124,7 @@ class LanguageModel:
                 )
             self.vocab.update(vocabulary_text)
         self.counts.update(self.vocab.lookup(sent) for sent in text)
+        self.cache.update(self.order, self.vocab, self.counts)
 
     def score(self, word, context=None):
         """Masks out of vocab (OOV) words and computes their model score.
--- a/nltk/lm/models.py
+++ b/nltk/lm/models.py
@@ -46,7 +46,7 @@ class Lidstone(LanguageModel):
         counts = self.context_counts(context)
         word_count = counts[word]
         norm_count = counts.N()
-        return (word_count + self.gamma) / (norm_count + len(self.vocab) * self.gamma)
+        return (word_count + self.gamma) / (norm_count + self.cache.vocab_len * self.gamma)
 
 
 class Laplace(Lidstone):
@@ -70,7 +70,7 @@ class InterpolatedLanguageModel(LanguageModel):
         assert issubclass(smoothing_cls, Smoothing)
         params = kwargs.pop("params", {})
         super().__init__(order, **kwargs)
-        self.estimator = smoothing_cls(self.vocab, self.counts, **params)
+        self.estimator = smoothing_cls(self.vocab, self.counts, self.cache, **params)
 
     def unmasked_score(self, word, context=None):
         if not context:
--- a/nltk/lm/smoothing.py
+++ b/nltk/lm/smoothing.py
@@ -20,8 +20,8 @@ def _count_non_zero_vals(dictionary):
 class WittenBell(Smoothing):
     """Witten-Bell smoothing."""
 
-    def __init__(self, vocabulary, counter, **kwargs):
-        super().__init__(vocabulary, counter, **kwargs)
+    def __init__(self, vocabulary, counter, cache, **kwargs):
+        super().__init__(vocabulary, counter, cache, **kwargs)
 
     def alpha_gamma(self, word, context):
         alpha = self.counts[context].freq(word)
@@ -30,7 +30,7 @@ class WittenBell(Smoothing):
 
     def _gamma(self, context):
         n_plus = _count_non_zero_vals(self.counts[context])
-        return n_plus / (n_plus + self.counts[len(context) + 1].N())
+        return n_plus / (n_plus + self.cache.countsN.get(len(context) + 1, 0))
 
     def unigram_score(self, word):
         return self.counts.unigrams.freq(word)
@@ -39,12 +39,12 @@ class WittenBell(Smoothing):
 class KneserNey(Smoothing):
     """Kneser-Ney Smoothing."""
 
-    def __init__(self, vocabulary, counter, discount=0.1, **kwargs):
-        super().__init__(vocabulary, counter, **kwargs)
+    def __init__(self, vocabulary, counter, cache, discount=0.1, **kwargs):
+        super().__init__(vocabulary, counter, cache, **kwargs)
         self.discount = discount
 
     def unigram_score(self, word):
-        return 1.0 / len(self.vocab)
+        return 1.0 / self.cache.vocab_len
 
     def alpha_gamma(self, word, context):
         prefix_counts = self.counts[context]


iliakur:
@palasso I'm trying to wrap up my thesis by the end of this month, so I'll take a look at this sometime in the first two weeks of March. I hope to use this excuse to experiment with adding some timing benchmarks to the automatic tests. If you're feeling ambitious, feel free to do so as well!

My initial impression is that it's a decent optimization, primarily because it is simple :)

iliakur:
Oh, I just noticed that you are indeed running tests. When you have them in good shape, would be wonderful to get them added to NLTK!

stevenbird:
[CI:retest]

iliakur:
@palasso I've submitted my thesis, but adjusting to work life (and COVID-19-related issues) is taking longer than anticipated.

I'd expect to get to this within the next 2-3 weeks.

Thanks for your patience!

palasso:
Thank you @iliakur. I don't know the release schedule of NLTK but I hope it makes in time for the next release of 3.5.

iliakur:
Hi gang, I finally got a moment to look at this properly.

@palasso thanks again for sending it in and being so patient :)

I have a few questions/concerns:

This seems to enable caching by default all the time. Disabling the cache seems a bit hairy, definitely not something to expect from a regular user. As we all know, caches are nice, but occasionally can lead to annoying behaviour (invalidating them is a pain, for instance). I wonder if we should consider providing an easy way to disable the cache.

Looking at the concrete way you implemented the cache, I noticed you had to touch literally all the main files in the module: `api`, `smoothing`, `models`. I suspect this won't be as easy to maintain as keeping all cache-related logic in one place.

With this in mind, what do you think of this idea (simplified and abridged to include only relevant parts)?

```python
# api.py

class LanguageModel:
    def __init__(self, order, vocabulary=None, counter=None, cache=True):
        # we enable caching by default, but it's easy to turn off
        self.order = order
        self.vocab = Vocabulary() if vocabulary is None else vocabulary
        self.counts = NgramCounter() if counter is None else counter
        if cache:
            self.score = cached(self.score)
```
Where `cached` is some standard implementation of a cache that just remembers its arguments. Here's a rough draft:
```python
from functools imports wraps

def cached(score_func):
     cache = {}

    @wraps(score_func)
    def wrapper(word, context=None):
        k = (word, context)
        if k in cache:
            score = cache[k]
        else:
            score = cache[k] = score_func(word, context)
        return score
    return wrapper
```

Another, perhaps even simpler idea would be to include the caching logic inside of the `LanguageModel.score` method itself. In any case we are keeping the caching localized to only one place.

I noticed these proposals don't cache the `len(vocab)` call, but I'm honestly not expecting a big slowdown from that. Would be open to empirically verifying this though.

alvations:
BTW, this is a good use case for this fun feature https://docs.python.org/3/library/functools.html#functools.lru_cache =)

And example is how it's applied in Levenshtein distance. https://gist.github.com/alvations/989dacca76c820c48ef18c4637515b39#file-sent_ter-py-L4

iliakur:
Is LRU a good caching scheme for language models?

This does bring up another idea though: what if the `cache` argument to `__init__` were to accept not only a simple boolean, but either `None` or a caching function? This would let the user roll their own caching mechanism if they want to or use something that exists like `lru_cache`. In any case, the user has options and the freedom to tweak things to their particular needs. Kudos @alvations :)

palasso:
> I noticed these proposals don't cache the len(vocab) call, but I'm honestly not expecting a big slowdown from that. Would be open to empirically verifying this though.

From the testing I have performed the `len(vocab)` call costs the most in terms of speed while its result usually doesn't change.

iliakur:
Oh, I'm surprised by that. I would personally focus our optimization efforts on it then. This week I'll sit down and try timing and tweaking it a bit. If you have numbers examples that target explicitly that call (I checked the linked notebook but couldn't find any there), more information is welcome :)

alvations:
Note that unless the instance value changes, the `len(x)` function might not actually be recounting all the time in Python https://stackoverflow.com/questions/29057153/does-the-len-built-in-function-iterates-through-the-collection-to-calculate-it

iliakur:
Hmm, I suspect this doesn't really apply to classes that implement their own `__len__` which involves a comprehension the way `Vocabulary` currently does. This is probably what needs to be tweaked.

iliakur:
**EDIT: fixed a small bug in init method**

@palasso I can confirm that the running time of `len(vocab)` increases linearly with the size of the vocabulary.

I got it to be constant with the following tweaks:

```python
class Vocabulary:

    def __init__(
            self, 
            counts=None, 
            unk_cutoff=1, 
            unk_label="<UNK>"
        ):
        self.unk_label = unk_label
        if unk_cutoff < 1:
            raise ValueError(
                "Cutoff value cannot be less than 1. Got: {0}".format(unk_cutoff)
            )
        self._cutoff = unk_cutoff

        self.counts = Counter()
        self.update(counts if counts is not None else "")

    def update(self, *counter_args, **counter_kwargs):
        self.counts.update(*counter_args, **counter_kwargs)
        self._len = sum(1 for _ in self)

    def __len__(self):
        """Computing size of vocabulary reflects the cutoff."""
        return self._len
```

Would you be interested in submitting this either in this PR or as a new one? It would still need to be tested, and it would be cool if we could try plugging it into your benchmark suite.

Let me know if you don't think you'll have the resources for it, I have some time on my hands and this kind of tinkering is fun :)

palasso:
I run the benchmark again with the modification of `Vocabulary` and without.

```
Original Vocabulary
trigrams:
{'LID': 25.451935806000165,
 'LIDM': 0.025555172999702336,
 'KNI': 27.028115159000208,
 'KNIM': 0.24075353199987148,
 'WBI': 86.52406588799977,
 'WBIM': 0.2683036789999278}

fivegrams:
{'LID': 30.00634654300029,
 'LIDM': 0.03528844099992057,
 'KNI': 29.600673633000042,
 'KNIM': 0.3856433239998296,
 'WBI': 694.1382620029999,
 'WBIM': 0.46601980600007664}


New Vocabulary
trigrams:
{'LID': 0.08882101800008968,
 'LIDM': 0.043921537000642275,
 'KNI': 0.26541022899982636,
 'KNIM': 0.25694879999991826,
 'WBI': 89.90805916800036,
 'WBIM': 0.2864626069995211}

fivegrams:
{'LID': 0.04899713899976632,
 'LIDM': 0.044580556000255456,
 'KNI': 0.32740841299983003,
 'KNIM': 0.3006070129995351,
 'WBI': 686.0069909860003,
 'WBIM': 0.39238316000046325}
```

As can be observed, caching the vocabulary length within the `Vocabulary` class, results in having inference time of the same order with or without my modifications for:
* Lidstone Smoothing
* Kneser-Ney Smoothing with Interpolation

That is not the case with Witten-Bell Smoothing with Interpolation. This is because of the `counts[n].N()` operation which costs a lot and can also be cached. Do you think this should also be cached in `NgramCounter`? In that case we could remove the `Cache` I made and just keep cache in `Vocabulary` and `NgramCounter` so the code should be cleaner. We could verify the effectiveness of this solution by re-running a new benchmark.

iliakur:
Ah, this becomes tricky. `counts[n].N()` calls the `N` method of the `FreqDist` and `ConditionalFreqDist` classes. `FreqDist` caches that, but `ConditionalFreqDist` doesn't. I suggest we tackle this in a separate pull request because I don't feel confident messing with that code until we've added some tests for it (i.e. more than it currently has). That makes it a bit of a project on its own and expands the scope of our operation significantly.

If you really cannot tolerate a slow WBI for your system and require that functionality ASAP, feel free to keep the caching mechanism in your fork. When it comes to this PR, I'd consider speeding up the `Vocabulary` a very worthy contribution for the time being.

I think we should focus on solidifying the progress we've made. This mostly has to do with tests. I think the best way to both demonstrate and confirm the `len` speedup is by showing that it runs in constant time no matter the size of the vocabulary. To this end we can create an "obviously small" (<10 items?) instance and an "obviously large" (a whole corpus?) one and show that calculating the `len` of both takes *the same amount of time*. This has the advantage of being very portable: no matter how powerful or weak your processor, constant time operations should take, well, *constant time* :)

iliakur:
@palasso anything else I can do to help move this forward?

palasso:
@iliakur Thank you for your help, your suggestions and the code changes. Next weekend I'll have more time and I will rework the PR and the benchmarks as well to test all the combinations. I am also thinking to include training time just to make sure there were no regressions there.

palasso:
I updated the branch with the benchmarks in order to test the performance of the various implementations and on whether there were any regressions.

I did not test Witten-Bell Smoothing with Interpolation because we already know that the fix in Vocabulary does not improve it and it would take much more time to benchmark.

```
LID: Lidstone-smoothing
KNI: Kneser-Ney Smoothing with Interpolation
M: Modified language model implementation
V: Modified Vocabulary implementation

trigram

training time
LID: 102.82990377700116
LIDM: 105.46359531199778
LIDMV: 106.86554181099928
LIDV: 107.51525854600186
KNI: 104.75891448299808
KNIM: 107.13869005499873
KNIMV: 110.7783844379992
KNIV: 109.84420230100295

inference time
LID: 285.03596957300033
LIDM: 0.27742780200060224
LIDMV: 0.27167368099981104
LIDV: 0.2825343350014009
KNI: 297.60491606600044
KNIM: 2.921886004998669
KNIMV: 3.961752424998849
KNIV: 4.490208731000166

fivegram

training time
LID: 249.4098840680017
LIDM: 250.5333112490007
LIDMV: 249.76111601699813
LIDV: 244.03447735400187
KNI: 399.9116396120007
KNIM: 507.21733781599687
KNIMV: 266.91502224399665
KNIV: 249.80434543500087

inference time
LID: 458.04469004699786
LIDM: 0.6001836180003011
LIDMV: 0.5922620089986594
LIDV: 0.6049938080032007
KNI: 519.8177143770008
KNIM: 6.205774944999575
KNIMV: 4.547874755997327
KNIV: 4.407776099997136
```

We can observe that the change on `Vocabulary` is enough to significantly improve inference time for both Lidstone Smoothing and Kneser-Ney Smoothing with Interpolation without any regression in training time.

Additionally, I run the same `nltk.ipynb` file again by replacing the two occurrences of `english.sents()` with `english.sents()[:1000]` (out of 19899 sentences), to test on whether the inference time is constant compared to vocabulary size.

```
LID: 53.094840910998755
LIDM: 0.29920799500177964
LIDMV: 0.2889021379996848
LIDV: 0.30506811599843786
KNI: 53.458264933000464
KNIM: 1.0204762119974475
KNIMV: 1.0211700770014431
KNIV: 1.0307058569997025

LID: 53.63951414499752
LIDM: 0.3840107810028712
LIDMV: 0.3840423500005272
LIDV: 0.38946822899742983
KNI: 56.61550912999883
KNIM: 1.5345575059982366
KNIMV: 1.6229681520017039
KNIV: 1.5848969139988185
```

Inference time for Lidstone smoothing seems to be close to constant compared to the size of the vocabulary. For Kneser-Ney Smoothing with Interpolation it seems that it was 3 times faster with vocabulary size of 1k (as opposed to 20k).

palasso:
Since we're going towards the direction of changing only `Vocabulary` I closed this PR and made a new PR #2555

PR #2502, opened 2020-02-23
FreqDist ordering based on partial order

### LINKED ISSUE #1457: Axe FreqDist ordering/comparisons?
This [StackOverflow question](http://stackoverflow.com/q/38215686/4501212) made me aware that the `FreqDist` class implements ordering comparisons. As the question shows, the implementation is buggy so we should do something about it. 
I propose we get rid of it altogether, considering that its parent class `Counter` doesn't support comparisons and that it's not entirely clear what it means for one frequency distribution to be "greater than" another.

Thoughts?

pyfisch:
Change the implementation of <= and >= for FreqDist
so they are both a partial order.

With the old implementation for a pair of two frequency
distributions each could be bigger than the other.

```
>>> a = FreqDist('aab')
>>> b = FreqDist('abb')
>>> a > b
True
>>> b > a
True
```

This is very unintuitive as it ignores both common sense
and the mathematical definition of order.

It should be noted that the new ordering scheme is a not
a total order and that both a <= b and b <= a are False.

Closes #1457

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -389,16 +389,47 @@ class FreqDist(Counter):
         return self.__class__(super(FreqDist, self).__and__(other))
 
     def __le__(self, other):
+        """
+        Returns True if this frequency distribution is a subset of the other
+        and for no key the value exceeds the value of the same key from
+        the other frequency distribution.
+
+        The <= operator forms partial order and satisfying the axioms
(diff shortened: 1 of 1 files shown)


stevenbird:
Agreed, like inclusion for plain sets, this should be a partial order.

Issue #2296, opened 2019-05-10
Dropping Python 2.7 Support


alvations:
From https://travis-ci.org/nltk/nltk/jobs/530566954, some of the dependencies NLTK is dependent on no longer supports Python 2.7. I think it's good timing to also drop support for Python 2.7 so that our CI continues to work and move the library forward.

stevenbird:
I see, ok, it's time to move on!

alvations:
Starting from these `__future__` =) 

```python
~/git-stuff/my-contrib/nltk$ grep "__future__" nltk/**/*
nltk/app/chartparser_app.py:from __future__ import division
nltk/app/chunkparser_app.py:from __future__ import division
nltk/app/collocations_app.py:from __future__ import division
nltk/app/rdparser_app.py:from __future__ import division
nltk/app/wordnet_app.py:from __future__ import print_function
nltk/ccg/api.py:from __future__ import unicode_literals
nltk/ccg/chart.py:from __future__ import print_function, division, unicode_literals
nltk/ccg/combinator.py:from __future__ import unicode_literals
nltk/ccg/lexicon.py:from __future__ import unicode_literals
nltk/chat/__init__.py:from __future__ import print_function
nltk/chat/eliza.py:from __future__ import print_function
nltk/chat/iesha.py:from __future__ import print_function
nltk/chat/rude.py:from __future__ import print_function
nltk/chat/suntsu.py:from __future__ import print_function
nltk/chat/util.py:from __future__ import print_function
nltk/chat/zen.py:from __future__ import print_function
nltk/chunk/named_entity.py:from __future__ import print_function
nltk/chunk/named_entity.py:from __future__ import unicode_literals
nltk/chunk/regexp.py:from __future__ import print_function, unicode_literals
nltk/chunk/regexp.py:from __future__ import division
nltk/chunk/util.py:from __future__ import print_function, unicode_literals, division
nltk/classify/decisiontree.py:from __future__ import print_function, unicode_literals, division
nltk/classify/maxent.py:from __future__ import print_function, unicode_literals
nltk/classify/megam.py:from __future__ import print_function
nltk/classify/naivebayes.py:from __future__ import print_function, unicode_literals
nltk/classify/rte_classify.py:from __future__ import print_function
nltk/classify/scikitlearn.py:from __future__ import print_function, unicode_literals
nltk/classify/senna.py:    >>> from __future__ import unicode_literals
nltk/classify/senna.py:from __future__ import unicode_literals
[...]

PR #2503, opened 2020-02-23
Remove Python 2 compatibility layer

pyfisch:
* Remove all `__future__` imports
* Remove dependency on `six`

See also #2296

There are still remaining changes like the `nltk/compat.py` file and various version-dependent branches. Additionally the documentation needs an update.


stevenbird:
Thanks @pyfisch - great contribution!

PR #2504, opened 2020-02-24
Remove more Python 2 artifacts

pyfisch:
This removes most code from the nltk/compat.py file including the Unicode compatibility handling.
This means if external code called e. g. `FreqDist().unicode_repr()` it needs to be changed to `repr(FreqDist())`. 

Additionally I removed most names from the compat.py file but as they did not appear in the documentation I doubt external code relies on them.

If you are concerned about maybe breaking some users code we can discuss which names I still need to export for retaining compatibility.


stevenbird:
Thanks @pyfisch. I'm happy for people using internal names to update their code.

alvations:
Thanks @pyfisch!! This is so awesome that we can remove `compat.py` and all the try-excepts !!!

PR #2506, opened 2020-02-24
Remove u'spam' strings from doctests

pyfisch:
Unicode literals in doctests were needed for Python 2 compat.
Remove DoctestFix nose plugin which dealt with these literals.


stevenbird:
Thanks!

PR #2508, opened 2020-02-25
Remove obsolete compatibility code

pyfisch:


pyfisch:
BTW I added Python 3.8 to the list of supported versions in `setup.py`.

There is a similar list of attributes in `nltk/init.py` but they appear to be very outdated. For example `__classifiers__` does only mention Python 2.6 and Python 2.7 as supported versions.

stevenbird:
Thanks @pyfisch!

PR #2510, opened 2020-02-25
Check code formatting on CI

pyfisch:
After running the tests in Travis check with black if the code is correctly formatted.
Formatted all the code.

#2163 formatted all the files in nltk with `black`. Since this PR there were many changes to nltk and some used a different formatting, so there are now 45 files again that have a different formatting than expected by black.

In the Go and Rust communities it is very common to include formatting as part of the CI pipeline and to only merge commits after the were formatted either with gofmt or rustfmt. I really like this because I need to think very little about formatting anymore.

If you don't want to enforce style in CI just let me know.


stevenbird:
Thanks @pyfisch. I'm happy for this to be part of our CI. What do you say @alvations?

alvations:
Black would be good to be part of the CI! 

Lets try and see whether this is right way to do it on the pre-commits: https://github.com/nltk/nltk/pull/2511   =)

PR #2511, opened 2020-02-26
[WIP] Added pre-commit flake8 and black

alvations:
I'm not sure whether I've done it correctly. 

Locally `black` runs when I do `git commit` but would it also run before the PR is merged? 

@pyfisch any ideas?


iliakur:
What was wrong with `black`'s default line length? 120 is definitely too long for me, at that point we're in Java-land imo.

alvations:
I would suggest 120 mainly because of the unittests with long long string checks. Also, because of the the initialization in LazyCorpusReader is kind of long too. 

Any particular length we should follow? default black's 88 or 89?

pyfisch:
Hi @alvations,

having black run as a pre-commit hook is a good idea. But I don't think it is necessary to depend on the pre-commit framework for this. It requires additional configuration in YAML and apparently includes an installer/package manager to download the correct hooks.

To execute the pre-commit-framework on Travis (and not only locally) you need to add it to the scripts section of .travis.yml.

Two notes on the formatting:

* NLTK was formatted a while back with 88 characters per line (the black default value). I personally don't see a reason to change this and I don't like longer lines as they don't fit the laptop screen.
* By default black replaces all single quoted strings with double quotes. With the --skip-string-normalization (or -S for short) option the strings are left unchanged. This reduces the diff size from around 150 files to about 45 files, if the line length is not changed.

iliakur:
Personally I'm comfortable with up to 100, so black's default is good.
Raymond Hettinger famously said "90ish" is a reasonable limit and I think he's a pretty good authority on clarity and readability. @pyfisch also brought up laptop screens (a good point), but even external desktop screens can be too small for 120 if used in vertical position (my preferred setup, easier to scan long code files without scrolling).

Long strings should be easy to break up, I could take that on sometime this month. I don't know about LazyCorpusReader, but I somehow doubt there's anything there that can't be solved with some creative use of parentheses.

pyfisch:
One can also add black to tox: https://opensource.com/article/19/5/python-tox

> Long strings should be easy to break up, I could take that on sometime this month. 

I don't think this is needed to use black as it does not complain about any files it can't reformat.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,8 @@
+- id: black
```
pyfisch, line 1: https://black.readthedocs.io/en/stable/version_control_integration.html

The black docs recommend storing the configuration in a pyproject.toml to ensure editors and direct calls to black behave consistently.

iliakur: Can't +1 this enough. Some folks prefer to run these formatting commands manually from their CLI in addition to all the hooks and whatnot. Consistency across these two invocation styles is important!

I've run into problems with this at work.

alvations:
We'll come back to this later.

PR #2512, opened 2020-02-26
Test black formatting

pyfisch:
Run black with default formatting (except for --skip-string-normalization) and only on Python version 3.8 as running formatting on multiple versions is unnecessary.


iliakur:
Did I understand correctly that string normalization was skipped only to keep the diff small(er)?

Does this mean that at some point we'll normalize the strings too? If yes, would it make sense to wrap that into the big formatting push and just be done with the whole topic instead of spreading it over multiple PRs?

pyfisch:
> Did I understand correctly that string normalization was skipped only to keep the diff small(er)?

Yes mostly. A previous PR that formatted all of nltk made the same choice so I did not want format those files again.

PR #2514, opened 2020-03-02
fix porter stemmer failed to return lowercase

### LINKED ISSUE #2507: Porter stemmer returns a capital instead of lowercase
This output is unexpected. The `In` returns the capitalize `In` from PorterStemmer's output.

```python
>>> from nltk.stem import PorterStemmer
>>> porter = PorterStemmer()
>>> porter.stem('In')
'In'
```

More details on https://stackoverflow.com/q/60387288/610569

PhanatosZou:
Fix #2507

### DIFF
--- a/nltk/stem/porter.py
+++ b/nltk/stem/porter.py
@@ -652,13 +652,16 @@ class PorterStemmer(StemmerI):
         stem = word.lower()
 
         if self.mode == self.NLTK_EXTENSIONS and word in self.pool:
-            return self.pool[word]
+            return self.pool[stem]
 
         if self.mode != self.ORIGINAL_ALGORITHM and len(word) <= 2:
             # With this line, strings of length 1 or 2 don't go through
             # the stemming process, although no mention is made of this
             # in the published algorithm.
-            return word
+            if word.isupper()==True:
+                return word
+            else:
+                return stem
 
         stem = self._step1a(stem)
         stem = self._step1b(stem)
--- a/nltk/stem/porter.py
+++ b/nltk/stem/porter.py
@@ -648,8 +648,11 @@ class PorterStemmer(StemmerI):
             word, [("ll", "l", lambda stem: self._measure(word[:-1]) > 1)]
         )
 
-    def stem(self, word):
-        stem = word.lower()
+    def stem(self, word, to_lowercase=True):
+        """
+        :param to_lowercase: if `to_lowercase=True` the word always lowercase
+        """
+        stem = word.lower() if to_lowercase else word
 
         if self.mode == self.NLTK_EXTENSIONS and word in self.pool:
             return self.pool[stem]
@@ -658,10 +661,8 @@ class PorterStemmer(StemmerI):
             # With this line, strings of length 1 or 2 don't go through
             # the stemming process, although no mention is made of this
             # in the published algorithm.
-            if word.isupper()==True:
-                return word
-            else:
-                return stem
+            return word if word.isupper() and to_lowercase else stem
+
 
         stem = self._step1a(stem)
         stem = self._step1b(stem)
--- a/nltk/stem/porter.py
+++ b/nltk/stem/porter.py
@@ -661,7 +661,7 @@ class PorterStemmer(StemmerI):
             # With this line, strings of length 1 or 2 don't go through
             # the stemming process, although no mention is made of this
             # in the published algorithm.
-            return word if word.isupper() and to_lowercase else stem
+            return stem if to_lowercase else word
 
 
         stem = self._step1a(stem)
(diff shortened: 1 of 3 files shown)


nltk/stem/porter.py
```diff
@@ -652,13 +652,16 @@ def stem(self, word):
...
 
         if self.mode != self.ORIGINAL_ALGORITHM and len(word) <= 2:
             # With this line, strings of length 1 or 2 don't go through
             # the stemming process, although no mention is made of this
             # in the published algorithm.
-            return word
+            if word.isupper()==True:
+                return word
+            else:
+                return stem
```
alvations: Perhaps the ternary operator here is appropriate:

```
return word if word.isupper() else stem
```

But I think it'll be best to assume that all stems are lowecased by default, so 

```python
return word if word.isupper() and to_lowercase else stem
```

Also, when checking for truthiness of string, `if word.isupper()` should be sufficient without the explicity `==True` =)

nltk/stem/porter.py
```diff
@@ -652,13 +652,16 @@ def stem(self, word):
         stem = word.lower()
```
alvations: Add a default behavior to always lowercase and allow user to undo the lowercasing if deemed necessary.

```python
def stem(self, word, to_lowercase=True):
    stem = word.lower() if to_lowercase else word 
```

alvations:
It'll be good to add certain test cases here https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_stem.py#L145 

```python 
    def test_lowercase_option(self):
        """Test for improvement on https://github.com/nltk/nltk/issues/2507
        Ensures that stems are lowercased when `to_lowercase=True`
        """
        porter = PorterStemmer()
        assert porter.stem('On') == 'on'
        assert porter.stem('I') == 'i'
        assert porter.stem('I', to_lowercase=False) == 'I'
        assert porter.stem('GitHub') == 'github'
        assert porter.stem('GitHub', to_lowercase=False) == 'GitHub'
```

iliakur:
What about strings like this: "GitHub"? What's the expected behaviour there?

alvations:
Possibly the same manner, lowercase by default, keep case if `to_lowercase=False`?

nltk/stem/porter.py
PhanatosZou: Sure. I'll fix it.

iliakur:
Would be nice to have test(s) for that, just to be explicit about what's happening :)

PhanatosZou:
Hi @alvations sorry for the long delay. I'm kind of new here. My last PR didn't pass the test and I'm kind of stuck here. Is there any suggestions to fix the issue? Thank you in advance!

nltk/stem/porter.py
```diff
@@ -648,17 +648,21 @@ def _step5b(self, word):
...
         if self.mode == self.NLTK_EXTENSIONS and word in self.pool:
-            return self.pool[word]
+            return self.pool[stem]
 
         if self.mode != self.ORIGINAL_ALGORITHM and len(word) <= 2:
             # With this line, strings of length 1 or 2 don't go through
             # the stemming process, although no mention is made of this
             # in the published algorithm.
-            return word
+            return word if word.isupper() and to_lowercase else stem
```
dannysepler, line 664:
```suggestion
            return stem
```

hi @PhanatosZou i think the broken test is because you're returning `stem` here :)

dannysepler:
```suggestion
            return stem if to_lowercase else word
```

hi @PhanatosZou i _think_ you're just backwards is all :) in fact, it's probably simpler to simply `return stem` always, since the lowercasing is handled above

PhanatosZou: It make sense to me! Thank you so much:)

PhanatosZou: > hi @PhanatosZou i _think_ you're just backwards is all :) in fact, it's probably simpler to simply `return stem` always, since the lowercasing is handled above

Make sense to me! Thank you so much:)

PhanatosZou:
Just want to follow up with this issue. When this pull request be merged to the base branch? I appreciate if anyone can help. Thanks in advance!

nltk/stem/porter.py
```diff
@@ -648,17 +648,21 @@ def _step5b(self, word):
...
         if self.mode == self.NLTK_EXTENSIONS and word in self.pool:
-            return self.pool[word]
+            return self.pool[stem]
 
         if self.mode != self.ORIGINAL_ALGORITHM and len(word) <= 2:
             # With this line, strings of length 1 or 2 don't go through
             # the stemming process, although no mention is made of this
             # in the published algorithm.
-            return word
+            return stem if to_lowercase else word
```
dannysepler: you could just return stem, no?

PhanatosZou: > you could just return stem, no?

Yes you are right! Since the stem already takes care of the lowercase. Thank you:)

stevenbird:
Thanks @dannysepler... sorry for the delay!

PR #2515, opened 2020-03-10
Removed unecessary keys call

Hk-tang:
It is not necessary to call keys() in the argument to sorted. Iterating over a dict will give you the keys. self._reflections is sufficient.


stevenbird:
Thanks @Hk-tang

PR #2536, opened 2020-05-01
Fix punkt tokenizer complexity for very long tokens

mariokostelac:
This change prevents N^2 complexity for inputs with very long tokens (N being the length of the longest token). Instead of using `finditer` with regex that starts from any point in the string, we use match method that starts only at the string beginning and moves from one opportunity to another (either end of next blank sequence of previous break).

**Why should we bother changing complexity**
I've discovered that while processing some real world conversational data. The current implementation choked on some strings having tokens 60k characters long (yes, over time, users will find all sorts of bottlenecks in the real system 🤷‍♂️), taking > 8s to tokenise that that string. Reducing to complexity to this brings the execution below few millis. I can't remember exact numbers, but can create some benchmark if it's bringing extra value here.

**What was N^2 exactly?**
Used regex can start at any position in the string. If we have a token
like 01234567890123...9, \S* starts matching 0 and matches everything till
the last 9 and then figures out that it does not have potential sentence
ending character.

Since finditer returns matches starting at any position, it tries matching with character 1, matches everything till the end and then does not find potential sentence ending, ... It does that matching N times, each time going through N/2 characters (on average).

**Complexity test**
I've found regression tests like the one I wrote useful. If you are ok with it, I can improve the robustness.

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -198,7 +198,7 @@ Two instances of PunktSentenceTokenizer should not share PunktParameters.
     >>> pst2 = PunktSentenceTokenizer()
     >>> pst._params is pst2._params
     False
-    
+
 Testing mutable default arguments for https://github.com/nltk/nltk/pull/2067
 
     >>> from nltk.tokenize.punkt import PunktBaseClass, PunktTrainer, PunktSentenceTokenizer
@@ -214,7 +214,7 @@ Testing mutable default arguments for https://github.com/nltk/nltk/pull/2067
     >>> pst = PunktSentenceTokenizer(lang_vars=None)
     >>> type(pst._lang_vars)
     <class 'nltk.tokenize.punkt.PunktLanguageVars'>
-    
+
 
 Regression Tests: align_tokens
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -5,6 +5,7 @@ See also nltk/test/tokenize.doctest
 """
 
 
+import timeit
 import unittest
 
 from nose import SkipTest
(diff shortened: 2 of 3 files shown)


mariokostelac:
Just saw there are more rests in doctest files. Will make sure all pass.

mariokostelac:
~~Interesting, I can replicate sentiment doctest failures locally, but not tokenize doctest failures.~~ I just did not know how to run all doctests 🤦‍♂️

mariokostelac:
Added more context on the effect of this change on performance.

mariokostelac:
@stevenbird any chance you know when somebody would have time to review this? This is my first contribution here so I am not sure what's the usual cadence.

stevenbird:
@mariokostelac, thanks for your PR. I'll review it during the week and respond here.

stevenbird:
@mariokostelac, can you please explain the context where you are sentence tokenizing with tokens up to 60kB? There needs to be a case for adding to the complexity of the code. We subscribe to the [worse is better](https://www.jwz.org/doc/worse-is-better.html) doctrine, where simplicity of an implementation trumps completeness.

mariokostelac:
That's interesting, I've never heard of that doctrine, thanks for linking it.

I've stumbled upon this performance issue while debugging some ETL jobs I've created in Intercom. They process real world, text data entered by users in messenger (you can see one on the bottom right of https://biasandvariance.com/).

I'd be less concerned if I found this theoretically, but it actually took a while to find a needle in the haystack and figure out it's part of NLTK library. Of course there is a way to avoid it before it hits NLTK (deleting long tokens, for example; that's what I do now), but then it means I have to to think about it every time I'm putting NLTK in production, which makes the library less usable.

Edit: turns out I just did not the know that doctrine under that name.

jnothman (approved):

mariokostelac:
If you're ok with the approach, I should probably change https://github.com/nltk/nltk/blob/7fd2bf5d52fabea678d546cef632463e7c9f13ae/nltk/tokenize/punkt.py#L1282 before merging, too.

stevenbird:
@mariokostelac sorry for the long delay... I've been reluctant to add complexity to the codebase, but if you're still keen, please go ahead with your other proposed change, and please resolve the merge conflicts, and then I'll merge. Thanks.

PR #2537, opened 2020-05-10
Use project_urls key to show additional PyPI links

bsolomon1124:
Reference:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls

Example (from `requests`):

Screen Shot 2020-05-10 at 9 42 50 AM


stevenbird:
Thanks @bsolomon1124

PR #2541, opened 2020-05-13
Mistake/bleu score.py

shushengyuan:
i just found a little mistake in nltk/translate/bleu_score.py where the comments of function Smoothing7 is wrong.
It should be
"
| Smoothing method 7:
| Interpolates methods 4 and 5.
"
rather then
"
| Smoothing method 7:
| Interpolates methods 5 and 6.
"


stevenbird:
Thanks @shushengyuan

PR #2547, opened 2020-05-27
Fix check for existing trigram_model in Text.generate()

thomasbird:
When generating text with `nlt.text.Text.generate()` a trigram model is created if one doesn't already exist, but the check to see if one exists has a small bug.

The check looks for the model at `self.trigram_model`, but it should look at `self._trigram_model`. This is defined two lines down. Since this check is never successful, it always creates a new model each time `generate()` is called.

This MR fixes this issue and speeds up multiple calls of `generate()`.


stevenbird:
Thanks @thomasbird!

PR #2555, opened 2020-06-14
Reduce inference time for some ngram language models

palasso:
A minimal set of changes to reduce inference time by order 100-1000x less.
This is a rebase of develop and the result of discussion in PR #2496 in how to fix the performance issue for ngram language models with:

- Lidstone smoothing
- Kneser-Ney Smoothing with Interpolation

This does not reduce inference time for the ngram language models with Witten-Bell Smoothing with Interpolation.

Relevant information and discussions have taken place in PR #2496 and Issue #2495

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -261,6 +261,7 @@
 - Alexandre H. T. Dias <https://github.com/alexandredias3d>
 - Jacob Weightman <https://github.com/jacobdweightman>
 - Bonifacio de Oliveira <https://github.com/Bonifacio2>
+- Vassilis Palassopoulos <https://github.com/palasso>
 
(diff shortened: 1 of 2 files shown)


iliakur:
:+1: from me to merge this.

In the next month or so I'll try to come up with a way to automate tests that confirm this behaviour based on what we used ased in the discussion of #2496

stevenbird:
Looks good, thanks @palasso!

palasso:
Thank you all for the help in improving it and moving it forward!

iliakur:
Thanks for sticking through the process. Feel free to poke around some more for places to optimize  if you have time :)

PR #2556, opened 2020-06-17
Use chain.from_iterable in api.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.


stevenbird:
Thanks @cool-RR

Issue #2557, opened 2020-06-18
TweetTokenizer - OverflowError: signed integer is greater than maximum


louis925:
```
File "/usr/local/lib/python3.7/dist-packages/nltk/tokenize/casual.py", line 286, in tokenize
    text = _replace_html_entities(text)
  File "/usr/local/lib/python3.7/dist-packages/nltk/tokenize/casual.py", line 249, in _replace_html_entities
    return ENT_RE.sub(_convert_entity, _str_to_unicode(text, encoding))
  File "/usr/local/lib/python3.7/dist-packages/nltk/tokenize/casual.py", line 243, in _convert_entity
    return chr(number)
OverflowError: signed integer is greater than maximum
```

In the source code, it try to catch errors but it doesn't include the Overflow cases.
https://github.com/nltk/nltk/blob/88c437b0f84cffc7bb63545cd809ebc92239c477/nltk/tokenize/casual.py#L241-L245.

In python 3.7, `chr()` can raise `OverflowError` when value >= 0x80000000 (or < -0x80000000). Suggest to fix the line as 
```python
except ValueError, OverflowError:
    pass
```

louis925:
Fixed by the PR

PR #2558, opened 2020-06-18
FIX OverflowError in TweetTokenizer

louis925:
In python 3.7, `chr()` can raise `OverflowError` when value >= 0x80000000 (or < -0x80000000). This is encountered in https://github.com/nltk/nltk/issues/2557


stevenbird:
Thanks @louis925

PR #2559, opened 2020-06-19
Fix exception causes in 13 modules

cool-RR:
@stevenbird This is a continuation of #2553


stevenbird:
Thanks very much @cool-RR

PR #2560, opened 2020-06-20
Use chain.from_iterable all over the codebase

cool-RR:
@stevenbird  This is a continuation of #2556


stevenbird:
Thanks @cool-RR!

PR #2563, opened 2020-06-21
performance test for vocabulary len being independent of its size

iliakur:
Small follow-up to recent work by @palasso.

I'm still not sure how robust this test will be to executing on other machines to be honest.

It also takes 2ish seconds on my computer, which means it'll add some overhead to the test suite overall. Not a big deal for CI, but may impact local development. At the same time, I personally don't actually run *all* the NLTK tests when I'm working on the `lm` module but only its tests.

If we at any point switch to `pytest` for managing tests, I could mark it (and some others) with a label so that they can be ignored when speed of test execution matters.

### DIFF
--- a/nltk/test/unit/lm/test_vocabulary.py
+++ b/nltk/test/unit/lm/test_vocabulary.py
@@ -7,6 +7,7 @@
 
 import unittest
 from collections import Counter
+from timeit import timeit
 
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @iliakur; sorry for the delay!

PR #2564, opened 2020-06-22
Bug fix/accommodate iteratables in everygrams resolves #2562

### LINKED ISSUE #2562: nltk.everygrams() does not work with non-reusable iterables and iterables without lengths.
`nltk.util.everygrams()` ([code](https://github.com/nltk/nltk/blob/develop/nltk/util.py#L577-L600), [docs](https://nltk.readthedocs.io/en/latest/api/nltk.html?highlight=everygrams#nltk.util.everygrams)) claims to support both sequences and iterables. However, for iterables that are not reusable or iterables without a length function, everygrams either raises an exception or behaves incorrectly. In particular:
* When calling `everygrams(obj)` without its optional `max_len` argument, it attempts to take the length of its first argument, which may not support `len(obj)` resulting in a `TypeError`.
* When calling `everygrams(obj, max_len=2)` for example, if obj is an iterable that is not reusable it will be exhausted after only the ngrams of length `1` have been found and `everygrams()` ends up returning only the ngrams of length `1`.

I am happy to make a PR for this.

## Non-Working Examples

For example, starting with the example from the documentation
```python
>>> from nltk import everygrams
>>> sent = 'a b c'.split()
>>> list(everygrams(sent))
[('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c'), ('a', 'b', 'c')]
>>> list(everygrams(sent, max_len=2))
[('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c')]
```

We should expect that `everygrams(iter(sent))` should have the same output as `everygrams(sent)`. Instead, it raises a `TypeError`.
```python
>>> list(everygrams(sent))
[('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c'), ('a', 'b', 'c')]
>>> list(everygrams(iter(sent)))
  File "/.../python3.8/site-packages/nltk/util.py", line 603, in everygrams
    max_len = len(sequence)
TypeError: object of type 'list_iterator' has no len()
```
Likewise, we should expect that `everygrams(iter(sent), max_len=2)` should have the same output as `everygrams(sent, max_len=2)`. This is not the case.
```python
>>> list(everygrams(sent, max_len=2))
[('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c')]
>>> list(everygrams(iter(sent), max_len=2))
[('a',), ('b',), ('c',)]
```

## Possible Solutions

There are several possible solutions:
1. Do not accept iterables.
    * Since nltk is commonly used to process large amounts of text data, iterables are the most natural and practical approach to avoid loading data into memory all at once.
2. If the argument is an iterable, cache it, computing its length in the process, in case it cannot be reused.
    * This would require loading the data into memory all at once, which should be avoided.
3. Use a `history` sliding window of size `max_len` to produce the ngrams. This strategy is similar to `history` in the implementation of ngram.
    * This would be my suggested approach since it provides the most functionality. However, it would have the drawback of changing the order of the output ngrams. For example, 
      ```python
      >>> list(old_everygrams(iter(sent)))
      [('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c'), ('a', 'b', 'c')]
      >>> list(new_everygrams(iter(sent)))
      [('a',), ('a', 'b'), ('a', 'b', 'c'), ('b',), ('b', 'c'), ('c',)]
      ```

I am happy to make a PR for this.

dmmolitor:
This is a PR for [issue 2562](https://github.com/nltk/nltk/issues/2562). It modifies the everygrams function to work with iterables. It modifies the necessary tests to accommodate the difference in order output of the grams.

### DIFF
--- a/nltk/test/lm.doctest
+++ b/nltk/test/lm.doctest
@@ -129,3 +129,19 @@ For speed take only the first 100 sentences of reuters. Shouldn't affect the tes
     >>> lm.fit(train_data, vocab_data)
     >>> lm.score("said", ("",)) < 1
     True
+
+
+
+
+Issue 2562
+---------
+https://github.com/nltk/nltk/issues/2562
+
+
+Test that everygrams works for iterables.
+    >>> from nltk import everygrams
+    >>> sent = 'a b c'.split()
+    >>> list(everygrams(iter(sent)))
+    [('a',), ('a', 'b'), ('a', 'b', 'c'), ('b',), ('b', 'c'), ('c',)]
+    >>> list(everygrams(iter(sent), max_len=2))
+    [('a',), ('a', 'b'), ('b',), ('b', 'c'), ('c',)]
--- a/nltk/test/lm.doctest
+++ b/nltk/test/lm.doctest
@@ -129,19 +129,3 @@ For speed take only the first 100 sentences of reuters. Shouldn't affect the tes
     >>> lm.fit(train_data, vocab_data)
     >>> lm.score("said", ("",)) < 1
     True
-
-
-
-
-Issue 2562
----------
-https://github.com/nltk/nltk/issues/2562
-
-
-Test that everygrams works for iterables.
-    >>> from nltk import everygrams
-    >>> sent = 'a b c'.split()
-    >>> list(everygrams(iter(sent)))
-    [('a',), ('a', 'b'), ('a', 'b', 'c'), ('b',), ('b', 'c'), ('c',)]
-    >>> list(everygrams(iter(sent), max_len=2))
-    [('a',), ('a', 'b'), ('b',), ('b', 'c'), ('c',)]
--- a/nltk/test/unit/lm/test_preprocessing.py
+++ b/nltk/test/unit/lm/test_preprocessing.py
@@ -13,15 +13,15 @@ class TestPreprocessing(unittest.TestCase):
     def test_padded_everygram_pipeline(self):
         expected_train = [
             [
-                 ('<s>',),
-                 ('<s>', 'a'),
-                 ('a',),
-                 ('a', 'b'),
-                 ('b',),
-                 ('b', 'c'),
-                 ('c',),
-                 ('c', '</s>'),
-                 ('</s>',)
+                 ("<s>",),
+                 ("<s>", "a"),
+                 ("a",),
+                 ("a", "b"),
+                 ("b",),
+                 ("b", "c"),
+                 ("c",),
+                 ("c", "</s>"),
+                 ("</s>",)
              ]
         ]
         expected_vocab = ["<s>", "a", "b", "c", "</s>"]
--- a/nltk/test/unit/lm/test_preprocessing.py
+++ b/nltk/test/unit/lm/test_preprocessing.py
@@ -13,15 +13,15 @@ class TestPreprocessing(unittest.TestCase):
     def test_padded_everygram_pipeline(self):
         expected_train = [
             [
-                 ("<s>",),
-                 ("<s>", "a"),
-                 ("a",),
-                 ("a", "b"),
-                 ("b",),
-                 ("b", "c"),
-                 ("c",),
-                 ("c", "</s>"),
-                 ("</s>",)
+                ("<s>",),
+                ("<s>", "a"),
+                ("a",),
+                ("a", "b"),
+                ("b",),
+                ("b", "c"),
+                ("c",),
+                ("c", "</s>"),
+                ("</s>",)
              ]
         ]
         expected_vocab = ["<s>", "a", "b", "c", "</s>"]
--- a/nltk/test/unit/lm/test_preprocessing.py
+++ b/nltk/test/unit/lm/test_preprocessing.py
@@ -13,16 +13,16 @@ class TestPreprocessing(unittest.TestCase):
     def test_padded_everygram_pipeline(self):
         expected_train = [
             [
-                ("<s>",),
-                ("a",),
-                ("b",),
-                ("c",),
-                ("</s>",),
-                ("<s>", "a"),
-                ("a", "b"),
-                ("b", "c"),
-                ("c", "</s>"),
-            ]
+                 ('<s>',),
+                 ('<s>', 'a'),
+                 ('a',),
+                 ('a', 'b'),
+                 ('b',),
+                 ('b', 'c'),
+                 ('c',),
+                 ('c', '</s>'),
+                 ('</s>',)
+             ]
         ]
         expected_vocab = ["<s>", "a", "b", "c", "</s>"]
         train_data, vocab_data = padded_everygram_pipeline(2, [["a", "b", "c"]])
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -13,29 +13,6 @@ class TestEverygrams(unittest.TestCase):
         """Form test data for tests."""
         self.test_data = 'a b c'.split()
 
-    def test_everygrams_pad_left(self):
-        expected_output = set(
-            [
-                (None),
-                (None, None),
-                (None, None, 'a'),
-                (None),
-                (None, 'a'),
-                (None, 'a', 'b'),
-                ('a',),
-                ('a', 'b'),
-                ('a', 'b', 'c'),
-                ('b',),
-                ('b', 'c'),
-                ('c',),
-            ]
-        )
-        output = list(everygrams(self.test_data, max_len=3, pad_left=True))
-        # Test for expected output.
-        self.assertEqual(set(output), expected_output)
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
-
     def test_everygrams(self):
         test_data = 'a b c'.split()
         # Test everygrams with defaults.
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -14,46 +14,27 @@ class TestEverygrams(unittest.TestCase):
         self.test_data = 'a b c'.split()
 
     def test_everygrams(self):
-        test_data = 'a b c'.split()
-        # Test everygrams with defaults.
-        expected_output = set(
-            [
-                ('a',),
-                ('a', 'b'),
-                ('a', 'b', 'c'),
-                ('b',),
-                ('b', 'c'),
-                ('c',),
-            ]
-        )
-        output = list(everygrams(self.test_data))
-        # Test for expected output.
-        self.assertEqual(set(output), expected_output)
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
-
-        # Test that everygrams works with iterables.
-        output = list(everygrams(iter(test_data)))
-        # Test for expected output.
-        self.assertEqual(set(output), expected_output)
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
+        # Test everygrams without padding.
+        expected_output = [
+            ('a',),
+            ('a', 'b'),
+            ('a', 'b', 'c'),
+            ('b',),
+            ('b', 'c'),
+            ('c',),
+        ]
+        output = everygrams(iter(self.test_data))
+        self.assertCountEqual(output, expected_output)
 
         # Test everygrams with a max length of 2.
-        expected_output = set([('a',), ('a', 'b'), ('b',), ('b', 'c'),('c',),])
-        output = list(everygrams(self.test_data, max_len=2))
-        # Test for expected output.
-        self.assertEqual(set(output), expected_output)
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
+        expected_output = [('a',), ('a', 'b'), ('b',), ('b', 'c'),('c',),]
+        output = everygrams(iter(self.test_data), max_len=2)
+        self.assertCountEqual(output, expected_output)
 
-        # Test everygrams with a max length of 2.
-        expected_output = set([('a', 'b'), ('b', 'c'), ('a', 'b', 'c'),])
-        output = list(everygrams(self.test_data, min_len=2))
-        # Test for expected output.
-        self.assertEqual(set(output), expected_output)
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
+        # Test everygrams with a min length of 2.
+        expected_output = [('a', 'b'), ('b', 'c'), ('a', 'b', 'c'),]
+        output = everygrams(iter(self.test_data), min_len=2)
+        self.assertCountEqual(list(output), expected_output)
 
     def test_everygrams_pad_right(self):
         expected_output = [
@@ -70,12 +51,8 @@ class TestEverygrams(unittest.TestCase):
             (None, None),
             (None,),
         ]
-        output = list(everygrams(self.test_data, max_len=3, pad_right=True))
-        # Test for expected output.
-        self.assertEqual(set(output), set(expected_output))
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
-
+        output = everygrams(iter(self.test_data), max_len=3, pad_right=True)
+        self.assertCountEqual(list(output), expected_output)
 
     def test_everygrams_pad_left(self):
         expected_output = [
@@ -92,29 +69,5 @@ class TestEverygrams(unittest.TestCase):
             ('b', 'c'),
             ('c',),
         ]
-        output = list(everygrams(self.test_data, max_len=3, pad_left=True))
-        # Test for expected output.
-        self.assertEqual(set(output), set(expected_output))
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
-
-    def test_everygrams_iterable(self):
-        test_data = 'a b c'.split()
-        # Test everygrams with defaults.
-        expected_output = set(
-            [
-                ('a',),
-                ('a', 'b'),
-                ('a', 'b', 'c'),
-                ('b',),
-                ('b', 'c'),
-                ('c',),
-            ]
-        )
-
-        # Test that everygrams works with iterables.
-        output = list(everygrams(iter(test_data)))
-        # Test for expected output.
-        self.assertEqual(set(output), expected_output)
-        # Test that no grams are repeated.
-        self.assertEqual(len(output), len(expected_output))
+        output = everygrams(iter(self.test_data), max_len=3, pad_left=True)
+        self.assertCountEqual(list(output), expected_output)
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -33,7 +33,7 @@ class TestEverygrams(unittest.TestCase):
     def test_everygrams_min_len(self):
         expected_output = [('a', 'b'), ('b', 'c'), ('a', 'b', 'c'),]
         output = everygrams(self.test_data, min_len=2)
-        self.assertCountEqual(list(output), expected_output)
+        self.assertCountEqual(output, expected_output)
 
     def test_everygrams_pad_right(self):
         expected_output = [
@@ -51,7 +51,7 @@ class TestEverygrams(unittest.TestCase):
             (None,),
         ]
         output = everygrams(self.test_data, max_len=3, pad_right=True)
-        self.assertCountEqual(list(output), expected_output)
+        self.assertCountEqual(output, expected_output)
 
     def test_everygrams_pad_left(self):
         expected_output = [
@@ -69,4 +69,4 @@ class TestEverygrams(unittest.TestCase):
             ('c',),
         ]
         output = everygrams(self.test_data, max_len=3, pad_left=True)
-        self.assertCountEqual(list(output), expected_output)
+        self.assertCountEqual(output, expected_output)
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -8,9 +8,7 @@ from nltk.util import everygrams
 
 
 class TestEverygrams(unittest.TestCase):
-    # test_data = 'a b c'.split()
 
-    # @classmethod
     def setUp(self):
         """Form test data for tests."""
         self.test_data = 'a b c'.split()
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -11,10 +11,9 @@ class TestEverygrams(unittest.TestCase):
 
     def setUp(self):
         """Form test data for tests."""
-        self.test_data = 'a b c'.split()
+        self.test_data = iter('a b c'.split())
 
-    def test_everygrams(self):
-        # Test everygrams without padding.
+    def test_everygrams_without_padding(self):
         expected_output = [
             ('a',),
             ('a', 'b'),
@@ -23,17 +22,17 @@ class TestEverygrams(unittest.TestCase):
             ('b', 'c'),
             ('c',),
         ]
-        output = everygrams(iter(self.test_data))
+        output = everygrams(self.test_data)
         self.assertCountEqual(output, expected_output)
 
-        # Test everygrams with a max length of 2.
+    def test_everygrams_max_len(self):
         expected_output = [('a',), ('a', 'b'), ('b',), ('b', 'c'),('c',),]
-        output = everygrams(iter(self.test_data), max_len=2)
+        output = everygrams(self.test_data, max_len=2)
         self.assertCountEqual(output, expected_output)
 
-        # Test everygrams with a min length of 2.
+    def test_everygrams_min_len(self):
         expected_output = [('a', 'b'), ('b', 'c'), ('a', 'b', 'c'),]
-        output = everygrams(iter(self.test_data), min_len=2)
+        output = everygrams(self.test_data, min_len=2)
         self.assertCountEqual(list(output), expected_output)
 
     def test_everygrams_pad_right(self):
@@ -51,7 +50,7 @@ class TestEverygrams(unittest.TestCase):
             (None, None),
             (None,),
         ]
-        output = everygrams(iter(self.test_data), max_len=3, pad_right=True)
+        output = everygrams(self.test_data, max_len=3, pad_right=True)
         self.assertCountEqual(list(output), expected_output)
 
     def test_everygrams_pad_left(self):
@@ -69,5 +68,5 @@ class TestEverygrams(unittest.TestCase):
             ('b', 'c'),
             ('c',),
         ]
-        output = everygrams(iter(self.test_data), max_len=3, pad_left=True)
+        output = everygrams(self.test_data, max_len=3, pad_left=True)
         self.assertCountEqual(list(output), expected_output)
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -4,11 +4,7 @@ Unit tests for nltk.util.
 """
 
 import unittest
-from nltk.util import (
-    pad_sequence,
-    ngrams,
-    everygrams,
-)
+from nltk.util import everygrams
 
 
 class TestEverygrams(unittest.TestCase):
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+"""
+Unit tests for nltk.util.
+"""
+
+import unittest
+from nltk.util import (
+    pad_sequence,
+    ngrams,
+    everygrams,
+)
+
+
+class TestEverygrams(unittest.TestCase):
+    # test_data = 'a b c'.split()
+
+    # @classmethod
+    def setUp(self):
+        """Form test data for tests."""
+        self.test_data = 'a b c'.split()
+
+    def test_everygrams_pad_left(self):
+        expected_output = set(
+            [
+                (None),
+                (None, None),
+                (None, None, 'a'),
+                (None),
+                (None, 'a'),
+                (None, 'a', 'b'),
+                ('a',),
+                ('a', 'b'),
+                ('a', 'b', 'c'),
+                ('b',),
+                ('b', 'c'),
+                ('c',),
+            ]
+        )
+        output = list(everygrams(self.test_data, max_len=3, pad_left=True))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+    def test_everygrams(self):
+        test_data = 'a b c'.split()
+        # Test everygrams with defaults.
+        expected_output = set(
+            [
+                ('a',),
+                ('a', 'b'),
+                ('a', 'b', 'c'),
+                ('b',),
+                ('b', 'c'),
+                ('c',),
+            ]
+        )
+        output = list(everygrams(self.test_data))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+        # Test that everygrams works with iterables.
+        output = list(everygrams(iter(test_data)))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+        # Test everygrams with a max length of 2.
+        expected_output = set([('a',), ('a', 'b'), ('b',), ('b', 'c'),('c',),])
+        output = list(everygrams(self.test_data, max_len=2))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+        # Test everygrams with a max length of 2.
+        expected_output = set([('a', 'b'), ('b', 'c'), ('a', 'b', 'c'),])
+        output = list(everygrams(self.test_data, min_len=2))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+    def test_everygrams_pad_right(self):
+        expected_output = [
+            ('a',),
+            ('a', 'b'),
+            ('a', 'b', 'c'),
+            ('b',),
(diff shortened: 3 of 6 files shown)


alvations:
Thanks @dmmolitor, this is an interesting contribution!! 

It seems like there's no easy way to go around without materializing the sequence first, at least partially it has to be. 

Let me suggest that we implement a generic sliding window function first, e.g. :

```python
from itertools import tee

def sliding_window(iterable, n):
    iterables = tee(iterable, n)
    for i, sub_iterable in enumerate(iterables):
        for _ in range(i):
            next(sub_iterable, None)
    return zip(*iterables)
```

And then the everygram loops can be simplified =)

```python
from itertools import islice
from nltk.util import pad_sequence, ngrams

def everygrams(
    sequence,
    min_len=1,
    max_len=-1,
    pad_left=False,
    pad_right=False,
    left_pad_symbol=None,
    right_pad_symbol=None,
):
    # Get max_len for padding.
    if max_len == -1:
        try:
            max_len = len(sequence)
        except TypeError:
            sequence = list(sequence)
            max_len = len(sequence)

    # Pad if indicated using max_len.
    sequence = pad_sequence(
        sequence, max_len, pad_left, pad_right, left_pad_symbol, right_pad_symbol
    )
    
    for subsequence in sliding_window(sequence, max_len):
        for n in range(min_len, max_len + 1):
            for ng in ngrams(subsequence, n):
                yield ng
```

Possibly a nice test for the sliding window could be :

```python
>>> list(sliding_window(iter('a b c'.split()), n=2))
[('a', 'b'), ('b', 'c')]
```

This way the full sequence doesn't need to be materialized and we will also have a more standardized way of implementing windowing functions in the future =)

----

EDITED
===

Come to think of it. We can do even better, I think we can replace the sliding window that takes in the iterable as the new `ngrams()`, if we refresh our ngrams with the sliding window code:


```python
from itertools import tee

def ngrams(iterable, n):
    iterables = tee(iterable, n)
    for i, sub_iterable in enumerate(iterables):
        for _ in range(i):
            next(sub_iterable, None)
    return zip(*iterables)
```

And then in `everygrams()`.

```python
def everygrams(...):
    ...
    # Evaluate/Materialize the subsequence.
    for subsequence in ngrams(sequence, max_len):
        # Iterate through the n-grams orders.
        for n in range(min_len, max_len + 1):
            # Generate the ngrams requested for this subsequence.
            for ng in ngrams(subsequence, n):
                yield ng
```

Finally for the doctest, maybe it's good to show how to get the order of items from the old version through:

```python
# Old version outputs for everygrams.
>>> sorted(everygrams(iter('a b c'.split()), max_len=2), key=len)
[('a',), ('b',), ('b',), ('c',), ('a', 'b'), ('b', 'c')]

# New outputs for everygrams.
>>> list(everygrams(iter('a b c'.split()), max_len=2))
[('a',), ('b',), ('a', 'b'), ('b',), ('c',), ('b', 'c')]
```

@dmmolitor what do you think about the suggestions above?

iliakur:
Perhaps related functions from [here](https://more-itertools.readthedocs.io/en/stable/index.html) can be an inspiration for API design and even behavior?

nltk/test/unit/lm/test_preprocessing.py
```diff
@@ -13,16 +13,16 @@ class TestPreprocessing(unittest.TestCase):
...
-                ("a",),
-                ("b",),
-                ("c",),
-                ("</s>",),
-                ("<s>", "a"),
-                ("a", "b"),
-                ("b", "c"),
-                ("c", "</s>"),
-            ]
+                 ('<s>',),
```
iliakur: Hmm, how did this change happen? The automatic formatter we use (`black`) enforces both double quotes and commas at the end of lists.

Did you get a chance to run it?

nltk/test/lm.doctest
```diff
@@ -129,3 +129,19 @@ For speed take only the first 100 sentences of reuters. Shouldn't affect the tes
     >>> lm.fit(train_data, vocab_data)
     >>> lm.score("said", ("",)) < 1
     True
+
+
+
+
+Issue 2562
+---------
+https://github.com/nltk/nltk/issues/2562
```
iliakur: To think of it, this isn't really a bug with the language model module, right? All the lm code works with both sequences and iterables iirc and is documented as such. Would it be clearer if we add this regression tests to the `utils.doctest` file?

dmmolitor: I wasn't sure where to add the tests. I am happy to move them where ever you think is appropriate. Are you referring to the `util.doctest` file in the test folder? It could be beneficial to add tests for iterables with other related functions as well. What are your thoughts on this?

Most of the `*grams` functions (`ngrams`, `bigrams`, etc.) work with iterables, but the current implementation of everygrams assumes that the iterable has a `len` function defined (if `max_len` is not given) and potentially depletes the iteratable on the first call to `ngrams`.

```python
>>> list(everygrams(sent))
[('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c'), ('a', 'b', 'c')]
>>> list(everygrams(iter(sent)))
  File "/.../python3.8/site-packages/nltk/util.py", line 603, in everygrams
    max_len = len(sequence)
TypeError: object of type 'list_iterator' has no len()
>>> list(everygrams(sent, max_len=2))
[('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c')]
>>> list(everygrams(iter(sent), max_len=2))
[('a',), ('b',), ('c',)]
```

nltk/test/unit/lm/test_preprocessing.py
dmmolitor: Thanks for pointing this out. I only ran `tox -e py38` as directed by [CONTRIBUTING.md](https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md). I am currently getting an `InvocationError` when running `tox -e py-travis` and need to look into this. I can revert the formatting here though!

dmmolitor:
@alvations,  I like `next(sub_iterable, None)` instead of the try/except blocks to catch `StopIteration`. The staggered iterables using `tee` is also clever and is much faster than the current implementation of `ngrams`.  Using `islice(sub_iterable, i, None)` get staggered iterables also works, but is slower than `for _ in range(i): next(sub_iterable, None)`.
```python
>>> l=list(range(10000))                                                      
>>> %timeit list(ngrams(l, 3))                                                
2.54 ms ± 9.73 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
>>> %timeit list(suggested_ngrams(l, 3))                                      
709 µs ± 66.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

I timed code with the current list implementation of the sliding `history` window and it seems to be slightly faster than when using a `deque` as is done in `more-itertools`'s `windowed` function.

The suggested `everygrams` implementation does not currently produce the expected results since `ngrams` is called on each tuple of maximal length. This leads to repeated grams of smaller size. For example,
```python
>>> list(everygrams('a b c d'.split(' '), max_len=3))  
[('a',),
 ('b',),
 ('c',),
 ('a', 'b'),
 ('b', 'c'),
 ('a', 'b', 'c'),
 ('b',),
 ('c',),
 ('d',),
 ('b', 'c'),
 ('c', 'd'),
 ('b', 'c', 'd')]
```

Adjusting this naively to 
```python
def everygrams(...):
     ...
     # Evaluate/Materialize the subsequence.
     for long_gram in ngrams(sequence, max_len, **kwargs):
        # Iterate through the n-grams orders.
        for ngram_len in range(min_len, len(long_gram) + 1):
            # Generate the ngrams requested for this subsequence.
            yield long_gram[:ngram_len]
```
misses the smaller grams at the end of the sequence:
```python
>>> list(everygrams('a b c'.split(' '), max_len=3))                           
[('a',), ('a', 'b'), ('a', 'b', 'c')]
```

I can update the code to incorporate your comments later today.

nltk/test/unit/lm/test_preprocessing.py
iliakur: Hmm, indeed the automatic formatting hasn't made it into the guidelines, sorry about that. @alvations was there any interest in setting up pre-commit hooks to automate this?

nltk/test/lm.doctest
iliakur: I'm honestly leaning towards adding a `tests/unit/test_utils.py` file. I have a mind to move those `lm` regression tests out of the `doctest` file, because the formatting forces us to overspecify things like the order of sequences even though we don't really care about it. You've run into this already with the whole need for `sorted`... In a unit test that would be done with `assertCountEqual`.

nltk/test/unit/lm/test_preprocessing.py
alvations: Yes but I didn't manage to get the black to work pre-commit. In any case the examples in the test cases shouldn't change convention, i.e. doubles quotes and black formatted.

nltk/test/lm.doctest
alvations: IMHO, both doctest and unittest should exists, though it seem redundant. Partly because `help()` will show the docs when users want to learn more about it.

alvations:
@dmmolitor great, look forward to the edits! 

On the point of naively using slices in the yield missing out the 

```python
# Generate the ngrams requested for this subsequence.
     yield long_gram[:ngram_len]
```

It's better to put the subsequence through another `ngrams()`:

```python
# Generate the ngrams requested for this subsequence.
for ng in ngrams(subsequence, n):
    yield ng
````

e.g. something like this should work:


```python
from itertools import tee
from nltk.util import pad_sequence, ngrams

def ngrams(iterable, n):
    iterables = tee(iterable, n)
    for i, sub_iterable in enumerate(iterables):
        for _ in range(i):
            next(sub_iterable, None)
    return zip(*iterables)

def everygrams(
    sequence,
    min_len=1,
    max_len=-1,
    pad_left=False,
    pad_right=False,
    left_pad_symbol=None,
    right_pad_symbol=None,
):

    # Get max_len for padding.
    if max_len == -1:
        try:
            max_len = len(sequence)
        except TypeError:
            sequence = list(sequence)
            max_len = len(sequence)

    # Pad if indicated using max_len.
    sequence = pad_sequence(
        sequence, max_len, pad_left, pad_right, left_pad_symbol, right_pad_symbol
    )
    
    # Evaluate/Materialize the subsequence.
    for subsequence in ngrams(sequence, max_len):
        # Iterate through the n-grams orders.
        for n in range(min_len, max_len + 1):
            # Generate the ngrams requested for this subsequence.
            for ng in ngrams(subsequence, n):
                yield ng
```

----

P/S: It's awesome that you did a benchmark, I was using `next(sub_iterable, None)` for other codebases and kind of roughly felt it's faster but didn't think of benchmarking it properly. Thank you!!

nltk/test/lm.doctest
iliakur: Agreed! I'm just wondering if someone who calls `help()` is interested in reading the regression tests :)

dmmolitor:
The inner `ngrams` of 
```python
# Evaluate/Materialize the subsequence.
    for subsequence in ngrams(sequence, max_len):
        # Iterate through the n-grams orders.
        for n in range(min_len, max_len + 1):
            # Generate the ngrams requested for this subsequence.
            for ng in ngrams(subsequence, n):
                yield ng
```
causes the smaller grams to be repeated. The above implementation produces the following output.
```python
>>> list(everygrams('a b c d'.split(' '), max_len=3))  
[('a',),
 ('b',),
 ('c',),
 ('a', 'b'),
 ('b', 'c'),
 ('a', 'b', 'c'),
 ('b',),
 ('c',),
 ('d',),
 ('b', 'c'),
 ('c', 'd'),
 ('b', 'c', 'd')]
```

The original code I proposed also had the issue of producing ngrams that only include the padding symbol. I have updated the code to prevent this so that the output matches the previous implementation, although the code is not as succinct. I could add a test for this since this issue wasn't caught by previous tests.

alvations:
Good catch on the mistake @dmmolitor !! 

There's a hack to get around this but I think your implementation might be better:

```python
    
    # Evaluate/Materialize the subsequence.
    for i, subsequence in enumerate(ngrams(sequence, max_len)):
        # Iterate through the n-grams orders.
        for n in range(min_len, max_len + 1):
            if i == 0:
                # Generate the ngrams requested for this subsequence.
                for ng in enumerate(ngrams(subsequence, n)):
                    yield ng
            else:
                # Yield only the last element in the ngrams. 
                for ng in ngrams(subsequence, n):
                    pass
                yield ng
```

### Test

```python
sorted(everygrams('a b c d e f'.split(' '), max_len=4), key=len)  
```

[out]:

```
[('a',),
 ('b',),
 ('c',),
 ('d',),
 ('e',),
 ('f',),
 ('a', 'b'),
 ('b', 'c'),
 ('c', 'd'),
 ('d', 'e'),
 ('e', 'f'),
 ('a', 'b', 'c'),
 ('b', 'c', 'd'),
 ('c', 'd', 'e'),
 ('d', 'e', 'f'),
 ('a', 'b', 'c', 'd'),
 ('b', 'c', 'd', 'e'),
 ('c', 'd', 'e', 'f')]
```

nltk/test/lm.doctest
dmmolitor: Should I add @iliakur's suggested `tests/unit/test_utils.py` file and move the tests there?

iliakur: Yes please! Make sure to name it `test_util.py` though, to stay consistent with the name of the module you're testing.

nltk/test/unit/lm/test_preprocessing.py
```diff
@@ -13,16 +13,16 @@ class TestPreprocessing(unittest.TestCase):
     def test_padded_everygram_pipeline(self):
         expected_train = [
             [
-                ("<s>",),
```
iliakur, line 16: I've been thinking about these tests a bit. We don't really care about the order of the items in the sequence right? The important thing is that we "see" all the different ngram orders. Let's future-proof our tests by using sets instead of lists wherever possible. For instance, doctests still care about the order of items because they check the raw output of `print`, but all of our unittests would really benefit from the conversion.

iliakur:
@dmmolitor thanks for tackling this and so thoroughly too :)

I also learned something new from your benchmarks. I think your `ngrams` re-implementation is elegant and props for consistently applying [this rule](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html) to your inline comments :+1: 

Do you plan to add all the different examples you've mentioned here and in the issue as explicit tests? I think this would really make the most of all the work you and @alvations put into spelling out the expected behavior of this code. And I guess now we have a place where to add them, `test_util.py` ;)

nltk/test/unit/lm/test_preprocessing.py
dmmolitor: I think using sets here makes sense!

dmmolitor:
@iliakur Thank you for the feedback. I can add tests for the issues I've discussed in `test_util.py`.

iliakur:
Sorry for the spam, but it's late here and my brain is slow: does this change make `lm.preprocessing.padded_everygrams` obsolete? If yes, we can add a deprecation warning to it and remove it. I've always felt weird about the `preprocessing` module and would not at all mind dropping it. I could take care of the deprecation warning stuff, I've been wanting to learn how to do that :)

dmmolitor:
Looking at `lm.preprocessing.padded_everygrams`, the implementation seems to be slightly different than using `everygrams(sequence, max_len=order, pad_left=True, pad_right=True)` as `lm.preprocessing.padded_everygrams` pads the sequence before passing it to `everygrams`. This will potentially produce grams with only pad symbols. For example, I expect:
```python
>>> lm.preprocessing.padded_everygrams(2, ['a', 'b'])
[('</s>',), ('</s>', 'a'), ('a',), ('a', 'b'), ('b'), ('b', '</s>'), ('</s>',)]
```
while
```python
>>> everygrams(2, ['a', 'b'])
[ ('</s>', 'a'), ('a',), ('a', 'b'), ('b'), ('b', '</s>')]
```
This seems to have been the case with the previous implementations as well (although the output would have been ordered differently).

It would be great if you could deal with deprecation warnings @iliakur.

nltk/util.py
```diff
@@ -574,30 +549,72 @@ def trigrams(sequence, **kwargs):
...
+    sequence = pad_sequence(sequence, max_len, pad_left, pad_right, **kwargs)
+
+    # Sliding window to store grams.
+    history = list(islice(sequence, max_len))
+
+    # Prevent ngrams with only the padding symbol.
+    if pad_left:
+        min_len_start = max_len
+    else:
+        min_len_start = min_len
```
alvations: I think this is okay to expect the unigrams with the just the `<s>` and `<\s>` so that it's consistent with the lm preprocessing function.

nltk/util.py
```diff
@@ -504,34 +497,16 @@ def ngrams(
...
     :param n: the degree of the ngrams
     :type n: int
-    :param pad_left: whether the ngrams should be left-padded
-    :type pad_left: bool
-    :param pad_right: whether the ngrams should be right-padded
-    :type pad_right: bool
-    :param left_pad_symbol: the symbol to use for left padding (default is None)
-    :type left_pad_symbol: any
-    :param right_pad_symbol: the symbol to use for right padding (default is None)
-    :type right_pad_symbol: any
```
alvations, line 514: Although it's not explicit in the function definition (since it's implied by the kwargs), we shouldn't remove the docs, so that the user know these arguments are possible in the ngram function. I think it's better to put it back.

dmmolitor: Done!

nltk/util.py
dmmolitor: I can revert the implementation so that the output is the same as that of `lm.preprocessing.padded_everygrams`. To clarify, this means that if there is both right and left padding, then everygrams would produce two ngrams with only padding symbols for each length from `min_len` to `max_len-1`.

nltk/test/unit/lm/test_preprocessing.py
iliakur: coolio, thanks for taking care of this!

nltk/util.py
alvations: Yes I think it's good to be consistent with the `lm.preprocessing.padded_everygrams` outputs. What do you think @iliakur ?

iliakur:
Do you mind if we wait with this PR until tomorrow evening CET? I will have time to sit down then and try to understand the interface we're presenting with `padded_everygrams` and `everygrams(seq, pad_left=True, pad_right=True)`.

I have a feeling that if we're not clear about it, then it must be very confusing to the users who don't see the code itself.

I would ideally want to condense them into just one function (even if it means adding an argument to it). Barring that I think the docstrings of both should explicitly reference the other one and explain/warn about the difference.

iliakur:
Ok, conceptually `everygrams` is just a recursive application of `ngrams`. Since `ngrams` ignores padding when generating unigrams, it makes sense that `everygrams` behave the same way.

That being said, my impression is that in practice when using padding and `everygrams` the preferred behavior is that of `padded_everygrams`, i.e. that the padding symbols are also added at the unigram level. Is this a fair assessment? 

If it is, would it be hard to add support for this kind of padding in `everygrams`, perhaps with the addition of another boolean flag like `pad_unigrams=False`? I think this should remain backwards compatible and we can also begin removing `padded_everygrams` and consolidating the interface a bit.

@alvations @stevenbird what do you think?

P.S.
I'm a bit busy this week and likely next so may take a while to respond.

alvations:
Streamlining the interface is good, having `nltk.utils.everygram` instead of `nltk.lm.preprocessing.padded_everygrams` would be preferred since it's quite generic. 

I think back-compatibility for the padded everygram output can be dropped, i.e. to just generate `<s>` and `</s>` unigrams when padding is called. Users can choose to drop them later on if necessary. So the default behavior should be 

 - (no padding) to iterate through all subsequences and generate for all ngrams order, the individual ngrams
 - (with padding) to pad the sequence with `n` value set to `max_len`, then iterate through all subsequences and generate for all ngrams order, the individual ngrams

I.e. for this PR, we need to check that:

```python
>>> list(everygrams(padded_bigrams, max_len=2))
[('<s>',), ('<s>', 'a'), ('a',), ('a', 'b'), ('b',), ('b', 'c'), ('c',), ('c', '</s>'), ('</s>',)]
```

@iliakur, is the above the expected behavior of `nltk.lm.preprocessing.padded_everygrams` too? Or should it do:

```python
>>> list(everygrams(padded_bigrams, max_len=2))
[('<s>', '<s>',), ('<s>',), ('<s>', 'a'), ('a',), ('a', 'b'), ('b',), ('b', 'c'), ('c',), ('c', '</s>'), ('</s>',), ('</s>', '</s>',)]
```

----

Then we should create another issue to track the tasks of:

 - Add deprecation warning to `lm.preprocessing.padded_everygrams`
 - Refactor functions that uses `lm.preprocessing.padded_everygrams` to `nltk.utils.everygram` with appropriate `*args` changes.

iliakur:
The first example you posted is the output of `padded_everygrams` for `max_len=2`. This is why I'm not sure how specialized this is: when working on the `lm` package I discovered that the convention for padding a sentence given an ngram order N is to insert N-1 padding symbols at both ends of it. Should this be the default behavior of `everygrams` too (provided it's passed a flag)?

dmmolitor:
Both `padded_everygrams` and `pad_sequence`, which is used for `ngrams` and `everygrams` pad with N-1 padding symbols. It seems padding with N-1 was previously the default for both.

iliakur:
Just letting you know, I'm not likely to get to this until next week.

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+"""
+Unit tests for nltk.util.
+"""
+
+import unittest
+from nltk.util import everygrams
+
+
+class TestEverygrams(unittest.TestCase):
+    # test_data = 'a b c'.split()
```
iliakur: Please remove the commented out lines.

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # @classmethod
+    def setUp(self):
+        """Form test data for tests."""
+        self.test_data = 'a b c'.split()
+
+    def test_everygrams_pad_left(self):
+        expected_output = set(
```
iliakur: A less verbose way to declare a set is to use the curly brackets:
```python
{
    (None,),
    (None, None)
    ...
}
```

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+            [
+                (None),
```
iliakur: Add a comma here to make it a tuple: `(None,)`

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+                ('b',),
+                ('b', 'c'),
+                ('c',),
+            ]
+        )
+        output = list(everygrams(self.test_data, max_len=3, pad_left=True))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
```
iliakur: Here and elsewhere: I think this test is redundant and covered by the previous assertion, you can get rid of it :)

If you're afraid that you're generating the same ngram multiple times, check out [this assert method](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual).

The idea would be to define your expectation as a set (see my other comment about sets vs lists), then cast it as a list and compare it to the output of your function:
```python
self.assertCountEqual(list(expected_output), output)
```

You could also assert that the list and set version of `output` contain the same counts of ngrams.

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+                ('a',),
+                ('a', 'b'),
+                ('a', 'b', 'c'),
+                ('b',),
+                ('b', 'c'),
+                ('c',),
+            ]
+        )
+        output = list(everygrams(self.test_data, max_len=3, pad_left=True))
+        # Test for expected output.
```
iliakur: Feel free to remove this comment, it only describes *what* the code does, not *why* and that question is clearly answered by the code because you chose a good variable name `expected_output`.

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+        # Test everygrams with a max length of 2.
+        expected_output = set([('a', 'b'), ('b', 'c'), ('a', 'b', 'c'),])
+        output = list(everygrams(self.test_data, min_len=2))
+        # Test for expected output.
+        self.assertEqual(set(output), expected_output)
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+    def test_everygrams_pad_right(self):
+        expected_output = [
```
iliakur, line 39: Here and elsewhere: I don't think you should make any commitments about the order in which the ngrams are generated. People will start relying on that and it ties our hands in case we decide to tweak the implementation for whatever reason. That's why I'm putting emphasis on using sets as much as possible instead of lists. The crucial expectation that we have of this function is that it yields all the relevant ngrams, *that's* what we want to test exhaustively.

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+            ('c',),
+        ]
+        output = list(everygrams(self.test_data, max_len=3, pad_left=True))
+        # Test for expected output.
+        self.assertEqual(set(output), set(expected_output))
+        # Test that no grams are repeated.
+        self.assertEqual(len(output), len(expected_output))
+
+    def test_everygrams_iterable(self):
+        test_data = 'a b c'.split()
```
iliakur: Why can't we use `self.test_data` here?

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+                ('a', 'b'),
+                ('a', 'b', 'c'),
+                ('b',),
+                ('b', 'c'),
+                ('c',),
+            ]
+        )
+
+        # Test that everygrams works with iterables.
+        output = list(everygrams(iter(test_data)))
```
iliakur: I think you don't necessarily need to test lists if your tests work with iterables. Lists and sequences in general are also iterable, which means that if you prove something works with `iter(seq)`, you don't really need to prove it also works with `seq`. Just as a sanity-check maybe throw in one test that has a list as input, but I'd recommend having the absolute majority operate on `iter(seq)` as input.

nltk/util.py
```diff
@@ -503,35 +496,25 @@ def ngrams(
...
     :param n: the degree of the ngrams
-    :type n: int
-    :param pad_left: whether the ngrams should be left-padded
-    :type pad_left: bool
-    :param pad_right: whether the ngrams should be right-padded
-    :type pad_right: bool
-    :param left_pad_symbol: the symbol to use for left padding (default is None)
-    :type left_pad_symbol: any
-    :param right_pad_symbol: the symbol to use for right padding (default is None)
+    :type n: int
```
iliakur: What editor do you use? It seems to be adding whitespace at the end of lines or at least not removing it. Let me know if you're having trouble getting it to remove such whitespace, I can quickly do it for you.

nltk/util.py
```diff
@@ -574,30 +557,63 @@ def trigrams(sequence, **kwargs):
...
+    Old version outputs for everygrams.
+        >>> sorted(everygrams(sent), key=len)
         [('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c'), ('a', 'b', 'c')]
+
         >>> list(everygrams(sent, max_len=2))
-        [('a',), ('b',), ('c',), ('a', 'b'), ('b', 'c')]
+        [('a',), ('a', 'b'), ('b',), ('b', 'c'), ('c',)]
 
-    :param sequence: the source data to be converted into trigrams
+    :param sequence: the source data to be converted into ngrams
```
iliakur: Should we mention here that this sequence will be instantiated if the length isn't passed?

If the sequence is long and the user deliberately turned it into an iterable to save memory, it may come as a surprise that the sequence is suddenly turned into a list.

nltk/test/unit/lm/test_preprocessing.py
```diff
@@ -13,16 +13,16 @@ class TestPreprocessing(unittest.TestCase):
...
-                ("a",),
-                ("b",),
-                ("c",),
-                ("</s>",),
-                ("<s>", "a"),
-                ("a", "b"),
-                ("b", "c"),
-                ("c", "</s>"),
-            ]
+                 ("<s>",),
```
iliakur: It seems like your editor is inserting whitespace again :)

iliakur (changes requested):

nltk/test/unit/test_util.py
dmmolitor: Yes, we should. Thanks for catching this.

nltk/test/unit/test_util.py
dmmolitor: I changed the tests to use `assertCountEqual` throughout.

nltk/test/unit/test_util.py
dmmolitor: Changed tests to use `assertCountEqual` to remove order dependence.

nltk/util.py
dmmolitor: I have been using Atom. I will try to keep a closer eye out for this.

nltk/test/unit/test_util.py
dmmolitor: I removed many of the comments in the tests.

nltk/util.py
iliakur: This may make your life easier:
https://stackoverflow.com/a/30403353/4501212

nltk/test/unit/test_util.py
iliakur: Looks good, so much simpler :)

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class TestEverygrams(unittest.TestCase):
+
+    def setUp(self):
+        """Form test data for tests."""
+        self.test_data = 'a b c'.split()
+
+    def test_everygrams(self):
+        # Test everygrams without padding.
```
iliakur: Let's move this comment into the function name: `test_everygrams_without_padding`

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,73 @@
...
+
+import unittest
+from nltk.util import everygrams
+
+
+class TestEverygrams(unittest.TestCase):
+
+    def setUp(self):
+        """Form test data for tests."""
+        self.test_data = 'a b c'.split()
```
iliakur: What if we set `test_data` to an iterable at this point, i.e. `iter('a b c'.split())`?

This would mean you can use the variable directly as is in your tests without having to call `iter` on it. You would also have to break up your first test method into 3 methods, which I actually think is a good thing.

iliakur (changes requested):
This is shaping up quite nicely, just a few more small tweaks :)

nltk/test/unit/test_util.py
dmmolitor: Done

nltk/test/unit/test_util.py
dmmolitor: Done.

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,72 @@
...
+        expected_output = [
+            ('a',),
+            ('a', 'b'),
+            ('a', 'b', 'c'),
+            ('b',),
+            ('b', 'c'),
+            ('c',),
+        ]
+        output = everygrams(self.test_data)
+        self.assertCountEqual(output, expected_output)
```
iliakur, line 26: Oh, this is cool, I didn't know `assertCountEqual` works with iterables, not just lists :)

nltk/test/unit/test_util.py
```diff
@@ -0,0 +1,72 @@
...
+        ]
+        output = everygrams(self.test_data, max_len=3, pad_right=True)
+        self.assertCountEqual(list(output), expected_output)
+
+    def test_everygrams_pad_left(self):
+        expected_output = [
+            (None,),
+            (None, None),
+            (None, None, 'a'),
+            (None,),
```
iliakur, line 61: Ok, one last sanity-check: we are not surprised that this `(None,)` item appears for the second time, right? This is, if I understand correctly, the major difference with `padded_everygrams` from the lm module.

dmmolitor: The same happens for `padded_everygrams` from the the lm module. For example, the following test passes:

```python
def test_padded_everygrams(self):
        expected_output = [
            ('<s>',),
            ('<s>', '<s>'),
            ('<s>', '<s>', 'a'),
            ('<s>',),
            ('<s>', 'a'),
            ('<s>', 'a', 'b'),
            ('a',),
            ('a', 'b'),
            ('a', 'b', 'c'),
            ('b',),
            ('b', 'c'),
            ('b', 'c', '</s>'),
            ('c',),
            ('c', '</s>'),
            ('c', '</s>', '</s>'),
            ('</s>',),
            ('</s>', '</s>'),
            ('</s>',),
        ]
        test_data = 'a b c'.split()
        output = padded_everygrams(3, iter(test_data))
        self.assertCountEqual(output, expected_output)
        output = everygrams(iter(test_data), max_len=3, pad_left=True, pad_right=True, left_pad_symbol='<s>', right_pad_symbol='</s>')
        self.assertCountEqual(output, expected_output)
```
`padded_everygrams` performs `everygrams` with specific defaults.

iliakur: I see. Very clear, thank you!

iliakur (approved):
Oki-doki, this looks good to me :)

dmmolitor:
Great, thanks! I am happy with it as well. Is it possible to link [issue 2562](https://github.com/nltk/nltk/issues/2562)? I'm not sure that I have the permissions to do this.

iliakur:
I don't think I have perms for that either. @alvations @stevenbird ?

alvations:
Sorry for chiming back late. A whirlwind of things for the last few months. 

@dmmolitor by use the #issue number on any comment in the pull-request would have automatically link the PR to the issues =)

There's one part that is a little terse though really clean, where we use the tee, zip to generate the ngrams. I've added the comments for that part, just in case we forget it (in the future). 

@iliakur @stevenbird yes it LGTM now! 

Thank you @dmmolitor and @jdmoorman  for all the work and @iliakur for the reviewing/mentoring through this issue.

stevenbird:
Thanks @dmmolitor, @iliakur, @alvations

PR #2568, opened 2020-07-01
Use the full license text and a separate notice file

ajorg-aws:
This is advised at https://infra.apache.org/apply-license.html

More importantly it makes it easier for license detection tools to discover the
license and includes the actual license in the project rather than rely on
external web resources.


stevenbird:
Thanks @ajorg-aws

Issue #2295, opened 2019-05-09
Treebank detokenizer fails to undo quote conversion


Dobatymo:
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
NLTK version: 3.4.1

Quotes `"` are converted to ` `` ` or `''` by the treebank tokenizer. The detokenizer is supposed to revert its regexes but fails to do so for these quotes. The space is lost also.

```python
>>> from nltk.tokenize.treebank import TreebankWordTokenizer, TreebankWordDetokenizer
>>> tmp = TreebankWordTokenizer().tokenize('How "are" you?')
>>> tmp
['How', '``', 'are', "''", 'you', '?']
>>> TreebankWordDetokenizer().tokenize(tmp)
'How``are"you?'
```

alvations:
Verifying bug. Yes this is particular to the detokenization of quotes. 

```python
>>> from nltk.tokenize.treebank import TreebankWordTokenizer, TreebankWordDetokenizer
>>> text = 'How "are" you?'
>>> t = TreebankWordTokenizer() 
>>> d = TreebankWordDetokenizer()
>>> t.tokenize(text)
['How', '``', 'are', "''", 'you', '?']
>>> d.detokenize(t.tokenize(text))
'How``are"you?'
>>> text = 'How are you?'
>>> d.detokenize(t.tokenize(text))
'How are you?'
```

Most probably caused by https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L286 and https://github.com/nltk/nltk/blob/develop/nltk/tokenize/treebank.py#L335

Issue #2220, opened 2019-01-24
Treebank detokenizer parenthesis issue


Dobatymo:
Python version: Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] on win32
NLTK version: 3.4

Parentheses in the input are not correctly detokenized.

```python
from nltk.tokenize.treebank import TreebankWordTokenizer, TreebankWordDetokenizer
t = TreebankWordTokenizer()
d = TreebankWordDetokenizer()
d.detokenize(t.tokenize("Hello (world)"))
```
Output: `'Hello (world )'` (notice the space)

ciaochiaociao:
same here

PR #2575, opened 2020-07-28
Fix treebank detokenizer

orsharir:
Fix a few issues with how the Treebank's word detokenizer mishandled quotes, punctuation marks, and parentheses, including #2295 and #2220.


stevenbird:
Thanks @orsharir. Sorry for the delay.

Issue #2584, opened 2020-08-19
Cannot import all translation modules


weissercn:
Some submodules under translate are implemented and described in the documentation, but cannot be imported and used. 

import nltk
nltk.translate.gleu_score.corpus_gleu

fails with this error:
AttributeError: module 'nltk.translate' has no attribute 'gleu_score'

weissercn:
I proposed a fix user pull request #2585

mohammedayub44:
I get the same error. Guess this has not been released yet.

tomaarsen:
This works now. If I run the following snippet:
```python
import nltk
nltk.translate.gleu_score.corpus_gleu()
```
Then the output I get is:
```python
TypeError: corpus_gleu() missing 2 required positional arguments: 'list_of_references' and 'hypotheses'
```
In short, you can use the function like this.

---

However, it's probably recommended to use something like this:
```python
from nltk.translate.gleu_score import corpus_gleu, sentence_gleu
```

Alternatively, you can import the `sentence_gleu` alias using either:
```python
from nltk.translate import gleu
```
or
```python
from nltk import gleu
```
In both of these cases, `gleu` is an alias for `sentence_gleu` from `nltk/translate/gleu_score.py`.

I'll close this as it seems to be resolved now.

- Tom Aarsen

PR #2585, opened 2020-08-19
Fix Issue #2584 Cannot import all translation modules

weissercn:
This fix allows users to use the following submodules of nltk.translate:
nist_score, chrf_score, gale_church, gdfa, gleu_score, phrase_based


stevenbird:
Thanks @weissercn

PR #2589, opened 2020-08-30
Make nltk.sentiment.vader deterministic

### LINKED ISSUE #2581: Bug introduced in nltk 3.5 leads to nondeterministic behavior in nltk.sentiment.vader
Test script test.py
```python
#!/usr/bin/env python3

import json
from nltk.sentiment.vader import SentimentIntensityAnalyzer

line="BUT fuck but thats it"

si = SentimentIntensityAnalyzer()
print(json.dumps(si.polarity_scores(line), sort_keys=True))
```

Run it multiple times and compare outputs
```shell
$ for ((i=0;i<10;i++)); do ./test.py; done | sort | uniq -c
      4 {"compound": -0.3071, "neg": 0.36, "neu": 0.64, "pos": 0.0}
      6 {"compound": -0.6956, "neg": 0.543, "neu": 0.457, "pos": 0.0}
```

The problem is this method
https://github.com/nltk/nltk/blob/b0cd83ded0c3c2394f878d8577d71187fa3f9ae4/nltk/sentiment/vader.py#L451-L460

The core of the probem is relying on order in sets which is random:
```shell
$ for ((i=0;i<10;i++)); do python -c "print(set(['BUT', 'but']))"; done | sort | uniq -c
      7 {'but', 'BUT'}
      3 {'BUT', 'but'}
```

So in one case `_but_check` triggers one `BUT`, in another case it triggers another `but`. That's why test.py dumps different scores for the same string.

The original text to trigger the error was this reddit comment:
```
We've got some studs prospects, but thats all they are, prospects. BUT FUCK WE WON THE LOTTERY
```

dannysepler:
Should resolve #2581 -- the issue was totally correct. It looks like the class would choose "but" or "BUT" at random as the first word in the set. Then, in the sentence given in the ticket that had both lowercase and uppercase, it would negate from the first "but" or "BUT" at random. My fix is to handle things case-insensitively.

I threw in a unit test for good measure.


nltk/test/sentiment.doctest
```diff
@@ -168,7 +168,7 @@ Vader
     Today SUX!
     compound: -0.5461, neg: 0.779, neu: 0.221, pos: 0.0,
     Today kinda sux! But I'll get by, lol
-    compound: 0.2228, neg: 0.195, neu: 0.531, pos: 0.274,
+    compound: 0.5249, neg: 0.138, neu: 0.517, pos: 0.344,
```
dannysepler, line 172: looks like another bug -- only `but` or `BUT` was caught before. it makes sense that fixing this bug would make the sentiment more positive

stevenbird:
Great contribution, thanks @dannysepler!

PR #2595, opened 2020-09-05
Run tests in pytest, not nose

### LINKED ISSUE #2590: Move from nose to pytest or nose2
https://nose.readthedocs.io/en/latest/ -- nose is on life support. I personally prefer pytest, but nose2 may also be considered. Has this been discussed very much yet?

dannysepler:
Resolves #2590. This is an attempt at swapping `runtests.py` to be a pytest runner rather than nose runner. Pytest is able to run unittest, nose, and pytests. This means we can migrate specific tests to a more pytest syntax at our own pace.

A few notes:
- Rednose wouldn't be necessary anymore, so removed references to it
- `runtests` isn't too useful anymore, so deleted it


nltk/test/corpus_fixt.py
```diff
@@ -1,3 +0,0 @@
-# -*- coding: utf-8 -*-
-
-from nltk.corpus import teardown_module
```
dannysepler, line 3: i was confused by files like these. they imported the teardown, but that teardown wasn't used in the fixture or corresponding doctest file

(EDIT: my b, realized that they're important. brought em back)

nltk/test/segmentation_fixt.py
```diff
@@ -1,11 +0,0 @@
-# -*- coding: utf-8 -*-
```
dannysepler, line 1: this file had no corresponding doctest file

iliakur:
I would love to take a look at this, but will only have time to do so in 2ish weeks. Thanks in advance for your patience!

iliakur (commented):
Wow, a herculean effort, many thanks!!

Do I understand correctly that we're still relying on nose for testing against different versions of Python?

I also have a question about the teardown for doctests: will it get executed if a test fails or produces an error? If not, is that a problem?

The pattern of setting things up before and tearing them down after a test is common enough that we have some options to implement it robustly. 

1. We can turn doctests into `unittest` tests, [as the docs say](https://docs.python.org/3/library/doctest.html#unittest-api)
1. At the cost of introducing another dependency we could [use some sphinx doctest functionality](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html)
1. Finally, we could convert the tests that require setup/teardown into `pytest` tests with fixtures.

What do you think?

dannysepler:
what a helpful comment, thank you!!

> Do I understand correctly that we're still relying on nose for testing against different versions of Python?

yeah i think that's mostly accurate. (my understanding is) `runtests.py` (which is called by tox) is swapped to pytest, but the tests themselves can be written in nose / unittest / pytest, and they would still work. I chose this way since it wouldn't, as you had warned earlier, tear the rug from under people who may be comfortable with nose.

> I also have a question about the teardown for doctests: will it get executed if a test fails or produces an error? If not, is that a problem?

Good callout. I don't think the teardown would get executed. And yes, that is a problem since these doctests can install large corpuses (corpora?).

Re your list: thank you, it's very helpful! I'll spend some time reviewing each option, and can report back / diff in my preference. My concern is rocking the boat too much (after all, this is just my second diff into the repo haha). But I like the three options, they all seem very interesting!

dannysepler:
actually @iliakur how about:

4. adding `--doctest-continue-on-failure` to the `pytest.ini` -- so whenever a doctest is run by pytest (which should be every time), it always continues to the end of the test (the teardown, if applicable)

iliakur:
Will that continue if the test throws an exception (I don't know if doctest lumps these together with failures that come from mismatch between expected and actual outputs)?

dannysepler:
oh good point. seems like it! not totally sure how to test, but i gave this

```
>>> assert False
>>> print("hi")
bye
```

and got errors for both

iliakur:
Sorry for disappearing.

I think adding that doctest flag makes sense. Could you include a comment mentioning the alternatives we discussed next to it? That way if we have to change it for whatever reason these ideas will be easy to find.

iliakur:
ok, let me take one more look at the whole PR in the next couple of days, but I think this is looking good!

dannysepler:
great! no rush, i know this is a biggie. i'm also curious what you (and maybe also @stevenbird) think the next steps could be.

i'm considering either porting nose things into pytest (eventually removing nose as a dependency), or targeting some of the slower tests and speeding up the suite.

iliakur:
Rn off the top of my head I can only say that I'm looking forward to replacing the metaclass-approach with `pytest.mark.parametrize` in the language model tests!

I'll give this some more thought though, thanks for asking.

dannysepler:
Hi there! Any update?

iliakur:
Sorry Danny, some other stuff popped up. I have it on my stack for this week!

CONTRIBUTING.md
```diff
@@ -136,6 +136,15 @@ For a better design of your code, we recommend using a technique called
 where you write your tests **before** writing the actual code that implements
 the desired feature.
 
+We recommend `pytest` as the way to run your tests, no matter which type of test it is:
```
iliakur: Why is this *only* a recommendation? Is there some non-pytest way to run the tests? `runtests.py`?

nltk/corpus/__init__.py
```diff
@@ -482,7 +482,7 @@ def demo():
     # demo()
     pass
 
-# ** this is for nose **
+# ** this is for unit testing **
```
iliakur, line 485: One potential next step would be to turn this into a pytest fixture and move it into the testing code as opposed to having it in "production" code.

This applies to the `setup_module` function as well.

tools/travis/coverage-pylint.sh
```diff
@@ -17,9 +17,9 @@ echo "$(pwd)"  # Know which directory tox is running this shell from.
 
 #coverage
 coverage erase
-coverage run --source=nltk $(pwd)/runtests.py -v --with-xunit
-coverage xml --omit=$(pwd)/*
-iconv -c -f utf-8 -t utf-8 nosetests.xml > nosetests_scrubbed.xml
+coverage run --source=nltk $(pwd)/runtests.py -v
```
iliakur: There's a [pytest plugin for coverage](https://github.com/pytest-dev/pytest-cov). Could you look into what we'd gain by using that? At work we seem to use it directly without having to install coverage.

pip-req.txt
```diff
@@ -11,4 +11,5 @@ gensim>=0.11.1
 pyparsing>=2.0.3
 twython>=3.2.0
 regex>=2019.08.19
-
+pytest>=6.0.1
```
iliakur: Why is pytest in both these requirements and in `requirements-test.txt`?

tox.ini
```diff
@@ -50,21 +53,24 @@ commands =
 basepython = python3.5
 deps =
     nose >= 1.2.1
-    rednose
+    pytest
```
iliakur, line 56: I understand why we're overriding the `deps` from `testenv` in `testenv:pypy` (see the comment), but why do we have to override them in the `nodeps` environments? @alvations @stevenbird come to think of it I'm not sure I understand what the `nodeps` environments are for :)

iliakur (commented):
Left some concrete questions. Some general discussion points are below.

## The runtests.py script
I noticed you asked a question in your PR description about supporting `--with-id` and in general passing arguments to `runtests.py`.

I know this is backwards incompatible, but I'm against us supporting custom arguments. Pytest already [supports selecting tests](https://docs.pytest.org/en/stable/usage.html#specifying-tests-selecting-tests) out of the box in a multitude of ways, so it sounds like a waste to me if we try to roll our own.

Actually, what does `runtests.py` give us that we don't have with `pytest`? Can we remove this script completely? Currently I see that it changes into the test directory, but could we take advantage of the fact that [pytest supports this setup](https://docs.pytest.org/en/stable/goodpractices.html#tests-as-part-of-application-code)?

If we must keep the `runtests.py` script, I'd just pass any arguments directly to `pytest` and let it handle them.

## What's next?
Here's what I think would make sense to tackle next:

1. Definitely remove nose as a dependency. Pytest docs have some notes about what features of nose aren't supported out of the box and what can be done to address that. For example you can change classes to inherit from `unittest.TestCase` and their setup/teardown should work as expected.
1. Pytest has a rich ecosystem of plugins, consider adding some, like [this one that executes tests in random order](https://pypi.org/project/pytest-randomly/), or [this one that executes them in parallel](https://docs.pytest.org/en/3.0.1/xdist.html).

iliakur:
One more thing to do: there is a pytest warning about two tests that use `yield` statement, that they are being ignored. You could fix that so that pytest picks those two tests up again.

tox.ini
dannysepler: there's an explanation here `web/dev/local_testing.rst` -- it seems like these environments were intended to have test requirements but not much beyond that? i can't speak to how often they're used at the moment though

tools/travis/coverage-pylint.sh
dannysepler: good idea, swapped to this

pip-req.txt
dannysepler: The way this file has been written, it seems to define the versions of _every_ pip package (nose and coverage are here too). I think your point is good, and I moved the pytest packages (and their versions) to `requirements-test.txt`

dannysepler:
thanks so much for the thorough review! i agree that `runtests.py` doesn't do much beyond what pytest supports out of the box, so i deleted `runtests.py` and cleaned up its references. (we might not even need the directory change, because tox already does that.) i hope nltk contributors are okay with adjusting their testing workflow, but it should be simpler this way.

i generally agree with all your `next steps` (for instance, [pytest-xdist](https://docs.pytest.org/en/3.0.1/xdist.html) has been fun to play around with), but am reading these next steps as tasks for followup diffs. After all, this diff rocks the boat in a few ways, so I'm hesitant to keep piling on large refactors here.

iliakur:
hi, I'm back at this.

You understood me correctly about the `next steps`. I also think this is a big enough change that we shouldn't pile too much on top of it.

As for the transition, yes it's big, but also actually necessary due to `nose` no longer being maintained.

@stevenbird @alvations could you weigh in with suggestions for how to make the transition less painful? Perhaps a big disclaimer in the Readme or contributor's guide that warns readers about the change? Or what about keeping the `runtests.py` file which doesn't actually run the tests but tells the user to switch over to `pytest`? In this scenario we'd set a future release by which we'd get rid of that script completely and indicate this in the message we show whoever runs it.

tox.ini
iliakur: Ah, ok I understand what that's about then. We'll deal with that later :)

dannysepler:
thanks again for everything, Ilia! just want to bump this thread for @stevenbird and @alvations. i'm pretty excited to get cracking on the next steps, so happy to do whatever it takes to get this into a merge-able state

stevenbird:
@iliakur, @dannysepler, sorry for the delay. Thanks for all this work. I think it is fine to announce this in the release notes, so that anyone who is committed to using nose can delay upgrading.

stevenbird:
Thanks again @dannysepler and @iliakur – really great to have this update.

Issue #2596, opened 2020-09-06
Potential Regex Denial of Service (ReDoS) in xmldocs


yetingli:
Type of Issue
Potential Regex Denial of Service (ReDoS)

Description
The vulnerable regular expression is located in

https://github.com/nltk/nltk/blob/ca357e5cdcdb137f40c45346bb8bfea618dd863f/nltk/corpus/reader/xmldocs.py#L225

The ReDOS vulnerability of the regex is mainly due to the sub-pattern `\s*/?\s*` and can be exploited with the following string
`"<"+" " * 5000`

I think you can limit the input length or modify the regex.

PR #2597, opened 2020-09-06
BUG: fix regex to avoid DoS

yetingli:
The regex which can match over long parts of strings but then be rejected by nonmatching at the end can be used for
denial-of-service. This addresses #2596


tomaarsen:
Hello! 
I've prepared a little script that showcases the execution times of the original regex and the regex proposed in this pull request.

### Script
```python
import re, timeit

inputs = {
    "dangerous": "<" + " " * 1000,
    "normal": "<      tag"
}

regexes = {
    "original":     re.compile("<\s*/?\s*([^\s>]+)"),
    "modified":     re.compile("<\s*(/\s*)?([^\s>]+)"),
    "new modified": re.compile("<\s*(?:/\s*)?([^\s>]+)")
}

for regex_name, regex in regexes.items():
    for input_name, input_str in inputs.items():
        # Time a lambda function that uses a regular expression to match an input string, 100 times
        execution_time = timeit.timeit(lambda: regex.match(input_str), number=100)
        print(f"{regex_name.title() + ' regex took': <24}: {execution_time:.8f}s on {input_name} input.")
    print()
```
### Outputs
```python
Original regex took     : 1.18097560s on dangerous input.
Original regex took     : 0.00007420s on normal input.

Modified regex took     : 0.00623780s on dangerous input.
Modified regex took     : 0.00007510s on normal input.

New Modified regex took : 0.00596490s on dangerous input.
New Modified regex took : 0.00007550s on normal input.
```
First of all, pay attention to the difference in execution time of the dangerous input for the original regex and the modified regex. This is exactly what this pull request and the corresponding issue #2596 is about. Currently, someone could easily multiply the amount of spaces in the dangerous string by another 1000, and the regex would never terminate.

So, the changes proposes here are **good**. Well done @yetingli.
I only have one small proposal.

### Proposal
The modified regex introduces another capturing group, which is never used. As a result, commit 2d353875eb4afec196be40c03af04701090553e6 was needed to increment the group index for the actual information we want the regex to capture. Because it is possible that there is some use of `_XML_TAG_NAME` that was not updated, and because it's more intuitive to not capture the whitespace, I modified the regex once again.
The only difference between the two is that I've turned the whitespace group into a [non-capturing group](https://docs.python.org/3.7/howto/regex.html#non-capturing-and-named-groups) by replacing `(/\s*)` with `(?:/\s*)`. The only changes are marked below.

Modified regex from this pull request:
```regex
<\s*(/\s*)?([^\s>]+)
```
Proposed regex:
```regex
<\s*(?:/\s*)?([^\s>]+)
     ^^
```

**My proposed regex will have identical outputs to the original regular expression, as opposed to the regex in this pull request, while having identical performance as the modified regex. As a result, with my proposed regex, *only* the regex needs to be updated, while the rest of the codebase can be left alone!**

### How to implement
* Revert commit 2d353875eb4afec196be40c03af04701090553e6, as the capturing group will disappear.
* Replace line with regex in [xmldocs.py](https://github.com/nltk/nltk/pull/2597/files#diff-081c9d4d1f0b9da40469a65ae63a3d25R226) with the following line:
```python
   _XML_TAG_NAME = re.compile("<\s*(?:/\s*)?([^\s>]+)")
```

yetingli:
Well done, I can't agree with you more!  Thanks @CubieDev :)
I will revert commit and thanks again.

tomaarsen:
These tests failed due to inconsistencies with the Python 3.5 UnigramTagger. I'm investigating it now.

Edit: It seems like a test in `nltk.tags` is inconsistent, and failed, despite your changes working properly. I have pushed a PR with more information.

stevenbird:
Thanks @yetingli, @CubieDev.

PR #2603, opened 2020-09-19
Stop storing `_hash` as an attribute

corynezin:
The `_hash` property of `nltk.grammar.Production` is hard coded and record at `__init__` time.  This may help performance a small amount, however it also produces the side effect that an unpickled `Production` will probably not have the same hash as an equal `Production` created in the scope of the current Python process.  For example consider running the two programs:

```python
import pickle
import nltk.grammar

a = nltk.grammar.Production('a', [])
with open('production.pkl', 'wb') as f:
    pickle.dump(a, f)
```

```python
import pickle
import nltk.grammar

a = nltk.grammar.Production('a', [])
with open('production.pkl', 'rb') as f:
    b = pickle.load(f)

print(hash(a), hash(b))
```

We would hope they would be the same, but they are not.  This particularly causes issues when storing dictionaries of Productions and saving them, as we will not find them when we search the dictionary upon loading [as in this SO question](https://stackoverflow.com/questions/63962245/tfidfvectorizer-model-loaded-from-joblib-file-only-works-when-trained-in-same-se/63963023?noredirect=1#comment113108360_63963023).

This PR solves the issue in a backward compatible way by making `_hash` a property.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -267,6 +267,7 @@
 - Ram Rachum <https://github.com/cool-RR>
 - Denali Molitor <https://github.com/dmmolitor>
 - Jacob Moorman <https://github.com/jdmoorman>
+- Cory Nezin <https://github.com/corynezin>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 ### Contributors to the Porter Stemmer
--- a/nltk/grammar.py
+++ b/nltk/grammar.py
@@ -290,7 +290,10 @@ class Production(object):
             )
         self._lhs = lhs
         self._rhs = tuple(rhs)
-        self._hash = hash((self._lhs, self._rhs))
+
+    @property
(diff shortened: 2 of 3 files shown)


iliakur:
Nice catch/fix!

Could you perhaps add some documentation for posterity? Could be in the form of a docstring (with tests?) or as an inline comment. I think I prefer docstring, but I'm not too familiar with the conventions of that module, so someone like Steven Bird or @alvations could say for sure.

corynezin:
I added a test.  It's a bit ugly but as elegant as I could make it.  If anyone has better suggestions please let me know!  I also changed the `_hash` property on a couple other classes in the same file, and decided that, since it is a private attribute, it should be safe to remove and we can just move the functionality entirely to `__hash__`.  Should I implement the change for classes in other files as well?

stevenbird:
Thanks @corynezin!

PR #2606, opened 2020-09-25
Ensure wheel and sdist contain proper license info

pombredanne:
Explicitly adds licensing related files and notices to a build sdist
or wheel to be comprehensive.

Also setup should fail on non-supported Python versions before
3.5.

This also explicitly adds licensing related files and notices to a
built wheel to be comprehensive.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,14 @@
-include LICENSE.txt NOTICE.txt INSTALL.txt README.txt MANIFEST.in
-include setup.py
-include nltk/test/*.doctest
-include nltk/VERSION
-recursive-include *.txt Makefile
+include *.md
+include *.txt
+include *.ini
+include setup.*
+include ChangeLog
+include Makefile
+include MANIFEST.in
+
+
+graft nltk
+graft tools
+graft web
+
 global-exclude *~
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,5 @@
+[metadata]
+license_files =
+    LICENSE.txt
+    AUTHORS.md
+    README.md
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,7 @@ natural language processing.  NLTK requires Python 3.5, 3.6, 3.7, or 3.8.""",
         "Topic :: Text Processing :: Linguistic",
     ],
     package_data={"nltk": ["test/*.doctest", "VERSION"]},
+    python_requires='>=3.5.*',
     install_requires=[
         "click",
         "joblib",


pombredanne:
Oh well, this latest code should not be buildable on Python 2 I guess? So the first commit is a bit moot... But the setup.py is missing a `    python_requires='>=3.5.*',` to avoid even building on Python 2 IMHO

tomaarsen:
@pombredanne  
> 
> Oh well, this latest code should not be buildable on Python 2 I guess? So the first commit is a bit moot... But the setup.py is missing a ` python_requires='>=3.5.*',` to avoid even building on Python 2 IMHO

I was about to mention this to you, that NLTK is only supported on Python 3.5+, but was unable to find an actual reference for this apart from how the Travis CI tests are only performed on Python 3.5 onwards. This kind of information should probably be available in the `setup.py` at the very least, and also present in the README.
Perhaps it is listed somewhere, but I wasn't able to find it quickly; not on a README, not on the website.

pombredanne:
@CubieDev I fixed the patch to remove the universal thing and add a proper `python_requires`

pombredanne:
@CubieDev Note that this is listed alright in the setup.py description 
https://github.com/nltk/nltk/blob/54221dec0bae2642d1642d182d8a381c88b86bd0/setup.py#L67 
and metadata 
https://github.com/nltk/nltk/blob/54221dec0bae2642d1642d182d8a381c88b86bd0/setup.py#L95-L98 
but that's not as good as an "executable" `python_requires` that I suggest adding in this PR

tomaarsen:
@pombredanne 
> 
> 
> @CubieDev Note that this is listed alright in the setup.py description
> 
> https://github.com/nltk/nltk/blob/54221dec0bae2642d1642d182d8a381c88b86bd0/setup.py#L67

Ah, yes. Well spotted. I still believe this information should be available in the README however. Though it will likely be harder to get a change in the README merged than a fix in the `setup.py`.

---

I went and did some digging. The [wheel documentation](https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file) states:
> By default, `wheel` conveniently includes files matching the following glob patterns in the `.dist-info` directory:
> - AUTHORS*
> - COPYING*
> - LICEN[CS]E*
> - NOTICE*

Which (if I understand this correctly) means the explicit addition of the `AUTHORS.md` and `LICENSE.txt` like in your commit are not needed. Whether `README.md` should be packaged as a license file, I do not know.

---

The addition of the `python_requires` seems good. Some documentation on it: [setuptools python_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires).

pombredanne:
@CubieDev LICENSE.txt is the only one included here and somehow AUTHORS.md is missed 
https://github.com/pypa/wheel/blob/9cdf053333f521a549d564b64f9d12a1a1d58a57/src/wheel/bdist_wheel.py#L408 should catch it but does not because when you build from the PyPI-released sdist the MANIFEST.in did not include several files (such as AUTHORS and NOTICE). In all cases the README.md contains extra license notice information that would be missing and are not available anywhere else: https://github.com/nltk/nltk/blob/develop/README.md#redistributing

And frankly it is better to be explicit than implicit so including README.md is best IMHO. Let me also fix the  MANIFEST.in

pombredanne:
@CubieDev not that the implicit inclusions are enabled only if there is no listed files under `license_files` per https://github.com/pypa/wheel/blob/9cdf053333f521a549d564b64f9d12a1a1d58a57/src/wheel/bdist_wheel.py#L408 hence why I added a comprehensive list of file that would have been included automatically otherwise

stevenbird:
Great, thanks @pombredanne!

PR #2610, opened 2020-10-08
Wordnet infinite looping for closure and tree

### LINKED ISSUE #2314: Infinite loop in Wordnet closure and tree
When I try to compute the hyponymic closure of `restrain.v.01` I get stuck in an infinite loop. Example code:
```python
>>> import nltk
>>> nltk.__version__
'3.4.3'
>>> ss = wn.synset('restrain.v.01')
>>> list(ss.closure(lambda s:s.hyponyms()))
```
And from here it just hangs. I guess that the reason is that the method is based on [`breadth_first`](http://www.nltk.org/api/nltk.html#nltk.util.breadth_first), which explicitly states in the documentation that is not designed to handle loops.

The same happens when running
```python
>>> list(ss.tree(lambda s:s.hyponyms()))
```

Which fails with a `RecursionError`. 

This seems to be closely related to #52, which is now closed, but the bug seems to be still present nonetheless.

### LINKED ISSUE #2565: WordNet: Loop in hyponyms hierarchy
Hello,

navigating the hierarchy of hyponyms I end up in a loop:

```
>>> synset=wn.synset('restrain.v.01'); print(synset)
Synset('restrain.v.01')
>>> synset=synset.hyponyms()[3]; print(synset)
Synset('inhibit.v.04')
>>> synset=synset.hyponyms()[2]; print(synset)
Synset('restrain.v.01')
>>> synset=synset.hyponyms()[3]; print(synset)
Synset('inhibit.v.04')
>>> synset=synset.hyponyms()[2]; print(synset)
Synset('restrain.v.01')
....
```

Is this meant to be?

nltk Version: 3.4.5 on Python 3.7

Giovanni

tomaarsen:
Fixes #2314, Fixes #2565.

Hello!

---
## Pull request overview
This PR is about solving infinite loops from wordnet. In particular:
- Remove cycles from `closure`
- Remove infinite loops from `tree`
- Added tests from #2314 and #2565 that failed previously.

@alvations I know you are working on moving Wordnet to a separate repository. I checked, and these issues exist there too. As such, you might be interested in these changes too. If you'd like, I can add a PR there too.

---

### How to reproduce:

```python
list(wn.synset('restrain.v.01').closure(lambda s:s.hyponyms()))
```
This will never terminate. 

### Why does this bug exist?
Let's look at the implementation of `closure`:
https://github.com/nltk/nltk/blob/385fb3f22031ebfa87b5c5571f32008f967264fa/nltk/corpus/reader/wordnet.py#L550-L574

From lines 571 and 572 it is clear that we want to avoid duplicates. However, the used `breadth_first` implementation says [here](https://github.com/nltk/nltk/blob/385fb3f22031ebfa87b5c5571f32008f967264fa/nltk/util.py#L215) that it does not check for cycles.
So, if a `rel` is passed to `closure` that is cyclic starting from the given `Synset`, then the loop using `breadth_first` on line 570 will loop forever. After all elements of the loop have been yielded once, lines 571 or 572 will prevent these repeats from being yielded, but the loop will go on. That is why there is non-termination in `closure`.

### A fix
I implemented a version of BFS very similar to `breadth_first`, but one that discards repeats. The implementation is pretty straightforward, using a deque. The deque is popped left, extracting the first synset and depth. If this synset isn't the current or some previous synset, we yield it. Then, if we have reached the desired max depth, we don't extend the deque. Otherwise, we do. The deque is extended by the outputs of `rel(synset)`, but for each output is checked whether it has been yielded before.

---

### How to reproduce:

```python
list(wn.synset('restrain.v.01').tree(lambda s:s.hyponyms()))
```
This will throw a RecursionError.

### Why does this bug exist?
Let's look at the implementation of `tree`:
https://github.com/nltk/nltk/blob/385fb3f22031ebfa87b5c5571f32008f967264fa/nltk/corpus/reader/wordnet.py#L756-L791

On line 788 we recurse using outputs of `rel(self)`. There is no check for cycles at all. So, we need to check whether synsets have already appeared higher up on the same branch. As @ekaf mentioned in his comment [here](https://github.com/nltk/nltk/issues/2314#issuecomment-702528702), we cannot simply prune a branch if a synset is found that has been placed anywhere in the tree, as we will truncate the non-cyclic result of the following line improperly:
```python
wn.synset('dog.n.01').tree(lambda s: s.hypernyms())
```
Should result in:
```python
    [Synset('dog.n.01'),
     [Synset('canine.n.02'),
      [Synset('carnivore.n.01'),
       [Synset('placental.n.01'),
        [Synset('mammal.n.01'),
         [Synset('vertebrate.n.01'),
          [Synset('chordate.n.01'),
           [Synset('animal.n.01'),
            [Synset('organism.n.01'),
             [Synset('living_thing.n.01'),
              [Synset('whole.n.02'),
               [Synset('object.n.01'),
                [Synset('physical_entity.n.01'),
                 [Synset('entity.n.01')]]]]]]]]]]]]],
     [Synset('domestic_animal.n.01'),
      [Synset('animal.n.01'),
       [Synset('organism.n.01'),
        [Synset('living_thing.n.01'),
         [Synset('whole.n.02'),
          [Synset('object.n.01'),
           [Synset('physical_entity.n.01'), [Synset('entity.n.01')]]]]]]]]]
```
but would then result in:
```python
    [Synset('dog.n.01'),
     [Synset('canine.n.02'),
      [Synset('carnivore.n.01'),
       [Synset('placental.n.01'),
        [Synset('mammal.n.01'),
         [Synset('vertebrate.n.01'),
          [Synset('chordate.n.01'),
           [Synset('animal.n.01'),
            [Synset('organism.n.01'),
             [Synset('living_thing.n.01'),
              [Synset('whole.n.02'),
               [Synset('object.n.01'),
                [Synset('physical_entity.n.01'),
                 [Synset('entity.n.01')]]]]]]]]]]]]],
     [Synset('domestic_animal.n.01'),
      [Synset('animal.n.01')]]]
```

### A fix
So, we need to prune only if a synset is found that has already been placed on the **same** branch in the tree. We can easily do this by adding another parameter `traverse`, which is a set of already traversed synsets on the same branch. Then, we can recurse with
```python
tree += [x.tree(rel, depth - 1, cut_mark, {*traversed, x}) for x in rel(self) if x not in traversed]
```
instead of 
```python
tree += [x.tree(rel, depth - 1, cut_mark) for x in rel(self)]
```
(Note for those unaware, `{*traversed, x}` creates a new set with the contents of `traversed` and `x`)

Because `traversed` is an implementation detail, I hid it by moving it to `_tree`, which only exists to be called by `tree` and by itself recursively.

---

## Tests
I added 5 tests, 3 of which failed previously, and 2 of which failed at some point during my development process.
The 3 tests are from comments in #2314, and one is a duplicate from #2565. These issues can be closed if this is merged and passes the tests.

---

Thanks to @ekaf for his many helpful comments and inspiration in #2314. 

- Tom Aarsen


ekaf:
In this PR, @CubieDev proposes to handle the errors entirely within the WordNet module.

As an alternative solution, PR #2612 proposes to handle the cycles more generally (in nltk/util.py), considering that the source of the problem with the current "closure" function is that it calls a "breadth_first" function imported from nltk.util, where the comment ("No need to check for cycles") is unfortunate, since there really is a need to check for cycles, in order to avoid eventual endless loops. Cyclicity also occurs in other graphs than Wordnet, so a general-purpose solution outside of the Wordnet module could be useful to check other graphs as well.

Also, due to the potential severity of this bug, it is preferrable to raise a warning than just to ignore the cycles silently.

Maybe the case is weaker for implementing an "acyclic_depth_first" search function that detects cycles in nltk/util.py,
and the proposal in this PR by @CubieDev to keep it within the WordNet module is a good alternative, although the coding style should be sufficiently explicit, in order to be able to report the eventual cycles.

tomaarsen:
The PR #2612 by @ekaf proposes an elegant solution to the issue mentioned in both of our PR's. Well done Eric.

Our implementations followed a different mindset. I took the existing implementation and reworked it to fix the issue, while Eric looked at the core issue, and moved the true implementation from these very specific `tree` and `closure` functions to very general functions in `nltk/util.py` that may be useful for other projects.
I've messed around with Eric's implementation a bit to see if it would pass the tests I wrote, and they seemed to. (Note, that was an older version of Eric's PR)

Regardless of which implementation is chosen, I would advise updating `nltk/test/wordnet.doctest` with (some of) the tests from this PR. This way, there are some more tests to check whether cycles are handled properly.

I'd like to hear @stevenbird's take on the situation.

ekaf:
Thanks @CubieDev! Concerning nltk/test/wordnet.doctest, it might be advantageous to split this out as a separate PR, since the same tests are relevant, no matter how the cyclicity problem is fixed. Please note that the discussion in issue #2314 just rebounded with a theoretical prediction that closure and tree also hang forever with a rather large number (over 5,000) of symmetric relation links. Among those, adding a few of the more salient examples into wordnet.doctest could be informative.

One among many particularly nasty also_sees() examples, is the following, where the tree grows to an unmanageable size, and the computation can only terminate in reasonable time by implementing global pruning. Meanwhile, it is possible to obtain a few levels of the tree, by limiting the "depth" parameter in the query:

```
>>> import nltk
>>> wn=nltk.corpus.wordnet
>>> from pprint import pprint
>>> pprint(wn.synset('certified.a.01').tree(lambda s:s.also_sees(),cut_mark='...',depth=4))
```

stevenbird:
Thanks @CubieDev and @ekaf! After taking too long (sorry), I decided to go with the generic solution. I'd be glad if you would like to add those further tests as you proposed.

PR #2612, opened 2020-10-11
Fixes issue #2314 (Infinite loop in Wordnet closure and tree)

ekaf:
This fixes issue #2314 (Infinite loop in Wordnet closure and tree)

## 1. The Problem

The origin of the bugs mentioned in issue #2314  is that the "nltk_data/corpora/wordnet/data.verb" database file contains two endless cycles: one hyponyms() cycle between 'restrain.v.01' and 'inhibit.v.04', and one topic_domains() cycle between 'computer.n.01' and 'computer_science.n.01'.

As a consequence of these cycles in the database, the closure() function in the WordNet module loops forever on the hyponyms of both 'restrain.v.01' and 'inhibit.v.04', and the topic_domains of both 'computer.n.01' and 'computer_science.n.01', these combinations yielding 4 failing calls:

```
>>> wn=nltk.corpus.wordnet
>>> print(list(wn.synset('restrain.v.01').closure(lambda s:s.hyponyms())))
>>> print(list(wn.synset('inhibit.v.04').closure(lambda s:s.hyponyms())))
>>> print(list(wn.synset('computer.n.01').closure(lambda s:s.topic_domains())))
>>> print(list(wn.synset('computer_science.n.01').closure(lambda s:s.topic_domains())))
```
Likewise, the tree() function throws a RecursionError for any of the 4 corresponding calls. 

```
>>> from pprint import pprint
>>> pprint(wn.synset('restrain.v.01').tree(lambda s:s.hyponyms()))
>>> pprint(wn.synset('inhibit.v.04').tree(lambda s:s.hyponyms()))
>>> pprint(wn.synset('computer.n.01').tree(lambda s:s.topic_domains()))
>>> pprint(wn.synset('computer_science.n.01').tree(lambda s:s.topic_domains()))
```

So 8 calls fail in total, in addition to a few thousand symmetric links from the verb_groups, attributes and also_sees relations.

One solution is to patch the database (as was already done in the past), but this does not prevent similar bugs in future db versions.

PR #2610 by @CubieDev proposes to handle the errors entirely within the WordNet module.

On the other hand, the present PR proposes to handle the problem more generally in nltk/util.py, considering that the source of the problem with the current "closure" function is that it calls a "breadth_first" function imported from nltk.util, where the comment ("No need to check for cycles") is unfortunate, since there really is a need to check for cycles, in order to avoid eventual endless loops. Cyclicity  also occurs in other graphs than Wordnet, so a general-purpose solution outside of the Wordnet module could be useful to check other graphs as well. 

Also, due to the potential severity of this bug, it is preferrable to raise a warning than just to ignore the cycles silently.

## 2. Fixing Closure

This fix implements an "acyclic_breadth_first" function in nltk/util.py, that detects and discards cycles. It is just a clone of the already existing "breadth_first" function, with a few additional lines to handle the cycles. Then wordnet.py needs only a simple wrapper to implement relation closure:

```
    def closure(self, rel, depth=-1):
        """
        Return the transitive closure of source under the rel
        relationship, breadth-first, discarding cycles:

        >>> computer = wn.synset('computer.n.01')
        >>> topic = lambda s:s.topic_domains()
        >>> print(list(computer.closure(topic)))
        [Synset('computer_science.n.01')]

        UserWarning: Discarded redundant search ([(Synset('computer.n.01'), 2)])


        Include redundant pathes (but only once), avoiding duplicate searches
        (from 'animal.n.01' to 'entity.n.01'):

        >>> dog = wn.synset('dog.n.01')
        >>> hyp = lambda s:s.hypernyms()
        >>> print(list(dog.closure(hyp)))
        [Synset('canine.n.02'), Synset('domestic_animal.n.01'), Synset('carnivore.n.01'),
        Synset('animal.n.01'), Synset('placental.n.01'), Synset('organism.n.01'),
        Synset('mammal.n.01'), Synset('living_thing.n.01'), Synset('vertebrate.n.01'),
        Synset('whole.n.02'), Synset('chordate.n.01'), Synset('object.n.01'),
        Synset('physical_entity.n.01'), Synset('entity.n.01')]

        UserWarning: Discarded redundant search ([(Synset('animal.n.01'), 7)])
        """

        from nltk.util import acyclic_breadth_first
        for synset in acyclic_breadth_first(self, rel, depth):
            if synset != self:
                yield synset

```

The current implementation of Wordnet closure does not include the caller synset in the output (unlike the 'tree' fumction), and this behaviour is maintained here.

Please note the  duplicate path (from 'animal.n.01' to 'entity.n.01') in the hypernyms of 'dog.n.01', which has not received much attention previously. Pruning this path indiscriminately would be incorrect, since it occurs legitimately in two different branches of the tree, so we want to keep both instances of it in the output of "tree" (but search it just once in "closure").

## 3. Fixing Tree

Maybe the case is weaker for implementing an "acyclic_depth_first" search function that detects cycles in nltk/util.py,
and the proposal in PR #2610 by @CubieDev to keep it within the WordNet module is a good alternative, although the coding style should be sufficiently explicit, in order to be able to report the eventual cycles, instead of ignoring them silently.

In the present PR, the Wordnet "tree" wrapper can simply be:

```
    def tree(self, rel, depth=-1, cut_mark=None):
        """
        Return the full relation tree, including self,
        discarding cycles:

        >>> from pprint import pprint
        >>> computer = wn.synset('computer.n.01')
        >>> topic = lambda s:s.topic_domains()
        >>> pprint(computer.tree(topic))
        [Synset('computer.n.01'), [Synset('computer_science.n.01')]]

        UserWarning: Discarded redundant search ([(Synset('computer.n.01'), -3)])


        But keep duplicate branches (from 'animal.n.01' to 'entity.n.01'):

        >>> dog = wn.synset('dog.n.01')
        >>> hyp = lambda s:s.hypernyms()
        >>> pprint(dog.tree(hyp))
        [Synset('dog.n.01'),
         [Synset('canine.n.02'),
          [Synset('carnivore.n.01'),
           [Synset('placental.n.01'),
            [Synset('mammal.n.01'),
             [Synset('vertebrate.n.01'),
              [Synset('chordate.n.01'),
               [Synset('animal.n.01'),
                [Synset('organism.n.01'),
                 [Synset('living_thing.n.01'),
                  [Synset('whole.n.02'),
                   [Synset('object.n.01'),
                    [Synset('physical_entity.n.01'),
                     [Synset('entity.n.01')]]]]]]]]]]]]],
         [Synset('domestic_animal.n.01'),
          [Synset('animal.n.01'),
           [Synset('organism.n.01'),
            [Synset('living_thing.n.01'),
             [Synset('whole.n.02'),
              [Synset('object.n.01'),
               [Synset('physical_entity.n.01'), [Synset('entity.n.01')]]]]]]]]]
        """

        from nltk.util import acyclic_branches_depth_first
        return acyclic_branches_depth_first(self, rel, depth, cut_mark)
```

The new function acyclic_branches_depth_first() traverses the nodes of a tree in depth-first order, discarding eventual cycles within the same branch, but keeping duplicate pathes in different branches. The branches are kept separate by recursing with a different "traversed" set for each child.

## 4. Tree-size  Limitation

But the worst problem by far, is the also_sees() relation (2692 links between Synsets), which has both 2490 symmetric links, and 202 non-symmetric links that involve Synsets that also have at least a symmetric link. This combination is particularly toxic, because it produces a huge Strongly Connected Component (SCC), counting 758 synsets, where any member of the SCC is transitively connected to all other members. I am not sure how to estimate the size of such a tree, but only guessing that it could approximately be the average number of children per node, raised to the 758 power. Then the full also_sees() tree, starting at any member of this SCC, for ex. 'concrete.a.01', could not fit on a single computer, no matter how efficiently trees are represented or computed.

So the proposed approach obviously cannot handle such harder cases, where the trees grow to an unmanageable size, and  the computation only terminates in reasonable time if the "depth" parameter is limited:

```
>>> import nltk
>>> wn=nltk.corpus.wordnet
>>> from pprint import pprint
>>> pprint(wn.synset('certified.a.01').tree(lambda s:s.also_sees(),cut_mark='...',depth=4))
[Synset('certified.a.01'),
 [Synset('authorized.a.01'),
  [Synset('lawful.a.01'),
   [Synset('legal.a.01'),
    "Cycle(Synset('lawful.a.01'),0,...)",
    [Synset('legitimate.a.01'), '...']],
   [Synset('straight.a.06'),
    [Synset('honest.a.01'), '...'],
    "Cycle(Synset('lawful.a.01'),0,...)"]],
  [Synset('legitimate.a.01'),
   "Cycle(Synset('authorized.a.01'),1,...)",
   [Synset('legal.a.01'),
    [Synset('lawful.a.01'), '...'],
    "Cycle(Synset('legitimate.a.01'),0,...)"],
   [Synset('valid.a.01'),
    "Cycle(Synset('legitimate.a.01'),0,...)",
    [Synset('reasonable.a.01'), '...']]],
  [Synset('official.a.01'), "Cycle(Synset('authorized.a.01'),1,...)"]],
 [Synset('documented.a.01')]]
```

To handle such cases, global pruning is needed, which is now implemented in a new "acyclic_tree" function. This uses depth first search to compute an arbitrary spanning tree. Alternatively, breadth-first search could compute a minimum spanning tree (MST), which may not necessarily be worthwile, since it would require some overhead in order to assemble a tree structure. The proposed solution can cope with all cycles, but has the drawback that it truncates some legitimate duplicate pathes:

```
>>> import nltk
>>> wn=nltk.corpus.wordnet
>>> from pprint import pprint
>>> pprint(wn.synset('dog.n.01').acyclic_tree(lambda s:s.hypernyms(),cut_mark='...'))
[Synset('dog.n.01'),
 [Synset('canine.n.02'),
  [Synset('carnivore.n.01'),
   [Synset('placental.n.01'),
    [Synset('mammal.n.01'),
     [Synset('vertebrate.n.01'),
      [Synset('chordate.n.01'),
       [Synset('animal.n.01'),
        [Synset('organism.n.01'),
         [Synset('living_thing.n.01'),
          [Synset('whole.n.02'),
           [Synset('object.n.01'),
            [Synset('physical_entity.n.01'),
             [Synset('entity.n.01')]]]]]]]]]]]]],
 [Synset('domestic_animal.n.01'), "Cycle(Synset('animal.n.01'),-3,...)"]]
```

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -91,6 +91,7 @@
 - Edward Ivanovic
 - Thomas Jakobsen
 - Nick Johnson
+- Eric Kafe
 - Piotr Kasprzyk
 - Angelos Katharopoulos
 - Sudharshan Kaushik
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -583,6 +583,9 @@ class Synset(_WordNetObject):
                 yield synset
 
 
+    from nltk.util import acyclic_depth_first as acyclic_tree
+
+
     def tree(self, rel, depth=-1, cut_mark=None):
         """
         Return the full relation tree, including self,
@@ -626,8 +629,8 @@ class Synset(_WordNetObject):
                [Synset('physical_entity.n.01'), [Synset('entity.n.01')]]]]]]]]]
         """
 
-        from nltk.util import acyclic_depth_first
-        return acyclic_depth_first(self, rel, depth, cut_mark)
+        from nltk.util import acyclic_branches_depth_first
+        return acyclic_branches_depth_first(self, rel, depth, cut_mark)
 
 
     def hypernym_paths(self):
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -559,22 +559,23 @@ class Synset(_WordNetObject):
         >>> print(list(computer.closure(topic)))
 
         UserWarning: Discarded redundant search ([(Synset('computer.n.01'), 2)])
+
         [Synset('computer_science.n.01')]
 
 
+        However, still search pathes once (from 'animal.n.01' to 'entity.n.01'):
+
         >>> dog = wn.synset('dog.n.01')
-        >>> topic = lambda s:s.hypernyms()
-        >>> pprint(computer.tree(hyp))
-        >>> print(list(dog.tree(hyp)))
+        >>> hyp = lambda s:s.hypernyms()
+        >>> print(list(dog.closure(hyp)))
 
-        [Synset('canine.n.02'), Synset('domestic_animal.n.01'),
-        Synset('carnivore.n.01'), Synset('animal.n.01'),
-        Synset('placental.n.01'), Synset('organism.n.01'),
-        Synset('mammal.n.01'), Synset('living_thing.n.01'),
-        Synset('vertebrate.n.01'), Synset('whole.n.02'),
-        Synset('chordate.n.01'), Synset('object.n.01'),
-        Synset('physical_entity.n.01'), Synset('entity.n.01')]
+        UserWarning: Discarded redundant search ([(Synset('animal.n.01'), 7)])
 
+        [Synset('domestic_animal.n.01'), Synset('animal.n.01'), Synset('organism.n.01'),
+        Synset('living_thing.n.01'), Synset('whole.n.02'), Synset('object.n.01'),
+        Synset('physical_entity.n.01'), Synset('entity.n.01'), Synset('canine.n.02'),
+        Synset('carnivore.n.01'), Synset('placental.n.01'), Synset('mammal.n.01'),
+        Synset('vertebrate.n.01'), Synset('chordate.n.01')]
         """
 
         from nltk.util import acyclic_breadth_first
@@ -593,14 +594,14 @@ class Synset(_WordNetObject):
         >>> pprint(computer.tree(topic))
 
         UserWarning: Discarded redundant search ([(Synset('computer.n.01'), -3)])
+
         [Synset('computer.n.01'), [Synset('computer_science.n.01')]]
 
 
         But keep duplicate branches (from 'animal.n.01' to 'entity.n.01'):
 
         >>> dog = wn.synset('dog.n.01')
-        >>> topic = lambda s:s.hypernyms()
-        >>> pprint(computer.tree(hyp))
+        >>> hyp = lambda s:s.hypernyms()
         >>> pprint(dog.tree(hyp))
 
         [Synset('dog.n.01'),
@@ -624,11 +625,10 @@ class Synset(_WordNetObject):
              [Synset('whole.n.02'),
               [Synset('object.n.01'),
                [Synset('physical_entity.n.01'), [Synset('entity.n.01')]]]]]]]]]
-
         """
 
         from nltk.util import acyclic_depth_first
-        return acyclic_depth_first(self, rel, depth)
+        return acyclic_depth_first(self, rel, depth, cut_mark)
 
 
     def hypernym_paths(self):
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -557,10 +557,9 @@ class Synset(_WordNetObject):
         >>> computer = wn.synset('computer.n.01')
         >>> topic = lambda s:s.topic_domains()
         >>> print(list(computer.closure(topic)))
+        [Synset('computer_science.n.01')]
 
         UserWarning: Discarded redundant search ([(Synset('computer.n.01'), 2)])
-
-        [Synset('computer_science.n.01')]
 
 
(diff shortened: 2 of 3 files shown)


ekaf:
One advantage of having generic transitive search functions in a separate module is, that it becomes possible to compute the closure of transitive lexical functions (relations between lemmas), even though in the wordnet.py module, closure and tree are only defined for relations between synsets.

For example:

```
>>> from nltk.util import acyclic_branches_depth_first as dfs_tree
>>> from pprint import pprint
>>> pprint(dfs_tree(wn.lemma('socioeconomically.r.01.socioeconomically'),lambda s:s.pertainyms()))
[Lemma('socioeconomically.r.01.socioeconomically'),
 [Lemma('socioeconomic.a.01.socioeconomic'),
  [Lemma('economic.a.02.economic'), [Lemma('economics.n.01.economics')]]]]
```

Alternatively, the new functions could be imported inside the Lemma class, in order to provide additional functionality for lemmas.

stevenbird:
Thanks @ekaf. Sorry for the delay.

PR #2619, opened 2020-10-26
Fix bug in meteor

### LINKED ISSUE #2618: Meteor score does not use `preprocess` input argument
Hello,

The `meteor_score` function defined [here](https://github.com/nltk/nltk/blob/385fb3f22031ebfa87b5c5571f32008f967264fa/nltk/translate/meteor_score.py#L421) takes the argument `preprocess` as an input, but does not pass it to the `single_meteor_score` function call. As a result, the function always uses the default `preprocess` function (`str.lower`).

AkshitaB:
Fixes https://github.com/nltk/nltk/issues/2618

Updated the `meteor_score` function to pass the `preprocess` argument to the `single_meteor_score` function call.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -270,6 +270,7 @@
 - Jacob Moorman <https://github.com/jdmoorman>
 - Cory Nezin <https://github.com/corynezin>
 - Matt Chaput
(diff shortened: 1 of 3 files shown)


stevenbird:
Thanks @AkshitaB

PR #2620, opened 2020-11-02
POS tagger throws IndexError when passed empty string

### LINKED ISSUE #2483: POS tagger crashes when text starts with a white space.
Dear NLTK team,

the POS parsing crashes when the text starts with one (or several) white space(s). The problem occurs in the normalize() call.

/usr/local/lib/python3.6/dist-packages/nltk/tag/perceptron.py in normalize(self, word)
    238         elif word.isdigit() and len(word) == 4:
    239             return '!YEAR'
--> 240         elif word[0].isdigit():
    241             return '!DIGITS'
    242         else:

IndexError: string index out of range

I guess that testing the length() of the word could solve the issue.
lstriping the input text is not a correct one, as for many reasons, the text has to be preserved untouched by ntlk.

Best regards

Jerome

tomaarsen:
Fixes #2483.

Hello!

---
## Pull request overview
This PR is about removing an IndexError that is thrown when an empty string is passed to either:
- `pos_tag()` in `nltk`
- `PerceptronTagger().tag()` in `nltk.tag.perceptron`
- `PerceptronTagger().train()` in `nltk.tag.perceptron`
- `PerceptronTagger().normalize()` in `nltk.tag.perceptron`

This error is reported in #2483.

---

### Why does this bug exist?
The first three functions (and methods) all call `PerceptronTagger().normalize()` for each "word" they are given. However, line 280 in that function checks the first character of that word:
https://github.com/nltk/nltk/blob/385fb3f22031ebfa87b5c5571f32008f967264fa/nltk/tag/perceptron.py#L267-L283
This will throw an IndexError when the passed word is an empty string.

### How to reproduce?
There are multiple functions affected, listed from most to least important. I'll provide ways to reproduce the issue for only the two most used functions:
```python
from nltk import pos_tag
words = ['', 'is', 'a', 'beautiful', 'day']
pos_tag(words)
```
```python
from nltk.tag.perceptron import PerceptronTagger
words = ['', 'is', 'a', 'beautiful', 'day']
tagger = PerceptronTagger()
tagger.tag(words)
```
For both of these, the last line will throw an IndexError.

### A fix
The fix, as proposed in f3ae03bcdfc3e57250bac511e3667d5ce421ce27 is to convert `elif word[0].isdigit():` to `elif word and word[0].isdigit()`, which causes the latter part to only be evaluated if `word` is non-empty.
This is only part of the fix however, as `pos_tag` and `PerceptionTagger().tag()` will still throw IndexErrors when extracting features from the word for use in the tagging. In particular, when getting the first character of the word. This is also resolved in the aforementioned commit by using an empty string if the word is an empty string.

### The consequences
With the pos_tagger that is shipped in nltk, an empty string is tagged as a `NN` with roughly `0.99` confidence. To me, this is way preferable over throwing an IndexError. If this really is undesired, an exception should be thrown explaining that empty strings should not be used.

### Tests
I added a test which previously failed, testing the first program listed in the "How to reproduce?" section.

---

### Notes
I also:
* Fixed a typo in a comment.
* Accidentally committed to the wrong branch in `CubieDev/nltk` and reverted afterward. As a result this PR has some commits that cancel each other out. This doesn't affect anything.

---

- Tom Aarsen

### DIFF
--- a/nltk/tag/__init__.py
+++ b/nltk/tag/__init__.py
@@ -108,7 +108,7 @@ def _get_tagger(lang=None):
 
 
 def _pos_tag(tokens, tagset=None, tagger=None, lang=None):
-    # Currently only supoorts English and Russian.
+    # Currently only supports English and Russian.
     if lang not in ["eng", "rus"]:
         raise NotImplementedError(
             "Currently, NLTK pos_tag only supports English and Russian "
--- a/nltk/tag/perceptron.py
+++ b/nltk/tag/perceptron.py
@@ -277,7 +277,7 @@ class PerceptronTagger(TaggerI):
             return "!HYPHEN"
         elif word.isdigit() and len(word) == 4:
             return "!YEAR"
-        elif word and word[0].isdigit():
+        elif word[0].isdigit():
             return "!DIGITS"
         else:
             return word.lower()
@@ -296,7 +296,7 @@ class PerceptronTagger(TaggerI):
         # It's useful to have a constant feature, which acts sort of like a prior
         add("bias")
         add("i suffix", word[-3:])
-        add("i pref1", word[0] if word else "")
+        add("i pref1", word[0])
         add("i-1 tag", prev)
         add("i-2 tag", prev2)
         add("i tag+i-2 tag", prev, prev2)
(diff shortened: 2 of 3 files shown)


stevenbird:
Thanks @CubieDev

Issue #2420, opened 2019-09-25
WordNet sense keys mismatch


etienne-v:
Hi,

I'm having problems with retrieving the correct definitions from the nltk wordnet corpus for a given sense key.  Please see [this](https://github.com/getalp/UFSAC/issues/2) related issue from the UFSAC repo.

The current ntlk WordNet version is 3.0:
```
import nltk
nltk.download('wordnet')
from nltk.corpus import wordnet as wn

print(wn.get_version())
```
gives:
```
3.0
```

Looking at the official WordNet 3.0 database (http://wordnetcode.princeton.edu/3.0/WordNet-3.0.tar.gz), the key "drive%1:04:03::" has a synset ID of 00572489 (according to the `dict/index.sense` file), and the definition related to this synset ID is "hitting a golf ball off of a tee with a driver" (according to the `dict/data.noun` file).

The key associated with the above definition from the official WordNet online search is also "drive%1:04:03::" (along with "driving%1:04:03::"), even though the online search version is 3.1.

Firstly, using the key "drive%1:04:03::" to look up the definition in ntlk as follows
```
print(wn.synset_from_sense_key("drive%1:04:03::"))
print(wn.synset_from_sense_key("drive%1:04:03::").definition())
print(wn.synset_from_sense_key("drive%1:04:03::").offset())
```
gives:
```
Synset('campaign.n.02')
a series of actions advancing a principle or tending toward a particular end
798245
```
This definition and ID doesn't correspond to either the WordNet online search or the database.  However, querying nltk with the key "drive%1:04:06::" (replacing 03 with 06)
```
print(wn.synset_from_sense_key("drive%1:04:06::"))
print(wn.synset_from_sense_key("drive%1:04:06::").definition())
print(wn.synset_from_sense_key("drive%1:04:06::").offset())
```
gives the expected definition and ID:
```
Synset('drive.n.06')
hitting a golf ball off of a tee with a driver
572489
```
For this key, why doesn't ntlk give the definition as recorded in the database and the online search?

Also, when using the key "driving%1:04:03::" (referring to the second synonym for that sense), to obtain the synset with
```
print(wn.synset_from_sense_key("driving%1:04:03::"))
```
I get the following error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
~/anaconda3/lib/python3.7/site-packages/nltk/corpus/reader/wordnet.py in synset(self, name)
   1333         try:
[...]

Issue #2171, opened 2018-11-08
Wordnet: synset_from_sense_key returns the wrong Synset


hlumin:
https://github.com/nltk/nltk/blob/15e7f74874cf562a44747957f006cf30869a15cd/nltk/corpus/reader/wordnet.py#L1532

**Issue**: synset_from_sense_key does not always return the correct Synset.

**Example**:

```python
sense_key = 'deal%1:10:00::'

synset_from_lemma = wn.lemma_from_key(sense_key).synset()
synset_from_key = wn.synset_from_sense_key(sense_key)

print(synset_from_lemma)
print(synset_from_key)
```
outputs

```
Synset('bargain.n.01')
Synset('deal.n.09')
```
While the same Synset should be returned by both

**Source of the issue**: It seems synset_from_sense_key directly concatenates the lemma name contained in the sense key to the newly created Synset ID, while the correct synset could be using an ID with a different main lemma.

begab:
As an additional note, the above issue can also result in a `WordNetError` being thrown in certain cases. There are 5 example sense keys to reproduce the error:
`problematic_sense_keys = 'remain%2:42:05:: put%2:31:13:: minister%2:41:12:: long%2:37:02:: aim%2:32:09::'.split()`

zhengyaogu:
Is this bug fixed? I am trying to use this method for my project.

HSalaberri:
Looks like this issue persists:

**Example**

`sense_key = 'leader%1:18:00::::'

synset_from_lemma = wn.lemma_from_key(sense_key).synset()
synset_from_key = wn.synset_from_sense_key(sense_key)

print(synset_from_lemma)
print(synset_from_key)`

**Output**

`Synset('leader.n.01')
Synset('drawing_card.n.02')`

Both synsets should be the same, instead, `synset_from_sense_key` returns the synset corresponding to the second sense of leader as a noun, which has sense key `leader%1:06:00::` not  `leader%1:18:00::::`.

PR #2621, opened 2020-11-09
Fixes Issue #2420 (WordNet sense keys mismatch)

ekaf:
Fixes Issues #2420 (WordNet sense keys mismatch) and #2171 (Wordnet: synset_from_sense_key returns the wrong Synset)

## The bug
For 25% of all sense keys, the synset_from_sense_key() function returns a wrong synset, or fails with an IndexError:

```
import nltk
print("NLTK v. %s" % nltk.__version__)
print("WordNet v. %s" % wn.get_version())

n = 0
wrong_ss = 0
no_ss = 0

for ss in wn.all_synsets():
    for lm in ss._lemmas:
        n += 1
        try:
            if ss != wn.synset_from_sense_key(lm._key):
                wrong_ss += 1
        except:
            no_ss += 1
#            print('%s not found' % lm._key)

errors = wrong_ss + no_ss
print("%s/%s errors (%s/100): %s wrong, %s not found" % 
   (errors, n, round(100.0*errors/n, 3), wrong_ss, no_ss))
```

```
NLTK v. 3.5
WordNet v. 3.0
51751/206978 errors (25.003/100): 50416 wrong, 1335 not found
```

## Why does this bug occur?

The synset_from_sense_key() function in the wordnet.py module is severely flawed, because it incorrectly uses the "lex_id" extracted from the sense key, to construct synset_ids that turn out to be bogus, since an index into the list of offsets for the corresponding lemma is expected instead. 

At first glance, it seems remarkable that 75% of the results are correct by pure chance, when the lex_id and the expected index have nothing to do with each other. But this result is actually not impressive at all, since 80% of the words in WordNet 3.0 are monosemous, i.e. they occur only in one synset, in which case it should not be possible to pick a wrong synset.

## Reproducing the bug

This retrieves an incorrect synset (Synset('campaign.n.02'), instead of the correct Synset('drive.n.06')):

```
print(wn.synset_from_sense_key("drive%1:04:03::"))
print(wn.synset_from_sense_key("drive%1:04:03::").definition())
print(wn.synset_from_sense_key("drive%1:04:03::").offset())
```

This raises an "IndexError: list index out of range", because lex_ids are neither consecutive nor necessarily numbered starting at 0 (their default value):

```
print(wn.synset_from_sense_key("driving%1:04:03::"))
```

Thanks to @etienne-v for raising this issue

## The fix

Rely on the lemma_from_key() function to provide the correct synset:

```
    def synset_from_sense_key(self, sense_key):
        """
        >>> import nltk
        >>> print(wn.synset_from_sense_key("drive%1:04:03::"))
        Synset('drive.n.06')

        >>> print(wn.synset_from_sense_key("driving%1:04:03::"))
        Synset('drive.n.06')
        """
        return self.lemma_from_key(sense_key).synset()
```

## The effect

After applying this PR, the number of wrong synsets drops to zero, but 8 ill-formed sense keys remain, pertaining to a different well-known problem (upper-case and lower-case forms of the same lemma are expected to share a common sense key):

```
NLTK v. 3.5
WordNet v. 3.0
ddc%1:06:01:: not found
ddi%1:06:01:: not found
earth%1:15:01:: not found
earth%1:17:02:: not found
moon%1:17:03:: not found
sun%1:17:02:: not found
kb%1:23:01:: not found
kb%1:23:03:: not found
8/206978 errors (0.004/100): 0 wrong, 8 not found
```

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1532,26 +1532,17 @@ class WordNetCorpusReader(CorpusReader):
                      Only used if sense is in an adjective satellite synset
         head_id:     uniquely identifies sense in a lexicographer file when paired with head_word
                      Only used if head_word is present (2 digit int)
+
+        >>> import nltk
+        >>> from nltk.corpus import wordnet as wn
+        >>> print(wn.synset_from_sense_key("drive%1:04:03::"))
+        Synset('drive.n.06')
+
+        >>> print(wn.synset_from_sense_key("driving%1:04:03::"))
+        Synset('drive.n.06')
         """
-        sense_key_regex = re.compile(r"(.*)\%(.*):(.*):(.*):(.*):(.*)")
-        synset_types = {1: NOUN, 2: VERB, 3: ADJ, 4: ADV, 5: ADJ_SAT}
-        lemma, ss_type, _, lex_id, _, _ = sense_key_regex.match(sense_key).groups()
+        return self.lemma_from_key(sense_key).synset()
 
-        # check that information extracted from sense_key is valid
-        error = None
-        if not lemma:
-            error = "lemma"
-        elif int(ss_type) not in synset_types:
-            error = "ss_type"
-        elif int(lex_id) < 0 or int(lex_id) > 99:
-            error = "lex_id"
-        if error:
-            raise WordNetError(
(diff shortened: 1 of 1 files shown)


stevenbird:
My goodness. Thank you @ekaf

Issue #2624, opened 2020-11-15
nltk.corpus.wordnet.synset_from_pos_and_offset broken in NLTK 3.5?


ghost:
Not an NLTK expert here, but this seems wrong. On a fresh install of NLTK 3.5 (Python 3.8.5) the following

```
from nltk.corpus import wordnet as wn
wn.synset_from_pos_and_offset(wn.NOUN, 1)
```

gives

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/.local/lib/python3.8/site-packages/nltk/corpus/reader/wordnet.py in _synset_from_pos_and_line(self, pos, data_file_line)
   1395             # parse out the definitions and examples from the gloss
-> 1396             columns_str, gloss = data_file_line.strip().split("|")
   1397             definition = re.sub(r"[\"].*?[\"]", "", gloss).strip()

ValueError: not enough values to unpack (expected 2, got 1)

During handling of the above exception, another exception occurred:

WordNetError                              Traceback (most recent call last)
<ipython-input-66-a20d7aaa0d8c> in <module>
----> 1 wn.synset_from_pos_and_offset(wn.NOUN, 1)

~/.local/lib/python3.8/site-packages/nltk/corpus/reader/wordnet.py in synset_from_pos_and_offset(self, pos, offset)
   1372         data_file.seek(offset)
   1373         data_file_line = data_file.readline()
-> 1374         synset = self._synset_from_pos_and_line(pos, data_file_line)
   1375         assert synset._offset == offset
   1376         self._synset_offset_cache[pos][offset] = synset

~/.local/lib/python3.8/site-packages/nltk/corpus/reader/wordnet.py in _synset_from_pos_and_line(self, pos, data_file_line)
   1478         # raise a more informative error with line text
   1479         except ValueError as e:
-> 1480             raise WordNetError("line %r: %s" % (data_file_line, e))
   1481 
   1482         # set sense keys for Lemma objects - note that this has to be

WordNetError: line ' 1 This software and database is being provided to you, the LICENSEE, by  \n': not enough values to unpack (expected 2, got 1)
```

When looking at the file that is accessed internally there seems to be some license info before the actual data:

```
wn.open("data.noun").read()[:5000].split("\n")
```

```
['  1 This software and database is being provided to you, the LICENSEE, by  ',
 '  2 Princeton University under the following license.  By obtaining, using  ',
 '  3 and/or copying this software and database, you agree that you have  ',
[...]

PR #2625, opened 2020-11-19
Fixes issue #2624 (WordNet offset validation)

ekaf:
In issue #2624 @poke1024 asks: "nltk.corpus.wordnet.synset_from_pos_and_offset broken in NLTK 3.5?"

The function is not broken, but the problem is that there is no check of the offset parameter, so when users pass an offset that does not correspond to an existing WordNet synset, a ValueError is (correctly) raised.

However, we can easily do better by checking that the offset parameter is an actual synset offset, since valid offsets equal the 8-digit 0-padded integer found at the start of the corresponding line in the WordNet db, and this allows to eventually return a more helpful error message.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -10,6 +10,8 @@
 #         Sim Wei Ying Geraldine
 #         Soe Lynn
 #         Francis Bond <bond@ieee.org>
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @ekaf (and @poke1024)

PR #2629, opened 2020-11-27
Fully remove nose in favor of pytest

dannysepler:
The most immediate followup of https://github.com/nltk/nltk/pull/2595 is to stop using nose in the repo entirely.

Also cleaned up:
- A few pytest warnings (there's a few more)
- A few random test code things (like unneeded / unreachable code)


nltk/parse/stanford.py
```diff
@@ -473,16 +473,6 @@ def _make_tree(self, result):
         return DependencyGraph(result, top_relation_label="ROOT")
 
 
-@skip("doctests from nltk.parse.stanford are skipped because it's deprecated")
+@pytest.mark.skip(reason="doctests from nltk.parse.stanford are skipped because it's deprecated")
 def setup_module(module):
-    from nose import SkipTest
```
dannysepler, line 478: this looks like unreachable code, right? since we marked the whole thing as skip

nltk/test/unit/test_seekable_unicode_stream_reader.py
```diff
@@ -96,41 +103,11 @@ def test_reader():
                 pass
 
 
-# nose shows the whole string arguments in a verbose mode; this is annoying,
```
dannysepler, line 99: this seems to no longer be true with pytest, so i added this back with the other strings

nltk/test/unit/test_seekable_unicode_stream_reader.py
```diff
@@ -96,41 +103,11 @@ def test_reader():
...
-            _check(encoding)
-
-        except UnicodeEncodeError:
-            pass
-
-
 def test_reader_stream_is_closed():
     reader = SeekableUnicodeStreamReader(BytesIO(b''), 'ascii')
-    assert reader.stream.closed is False
+    assert reader.stream.closed == False
```
iliakur: Actually `is` is preferred here. `False` and `True` are singletons so you can (and should, for better readability) check them by identity.

Another option is to say this:

```python
assert not reader.stream.closed
```

iliakur (changes requested):

iliakur:
I'm honestly not sure about all those skipped tests as well. I'm tempted to suggest to delete them, but since none of them are in the scope of my regular contributions/maintenance work I think it's best you try to ping the folks who wrote them (use `git blame` to find that out).

nltk/test/unit/test_seekable_unicode_stream_reader.py
dannysepler: didn't know about them being singletons. looks like the [python style guide](https://www.python.org/dev/peps/pep-0008/) suggests doing neither `==` nor `is`. Wow, TIL!

dannysepler:
re: skipped tests, totally! i spun out a new issue for CoreNLP. i'd rather not tie this diff into that conversation. (after all, maybe that test code is helpful and should stay)

stevenbird:
Thanks @dannysepler, @iliakur! Nice milestone 👍

pombredanne:
This change means that pytest is an undeclared dependency of nltk and this seems to break things with the new 3.6. release
https://github.com/nltk/nltk/releases/tag/3.6 breaks any https://github.com/nexB/scancode-toolkit pip installation

PR #2630, opened 2020-11-28
Minor improvements to the CI build

dannysepler:
One of @iliakur's comments on my original diff, which I really liked.

The alternative is to add this to the `pytest.ini` so that tests always run parallel -- but I don't like the experience of `pytest path/to/file.py` still running parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,12 @@ sudo: false
 language: python
 
 cache:
-  apt: true
+  pip: true
   directories:
   - $HOME/.cache/pip
   - $HOME/.ccache
   - $HOME/.pip-cache
   - $HOME/third
-  - $HOME/nltk_data
 
 dist: xenial
 
--- a/tox.ini
+++ b/tox.ini
@@ -46,16 +46,16 @@ deps =
     pytest-mock
     twython
 
-commands =
-    pytest
-
 [testenv:py35-nodeps]
 basepython = python3.5
 deps =
     nose >= 1.2.1
     pytest
     pytest-mock
-commands = pytest
+commands =
+    ; "nodeps" testenvs require a new commands list that doesn't
+    ; install scipy and scikit-learn
+    pytest
 
 [testenv:py36-nodeps]
 basepython = python3.6
--- a/tox.ini
+++ b/tox.ini
@@ -28,13 +28,12 @@ deps =
     python-crfsuite
     regex
 
-changedir = nltk/test
 commands =
     ; scipy and scikit-learn requires numpy even to run setup.py so
     ; they can't be installed in one command
     pip install scipy scikit-learn
 
-    ; pytest --cov=nltk --cov-report html:{envdir}/docs nltk/test/
+    ; pytest --cov=nltk --cov-report html:{envdir}/docs
     pytest
 
 [testenv:pypy]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,6 +145,15 @@ pytest unit/translate/test_nist.py  # unittest
 pytest  # all tests
 ```
 
+(For moderately faster tests, you can install [pytest-xdist](https://pypi.org/project/pytest-xdist/) locally and run your tests in parallel.)
+
+```
+pip install pytest-xdist
+
+# Run tests using 3 subprocesses
+pytest  -n3 nltk/test/
+```
+
 
 ## Continuous Integration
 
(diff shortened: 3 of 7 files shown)


dannysepler:
Followup: the speedup is a bit less than i thought it would be. So happy to abandon this diff if we don't think it's worth it!

**Before**:
Screen Shot 2020-11-28 at 11 17 45 AM

**After**: 
Screen Shot 2020-11-28 at 11 17 50 AM

tox.ini
```diff
@@ -47,31 +48,31 @@ deps =
     twython
 
 commands =
-    pytest
```
iliakur, line 50: Would it be possible to remove this line completely? If I understood tox correctly, `commands` should be inherited from `testenv`. As you have yourself experienced, the reduction in duplication should make future maintenance easier.

nltk/test/unit/lm/test_vocabulary.py
```diff
@@ -149,5 +149,6 @@ def test_len_is_constant(self):
         small_vocab_len_time = timeit("len(small_vocab)", globals=locals())
         large_vocab_len_time = timeit("len(large_vocab)", globals=locals())
 
-        # The timing should be the same order of magnitude.
-        self.assertAlmostEqual(small_vocab_len_time, large_vocab_len_time, places=1)
+        # The timing should be the same order of magnitude. The delta is pretty
```
iliakur: I'm kinda against this. We're changing an important assertion about the software based only on convenience testing it...

Here's an idea for how to deal with this: `pytest` has this concept of markers, we could introduce one for tests that we don't want to run in parallel. In the CI we execute our parallel tests first, then follow up by executing the non-parallel ones. I'd definitely have a look that save us any time as a result, perhaps the additional complication isn't really worth it?

I'm guessing that it should be though, given that tests take quite a while currently.

iliakur (changes requested):

iliakur:
My two cents. I'd be interested in running the pipeline several times for this PR to see if there is actually a speed-up compared to what happens normally. Maybe Travis.ci gives us such low-powered machines that they can't actually handle the extra load from parallelization :D


*[update]*
Just saw your post with the speedup numbers. Doesn't look like it's all too big, the speedup, so maybe we don't include this in the CI. That being said, what about some kind of setup that lets you easily (without any changes to config files) parallelize the tests locally?

I know tox can be run in parallel mode, perhaps that's already enough?

tox.ini
dannysepler: good eye! since pypy comes bundled with numpy, we can keep the parent `commands`. but with the nodeps environments, we need to redefine `commands` because otherwise it would try to install scipy and scikit-learn to those environments.

dannysepler:
totally agree! i don't see any reason for us to do anything beyond letting contributors know they can install `pytest-xdist` on their own. i'll certainly keep using it, since it cuts the test time in half (maybe more) for me. i found 3 processes to be my magic number, so i added that to the docs.

dannysepler:
FWIW, i know nothing about travis CI. but i did wonder what the extra 45 mins were up to.... but also at the end of the day, the switch to pytest is about readability / maintainability, and less about optimization. i'd rather not add too much complexity shaving a few mins off a thing that works pretty well!

**EDIT**: I'm a dummy. 1h15 is total compute time. 30m is wall time, cause it's parallelized.

dannysepler:
New before/after numbers! (Speedup is small, and primarily from not caching `nltk_data`)

Before:
before


After:
Screen Shot 2020-11-29 at 5 55 16 PM

.travis.yml
```diff
@@ -3,13 +3,12 @@ sudo: false
...
 
 cache:
-  apt: true
+  pip: true
   directories:
   - $HOME/.cache/pip
   - $HOME/.ccache
   - $HOME/.pip-cache
   - $HOME/third
-  - $HOME/nltk_data
```
dannysepler, line 12: it's not useful to cache this, since we download with `force=True` on the Travis CI

Also, the "apt" cache is deprecated. https://github.com/travis-ci/docs-travis-ci-com/commit/bd8b56d501dbb47f76d139831c99f55ac69d1314

stevenbird:
Is there consensus yet? My preference is less complexity, in the interests of maintainability.

dannysepler:
thanks for following up!

this pull request kind of morphed, but nothing in it is particularly pressing. i'm just gonna close it.

PR #2632, opened 2020-11-29
Minor improvements to CI build

dannysepler:
Apparently apt-cache is not supported, and they recommend using a pip cache. https://github.com/travis-ci/docs-travis-ci-com/commit/bd8b56d501dbb47f76d139831c99f55ac69d1314

Also moved pytest.ini to base-level (so you can run `pytest` from anywhere in the repo)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,12 @@ sudo: false
 language: python
 
 cache:
-  apt: true
+  pip: true
   directories:
   - $HOME/.cache/pip
   - $HOME/.ccache
   - $HOME/.pip-cache
   - $HOME/third
-  - $HOME/nltk_data
 
 dist: xenial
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,13 +3,12 @@ sudo: false
...
 
 cache:
-  apt: true
+  pip: true
   directories:
   - $HOME/.cache/pip
   - $HOME/.ccache
   - $HOME/.pip-cache
   - $HOME/third
-  - $HOME/nltk_data
```
dannysepler, line 12: this is our largest cache directory, but i don't think it's very useful since we end up downloading with `force=True`

dannysepler:
**Before**:
before

**After**:
Screen Shot 2020-11-29 at 4 55 36 PM

dannysepler:
eh, i'll merge it with the other. sorry for the noise!

PR #2641, opened 2020-12-20
Various test cleanups, and unused import cleanups

dannysepler:
Just some random things! If there's anything you don't like here, let me know and I'll revert it.

### DIFF
--- a/nltk/test/unit/conftest.py
+++ b/nltk/test/unit/conftest.py
@@ -7,9 +7,7 @@ def mock_plot(mocker):
 
     try:
         import matplotlib.pyplot as plt
-
+        mocker.patch.object(plt, 'gca')
         mocker.patch.object(plt, 'show')
     except ImportError:
-        pytest.skip(
-            "mock_plot imports matplotlib which doesn't exist!"
-        )
+        pass
--- a/nltk/test/unit/test_seekable_unicode_stream_reader.py
+++ b/nltk/test/unit/test_seekable_unicode_stream_reader.py
@@ -1,69 +1,37 @@
 # -*- coding: utf-8 -*-
-"""
-The following test performs a random series of reads, seeks, and
-tells, and checks that the results are consistent.
-"""
-import random
-import functools
 from io import BytesIO
+import os
+import pytest
+
 from nltk.corpus.reader import SeekableUnicodeStreamReader
 
 
-def check_reader(unicode_string, encoding, n=1000):
+def check_reader(unicode_string, encoding):
     bytestr = unicode_string.encode(encoding)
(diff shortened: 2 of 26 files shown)


nltk/test/unit/conftest.py
```diff
@@ -7,9 +7,7 @@ def mock_plot(mocker):
 
     try:
         import matplotlib.pyplot as plt
-
+        mocker.patch.object(plt, 'gca')
```
dannysepler, line 10: This is the reason why, when you run tests, it opens random python windows for a split-second. Shouldn't do that anymore!

nltk/test/unit/test_seekable_unicode_stream_reader.py
```diff
@@ -1,69 +1,37 @@
...
-import functools
 from io import BytesIO
+import os
+import pytest
+
 from nltk.corpus.reader import SeekableUnicodeStreamReader
 
 
-def check_reader(unicode_string, encoding, n=1000):
+def check_reader(unicode_string, encoding):
```
dannysepler, line 9: this was kinda a weird test, so i rewrote it. things i wasn't crazy about:

- Basing a test off randomization inherently feels odd?
- It looped a lot per string. The larger strings seem to loop about 6k times on average!
- the physical operations themselves were a bit hard to understand. Also sometimes they do nothing. Like, when ` op == 't'` it would `.tell()` but not use the result.

it feels better to just read in the information in a few ways, and check it against the original string

iliakur:
I can't really speak to the other tests, but the ones for the `lm` module could be taken a bit further in the direction of `pytest`.

The classes with the `setup_class` methods can be replaced with fixtures. The only reason for the existence of the metaclass is to add support for parametrization, which is missing in `unittest`. We can completely replace it with `pytest.mark.parametrize`.

I had it in my todo-list to deal with this, so I'm grateful you're tackling it :) I could help with it, if you want. If you're tight on time or don't want to muck with fixtures and parametrization I could also do it myself too.

dannysepler:
thanks, @iliakur! i tried to replace that part of the `lm` tests too, but couldn't think of how. mind explaining what you'd do a bit more?

stevenbird:
Thanks for all these clean ups @dannysepler. I'm happy to merge them, and for us to consider further changes in light of [@iliakur's suggestions](https://github.com/nltk/nltk/pull/2641#issuecomment-748869316) in a separate PR.

iliakur:
@dannysepler if you don't mind I'll just go ahead and do this one of these days and point you to the PR I open.

dannysepler:
Sounds good. Thanks, both!

PR #2645, opened 2020-12-27
Remove problematic naming

tconroy:
Hello! 👋 Hope you are doing well.

This PR refactors internal naming to drop a problematic name in favor of a more neutral one.

I am not a Python developer, and this is my first interaction with this library/codebase, so I apologize if I've failed to follow any particular convention. I found this repo through a discussion on Twitter and wanted to extend a hand to help make this codebase a more inclusive environment. 

Happy to work with the maintainers to find a more suitable name if the one I provided isn't good enough.

Thank you for your time!

### DIFF
--- a/nltk/chunk/regexp.py
+++ b/nltk/chunk/regexp.py
@@ -11,9 +11,9 @@ import re
 from nltk.tree import Tree
 from nltk.chunk.api import ChunkParserI
 
-##//////////////////////////////////////////////////////
-##  ChunkString
-##//////////////////////////////////////////////////////
+# //////////////////////////////////////////////////////
+# ChunkString
+# //////////////////////////////////////////////////////
 
 
 class ChunkString(object):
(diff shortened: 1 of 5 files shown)


dannysepler:
great PR! just a suggestion, but i think you're missing a few references, for example `IN_<TERM>_PATTERN`

a `git grep --ignore-case <TERM>` could probably help you catch the lingering ones

as far as the test failure, seems like a flake to me? 🤷 maybe it'll resolve itself next commit

tconroy:
Thank you @dannysepler! Great point, I think that will work much better. Will push up a fresh commit this evening when I'm back at my laptop.

tconroy:
Hmm — Not sure why the build is failing. It looks to me like the CI is encountering issues with the `pip install` command?

nltk/chunk/regexp.py
```diff
@@ -1426,7 +1428,8 @@ def demo():
     print("Demonstration of empty grammar:")
 
     cp = chunk.RegexpParser("")
-    print(chunk.accuracy(cp, conll2000.chunked_sents("test.txt", chunk_types=("NP",))))
+    print(chunk.accuracy(cp, conll2000.chunked_sents(
```
dannysepler, line 1431: Not the biggest deal, but curious why some formatting changes appeared? I kinda liked this one better before, although each is fairly subjective

dannysepler:
and looks like other open PRs are also failing on the py3.6 install, so i wouldn't worry about it :)

nltk/chunk/regexp.py
stevenbird: :+1: Longer than 80 characters I guess.

stevenbird:
Thanks @tconroy, @dannysepler

tconroy:
Thank you @stevenbird @dannysepler! Cheers, Hope you both have a great 2021.

PR #2640, opened 2020-12-18
Wordnet doctests for tree() and acyclic_tree()

ekaf:
# Wordnet doctests for tree() and acyclic_tree()

This PR adds tests to wordnet.doctests, to illustrate how the updated *tree()* function in the Wordnet module now avoids endless cycles, and how the new *acyclic_tree()* function handles intractable trees.

### 1. Endlessness

Until NLTK v. 3.5, the _tree()_ function looped forever on symmetric relations (verb_groups, attributes, and most also_sees). But in the current version, _tree()_ now detects and discards these cycles:

```
    >>> from pprint import pprint
    >>> pprint(wn.synset('bound.a.01').tree(lambda s:s.also_sees()))
    [Synset('bound.a.01'),
     [Synset('unfree.a.02'),
      [Synset('confined.a.02'),
       [Synset('restricted.a.01'), [Synset('classified.a.02')]]],
      [Synset('dependent.a.01')],
      [Synset('restricted.a.01'),
       [Synset('classified.a.02')],
       [Synset('confined.a.02')]]]]
```

Specifying the "cut_mark" parameter increases verbosity, so that the cycles are mentioned in the output, together with the level where they occur:

```
    >>> pprint(wn.synset('bound.a.01').tree(lambda s:s.also_sees(),cut_mark='...'))
    [Synset('bound.a.01'),
     [Synset('unfree.a.02'),
      "Cycle(Synset('bound.a.01'),-3,...)",
      [Synset('confined.a.02'),
       [Synset('restricted.a.01'),
        [Synset('classified.a.02')],
        "Cycle(Synset('confined.a.02'),-5,...)",
        "Cycle(Synset('unfree.a.02'),-5,...)"],
       "Cycle(Synset('unfree.a.02'),-4,...)"],
      [Synset('dependent.a.01'), "Cycle(Synset('unfree.a.02'),-4,...)"],
      [Synset('restricted.a.01'),
       [Synset('classified.a.02')],
       [Synset('confined.a.02'),
        "Cycle(Synset('restricted.a.01'),-5,...)",
        "Cycle(Synset('unfree.a.02'),-5,...)"],
       "Cycle(Synset('unfree.a.02'),-4,...)"]]]
```

### 2. Intractability

[...]

PR #2646, opened 2020-12-28
Compute Minimum Spanning Tree of unweighted graph using BFS

ekaf:
This PR provides a better solution to  the problem of intractable trees first mentioned in connection to  issue #2314, which was solved sub-optimally using Depth-first Search in PR #2612, and  complements the corresponding wordnet.doctest  section from PR #2640.
 
A Minimum Spanning Tree (MST) spans all the nodes of a relation graph once, while guaranteeing that each node is reached through the shortest path possible. In unweighted relation graphs like WordNet, a MST can be computed very efficiently in linear time, using Breadth-First Search (BFS). 

Like the acyclic_tree() function introduced in PR #2612, the new "unweighted_minimum_spanning_tree()" function (imported in the Wordnet module as "mst") handles the intractable trees discussed in PR #2640, for ex. "wn.synset('concrete.a.01').mst(lambda s:s.also_sees())".

But the MST is a better solution, because the also_sees() acyclic_tree of for ex. 'bound.a.01' is unnecessarily deep, and thus less optimal than the corresponding MST, since  depth-first search reaches 'classified.a.02' through a longer path, consisting in four links (bound.a.01 > unfree.a.02 > confined.a.02 > restricted.a.01 > classified.a.02):

bound-dfs
By contrast, the following MST is better, because the path to 'classified.a.02' is the shortest possible, consisting only in three links (bound.a.01 > unfree.a.02 > restricted.a.01 > classified.a.02):
```
    >>> pprint(wn.synset('bound.a.01').mst(lambda s:s.also_sees()))
    [Synset('bound.a.01'),
     [Synset('unfree.a.02'),
      [Synset('confined.a.02')],
      [Synset('dependent.a.01')],
      [Synset('restricted.a.01'), [Synset('classified.a.02')]]]]
```

bound-bfs

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -587,6 +587,7 @@ class Synset(_WordNetObject):
 
     from nltk.util import acyclic_depth_first as acyclic_tree
 
+    from nltk.util import unweighted_minimum_cost_spanning_tree as mst
 
     def tree(self, rel, depth=-1, cut_mark=None):
         """
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -587,7 +587,7 @@ class Synset(_WordNetObject):
 
     from nltk.util import acyclic_depth_first as acyclic_tree
 
-    from nltk.util import unweighted_minimum_cost_spanning_tree as mst
+    from nltk.util import unweighted_minimum_spanning_tree as mst
 
     def tree(self, rel, depth=-1, cut_mark=None):
         """
--- a/nltk/test/wordnet.doctest
+++ b/nltk/test/wordnet.doctest
@@ -727,7 +727,7 @@ A Minimum Spanning Tree (MST) spans all the nodes of a relation subgraph once,
 while guaranteeing that each node is reached through the shortest path possible.
 In unweighted relation graphs like WordNet, a MST can be computed very efficiently
 in linear time, using Breadth-First Search (BFS). Like acyclic_tree(), the new
-"unweighted_minimum_cost_spanning_tree()" function (imported in the Wordnet
+"unweighted_minimum_spanning_tree()" function (imported in the Wordnet
 module as "mst") handles intractable trees, such as the example discussed above:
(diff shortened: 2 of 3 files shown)


nltk/corpus/reader/wordnet.py
```diff
@@ -587,6 +587,7 @@ def closure(self, rel, depth=-1):
 
     from nltk.util import acyclic_depth_first as acyclic_tree
 
+    from nltk.util import unweighted_minimum_cost_spanning_tree as mst
```
stevenbird: I don't think we need "cost" in the name, do we?

nltk/test/wordnet.doctest
```diff
@@ -716,6 +720,49 @@ cycles are mentioned in the output, together with the level where they occur:
...
+    >>> pprint(wn.synset('bound.a.01').mst(lambda s:s.also_sees()))
+    [Synset('bound.a.01'),
+     [Synset('unfree.a.02'),
+      [Synset('confined.a.02')],
+      [Synset('dependent.a.01')],
+      [Synset('restricted.a.01'), [Synset('classified.a.02')]]]]
+
+
+Here also, specifying the "cut_mark" parameter increases verbosity, so that the
+cycles are mentioned in the output, together with the level where they occur:
```
stevenbird: @ekaf can you explain the motivation for this please?

nltk/corpus/reader/wordnet.py
ekaf: I totally agree, and changed the name accordingly,

nltk/test/wordnet.doctest
ekaf: Not necessarily: the increased verbosity is nice to have in the normal "tree" function, where cycles are not expected, and may reveal bugs in the Wordnet database. On the contrary, the motivation for computing a MST is when a tree is already known to be intractable, or to include many cycles, in which case one would prefer to only output the trimmed version. So the additional parameters were probably superfluous and have now been removed.

stevenbird:
@ekaf sorry for the delay... hope to get back to this soon

stevenbird:
[CI: retest]

PR #2650, opened 2020-12-30
Synset path_similarity commutativity

### LINKED ISSUE #2278: Wordnet similarity quirky when only one synsets needs root
Sometimes when only one synset needs root, the similarity between the synsets are not commutative:

```python
from nltk.corpus import wordnet as nltk_wn

ncat = nltk_wn.synset('cat.n.01')
nbuy = nltk_wn.synset('buy.v.01')

print(nltk_wn.path_similarity(nbuy, ncat), nltk_wn.wup_similarity(nbuy, ncat))
print(nltk_wn.path_similarity(ncat, nbuy), nltk_wn.wup_similarity(ncat, nbuy))
```

[out]:

```
0.058823529411764705 0.1111111111111111
None None
```

Details on https://stackoverflow.com/q/20075335/610569

snrore:
Fixes #2278

We see in this issue, that `path_similarity` is not commutative when comparing a noun and a verb.
This happens because the function `Synset._needs_root()` returns `True` for verbs, but returns `False` for nouns.  According to the docstring, this is because nouns already have a default root.

So, if the graph is not connected, and you do `verb.path_similarity(noun)`, `path_similarity` will call `verb._needs_root()` which returns `True`, which will add a fake root to connect the graph.
If the operation is done in the opposite direction, `noun.path_similarity(verb)`, the graph will also not be connected. But `noun._needs_root()` returns `False` so a fake root is not added.

The same problem appears in the function `synset.wup_similarity`. 
The function `synset.lch_similarity` also calls `synset.needs_root`, but `lch_similarity` checks that both synsets have the same part of speech, so this problem does not appear with this function. 

This PR adds tests to confirm that path_similarity is commutative for nouns and verbs, and updates `path_similarity` and `wup_similarity` so that a fake root is added if either synset needs a fake root.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -843,6 +843,7 @@ class Synset(_WordNetObject):
             could be found. 1 is returned if a ``Synset`` is compared with
             itself.
         """
+        
         distance = self.shortest_path_distance(
             other, simulate_root=simulate_root
         )
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -843,7 +843,7 @@ class Synset(_WordNetObject):
             could be found. 1 is returned if a ``Synset`` is compared with
             itself.
         """
-        
+    
         distance = self.shortest_path_distance(
             other, simulate_root=simulate_root
         )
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @snrore - great contribution!

PR #2652, opened 2020-12-30
Add syllable tokenizer based on legality and onset maximization

henchc:
This adds the second of the two syllable tokenizers based on the Legality Principle and Onset Maximization from #2203 .

### DIFF
--- a/nltk/tokenize/legality_principle.py
+++ b/nltk/tokenize/legality_principle.py
@@ -61,12 +61,12 @@ class LegalitySyllableTokenizer(TokenizerI):
         [['This'], ['is'], ['a'], ['won', 'der', 'ful'], ['sen', 'ten', 'ce'], ['.']]
     """
 
-    def __init__(self, tokenized_source_text, vowels=set("aeiouy"), legal_frequency_threshold=.001):
+    def __init__(self, tokenized_source_text, vowels="aeiouy", legal_frequency_threshold=.001):
         """
         :param tokenized_source_text: List of valid tokens in the language
         :type tokenized_source_text: list(str)
         :param vowels: Valid vowels in language or IPA represenation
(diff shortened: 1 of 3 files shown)


nltk/tokenize/legality_principle.py
```diff
@@ -0,0 +1,142 @@
...
+        >>> from nltk import word_tokenize
+        >>> from nltk.corpus import words
+        >>> text = "This is a wonderful sentence."
+        >>> text_words = word_tokenize(text)
+        >>> LP = LegalitySyllableTokenizer(words.words())
+        >>> [LP.tokenize(word) for word in text_words]
+        [['This'], ['is'], ['a'], ['won', 'der', 'ful'], ['sen', 'ten', 'ce'], ['.']]
+    """
+
+    def __init__(self, tokenized_source_text, vowels=set("aeiouy"), legal_frequency_threshold=.001):
```
cuspymd: I'm afraid It's not good idea to use `set` type as default value of function parameter. Since It is evaluated only once, the following unintended result may occur.
```
tonkenizer1 = LegalitySyllableTokenizer(some_text)
tonkenizer1.vowels.add('c')
print(tonkenizer1.vowels)  // { 'a', 'e', 'i', 'o', 'u', 'y', 'c' }

tonkenizer2 = LegalitySyllableTokenizer(some_text)
print(tonkenizer2.vowels)  // { 'a', 'e', 'i', 'o', 'u', 'y', 'c' }
```
It would be better to change the type of `vowels` to `string` or to pass `vowels` as string type and then convert it to `set` type in the constructor.

henchc:
Thanks, @cuspymd , updated!

stevenbird:
Thanks @henchc and @cuspymd!

PR #2657, opened 2021-01-12
adding in a variable stepSize.  This is based on Online EM for Unsupe…

hlncrg:
…rvised Models (Liang and Klein 2009).  This metric is used to mix previous translation probability with newly calculated translation probability per iteration step.  This allows mini-batch steps.

### DIFF
--- a/nltk/translate/ibm1.py
+++ b/nltk/translate/ibm1.py
@@ -104,7 +104,7 @@ class IBMModel1(IBMModel):
 
     """
 
-    def __init__(self, sentence_aligned_corpus, iterations, probability_tables=None):
+    def __init__(self, sentence_aligned_corpus, iterations, probability_tables=None, stepSize=1):
         """
         Train on ``sentence_aligned_corpus`` and create a lexical
         translation model.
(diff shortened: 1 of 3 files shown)


hlncrg:
An example of how to use this code would be:

```
step=0
for iter in range(4):
    for minibatchIter in range(0,len(bitext),minibatchSize):
        ibm1 = IBMModel1(bitext[minibatchIter:minibatchIter+minbatchSize], 1,probability_tables={"translation_table":tt}, stepSize=pow(step+2,-0.8))
        step+=1
```

nltk/translate/ibm1.py
```diff
@@ -104,7 +104,7 @@ class IBMModel1(IBMModel):
 
     """
 
-    def __init__(self, sentence_aligned_corpus, iterations, probability_tables=None):
+    def __init__(self, sentence_aligned_corpus, iterations, probability_tables=None, stepSize=1):
```
skang23, line 107: I think step_size would be a better naming for consistency?

nltk/translate/ibm1.py
```diff
@@ -125,6 +125,18 @@ def __init__(self, sentence_aligned_corpus, iterations, probability_tables=None)
             ``translation_table``.
             See ``IBMModel`` for the type and purpose of this table.
         :type probability_tables: dict[str]: object
+
+
+        :param stepSize: Optional. Use this to mix old probability table
```
skang23, line 130: I am curious if you have considered making the batch_size and alpha as parameters instead of step_size?
This would make it easier to use because we can just call ibm1 without having to wrap it in while loops.

stevenbird:
@hlncrg: this is waiting on your response to the review from @skang23

PR #2668, opened 2021-02-25
Fixed incorrect test regarding WordNet wup_similarity

tomaarsen:
Hello!

---
## Pull request overview
This PR is about fixing an incorrect test in `nltk/nltk/test/wordnet.doctest`. 
It is related to #2650. This PR fixed an issue that caused wup_similarity to sometimes return `None` when it shouldn't.

---

## Bug:
https://github.com/nltk/nltk/blob/df1b563049ffafa6ca7ef34b07bf88d98d22db70/nltk/test/wordnet.doctest#L461-L464

The wup_similarity is `0.14285...` since #2650, which *should* be correct. @snrore might be able to help confirm.

The `None` has been replaced with `0.14...` which should test correctly.

---

- Tom Aarsen


stevenbird:
@CubieDev: thanks for this, and sorry for the delay; so just the final commit is substantive?

tomaarsen:
> @CubieDev: thanks for this, and sorry for the delay; so just the final commit is substantive?

@stevenbird That is correct. My apologies, I did not mean for those other commits to become a part of this PR. The first four were small changes based on some Pylint output. I reverted them as they are not relevant to this fix, so they shouldn't be a part of this PR, and they weren't really worth creating a PR over either.

ekaf:
Great catch @CubieDev! This is the first PR to pass the CI-Travis test for a long time, and it seems that the no longer correct None value in wordnet.doctest  was the main reason for the failures.

Now that the header above the tests ("Bug 160: wup_similarity breaks when the two synsets have no common hypernym") has become inaccurate, what about updating it also, f. ex. "Bug 160 resolved: wup_similarity can now handle two synsets without a common hypernym"?

tomaarsen:
@ekaf I do not believe that the header has necessarily become inaccurate. As you can see with all of the tests there, the header describes what the bug *was*, rather than describe a current bug. You can see this is the case in the commit that added the header: https://github.com/nltk/nltk/commit/ad4c8b92bc89698f7726b053b22b1b667c371f80
The header describing the old bug is introduced alongside the fix for the bug. 

However, I do agree that perhaps it is best to document the original issue, the patch that fixed it, and the change made to the test. I'll be doing that now.

Thank you!

ekaf:
Ok, there are also headlines for some of the regression tests that use past tense (fx. used, ignored, returned) to indicate that the bug has been fixed.

stevenbird:
Thanks @CubieDev!

Issue #2648, opened 2020-12-29
Pos-tag returns strange output for 'as'


mina1987:
Any insights on how to handle 'as'?
Inltk.pos_tag('as')
[('a', DT), ('s','NN')

I also tried to check wordnet synsets, most of the synsets were chemical terms and only the last one was my desired synset: Synset('equally.r.01) r , any insights on how i can isolate this specific meaning and use it for pos_tag, etc?

12mohaned:
I believe there is a problem with the pos_tagging classifier, I tried the same sentence on textblob and it worked.
`from textblob import TextBlob`
 `text = ('as')`
 `blob = TextBlob(text)`
 `blob.tags` 

the output is 
[('as', 'IN')]

Updated answer : 
Actually, it does work, you have to tokenize the sentence before tagging it as the parser does not tokenize it for you before parsing. 

`from nltk.tokenize import word_tokenize`
 `nltk.pos_tag(word_tokenize(text))`
 
the output 
[('as', 'IN'), ('you', 'PRP'), ('know', 'VBP'), ('i', 'JJ'), ('love', 'VBP'), ('coffee', 'NN')]

goodmami:
@12mohaned's updated answer is correct, but I'll provide some context.

[nltk.pos_tag()](https://www.nltk.org/api/nltk.tag.html#nltk.tag.pos_tag) expects a list of tokens:

```python
>>> nltk.pos_tag(['as'])
[('as', 'IN')]
```

Since a list of tokens is a iterable of strings, and a single string acts like an iterable of 1-character substrings, they have the same "shape" as far as the function is concerned. That is, the function proceeds as if you'd passed the token list `['a', 's']`. Since passing a string as the argument to `pos_tag()` is a common mistake and not likely to have a valid use-case, the NLTK might be improved by raising a `TypeError` or at least warning users when a string is passed.

PR #2674, opened 2021-03-22
Throws exception when passing string

12mohaned:
Throw error if pos_tag take an argument of type string 

This pull request resolve a problem someone raised as an [issue.](https://github.com/nltk/nltk/issues/2648)

### DIFF
--- a/nltk/tag/__init__.py
+++ b/nltk/tag/__init__.py
@@ -115,8 +115,8 @@ def _pos_tag(tokens, tagset=None, tagger=None, lang=None):
             "(i.e. lang='eng' or lang='rus')"
         )
     # Throws Error if tokens is of string type
-    if str(tokens) == str:
-        raise TypeError('NLTK pos_tag tokens does not support the string type')
+    elif isinstance(tokens, str):
+        raise TypeError('tokens: expected a list of strings, got a string')
 
     else:
         tagged_tokens = tagger.tag(tokens)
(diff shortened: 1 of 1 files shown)


nltk/tag/__init__.py
```diff
@@ -114,6 +114,10 @@ def _pos_tag(tokens, tagset=None, tagger=None, lang=None):
             "Currently, NLTK pos_tag only supports English and Russian "
             "(i.e. lang='eng' or lang='rus')"
         )
+    # Throws Error if tokens is of string type
+    if str(tokens) == str:
```
goodmami: This `if` is inserted between another `if` and an `else`, which changes the control flow. In this case, however, both `if`s raise an uncaught exception (which exit the function immediately) so the behavior is the same. It would be better to use `elif`, particularly if we decide to issue a warning here instead of raise an error.

Also, for a type check it's clearer to check the type:

```python
elif isinstance(tokens, str):
    raise TypeError('tokens: expected a list of strings, got a string')
```

goodmami (commented):
Thanks for putting together a PR. I only suggested a TypeError or warning in #2648, so I'm not sure if the NLTK maintainers actually want such a change as it has the potential to break people's code (if, for some reason, they wanted to pos-tag individual characters). In any case, I have some suggestions (see comment in code).

nltk/tag/__init__.py
12mohaned: I did refactor the changes, I believe it will output wrong result so it should be handled, maybe we can display a warning message when the passed parameter is a string.

stevenbird:
Thanks @12mohaned, @goodmami

Issue #2676, opened 2021-03-24
Issue about BLEU smoothing method 4


OpenSource-fan:
NLTK3.5 implements smoothing method 4 proposed in the paper [https://www.aclweb.org/anthology/W14-3346/ ].
However, we suspect that the implementation is incorrect. The code at https://github.com/nltk/nltk/blob/d0f54c29b3f65ea13dd6e824e08eb9f193bcb958/nltk/translate/bleu_score.py#L582-L585 is problematic.
This leads to the p_n[i] greater than 100% in some case. When hyp_len < 4 in 4-gram, p_n[i] can be assigned with a percentage number that is much greater than 100% (or even > 500%).
In fact, the p_n[i] should not be greater than 100%.
The correct implementation is p_n[i] = (1 / incvnt) / p_i.denominator in line585.

stevenbird:
Thanks @anonymityBoy. Would you please submit a pull request?

stevenbird:
Thanks @anonymityBoy... can I trouble you to add a regression test or two [here](https://github.com/nltk/nltk/blob/develop/nltk/test/bleu.doctest) please?

OpenSource-fan:
Sure. I have added.

PR #2681, opened 2021-03-28
fix bug in calculating BLEU using smoothing 4

OpenSource-fan:
[ISSUE #2676 ] Fix issue about BLEU smoothing method 4

### DIFF
--- a/nltk/test/bleu.doctest
+++ b/nltk/test/bleu.doctest
@@ -25,6 +25,5 @@ http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf
 ...     ['It is a place of quiet contemplation .'.split()],
 ...     'It is .'.split(),
 ...     smoothing_function=SmoothingFunction().method4,
-... )
->>> round(bleu4*100 ,4)
-4.4267
+... )*100
+4.4267...
--- a/nltk/test/bleu.doctest
+++ b/nltk/test/bleu.doctest
@@ -21,9 +21,10 @@ http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf
 >>> from nltk.translate.bleu_score import sentence_bleu,SmoothingFunction
 
 
->>> sentence_bleu(
+>>> bleu4= sentence_bleu(
 ...     ['It is a place of quiet contemplation .'.split()],
 ...     'It is .'.split(),
 ...     smoothing_function=SmoothingFunction().method4,
-... )*100
-4.426730592500287
+... )
+>>> round(bleu4*100 ,4)
+4.4267
--- a/nltk/test/bleu.doctest
+++ b/nltk/test/bleu.doctest
@@ -12,3 +12,18 @@ If the candidate has no alignment to any of the references, the BLEU score is 0.
 ...     [1],
 ... )
 0
+
+This is an implementation of the smoothing techniques
+for segment-level BLEU scores that was presented in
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @anonymityBoy. Minor issue remaining:

```
Expected:
    4.4267
Got:
    0.04426730592500287
```

NB we probably also want `4.4267...` (ie including the dots)

nltk/test/bleu.doctest
```diff
@@ -12,3 +12,18 @@ If the candidate has no alignment to any of the references, the BLEU score is 0.
...
+http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf
+>>> from nltk.translate.bleu_score import sentence_bleu,SmoothingFunction
+
+
+>>> sentence_bleu(
+...     ['It is a place of quiet contemplation .'.split()],
+...     'It is .'.split(),
+...     smoothing_function=SmoothingFunction().method4,
+... )*100
+4.426730592500287
```
tomaarsen: @anonymityBoy 
I would very much advise using for example `4.4267...` (yes, with the dots) instead of the exact value of `4.426730592500287`.

The former will test correctly with any float that starts with (or rounds to, I forget) `4.4267`. This allows for results which are pretty much exactly correct, but differ due to some inconsistencies in float operations. The latter will likely cause some tests to arbitrarily fail due to these inconsistencies in float operations, e.g.:
```python
>>> 1.2 * 1.5
1.7999999999999998
```

OpenSource-fan: Good point. I format the output.

stevenbird:
@anonymityBoy the regression test can be simply:

```
>>> sentence_bleu(
...     ['It is a place of quiet contemplation .'.split()],
...     'It is .'.split(),
...     smoothing_function=SmoothingFunction().method4,
... ) * 100
4.4267...
```

OpenSource-fan:
Thanks. I have simplified the regression test

nltk/test/bleu.doctest
```diff
@@ -12,3 +12,18 @@ If the candidate has no alignment to any of the references, the BLEU score is 0.
...
+
+This is an implementation of the smoothing techniques
+for segment-level BLEU scores that was presented in
+Boxing Chen and Collin Cherry (2014) A Systematic Comparison of
+Smoothing Techniques for Sentence-Level BLEU. In WMT14.
+http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf
+>>> from nltk.translate.bleu_score import sentence_bleu,SmoothingFunction
+
+
+>>> bleu4= sentence_bleu(
```
tomaarsen: You are still assigning to this variable. This returns nothing, causing the test to fail.
Please see the test output here: https://travis-ci.org/github/nltk/nltk/jobs/765303038#L2080-L2083
Remove the `blue4= ` and it will work:
```suggestion
>>> sentence_bleu(
```

OpenSource-fan: Thanks for your suggestion. I have revised it.

tomaarsen: Wonderful! Seems like the test passes now. Well done.

OpenSource-fan: hahaha.

stevenbird:
@dimazest I think you had a hand in writing this code... would you mind giving it a review please? (particularly the new formula)

PR #2683, opened 2021-04-02
Remove usage of thread.isAlive

JelleZijlstra:
The thread.isAlive method was removed in Python 3.9; see https://bugs.python.org/issue37804. It was an alias for thread.is_alive, so use that instead.


stevenbird:
Thanks @JelleZijlstra

Issue #2686, opened 2021-04-04
Travis CI test fails with Python 3.9


ekaf:
Travis CI now also tests with Python 3.9.

The output reveals that the XML Parser used by NLTK's corpus.reader is no longer compatible with the newest Python 3.9 version.

So all NLTK builds will fail Travis CI testing with Python 3.9 until this is fixed:

=================================== FAILURES ===================================
___________________________ [doctest] corpus.doctest ___________________________
615 
616 semcor
617 ------
618 The Brown Corpus, annotated with WordNet senses.
619 
620     >>> from nltk.corpus import semcor
621     >>> semcor.words('brown2/tagfiles/br-n12.xml')  # doctest: +ELLIPSIS
622     ['When', 'several', 'minutes', 'had', 'passed', ...]
623     >>> sent = semcor.xml('brown2/tagfiles/br-n12.xml').findall('context/p/s')[0]
624     >>> for wordform in sent.getchildren():
UNEXPECTED EXCEPTION: AttributeError("'xml.etree.ElementTree.Element' object has no attribute 'getchildren'")
Traceback (most recent call last):
  File "/opt/python/3.9.1/lib/python3.9/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest corpus.doctest[135]>", line 1, in <module>
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'
/home/travis/build/nltk/nltk/nltk/test/corpus.doctest[0m:624: UnexpectedException
920 rte
921 ---
922 The RTE (Recognizing Textual Entailment) corpus was derived from the
923 RTE1, RTE2 and RTE3 datasets (dev and test data), and consists of a
924 list of XML-formatted 'text'/'hypothesis' pairs.
925 
926     >>> from nltk.corpus import rte
927     >>> print(rte.fileids()) # doctest: +ELLIPSIS
928     ['rte1_dev.xml', 'rte1_test.xml', 'rte2_dev.xml', ..., 'rte3_test.xml']
929     >>> rtepairs = rte.pairs(['rte2_test.xml', 'rte3_test.xml'])
UNEXPECTED EXCEPTION: AttributeError("'xml.etree.ElementTree.Element' object has no attribute 'getiterator'")

stevenbird:
[...]

PR #2687, opened 2021-04-04
Fixes XML ElementTree compatibility (issue #2686)

ekaf:
In Python 3.9, some functions that were previously flagged as deprecated in the xml package have now been removed. 

This PR applies the recommended replacements, still maintaining compatibility with older Python versions, by keeping two versions of the calls. However this is too cumbersome with one test in corpus.doctest, because the very long output is slightly different in each case, so that test has simply been removed for the moment.


stevenbird:
Thanks @ekaf!

Issue #2690, opened 2021-04-07
nltk 3.6 has pytest as unlisted requirement


cvanweelden:
From version 3.6 pytest is a requirement, e.g. it's imported in https://github.com/nltk/nltk/blob/24d6e0d4123a5cfea3ede7dbdfe77fe0ec54b83c/nltk/parse/corenlp.py line 23. However it is not listed as requirement in setup.py. This probably works fine if you set up the dev environment, but for a normal user install it results in an error.

Reproduce (in fresh env):
```
$ pip install nltk
...
$ python -m nltk
...
ModuleNotFoundError: No module named 'pytest'
```

yokotoka:
Vote for it! We don't need pytest in the production environment. 🙃

LucaCappelletti94:
Same issue! Please move the pytest-related code to the test suite!

pombredanne:
See https://github.com/nltk/nltk/pull/2629#issuecomment-815036244 as the problematic commit.
This is breaking all of my users pip installation as I was not putting an upper bound on nltk in my deps.
@dannysepler gentle ping since you made the commit that introduced pytest

pombredanne:
For me the solution is to change our deps this way https://github.com/nexB/scancode-toolkit/commit/75b95c51ba977e13cca0c716cbb0be2773f93d9a

`nltk >= 3.2, !=3.6, < 4.0`

but this is creating quite a bit of stir and churn for users as this will require a new release to be pushed and our previous release to be potentially yanked too, unless nltk 3.6 is yanked soon enough (like in thew next few hours)

daltonpearson:
Just created a hotfix and submitted the pull request: https://github.com/nltk/nltk/pull/2693

pombredanne:
> Just created a hotfix and submitted the pull request: #2693

Thanks, but somehow when I look at the fix this brings badzillion questions. Why would test runner code even be used there at all? At least using unittest as before while as problematic was standard library. IMHO test code should be removed/moved elsewhere from core code.

dannysepler:
Sorry for not thinking about this in my diff, and thanks for surfacing it so quickly! @daltonpearson 's diff makes sense to me. `nose` code existed in non-test folders before my diff, so following the previous pattern of only importing nose inside `setup_module` tests makes sense for the short term. But as a non-hotfix, i created https://github.com/nltk/nltk/issues/2694 so that we can solve this in a better way!

pombredanne:
[...]

PR #2693, opened 2021-04-07
Hotfix/pytest production dependancy

daltonpearson:
A simple fix to remove the need to install pytest when using nltk in production. Solves issue: https://github.com/nltk/nltk/issues/2690

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -278,6 +278,7 @@
 - Akshita Bhagia <https://github.com/AkshitaB>
 - Pratap Yadav <https://github.com/prtpydv>
 - Hiroki Teranishi <https://github.com/chantera>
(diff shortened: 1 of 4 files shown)


dannysepler:
thank you so much!! how about `nltk/app/__init__.py` as well?

daltonpearson:
> thank you so much!! how about `nltk/app/__init__.py` as well?

Good catch! I've added the fix there as well.

dannysepler:
I gave another read through and everything else is either in a `setup_module` or `test_draw` in the case of `nltk/sem/drt.py` (which i also assume is a test function). 

Really appreciate you solving this so quickly!! @stevenbird what do you think?

stevenbird:
Thanks @daltonpearson and @dannysepler! Released now.

PR #2696, opened 2021-04-11
Add a "no dependency" runtime check to tox

dannysepler:
I was scratching my head as to how some check didn't catch that "pytest" issue earlier this week. And then I realized, all of the relevant tox targets are unit tests, meaning they all have pytest installed as extras! So this adds a NEW target, which has nothing installed beyond what comes with nltk.

The test passes now but I reverted https://github.com/nltk/nltk/pull/2693 and the test failed


stevenbird:
Thanks @dannysepler. I was wondering the same thing.

PR #2697, opened 2021-04-11
Move all pytest code to nltk/test/

### LINKED ISSUE #2694: Move all pytest code into `nltk/test/`
A better way to solve https://github.com/nltk/nltk/issues/2690 may be to remove all pytest imports that are outside of `nltk/test/`

dannysepler:
Closes https://github.com/nltk/nltk/issues/2694. A lot of these setup_module functions actually aren't even part of files that are tested! In instances like those, I just deleted the setup_module entirely. Also tried to teardown corpora in a more "pytest" way.

**Testing:**
Now, `git grep pytest nltk/ | grep -v nltk/test/` returns  nothing. I tested via `pytest nltk/test/` but also wanna make sure all the tox environments pass on this build.


stevenbird:
Brilliant!

stevenbird:
Released in NLTK 3.6.2

Issue #2689, opened 2021-04-06
More Python3.9 deprecations


ekaf:
Running pytest with the new Python 3.9.4 shows a number of deprecation warnings about future failures to anticipate:

============================= test session starts ==============================
platform linux -- Python 3.9.4, pytest-6.2.0, py-1.10.0, pluggy-0.13.1

[....]
=============================== warnings summary ===============================
chunk.doctest::chunk.doctest
  <doctest chunk.doctest[11]>:1: DeprecationWarning: invalid escape sequence \.

classify.doctest::classify.doctest
classify.doctest::classify.doctest
classify.doctest::classify.doctest
classify.doctest::classify.doctest
classify.doctest::classify.doctest
nltk/util.py:64: DeprecationWarning: `formatargspec` is deprecated since Python 3.5. Use `signature` and the `Signature` object directly
    argspec = inspect.formatargspec(args, varargs, varkw, defaults)

corpus.doctest::corpus.doctest
  <doctest corpus.doctest[232]>:2: DeprecationWarning: invalid escape sequence \.

corpus.doctest::corpus.doctest
  <doctest corpus.doctest[272]>:1: DeprecationWarning: invalid escape sequence \.

data.doctest::data.doctest
  <doctest data.doctest[63]>:1: DeprecationWarning: 
    Function BufferedGzipFile() has been deprecated.  Use gzip.GzipFile
    instead as it also uses a buffer.

data.doctest::data.doctest
  <doctest data.doctest[67]>:1: DeprecationWarning: 
    Function BufferedGzipFile() has been deprecated.  Use gzip.GzipFile
    instead as it also uses a buffer.

gensim.doctest::gensim.doctest
  <doctest gensim.doctest[7]>:1: DeprecationWarning: Call to deprecated `__getitem__` (Method will be removed in 4.0.0, use self.wv.__getitem__() instead).

gensim.doctest::gensim.doctest
  /usr/lib64/python3.9/site-packages/gensim/models/keyedvectors.py:877: FutureWarning: arrays to stack must be passed as a "sequence" type such as list or tuple. Support for non-sequence iterables such as generators is deprecated as of NumPy 1.16 and will raise an error in the future.
    vectors = vstack(self.word_vec(word, use_norm=True) for word in used_words).astype(REAL)

gluesemantics.doctest::gluesemantics.doctest
  <doctest gluesemantics.doctest[55]>:1: DeprecationWarning: invalid escape sequence \P

[...]

PR #2698, opened 2021-04-14
Fix deprecation warnings in doctests

rmalouf:
This fixes some miscellaneous warnings that come up when running tests


rmalouf:
Hah - I just now saw #2689. This fixes some of those warnings.

stevenbird:
Thanks @rmalouf

PR #2700, opened 2021-04-15
Compatibility for Pypy v7.3.4.

rmalouf:
These changes bring compatibility with pypy v7.3.4.

The biggest issue is that pypy's gc doesn't automatically clean up open files, so unless we're careful about closing them we'll quickly exceed the open file limit. Closing files when we're done with them is probably not a bad practice in cpython too.


stevenbird:
Thank you @rmalouf!! This was long overdue.

stevenbird:
Released in NLTK 3.6.2

PR #2701, opened 2021-04-17
Speedup two tests

dannysepler:
Philosophically, I don't think unit tests should be responsible for loading whole corpora. Ideally, they should load mock datasets. But for now we can downsample their input size to make them a bit faster.

`rte_classify`: ~60s to 4s
`gensim.doctest`: 45s to 25s

This is not to say this diff shaves 80 seconds off of the whole test suite. It's more around the lines of 30 or 40 seconds.

### DIFF
--- a/nltk/test/conftest.py
+++ b/nltk/test/conftest.py
@@ -15,12 +15,15 @@ def mock_plot(mocker):
         pass
 
 
-@pytest.fixture(scope="session", autouse=True)
-def teardown_loaded_corpora(request):
+@pytest.fixture(scope="module", autouse=True)
+def teardown_loaded_corpora():
     """
     After each test session ends (either doctest or unit test),
     unload any loaded corpora
     """
+
+    yield  # first, wait for the test to end
+
     import nltk.corpus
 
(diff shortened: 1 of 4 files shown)


nltk/test/conftest.py
```diff
@@ -15,12 +15,15 @@ def mock_plot(mocker):
...
-@pytest.fixture(scope="session", autouse=True)
-def teardown_loaded_corpora(request):
+@pytest.fixture(scope="module", autouse=True)
+def teardown_loaded_corpora():
     """
     After each test session ends (either doctest or unit test),
     unload any loaded corpora
     """
+
+    yield  # first, wait for the test to end
```
dannysepler, line 25: this was my bad (oops) this fixture was running before the tests rather than after. ([Relevant pytest docs.](https://docs.pytest.org/en/stable/fixture.html#yield-fixtures-recommended)) Also I changed it from "session" to "module" because waiting till the end of all the tests to teardown is slower than doing smaller teardowns after each test. I'm not totally sure why.

stevenbird:
Good idea, thanks

PR #2702, opened 2021-04-17
Remove redundant doctest option flags

dannysepler:
There's no reason for these anymore, since they're in the `pytest.ini` by default. (I "skipped" `SKIP`, since it's not in the pytest.ini)

I did this by the following. On a mac, so this syntax may or may not work on your computer.
```
$ cd nltk/test
$ ls | grep doctest | grep -v "portuguese" | xargs sed -i '' '/SKIP/! s/[[:space:]]*# doctest:[[:space:]]*.*//'
```

portuguese had some characters that didn't play well with sed, so removed those by hand


stevenbird:
Nice, thanks!

PR #2659, opened 2021-01-22
Fixing Kneser-Ney LM Implementation

manujosephv:
@iliakur 

The most commonly stated example to get the intuition of Kneser Ney is below:

There is a sentence - "I can't see without my reading _" - and we need to complete this sentence. But the rule is that you are only allowed to look at the last word, just like a language model. Intuitively, the guess would be "glasses". But what about "Francisco"? If you just look at the pure unigram frequencies, Francisco has a higher probability than glasses.

Suppose, we do not have the bigram "reading glasses" in our corpus. So under Absolute Discounting, we will be using the unigram probabilities of the word "glasses" and "Francisco" to estimate the next word and because of the higher frequency of "Francisco" it is more likely to be picked as the word than "glasses".

This is the problem Kneser-Ney tries to tackle. Eventhough "Francisco" is more common, it always follows "San". So, can we use this idea when we are backing-off to a lower order probability. This idea is formalized as Continuation Probability - how likely is a word to appear as a novel continuation?

image

image

#### References

- [Mini Example](http://idiom.ucsd.edu/~rlevy/lign256/winter2008/kneser_ney_mini_example.pdf)
- [PPT](http://www.cs.virginia.edu/~kc2wc/teaching/NLP16/slides/04-LMeval.pdf)
- [Dan Jurafsky](https://www.youtube.com/watch?v=ody1ysUTD7o)
- [Simple calculation-Medium](https://medium.com/@dennyc/a-simple-numerical-example-for-kneser-ney-smoothing-nlp-4600addf38b8)
- [Another doc](https://www.cl.uni-heidelberg.de/courses/ss15/smt/scribe6.pdf)
- [Original Paper](https://www-i6.informatik.rwth-aachen.de/publications/download/951/Kneser-ICASSP-1995.pdf)

PR #2704, opened 2021-04-24
Adding New Language Models - Absolute Discounting, StupidBackoff

manujosephv:
@iliakur 
Have created a separate branch where I have added these two models. This is an addition on top of the existing PR for KneserNey #2659

Step1: merge KneserNey
Step2: merge this PR

# Absolute Discounting
Just subtract a constant discount from the probability mass on seen words to distribute among unseen.
Let's look at the formula for it:

image

Now, let's see what $\lambda$ is. To make sure this still sums up to one, we need to allocate the probability mass we offset from the first term to the second term.  What is the probability mass we discounted in the first term? 

image

where $N_{1+}(w_{i-1}, \bullet)$ is the number of unique words which follow the history.

The recursion in the formula ends at $P_{unigram}(w)$, which is just the plain unigram probabilities of the word.

[2] - Kenneth W. Church, William A. Gale,A comparison of the enhanced Good-Turing and deleted estimation methods for estimating probabilities of English bigrams, Computer Speech & Language, Volume 5, Issue 1

# Stupid Backoff
Last, but not the least, let's take a look at Stupid Backoff. In 2007, researchers at Google came up with a new way of backing-off to lower models. Kneser-Ney smoothing had already established itself as the state-of-the-art, but it was computationally intensive. So, when dealing with huge corpus like the Goole n-gram corpus with 1 Trillion words, it was a challenge. So, these researchers proposed something completely out of the box. We were doing all of this probability circus so that the probabilities always add up to one and stay a true probability. What if we let go of that property, and get away without smoothing the model. In other words:

image

We are calling it explicitly, S, and not a P to drive home the fact that this does not give you a probability, but more like a score. The backoff value, $\lambda$, can be calculated empirically. We can have different factors for each length context, or have a constant one. In the experiments that the google researchers did, it was shown that $\lambda = 0.4$ was a value that worked well.

[Paper](https://www.aclweb.org/anthology/D07-1090.pdf)


manujosephv:
closing because #2709 supersedes this

PR #2709, opened 2021-05-03
Lm 2021 facelift

### LINKED ISSUE #2599: Unigram scores incorrectly computed for Kneser-Ney smoothing causing high perplexities
The unigram scores in KneserNey smoothing introduced in this commit https://github.com/nltk/nltk/commit/2759b03d85983fedd29b646380c713c3f4ccff66 are computed as a uniform distribution over the vocabulary, which is clearly not how any variant of KneserNey computes it. 

https://github.com/nltk/nltk/blob/5023d6b933ef1a5b1f25fba1d5ed11a8a43a47e4/nltk/lm/smoothing.py#L47

iliakur:
This supercedes #2659 and #2704, they can both be closed.

All credit for the conceptual work goes to @manujosephv I just translated his intents into idiomatic Python.

@manujosephv please take a look and let me know if anything is unclear. 

I think this also fixes #2599. @osmanio2 can you confirm?


nltk/lm/models.py
```diff
@@ -92,8 +119,23 @@ def __init__(self, order, **kwargs):
...
+    def __init__(self, order, discount=0.75, **kwargs):
+        super().__init__(
+            AbsoluteDiscounting, order, params={"discount": discount}, **kwargs
+        )
+
+
 class KneserNeyInterpolated(InterpolatedLanguageModel):
     """Interpolated version of Kneser-Ney smoothing."""
 
     def __init__(self, order, discount=0.1, **kwargs):
```
iliakur, line 134: @manujosephv I undid your change of default value here. I think we add enough entropy for existing users by changing the scores, let's not throw this in as well. I've set the value to `0.75` in the tests, so they didn't need to be changed.

nltk/lm/smoothing.py
```diff
@@ -72,66 +72,52 @@ def unigram_score(self, word):
...
-                if is_key:
-                    unique_continuation_keys += [
-                        ngram + (word,) for word in self.counts[key_len][ngram].keys()
-                    ]
-                    if is_value:
-                        continuation_keys.append(ngram + (value,))
-        continuation_count = len(set(continuation_keys))
-        unique_continuation_count = len(set(unique_continuation_keys))
-        return (continuation_count, unique_continuation_count)
+    def _continuation_counts(self, word, context=tuple()):
```
manujosephv, line 111: A bit skeptical about this function. The idea is to find number of unique continuation. for eg. if the context is "ab" and c follows it 3 times, d follows 1, and e follows 1, the the unique continuation is only 3(c, d, and e), and not 5

Let me check this function a little more deeply and figure out if it works out the same or not.

iliakur: I'm taking advantage of the fact that all our frequency distributions already account for uniqueness when it comes to the keys. We never look at actual counts here.

manujosephv: I guess you are right. The two functions are performing identically.

manujosephv (approved):
Liked the use of ngram order to track the recursion order. And all in all good refactoring of the code as well. Much more pythonic now :)

nltk/lm/smoothing.py
iliakur: Cool!

I think next weekend I'll try to come up with some test cases that might highlight the types vs tokens distinction.

iliakur: Past weekend ended up being hectic, I'll get back with the test examples (if any) on Thursday/Friday.

iliakur: After that this should be ready to merge I believe.

iliakur: Ok, I noticed that the comments explaining the tests were in need of some love, so I did that instead.

I think before I add more tests I should switch the whole setup to `pytest`, I'll do that in a separate PR. I already have a list of phenomena I want to test.

@manujosephv please double-check the accuracy of the new comments.

stevenbird:
Thanks for your work @iliakur

nltk/lm/smoothing.py
manujosephv: Double checked the KneserNey, AbsoluteDiscounting, and StupidBackoff testcase comments, they seem to be correct.

iliakur:
@stevenbird I think we can merge this, no more feedback is likely.

stevenbird:
Thanks @manujosephv and @iliakur!!

PR #2713, opened 2021-05-10
Fix bug in nltk.metrics.aline, C_skip = -10

### LINKED ISSUE #2708: Wrong default value in aline.py
I believe the value of `C_skip` in https://github.com/nltk/nltk/blob/develop/nltk/metrics/aline.py#L51 should be negative. In particular, the value in the original paper seems to be `-10`, not `10` (see https://webdocs.cs.ualberta.ca/~kondrak/papers/thesis.pdf, p. 54).

DanielLoney:
Fixes #2708


tomaarsen (approved):
Whether the tests are "correct" per se is a hard question. We simply cannot use [Kondrak's Thesis](https://www.cs.toronto.edu/pub/gh/Kondrak-PhD-thesis-2002.pdf) as a source of tests, as many small modifications are made on this version relative to his thesis.

However, it does appear like C_skip should be -10, rather than 10. In the aforementioned thesis, this can be very vaguely seen on page 54. I can also see how someone could read over this. On [another publication](https://aclanthology.org/A00-2038.pdf) of Kondrak at page 292, the value of -10 is more visible.

So, I'm confident in two things:
- Our ALINE implementation differs from Kondrak's. Likely not by much, as it only differs in terms of tuning (the rest seems identical to the thesis). This makes the tests somewhat moot beyond "This shows that it doesn't crash", as we can't use it to verify that it's exactly the same as Kondrak's ALINE, but that is not necessarily that bad.
  - We either accept that we cannot make Kondrak's tests pass, or we try to "revert" to exactly Kondrak's version.
- The value of C_skip was intended to be -10, not 10. As a result, I approve of these changes.

If this is merged, then #2708 and #2712 can be closed.

tomaarsen:
Resolved a merge conflict between the updated test expectations and the pytest-ifying of tests from #2743.

tomaarsen:
The (what I assume to be) merge conflict that was resolved in 3761caf has reverted the test fixes from 8c9a7d55c40073ac217757fdbd640e8e36eb0490, causing the failing tests. I'm confident that the tests from 8c9a7d55c40073ac217757fdbd640e8e36eb0490 are correct, as they are taken directly from [Kondrak's thesis](https://www.cs.toronto.edu/pub/gh/Kondrak-PhD-thesis-2002.pdf) (note, not all of Kondrak's tests pass on our system, but these do).

stevenbird:
Sorry, my mistake...

PR #2715, opened 2021-05-16
Hotfix empty trees

ekaf:
This PR fixes the handling of empty trees in the new Tree.fromlist() function.

Currently, when a tree is empty, Tree.fromlist() returns the input node as a string, so further processing of the result will fail with functions that expect a tree:

import nltk
print("NLTK v. %s" % nltk.__version__)
> NLTK v. 3.6

from nltk.tree import Tree

t=Tree.fromlist(wn.synset('entity.n.01').tree(lambda s:s.hypernyms()))
print(t)
> Synset('entity.n.01')

print(type(t))
> <class 'str'>

This fails:
print(t.pformat())
> AttributeError: 'str' object has no attribute 'pformat'

After applying this PR, a tree is returned instead of a string:

t=Tree.fromlist(wn.synset('entity.n.01').tree(lambda s:s.hypernyms()))
print(t)
> (Synset('entity.n.01') )

print(type(t))
> <class 'nltk.tree.Tree'>

This succeeds:
print(t.pformat())
> (Synset('entity.n.01') )


iliakur:
Don't forget to revert the permissions on those other files. `644` seems a bit too restrictive for a python library ;)

ekaf:
This PR introduces a bigger problem, because trees would have no leaves, There doesn't seem to be a solution to the empty tree case.

PR #2720, opened 2021-05-29
setup.py: fix python_requires

mgorny:
Fix python_requires to specify a valid version.  Apparently, combining
`>=` and `.*` is disallowed, and I'm pretty sure 'just' `>=3.5` means
the same.

To reproduce the problem, you can:

    $ pip install nltk distlib
    $ python -c "import distlib.database; \
      distlib.database.DistributionPath().get_distribution('nltk')"

which yields a (not very readable) error of:

    ValueError: '.*' not allowed for '>=' constraints


stevenbird:
Precisely. Thanks @mgorny

PR #2723, opened 2021-06-02
Fixed broken vocabulary test

tomaarsen:
Fixes the flaky test mentioned in #2722.

Hello!

# Pull request overview
* Fixed a single test in `nltk/test/unit/lm/test_vocabulary.py`

## The bug
Previously, the following line would not correctly do as the comment states:
https://github.com/nltk/nltk/blob/e4444c9b15762e6d86f5f6c4f5faadb87632c72a/nltk/test/unit/lm/test_vocabulary.py#L152-L153

In particular, the broken test didn't check whether the timings are in indeed the same order of magnitude. 

As per the [assertAlmostEqual documentation](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual), `assertAlmostEqual(a, b, places)` checks that `round(a-b, places) == 0`. Then, as can be seen in [this failing Travis CI](https://travis-ci.org/github/nltk/nltk/jobs/773245334#L2097-L2117) run, `0.0400...` and `0.4656...` are not equal within 1 place.

This makes sense, as `round(0.0400... - 0.4656..., 1) => round(-0.065217..., 1) => -0.1`, and this is not equal to `0.0`.

## The fix
So, if we truly want to (roughly) check whether the two are in the same order of magnitude, then one simple way to do so is to ensure that `a < b * 10`, and `b < a * 10`. So, this only holds if `a` is not more than 10 times as large as `b`, and not more than 10 times smaller than `b`.

This is implemented through
```python
        # The timing should be the same order of magnitude.
        self.assertLess(large_vocab_len_time, 10 * small_vocab_len_time, 
            "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
        self.assertLess(small_vocab_len_time, 10 * large_vocab_len_time, 
            "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
```

---

This test should pass, and should conform to the desired behaviour described in the test comment, unlike the current test.
Upon merging, the tests from #2722 should pass again.

cc: @stevenbird @goodmami @iliakur 

- Tom Aarsen

### DIFF
--- a/nltk/test/unit/lm/test_vocabulary.py
+++ b/nltk/test/unit/lm/test_vocabulary.py
@@ -150,4 +150,6 @@ class NgramModelVocabularyTests(unittest.TestCase):
         large_vocab_len_time = timeit("len(large_vocab)", globals=locals())
 
         # The timing should be the same order of magnitude.
-        self.assertAlmostEqual(small_vocab_len_time, large_vocab_len_time, places=1)
+        self.assertLess(large_vocab_len_time, 10 * small_vocab_len_time, 
+            "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
+
--- a/nltk/test/unit/lm/test_vocabulary.py
+++ b/nltk/test/unit/lm/test_vocabulary.py
@@ -150,8 +150,8 @@ class NgramModelVocabularyTests(unittest.TestCase):
         large_vocab_len_time = timeit("len(large_vocab)", globals=locals())
 
         # The timing should be the same order of magnitude.
-        self.assertLess(large_vocab_len_time, 10 * small_vocab_len_time, 
-            "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
-        self.assertLess(small_vocab_len_time, 10 * large_vocab_len_time, 
-            "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
+        self.assertLess(large_vocab_len_time, 2 * small_vocab_len_time, 
+            "Computing the length of a small Vocabulary should take roughly as long as computing the length of a large Vocabulary.")
+        self.assertLess(small_vocab_len_time, 2 * large_vocab_len_time, 
+            "Computing the length of a small Vocabulary should take roughly as long as computing the length of a large Vocabulary.")
 
--- a/nltk/test/unit/lm/test_vocabulary.py
+++ b/nltk/test/unit/lm/test_vocabulary.py
@@ -152,4 +152,6 @@ class NgramModelVocabularyTests(unittest.TestCase):
         # The timing should be the same order of magnitude.
         self.assertLess(large_vocab_len_time, 10 * small_vocab_len_time, 
             "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
+        self.assertLess(small_vocab_len_time, 10 * large_vocab_len_time, 
+            "Computing the length of a small Vocabulary should take within the same order of magnitude as computing the length of a large Vocabulary.")
 


iliakur:
While I understand that "order of magnitude" _typically_ means 10x, I don't think precisely that value was meant by this comment. The point of the test is to make sure that retrieving the vocabulary size is near-constant and doesn't depend on the size of the vocabulary.

When we added the test we refrained from picking any arbitrary "order of magnitude", but we could settle for `2` if that helps make the test less flaky. `10` is definitely way too large of a margin to be informative.

tomaarsen:
You're completely right. I took the definition too literally, but clearly if one of the lengths takes 1 second to compute, and the other one 9, then it's not like the `len_is_constant`, like the name of the test suggests. 

I'll reduce it to 2, meaning that computing the length of either vocabulary may not be twice as large or small as the computation time of the length of the other vocabulary.

iliakur:
Apologies if my first comment was a bit curt. I should have waited until my workday was over and I could phrase it more kindly.

Thanks for looking into this and sticking with it. After some more thought I still have concerns. The factor `2` seems a bit arbitrary. Also note that even with `10` as a factor the test would fail given the timing values you showed in the issue description. I haven't had time to collect more data points but maybe this approach will ultimately **not** solve the flakiness problem?

Honestly, I'm feeling out of my depth here, so I'll search/ask about this on StackOverflow in the next couple of days. I suspect it's a common enough problem that there are known ways to tackle it. When we originally wrote the test we really didn't do much research into performance testing best practices, this is a good opportunity to learn!

If you're interested in resolving the immediate pain so that you and others can work peacefully, I'd suggest disabling the test for now as an easy fix. The issue is on my radar and will get addressed properly in the near future, so the test won't go dead on us :)

tomaarsen:
Whatever you wish, I don't have a preference on the matter. This PR was meant as a quick and simple solution that preserves usefulness of the test, meant to allow other PRs to pass their tests like normal. A more permanent solution can always be implemented later on.

goodmami:
I think I'm waiting on this for #2722. I'm wondering why we are testing performance in a unit test. @iliakur, since it looks like you added this test in faf7152869c94f9ce0a884ce2386fa6e24f60f50, can you provide some context? Related to #2496, perhaps?

If it's important to keep the test, we might also try bumping up the number of iterations for the `timeit` call. When the time to complete is so small, deltas become less stable.

iliakur:
Hi @goodmami you correctly guessed the related issue.

As for why we're testing performance in a "unit" test, afaik nltk doesn't explicitly make a distinction between unit and integration tests, so I didn't know where else to put it. Not to mention that generally the line between the two test types in  my experience is much fuzzier than it may seem.

As for this specific test, we could definitely increasing the sample size. Do you have an idea for what to try?

goodmami:
Thanks for explaining. I'm happy to skip the test as in #2728, but I'll address your question below.

> As for this specific test, we could definitely increasing the sample size. Do you have an idea for what to try?

Actually, rather than increasing the number of iterations, it might be better to do several rounds with the `timeit.repeat()` function, similar to the command-line usage of `timeit`. With `repeat` you can see the variance in multiple rounds:

```pycon
>>> import timeit
>>> from nltk.lm import Vocabulary
>>> from nltk.corpus.europarl_raw import english
>>> small_vocab = Vocabulary('abcde')
>>> large_vocab = Vocabulary(english.words())
>>> timeit.repeat('len(small_vocab)', globals=locals(), repeat=5)
[0.22175809400005164, 0.154130116000033, 0.15880793500036816, 0.15081139299991264, 0.14879256899985194]
>>> timeit.repeat('len(large_vocab)', globals=locals(), repeat=5)
[0.2633845920004205, 0.16705999699979657, 0.16612389200008693, 0.16602043500006403, 0.16873413400026038]
```

You might discard the first result if it tends to vary the most. Either way, the average should be more stable than any two random readings:

```pycon
>>> sum(timeit.repeat('len(small_vocab)', globals=locals(), repeat=5)) / 5
0.1796754442000747
>>> sum(timeit.repeat('len(large_vocab)', globals=locals(), repeat=5)) / 5
0.17757056319996992
```

I think the performance on test runners on Travis, GitHub Actions, etc. can be erratic if you don't have a dedicated server, as the same machine may be running many other jobs simultaneously. You could even test the stdev of the samples of `timeit.repeat()` and do something special if it's too high, like repeat the test up to a few times, or just skip the test with a warning.

iliakur:
Thanks for these pointers @goodmami they give me a good starting point to play with. I really appreciate it!

stevenbird:
Resolved by #2728

PR #2726, opened 2021-06-11
create method word_as_freq at WordListCorpusReader

fernandoasevedo:
Create method `word_as_freq` at `nltk.corpus.reader.wordlist.WordListCorpusReader`

This method will return a `dict(str -> int)` in which the key is the word and the value the respective frequency on the corpus.
This could be intresting for `nltk.corpus.stopwords`, as well, when the developer could call it and (without more codings) apply the `dict` to remove stopwords from the input-texts faster (faster than list)


stevenbird:
For this functionality, please use `FreqDist` to wrap a call to the corpus reader

PR #2733, opened 2021-06-17
Update wsd.py  by  Adding updated porter stemmer 

DASHANANT:
performing stem forms of context  words and words in definations will give better intersection count.

Example: 
'playing' word in sentence , 'plays' word in sense definition  
will give null intersection 
but if we perform stemming both words will be play and gives intersection as 1.
which may helps to predict best sense.

### DIFF
--- a/nltk/wsd.py
+++ b/nltk/wsd.py
@@ -1,4 +1,4 @@
-# Natural Language Toolkit: Word Sense Disambiguation Algorithms
+# Natural Language Toolkit: Word Sense Disambiguation Algorithm
 #
 # Authors: Liling Tan <alvations@gmail.com>,
 #          Dmitrijs Milajevs <dimazest@gmail.com>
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks for your contribution @DASHANANT. This is a proposal to modify the definition of the Lesk algorithm. Minimally it would have to be controlled by a flag, defaulting to the existing behaviour. However, the proposal has a linguistic flaw: the senses of a word are not necessarily included in the senses of the stemmed version of that word. For example, in Wordnet, [computer](http://wordnetweb.princeton.edu/perl/webwn?s=computer) has two senses, while [compute](http://wordnetweb.princeton.edu/perl/webwn?s=compute) only has one.

DASHANANT:
Thank You, The flaw can be solved using a simple if condition .
def stem(i):
    if wordnet.synsets(ps.stem(i))==wordnet.synsets(i):
        return PorterStemmer.stem(i)
    else:
        return i

we change the stem function according to wordnet.
I have simulated it, now it is giving better results than previous.
Thank You again.

Please check it.

PR #2736, opened 2021-06-22
Edit_distance now computes the actual Damerau-Levenshtein edit-distance

### LINKED ISSUE #2734: edit_distance() does not compute the actual Damerau-Levenshtein distance when transpositions is set to True
I have recently noticed that `edit_distance(s1, s2, transpositions=True)` in `nltk.metrics.distance` does not compute the Damerau-Levenshtein distance between `s1` and `s2` -- i.e. the minimal number of operations among insertion,  deletion, substitution and transpositions required to transform `s1` into `s2`. As an example, 

`edit_distance("ca", "abc", transpositions=True)` returns 3, where the distance is actually 2 (it can be achieved by first transposing c and a, then inserting b in the middle).

I am unsure whether this is intended behavior (apologies if it is). It seems that what the function computes in its present state, is what is referred to as 'Optimal String Alignment Distance' on Wikipedia (https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance) -- which imho does not really have a natural and appealing interpretation. As a side note, despite its name, Optimal String Alignment is not a distance in the mathematical sense of the term (it does not satisfies the triangle inequality) which contradicts a comment in the heading of `nltk.metrics.distance`. 

Best,
Antoine

avena554:
This is a (rather quick) fix to #2734 () -- assuming one wants `edit_distance(s1, s2, transpositions=True)` function in nltk to computes the Damerau-Levenshtein distance rather than Optimal String Alignment, this does it (with minimal changes).

Best,
Antoine


stevenbird:
@avena554 thanks for your contribution, and sorry for the delay!
Could I trouble you to please provide some test cases in a new file:
`nltk/develop/nltk/test/unit/test_distance.py`

avena554:
Dear  @stevenbird ,

sorry for my own response delay. I added a couple unit tests checking the relevant behavior, though very basic ones, at the requested location.

Best,
Antoine

tomaarsen:
Hello @avena554,

Thank you for providing some tests for us to look at. I took some time to modify the tests and add some more, and ran into the following quirk, which I believe is not quite right. A simple program to reproduce the issue is the following:
```python
from nltk.metrics import edit_distance

print(edit_distance("wants", "wasp", transpositions=True))
print(edit_distance("wasp", "wants", transpositions=True))
```
which outputs:
```python
2
3
```
I might be mistaken, but it should not be possible to change `"wants"` into `"wasp"` with just 2 modifications, right? Even when transpositions are allowed.

---

For comparison, this same program outputs
```python
3
3
```
on the develop branch.

avena554:
Dear @tomaarsen ,

Thanks a lot, you are definitively right. There is something wrong with my changes too. The "abc" -> "ca" example is done right, but there is something quite wrong with my changes. 

I'll try and fix it.

Best,
Antoine

tomaarsen:
@avena554 Would a possible problem be that `d` in https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance (which is called `lev` in our code), is still instantiated to have dimensions `length(a)+1, length(b)+1`, instead of the correct `length(a)+2, length(b)+2`?
As you can see in the Wikipedia algorithm, `d` is -1 indexed, and the first row and column (labeled as -1 in the pseudocode) should be `maxdist`. 

I believe this detail was not updated in the conversion from the Optimal String Alignment algorithm to the correct Damerau-Levenshtein edit-distance. Updating it will likely cause some nasty off-by-one errors though.

avena554:
Hi @tomaarsen ,

the main problem seems to be more stupid than that -- from what I see your counterexamples  "wants" ->  "wasp" fails in my code even with the parameter `transpositions=False`. I wrongly believed to had made the minimal changes so that it would bear no change to the algorithm when the transpositions flag is not set. I just printed some debug info, and it seems indeed that even when "transpositions=False" the algorithm ends up choosing its final value with a `min(a, b, c, d)` in  `edit_distance_step` where `c = 2` and `d = 1`. This is unexpected since we should have `d = c + 1` always when transpositions is set to `False`. 

Of course this does not exclude another mistake of mine in the direction your pointing -- I'll have to think carefully about the offsets to answer -- but for the problem at hand I tend to believe it must be something quite stupid (e.g. a bad ternary if condition) :).

tomaarsen:
@avena554 I'm not sure the output is also incorrect with `transpositions=False`. If I run the following program:
```python
from nltk.metrics import edit_distance
print(edit_distance("wants", "wasp", transpositions=False))
```
Then I get `3`, which seems to be correct? For example:
```python
# "wants" -> "wats" -> "was" -> "wasp"
# Two deletions followed by an insertion
```
That said, the way it finds that output might still be incorrect.

avena554:
@tomaarsen 

Yes, this was nonsense from me sorry.

I am in a bit of a rush now, but I'll be able to look at your suggestion in a couple hours !

Thanks !

tomaarsen:
@avena554 Wonderful, I'm curious to see what caused the issue!

avena554:
Hello again @tomaarsen ,

I think I have now fixed the problem. I don't think there was any array size or indices offset problems (though I haven't extensively tested for that). What caused the mistake was a mix up between the first and the second argument strings towards the end of the main loop in `edit_distance`, (`s2` instead of `s1`). The thing is, `last_left_t` is supposed to keep track, for every symbol in the alphabet, of the index at which it was last seen in `s1`. Now I updated it based on `s2` instead. I guess it was luck or (lack thereof) that it worked on the abc -> ca example :)

I have refactored the unit tests accordingly (you said you added some of yours, I don't really know how to merge those with my branch, sorry about that). Many thanks again for your vigilance, I hope it works now.

tomaarsen:
@avena554 I'll have a little look when I have a moment! And don't worry, I believe I have permissions to add commits to this PR, so I can just add my tests myself, if that's okay with you.

tomaarsen (approved):
@avena554 I manually determined some simple test cases, and they all seem to pass. I have some confidence that the program is now correct. Thank you for the fix!

For the record, if I copy the new tests to the develop branch, then the following fail:
```python
FAILED nltk\test\unit\test_distance.py::TestEditDistance::test_with_transpositions[abc-ca-1-expecteds0] - assert 3 == 2
FAILED nltk\test\unit\test_distance.py::TestEditDistance::test_with_transpositions[abc-ca-5-expecteds1] - assert 3 == 2
FAILED nltk\test\unit\test_distance.py::TestEditDistance::test_with_transpositions[wants-swim-2-expecteds15] - assert 7 == 6
```
The remainder pass for the develop branch too.

avena554:
@tomaarsen 

Thanks for checking !

So if the function's correct, did I mess up something with my tests, or do you mean than adding the new tests to the develop branch -- before adding the fix in `edit_distance` -- some tests using transpositions fail, which was expected ?

If that's the former, sorry about that !

tomaarsen:
@avena554 The latter. Some of the new tests to the develop branch (before adding the fix in `edit_distance`) fail. This was indeed to be expected.
Your tests were correct, I believe.

stevenbird:
Great work @avena554 and @tomaarsen!

PR #2737, opened 2021-06-25
add broken-heart emoj and my name on contributors list

12mohaned:
Add a new emoji to the emoticons list.

### DIFF
--- a/nltk/tokenize/casual.py
+++ b/nltk/tokenize/casual.py
@@ -67,6 +67,7 @@ EMOTICONS = r"""
       [<>]?
       |
       <3                         # heart
+      </3                        # Broken-heart
     )"""
 
 # URL pattern due to John Gruber, modified by Tom Winzig. See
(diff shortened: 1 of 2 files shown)


nltk/tokenize/casual.py
```diff
@@ -67,6 +67,7 @@
       [<>]?
       |
       <3                         # heart
+      </3                        # Broken-heart
```
tomaarsen: This does not work. It will try to match `<3</3`. Merging this would prevent the TweetTokenizer from recognising hearts.

What could work is 
```python
...
|
<3                   # heart
|
</3                  # broken heart
```
But at that point, just use regex to modify the existing heart:
```python
...
|
</?3                 # heart, perhaps broken
```

tomaarsen (changes requested):

nltk/tokenize/casual.py
12mohaned: Yes, I totally forgot that, thanks!

stevenbird:
Thanks @12mohaned, @tomaarsen

Issue #2738, opened 2021-06-26
Use Graphviz for Wordnet graph visualisations


ekaf:
Currently, NLTK can draw the tree of hypernyms of one sense of "person" at a time, but it would be difficult to merge those trees into one graphic like the following interesting "Figure 1: Hypernyms of the word person (all senses)", extracted from the article "WordNet and Prolog: why not?",  Julian-Iranzo and Saenz-Perez, 2019 Conference of the International Fuzzy Systems Association and the European Society for Fuzzy Logic and Technology (EUSFLAT 2019).

![person-dot](https://user-images.githubusercontent.com/4782556/123507839-0abe7c80-d66c-11eb-8691-5909fa987436.png)

However, it turns out that the authors used a pretty simple approach: collect the set of edges of this graph, print that set in the format expected by the "dot" program from the "graphviz" package, and then rely on "dot" to draw the graph.

The same approach is quite easy to implement in NLTK, since a few NLTK modules (api, chart, dependencygraph) already interface with the "dot" program. Wouldn'it be nice to have this kind of more powerful graphical capabilities available for Wordnets also?

tomaarsen:
Implemented in 8f91325, thank you for the idea and contribution!

PR #2685, opened 2021-04-04
Added Tree.fromlist()

ekaf:
Solves issue #2684

The new class method Tree.fromlist() can be used to parse trees that are expressed as nested lists, such as those produced by the tree() function from the wordnet module.

This allows to use nice functionality provided by the NLTK Tree class:

    >>> from nltk.corpus import wordnet as wn
    >>> t=Tree.fromlist(wn.synset('dog.n.01').tree(lambda s:s.hypernyms()))
    >>> print(t.height())
    14
    >>> print(t.leaves())
    ["Synset('entity.n.01')", "Synset('entity.n.01')"]
    >>> t.pretty_print()
                      Synset('dog.n.01')                  
             _________________|__________________          
    Synset('canine.n.                            |        
           02')                                  |        
            |                                    |         
     Synset('carnivor                            |        
         e.n.01')                                |        
            |                                    |         
     Synset('placenta                            |        
         l.n.01')                                |        
            |                                    |         
    Synset('mammal.n.                            |        
           01')                                  |        
            |                                    |         
     Synset('vertebra                            |        
        te.n.01')                                |        
            |                                    |         
    Synset('chordate.                     Synset('domestic
          n.01')                           _animal.n.01') 
            |                                    |         
    Synset('animal.n.                    Synset('animal.n.
           01')                                 01')      
            |                                    |         
    Synset('organism.                    Synset('organism.
          n.01')                               n.01')     
            |                                    |         
     Synset('living_t                     Synset('living_t
       hing.n.01')                          hing.n.01')   
            |                                    |         
     Synset('whole.n.                     Synset('whole.n.
           02')                                 02')      
            |                                    |         
    Synset('object.n.                    Synset('object.n.
           01')                                 01')      
            |                                    |         
     Synset('physical                     Synset('physical
      _entity.n.01')                       _entity.n.01') 
            |                                    |         
    Synset('entity.n.                    Synset('entity.n.
           01')                                 01')      
    >>> t.draw()
dog

PR #2740, opened 2021-06-30
Visualize WordNet relation graphs using Graphviz

ekaf:
Solve #2738.

Introduce the new 'digraph' function in the wordnet module, to produce a graphical representation of a Wordnet synset relation, for drawing with the 'dot' graph visualisation program from the Graphviz package. According to the 'dot' documentation, "the layouts are not necessarily stable".

Output a string in the DOT graph file language, which can then be converted to an image (defaulting to the SVG format), by _nltk.parse.dependencygraph.dot2img(dot_string)_,  eventually using the optional 't' argument to specify another image file format, for ex.  'png' or 'jpeg' (Running 'dot -T:' lists all available formats).

The input is a list of one or more Synsets, Lemmas and/or words, eventually mixed. If inputs are words or Lemmas,
their respective Synsets are collected, before applying the specified synset relation, which defaults to _hypernyms_.

Unlike trees, directed graphs expose shared stucture, like the common path from Synset('animal.n.01') to Synset('entity.n.01'), which is not immediately apparent in the corresponding tree (cf. #2685).

For example:

> import nltk
> from nltk.corpus import wordnet as wn
> from nltk.parse.dependencygraph import dot2img
> dot2img(wn.digraph([wn.synset('dog.n.01')]), t='png')

doghyp

The optional 'attr' parameter is a dictionary of global graph attributes ('BT' specifies a "bottom to top" orientation, 'LR draws from left to right; 'dot' uses 'TB'  as default). 'shapes' is a dictionary of strings that trigger the specified shapes:

> dot2img(wn.digraph(['person'], shapes={'Lemma':'box','Synset':'plaintext'}, attr={'rankdir':'BT'}), t='png')

person

Eventually, draw other relations than the default hypernyms. The cycles in the also_sees() relation are discarded by nltk.util.edge_closure().

> dot2img(wn.digraph([wn.synset('bound.a.01')], lambda s:s.also_sees()), t='png')

Most (but not all) _also_sees_ links are bidirectional. Note that more layouts are possible with bidirectional links, and 'dot' outputs _one_ of them arbitrarily:

bound1

In another run of the previous example, 'dot' might draw the same graph  with a different layout:
bound2

By contrast, 'dot' seems to draw a completely stable layout of the corresponding Minimum Spanning Tree (cf. #2646): 

> from nltk.util import unweighted_minimum_spanning_digraph as umsd
> dot2img(umsd(wn.synset('bound.a.01'), lambda s:s.also_sees(), shapes={'':'plaintext'}), t='png')

boundumsd

Eventually, specify more input nodes. The 'maxdepth' option limits the longest path:

> dot2img(wn.digraph([wn.synset('cat.n.01'),wn.synset('dog.n.01')], maxdepth=3), t='jpeg')

catdog

NB: the generic graph drawing functions are outside the WordNet module, for use with other graphs as well:

>>> from nltk.util import edges2dot,edge_closure
>>> dot2img(edges2dot(edge_closure('A', lambda x:{'A':'BC', 'B':'AC', 'C':''}[x])), t='png')

ABC


ekaf:
Note: the new _dot2img_ function uses sys.stdout instead of subprocess.PIPE, to bypass a very common encoding/decoding problem in Python's _subprocess_ module.

The problem seems to be that the _subprocess_ module enforces input and output to be in the _same_ encoding, which cannot work here, since the input string is in 'utf-8', while the image writers for jpeg and png produce output that cannot be decoded as 'utf-8'. So using subprocess.PIPE leads to the following error for PNG:

> UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte

and for JPEG:

> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

Using sys.stdout instead solves this problem, since according to the _subprocess_ documentation,  "By default, file objects are opened in binary mode".

nltk/corpus/reader/wordnet.py
```diff
@@ -2051,7 +2051,7 @@ def custom_lemmas(self, tab_file, lang):
     ######################################################################
 
 
-    def digraph(self, inputs, rel=lambda s:s.hypernyms(), pos=None, o='default', maxdepth=-1, boxes=[], verbose=False):
+    def digraph(self, inputs, rel=lambda s:s.hypernyms(), pos=None, maxdepth=-1, shapes=dict(), attr=dict(), verbose=False):
```
tomaarsen: `dict()` is a dangerous default value for an argument. The reason for this, is that this `dict` is only created once, and used between all calls to the method. If this `dict` is updated once in call 1 to the method, then the dictionary is still updated in the second call to the method. Wonderful for caching, horrible for all other situations. It can cause some *really* confusing bugs. 

This is generally avoided with:
```python
def digraph(self, ..., shapes=None, attr=None):
    """
    ...
    """
    if not shapes:
        shapes = dict()
    if not attr:
        attr = dict()
    ...
```

nltk/util.py
```diff
@@ -253,17 +253,16 @@ def edge_closure(tree, children=iter, maxdepth=-1, verbose=False):
                 pass
 
 
-def edges2dot(edges, boxes=[], o='default'):
+def edges2dot(edges, shapes=dict(), attr=dict()):
```
tomaarsen: Same as with `digraph`.

nltk/util.py
```diff
@@ -277,15 +276,13 @@ def edges2dot(edges, boxes=[], o='default'):
...
     """
     dot_string = 'digraph G {\n'
-    for (source,target) in edges:
-        if o == 'reverse':
-            pair = (target,source)
-        else:
-            pair = (source,target)
-        for box in boxes:
+    for tup in attr.items():
+        dot_string += "%s=%s;\n" % tup
```
tomaarsen: The [`str.format()`](https://docs.python.org/3/library/string.html#format-examples) methodology for formatting is preferred over the Python 2 `%`-style.
In fact, now that we've deprecated Python 3.5, we can also start using the wonderful [`f-string formatting`](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings).

See the [CONTRIBUTING guidelines](https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md#code-guidelines).

This also applies a few times below.

tomaarsen (changes requested):
I've added some suggestions for this one commit. I haven't had time to look at the others.

nltk/util.py
ekaf: Thanks @tomaarsen. I have applied all your suggestions.

tomaarsen (approved):

stevenbird:
@ekaf great contribution!
@tomaarsen thanks for reviewing

PR #2741, opened 2021-07-06
[cleanup] Remove py2-only code

dannysepler:
- All py3 classes inherit from `object` by default
  - scripted via: `find . -name "*.py" -type f | xargs sed -i -e 's/(object)//g'`
- `# -*- coding: utf-8 -*-` happens by default
  - scripted via:  `find . -name "*.py" -type f | xargs sed -i -e '/coding: utf-8/d'`
- Remove py2-only if statements


stevenbird:
@dannysepler – thank you!

PR #2743, opened 2021-07-07
Pytest-ify a few more tests

dannysepler:


nltk/test/unit/test_json2csv_corpus.py
```diff
@@ -9,223 +9,199 @@
...
+
+subdir = Path(__file__).parent / 'files'
+
+@pytest.fixture
+def infile():
+    with open(twitter_samples.abspath("tweets.20150430-223406.json")) as infile:
+        return [next(infile) for x in range(100)]
+
+
+def test_textoutput(tmp_path, infile):
```
dannysepler, line 35: used the [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture instead of TemporaryDirectory. a lot of the subsequent cleanup is to this file from using pathlib functions

stevenbird:
Thanks @dannysepler... really great to have all these clean ups.

PR #2746, opened 2021-07-10
Check out this file and Create WSD_LeskUpgrade.py

DASHANANT:
This is upgradation of standard Lesk Algorithm For Word sense Disambiguation


stevenbird:
I'm sorry @DASHANANT... in the case of established methods we are only taking bug fixes.

DASHANANT:
Okay , but is it working correctly?

Issue #2749, opened 2021-07-21
travis-ci.org no longer runs builds


purificant:
![travis_ci_org_eol](https://user-images.githubusercontent.com/4669013/126481562-e9d7475f-3c1f-499c-9416-c04b8514d793.png)
The above message is currently displayed on https://travis-ci.org/github/nltk/nltk

travis-ci.org no longer runs ci pipeline for nltk, they suggest migrating to travis-ci.com
At the time of writing there are no CI builds running for latest pull requests and merges.

Possible courses of action:
1. Migrate from travis-ci.org to travis-ci.com
2. Use github actions to run CI pipeline and builds

tomaarsen:
Continuous Integration is now handled through GitHub Actions instead of Travis CI.

Issue #2744, opened 2021-07-07
Add pre-commit hooks?


dannysepler:
Generally curious if we've given thought into this. A lot of repos are now using pre-commit hooks to keep code nice and consistent.

- We could dip our toes in the water with some of these https://github.com/pre-commit/pre-commit-hooks -- they generally don't change the code much

The following two are noisier to set up but I really like them both:
- [Pyupgrade](https://github.com/asottile/pyupgrade/)
- [isort](https://github.com/PyCQA/isort/issues)

iliakur:
I'm in favor, especially for auto-formatting like `isort` and `black`. I haven't used `pyupgrade`, can't say if its noisiness is worth it.

I think if you search the google group there's a thread about this as well.

tomaarsen:
I've personally used this as well, and support the inclusion of this in NLTK. In particular I like the idea of an auto-formatter, but If that's a bridge too far (for now?), then I can recommend the `trailing-whitespace` trimmer from https://github.com/pre-commit/pre-commit-hooks.

dannysepler:
thanks both! i see the [relevant post](https://groups.google.com/g/nltk-dev/c/G7Iz-bjWtqk/m/DMw58fQqAQAJ) in the google group but do not have permissions to post to it?

i agree with your point, Tom. We can start with lighter ones like `trailing-whitespace` and perhaps also `end-of-file-fixer`, document how to do pre-commit stuff for contributors, and then do a noisier one if it goes okay?

stevenbird:
@iliakur would you be interested in managing this? What permissions would you need?

iliakur:
I'm interested, but wary of the commitment. My first-born is 3 weeks old and I won't be able to reliably find time to look into this for a few more weeks, maybe even more.

stevenbird:
@dannysepler I like your suggestion

PR #2750, opened 2021-07-21
implement continuous integration pipeline based on github actions

purificant:
This PR implements CI pipeline based on Github Actions. It is intended to supplement and possibly replace existing Travis integration that is not working at the moment, see https://github.com/nltk/nltk/issues/2749 for details.

Notable differences from Travis:
- faster runtime
- caches nltk_data directory saving gigabytes of downloads on every run
- caches python dependencies
- unifies pipeline package dependencies into a single requirements file instead of specifying them in multiple config files
- runs tests in parallel using number of workers based on available cpu cores
- runs tests on macos in addition to linux

Possible future work that can build on this:
- run tests on windows
- automate code linting with black / pylint
- test third party dependencies: stanford nlp / parser / postagger

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,60 @@
+name: ci-workflow
+# run workflow for these events
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  cache_nltk_data:
+    name: cache nltk_data
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Cache nltk data
+        uses: actions/cache@v2
+        with:
+          path: ~/nltk_data
+          key: nltk_data
+
+      - name: Download nltk data packages
+        run: |
+          pip install regex # dependencies needed to download nltk data
+          python -c "import nltk; nltk.download('all')"
+        shell: bash
+
+  test:
+    name: Python ${{ matrix.python-version }} on ${{ matrix.os }}
+    needs: cache_nltk_data
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          path: ~/nltk_data
+          key: nltk_data
+
+      - name: Download nltk data packages
+        run: python -c "import nltk; nltk.download('all')"
+        shell: bash
+
+      - name: Run pytest
+        run: pytest
```
dannysepler: can you change this to `pytest nltk/test`? for some reason that catches both doctests and unit tests, whereas running this way only catches unit tests

dannysepler:
so cool! i was wondering how to make the CI faster

if we were gonna do black, another route could be via [pre-commit](https://github.com/nltk/nltk/issues/2744) so that it's automatic

.github/workflows/ci.yaml
purificant: Good call

Running `pytest` will use a different root directory for running tests and will not include configuration from `pytest.ini` which is what allows doctests to run when using `pytest nltk/test`. I was originally going to leave doctests as an item for future work but have now implemented running doctests in the current version of this PR.

Also added caching for python dependencies and added parallel execution for tests.

stevenbird:
Thanks @purificant and @dannysepler. Maintaining our CI has been an ongoing hassle, so I'm excited that there may be a way to do it within github itself

.github/workflows/ci.yaml
dannysepler: So awesome! Great job

purificant: Thanks! :heart:

PR #2753, opened 2021-07-24
Initial pre-commit hooks and documentation

dannysepler:
Closes https://github.com/nltk/nltk/issues/2744

I'd like to try a few simple pre-commit hooks with some documentation, see if it goes okay, and then add more

PR #2754, opened 2021-07-25
revert whitespace changes that affected tests

purificant:
https://github.com/nltk/nltk/pull/2753 had an unfortunate side effect on a few tests, this is due to test code expecting files on disk to be formatted in a specific way. As a workaround, this PR reverts white space changes to test data files in order for tests to pass.

More details here: https://github.com/nltk/nltk/actions/runs/1063540909


dannysepler:
Oh man. Thanks. I'm a little iffy on this since the CSV files will just get the white space removed next time they're edited. I may look into making the test itself better

purificant:
@dannysepler Yeah, this is a workaround to make CI build green. It may be possible that a smaller diff is sufficient to pass the tests, or that some other change like a different data format for test data files can resolve this, or changing the nature of the tests themselves.

PR #2756, opened 2021-07-25
cache improvements, enable gensim tests, more verbose test output for skipped and failed tests

purificant:
A few small updates to Github Actions CI:

- enable more verbose output for skipped and failed tests with `pytest -rsx`, this is useful when a test is skipped due to a missing dependency or third party package
- enable test that requires gensim package
- skip cache regeneration step on cache hit, this further speeds up the build with warm cache


stevenbird:
Excellent!

PR #2761, opened 2021-07-28
Replaced Travis info with GitHub Actions info in CONTRIBUTING.md

tomaarsen:
Documents changes from #2750.

---

Hello!

### Pull request overview
* Replaced explanation of `.travis.yml` with a short overview of `.github/workflows/ci.yaml`, mentions deprecation of Travis as our CI server.

---

### Note
The CI workflow provided by @purificant runs tests on all *officially supported* versions of Python, i.e. (currently) 3.6, 3.7, 3.8 and 3.9. However, this clashes with our `setup.py` and our `tox.ini`, both of which still mention support for 3.5.

We need to make a call. Some options are:
* We drop 3.5 support. This would be in accordance to the recommendation by the Python team to update away from 3.5.
* We keep 3.5 support, allowing developers to make their own call on which version they use, even when that choice counters the advise from the Python team.

I'm personally very much in favour of the first option. I'd like some feedback on that. I should have some time to work on whichever option is preferred, but the current situation of *"Our setup.py supports Python 3.5, but we don't run tests for it"* is definitely not going to work.

- Tom Aarsen

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -158,19 +158,19 @@ NLTK uses [GitHub Actions](https://github.com/nltk/nltk/actions) for continuous 
 The [`.github/workflows/ci.yaml`](https://github.com/nltk/nltk/blob/develop/.github/workflows/ci.yaml) file configures the CI:
 
  - `on:` section
-   - Ensures that this CI is run on code pushes, pull request, or through the GitHub website via a button.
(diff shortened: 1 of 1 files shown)


stevenbird:
@tomaarsen yes it's time to drop Python 3.5 support

purificant (approved):

stevenbird:
Thanks @tomaarsen, @purificant

PR #2763, opened 2021-07-28
Fixed large error in METEOR score, see #2751

### LINKED ISSUE #2729: Meteor unmatched-lists calculation
Hi, 

I think [unmatched-lists calculation by _enum_stem_match()](https://github.com/nltk/nltk/blob/e4444c9b15762e6d86f5f6c4f5faadb87632c72a/nltk/translate/meteor_score.py#L109) is wrong,
```
enum_unmat_hypo_list = (
        list(zip(*enum_unmat_hypo_list)) if len(enum_unmat_hypo_list) > 0 else []
    )
enum_hypothesis_list = list(
        filter(lambda x: x[0] not in enum_unmat_hypo_list, enum_hypothesis_list)
    )
```
because the upper zip(*) generates [(unmatched indices), (unmatched stems)]
but what we want is to filter by (unmatched indices).
Also, "not in" would filter **out** unmatched, not the opposite.

When I fix the calculation to:
```
enum_unmat_hypo_list = (
        list(zip(*enum_unmat_hypo_list))[0] if len(enum_unmat_hypo_list) > 0 else []
    )
enum_hypothesis_list = list(
       filter(lambda x: x[0] in enum_unmat_hypo_list, enum_hypothesis_list)
    )
```

the score below is 0.655 with matches = [(0, 0), (1, 2), (2, 3)]
(Before fixing, it was 0.809 with matches = [(0, 0), (1, 2), (1, 2), (2, 3)] which contains duplicated elements)
```
hypothesis = 'I likes pizza'
reference = 'I really like pizza'
single_meteor_score(reference, hypothesis)
```

I would like to know if this is wrong.
Thanks.

### LINKED ISSUE #2751: Is not the max of meteor score 100?  (nltk.translate.meteor_score)
I used [nltk.translate.meteor_score](https://github.com/nltk/nltk/blob/develop/nltk/translate/meteor_score.py) to calculate the meteor score of a example, strangely I got a score of 105.6. 
Is not the max of meteor score 100?

The example is following:

> **candidate:**   creates or updates a virtual machine scale set
**reference:**   create or update a vm scale set
**METEOR score:**   105.6

Is there anyone know that? Thank you very much!

tomaarsen:
Fixes #2751, fixes #2729

---

Hello!

### Pull request overview
* Fixed `_enum_stem_match`, which previously did not remove matched words from `enum_hypothesis_list` and `enum_reference_list`. This caused very incorrect METEOR scores!
* Added a new test with the example reference and hypothesis from #2751 by @treeAndRiver in `nltk/test/meteor.doctest`.
  * This test failed previously, but passes now.
* Pass `wordnet` from `single_meteor_score` to `_enum_align_words`. (Previously, giving a different value for the `wordnet` parameter did nothing)
* Fixed typos in `_enum_allign_words` and `allign_words` (!). These are now (correctly) called `align_...`.
* Fixed a few typos in comments and docstrings.
* Removed an unnecessary import
* Added myself as a contributor to the meteor_score.py file.

### In short:
**One of the functions used in the METEOR score will try to "match" words together, but failed to remove these newly matched words from the list of yet-to-match words, causing another matching method to find the same matches, resulting in duplicates matches.**
**This produced very incorrect METEOR scores whenever sentences include words which nearly match, such as `"create"` and `"creates"`.**
This PR has fixed this issue.

---

### Overview
The bug itself is mentioned in #2751. @treeAndRiver mentions using a candidate (hypothesis) of:
>  creates or updates a virtual machine scale set

and a reference of
>  create or update a vm scale set

and reports getting a METEOR score over 100 (percent).

The reasoning for the bug is also mentioned in #2729.

---

### Reproduction
```python
from nltk.translate.meteor_score import single_meteor_score, meteor_score

reference = "create or update a vm set"
hypothesis = "creates or updates a virtual machine scale set"

print(single_meteor_score(reference, hypothesis))
```

---

### Output
`1.023699802501646`

This should not be a possible output, as identical sentences should give the highest possible score of 1.

---

### Correct output
First of all, the METEOR score that is implemented is one of the earlier versions, following [this paper](http://www.cs.cmu.edu/~alavie/METEOR/pdf/Lavie-Agarwal-2007-METEOR.pdf).
This algorithm is quite simple, and we can follow it by hand to get the result that we *should* be getting:
```python
# Given parameters
alpha=0.9
beta=3
gamma=0.5

# The input values
reference = "create or update a vm set"
hypothesis = "creates or updates a virtual machine scale set"

# How many words each of the input values have
reference_length = len(reference.split()) = 6 # 6 words
hypothesis_length = len(hypothesis.split()) = 8 # 8 words

# Now, the interesting part, the "matches". 
# Here, the indices of matching words between the reference and hypothesis are listed.
matches = [
(0, 0) # create creates
(1, 1) # or or
(2, 2) # update updates
(3, 3) # a a
(7, 5) # set set
]
# As can be seen from this, the first 4 words are roughly the same, 
# after which word with index 5 for the reference and index 7 for the hypothesis are the same.

# So, clearly, 5 matches
matches_count = 5

# This gives the following precision and recall
precision = matches_count / hypothesis_length = 5 / 8
recall = matches_count / reference_length = 5 / 6

# And when following the formula from the paper, the fmean is roughly 0.8
fmean = (precision * recall) / (alpha * precision + (1 - alpha) * recall) \
      = ((5 / 8) * (5 / 6)) / (0.9 * (5 / 8) + 0.1 * (5 / 6)) \
      = 0.8064516129032259

# Now, these matches represent 2 chunks (indices 0 to 3 for (roughly) "create or update a",
# and the last match for "set".
chunk_count = 2

# This means the fragment fraction is 0.4
frag_frac = chunk_count / matches_count = 2 / 5
# Which gives this penalty
penalty = gamma * frag_frac ** beta \
        = 0.5 * (2 / 5)**3 \
        = 0.032

# After which the result can easily be computed to be 0.7806
result = (1 - penalty) * fmean \
       = 0.968 * fmean \
       = 0.7806451612903226
```
So, the output *should* be 0.7806, and **definitely** not 1.0237.

This bug can be experienced more heavily for a reference of `"walking running"` and a hypothesis of `"walk run"`. This gives a METEOR score of 1.5781.

---

## The bug, and the fix
The algorithm deviates from my manual calculation at deriving the matches. The matches the current METEOR implementation finds are:
```python
[(0, 0), (0, 0), (1, 1), (2, 2), (2, 2), (3, 3), (7, 5)]
```
As you can see, there are duplicates, and this deviates from my manual findings of
```python
[(0, 0), (1, 1), (2, 2), (3, 3), (7, 5)]
```

The error lies in `_enum_allign_words` (now called `_enum_align_words`):
https://github.com/nltk/nltk/blob/6411fabf05a75e2beb82b29fed614dbd12688f7f/nltk/translate/meteor_score.py#L221-L239

These 3 sections will use the `exact`, `stemming` and `wordnet` methods respectively to match words. If a match is found, e.g. on `(5, "set")` and `(7, "set")`, then this tuple should be removed from `enum_hypothesis_list` and `enum_reference_list`. This way, the method that follows does not find the exact same match. 
**However, the middle method, `_enum_stem_match` fails to do so.**

This method will find the matches of e.g. `(0, "creates")` and `(0, "create")` and `(2, "updates")` and `(2, "update")` (by way of stemming, and then matching exactly) and return them under `stem_matches`. But, because these matches are not removed from `enum_hypothesis_list` and `enum_reference_list`, the next method, `_enum_wordnetsyn_match`, will find the exact same matches!

This is why `(0, 0)` and `(2, 2)` are in the list of matches twice.

The body of the failing function is as follows:
https://github.com/nltk/nltk/blob/6411fabf05a75e2beb82b29fed614dbd12688f7f/nltk/translate/meteor_score.py#L96-L124

Step by step, this method will:
* Stem the hypothesis words
* Stem the reference words
* Match the now-stemmed hypothesis and reference using the exact method.
  Then, it gets a little weird. It will:
* Convert e.g. `[(4, 'virtual'), (5, 'machin'), (6, 'scale')]` to `[(4, 5, 6), ('virtual', 'machin', 'scale')]` (for hypothesis)
* Do the same for the reference.
* Filter `enum_hypothesis_list` (e.g. `[(4, 'virtual'), (5, 'machin'), (6, 'scale')]`) by ensuring that the first value of each tuple in this list is *not* in `enum_unmat_hypo_list` (which might look like `[(4, 5, 6), ('virtual', 'machin', 'scale')]`).
  However, this must always be the case, as we are checking whether an integer is `not in` a list of tuples?
* Do the same for the reference.

So, the filtering is moot. It does not modify `enum_hypothesis_list` whatsoever. Then, the modification of `enum_unmat_hypo_list` from e.g. `[(4, 'virtual'), (5, 'machin'), (6, 'scale')]` to `[(4, 5, 6), ('virtual', 'machin', 'scale')]` is not necessary either.
In fact, I don't know why this was even written, because the goal for this section is to **filter the input of enumerated words to exclude the newly-matched words**, but this functionality is already provided in `_match_enums`. That is *exactly* what `_match_enums` does. So, we can literally return at line 104, just returning the values from `_match_enums`.

This is the core of the fix.

---

#### Additional changes
The PR Overview at the top lists all other small additional changes, but one that should be considered is that `allign_words` was renamed. Perhaps you want to reintroduce the incorrectly written `allign_words` as a wrapper for `align_words`, with a deprecation warning?

Furthermore, a test was introduced based on #2751, which passes.

---

Let me know if you need anything more from me for this.
Also, thank you @treeAndRiver for reporting this issue, and thank you @yunju63 for guiding me in the right direction.

- Tom Aarsen


tomaarsen:
The previous commit resolved a merge conflict caused by a typo in the `meteor_score.py` file that #2764 solved.

stevenbird:
Sorry for delay... hoping to have someone verify first

stevenbird:
I have pinged nltk-dev for someone to verify this PR.
https://groups.google.com/g/nltk-dev/c/vsJY1K1eXDw

stevenbird:
@tomaarsen thanks for the clear explanation, and the clean fix

Issue #2735, opened 2021-06-22
Importing words throws numpy deprecation warning


ihawn:
Warning: 

DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  max_n_alphas=1000, n_jobs=None, eps=np.finfo(np.float).eps

tomaarsen:
NLTK does not use `np.float` directly, only `np.float32` and `np.float64`. 
It seems your warning came from scikit-learn, as `max_n_alphas=1000, n_jobs=None, eps=np.finfo(np.float).eps` can be found [here](https://github.com/scikit-learn/scikit-learn/blob/0fb307bf39bbdacd6ed713c00724f8f871d60370/sklearn/linear_model/_least_angle.py#L1379) in that repo.

purificant:
I wonder if `gaac.py` line 51 could be the culprit

PR #2765, opened 2021-07-28
resolve numpy deprecation warning about numpy.float alias

purificant:
This PR replaces single use of a deprecated numpy alias with an identical equivalent.

Steps to reproduce:
```python
>>> import numpy
>>> numpy.float
<stdin>:1: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```
I spotted this code when looking at https://github.com/nltk/nltk/issues/2735, thanks @ihawn
Details: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

### DIFF
--- a/nltk/cluster/gaac.py
+++ b/nltk/cluster/gaac.py
@@ -48,7 +48,7 @@ class GAAClusterer(VectorSpaceClusterer):
 
         # construct the similarity matrix
         dims = (N, N)
-        dist = numpy.ones(dims, dtype=numpy.float) * numpy.inf
+        dist = numpy.ones(dims, dtype=float) * numpy.inf
         for i in range(N):
             for j in range(i + 1, N):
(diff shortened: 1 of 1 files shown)


tomaarsen:
Oh, apologies, I found no cases of `np.float`, not realising that it might also have been used under `numpy.float`.

tomaarsen (approved):

purificant:
@tomaarsen I believe you are right regarding https://github.com/nltk/nltk/issues/2735, that it's an issue with another package
This PR is to fix an unrelated instance that I noticed, which should be the last deprecated use in nltk codebase

PR #2769, opened 2021-07-31
Prevent `debug_decisions` in Punkt from throwing IndexError

### LINKED ISSUE #2519: PunktSentenceTokenizer doesn't work with custom end sentence character
Example:

```
import nltk
from nltk.tokenize.punkt import PunktLanguageVars

class ExtLangVars(PunktLanguageVars):
    sent_end_chars = ('.', '?', '!', '^')

punkt_tk = nltk.data.load('tokenizers/punkt/english.pickle')
punkt_tk._lang_vars = ExtLangVars()

txt1 = 'Subject: Some subject. Attachments: Some attachemnts'
txt2 = 'Subject: Some subject^ Attachments: Some attachemnts'

r = punkt_tk.tokenize(txt1)
print(r)

r = punkt_tk.tokenize(txt2)
print(r)

d = punkt_tk.debug_decisions(txt1)
for x in d:
    print(nltk.tokenize.punkt.format_debug_decision(x))

#d = punkt_tk.debug_decisions(txt2)
#for x in d:
#    print(nltk.tokenize.punkt.format_debug_decision(x))

```
Also if you uncomment last lines, it will fail with index out of bounds. I think it's because of hardcoded 

`        self.period_final = tok.endswith('.')`

in PunktToken. Tried many ways to work around this, but no luck. Any suggestions would be appreciated.

tomaarsen:
Fixes #2519

---

Hello!

### Pull request overview
* Prevent `debug_decisions` in Punkt from throwing IndexError whenever the passed sentence contained a sentence ending in `?`, `!` or a custom defined end character (under `_lang_vars.sent_end_chars`). 
* Added 4 tests to show that this no longer throws an Exception.
* Modified `nltk/test/unit/test_tokenize.py` to only skip relevant tests if Stanford Segmenter/Tokenizer jars are missing, instead of skipping all tokenizer tests.
* Modernised the code slightly, e.g. update a format string to the `.format()` style.

---

### The bug & the fix
The `IndexError` is thrown in the following two lines, where `tokens` is a list of `PunktToken` instances:
https://github.com/nltk/nltk/blob/1244ab9fc4b130ae20127267d49eea506e30a9ae/nltk/tokenize/punkt.py#L1288-L1289

If no element in this list ends with a period, then the while loop will only terminate through an `IndexError` when `tokens` becomes empty. 
The loop here aims to remove all arbitrary words that happen before the sentence break. This way, only the word before and after the sentence break are shown. However, a sentence break may also occur on a `?`, or a `!`. In these cases, the loop will throw an Exception.

So, instead of discarding until there's a token that ends with a period, it now discards until there's a token that ends with any sent_end_char (by default: `(".", "?", "!")`).
This has removed the issue. However, just to be safe, I also added `while tokens and tokens[0]...`. This way the loop can never throw an `IndexError`.

The bug can be reproduced by running any of the new tests on the old codebase.

---

### Additional changes
I noticed that `nltk/test/unit/test_tokenize.py` does not run any tests whatsoever if the Stanford segmenter/tokenizer jars are not installed. The following function is responsible for this:
https://github.com/nltk/nltk/blob/1244ab9fc4b130ae20127267d49eea506e30a9ae/nltk/test/unit/test_tokenize.py#L17-L34

(Furthermore, `StanfordTokenizer` is never imported?)

I've modified this slightly to create a `pytest.mark.skipif` object (See [here](https://docs.pytest.org/en/latest/how-to/skipping.html#id1) for documentation) which is used on the two methods that test the Stanford Segmenter. The `try-except` around these two methods were also removed as they are no longer needed. Mentions of the Stanford Tokenizer are removed, as they weren't actually tested. 

The main goal of these changes is to keep testing the other tokenizers (tweet, punkt, etc.) even if the Stanford files are not installed.
@dannysepler I know this is your domain, so I'm asking for you to have a quick look at this. 

I also updated some small things, like moving to `.format()` formatting, 

- Tom Aarsen

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -25,7 +25,7 @@ def load_stanford_segmenter():
         return False
 
 
-stanford_segmenter = pytest.mark.skipif(
+check_stanford_segmenter = pytest.mark.skipif(
     not load_stanford_segmenter(), reason="NLTK was unable to find stanford-segmenter.jar."
 )
 
@@ -71,7 +71,7 @@ class TestTokenize:
         tokens = tokenizer.tokenize(test_word)
         assert tokens == ['won', 'der', 'ful']
 
-    @stanford_segmenter
+    @check_stanford_segmenter
     def test_stanford_segmenter_arabic(self):
         """
         Test the Stanford Word Segmenter for Arabic (default config)
@@ -95,7 +95,7 @@ class TestTokenize:
             'المشكلات',
         ]
 
-    @stanford_segmenter
+    @check_stanford_segmenter
     def test_stanford_segmenter_chinese(self):
         """
         Test the Stanford Word Segmenter for Chinese (default config)
@@ -468,7 +468,7 @@ class TestTokenize:
         tokenizer = punkt.PunktSentenceTokenizer()
         txt = "Subject: Some subject. Attachments: Some attachemnts"
 
-        # Test that the sentence is split into two sections, 
+        # Test that the sentence is split into two sections,
         # with one split and hence one decision.
         assert len(tokenizer.tokenize(txt)) == 2
         assert len(list(tokenizer.debug_decisions(txt))) == 1
@@ -478,16 +478,16 @@ class TestTokenize:
         tokenizer = punkt.PunktSentenceTokenizer()
         txt = "Subject: Some subject! Attachments: Some attachemnts"
 
-        # Test that the sentence is split into two sections, 
+        # Test that the sentence is split into two sections,
         # with one split and hence one decision.
         assert len(tokenizer.tokenize(txt)) == 2
         assert len(list(tokenizer.debug_decisions(txt))) == 1
 
     def test_punkt_debug_decisions_custom_end(self):
-        # Test debug_decisions on a text with two sentences, 
+        # Test debug_decisions on a text with two sentences,
         # split by a custom end character, based on Issue #2519
         tokenizer = punkt.PunktSentenceTokenizer()
-        
+
         class ExtLangVars(punkt.PunktLanguageVars):
             sent_end_chars = ('.', '?', '!', '^')
 
@@ -495,13 +495,13 @@ class TestTokenize:
 
         txt = "Subject: Some subject^ Attachments: Some attachemnts"
 
-        # Test that the sentence is split into two sections, 
+        # Test that the sentence is split into two sections,
         # with one split and hence one decision.
         assert len(tokenizer.tokenize(txt)) == 2
         assert len(list(tokenizer.debug_decisions(txt))) == 1
 
     def test_punkt_debug_decisions_no_split(self):
-        # Test debug_decisions on a text with one sentences, 
+        # Test debug_decisions on a text with one sentences,
         # which is not split.
         tokenizer = punkt.PunktSentenceTokenizer()
 
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -499,3 +499,14 @@ class TestTokenize:
         # with one split and hence one decision.
         assert len(tokenizer.tokenize(txt)) == 2
         assert len(list(tokenizer.debug_decisions(txt))) == 1
+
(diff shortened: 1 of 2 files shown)


purificant (approved):

purificant:
Like how this now runs more tests instead of skipping all. :+1:

stevenbird:
@tomaarsen - I really appreciate the solid documentation and careful testing.

nltk/test/unit/test_tokenize.py
```diff
@@ -14,24 +14,20 @@
...
-        StanfordTokenizer()
+        return True
     except LookupError:
-        pytest.skip(
-            "Tests for nltk.tokenize.stanford are skipped because the stanford postagger jar doesn't exist"
-        )
+        return False
+
+
+stanford_segmenter = pytest.mark.skipif(
```
iliakur: I would suggest naming this `check_stanford_segmenter`, that makes it clearer what it does.

tomaarsen: That's a great idea. The current name is quite confusing. I've made the change, thank you for the suggestion.

nltk/test/unit/test_tokenize.py
```diff
@@ -470,3 +462,51 @@ def test_punkt_tokenize_no_custom_lang_vars(self):
...
+        tokenizer._lang_vars = ExtLangVars()
+
+        txt = "Subject: Some subject^ Attachments: Some attachemnts"
+
+        # Test that the sentence is split into two sections,
+        # with one split and hence one decision.
+        assert len(tokenizer.tokenize(txt)) == 2
+        assert len(list(tokenizer.debug_decisions(txt))) == 1
+
+    def test_punkt_debug_decisions_no_split(self):
```
dannysepler: optional: you could do `test_punkt_debug_decisions_no_split`, `test_punkt_debug_decisions_exclamation_mark`, and `test_punkt_debug_decisions_dot` as one test with `pytest.mark.parametrize`

nltk/test/unit/test_tokenize.py
```diff
@@ -470,3 +462,51 @@ def test_punkt_tokenize_no_custom_lang_vars(self):
         expected = ["উপরাষ্ট্রপতি শ্রী এম ভেঙ্কাইয়া নাইডু সোমবার আই আই টি দিল্লির হীরক জয়ন্তী উদযাপনের উদ্বোধন করেছেন। অনলাইনের মাধ্যমে এই অনুষ্ঠানে কেন্দ্রীয় মানব সম্পদ উন্নয়নমন্ত্রী শ্রী রমেশ পোখরিয়াল ‘নিশাঙ্ক’  উপস্থিত ছিলেন। এই উপলক্ষ্যে উপরাষ্ট্রপতি হীরকজয়ন্তীর লোগো এবং ২০৩০-এর জন্য প্রতিষ্ঠানের লক্ষ্য ও পরিকল্পনার নথি প্রকাশ করেছেন।"]
 
         assert obj.tokenize(sentences) == expected
+
+    def test_punkt_debug_decisions_dot(self):
+        # Test debug_decisions on a text with two sentences, split by a dot.
+        tokenizer = punkt.PunktSentenceTokenizer()
+        txt = "Subject: Some subject. Attachments: Some attachemnts"
```
dannysepler: spelling of `attachemnts`

dannysepler:
agree with everyone -- great idea, and very thoughtful work

nltk/test/unit/test_tokenize.py
tomaarsen: I've done so! Thank you for the reminder and recommendation.

nltk/test/unit/test_tokenize.py
tomaarsen: I've resolved this. The typo was already present in the example I took it from, but I suppose that the issue (obviously) still persists with the non-typo version. So, it's probably preferable to fix the typo, rather than stay true to the original example.

tomaarsen:
I've applied your suggestions and quickly ran black over the `test_tokenize` file.

stevenbird:
@tomaarsen - great work!

PR #2773, opened 2021-08-03
Ran black on codebase

tomaarsen:
Hello!

### Pull request overview
- Ran `black` reformatter on entire codebase. (edited: `version 21.7b0`)

---

Running `black` has been done previously, and hence only files that have been modified in the last three years have seen changes (or if `black` has changed its formatting rules somewhat in the last three years)

### What now?
If we decide to merge this, then this might be a good opportunity to add `black` to `pre-commit`. This way, whenever a developer pushes a commit, their code will be automatically formatted according to `black`, if they hadn't done so already. (This is, assuming they use `pre-commit`, which is now recommended within NLTK)
Making this addition to `pre-commit` is preferably done soon after the entire codebase is `black`ed, as that means only newly written code will be modified with the `pre-commit`. 
However - whether we want to add `black` to `pre-commit` at all is the question that remains.

And now that we're on the topic:

Other pre-commit hooks that seem useful are:
- `debug_statements`: Check for debugger imports and py37+ `breakpoint()` calls in python source.
- `no-commit-to-branch`: Protect specific branches from direct checkins.
- Hooks that attempt to load specific files to verify syntax. (`check-xml`, `check-yaml`, `check-toml`, `check-json`)

Some other hooks that have niche uses, but might be nice are:
- `check-case-conflict`: Check for files with names that would conflict on a case-insensitive filesystem like MacOS HFS+ or Windows FAT.
- `check-merge-conflict`: Check for files that contain merge conflict strings.

*Perhaps I should've turned this pre-commit section into an issue instead.*

---

Furthermore, merging this will cause some merge conflicts with currently active pull requests. Luckily, these shouldn't be too complicated.
Alternatively, merging the other PR's before this will also cause merge conflicts. Again, shouldn't be too big of a deal.

---

- Tom Aarsen


purificant:
Out of curiosity, what version of black was used to make this PR?

nltk/corpus/reader/mte.py
```diff
@@ -284,7 +284,7 @@ def paras(self, fileids=None):
 
     def lemma_words(self, fileids=None):
         """
-	    :param fileids: A list specifying the fileids that should be used.
```
purificant, line 287: That's odd ... why would it remove `:`

tomaarsen:
@purificant 
> 
> 
> Out of curiosity, what version of black was used to make this PR?

```
> black --version
black, version 21.7b0
```
Also, I didn't use a custom config file or something. I simply used:
```
black ./nltk
```

According to https://github.com/psf/black/releases, it's the newest release from 18 days ago.

purificant (approved):

purificant:
Looks good. Updates codebase to new default `black` code formatting rules and formats some new code with the same ruleset. :+1:

nltk/corpus/reader/mte.py
tomaarsen: Wow, well spotted. There were a few occassions where black would incorrectly format as it would replace a tab with 8 spaces instead of 4. Or at least, that's what I assume the problem was. For these cases I manually fixed it. In this case, I must've accidentally removed the colon as well, and not realised.

Another common problem was:
```python
                self.random_code_here()
#              this is where we do random code
                self.more_random_code_here()
```
converting to
```python
                self.random_code_here()
                #              this is where we do random code
                self.more_random_code_here()
```

I manually changed these to
```python
                self.random_code_here()
                # this is where we do random code
                self.more_random_code_here()
```

stevenbird:
>  might be a good opportunity to add black to pre-commit

@tomaarsen great idea, I see no reason not to do this

> Perhaps I should've turned this pre-commit section into an issue instead.

@tomaarsen yes please

stevenbird:
@tomaarsen would you mind resolving this merge conflict please?

stevenbird:
thanks @tomaarsen, @purificant

PR #2774, opened 2021-08-03
Added Black as pre-commit hook

tomaarsen:
Hello!

### Pull request overview
- Added `black` as a hook to pre-commit, helping developers format, and preventing them from (accidentally) committing code that does not match formatting standards.

Upon committing, black will run (after the other hooks), and if it wants to modify a file, it will prevent the commit from going through, and modify the code. You can then git add again, and re-commit, now with formatted changes.

Note that all of this only works if the user has done `pre-commit install` at some point, like is now recommended for nltk. Only then will pre-commit do anything.

See #2773 for some discussion of adding `black` to pre-commit. Also see #2744 for more discussion on pre-commit itself.

---

- Tom Aarsen

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,3 +6,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: requirements-txt-fixer
+  - repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black


iliakur:
:+1: from me!

stevenbird:
> Note that all of this only works if the user has done pre-commit install at some point, like is now recommended for nltk. Only then will pre-commit do anything.

For the record, the recommendation is found here:
https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md

stevenbird:
Thanks @tomaarsen, @iliakur!

iliakur:
I think long-term we should run the pre-commit hooks as checks in the CI as well. From experience I wouldn't expect most one-time contributors to consistently set up the hooks. Hosting these hooks can be done on [pre-commit.ci](https://pre-commit.ci/)

PR #2779, opened 2021-08-07
improvement in WSD LESK

DASHANANT:
Usage of wordnet properties Like (Hypernyms And Hyponyms)
-to improve WSD Predictions


stevenbird:
@DASHANANT sorry, we cannot accept this PR. You would need to make a clear case for new functionality and for modifying the behaviour of a long-established module. You would also need to follow PEP-8 style (e.g. the PR leaves comments as orphan strings).

PR #2780, opened 2021-08-09
Apply 'isort' to nltk code

dannysepler:
Noting random things here!

- We can choose a profile we'd like from here https://pycqa.github.io/isort/reference/isort/profiles.html -- i chose "black" but happy to choose another
- One thing I like about isort is the ability to customize options. If there are any behaviors we don't like, we could check this doc https://pycqa.github.io/isort/docs/configuration/options.html and see if we can turn it off
- I skipped a few files where import order seemed to matter with `isort:skip_file`
- Also pre-commit asked me to set `black` to a specific version. I ran `pre-commit autoupdate` and this is the version it came up with


nltk/tbl/__init__.py
```diff
@@ -14,16 +14,15 @@
 currently used by nltk.tag.BrillTagger.
 """
 
+from nltk.tbl.erroranalysis import error_list
```
tomaarsen: Perhaps this file would rather not be sorted? Considering it had comments for each import previously.
That said, the comments aren't particularly useful or anything.

tomaarsen (approved):
Looks great - couldn't find any import issues at a glance, and I way prefer that everything is sorted now. Also, we should probably stick with the `black` profile, as otherwise the black formatter might "disagree" with isort, and they would both try to revert each other's formatting changes all the time. Well done!

nltk/tbl/__init__.py
dannysepler: good catch! haha may as well skip it

stevenbird:
Thanks @dannysepler, @tomaarsen!

Issue #2775, opened 2021-08-03
Suggestions and discussions regarding `pre-commit`


tomaarsen:
Hello!

Now that `pre-commit` has been in place since #2753, we should take a moment to reflect on it, and consider whether we want to remove or add certain hooks to it. This issue will be a place for hook suggestions and discussions.

Previously, discussion about `pre-commit` has occurred in #2744.
WIthin this issue, the following hooks were mentioned:
- [PyUpgrade](https://github.com/asottile/pyupgrade/): A hook to automatically upgrade syntax for newer verions of Python. I like the sound of this one.
- [isort](https://github.com/PyCQA/isort): Import sorting. I like the sound of this one.
- [black](https://github.com/psf/black): Python formatter, implemented if #2774 gets merged.

Beyond these, the `pre-commit` team has provided some more hooks, which can be found [here](https://github.com/pre-commit/pre-commit-hooks). Some of them I find particularly interesting, such as:
- `debug_statements`: Check for debugger imports and py37+ `breakpoint()` calls in python source.
- `no-commit-to-branch`: Protect specific branches from direct checkins.
- Hooks that attempt to load specific files to verify syntax. (`check-xml`, `check-yaml`, `check-toml`, `check-json`)

Some other hooks provided by the `pre-commit` team have niche uses, but might be nice, such as:
- `check-case-conflict`: Check for files with names that would conflict on a case-insensitive filesystem like MacOS HFS+ or Windows FAT.
- `check-merge-conflict`: Check for files that contain merge conflict strings.

Other custom hooks I've used myself are:
- [Bandit](https://github.com/PyCQA/bandit): For security vulnerabilities. Might not be very applicable for nltk.
- [mypy](https://github.com/pre-commit/mirrors-mypy): Add type hints ([PEP 484](https://www.python.org/dev/peps/pep-0484/)) (e.g. `def func(n: int) -> bool`) and statically check correctness of these type hints. I like this one, but it might require solving a bunch of warnings before it can be incorporated into `pre-commit`.
- [pylint](https://github.com/PyCQA/pylint): Very interesting if you're very interested in improving code quality, but this hook will only pass if there are no warnings, meaning that a developer can only push their commit if all warnings are resolved. Currently, there are roughly 15000 warnings.
- [codespell](https://github.com/codespell-project/codespell): Points out potential spelling errors in code comments, with the option to automatically resolve them. Interesting, but often marks correct words as incorrect.

[...]

PR #2782, opened 2021-08-12
Run CI tests on Windows, too

tomaarsen:
Hello!

### Pull request overview
- Made CI tests run on Windows too. This involves:
  - Ensuring that `nltk.download()` data is placed in `~/nltk_data`.
- Added `${{ secrets.CACHE_VERSION }}` to cache key. **This requires setting `CACHE_VERSION` in Settings -> Secrets to e.g. `v1`.** This is the only way to "reset" the cache in case of a broken cache. The `CACHE_VERSION` in Secrets can be edited to e.g. `v2`, and a new cache will be created and used instead.

---

### Adding Windows support
We cannot simply add `windows-latest` to the `test` job. If we did, then the step to use the cached `nltk.download()` data will fail or return no data, and then a bunch of the `pytest` tests will fail due to lacking the required `nltk.download()` files.

So, we need to add a cache for `windows-latest` as well. 

Simply adding `windows-latest` to the `cache_nltk_data` job doesn't do it either - as `nltk.download('all')` will place the files in `~/nltk_data` for Ubuntu and MacOS, but in `C:\Users\...\AppData\Roaming\nltk_data` for Windows. So, either:
* The download location is fixed.
* The location where to read the cache is OS specific.

I went with the first option, and replaced
```
python -c "import nltk; nltk.download('all')"
```
with
```
python -c "import nltk; from pathlib import Path; path = Path('~/nltk_data').expanduser(); path.mkdir(exist_ok=True); nltk.download('all', download_dir=path)"
```
This ensures the folder exists (which is required to prevent an Exception on windows), and then places the downloaded files on `~/nltk_data`, regardless of OS.

---

### CACHE_VERSION Secret
During testing, the cache was frequently wrong. However, the only way to "clear" a cache in GitHub Actions is to use a different cache key instead, forcing it to make a new cache instead of using the old broken one. However, pushing a new `ci.yaml` with a modified key every time we want to reset the cache is not convenient at all. (There are other ways, but they involve not pushing code for 7 days, or filling up the cache with an additional 5+ GB, after which GitHub Actions clears the cache for exceeding their limit)

That's why it's recommended to use a Secret which can be modified on the fly, after which the CI can just be re-run with a button on the Actions tab on GitHub. 

For this, someone with Secrets privileges has to go to Settings -> Secrets, or follow this link: https://github.com/nltk/nltk/settings/secrets/actions. Once there, click on New Repository Secrets, and set `CACHE_VERSION` as the name, with `v1` as the value. Whenever there is an issue with the cache, we can clear it by updating the Secret to e.g. `v2` (or just a randomly generated string, it doesn't matter).

However, the cache is only likely to have issues if we mess with the CI, which should rarely happen.

---

### Further improvements
Something we should still tackle is adding `pre-commit` to the CI. I believe this can be done by simply adding
```yaml
- uses: pre-commit/actions@v2.0.0
```
But I've yet to really look into it. See https://github.com/nltk/nltk/issues/2775#issuecomment-894076955 for more information.

---

Also, my actual changes were iteratively tried in my `add_windows_ci` branch, and not this `add_windows_ci_reduced` branch. However, I didn't want to create this PR with 19 unnecessary commits, so I just created a new branch from develop instead. For those interested, see https://github.com/tomaarsen/nltk/tree/add_windows_ci for the individual attempts I made.

@purificant This PR might interest you, given that you worked on the CI file.

- Tom Aarsen

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,10 @@ on: [push, pull_request, workflow_dispatch]
 jobs:
   cache_nltk_data:
     name: cache nltk_data
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
@@ -15,12 +18,12 @@ jobs:
         id: restore-cache
         with:
           path: ~/nltk_data
-          key: nltk_data
+          key: nltk_data_${{ secrets.CACHE_VERSION }}
 
       - name: Download nltk data packages on cache miss
         run: |
           pip install regex # dependencies needed to download nltk data
-          python -c "import nltk; nltk.download('all')"
+          python -c "import nltk; from pathlib import Path; path = Path('~/nltk_data').expanduser(); path.mkdir(exist_ok=True); nltk.download('all', download_dir=path)"
         shell: bash
         if: steps.restore-cache.outputs.cache-hit != 'true'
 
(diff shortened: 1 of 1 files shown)


stevenbird:
@tomaarsen nice work, and great to have this

PR #2784, opened 2021-08-14
Run pre-commit on each diff

dannysepler:
Following the very helpful guidance here https://github.com/nltk/nltk/issues/2775#issuecomment-894076955

This should run pre-commit on each NLTK diff. I decided to put pre-commit in its own "job" rather than adding it in another job. I chose this so that:

1. It's easier to understand what failed when it's a job called "pre-commit", rather than a test job
2. We only run pre-commit once, rather than once per environment


stevenbird:
@dannysepler I see some extraneous changes in this PR

dannysepler:
Yep sorry, should have explained that! Those changes are needed to make pre-commit pass. I'm guessing they've just been missed as we've enabled various hooks?

CONTRIBUTING.md
```diff
@@ -112,9 +112,9 @@ Summary of our git branching model:
   code](http://www.codinghorror.com/blog/2008/07/coding-without-comments.html);
 - Name identifiers (variables, classes, functions, module names) with readable
   names (`x` is always wrong);
-- When manipulating strings, use [Python's new-style
-  formatting](http://docs.python.org/library/string.html#format-string-syntax)
-  (`'{} = {}'.format(a, b)` instead of `'%s = %s' % (a, b)`);
+- When manipulating strings, use [Python's f-string
+  formatting](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals)
+  (f`"{a} = {b}"` instead of `'{} = {}'.format(a, b)` or `'%s = %s' % (a, b)`);
```
tomaarsen: I believe these are a part of the extraneous changes that Steven saw in this PR.
Furthermore, I would add that the format-string syntax is still supported, but the f-string formatting is generally preferred.

CONTRIBUTING.md
```diff
@@ -178,6 +178,12 @@ The [`.github/workflows/ci.yaml`](https://github.com/nltk/nltk/blob/develop/.git
       - Load cached `nltk_data` loaded via `cache_nltk_data`.
       - Run `pytest --numprocesses auto -rsx nltk/test`.
 
+ - The `pre-commit` job
+   - performs these steps:
+     - Downloads the `nltk` source code.
+     - Runs pre-commit on all files in the repository. (Similar to `pre-commit run --all-files`)
+     - Fails if any hooks performed a change
```
tomaarsen:
```suggestion
     - Fails if any hooks performed a change.
```
Added a dot to the sentence.

tomaarsen (commented):
@dannysepler 

Looks solid! Something to sonsider, however:
This CI job is completely separate from the other two, meaning that it runs concurrently with the two other jobs, which have s a task to ensure that the tests run properly. This has several consequences:
* If `pre-commit` passes, then the test jobs have already started, meaning that no time was wasted waiting on `pre-commit` to give the all-good. Wonderful. **But,**
* If `pre-commit` fails, then the test jobs have already started, despite them not being necessary, as the CI will fail regardless. In this case, will the test jobs stop once `pre-commit` has failed? Or will the test jobs continue to completion? I'm not 100% but I suspect they'll just keep going, given that `fail-fast` was set to `false` for `test`:

https://github.com/nltk/nltk/blob/eeece6ca5bb2d8c26b6821db6f46d63e90823818/.github/workflows/ci.yaml#L30-L37

I suspect the latter consequence means that 10+ min CI runs will be started even when the CI knew after 20 seconds that it would fail due to a trailing space. In those cases, people will be shown that the CI is still executing for these 10 minutes, when they could already be shown their fault after half a minute. If the developer knew their mistake in half a minute, they could fix the issues way more quickly. Furthermore, it'll be even clearer if there is only one failing test: `pre-commit`, and the others never even run - not until `pre-commit` passes at least. 

So, I propose one change, and I welcome discussion on this: 
Add `pre-commit` as a sequential job, instead of a concurrent one - which runs *before* the test jobs. 

This can easily be implemented by adding `needs: pre-commit` to the `cache_nltk_data` job, like so:
```yaml
...

jobs:
  cache_nltk_data:
    name: cache nltk_data
    needs: pre-commit
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest]
    runs-on: ${{ matrix.os }}
    steps:
      ...
```

This way, the CI will become a chain of `pre-commit -> cache_nltk_data -> test`, where a later job in the chain cannot run until all previous were executed. After that, I would also recommend moving the `pre-commit` job up to be the first job in the `ci.yaml` file, as then the jobs are ordered chronologically to their execution.
I'd love to hear what you all think about this.

- Tom Aarsen

tomaarsen:
> 
> 
> Yep sorry, should have explained that! Those changes are needed to make pre-commit pass. I'm guessing they've just been missed as we've enabled various hooks?

The other formatting and whitespace changes that are in this PR are all related to changes from #2763 about METEOR, which was a PR made before most of the pre-commit hook changes. That is why they were missed previously. These changes do need to be included in this PR in order for it to pass the checks.

CONTRIBUTING.md
dannysepler: makes sense! thanks

dannysepler:
re: the order, that's a good point! i didn't intentionally think about consecutively vs not. at the moment, i believe if pre-commit fails tests will continue until completed. that said, the wheel here will begin to show red the moment pre-commit fails:
 
Screen Shot 2021-08-15 at 6 56 27 PM

so a user should get some notice right away. BUT, perhaps it's overwhelming? i could see an argument for doing both so that users know upfront all the things to change. i could also see an argument for only running tests if pre-commit passes, so that there's less to think about

i don't have much stake in the matter! appreciate you teaching me how to have one task depend on the other, looks very doable for this pull request if that's what we'd like to go with

stevenbird:
Let's try it @tomaarsen's way, only running tests if pre-commit passes.

tomaarsen (approved):
Looks great to me! And regarding the extraneous changes - that should stop happening once this PR is merged. By then, no other PR can be merged until they conform to the pre-commit standards.

Furthermore, merging this will prevent develop and other PR's from failing.

I'm very happy to have this included. Thanks Danny for putting in the work!

iliakur:
I know this is post factum at this point, but I'm also in favor of having tests "depend" on a passing pre-commit job. In my experience pre-commit tends to take less time, so we would be "failing early" in this case.

tomaarsen:
@iliakur Perhaps I'm misunderstanding, but that is exactly the situation since this PR. You can see the effects [here](https://github.com/nltk/nltk/actions/runs/1145022504): Tests can only run if `pre-commit` passes.

iliakur:
yup, exactly! that's why my vote in favor was more for the record

tomaarsen:
Oh, wonderful! I misunderstood it as a suggestion. Also - hope you and the little one are well.

dannysepler:
Cool! Glad this is the preferred way. Also pretty satisfied so far with how easy this was to enable

PR #2786, opened 2021-08-17
Add optional `show` argument to `FreqDist.plot()`

### LINKED ISSUE #2663: Remove plt.show() from FreqDist.plot()
I think it would be a good idea to remove plt.show() from the FreqDist.plot() function here: https://github.com/nltk/nltk/blob/3d7960008dd9f3802a318f5ed20fa4a63584721e/nltk/probability.py#L302

This is because you cannot customize the plot further after calling plt.show(). For example, I'd prefer to change the x-axis label to 'word', save the figure with plt.savefig(), use plt.tight_layout(), and so on. The only way to currently do this is to run the code in IPython after using the %matplotlib magic command.

I'd be happy to complete a pull request but it's such a tiny change, maybe it's best to do it another way?

tomaarsen:
Closes #2663 

Hello!

### Pull request overview
- Added `show` argument to FreqDist's `plot` method. 
- Made optional function arguments explicit, allowing IDE's to autocomplete these.
  - E.g. put `cumulative=False` in the function header, instead of only showing in the function comments that this is a parameter.

---

### Adding `show` to FreqDist's `plot`
As mentioned by #2663, the following line will automatically plot the frequency distribution:
https://github.com/nltk/nltk/blob/3d7960008dd9f3802a318f5ed20fa4a63584721e/nltk/probability.py#L302

This is great - it's exactly what the FreqDist's `plot` function implies, but it prevents the plot from being modified in any way. What if the user wishes to just `.savefig` these instead of showing them? Or if they want to use `.tight_layout` before plotting? They would need to let the function plot, close the plot, and work with the `ax` that was returned.

However, instead of simply removing this line and changing the functionality of this method (like suggested in #2663), I've opted to add a simple parameter `show`, which is True by default. If `show` is `True`, then `plt.show()` will be executed, and the distribution will be plotted exactly like before. If it is `False`, then it's not plotted. Like always, the `ax` will be returned, allowing it to be modified before the user themselves calls `plt.show()`.

---

### Made optional function arguments explicit
Currently, the method signature is this:
https://github.com/nltk/nltk/blob/3d7960008dd9f3802a318f5ed20fa4a63584721e/nltk/probability.py#L249
And certain arguments are extracted like this:
https://github.com/nltk/nltk/blob/3d7960008dd9f3802a318f5ed20fa4a63584721e/nltk/probability.py#L274-L275
These arguments are briefly mentioned in the method's docstring. These two lines extracts the values (if they exists, otherwise it takes the third parameter, the default) and removes them from the `kwargs` dict. 

Afterwards, `**kwargs` is used in `matplotlib`'s plotting function. 

I've now changed that to list these arguments in the method signature, so that IDE's can autocomplete and autosuggest these arguments.  Beyond that, this change does not impact the functionality of this method at all.

---

### Note
This PR should not change the execution of existing programs. All it does is add the functionality of optionally not plotting, and improve how arguments are taken. 

@stevenbird Whether you want to include this in NLTK 3.6.3 is up to you.

- Tom Aarsen

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -1955,6 +1955,8 @@ class ConditionalFreqDist(defaultdict):
 
         if not conditions:
             conditions = self.conditions()
+        else:
+            conditions = [c for c in conditions if c in self]
         if not samples:
             samples = sorted({v for c in conditions for v in self[c]})
         if "linewidth" not in kwargs:
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -1916,7 +1916,17 @@ class ConditionalFreqDist(defaultdict):
         """
         return sum(fdist.N() for fdist in self.values())
 
-    def plot(self, *args, **kwargs):
+    def plot(
+        self,
+        *args,
+        samples=None,
+        title="",
+        cumulative=False,
+        percents=False,
+        conditions=None,
+        show=True,
+        **kwargs,
+    ):
         """
         Plot the given samples from the conditional frequency distribution.
         For a cumulative plot, specify cumulative=True.
@@ -1926,8 +1936,14 @@ class ConditionalFreqDist(defaultdict):
         :type samples: list
         :param title: The title for the graph
         :type title: str
+        :param cumulative: Whether the plot is cumulative. (default = False)
+        :type cumulative: bool
+        :param percents: Whether the plot uses percents instead of counts. Only when cumulative is True.
+        :type percents: bool
         :param conditions: The conditions to plot (default is all)
         :type conditions: list
+        :param show: Whether to show the plot, or only return the ax.
+        :type show: bool
         """
         try:
             import matplotlib.pyplot as plt  # import statement fix
@@ -1937,19 +1953,14 @@ class ConditionalFreqDist(defaultdict):
                 "See http://matplotlib.org/"
             ) from e
 
-        cumulative = _get_kwarg(kwargs, "cumulative", False)
-        percents = _get_kwarg(kwargs, "percents", False)
-        conditions = [
-            c for c in _get_kwarg(kwargs, "conditions", self.conditions()) if c in self
-        ]  # conditions should be in self
-        title = _get_kwarg(kwargs, "title", "")
-        samples = _get_kwarg(
-            kwargs, "samples", sorted({v for c in conditions for v in self[c]})
-        )  # this computation could be wasted
+        if not conditions:
+            conditions = self.conditions()
+        if not samples:
+            samples = sorted({v for c in conditions for v in self[c]})
         if "linewidth" not in kwargs:
             kwargs["linewidth"] = 2
         ax = plt.gca()
-        if len(conditions) != 0:
+        if conditions:
             freqs = []
             for condition in conditions:
                 if cumulative:
@@ -1979,7 +1990,9 @@ class ConditionalFreqDist(defaultdict):
                 ax.set_title(title)
             ax.set_xlabel("Samples")
             ax.set_ylabel(ylabel)
-        plt.show()
+
+        if show:
+            plt.show()
 
         return ax
 
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -244,7 +244,9 @@ class FreqDist(Counter):
             )
         return self.most_common(1)[0][0]
 
-    def plot(self, *args, **kwargs):
+    def plot(
+        self, *args, title="", cumulative=False, percents=False, show=True, **kwargs
+    ):
         """
         Plot samples from the frequency distribution
         displaying the most frequent sample first.  If an integer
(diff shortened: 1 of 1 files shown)


tomaarsen:
Converted to a draft, as I found out that `ConditionalFreqDist` has a similar `plot` method. I'll make these same changes there when I have time.

tomaarsen:
I'm open to feedback here (as always), including argument order and docstring argument documentation.

dannysepler:
Looks pretty solid to me!

nltk/probability.py
```diff
@@ -244,18 +244,24 @@ def max(self):
...
         (Requires Matplotlib to be installed.)
 
-        :param title: The title for the graph
+        :param title: The title for the graph.
         :type title: str
-        :param cumulative: A flag to specify whether the plot is cumulative (default = False)
-        :type title: bool
+        :param cumulative: Whether the plot is cumulative. (default = False)
+        :type cumulative: bool
+        :param percents: Whether the plot uses percents instead of counts. Only when cumulative is True.
```
stevenbird: @tomaarsen percents is just the unit, and can be used independently of whether the plot is cumulative, right?

iliakur:
It would be nice to get rid of the duplication somehow :thinking: 

Also, this PR would be a good opportunity to include a word or two about `kwargs` in the docstring. Right now the only way to find out that they are passed to matplotlib is to read the code :/

nltk/probability.py
stevenbird: By which I mean something like the following:

```
        if cumulative:
            freqs = list(self._cumulative_frequencies(samples))
            ylabel = "Cumulative "
        else:
            freqs = [self[sample] for sample in samples]
            ylabel = ""


        if percents:
            freqs = [100 * f / self._N for f in freqs]
            ylabel += "Percents"
        else:
            ylabel += "Counts"
```

tomaarsen: I'll build it in when I have some time!

tomaarsen:
> It would be nice to get rid of the duplication somehow 🤔

It would be. I looked into it a bit, but the functions do seem to differ too much for some attempt at reducing duplication to really help. Perhaps I'll look into it later again.

> Also, this PR would be a good opportunity to include a word or two about `kwargs` in the docstring. Right now the only way to find out that they are passed to matplotlib is to read the code :/

I've added it, thank you!

nltk/probability.py
tomaarsen: I've added it. Thank you for your implementation, as well!

tomaarsen:
Some scripts to test the recent changes to `percents` etc.:

### FreqDist

```python
from nltk.probability import FreqDist

import matplotlib.pyplot as plt

fd = FreqDist({'the': 3, 'dog': 2, 'not': 1})
# Test for `show`
ax = fd.plot(cumulative=False, percents=False, show=False)

plt.tight_layout()
plt.show()

# The other 3 configurations of cumulative and percents
ax = fd.plot(cumulative=False, percents=True)
ax = fd.plot(cumulative=True, percents=True)
ax = fd.plot(cumulative=True, percents=False)
```

### ConditionalFreqDist

```python
from nltk.probability import ConditionalFreqDist
from nltk.tokenize import word_tokenize
sent = "the the the dog dog some other words that we do not care about"
cfdist = ConditionalFreqDist()
for word in word_tokenize(sent):
    condition = len(word)
    cfdist[condition][word] += 1
# All 4 configurations of cumulative and percents
cfdist.plot()
cfdist.plot(cumulative=True)
cfdist.plot(cumulative=True, percents=True)
cfdist.plot(cumulative=False, percents=True)
```

iliakur:
Sorry @tomaarsen you're gonna hate me, but here's a question after almost a full night of sleep: what if we always return the plot and have the user invoke the `plt.show()` method?

I generally find the interface "perform I/O (i.e. showing plot) **by default**" unintuitive. Is this for backwards compatibility?

Note that inside a Jupyter notebook (a common work environment for many users) the plot is rendered even if we only return it. For environments that require explicitly rendering the plot, I think it's easier for a user to invoke a well-known matplotlib method (`plt.show()`) than reading through our whole docstring to understand the meaning of a custom argument. In general, whenever interacting with such well-established interfaces as that of `matplotlib`, I'd lean towards using the well-established interfaces instead of rolling our own stuff.

I know this is coming late and I really am sorry about that.

tomaarsen:
This is indeed done for backwards compatibility. The goal was to only improve and add, and keep existing programs the same. See Note for a short message in this direction:

> This PR should not change the execution of existing programs. All it does is add the functionality of optionally not plotting, and improve how arguments are taken.

I envisioned these functions to be used exclusively in debugging, when some developer wants to see what's happening in their FreqDist really quickly, hence the default show being True. However, if simply returning the `ax` object is a preferred default, then I have no issues with simply setting the default for the new `show` to False.

Thoughts on that?

stevenbird:
I like this suggestion, but how about we save up breaking changes for the next major release, i.e. version 3.7. So, label such issues and PRs with nltk3.7?

tomaarsen:
@stevenbird Are you suggesting to keep this PR as it is, where it only adds features without changing existing functionality, and then open an issue to eventually implement the suggestion? If so, then we can create an issue for the suggestion once this is merged.

stevenbird:
@tomaarsen precisely

stevenbird:
Thanks @tomaarsen, + @dannysepler, @iliakur

iliakur:
this sounds good. Since we're maintaining backwards compat here, the default value of the `show` argument should be whatever keeps existing behavior.

I agree with the call to keep this MR as is and improve the interface in a subsequent release (I see you've already created an issue for it :) )

tomaarsen:
Yup! Whoever feels like it at the time can work on that issue in preparation for another release.

PR #2787, opened 2021-08-18
Made `nltk/__init__.py` equivalent to `setup.py`

### LINKED ISSUE #2509: Remove outdated names in init.py
Cf #2508

tomaarsen:
Fixes #2509

Hello!

### Pull request overview
- Made `nltk/__init__.py` equivalent to `setup.py` in terms of project description, classifiers, etc.

---

### Changes
I simply updated all dunder variables in `nltk/__init__.py` to be equivalent to their corresponding variables in `setup.py`. This includes changes to:
* Long description.
* Maintainer.
* Trove classifiers.

---

### Note
I simply copied the values from `setup.py`, but as a consequence two names are removed from the maintainer list here!

- Tom Aarsen

### DIFF
--- a/nltk/__init__.py
+++ b/nltk/__init__.py
@@ -52,7 +52,7 @@ __license__ = "Apache License, Version 2.0"
 # Description of the toolkit, keywords, and the project's primary URL.
 __longdescr__ = """\
 The Natural Language Toolkit (NLTK) is a Python package for
-natural language processing.  NLTK requires Python 2.6 or higher."""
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @tomaarsen, good catch!

iliakur:
Would it be possible to reuse this string in `setup.py` directly? I dunno if that'll work, but seems like a good way to reduce duplication. Just to minimize the number of things we have to keep consistent...

Issue #2112, opened 2018-09-10
CoreNLPParser tag() should allow properties overloading


alvations:
With the current `CoreNLPParser.tag()`, the "retokenization" by Stanford CoreNLP is unexpected:

```python
>>> from nltk.parse.corenlp import CoreNLPParser
>>> ner_tagger = CoreNLPParser(url='http://localhost:9000', tagtype='ner')
>>> sent = ['my', 'phone', 'number', 'is', '1111', '1111', '1111']
>>> ner_tagger.tag(sent)
[('my', 'O'),
 ('phone', 'O'),
 ('number', 'O'),
 ('is', 'O'),
 ('1111\xa01111\xa01111', 'NUMBER')]
```

The expected behavior should be:

```python
>>> from nltk.parse.corenlp import CoreNLPParser
>>> ner_tagger = CoreNLPParser(url='http://localhost:9000', tagtype='ner')
>>> sent = ['my', 'phone', 'number', 'is', '1111', '1111', '1111']
>>> ner_tagger.tag(sent)
[('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111', 'DATE'), ('1111', 'DATE'), ('1111', 'DATE')]
```

Proposed solution is to allow `properties` arguments overloading for `.tag()` and `.tag_sents()`, i.e. at https://github.com/nltk/nltk/blob/develop/nltk/parse/corenlp.py#L348 and by default use `properties = {'tokenize.whitespace':'true'}` because we are concatenating the tokens by spaces in `tag_sents()`.

```python

    def tag_sents(self, sentences, properties=None):
        """
        Tag multiple sentences.

        Takes multiple sentences as a list where each sentence is a list of
        tokens.

        :param sentences: Input sentences to tag
        :type sentences: list(list(str))
        :rtype: list(list(tuple(str, str))
        """
        # Converting list(list(str)) -> list(str)
        sentences = (' '.join(words) for words in sentences)
        if properties == None:
            properties = {'tokenize.whitespace':'true'}
        return [sentences[0] for sentences in self.raw_tag_sents(sentences, properties)]

    def tag(self, sentence, properties=None):
        """
        Tag a list of tokens.

        :rtype: list(tuple(str, str))

        >>> parser = CoreNLPParser(url='http://localhost:9000', tagtype='ner')
        >>> tokens = 'Rami Eid is studying at Stony Brook University in NY'.split()
        >>> parser.tag(tokens)
[...]

PR #2789, opened 2021-08-29
Allowed properties overloading for CoreNLPParser tag, Authors.md

BatMrE:
This is a fix to #2112
Added the overloading properties in code as per the standards

before fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111\xa01111\xa01111', 'O')]
after fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111', 'DATE'), ('1111', 'DATE'), ('1111', 'DATE')]


tomaarsen:
For those interested in the reasoning behind the changes in this PR, feel free to have a look at #2112 itself, where all the changes are already described and laid out.

nltk/parse/corenlp.py
```diff
@@ -339,9 +339,11 @@ def tag_sents(self, sentences):
         """
         # Converting list(list(str)) -> list(str)
         sentences = (" ".join(words) for words in sentences)
-        return [sentences[0] for sentences in self.raw_tag_sents(sentences)]
+        if properties is None:
```
dimazest, line 342: Please add a description for the `properties` argument in the docstring, also mention that `"tokenize.whitespace"` is set to `True` by default (but it won't if `properties` are specified.

dimazest:
> before fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111\xa01111\xa01111', 'O')]
after fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111', 'DATE'), ('1111', 'DATE'), ('1111', 'DATE')]

would it be possible to make a test out of this example?

BatMrE:
> > before fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111\xa01111\xa01111', 'O')]
> > after fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111', 'DATE'), ('1111', 'DATE'), ('1111', 'DATE')]
> 
> would it be possible to make a test out of this example?

Sure , bit busy, will add over weekend

BatMrE:
> > before fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111\xa01111\xa01111', 'O')]
> > after fix : [('my', 'O'), ('phone', 'O'), ('number', 'O'), ('is', 'O'), ('1111', 'DATE'), ('1111', 'DATE'), ('1111', 'DATE')]
> 
> would it be possible to make a test out of this example?

@tomaarsen @dimazest  
any idea how to create 'api_return_value' dictionary for a sample?? like in:
https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_corenlp.py#L261

dimazest:
you should be able to manually use CoreNLP Rest API and intercept returned result

tomaarsen:
Keep in mind that the CoreNLP tests are being skipped, because our GitHub Actions CI doesn't currently support downloading third party tools such as the Stanford models. Preferably we would still have tests for this PR though, as we likely want to support third party tools eventually.

dimazest:
The actual call to corenlp should be mock, then tests won't need corenlp running.

tomaarsen:
Since #2820 there are automated tests for CoreNLP. We would like to see some tests to back up these changes, is possible.

nltk/parse/corenlp.py
```diff
@@ -339,9 +339,11 @@ def tag_sents(self, sentences):
...
+            properties = {"tokenize.whitespace": "true"}
```
tomaarsen, line 343: I've merged develop into this PR. Because the CoreNLP tests now run on the CI, we can see that some tests fail due to this change.

tomaarsen (changes requested):

nltk/parse/corenlp.py
BatMrE: @tomaarsen apart from automated tests for CoreNLP is there any other test that needs to be added

tomaarsen: The PR should have some tests that show that the new behaviour works like intended, so you would only need to add tests for the functionality that you added. So, just for these functions.

stevenbird:
@BatMrE: just wondering if there's interest in wrapping this up. I think it's just a question of adding a test or two.

stevenbird:
Closing. Hopefully someone will resurrect this if it's still an issue.

PR #2791, opened 2021-08-29
Fixing TweetTokenizer format

12mohaned:
Fix PEP8 issues in TweetTokenizer class.


tomaarsen (approved):
Generally speaking, I'm personally not particularly worried about line lengths of comments or regexes that much, but I recognise that enforcing a line length more strictly is more "proper". Considering the before and after are functionally identical, I have no issues approving this.

12mohaned:
I totally agree with you, but I do find reading long line hard and little bit distributing.

stevenbird:
Thanks @12mohaned, @tomaarsen

PR #2798, opened 2021-09-06
feat: enable phone number regex to recognize phone numbers

### LINKED ISSUE #1799: TweetTokenizer quirk with long numbers
When there's a number token that's >8 length, the `TweetTokenizer` tries to iteratively parse the first 8-10 digits as a phone number if it matches the phone number regex on https://github.com/alvations/nltk/blob/develop/nltk/tokenize/casual.py#L125

Maybe, there should be an option to allow users to keep long digits and not try to recognize the first part as a number. 

Or better yet, if there's regex fix that can prevent the number from getting split if it's longer than 10. 

Details on https://stackoverflow.com/questions/45425946/tokenize-in-nltk-tweettokenizer-returning-integers-by-splitting/45431922#45431922

12mohaned:
This pull request implement feature that solve a problem raised by this [issue](https://github.com/nltk/nltk/issues/1799) by allowing phoneRegex feature to be enabled when initialising the tokenizer. 

I had to use the global keyword to be able to access the global variable **WORD_RE** in the function scope, which suggests that the structure of this class needs further refactoring and improvement, for example(labelling not-related methods to the twitter tokeniser with static decorator, and placing global variables in the right setting, etc..).


tomaarsen:
@12mohaned Heya! We've started using https://pre-commit.com/ pretty recently. With this program, some tests are ran before it lets you commit. With [our configuration](https://github.com/nltk/nltk/blob/develop/.pre-commit-config.yaml), it means that some commands are ran to ensure some code quality. In particular, the `black` formatter. As you can see in the [test output](https://github.com/nltk/nltk/pull/2798/checks?check_run_id=3526840268), the `black` formatter complained that the changes you are trying to push are not formatted according to the `black` formatting rules.

My recommendation is to run `pre-commit install`. This will install the pre-commit hooks, causing it to run all of those hooks (e.g. formatters, verifiers, etc.) when you make another commit. If you then try to push another commit, the `black` hook will reformat your changes, and you can push the correctly formatted changes. This might seem a little bit confusing, but it's more intuitive in practice.

Let me know if you have any issues. If you do not try this method, then our Continuous Integration will complain about the formatting of the file, and no changes can even be merged.

tomaarsen:
Well done! From now onwards, pre-commit should ensure that your code (for NLTK) is properly formatted (among other things) without any extra work.

I don't have time to really review your work yet, but I did notice that you didn't provide any tests with your work. Perhaps it would be possible to add a few, as you're adding some new functionality to TweetTokenizer.

tomaarsen (changes requested):
I'm a tad confused. Your new `PHONE_REGEX` seems to be a copy of the existing phone regex section in `REGEXPS`, and `REGEXPS_PHONE` is just a copy of `REGEXPS`, but with `PHONE_REGEX` instead of the existing phone regex section. I might be wrong, but isn't `REGEXPS_PHONE` just functionally identical to `REGEXPS` with your changes?

In short, I'm just seeing a lot of code duplication, and I'd prefer a solution without any.

Beyond that, with your idea of modifying the global `WORD_RE`, doesn't that prevent the original `WORD_RE` from being used again? As in,
```python
tok_one = TweetTokenizer(phone_number_regex=True)
tok_two = TweetTokenizer(phone_number_regex=False)
tok_one.tokenize("My text")
tok_two.tokenize("My text") # Does this one use the original WORD_RE, or the updated one with REGEXPS_PHONE?
```
That might be an issue.

Personally I'm not a big fan of modifying those variables (I consider them to be constants). Would it be easier to simply go for e.g.?:
```python
    if self.phone_number_regex:
        words = PHONE_WORD_RE.findall(safe_text)
    else:
        words = WORD_RE.findall(safe_text)
```
However, compiling a second big regex should preferably only be done if strictly necessary. Perhaps this can even be fixed by modifying the existing regular expression section for phone numbers.

12mohaned:
Yes, sorry for the confusion I forgot to delete the phone regex from the REGEPX variable. In a nutshell, the idea is to have two pipelines, one for recognising phone numbers and one without the phone regex expression. In the example of tok_one and tok_two, the tokeniser will use the updated word_re which is a problem as you mentioned. Also, I totally agree even after deleting the phone regex it would be more consuming to go with the option of creating two pipelines that contains the same components with the difference of phone regex. 

Thanks for the constructive feedback.

nltk/tokenize/casual.py
```diff
@@ -350,17 +334,14 @@ def tokenize(self, text):
...
-        if self.phone_number_regex:
-            global WORD_RE
-            WORD_RE = regex.compile(
-                r"""(%s)""" % "|".join(REGEXPS_PHONE),
-                regex.VERBOSE | regex.I | regex.UNICODE,
-            )
         # Shorten problematic sequences of characters
         safe_text = HANG_RE.sub(r"\1\1\1", text)
         # Tokenize:
         words = WORD_RE.findall(safe_text)
```
tomaarsen: This still doesn't seem quite right - if `self.phone_number_regex == True`, then `words = WORD_RE.findall(safe_text)` is ran, and then this `words` is overridden. So the `words = WORD_RE.findall(safe_text)` was not necessary if `self.phone_number_regex == True`.

12mohaned:
This feature will introduce errors in some test cases related to the TweetTokenizer which contain phone regular expression by default. For example, test_phone_tokenizer in test_tokenize.py

tomaarsen:
@12mohaned If it's okay with you - I added some changes like you suggested in your first comment, when you said
> I had to use the global keyword to be able to access the global variable WORD_RE in the function scope, which suggests that the structure of this class needs further refactoring and improvement, for example(labelling not-related methods to the twitter tokeniser with static decorator, and placing global variables in the right setting, etc..).

I agreed, and I went and turned `WORD_RE` and `PHONE_WORD_RE` into properties of the TweetTokenizer class. Preferably I would like to have moved `EMOTICON_RE`, `HANDLES_RE`, `ENT_RE` and `HANG_RE`, as well as `_str_to_unicode` and `_replace_html_entities` into `TweetTokenizer`, but I'm not sure which of those variables are used by developers. 
In fact, I tried to move them into the `TweetTokenizer` scope, but I got an error when `json2csv_preprocess` in `nltk/sentiment/util.py` uses `EMOTICON_RE`.

I'll quickly elaborate my changes:
* Turned `WORD_RE` and `PHONE_WORD_RE` into properties of the `TweetTokenizer` class. These can be accessed with `TweetTokenizer().WORD_RE`, and they are compiled only once. Even if you have multiple objects like
  ```python
  tok_one = TweetTokenizer(reduce_len=False)
  tok_two = TweetTokenizer(reduce_len=True)
  tok_one.tokenize("This is the text I'm tokenizing")
  tok_two.tokenize("This is the text I'm tokenizing")
  ```
  Even then, the regex is only compiled once. That way, if someone only uses `match_phone_numbers=True`, then `WORD_RE` is never compiled, and `PHONE_WORD_RE` is compiled only once. And if someone uses `match_phone_numbers=False`, then `PHONE_WORD_RE` is not compiled, and `WORD_RE` only once.
* I removed some code duplication when making `PHONE_REGEXPS`. That is now:
  ```python
  REGEXPS_PHONE = (REGEXPS[0], PHONE_REGEX, *REGEXPS[1:])
  ```
* I took your tests (thank you for making them!) and placed them in `test_tokenize.py` instead. That's also where I modified the phone number test to use `match_phone_numbers=True`.
* I renamed `phone_number_regex` to `match_phone_numbers`. I did this because with `phone_number_regex` it sounds like you have to supply the regex yourself, instead of it being a boolean. I'm definitely open to other names for this function parameter!
* Modified the comment at the top of the document, as well as added a docstring for `__init__`.

Let me know how you feel about these changes. 

Also, this PR would change the default behaviour of the `TweetTokenizer`, as `match_phone_numbers` defaults to `False`. We need to be certain that this is a change we want to make, and perhaps have a way to notify developers of this.

tomaarsen (approved):

nltk/test/unit/test_tokenize.py
```diff
@@ -54,6 +54,77 @@ def test_tweet_tokenizer(self):
         ]
         assert tokens == expected
 
+    @pytest.mark.parametrize(
+        "match_phone_numbers,test_input,expected",
```
dannysepler: nit: with parametrize, i like putting a space between args, like `"match_phone_numbers, test_input, expected"`. feels easier to read!

also i wonder if you could squeeze some of these test cases into one-liners. (okay if not haha)

nltk/tokenize/casual.py
```diff
@@ -299,15 +341,40 @@ def tokenize(self, text):
...
+        # Compiles the regex for this and all future instantiations of TweetTokenizer.
+        if not type(self)._WORD_RE:
+            type(self)._WORD_RE = regex.compile(
+                f"({'|'.join(REGEXPS)})",
+                regex.VERBOSE | regex.I | regex.UNICODE,
+            )
+        return type(self)._WORD_RE
+
+    @property
+    def PHONE_WORD_RE(self) -> regex.Pattern:
```
dannysepler, line 368: naming nit: maybe `PHONE_NUM_RE` since it's a phone number?

dannysepler:
looks like a useful feature, thanks @12mohaned! also i always learn from your explanations, @tomaarsen!

nltk/test/unit/test_tokenize.py
tomaarsen: I've placed a space between the args, and I've shrunk down the test cases a bit, to avoid needing to duplicate the `test_input`. Perhaps that's what you were after with the one-liner stuff? I'm not quite sure.

nltk/tokenize/casual.py
tomaarsen: I'm not sold on this - `PHONE_NUM_RE` sounds like it's *just* a regex for phone numbers, while it's actually `WORD_RE` *plus* phone numbers. That said, `PHONE_WORD_RE` isn't a great name.

12mohaned:
@tomaarsen Can you give me a couple of days before merging to check the changes.

tomaarsen:
@12mohaned Of course!

Before I would consider merging this, I would like to have some confirmation from the other maintainers on whether it would be a good idea to change the default behaviour of TweetTokenizer (like this PR is doing). 
In short, if we merge this PR, then we impose the following functionality changes:
* TweetTokenizer will no longer match phone numbers by default, and will perform better when matching long numbers like in `The identification number is 471274937125`.
* TweetTokenizer will support a `match_phone_numbers` boolean parameter to still allow matching phone numbers. If this is set to `True`, then the behaviour is identical to before. By default, this parameter is set to `False` (hence the behaviour on the item above).

Beyond that, there are some potential performance improvements in the compilation of regexes (which may or may not actually have an impact, depending on the internal `regex` cache, which caches at most 15 compilations). These performance improvements are accompanied with a bit of refactoring/moving things around.

nltk/test/unit/test_tokenize.py
dannysepler: Yup, exactly what I meant! Also nice, our first instance of typing in nltk 👍

stevenbird:
@12mohaned perhaps you already said somewhere, but I missed it sorry. What is the case for changing the default behaviour?

12mohaned:
@stevenbird This pull requests addresses this [issue](https://github.com/nltk/nltk/issues/1799) where the REGEX components recognise first part of long numbers (>8) as a phone number.

tomaarsen:
I see two options for us:
* Set `match_phone_numbers=True` as the default. **This is equivalent to the current behaviour**. Then:
  * Phone numbers such as `06-46124080`, `601-984-4813` or even `(393)  928 -3010` are matched **more effectively**. These are all tokenized as one token. For example:
    ```python
    >>> TweetTokenizer().tokenize("My number is 06-46124080, except it's not.")
    ['My', 'number', 'is', '06-46124080', ',', 'except', "it's", 'not', '.'] # Correct
    >>> TweetTokenizer().tokenize("My number is 601-984-4813, except it's not.")
    ['My', 'number', 'is', '601-984-4813', ',', 'except', "it's", 'not', '.'] # Correct
    >>> TweetTokenizer().tokenize("My number is (393)  928 -3010, except it's not.")
    ['My', 'number', 'is', '(393)  928 -3010', ',', 'except', "it's", 'not', '.'] # Correct
    ```
  * Long numbers and certain lengths of numbers are matched **less effectively**. For example:
    ```python
    >>> TweetTokenizer().tokenize("The product identification number is 48103284512.")
    ['The', 'product', 'identification', 'number', 'is', '4810328451', '2', '.'] # The number is split
    >>> TweetTokenizer().tokenize("My favourite substraction is 240 - 1353.")
    ['My', 'favourite', 'substraction', 'is', '240 - 1353', '.'] # The `'240 - 1353'` is not split
    ```
* Set `match_phone_numbers=False` as the default. Then: 
  * Phone numbers are matched **less effectively**. For example:
    ```python
    >>> TweetTokenizer().tokenize("My number is 06-46124080, except it's not.")
    ['My', 'number', 'is', '06-46124080', ',', 'except', "it's", 'not', '.'] # Correct
    >>> TweetTokenizer().tokenize("My number is 601-984-4813, except it's not.")
    ['My', 'number', 'is', '601-984-', '4813', ',', 'except', "it's", 'not', '.'] # Phone number is split weirdly
    >>> TweetTokenizer().tokenize("My number is (393)  928 -3010, except it's not.")
    ['My', 'number', 'is', '(', '393', ')', '928', '-', '3010', ',', 'except', "it's", 'not', '.'] # Completely tokenized
    ```
  * Long numbers and certain lengths of numbers are matched **more effectively**. For example:
    ```python
    >>> TweetTokenizer().tokenize("The product identification number is 48103284512.")
    ['The', 'product', 'identification', 'number', 'is', '48103284512', '.'] # Correct
    >>> TweetTokenizer().tokenize("My favourite substraction is 240 - 1353.")
    ['My', 'favourite', 'substraction', 'is', '240', '-', '1353', '.'] # Correct
    ```

---

Currently this PR implements the latter, while the `develop` branch implements the former version.
I'd like to get some feedback on which of the two versions is preferred.

stevenbird:
@tomaarsen thanks for the analysis.

Your work might be nice to preserve in our tests, in case of future revisions https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_tokenize.py 

If no-one has argued for a change, then I don't think there's anything to discuss, and it's a question of preserving, or staying close to, the existing behaviour.

tomaarsen:
I've set `match_phone_numbers=True` as the default, which means the default behaviour is equivalent to the previous behaviour. I've also added some tests from my comment earlier. 
@12mohaned Do you agree with these changes?

As far as I'm concerned, this PR is ready for merging.

12mohaned:
I believe it is okay, Thanks for the changes and the elegant analysis @tomaarsen .

12mohaned:
I believe this changes should be reflected in the documentation to keep NLTK developers updated with these changes.

stevenbird:
Thanks @12mohaned and @tomaarsen - great effort

PR #2799, opened 2021-09-07
Improved the removal of twitter handles

### LINKED ISSUE #2792: TweetTokenizer fails to remove user handles in text.
The below input should remove all user handles which start with "@".

input: @remy:This is waaaaayyyy too much for you!!!!!!@adam 
output : [':', 'This', 'is', 'waaayyy', 'too', 'much', 'for', 'you', '!', '!', '!', '@adam']

The TweetTokenizer fail to remove the user handle of Adam.

I would like to open a pull request that solves the following issues:- 

- Improve the regular expression pattern to detect in-text handles properly.
- Adjust the length of characters to be 15 instead of 20.

404-P4rziv4L:
Have changed the approach for removal of twitter handles from pure Regex to directly striping the text with an appropriate regex within the regex.findall(). Passes the test-case mentioned in this [issue](https://github.com/nltk/nltk/issues/2792).


12mohaned:
Great work, but please bear in mind the following points. 

- The regular expression does not specify the length of characters can be followed by the @ sign, and the twitter handle can not exceed 15 character in length, so the regular expression will match any keyword after @ sign regardless of the length, so * should be removed. 
-  The test cases I mentioned are not exhaustive by any mean, maybe these [test cases ](https://gist.github.com/hdlim15/2419dd8f6bb3987f4a3536a4008b0544)might provide more good example of how the regular expression should handle the twitter handles.

Thanks.

tomaarsen:
I totally agree with @12mohaned. It's always a good idea to add a few tests that used to fail, that now pass, to help motivate your changes. 

In addition, have a look at `regex.finditer`. The documentation for `re` (note, not for `regex`), can be found [here](https://docs.python.org/3/library/re.html#re.finditer). It's similar to your `findall`, but yields match objects instead. With these objects, you can use `match.start` and `match.end` (or equivalently `match.span`) as string indices to remove the handle. 

Using this just saves some time, as then we won't have to *both* do `findall` and multiple `sub`'s.

tomaarsen:
Another thing to consider is e.g. email addresses. The `@example` in `example@example.com` might get picked out and removed. Furthermore, simply `@` without any other text would get removed too, despite not being a handle.

404-P4rziv4L:
Is this effective in changing the length to 15?

404-P4rziv4L:
I think the test-cases are failing because some of them have handles which are greater than 15 character length.

404-P4rziv4L:
I apologise for accidentally closing the pull request.

tomaarsen:
Don't worry about it 👍 
I'll have a looksie at the code and the tests when I have some time.

tomaarsen:
I've modified the tests that were built under the assumption that a Twitter handle can only contain 20 characters. The assumption is now that they can contain 15 characters. Your changes work for these tests.
Beyond that, I reworked the regex slightly. To explain, I'll use some abbreviations (not actual code, just pseudocode):
```python
with A = r"(?<![A-Za-z0-9_!@#\$%&*])@",
B = r"(([A-Za-z0-9_]){20}(?!@))"
and C = r"(([A-Za-z0-9_]){1,19})(?![A-Za-z0-9_]*@)"

The old HANDLE_RE was in the form of:
(AB)|(AC)
And the new HANDLES_RE is:
A(B|C)
```
I only moved these parts around, the regex's functionality didn't change from @Abhijnan-Bajpai 's version.

Does that look good to you?

404-P4rziv4L:
Yea it looks fine @tomaarsen

404-P4rziv4L:
Should I close this PR @tomaarsen ?

tomaarsen:
@Abhijnan-Bajpai No, you should leave it open. I'll give other team members or contributors some days to have a look at these changes, in case they have some more suggestions. Once it seems that nobody has comments, then I'll merge the changes with the develop branch, and your updates will be added.

tomaarsen:
Thank you for your contribution @Abhijnan-Bajpai!

404-P4rziv4L:
> Thank you for your contribution @Abhijnan-Bajpai!

My pleasure!

PR #2801, opened 2021-09-11
Final py3.9 deprecation

### LINKED ISSUE #2689: More Python3.9 deprecations

dannysepler:
Closes #2689. All the other deprecations have been taken care of here #2698

This also added a few tests (and snuck in a pytest refactor)

### DIFF
--- a/nltk/util.py
+++ b/nltk/util.py
@@ -37,32 +37,38 @@ from nltk.internals import raise_unorderable_types, slice_bounds
 ######################################################################
 
 
-def usage(obj, selfname="self"):
+def usage(obj):
     str(obj)  # In case it's lazy, this will load it.
 
     if not isinstance(obj, type):
         obj = obj.__class__
 
-    print("%s supports the following operations:" % obj.__name__)
+    print(f"{obj.__name__} supports the following operations:")
     for (name, method) in sorted(pydoc.allmethods(obj).items()):
         if name.startswith("_"):
             continue
         if getattr(method, "__deprecated__", False):
             continue
 
-        getargspec = inspect.getfullargspec
-        args, varargs, varkw, defaults = getargspec(method)[:4]
-        if (
-            args
-            and args[0] == "self"
-            and (defaults is None or len(args) > len(defaults))
-        ):
-            args = args[1:]
-            name = f"{selfname}.{name}"
-        argspec = inspect.formatargspec(args, varargs, varkw, defaults)
+        try:
+            sig = str(inspect.signature(method))
+        except ValueError as e:
+            # builtins sometimes don't support introspection
+            if "builtin" in str(e):
+                continue
+            else:
+                raise
+
+        args = sig.lstrip("(").rstrip(")").split(", ")
+        meth = inspect.getattr_static(obj, name)
+        if isinstance(meth, (classmethod, staticmethod)):
+            name = f"cls.{name}"
+        elif args and args[0] == "self":
+            name = f"self.{name}"
+            args.pop(0)
         print(
             textwrap.fill(
-                f"{name}{argspec}",
+                f"{name}({', '.join(args)})",
                 initial_indent="  - ",
                 subsequent_indent=" " * (len(name) + 5),
             )
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -50,19 +50,9 @@ def test_usage_with_cls(capsys):
 
 
 def test_usage_on_builtin(capsys):
-    class MyClass(dict):
-        ...
-
-    usage(MyClass)
-
-    captured = capsys.readouterr()
-    assert captured.out == (
-        "MyClass supports the following operations:\n"
-        "  - fromkeys(iterable, value=None, /)\n"
-        "  - self.get(key, default=None, /)\n"
-        "  - self.popitem(/)\n"
-        "  - self.setdefault(key, default=None, /)\n"
-    )
+    # just check the func passes, since
+    # builtins change each python version
+    usage(dict)
 
 
 @pytest.fixture
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -49,7 +49,7 @@ def test_usage_with_cls(capsys):
     )
 
 
-def test_usage_on_builtin(capsys):
+def test_usage_on_builtin():
     # just check the func passes, since
     # builtins change each python version
     usage(dict)
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -1,81 +1,144 @@
-"""
-Unit tests for nltk.util.
-"""
+import pytest
 
-import unittest
-
-from nltk.util import everygrams
+from nltk.util import everygrams, usage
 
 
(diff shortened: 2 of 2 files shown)


nltk/util.py
```diff
@@ -37,32 +37,30 @@
 ######################################################################
 
 
-def usage(obj, selfname="self"):
+def usage(obj):
```
dannysepler, line 40: could alternatively delete this method? it's only used in `nltk/test/classify.doctest` and that usage feels unnecessary?

tomaarsen (approved):

nltk/util.py
tomaarsen: I was unaware of this function, but I could definitely see how it has a nice educatory use. Being able to quickly see the useful methods of a class can be quite useful if you're not entirely aware of the codebase, especially considering that NLTK is definitely not as well documented as I'd like.

There are some classes in NLTK for which users definitely might like to get an overview of the methods that it supports. `nltk.tree.Tree`, `nltk.BigramCollocationFinder` and `nltk.FreqDist` come to mind, but upon trying the last one, I get the following exception:

```python
>>> import nltk
>>> nltk.usage(nltk.FreqDist)
FreqDist supports the following operations:
  - self.B()
  - self.N()
  - self.Nr(r, bins=None)
  File "[sic]\nltk\util.py", line 53, in usage
    signature = inspect.signature(method)
  File "[sic]\lib\inspect.py", line 3130, in signature    
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "[sic]\lib\inspect.py", line 2879, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "[sic]\lib\inspect.py", line 2334, in _signature_from_callable
    return _signature_from_builtin(sigcls, obj,
  File "[sic]\lib\inspect.py", line 2145, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <method 'clear' of 'dict' objects>
```

Potentially this is because `FreqDist` subclasses collections' `Counter`, which again subclasses `dict`. Note that I also got a similar exception before this PR.

As far as I'm concerned, this can still be merged, as the bug existed before this already regardless. We can just open an issue for it then. However, if you feel like it, you can also include a fix for this in this PR already.

dannysepler: good catch!! you're totally right, thanks!

turns out that `inspect` and _some_ builtin functions don't always work out together. i can't really figure out how to get the failing ones working, my best workaround is to skip them?

iliakur: Hmm, what about Python's built-in `help` command? It gives you both methods and attributes as well as their docstrings. I use it extensively with all kinds of codebases. 

I completely agree that the docs need some love, but that's a separate issue imo.

tomaarsen: It's indeed a separate issue I'd say.

I'd even consider `usage` failing on some builtin functions to be a separate issue. If you'd like to get this merged, then we can just open an issue saying that `usage` is broken in some cases. That is, assuming that the bug still exists.

Regarding the tests, it might be a good idea to check if the output `.startswith("MyClass supports the following operations:")` or something, or whether the output is indeed a string with at least length ~35 or something. It's not much, but then at least we have some tests to verify that it prints a string with `"... supports the following operations: ..."`.

stevenbird: Yes, let's just treat that as a separate issue.

stevenbird:
Thanks @dannysepler, @tomaarsen, @iliakur

PR #2803, opened 2021-09-11
build: add matplotlib in deps section

12mohaned:
Add dependency for missing package **Matplotlib**

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,6 +29,7 @@ deps =
     click
     joblib
     tqdm
+    matplotlib
 
 changedir = nltk/test
 commands =


tomaarsen (approved):
This seems necessary indeed. Running e.g. `tox -e py39` will currently show failing tests, because of a the following test:
https://github.com/nltk/nltk/blob/d4e8c3c0e99de11f23d19cfe060591a7afb738c3/nltk/test/unit/test_cfd_mutation.py#L16-L20 

This test will throw a `ModuleNotFoundError` with `No module named 'matplotlib'`. Similarly, a `ValueError: The plot function requires matplotlib to be installed. See http://matplotlib.org/` is thrown.

Adding `matplotlib` like so to `tox.ini` has resolved this issue.

dannysepler:
+1, very nice!

PR #2804, opened 2021-09-13
refactor: refactor sentiment analyzer by removing dead and slow perfomance code

12mohaned:
The pull request refactors the sentiment analyser class by generally doing the following points:

- Removing dead code like sentiment value in for loop(which made my curious why it is defined in all_words while not used) 
- Replace '==' in checking boolean values with the variable value. For example refactor if value == False with if not value: (performance reasons) 

A single if statement or even dozens won not be that bad, but hundred of if's statement, this may cause some minor performance issues. The below numbers illustrates better. 

`python -m timeit -s "variable=False" "if variable == True: pass"
`
output: 

> 10000000 loops, best of 5: 28.7 nsec per loop

`python -m timeit -s "variable=True" "if variable is True: pass"`

> 10000000 loops, best of 5: 21 nsec per loop

`python -m timeit -s "variable=True" "if variable: pass"`

> 20000000 loops, best of 5: 13.6 nsec per loop

I believe this may affect the performance of the sentiment analyser in a minor way.

### DIFF
--- a/nltk/sentiment/sentiment_analyzer.py
+++ b/nltk/sentiment/sentiment_analyzer.py
@@ -47,10 +47,10 @@ class SentimentAnalyzer:
         all_words = []
         if labeled is None:
             labeled = documents and isinstance(documents[0], tuple)
-        if labeled == True:
-            for words, sentiment in documents:
-                all_words.extend(words)
-        elif labeled == False:
+        if labeled is True:
+            for words in documents:
+                all_words.extend(words[0])
+        elif labeled is False:
             for words in documents:
                 all_words.extend(words)
         return all_words
@@ -218,7 +218,7 @@ class SentimentAnalyzer:
             classifier = self.classifier
         print(f"Evaluating {type(classifier).__name__} results...")
         metrics_results = {}
-        if accuracy == True:
+        if accuracy is True:
             accuracy_score = eval_accuracy(classifier, test_set)
             metrics_results["Accuracy"] = accuracy_score
 
@@ -232,22 +232,22 @@ class SentimentAnalyzer:
             test_results[observed].add(i)
 
         for label in labels:
-            if precision == True:
+            if precision is True:
                 precision_score = eval_precision(
                     gold_results[label], test_results[label]
                 )
                 metrics_results[f"Precision [{label}]"] = precision_score
-            if recall == True:
+            if recall is True:
                 recall_score = eval_recall(gold_results[label], test_results[label])
                 metrics_results[f"Recall [{label}]"] = recall_score
-            if f_measure == True:
+            if f_measure is True:
                 f_measure_score = eval_f_measure(
                     gold_results[label], test_results[label]
                 )
                 metrics_results[f"F-measure [{label}]"] = f_measure_score
 
         # Print evaluation results (in alphabetical order)
-        if verbose == True:
+        if verbose is True:
             for result in sorted(metrics_results):
                 print(f"{result}: {metrics_results[result]}")
 
--- a/nltk/sentiment/sentiment_analyzer.py
+++ b/nltk/sentiment/sentiment_analyzer.py
@@ -47,10 +47,10 @@ class SentimentAnalyzer:
         all_words = []
         if labeled is None:
             labeled = documents and isinstance(documents[0], tuple)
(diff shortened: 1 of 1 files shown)


nltk/sentiment/sentiment_analyzer.py
```diff
@@ -47,10 +47,10 @@ def all_words(self, documents, labeled=None):
         all_words = []
         if labeled is None:
             labeled = documents and isinstance(documents[0], tuple)
-        if labeled == True:
-            for words, sentiment in documents:
-                all_words.extend(words)
-        elif labeled == False:
+        if labeled is True:
+            for words in documents:
```
purificant: small note, maybe call this variable something else? Because this is no longer `words` and is now a (words, label) tuple of type (list(str), str)

tomaarsen:
I suppose that `is` is indeed slightly faster than `==`. For those unaware:
* `==` is for value equality. For when we want to ensure that two objects have the same value.
* `is` is for reference equality. For when we want to ensure that two references refer to the same object.

See [this SO](https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is) for more information.

With `is`, only the pointers need to be compared, while `==` has the additional step of traversing to the pointer's location and reading the value. I can see how this is (insignificantly) faster. 

Generally, the use of `is` is not recommended ([with the exception of singletons like `None`](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)), as
```python
>>> a = 2000
>>> b = 2000
>>> a == b
True
>>> a is b
False
```
It can also cause some really confusing bugs:
```python
>>> a = 2000
>>> b = 2000
>>> a is b
False
>>> c = 200
>>> d = 200
>>> c is d
True
```
[(more information on this "bug")](https://stackoverflow.com/a/28864111/541136)

In conclusion:
```python
Yes:   if greeting:
No:    if greeting == True:
Worse: if greeting is True:
```
[(source)](https://stackoverflow.com/a/48350377)

So, in short, I'm in favor of this PR, with the exception of the following:

nltk/sentiment/sentiment_analyzer.py
```diff
@@ -47,10 +47,10 @@ def all_words(self, documents, labeled=None):
         all_words = []
         if labeled is None:
             labeled = documents and isinstance(documents[0], tuple)
-        if labeled == True:
-            for words, sentiment in documents:
-                all_words.extend(words)
-        elif labeled == False:
+        if labeled:
+            for words in documents:
+                all_words.extend(words[0])
```
tomaarsen:
```suggestion
            for words, _sentiment in documents:
                all_words.extend(words)
```
I believe this is the best solution - this way no decent IDE or linter will complain about `_sentiment` not being used, but a developer can still read and understand that `document` must be a tuple of `(words, sentiment)`, which is valuable information in understanding the code.

That said, others might prefer other solutions, and in a previous life I would have preferred
```python
for words, _ in documents:
    all_words.extend(words)
```
So if someone can convince me that your solution is best, then I won't have issues with it.

tomaarsen (changes requested):

purificant:
I like this, using `if foo:` is more pythonic than doing `if foo == True:` or `if foo is True:` :+1:

nltk/sentiment/sentiment_analyzer.py
12mohaned: I believe that the below snippet is somehow misleading. 

```
for words, _ in documents:
    all_words.extend(words)
```
Since you removed the sentiment variable, so it does not clarify what it really does. 

I am okay with the changes, but I am going to leave it for a while, if someone has different POV.

12mohaned:
I did add the suggested changes.

tomaarsen (approved):

purificant (approved):

nltk/sentiment/sentiment_analyzer.py
tomaarsen: @purificant I believe this is outdated - the PR now uses:
```python
            for words, _sentiment in documents:
                all_words.extend(words)
```

tomaarsen:
@12mohaned Thank you for your changes! I can always respect a code quality PR.
@purificant Thank you for the reviews. 👍

PR #2808, opened 2021-09-18
replace travis badge with github actions badge

purificant:
This PR replaces Travis badge with Github Actions badge. The new badge is pinned for builds in `develop` branch.

The change is:
travis
=>
github_actions

"ci-workflow" text comes from the name of the workflow in https://github.com/nltk/nltk/blob/1d87106312be12b8510890be42c329c00e1b425f/.github/workflows/ci.yaml#L1

Ideas are welcome if we should rename that to something else.


dannysepler:
thanks @purificant!

PR #2810, opened 2021-09-19
Deprecate `nltk.usage(obj)` in favor of `help(obj)`

### LINKED ISSUE #2807: usage function: repair or drop?
Do we need to support this usage method?
https://github.com/nltk/nltk/blob/develop/nltk/util.py#L40

Here's a discussion of it's shortcomings:
https://github.com/nltk/nltk/pull/2801#discussion_r706567025

tomaarsen:
Fixes #2807 

Hello!

### Pull request overview
- Started deprecation of `nltk.usage(obj)`. In this PR, it throws a warning and then calls `help(obj)`.
- Removed tests for `nltk.usage`.
- Removed a `classify.doctest` test that used `nltk.usage`, and reworded the documentation there.

See #2807 for discussion on the deprecation.

---

### Changes
My idea is to roll out a somewhat smooth deprecation. Rather than simply removing the function in the next version, it'll start throwing a warning indicating the pending deprecation. Then, because `nltk.usage()` was broken for certain classes (e.g. `nltk.FreqDict`), it now calls `help(obj)` instead.

If this is merged, then another PR can be created some time in the future, for a future version, which fully removes this function.

Please post any discussion regarding whether `nltk.usage` ought to be deprecated or not in #2807, and feel free to comment here on any suggestions about how the deprecation should be rolled out.

### Note
The warning will only be displayed if Python is ran with e.g. a `-Wd` or `-Wall` flag, indicating that warnings must be shown.

- Tom Aarsen

### DIFF
--- a/nltk/test/classify.doctest
+++ b/nltk/test/classify.doctest
@@ -11,16 +11,13 @@
 Classifiers label tokens with category labels (or *class labels*).
 Typically, labels are represented with strings (such as ``"health"``
 or ``"sports"``.  In NLTK, classifiers are defined using classes that
-implement the `ClassifyI` interface:
+implement the `ClassifierI` interface, which supports the following operations:
 
-    >>> import nltk
-    >>> nltk.usage(nltk.classify.ClassifierI)
-    ClassifierI supports the following operations:
-      - self.classify(featureset)
-      - self.classify_many(featuresets)
-      - self.labels()
-      - self.prob_classify(featureset)
-      - self.prob_classify_many(featuresets)
+- self.classify(featureset)
+- self.classify_many(featuresets)
+- self.labels()
+- self.prob_classify(featureset)
+- self.prob_classify_many(featuresets)
 
 NLTK defines several classifier classes:
 
@@ -42,6 +39,7 @@ We define a very simple training corpus with 3 binary features: ['a',
 that the correct answers can be calculated analytically (although we
 haven't done this yet for all tests).
 
+    >>> import nltk
     >>> train = [
     ...     (dict(a=1,b=1,c=1), 'y'),
     ...     (dict(a=1,b=1,c=1), 'x'),
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -1,58 +1,6 @@
 import pytest
 
-from nltk.util import everygrams, usage
-
-
-def test_usage_with_self(capsys):
-    class MyClass:
-        def kwargs(self, a=1):
-            ...
-
-        def no_args(self):
-            ...
-
-        def pos_args(self, a, b):
-            ...
-
-        def pos_args_and_kwargs(self, a, b, c=1):
-            ...
-
-    usage(MyClass)
-
-    captured = capsys.readouterr()
-    assert captured.out == (
-        "MyClass supports the following operations:\n"
-        "  - self.kwargs(a=1)\n"
-        "  - self.no_args()\n"
-        "  - self.pos_args(a, b)\n"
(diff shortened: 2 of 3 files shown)


stevenbird:
@tomaarsen good idea

iliakur:
Would it make sense to use `logging.warning` here to display the message without expecting the user to pass flags to Python?

tomaarsen:
`warnings.warn` is used some 34 times in the codebase, while `logging.warn` is never used.
That said, using `logging` does mean that all users are shown the warning, unless they change the logging settings.
I'm up for either warnings or logging.

iliakur:
I'd say let's stay consistent in this case and use warnings. I wonder how many people actually know that the output of that module is suppressed by default. I was definitely surprised to learn this.

tomaarsen:
It definitely surprised me too. I wouldn't have expected it.

tomaarsen:
It seems that one place in the codebase use this sneaky workaround to ensure that the warning is thrown:
https://github.com/nltk/nltk/blob/d21646dbd547cdd02d0c60f8e23d1d28a9fd1266/nltk/tokenize/stanford_segmenter.py#L71-L82

However, I'm definitely not a fan of exactly how it is done, as people often have reasons for their warning filtering, and this just sets it back to `"ignore"` afterwards, even if users set it to `"always"` themselves.

---

But, perhaps gathering the current filter(s), setting to `"always"`, raising the DeprecationWarning, and resetting the filter to what it was previously might be smart. I'm not sure if that's "proper" though.

Oh, and regarding the StanfordSegmenter snippet, I'll have a look at that.

tomaarsen:
Actually, we can also just use this:
https://github.com/nltk/nltk/blob/d21646dbd547cdd02d0c60f8e23d1d28a9fd1266/nltk/internals.py#L405-L434

iliakur:
Ah, good find! Use that :)

One thing to check for the future is if [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps) can be used to update the meta-data of the function being decorated. My guess is yes, but maybe there's some backwards-compatibility issue I'm not aware of.

tomaarsen:
@iliakur I was under the impression that `functools.wraps` has issues preserving the docstring and signature of functions of wrapped methods. At least, that was my conclusion when I attempted to use it a few years back. 

I managed to find the commit that described my issue: https://github.com/tomaarsen/TheNounProjectAPI/commit/d338927f62ee43995c51cb4fc28a8355ba93276d.

When using `@wraps` on a method, the signatures of wrapped functions into `*args, **kwargs`, i.e. the signature of the wrapper. That said, I might have just used it wrong.

And I will simply start using `@deprecated`.

iliakur:
Ah, I'll have to look into this, it's something new to me. I don't actually use `wraps` that often myself. We could consider adding `wrapt` as a dependency. Or [the decorator package](https://pypi.org/project/decorator/)?

tomaarsen:
I've made the update - it now uses `@deprecated` (Also, apologies to the GitHub user who has that as their username, I didn't mean to tag you in my commit message)

The output is now as follows:
```python
>>> import nltk
>>> nltk.usage(nltk.tokenize.NLTKWordTokenizer) 
<stdin>:1: DeprecationWarning: 
  Function usage() has been deprecated.  Use help(obj) instead.
NLTKWordTokenizer supports the following operations:
  - self.span_tokenize(s)
  - self.span_tokenize_sents(strings)
  - self.tokenize(text, convert_parentheses=False, return_str=False)
  - self.tokenize_sents(strings)
```

iliakur (approved):

stevenbird:
Thanks @tomaarsen, @iliakur

PR #2813, opened 2021-09-22
refactor:tokenize the text if the input is string.

12mohaned:
The pos_tag method past implementation had a problem with tagging un-tokenized strings. This [issue](https://github.com/nltk/nltk/issues/2648) illustrates the problem.

The suggested changes to solve the problem was raising an error or warning to the user that the pos_tag does not support the type string.

I believe a better approach might be tokenising the string into tokens.

### DIFF
--- a/nltk/tag/__init__.py
+++ b/nltk/tag/__init__.py
@@ -90,7 +90,7 @@ from nltk.tag.senna import SennaTagger, SennaChunkTagger, SennaNERTagger
 from nltk.tag.mapping import tagset_mapping, map_tag
 from nltk.tag.crf import CRFTagger
 from nltk.tag.perceptron import PerceptronTagger
-
+from nltk.tokenize import word_tokenize
 from nltk.data import load, find
 
 RUS_PICKLE = (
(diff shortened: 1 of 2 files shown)


tomaarsen:
I'm not sure that tokenising the input if the user doesn't use the function correctly is the best course of action. Users may start thinking that `pos_tag` works with strings, and they might not get the desired tokenization for their use case. They would never realise that there are other tokenization options out there for them to use.

NLTK itself provides 15+, and there are countless of other modules that provide tokenization too:
https://github.com/nltk/nltk/blob/d21646dbd547cdd02d0c60f8e23d1d28a9fd1266/nltk/tokenize/__init__.py#L65-L92

I'm all in favor of helping out the average developer, but I believe this is restricting them. I would rather keep throwing a `TypeError` and let them know to use one of the many tokenization options before using `pos_tag`.

Beyond that argument, I'm of the opinion that preserving tokenization as a "real" step in the NLP pipeline is somewhat important. I'd rather not have developers think they have no influence over that.

Perhaps others might disagree with me here, and I'm happy to go with whatever option seems best according to the rest of the team, but these are my 2 cents.

My perfect solution to this problem is simply to edit the `TypeError` to say, for example:
```python
raise TypeError("tokens: expected a list of strings, got a string.\nConsider tokenization, for example with `nltk.word_tokenize`.")
```
<sup> (phrasing can be modified to whatever sounds best, this is just to give an idea) </sup>

This way, developers are notified that they should tokenize. It mentions a quick and easy solution, but the developers know that they can modify the tokenization to their likings.

12mohaned:
I agree with the fact that tokenisation is an important preprocessing step in the NLP pipeline, and changing the format of input will theoretically affect the flow of the system. But, I though of this as a work around of solving the problem instead of throwing an error. In fact, anyone with humble knowledge with NLTK and NLP in general will know the processing of tagging tokens, but there is simple solutions like TextBlob, which tokenise string directly as far as I know. But of course, this change might confuse developers using NLTK.

stevenbird:
> I'm not sure that tokenising the input if the user doesn't use the function correctly is the best course of action. Users may start thinking that pos_tag works with strings, and they might not get the desired tokenization for their use case. They would never realise that there are other tokenization options out there for them to use.

Precisely. Ad hoc polymorphism is not going to do us any favours, just encourage sloppiness.

Issue #2770, opened 2021-07-31
GitHub Actions CI does not download third party dependencies


tomaarsen:
Hello!

The Travis CI used to use https://github.com/nltk/nltk/blob/develop/tools/travis/third-party.sh to ensure that the Stanford CoreNLP, Parser, POSTagger and SENNA are downloaded before running the CI. The current Github Actions CI workflow does not do this, and more files are lacking too.

These are all the skipped and xfailed tests on the current develop branch:
```
=========================== short test summary info ============================
SKIPPED [2] nltk/test/childes_fixt.py:9: The CHILDES corpus is not found. It should be manually downloaded and saved/unpacked to [NLTK_Data_Dir]/corpora/childes/
SKIPPED [1] nltk/test/inference_fixt.py:10: Mace4/Prover9 is not available so inference.doctest was skipped
SKIPPED [1] nltk/test/nonmonotonic_fixt.py:10: Mace4/Prover9 is not available so nonmonotonic.doctest was skipped
SKIPPED [1] nltk/test/unit/test_brill.py:33: Should be tested in __main__ of nltk.tbl.demo
SKIPPED [1] nltk/test/unit/test_classify.py:35: 

===========================================================================
NLTK was unable to find the megam file!
Use software specific configuration parameters or set the MEGAM environment variable.

  For more information on megam, see:
    <http://www.umiacs.umd.edu/~hal/megam/index.html>
===========================================================================
SKIPPED [6] nltk/test/unit/test_corenlp.py: Skipping all CoreNLP tests.
SKIPPED [1] nltk/test/unit/test_corpora.py:213: A full installation of the Penn Treebank is not available
SKIPPED [1] nltk/test/unit/test_corpora.py:190: A full installation of the Penn Treebank is not available
SKIPPED [1] nltk/test/unit/test_corpora.py:207: A full installation of the Penn Treebank is not available
SKIPPED [2] nltk/test/unit/test_corpora.py: Skipping test for mwa_ppdb.
SKIPPED [1] nltk/test/unit/test_rte_classify.py:88: Skipping tests with dependencies on MEGAM
SKIPPED [4] ../../../../../opt/hostedtoolcache/Python/3.6.14/x64/lib/python3.6/site-packages/_pytest/unittest.py:153: Requires Senna executable
SKIPPED [1] nltk/test/gluesemantics_malt_fixt.py:8: MaltParser is not available
[...]

PR #2820, opened 2021-09-27
Enable automated CoreNLP, Senna, Prover9/Mace4, Megam, MaltParser CI tests

tomaarsen:
Related: #2770

Hello!

### Pull request overview:
* Create bash script that downloads third party tools, including:
  * [Stanford tools](https://nlp.stanford.edu/software/):
    * CoreNLP
    * POSTagger
    * Parser
  * [SENNA](https://ronan.collobert.com/senna/)
  * [MEGAM](https://users.umiacs.umd.edu/~hal/megam/index.html)
  * [Prover9/Mace4](https://www.cs.unm.edu/~mccune/prover9/)
  * [MaltParser](http://www.maltparser.org/)
* Add CI job that uses and caches the output from this bash script.
* Deprecated `BFGS` as a valid algorithm for `rte_classifier`.
* Added `expanduser` in `find_jar_iter` in `internals.py`.
* Updated `MaltParser` doctest.
* Resolved replace-all issue in `inference.doctest`, and modified test output.
* Resolved incorrect import of `CoreNLPServer`.
* Removed automatic skip of `CoreNLP` and `MEGAM` tests.

### Consequences in short
The GitHub Actions Continuous Integration will download and cache some third party tools for the Linux test runner, allowing testing of code that uses third party tools. This has reduced the number of skipped tests from 31 to 16 for `ubuntu-latest` tests. `macos-latest` and `windows-latest` tests remain largely unchanged.

### Main Changes
A new job, `cache_third_party` was created in `.github/workflows/ci.yaml`. This job executes sequentially to `cache_nltk_data`, and runs only on `ubuntu-latest`. All it does is run a bash script called `tools/github_actions/third-party.sh`, and cache the generated folder. 
This bash script generates a folder with the path of `~/third`. Within this directory, more folders are created with the automatically downloaded tools. The script is based on `tools/travis/third-party.sh`, but downloads even more tools. It downloads Stanford's CoreNLP, POSTagger, Parser, SENNA, Prover9/Mace4, MEGAM and MaltParser, and includes some commented code to download TADM.
TADM is tricky, as it requires more installation after the download. This is something that can be done in the future, but I've left it out for now. 

As of right now, this downloads and then caches roughly 1 GB of data.

In order to get the tests for these third party tools to pass, a bunch of changes needed to be made. I'll go over each of them here, to explain why I made them:

### Deprecated `BFGS` as a valid algorithm for `rte_classifier`.
This involves the function `rte_classifier` in `nltk/classify/rte_classify.py`, which according to its documentation supports these algorithms:
https://github.com/nltk/nltk/blob/e7420b1500fd20133c2bed0eb2cf3611f7574c25/nltk/classify/rte_classify.py#L175-L178

Out of these, `"MEGAM"` and `"BFGS"` are both "MEGAM based algorithms", and require MEGAM to be configured. The `check_megam_config()` that existed was broken, so I removed that, as it wasn't strictly necessary: the check was already being done more downstream regardless. 
Then, `MaxentClassifier.train` gets called  with the algorithm as a parameter, but this function doesn't support an algorithm called `"BFGS"`:
https://github.com/nltk/nltk/blob/e7420b1500fd20133c2bed0eb2cf3611f7574c25/nltk/classify/maxent.py#L242-L244

I can't find any evidence that it ever did, either. Now, I've never used `rte_classifier`, or anything from the `classify` folder, but it seems that `rte_classifier` shouldn't support `"BFGS"` either, if it relies on `MaxentClassifier.train`. So, I removed it as one of the options.

### Added `expanduser` in `find_jar_iter` in `internals.py`.
I found out that if a user uses `~` in an environment variable, then that path is never expanded to the home directory of the user's operating system. It is simply left as a string `~`, and will then never result in a file being found. At least, that seems to be the behaviour.
I ended up not using `~` in the environment variables for the CI because of this, but simply adding `path = os.path.expanduser(path)` should solve this issue. I've only added it for `find_jar_iter` for now, but it likely should be added everywhere, allowing NLTK users everywhere to use `~` in their environment variables.

However, it should be noted that files on e.g. linux can include `~`, without it specifying something about the path relative from the home directory. In those cases this wouldn't work.

### Updated `MaltParser` doctest.
The `MaltParser` doctest was broken, as it's constructor requires a tagger *function*, and a tagger *object* was given instead. Beyond that, the mandatory `parser_dirname` value was left out. After that, the `try-except` in `nltk/test/gluesemantics_malt_fixt.py` also requires an `AssertionError`, as that error is sometimes thrown when the MaltParser tool cannot be found.

However, the fact that `MaltParser` takes a `parser_dirname` parameter means that the tests are also only configured for one specific version: `1.7.2`, while this is not the most recent version whatsoever (which is now `1.9.1` or something). Perhaps this is something to change in the future.

### Resolved replace-all issue in `inference.doctest`, and modified test output.
In `inference.doctest`, there was an occurrence of `Alread_expression`, while `AllExpression` was way more logical. As it turns out, an old commit c2988cc0f5c5a950544f8ef1024f38cea01373d9 replaced all occurrences of `llexpr` with `read_expr`, when a function was being renamed. This replace-all has also accidentally modified this test. I've fixed this now.

Beyond that, the expected output and the output which was reported differed slightly. For example, `0.000` was expected, while `0` was given by the tool, and the tool gave one blank line less. I've changed this now, but it might mean that a test only works for a specific version of the tool. For context, I've only tested with this one version.

### Resolved incorrect import of `CoreNLPServer`.
Previously, importing was done like so:
https://github.com/nltk/nltk/blob/43f3e3096ade1a2564a20a66172d53bd140ac983/nltk/test/unit/test_corenlp.py#L10-L20

However, `CoreNLPServer` (and `CoreNLPServerError`) are not actually imported - only `corenlp` is. I've resolved that by using `corenlp.CoreNLPServer`.

### Removed automatic skip of `CoreNLP` and `MEGAM` tests.
As the linux CI tests now want to do tests using third party tools, the `CoreNLP` and `MEGAM` tests are no longer automatically skipped, but skipped only if the tool is not installed.

---

### Future changes
This PR is extensive, but doesn't include all features that could be thought of. During development, I came up with some features that might be developed in the future.
* **Support for other tools**
  The number of skipped tests has been reduced from 31 to 16, but can be reduced even further if more tools are automatically downloaded, such as:
    - TADM
    - Penn Treebank
    - CHILDES corpus
* **Expand user in internals**
  Expand `~` in paths used in environment variables to the home path of the user's operating system. This is as simple as wrapping a path with `path = os.path.expanduser(path)`. This probably ought to be done in some functions in `internals.py`. **However**, files on linux are allowed to have `~` in them. In some of those cases, you wouldn't want to expand the `~`.
* **MaltParser is fixed to 1.7.2**
  As you can see in the diff of this PR, the tests for MaltParser are fixed to version 1.7.2. Perhaps this means that users who have another version for personal use won't have passing tests? If so, that is a problem. Perhaps the best solution is to not even have to specify a version in the constructor. I'm not even sure it gets used - as I believe the MaltParser executable is actually loaded through the `MALT_PARSER` environment variable.
* **Download tools for Windows and Mac too**
  This PR downloads some cross-platform tools, and also some Unix only tools. Some of them would work for Mac and Ubuntu too, and some of them simply require downloading slightly different versions.
* **Move the script to a seperate repository, and simply check out that repository**
  This is an idea from #2770. We should store fixed versions of these third party tools in a separate repository, so we don't make too many requests of automatic downloads to these websites. I considered doing this already, but figured it was an extra step which I don't have the time investment for. 
  This would work with the previous change too - we can have e.g. 3 branches, for `ubuntu-latest`, `macos-latest` and `windows-latest`, respectively.
* **Update third party tools to newer versions (Stanford, MaltParser)**
  The Stanford and MaltParser tools have had newer versions released than the ones used in `third-party.sh`. Perhaps it would be wise to update these versions. This would be as simple as modifying the download link to point to the new version.

---

This has been a long time coming, and I'm glad to be able to release this "minimal viable product" of having third party downloads for CI.
I'd love to get some reviews on this - not just on the general changes, but also on the smaller changes that needed to be made to the tests, like the `"BFGS"` "deprecation" or test modifications.

- Tom Aarsen

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,12 +7,12 @@ jobs:
     name: Run pre-commit
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
-    - uses: pre-commit/action@v2.0.0
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.0
 
   cache_nltk_data:
-    name: cache nltk_data
+    name: Cache nltk_data
     needs: pre-commit
     strategy:
       matrix:
@@ -36,9 +36,35 @@ jobs:
         shell: bash
         if: steps.restore-cache.outputs.cache-hit != 'true'
 
+  cache_third_party:
+    name: Cache third party downloads
+    needs: pre-commit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Java
+        uses: actions/setup-java@v1
+
+      - name: Cache third party downloads
+        uses: actions/cache@v2
+        id: restore-cache
+        with:
+          path: ~/third_party
+          key: third_party_${{ secrets.CACHE_VERSION }}
+
+      - name: Download third party data
+        run: |
+          # Ensure a third_party folder exists.
+          mkdir -p third_party
+          cd third_party
+          # Download the zip
+          wget "https://huggingface.co/stanfordnlp/CoreNLP/resolve/main/stanford-corenlp-latest.zip" -O core_nlp.zip
+          unzip core_nlp.zip -d ./core_nlp
+          rm core_nlp.zip
+          export CLASSNAME="~/third_party/core_nlp"
+
   test:
     name: Python ${{ matrix.python-version }} on ${{ matrix.os }}
-    needs: cache_nltk_data
+    needs: [cache_nltk_data, cache_third_party]
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,7 +52,10 @@ jobs:
           key: third_${{ secrets.CACHE_VERSION }}
 
       - name: Download third party data
-        run: ./tools/github_actions/third-party.sh
+        run: |
+          pwd
+          ls
+          ./tools/github_actions/third-party.sh
         if: steps.restore-cache.outputs.cache-hit != 'true'
 
   test:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,4 +103,4 @@ jobs:
         run: pytest -v --numprocesses auto -rsx nltk/test
         shell: bash
         env:
-          CLASSNAME: "~/third_party/core_nlp"
+          CLASSPATH: "~/third_party/core_nlp"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,10 +53,8 @@ jobs:
 
       - name: Download third party data
         run: |
-          pwd
-          ls
-          file ./tools/github_actions/third-party.sh
-          dos2unix ./tools/github_actions/third-party.sh
+          chmod +x ./tools/github_actions/third-party.sh
+          ./tools/github_actions/third-party.sh
         if: steps.restore-cache.outputs.cache-hit != 'true'
 
   test:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,6 +102,9 @@ jobs:
         with:
           path: ~/third
           key: third_${{ secrets.CACHE_VERSION }}
+        if: runner.os == 'Linux'
+
+      - name: Set up third party environment variables
         shell: bash
         run: |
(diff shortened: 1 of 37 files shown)


dimazest (approved):
Looks good!

purificant (approved):

purificant:
Great job! LGTM :+1:

stevenbird:
Thank you @tomaarsen... big effort! 👏
Thanks for the reviews @dimazest, @purificant

PR #2822, opened 2021-09-27
Accept pre-tokenized references & hypothesis for METEOR calculation

yutanakamura-tky:
I added an option to input pre-tokenized sentence(s) for METEOR calculation.

At present, references & hypothesis are always tokenized with `str.split()` (Lines 30-31 of `meteor_score.py`):

```
    hypothesis_list = list(enumerate(preprocess(hypothesis).split()))
    reference_list = list(enumerate(preprocess(reference).split()))
```

However, we have trouble calculating METEOR in a language that does not separate words with spaces (e.g., Japanese).

In this pull request, I have changed `_generate_enums` to accept pre-tokenized reference & hypothesis in addition to untokenized reference & hypothesis so that we can use a customized tokenizer.

Other functions have been changed accordingly.

### DIFF
--- a/nltk/translate/meteor_score.py
+++ b/nltk/translate/meteor_score.py
@@ -56,7 +56,9 @@ def exact_match(hypothesis, reference, preprocess=str.lower):
              enumerated unmatched reference tuples
     :rtype: list of 2D tuples, list of 2D tuples,  list of 2D tuples
     """
-    hypothesis_list, reference_list = _generate_enums(hypothesis, reference, preprocess=preprocess)
+    hypothesis_list, reference_list = _generate_enums(
+        hypothesis, reference, preprocess=preprocess
+    )
     return _match_enums(hypothesis_list, reference_list)
 
 
@@ -134,7 +136,9 @@ def stem_match(hypothesis, reference, stemmer=PorterStemmer(), preprocess=str.lo
              enumerated unmatched reference tuples
     :rtype: list of 2D tuples, list of 2D tuples,  list of 2D tuples
     """
-    enum_hypothesis_list, enum_reference_list = _generate_enums(hypothesis, reference, preprocess=preprocess)
+    enum_hypothesis_list, enum_reference_list = _generate_enums(
+        hypothesis, reference, preprocess=preprocess
+    )
     return _enum_stem_match(enum_hypothesis_list, enum_reference_list, stemmer=stemmer)
 
 
@@ -190,7 +194,9 @@ def wordnetsyn_match(hypothesis, reference, wordnet=wordnet, preprocess=str.lowe
     :return: list of mapped tuples
     :rtype: list of tuples
     """
-    enum_hypothesis_list, enum_reference_list = _generate_enums(hypothesis, reference, preprocess=preprocess)
+    enum_hypothesis_list, enum_reference_list = _generate_enums(
+        hypothesis, reference, preprocess=preprocess
+    )
     return _enum_wordnetsyn_match(
         enum_hypothesis_list, enum_reference_list, wordnet=wordnet
     )
@@ -237,7 +243,13 @@ def _enum_align_words(
     )
 
 
-def align_words(hypothesis, reference, stemmer=PorterStemmer(), wordnet=wordnet, preprocess=str.lower):
+def align_words(
+    hypothesis,
+    reference,
+    stemmer=PorterStemmer(),
+    wordnet=wordnet,
+    preprocess=str.lower
+):
     """
     Aligns/matches words in the hypothesis to reference by sequentially
     applying exact match, stemmed match and wordnet based synonym match.
@@ -257,7 +269,9 @@ def align_words(hypothesis, reference, stemmer=PorterStemmer(), wordnet=wordnet,
     :return: sorted list of matched tuples, unmatched hypothesis list, unmatched reference list
     :rtype: list of tuples, list of tuples, list of tuples
     """
-    enum_hypothesis_list, enum_reference_list = _generate_enums(hypothesis, reference, preprocess=preprocess)
+    enum_hypothesis_list, enum_reference_list = _generate_enums(
+        hypothesis, reference, preprocess=preprocess
+    )
     return _enum_align_words(
         enum_hypothesis_list, enum_reference_list, stemmer=stemmer, wordnet=wordnet
     )
--- a/nltk/translate/meteor_score.py
+++ b/nltk/translate/meteor_score.py
@@ -33,7 +33,7 @@ def _generate_enums(hypothesis, reference, preprocess=str.lower):
     return hypothesis_list, reference_list
 
 
-def exact_match(hypothesis, reference, preprocess=str.lower):
+def exact_match(hypothesis, reference, preprocess):
     """
     matches exact words in hypothesis and reference
     and returns a word mapping based on the enumerated
@@ -43,7 +43,7 @@ def exact_match(hypothesis, reference, preprocess=str.lower):
     :type hypothesis: list(str)
     :param reference: pre-tokenized reference
     :type reference: list(str)
-    :preprocess: preprocessing method (default str.lower)
+    :preprocess: preprocessing method
     :type preprocess: method
     :return: enumerated matched tuples, enumerated unmatched hypothesis tuples,
              enumerated unmatched reference tuples
@@ -111,7 +111,7 @@ def _enum_stem_match(
     return _match_enums(stemmed_enum_list1, stemmed_enum_list2)
 
 
-def stem_match(hypothesis, reference, stemmer=PorterStemmer(), preprocess=str.lower):
+def stem_match(hypothesis, reference, stemmer=PorterStemmer()):
     """
     Stems each word and matches them in hypothesis and reference
     and returns a word mapping between hypothesis and reference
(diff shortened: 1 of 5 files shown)


tomaarsen:
@yutanakamura-tky Heya! We've started using https://pre-commit.com/ pretty recently. With this program, some tests are ran before it lets you commit. With [our configuration](https://github.com/nltk/nltk/blob/develop/.pre-commit-config.yaml), it means that some commands are ran to ensure some code quality. In particular, the `black` formatter. As you can see in the [test output](https://github.com/nltk/nltk/pull/2822/checks?check_run_id=3722726809), the `black` formatter complained that the changes you are trying to push are not formatted according to the `black` formatting rules.

My recommendation is to run `pre-commit install` (`pre-commit` should have been `pip install`-ed already, but otherwise you can use `pip install pre-commit`). This will install the pre-commit hooks, causing it to run all of those hooks (e.g. formatters, verifiers, etc.) when you make another commit. If you then try to push another commit, the `black` hook will reformat your changes, and you can push the correctly formatted changes. This might seem a little bit confusing, but it's more intuitive in practice. Oh, also, `pre-commit` only affects this repository, you don't have to worry about it affecting other projects.

Upon installing, you may run `pre-commit run --all`, which applies formatting and such on all files, and in particular on your changes. You may push these changes, so the Continuous Integration stops complaining.

Let me know if you have any issues.

yutanakamura-tky:
@tomaarsen 
Thank you! I have updated this pull request using `pre-commit`.

stevenbird:
I would prefer to require tokenised input (a breaking change) than to have `str or list(str)` parameter

yutanakamura-tky:
@stevenbird 
Thank you! I have made the following changes:

#### 1. Enforcement of tokenized inputs

I have updated the function implementations, docstrings, and examples.

#### 2. Modification of the test code
I have deleted the following code in `nltk/test/unit/translate/test_meteor.py` because it assumed that the input was given not in `list(str)` but `str` (Lines 12-13):

```
        # no `preprocess` argument
        self.assertRaises(TypeError, meteor_score, reference, candidate)
```

I have also changed the code as follows to examine better if the preprocessing works correctly:
```
        reference = [["this", "is", "a", "test"], ["this", "is" "test"]]
        candidate = ["THIS", "Is", "a", "tEST"]

        score = meteor_score(reference, candidate, preprocess=str.lower)
```

I have confirmed locally that it passes `pytest nltk/test/unit/translate/test_meteor.py`.

yutanakamura-tky:
The force-push from 5db9c72 to ea66229 was made to amend the commit message of 3d0d272 from "accept pre-tokenized references & hypothesis" to "applied code formatter."
Sorry for any inconvenience.

nltk/translate/meteor_score.py
```diff
@@ -38,10 +49,10 @@ def exact_match(hypothesis, reference):
...
     word id between hypothesis and reference
 
-    :param hypothesis: hypothesis string
-    :type hypothesis: str
-    :param reference: reference string
-    :type reference: str
+    :param hypothesis: pre-tokenized hypothesis
+    :type hypothesis: array-like(str)
+    :param reference: pre-tokenized reference
+    :type hypothesis: array-like(str)
```
tomaarsen:
```suggestion
    :type reference: array-like(str)
```
I believe this is a small copy-paste error. No biggie.

nltk/translate/meteor_score.py
```diff
@@ -111,10 +122,10 @@ def stem_match(hypothesis, reference, stemmer=PorterStemmer()):
...
     and returns a word mapping between hypothesis and reference
 
-    :param hypothesis:
-    :type hypothesis:
-    :param reference:
-    :type reference:
+    :param hypothesis: pre-tokenized hypothesis
+    :type hypothesis: array-like(str)
+    :param reference: pre-tokenized reference
+    :type hypothesis: array-like(str)
```
tomaarsen:
```suggestion
    :type reference: array-like(str)
```
idem.

nltk/translate/meteor_score.py
```diff
@@ -296,13 +316,13 @@ def single_meteor_score(
...
-    >>> round(meteor_score('this is a cat', 'non matching hypothesis'),4)
+    >>> round(meteor_score(['this', 'is', 'a', 'cat'], ['non', 'matching', 'hypothesis']),4)
     0.0
 
-    :param reference: reference sentence
-    :type reference: str
-    :param hypothesis: a hypothesis sentence
-    :type hypothesis: str
+    :param reference: pre-tokenized reference
+    :type hypothesis: array-like(str)
```
tomaarsen:
```suggestion
    :type reference: array-like(str)
```
idem.

nltk/translate/meteor_score.py
```diff
@@ -376,13 +396,13 @@ def meteor_score(
...
-    >>> round(meteor_score(['this is a cat'], 'non matching hypothesis'),4)
+    >>> round(meteor_score([['this', 'is', 'a', 'cat']], ['non', 'matching', 'hypothesis']),4)
     0.0
 
-    :param references: reference sentences
-    :type references: list(str)
-    :param hypothesis: a hypothesis sentence
-    :type hypothesis: str
+    :param references: pre-tokenized reference sentences
+    :type hypothesis: 2D array-like(str)
```
tomaarsen:
```suggestion
    :type references: 2D array-like(str)
```
idem.

nltk/translate/meteor_score.py
```diff
@@ -8,27 +8,38 @@
 
 
 from itertools import chain, product
+from typing import Iterable
```
tomaarsen: Does this get used? I'm personally down to use the Python 3.5+ typing in signatures, and use e.g.
```python
def _generate_enums(
    hypothesis: Iterable[str],
    reference: Iterable[str],
    preprocess: Callable[[str], str] = str.lower
) -> Tuple[List[Tuple[str, str]], List[Tuple[str, str]]]:
    """...
    
    :param hypothesis: pre-tokenized hypothesis
    :type hypothesis: Iterable[str]
    :param reference: pre-tokenized reference
    :type reference: Iterable[str]
    :preprocess: preprocessing method (default str.lower)
    :type preprocess: Callable[[str], str]
    :return: enumerated words list
    :rtype: list of 2D tuples, list of 2D tuples
    :rtype: Tuple[List[Tuple[str, str]], List[Tuple[str, str]]]
    """
    ...
```
*However*, this isn't necessarily in the scope of your PR. This is a change I'd like to see throughout the entire codebase, so you can just leave it like it was if you prefer. At that point it might also be overkill to specify the type *both* in the signature and in the docstring.

tomaarsen:
> I would prefer to require tokenised input (a breaking change) than to have `str or list(str)` parameter

I believe that all the other translation scores, i.e. BLEU, CHRF, GLEU, NIST and RIBES also use pre-tokenized inputs (Correct me if I'm wrong). As a result, I'm all in favor of modifying the METEOR score so that it also takes tokenised input.

tomaarsen:
I've applied tokenization on the hypotheses and references in `meteor.doctest`, so the tests use tokenized inputs rather than the original string ones. The tests pass now! 🎉

yutanakamura-tky:
@tomaarsen 
Thank you for your review comments & updating `meteor.doctest`!
I have made some changes as follows:

### 1. Addition of Type Hints
I have added type hints to all methods in `meteor_score.py`.
Accordingly, I have removed type descriptions in the docstrings.

### 2. Minor Improvement of Local Variable Names
I have changed some local variable names to make their roles clearer.

- `_generate_enums()`
  - `hypothesis_list` -> `enum_hypothesis_list` 
  - `reference_list` -> `enum_reference_list` 
- `exact_match()`
  - `hypothesis_list` -> `enum_hypothesis_list` 
  - `reference_list` -> `enum_reference_list` 
- `_enum_stem_match()`
  - `stemmed_enum_list1` -> `stemmed_enum_hypothesis_list` 
  - `stemmed_enum_list2` -> `stemmed_enum_reference_list`

tomaarsen (approved):
Looks good to me! I'm a great fan of the modernisation, and the modified functionality is wonderful too, it makes the METEOR score calculation similar to the other translation score metrics.

yutanakamura-tky:
@tomaarsen 
Thank you so much!
I want just one more piece of advice about variable typing.

The ongoing docstring in the `develop` branch says that `meteor_score()` takes `float` for the variable `beta`, while it defaults to 3:

```
def meteor_score(
    references: Iterable[Iterable[str]],
    hypothesis: Iterable[str],
    preprocess: Callable[[str], str] = str.lower,
    stemmer: StemmerI = PorterStemmer(),
    wordnet: WordNetCorpusReader = wordnet,
    alpha: float = 0.9,
    beta: float = 3,
    gamma: float = 0.5,
)
```

The "3" derives from the power of penalty in the original METEOR article.
I am wondering which is better, `beta: Union[int, float]=3` or `beta: float=3.0`.

tomaarsen:
@yutanakamura-tky I think the best solution in that situation is simply to change `beta: float = 3` into `beta: float = 3.0`. That way it's extra clear that this value can be/is meant to be a float. That said, users are free to use e.g. `4`, because Python automatically converts ints to floats whenever requires anyways. 
Good catch, by the way!

yutanakamura-tky:
@tomaarsen 
Thank you so much! I have also made the same change to `single_meteor_score()`.
There are no further changes that I am planning to make.

tomaarsen:
Wonderful, it all looks great to me. I'll leave the PR like this for now, so perhaps another team member can have a look at it too.

dannysepler:
nice! i love the addition of type hints anywhere in NLTK

~~given that this is a breaking change, should we have a changelog entry? after that, lgtm!~~ (EDIT: Realizing now that only Steven has edited that file in recent history. Sorry!)

dannysepler (approved):

stevenbird:
Thank you @yutanakamura-tky, @tomaarsen, @dannysepler.

Witiko:
@stevenbird Please, consider incrementing at least the minor version the next time you introduce a breaking change.

stevenbird:
@Witiko: noted, my apologies

PR #2832, opened 2021-09-30
Updating third party tools to newer versions and removing MaltParser fixed version 

### LINKED ISSUE #2826: MaltParser is fixed to 1.7.2
From #2802 (@tomaarsen)

The tests for MaltParser are fixed to version 1.7.2. Perhaps this means that users who have another version for personal use won't have passing tests? If so, that is a problem. Perhaps the best solution is to not even have to specify a version in the constructor. I'm not even sure it gets used - as I believe the MaltParser executable is actually loaded through the `MALT_PARSER` environment variable.

### LINKED ISSUE #2827: Update third party tools to newer versions
From #2820 (@tomaarsen)

The Stanford and MaltParser tools have had newer versions released than the ones used in `third-party.sh`. This may be as simple as modifying the download link to point to the new versions.

gavishpoddar:
Hello,

This micro pull request updates third party tools to newer versions (Stanford  `corenlp`, `postagger`, `parser`) (#2826) and removes MaltParser fixed version. (#2827)

I am trying the fix the above-listed issues please let me know if changes are required.

Thanks


tools/travis/third-party.sh
```diff
@@ -8,8 +8,8 @@ pushd ${HOME}
 pushd 'third'
 
 # Download nltk stanford dependencies
-#stanford_corenlp_package_zip_name=$(curl -s 'https://stanfordnlp.github.io/CoreNLP/' | grep -o 'stanford-corenlp-full-.*\.zip' | head -n1)
-stanford_corenlp_package_zip_name="stanford-corenlp-full-2017-06-09.zip"
+#stanford_corenlp_package_zip_name=$(curl -s 'https://stanfordnlp.github.io/CoreNLP/history.html' | grep -o 'stanford-corenlp-full-.*\.zip' | head -n1)
```
tomaarsen: For context, the two issues you're trying to solve were listed in #2820. These issues were meant as "future changes" for after that PR is merged. As you might be able to see in that PR, there is a new file `tools/github_actions/third-party.sh`, which is based on this `travis` file.

We don't use Travis CI (https://travis-ci.org/) anymore, as we've replaced it with GitHub Actions. As a result, this file is no longer being used, and doesn't require updating.

nltk/test/gluesemantics_malt_fixt.py
```diff
@@ -4,6 +4,6 @@ def setup_module():
     from nltk.parse.malt import MaltParser
 
     try:
-        depparser = MaltParser("maltparser-1.7.2")
+        depparser = MaltParser("maltparser")
```
tomaarsen: I would have preferred to have a fix that involves no longer requiring `MaltParser` to take a `parser_dirname` parameter. It can become optional, but right now the flow is as follows:
* MaltParser gets initialized with a mandatory `parser_dirname`.
* This `parser_dirname` is used to find the `MaltParser` executable. If this fails, then the `MALT_PARSER` environment variable is looked for. If this also fails, then an error is thrown. 

However, this means that if a user wishes to initialize `MaltParser` with an environment variable (which is how I would prefer it), then the user must also still give a path to `MaltParser`? That is weird. Perhaps we can make `parser_dirname` optional, allowing for 2 ways to initialize the parser.

tomaarsen (commented):
Regarding the changes to the travis file - I'm afraid we won't need those. And for the MaltParser, we're looking for a more involved fix that tackles the weird dynamic I've described in my other comment.

It's up to you whether you want to try and tackle those issues, or if you want to leave it. If you choose the former, then maybe it would be best to wait until #2820 is either closed or merged.

gavishpoddar:
Thanks, @tomaarsen, I see that is now merged and I am trying to update this PR.

And if there is anything I can contribute please let me know 

Thanks

tomaarsen:
Of course! Perhaps the links in https://github.com/nltk/nltk/blob/develop/tools/github_actions/third-party.sh that gather the third party applications can be updated. Beyond that, any problem from https://github.com/nltk/nltk/issues can be tackled.

tomaarsen:
Apologies, I did not mean to close this.

tools/travis/third-party.sh
gavishpoddar: Updates 👍🏻

nltk/test/gluesemantics_malt_fixt.py
gavishpoddar: HI @tomaarsen, I have changed the default function param to `""` (empty string) this fixing the `parser_dirname` requirement issue. This somewhat archives the above goals. Please suggest if this is fine or changes are required or if the default param issues need to be solved in a more elegant way.

tomaarsen:
I'm trying to do some testing, but it seems that GitHub is having some issues with their `windows-latest` continuous integration, which is making it all a bit more difficult. I'll get back to this later.

tomaarsen (approved):
I've removed the `parser_dirname` parameter from a few test cases, where the path to the actual executable should be supplied via the `MALT_PARSER` environment variable. This would completely fix the issue #2826! I believe with your changes, this should pass just fine. 
If it does, then I'll merge this, and reset the CI cache so that it indeed downloads the new versions of MaltParser and the CoreNLP tools for all future test runs.

tomaarsen:
These changes look great. Thank you @gavishpoddar!

gavishpoddar:
Thanks, @tomaarsen for helping me throughout the PR and the merge

Issue #2833, opened 2021-10-01
[3.6.4] AttributeError: module 'regex' has no attribute 'Pattern'


astariul:
Since I upgraded `nltk` from `3.6.2` to `3.6.4`, the following code :

```python
from nltk.tokenize.casual import TweetTokenizer, casual_tokenize
```

raises the following error :

> Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/user/miniconda3/envs/test/lib/python3.7/site-packages/nltk/__init__.py", line 137, in <module>
    from nltk.text import *
  File "/home/user/miniconda3/envs/test/lib/python3.7/site-packages/nltk/text.py", line 29, in <module>
    from nltk.tokenize import sent_tokenize
  File "/home/user/miniconda3/envs/test/lib/python3.7/site-packages/nltk/tokenize/__init__.py", line 65, in <module>
    from nltk.tokenize.casual import TweetTokenizer, casual_tokenize
  File "/home/user/miniconda3/envs/test/lib/python3.7/site-packages/nltk/tokenize/casual.py", line 272, in <module>
    class TweetTokenizer:
  File "/home/user/miniconda3/envs/test/lib/python3.7/site-packages/nltk/tokenize/casual.py", line 357, in TweetTokenizer
    def WORD_RE(self) -> regex.Pattern:
**AttributeError: module 'regex' has no attribute 'Pattern'**

purificant:
Had a quick look and was unable to reproduce.

```
pip install --upgrade pip && pip install nltk==3.6.4
```

```python
>>> import nltk
>>> nltk.__version__
'3.6.4'
>>> from nltk.tokenize.casual import TweetTokenizer, casual_tokenize
>>> 
```

Edit: on python 3.9.7 / 3.7.12

tomaarsen:
~~It seems that the `re.Pattern` (and `regex.Pattern`) attributes were only introduced starting from Python 3.8+. How our tests didn't catch this - I have no idea. I'll make a fix.~~

Edit: This doesn't seem to be the issue - I think `regex.Pattern` is not defined on older versions of `regex`. A fix is updating `regex` to the newest version. See https://bitbucket.org/mrabarnett/mrab-regex/commits/26a320f29b8266562a67dfa2243feda1a8db40ac for more information.

tomaarsen:
~~@astariul For the time being, please use the develop branch instead of 3.6.4. This command ought to do it:~~

[...]

PR #2835, opened 2021-10-01
specify minimum regex version that supports regex.Pattern

purificant:
See excellent explanation by @tomaarsen here: https://github.com/nltk/nltk/pull/2834

This PR increases minimum required version of a dependency to solve a use case like https://github.com/nltk/nltk/issues/2833

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,7 +115,7 @@ natural language processing.  NLTK requires Python 3.6, 3.7, 3.8, or 3.9.""",
     install_requires=[
         "click",
         "joblib",
-        "regex",
+        "regex>=2021.8.3",
         "tqdm",
     ],
     extras_require=extras_require,


iliakur:
Hmm, the fact that we had to bump this in two places makes me wonder if we should consider introducing something like `poetry` to manage our dependency spec...

lgtm otherwise :)

iliakur (approved):

fernandesbreno:
@iliakur I was wondering If I could add poetry as a dependency manager in the project and unify all configs in the "pyproject.toml" to standardize the project like pep 518 says.

iliakur:
@breno-jesus-fernandes thanks for reaching out! I'm not very active in nltk development these days, so I think @stevenbird is in a better position to make this decision.

My two cents: the python package ecosystem is in a state of transition, so I'd hesitate to bet the farm on a specific tool right now. That being said, I do like poetry as a tool. I also like `pip-compile` though.

Issue #2637, opened 2020-12-08
SentimentIntensityAnalyzer() from nltk.sentiment.vader does not respond to hashtags. 


neldivad:
Example: 
```
SentimentIntensityAnalyzer.polarity_scores( 'Strings with hashtag #stupid #useless #BAD' )  
... compound: 0.0, 
... neg: 0.0, 
... neu: 1.0, 
... pos: 0.0,
```

PR #2840, opened 2021-10-04
feat: allow SentimentIntensityAnalyzer to preprocess hashtags

12mohaned:
This pull request address [issue](https://github.com/nltk/nltk/issues/2637). Hashtags affects the sentiment of text which affects the overall score and output in-accurate results.

This pull request does the following: 

- Add remove_hashtag function in the preprocessing pipeline to remove hashtags, in order to deal with words properly.

Regards.

### DIFF
--- a/nltk/sentiment/vader.py
+++ b/nltk/sentiment/vader.py
@@ -6,6 +6,7 @@
 #         Pierpaolo Pantone <24alsecondo@gmail.com> (modifications)
 #         George Berry <geb97@cornell.edu> (modifications)
 #         Malavika Suresh <malavika.suresh0794@gmail.com> (modifications)
+#         Mohaned Mashaly <>(modifications)
 # URL: <http://nltk.org/>
 # For license information, see LICENSE.TXT
 #
@@ -269,6 +270,7 @@ class SentiText:
         if not isinstance(text, str):
             text = str(text.encode("utf-8"))
         self.text = text
+        self.text = self.remove_hashtag(text)
         self.PUNC_LIST = punc_list
         self.REGEX_REMOVE_PUNCTUATION = regex_remove_punctuation
         self.words_and_emoticons = self._words_and_emoticons()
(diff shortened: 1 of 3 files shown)


stevenbird:
@12mohaned this change needs to be more carefully argued. Please consider the following.
– it removes hashes globally, when a hashtag is more specific
– other symbols could interfere with sentiment analysis
– any programmer concerned about these issues could preprocess the input

tomaarsen:
@stevenbird 
> – any programmer concerned about these issues could preprocess the input

I believe the core of the issue here is that developers might expect that hashtags will affect the sentiment analysis. With other words, this could also be resolved with a documentation update. 

Also, to add to your list, hashtags are not always just singular words. For example, `#terribleservice` wouldn't get picked up, regardless of whether there is a hashtag or not. Allowing "sentiment analysis of hashtags" might make developers suspect that even multi-word hashtags like my previous example would work.

12mohaned:
@stevenbird  I would not probably be concerned about removing hashes global, in my opinion I do not believe hashes are not that important since they do not carry any weight(Please, correct on this if I am wrong). Regarding the second point, there could a similar list to that of punctuation list which removes other sybmols (I do not know if this is a valid thing TBH). I partially agree with the third point, but I would argue that since NLTK is a preprocessing library. It might be useful to preprocess your input. For a safer option, we might add as an optional feature for **SentimentIntenseAnalyzer** class.

12mohaned:
@tomaarsen Well, It is a bit tricky, I do not suspect that removing a hashtag implies supporting multi-words. For example, I would not expect as a developer TweeterTokenizer to be able to deal with multi-words, just because it supports dealing with hashtags. Also, I am not sure wether this is hundred percent correct or not.

tomaarsen:
I do agree that there is a problem here: unexpected behaviour. That said, I don't know whether this problem is best solved with a documentation or a code change.

stevenbird:
In the case of `#terribleservice` we would want to remove the tag and tokenize the result. But in cases like `#metoo`, we might want to ascribe sentiment to the hashtag itself (and not to `metoo`, or `me`, or `too`).

I think an algorithmic solution should be informed by people who are active in this space, and that for now we just go with a note in the documentation.

12mohaned:
Great okay, I believe the changes are irrelevant now. So, I will close this PR.

Thanks @stevenbird  @tomaarsen for the useful insights and discussion.

stevenbird:
Thanks @12mohaned

PR #2760, opened 2021-07-26
checkpoint, test setup in CI

iliakur:
Addressing #2723 through sampling.

PR #2728, opened 2021-06-13
Skipped broken vocabulary test

tomaarsen:
Fixes the flaky test mentioned in #2722.
Alternative to #2723.

---

Hello!

### Pull request overview
* *Skipped* a single test in `nltk/test/unit/lm/test_vocabulary.py`.

This is a *temporary* solution, allowing other PR's (e.g. #2722 by @goodmami) tests to consistently pass again. In the future, @iliakur can propose a more permanent solution that properly corresponds to the initial goals of the test.
As an alternative, #2723 can be merged with the same purpose: being a temporary fix. 

See #2723 for information on the bug.

---

cc: @stevenbird @goodmami @iliakur 

- Tom Aarsen

PR #2841, opened 2021-10-04
Skip ReDoS test - performance testing isn't viable with cloud computing

tomaarsen:
Hello!

### Pull request overview
* Skipped ReDoS test introduced in #2816

### The issue
I'll keep this short - the GitHub actions run [here](https://github.com/nltk/nltk/runs/3756143856) from dacda6f50057e1de560483f5e432b5ccb0dd5876 failed due to this test being inconsistent. This PR is very similar to #2728. 

GitHub uses cloud computing for their CI, and it's not meant to be super consistent. If some other task on the physical machine that runs our test happens to require more resources at a given moment, then we simply can't rely on time-based performance results. This is something that @iliakur has also experienced with #2760.

A potential fix would be to somehow use some hack to keep track of function calls of built-in methods, and compare those. That said, I don't believe this is possible (or reasonable to build for just 2 tests).

So, for the time being, this test should be skipped. I really hoped that I had built in enough room for error, but it seems that perhaps there is no such thing as enough room for error.

- Tom Aarsen

### DIFF
--- a/nltk/test/corpus.doctest
+++ b/nltk/test/corpus.doctest
@@ -2192,5 +2192,5 @@ For now, we accept values below 30 (times as long), due to the potential
 for variance. This ensures that the ReDoS has certainly been reduced,
 if not removed.
 
-    >>> exec_times["long"] / exec_times["short"] < 30
+    >>> exec_times["long"] / exec_times["short"] < 30 # doctest: +SKIP
     True


purificant (approved):

stevenbird:
Thanks @tomaarsen, @purificant

iliakur:
Maybe we give up on running these tests in CI but run them locally? That way whoever's developing the library will catch any problems if they arise. We'd have to mark them somehow so that it's easy to skip them locally too. Perhaps a custom pytest mark for such tests?

tomaarsen:
@iliakur That could be a fix. We would need some way of knowing whether the code is currentlyt ran in a CI test or not. If we have that, then we can e.g. make a `pytest.skipif` decorator in `internals.py`, and slap that on wherever function we want.

iliakur:
I'm guessing gitlab CI defines some environment variables. We can check for their presence and use that as the condition to `skipif`

purificant:
We can mark some tests with a custom marker, like `performance` and only run them them when requested explicitly by the test command. We could also run these in parallel with the main test suite and pass CI regardless of output.

iliakur:
I think I'd prefer to run the tests locally by default without having to request them explicitly. We have put some effort into making sure they **do pass** given a reasonable amount of resources. And having a non-blocking CI stage in my experience results in everyone ignoring it, just due to human nature. This is at best noisy, at worst dangerous (the stage might fail for some other reason than CI flakiness, but we don't catch it).

Issue #2829, opened 2021-09-30
emojis are not tokenized very well


fcbond:
Maybe add something like this as a pre or post processing step?

It might make sense to download the emoji list and store it as part of the build, so people do not need to load the emojis module, ...

```
import emoji
from emoji import unicode_codes
import re


EMOJI_UNICODE = unicode_codes.EMOJI_UNICODE['en']
emojis = sorted(EMOJI_UNICODE.values(), key=len, reverse=True)
print (emojis, sep='\n')
emoji_regexp = f"({'|'.join(re.escape(u) for u in emojis)})"
EMOJI_XL = re.compile(rf"\B({emoji_regexp})", flags=re.UNICODE)
EMOJI_XR = re.compile(rf"({emoji_regexp})\B", flags=re.UNICODE)
EMOJI_WL = re.compile(rf"(\w)({emoji_regexp})", flags=re.UNICODE)
EMOJI_WR = re.compile(rf"({emoji_regexp})(\w)", flags=re.UNICODE)
EMOJI_REGEX = re.compile(rf"({emoji_regexp})", flags=re.UNICODE)

def split_emoji(text):
    text = EMOJI_REGEX.sub(r' \1 ', text)
    text = text.replace('  ', ' ')
    return text

test =  "🤔 🙈 me así, se😌 ds 💕👭👙 hello 👩🏾‍🎓 emoji hello 👨‍👩‍👦‍👦 how are 😊 you today🙅🏽🙅🏽"
#test = "They are going to start a direct flight soon😠"

print(test)

print(split_emoji(test))
```

tomaarsen:
I'm in favor of this, as long as it doesn't make the tokenization considerably slower. I don't think it would have to. 
Perhaps a good place for it would be to include it in the TweetTokenizer? The regexps for those are defined [here](https://github.com/nltk/nltk/blob/a31de6c687165afe64a490053c73c8e0b20ff6db/nltk/tokenize/casual.py#L146).

I would also prefer not having additional dependencies for this, but perhaps the dependency allows us to automatically update to newer emoji lists. After all, they're up to v14.0 now, and will continue updating.

tomaarsen:
I believe #2843 has significantly improved Emoji tokenization (for TweetTokenizer), so I'll close this accordingly. Feel free to let it be known if there is still some issue with Emoji tokenization that was overlooked.

PR #2843, opened 2021-10-05
Enhancement: Add ZWJ sequences Emoji and Skin Tone Modifier Emoji support to TweetTokenizer 

Saibo-creator:
Resolves #2829
Enable TweetTokenizer to handle Emoji containing ZWJ element or Skin Tone Modifier, i.e. tokenize them as a single token instead of several.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -280,6 +280,7 @@
 - Hiroki Teranishi <https://github.com/chantera>
 - Ruben Cartuyvels <https://github.com/rubencart>
 - Dalton Pearson <https://github.com/daltonpearson>
+- Saibo Geng <https://github.com/Saibo-creator>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -334,6 +334,54 @@ class TestTokenize:
         result = tokenizer.tokenize(test2)
         assert result == expected
 
+    def test_emoji_tokenizer(self):
+        """
+        Test a string that contains Emoji ZWJ Sequences and skin tone modifier
+        """
+        tokenizer = TweetTokenizer()
+
+        # A Emoji ZWJ Sequences, they together build as a single emoji, should not be split.
+        test1 = "👨‍👩‍👧‍👧"
+        expected = ["👨‍👩‍👧‍👧"]
+        result = tokenizer.tokenize(test1)
+        assert result == expected
+
+        # A Emoji with skin tone modifier, the two characters build a single emoji, should not be split.
+        test2 = "👨🏿"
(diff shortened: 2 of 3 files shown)


Saibo-creator:
In response to  Issue #2820 , I try to add post-processing to TweetTokenizer so that it will merge ZWJ characters and skin tone modifiers while dealing with Emojis. In terms of performance, I don't think this would slow down the tokenization too much.

tomaarsen:
Hello!

Although your post-processing modifications work, I might prefer a similar implementation that modifies the regex. i.e. adding:
```python
    # Zero-Width-Joiner and Skin tone modifier emojis
    u""".(?:
        [\U0001F3FB-\U0001F3FF]?(?:\u200d.[\U0001F3FB-\U0001F3FF]?)+
        |
        [\U0001F3FB-\U0001F3FF]
    )""",
```
I'll explain. We want two scenario's to work: 
* Characters with a skin tone modifier.
* Characters that are "tied together" with a Zero Width Joiner.

These need to work together. After all, the ZWJ can also tie together skin tone modified emotes.

The core of the regex is essentially this, where `M` matches a skin tone **M**odifier, and `Z` is a **Z**ero-width joiner.
```python
    # Zero-Width-Joiner and Skin tone modifier emojis
    u""".(?:
        M?(?:Z.M?)+
        |
        M
    )""",
```
So either we get a character followed by a skin tone modifier (the second branch), or we get an optional skin tone modifier, followed by a continuous pattern of: a ZWJ, any character, and an optional skin tone modifier.
This means that all characters followed directly with a skin tone modifier will have that modifier attached to the character, which is generally, but not necessarily an emoji. Furthermore, it means that all characters joined together with the ZWJ will also be joined together in the tokenization.

Adding the top 5 lines in this tuple will fix the problem as well:
https://github.com/nltk/nltk/blob/3dcfeb16c8c0cf23e8ab6e3bbc3e2fb63f27f90c/nltk/tokenize/casual.py#L145-L172

Your wonderful tests still pass with the regex fix.

---

I'm personally in favour of the fix with regex, as I'd like to keep the core tokenization mostly regex-based if possible. That said, I'm also biased because I wrote this fix, so I'll leave the decision to others. 
That said, this is a wonderful PR. The code has typing information, is correctly formatted, follows our standards, and introduces some useful tests!

With other words, well done. I approve of this PR regardless of which implementation of the fix is chosen.

- Tom Aarsen

Saibo-creator:
Hey,
        Thanks to your very detailed comment! I reviewed your solution with RegEx, it looks very good to me. I agree that given that Tokenizer's implementation is primarily RegEx-based, keep using RegEx makes the code easier to maintain. (I updated my PR to use your solution but keep the my tests unchanged )

stevenbird:
Great work @Saibo-creator!
Thanks for keeping this in regex land @tomaarsen

PR #2846, opened 2021-10-07
use latest version of pip in CI pipeline

purificant:
Small tweak to CI.

1 Use latest version of pip to install dependencies. There can be some variation depending on which python version is used, this ensures the latest version.
2 Use `--no-cache-dir` during dependency installation, this avoid creating local pip cache which is redundant because it is discarded between runs and because we are caching the entire python environment with all installed dependencies.

This PR should not conflict with https://github.com/nltk/nltk/pull/2820/, however, it's probably a good idea to merge https://github.com/nltk/nltk/pull/2820/ first.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,7 +103,9 @@ jobs:
           key: python-dependencies-${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('requirements-ci.txt') }}-${{ env.pythonLocation }}
 
       - name: Install dependencies on cache miss
-        run: pip install --upgrade --requirement requirements-ci.txt
+        run: |
+          pip install --no-cache-dir --upgrade pip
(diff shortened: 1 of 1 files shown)


tomaarsen:
Out of curiosity, do we know that pip isn't already initialized to the highest version by default when Python is set up?

purificant:
> Out of curiosity, do we know that pip isn't already initialized to the highest version by default when Python is set up?

Yes, I know this to be true from experience. Recently GitHub actions started to show a warning when this happens for py3.10. It's also a common pattern that I use when building containers and CI pipelines but it hasn't been implemented here initially. It's worth adding imo.

tomaarsen (approved):
Good, let's do this then!

purificant:
Rebased the branch to use latest code from develop now that https://github.com/nltk/nltk/pull/2820 was merged.

PR #2847, opened 2021-10-07
Implement get_refs function for DrtLambdaExpression

### LINKED ISSUE #2844: NotImplementedError in lambda DRT simplification.
The following piece of code gives NotImplementedError on line 264 of drt.py (3.6.4)

    import nltk
    expr="\P.\Q.\R.(([X], []) + R(X) + ((P(\Y.([],[ (X=Y) ]))) + (Q(\Z.([], [ (X=Z) ])))))(\X.(\P.(([X],[named(X, karla)]) + P(X) )(\X1.(X(X1)))))(\Y.(\P.\Q. (([X],[ant(X)]) + P(X) + Q(X) )(\Y1.(\X.([],[owner(X) ] )(Y1)))(\Z1.(Y(Z1)))))(\Z.(\P.\X.\F.(P(\E.(([],[nsubj(E,X)]) + F(E))))(\X2.(\F.(([E][arrive(E)]) + F(E) )(\Z2.(X2(Z2)))))(Z)(\Y2.(\F.( [],[] )(Y2)))))"
    dexpr = nltk.sem.drt.DrtExpression.fromstring(r"{}".format(expr))
    dexpr.simplify()
    print(str(dexpr))

When I track the cause, I find that it is happening when  get_refs is called in DrtLambdaExpression. Since it does not exist, the  get_refs function of the abstract class is called (line 264) giving a not implemented error.

RnDevelover:
This update implemets get_refs function for a DrtLambdaExpression. This function is called when a simplify() on DrtLambdaExpression is called. It returns a list of bound variables within the object of type DrtLambdaExpression. The issue caused without this function was reported in #2844 . 
Should fix #2844

### DIFF
--- a/nltk/sem/drt.py
+++ b/nltk/sem/drt.py
@@ -717,6 +717,12 @@ class DrtLambdaExpression(DrtExpression, LambdaExpression):
             + ["    " + blank + line for line in term_lines[3:]]
         )
 
+    def get_refs(self, recursive=False):
+        """:see: AbstractExpression.get_refs()"""
+        return (
+                [self.variable] + self.term.get_refs(True) if recursive else [self.variable]
+            )
+
 
 class DrtBinaryExpression(DrtExpression, BinaryExpression):
     def get_refs(self, recursive=False):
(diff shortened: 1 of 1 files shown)


stevenbird:
Thank you @RnDevelover!

PR #2851, opened 2021-10-09
LGTM.com errors

DimitriPapadopoulos:
Fix part of these LGTM.com errors:
https://lgtm.com/projects/g/nltk/nltk/?severity=error

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -238,7 +238,7 @@ def read_str(s, start_position):
             break
 
     # Process it, using eval.  Strings with invalid escape sequences
-    # might raise ValueEerror.
+    # might raise ValueError.
     try:
         return eval(s[start_position : match.end()]), match.end()
     except ValueError as e:
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -242,7 +242,7 @@ def read_str(s, start_position):
     try:
         return eval(s[start_position : match.end()]), match.end()
     except ValueError as e:
-        raise ReadError("invalid string (%s)" % e) from e
+        raise ReadError("valid escape sequence", start_position) from e
 
 
 _READ_INT_RE = re.compile(r"-?\d+")
--- a/nltk/test/internals.doctest
+++ b/nltk/test/internals.doctest
@@ -138,3 +138,23 @@ It works for classic classes, too:
     False
     >>> overridden(D().f)
     True
+
+
+read_str()
+~~~~~~~~~~~~
+    >>> from nltk.internals import read_str
+
+Test valid scenarios
+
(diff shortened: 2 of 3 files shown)


nltk/internals.py
```diff
@@ -242,7 +242,7 @@ def read_str(s, start_position):
     try:
         return eval(s[start_position : match.end()]), match.end()
     except ValueError as e:
-        raise ReadError("invalid string (%s)" % e) from e
+        raise ReadError("valid escape sequence", start_position) from e
```
dannysepler, line 245: should it say "invalid" instead?

DimitriPapadopoulos: The `ReadError` documentation reads:
```
    :param expected: What was expected when an error occurred.
```

purificant (approved):

nltk/internals.py
dannysepler: i see. i guess i am used to errors telling me what's wrong, rather than what is expected. but honestly, i don't even think this code works!! from my own poking around, `eval-ing` a string with an invalid escape sequence does NOT raise a ValueError, it instead raises a DeprecationWarning. This code dates back to 2008, so perhaps _then_ python used raise an error?

There seems to be [active discussion](https://bugs.python.org/issue32912) in the python community about this, and they may raise an error down the road.

all of this is to say, i know you're just trying to improve small bits of code quality and not anything more. So I added a quick unit test to check other errors work (they seem to), and this looks fine now!

dannysepler (approved):

dannysepler:
Thanks, @DimitriPapadopoulos !

nltk/internals.py
stevenbird: I think the error message needs to make it easy for people to fix. Doesn't that mean reporting the invalid string?

DimitriPapadopoulos: Perhaps, but then that's because the `ReadError` class is poorly designed. This could be fixed in a different PR.

purificant: ReadError class is defined further up in the file as:
```python
class ReadError(ValueError):
    """
    Exception raised by read_* functions when they fail.
    :param position: The index in the input string where an error occurred.
    :param expected: What was expected when an error occurred.
    """

    def __init__(self, expected, position):
        ValueError.__init__(self, expected, position)
        self.expected = expected
        self.position = position

    def __str__(self):
        return f"Expected {self.expected} at {self.position}"
```
This PR fixed a bug where constructor was only passed one argument instead of two and the new string displayed to the user in the context of line 245 will be `f"Expected valid escape sequence at {start_position}"`.

PR #2868, opened 2021-10-27
Rewind Wordnet data file after each lookup

### LINKED ISSUE #2867: Rewind Wordnet data files?
Before the recent merge of #2860, _wordnet.get_version()_ would fail whenever a read had already occurred from the _data.adj_ file:

> import nltk
> from nltk.corpus import wordnet as wn
> print(wn.get_version())

3.0

> print(wn.synset_from_pos_and_offset('a',1740))

Synset('able.a.01')

> print(wn.get_version())

None

The cause of the problem is that _synset_from_pos_and_offset_ goes to a given offset in a data file, reads a line, and stays there. This function is called all the time by other functions in the wordnet module. For ex., in wordnet.doctest, _synset_from_pos_and_offset_ is called indirectly over 10,000 times. 

But now that _get_version_ ensures that reading the data.adj starts at byte offset 0, it probably doesn't matter anywhere else, because there may not be another function that needs to read from the beginning of the data file.

It might be safer to add _data_file.seek(0)_ at the end of _synset_from_pos_and_offset_, but I am not sure that it is worth the additional processing. On the other hand, seek(0) is so cheap an operation that the time penalty would hardly be noticeable.

ekaf:
This fixes #2867. Rewinding Wordnet data files after each lookup may not be absolutely necessary, but it is safer.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1423,6 +1423,7 @@ class WordNetCorpusReader(CorpusReader):
             raise WordNetError(
                 f"No WordNet synset found for pos={pos} at offset={offset}."
             )
+        data_file.seek(0)
         return synset
 
     @deprecated("Use public method synset_from_pos_and_offset() instead")


tomaarsen (approved):
Looks great to me. This seems to be the only occurrence of `seek` other than in `get_version`. I'm all in favor of setting the file pointer back at the starting position like this. I'll give others a few days to comment on this in case they have something to add.
Thanks @ekaf!

stevenbird:
Thanks @ekaf

PR #2877, opened 2021-11-04
 Fixed several TreebankWordTokenizer and NLTKWordTokenizer bugs

### LINKED ISSUE #1750: span_tokenize failed when sentence contains double quotation

### LINKED ISSUE #2076: TreebankWordTokenizer span_tokenize still throwing exception for some corner-cases

### LINKED ISSUE #2876: TreebankWordDetokenizer is not always an inverse of TreebankWordTokenizer
For example, for the following script:

```python

import nltk
import nltk.tokenize

tokens = nltk.tokenize.TreebankWordTokenizer().tokenize("I wanna watch something")
print(tokens)
sentence = nltk.tokenize.treebank.TreebankWordDetokenizer().detokenize(tokens)
print(sentence)
```

The result is:
```
['I', 'wan', 'na', 'watch', 'something']
I wannawatch something
```

The detokenized sentence should be: `I wanna watch something`

tomaarsen:
Resolves #1750, resolves #2076, resolves #2876

Hello!

### Pull request overview
* (Bug 1) Correct tokenisation error when using `''` (two single quotes) as starting quotes.
* (Bug 2) Resolved issue with `'wanna'` in MacIntyreContractions for TreebankWordTokenizer and NLTKWordTokenizer. 
* Implement `span_tokenize` in NLTKWordTokenizer, just like in TreebankWordTokenizer.
* Allow TreebankWordDetokenizer to be imported with `from nltk.tokenize import TreebankWordDetokenizer`.

## Bug 1
Relevant issues: #1750 and #2076
### Reproduction
```python
from nltk.tokenize import TreebankWordTokenizer
tokenizer = TreebankWordTokenizer()
text = "''Hello'\""
print(list(tokenizer.span_tokenize(text)))
```
produces
```python
  File "[sic]\nltk\tokenize\util.py", line 290, in align_tokens
    start = sentence.index(token, point)
ValueError: substring not found

  File "[sic]\nltk_1750.py", line 42, in <module>
    print(list(tokenizer.span_tokenize(text)))
  File "[sic]\nltk\tokenize\treebank.py", line 190, in span_tokenize
    yield from align_tokens(tokens, text)
  File "[sic]\nltk\tokenize\util.py", line 292, in align_tokens
    raise ValueError(f'substring "{token}" not found in "{sentence}"') from e
ValueError: substring "''" not found in "''Hello'""
```
### Details & The fix
The true cause here is that the `span_tokenize` method expects tokens with quotes to be exclusively `''`, `"` or ``:
https://github.com/nltk/nltk/blob/ec1d49da924a69dcaaac8fceb64a34fcd3f9f1e0/nltk/tokenize/treebank.py#L183-L186

This is because `span_tokenize` assumes that the tokenized output will always split off these quotes into separate tokens. However, this is not the case currently. For example:
```python
from nltk.tokenize import TreebankWordTokenizer
tokenizer = TreebankWordTokenizer()
text = "''Hello''"
print(tokenizer.tokenize(text))
```
produces
```python
["''Hello", "''"]
```
This is the true underlying bug. This bug exists in both TreebankWordTokenizer and NLTKWordTokenizer!
The cause is the following ENDING_QUOTES rule:
https://github.com/nltk/nltk/blob/ec1d49da924a69dcaaac8fceb64a34fcd3f9f1e0/nltk/tokenize/treebank.py#L90

`(\S)(\'\')` requires a non-whitespace character before the `''`. Only then will it separate the quote and what comes before - and crucially: only then will it place a space on the right-hand side of the `''`. I don't believe this non-whitespace character requirement to matter. We can remove it, and create:
```python
    (re.compile(r"''"), " '' "),
```
Upon making this change, the last mentioned program outputs:
```python
["''", "Hello", "''"]
```
And the issues reported for `span_tokenize` disappear like expected. 

### Undesired consequences of the fix
This PR improves tokenization by separating these quotes in a better way, but it does have an undesired consequence: TreebankWordDetokenizer doesn't work as well as a result. Or rather, TreebankWordDetokenizer works identically, but the output of the tokenizer has changed somewhat. This is a consequence:
```python
from nltk.tokenize.treebank import TreebankWordDetokenizer
detokenizer = TreebankWordDetokenizer()
print(detokenizer.detokenize(["''", "Hello", "''"]))
```
produces
```
" Hello"
```
This is because the Detokenizer doesn't consider `''` as a start of quotation, but only as the end. After all, if we let it detokenize `["one", "''", "two"]`, how would it know whether `one" two` or `one "two` is correct? It defaults to `one" two`, as `"` is generally an ending quote, with "``" as a beginning quote.

### Notes for Bug 1
This PR modifies both NLTKWordTokenizer *and* TreebankWordTokenizer. I recognise that it might be preferable to keep the latter true to the original (as much as possible). If we want to do that, then I can simply revert the changes there. But do note that then the issues for `span_tokenize` will not be solved.
These issues can also be solved in different ways, though. We can also direct users to the (new) NLTKWordTokenizer's `span_tokenize` which this PR adds, but I'm not a great fan of that. 

---

## Bug 2
Relevant issue: #2876
### Reproduction
```python
from nltk.tokenize.treebank import TreebankWordTokenizer, TreebankWordDetokenizer

tokens = TreebankWordTokenizer().tokenize("I wanna watch something")
print(tokens)
sentence = TreebankWordDetokenizer().detokenize(tokens)
print(sentence)
```
produces
```
I wannawatch something
```

### Details & The fix
This issue originates from the last line of this set of CONTRACTION regexes:
https://github.com/nltk/nltk/blob/ec1d49da924a69dcaaac8fceb64a34fcd3f9f1e0/nltk/tokenize/destructive.py#L19-L28

Where the rule requires the phrase to end in a whitespace with `\s` (which is one character big), rather than `\b` (word boundary, 0 characters big). The issue is that the Detokenizer considers the whitespace (e.g. a space) to be a part of the regex rule match. This is then replaced, while it shouldn't be. 

A simple solution is to just use `\b`. However, this will fail for the edge case of `wanna-be`, where `wanna` should not be split.

So, another solution is to still check for `\s`, but don't include it in the match. This can be done with a positive lookahead:
```python
    r"(?i)\b(wan)(?#X)(na)(?=\s)", 
```
After this fix, both `wanna watch` and `wanna-be` work properly. There is a test for both of these cases.

---

## Added `span_tokenize` to NLTKWordTokenizer
As NLTKWordTokenizer is merely NLTK's improved version of TreebankWordTokenizer, it kind of surprised me that NLTKWordTokenizer didn't already provide all methods that TreebankWordTokenizer has. Because both classes are so similar, both methods are identical. 
However, if we choose not to update TreebankWordTokenizer, then `span_tokenize` for TreebankWordTokenizer will be broken, while the one for NLTKWordTokenizer should still work. I took the doctests from TreebankWordTokenizer's `span_tokenize`, and added them in `tokenize.doctest` too.

## Improved accessibility of TreebankWordDetokenizer
Currently, TreebankWordDetokenizer can only be imported with:
```python
from nltk.tokenize.treebank import TreebankWordDetokenizer 
```
I've added TreebankWordDetokenizer to `nltk/tokenize/__init__.py`, allowing users to import it like so:
```python
from nltk.tokenize import TreebankWordDetokenizer 
# or even
from nltk import TreebankWordDetokenizer 
```

## Future changes
Create some common interface for NLTKWordTokenizer and TreebankWordTokenizer. There is a good bit of code reuse, especially the new `span_tokenize`.

- Tom Aarsen

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -73,6 +73,9 @@ Testing treebank's detokenizer
     >>> s = "Well, we couldn't have this predictable, cliche-ridden, \"Touched by an Angel\" (a show creator John Masius worked on) wanna-be if she didn't."
     >>> detokenizer.detokenize(word_tokenize(s))
     'Well, we couldn\'t have this predictable, cliche-ridden, "Touched by an Angel" (a show creator John Masius worked on) wanna-be if she didn\'t.'
+    >>> s = "I wanna watch something"
+    >>> detokenizer.detokenize(word_tokenize(s))
+    'I wanna watch something'
     >>> s = "I cannot cannot work under these conditions!"
     >>> detokenizer.detokenize(word_tokenize(s))
     'I cannot cannot work under these conditions!'
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -43,7 +43,7 @@ Some test strings.
     ['It', "'s", 'more', "'n", 'enough', '.']
     >>> s12 = "''Hello, there!''"
     >>> word_tokenize(s12)
-    ["''", "Hello", ",", "there", "!", "''"]
+    ["''", 'Hello', ',', 'there', '!', "''"]
     >>> s13 = "''What a wonderful quote, this is'' - Someone (12 BC)"
     >>> word_tokenize(s13)
     ["''", 'What', 'a', 'wonderful', 'quote', ',', 'this', 'is', "''", '-', 'Someone', '(', '12', 'BC', ')']
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -3,10 +3,10 @@
 
     >>> from nltk.tokenize import *
 
-Regression Tests: Treebank Tokenizer
+Regression Tests: NLTKWordTokenizer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Some test strings.
+Tokenizing some test strings.
 
     >>> s1 = "On a $50,000 mortgage of 30 years at 8 percent, the monthly payment would be $366.88."
     >>> word_tokenize(s1)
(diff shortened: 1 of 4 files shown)


stevenbird:
@tomaarsen thanks for the analysis... I like your compromise which favours the correctness of tokenisation over detokenisation (which is less often required)

Issue #2836, opened 2021-10-01
Managing dependencies


stevenbird:
Consider managing dependencies more rigorously, cf https://github.com/nltk/nltk/pull/2835

PR #2881, opened 2021-11-08
Clean up root directory

tomaarsen:
Hello!

### Pull request overview
* Attempt to clean up the root directory somewhat.

### The cleanup
I've already deleted `.travis.yml`, `jenkins-job-config.xml` and `jenkins.sh` from the root in this PR, as we no longer use these files. If we ever decide to return to these services (which I doubt greatly), then we can always just look in the git history to find them again.

There are also jenkins and travis specific files elsewhere, specifically under `tools/travis` and `tools/jenkins`, but these files aren't really "in the way" as much. That said, I'm open to also deleting these.

### Other files
There are some other files in the root directory which might be somewhat unnecessary:

* Perhaps we can merge [requirements-test.txt](https://github.com/nltk/nltk/blob/develop/requirements-test.txt) and [requirements-ci.txt](https://github.com/nltk/nltk/blob/develop/requirements-ci.txt). They're both for testing, and are already very similar.
* I'm not sure whether [NOTICE.txt](https://github.com/nltk/nltk/blob/develop/NOTICE.txt) and [INSTALL.txt](https://github.com/nltk/nltk/blob/develop/INSTALL.txt) provide much at this point - I don't know why these are explicit files, instead of being a part of the README or something similar.
* It somewhat bothers me that `pip-req.txt` and `requirements-ci.txt`/`requirements-test.txt` don't have the same naming structure. Perhaps we can change this? (To be honest, I don't remember ever seeing a `pip-req.txt` file before NLTK - I only encounter `requirements.txt`)

---

- Tom Aarsen


iliakur:
+1 for deleting jenkins/travis tooling. As you said, we have version control in case we want to do some time-travel. In the unlikely event that we return to those services, though, I suspect we'll have to start from scratch anyway due to changes having accumulated since those files were written.

In general +1 for deleting anything that's not *absolutely necessary*. The catch is ofc figuring out what *absolutely necessary* is. For things like NOTICE and INSTALL, I'd ask around (slack, mailing list) if anyone needs those for whatever reason. I'm not sure everyone follows github that closely.

As for all the `requirements.txt` files, maybe #2836 is the right place to discuss that? For example, if we switch to Poetry (tbh my personal preference), we likely won't need any `requirements.txt` at all.

stevenbird:
Glad for this cleanup... thanks for thinking of it, and I hope its not too hard to work out which of those other files can go away now!

PR #2878, opened 2021-11-07
Improved Tokenize documentation + added TokenizerI as superclass for TweetTokenizer

tomaarsen:
Hello!

### Pull request overview
* Improved tokenize documentation:
  * Added docstrings for `span_tokenize` and `tokenize` for `NLTKWordTokenizer` and `TreebankWordTokenizer`.
  * Added Python 3.5+ typing to several methods.
* Made `TokenizerI` the superclass for `TweetTokenizer`, which adds the `tokenize_sents` method for free.

The documentation changes speak for themselves, so I'll only briefly discuss the other changes.

## New superclass for `TweetTokenizer`
`TweetTokenizer` is (I believe) the only tokenizer that didn't subclass from `TokenizerI` yet. Making it subclass from `TokenizerI` will automatically implement `tokenize_sents` for us, and allow easier implementation of `tokenize_span` in the future. 
I've added a test to show that this works correctly.

Oh, also, I made a small mistake with git, so there's one unrelated commit which I didn't see until just now. It doesn't affect the PR and It shouldn't matter if we just squash and merge anyways, but it can be removed from the squash commit message.

- Tom Aarsen

PR #2883, opened 2021-11-10
Deprecate 'return_str' parameter in NLTKWordTokenizer and TreebankWordTokenizer

### LINKED ISSUE #2879: Deprecate `return_str` parameter in `NLTKWordTokenizer` and `TreebankWordTokenizer`
Hello!

I'd like to discuss a potential enhancements of `NLTKWordTokenizer` and `TreebankWordTokenizer`. For those unaware, the former is the tokenizer that is most frequently used, and is used in the `word_tokenize` function. It's also based on the latter class: `TreebankWordTokenizer`.

An example usage as can be found in the documentation:

```python
>>> from nltk.tokenize import NLTKWordTokenizer
>>> s = '''Good muffins cost $3.88 (roughly 3,36 euros)\nin New York.  Please buy me\ntwo of them.\nThanks.'''
>>> NLTKWordTokenizer().tokenize(s)
['Good', 'muffins', 'cost', '$', '3.88', '(', 'roughly', '3,36',
'euros', ')', 'in', 'New', 'York.', 'Please', 'buy', 'me', 'two',
'of', 'them.', 'Thanks', '.']
>>> NLTKWordTokenizer().tokenize(s, convert_parentheses=True)
['Good', 'muffins', 'cost', '$', '3.88', '-LRB-', 'roughly', '3,36',
'euros', '-RRB-', 'in', 'New', 'York.', 'Please', 'buy', 'me', 'two',
'of', 'them.', 'Thanks', '.']
>>> NLTKWordTokenizer().tokenize(s, return_str=True)
' Good muffins cost  $ 3.88  ( roughly 3,36 euros ) \nin New York.  Please buy me\ntwo of them.\nThanks  .  '
```

### The enhancement
As you can see from the example, if `return_str` is True, then the `tokenize` method returns a space-separated string. However, the number of spaces is very inconsistent. Perhaps we would be better off stripping spaces on the ends, and replacing all sequences of multiple spaces with just one space.

E.g. 
```python
>>> NLTKWordTokenizer().tokenize(s, return_str=True)
'Good muffins cost $ 3.88 ( roughly 3,36 euros ) \nin New York. Please buy me\ntwo of them.\nThanks .'
``` 
instead of 
```python
>>> NLTKWordTokenizer().tokenize(s, return_str=True)
' Good muffins cost  $ 3.88  ( roughly 3,36 euros ) \nin New York.  Please buy me\ntwo of them.\nThanks  .  '
```

I figured I would create an issue for this to find out if others agree with this idea, before I put in the time to make this change for no reason. So, I'd like to hear your thoughts.

- Tom Aarsen

adamjhawley:
Resolves #2879 

As per the discussion on #2879, I have deprecated the use of the 'return_str' parameter in the `tokenize` method of both `NLTKWordTokenizer` and `TreebankWordTokenizer`.

I have currently done this in a way that raises a `DeprecationWarning` if a non-default value is passed to the parameter.

### DIFF
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -115,7 +115,7 @@ class NLTKWordTokenizer(TokenizerI):
 
     def tokenize(self, text, convert_parentheses=False, return_str=False):
 
-        if return_str is not False:
+        if return_str:
             warnings.warn(
                 "Parameter 'return_str' has been deprecated and should no "
                 "longer be used.",
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -8,6 +8,7 @@
 
 
 import re
+import warnings
 
 from nltk.tokenize.api import TokenizerI
 from nltk.tokenize.util import align_tokens
@@ -113,6 +114,15 @@ class NLTKWordTokenizer(TokenizerI):
     CONTRACTIONS3 = list(map(re.compile, _contractions.CONTRACTIONS3))
 
     def tokenize(self, text, convert_parentheses=False, return_str=False):
+
+        if return_str is not False:
+            warnings.warn(
+                "Parameter 'return_str' has been deprecated and should no "
+                "longer be used.",
+                category=DeprecationWarning,
+                stacklevel=2,
+            )
+
         for regexp, substitution in self.STARTING_QUOTES:
             text = regexp.sub(substitution, text)
 
@@ -147,7 +157,7 @@ class NLTKWordTokenizer(TokenizerI):
         # for regexp in self._contractions.CONTRACTIONS4:
         #     text = regexp.sub(r' \1 \2 \3 ', text)
 
-        return text if return_str else text.split()
+        return text.split()
 
     def span_tokenize(self, text):
         r"""
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -18,6 +18,7 @@ and available at http://www.cis.upenn.edu/~treebank/tokenizer.sed.
 """
 
 import re
+import warnings
 
 from nltk.tokenize.api import TokenizerI
 from nltk.tokenize.destructive import MacIntyreContractions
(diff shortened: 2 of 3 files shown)


nltk/tokenize/destructive.py
```diff
@@ -113,6 +114,15 @@ class NLTKWordTokenizer(TokenizerI):
     CONTRACTIONS3 = list(map(re.compile, _contractions.CONTRACTIONS3))
 
     def tokenize(self, text, convert_parentheses=False, return_str=False):
+
+        if return_str is not False:
+            warnings.warn(
+                "Parameter 'return_str' has been deprecated and should no"
+                "longer be used.",
```
tomaarsen:
```suggestion
                "Parameter 'return_str' has been deprecated and should no "
                "longer be used.",
```
This string will be concatenated with no space between `no` and `longer`.

nltk/tokenize/treebank.py
```diff
@@ -99,6 +100,15 @@ class TreebankWordTokenizer(TokenizerI):
     CONTRACTIONS3 = list(map(re.compile, _contractions.CONTRACTIONS3))
 
     def tokenize(self, text, convert_parentheses=False, return_str=False):
+
+        if return_str is not False:
+            warnings.warn(
+                "Parameter 'return_str' has been deprecated and should no"
+                "longer be used.",
```
tomaarsen:
```suggestion
                "Parameter 'return_str' has been deprecated and should no "
                "longer be used.",
```
idem.

tomaarsen (changes requested):
I have two small comments on your work. Beyond that, looks great!

Eventually we'll want the deprecation of `return_str` to be described in the method documentation, but the PR for that documentation is still open: #2878. That's something for when both this and that PR are merged (or at least one of the two).

adamjhawley:
@tomaarsen thanks for the review! I have amended the commit to contain your suggestions

tomaarsen (approved):
Looks good to me! Thank you.
I'll leave this open for a little bit so others can have a look at it.

stevenbird (approved):

12mohaned:
I have a nit comment, I would suggest refactoring the if's in Treebank and WordTokenizer to be:

```
if return_str:
  ..........
```

instead of

```
if return_str is not False:
  ..........
```

It is more readable and faster. Also, it is more pythonic and follows the PEP8 standard.

nltk/tokenize/destructive.py
```diff
@@ -113,6 +114,15 @@ class NLTKWordTokenizer(TokenizerI):
     CONTRACTIONS3 = list(map(re.compile, _contractions.CONTRACTIONS3))
 
     def tokenize(self, text, convert_parentheses=False, return_str=False):
+
+        if return_str is not False:
```
tomaarsen:
```suggestion
        if return_str:
```
Well spotted @12mohaned! This is also more in line with the old behaviour of `text if return_str else text.split()`.

stevenbird:
Thanks @adamjhawley, @12mohaned, @tomaarsen

PR #2897, opened 2021-11-28
Fix decode error for bllip parser

### LINKED ISSUE #1920: BllipParser: AttributeError: 'str' object has no attribute 'decode' on Python 3.5
When calling next(BllipParser.parse()) on Python 3.5, I receive the following error:

File "/home/andrew/.local/lib/python3.5/site-packages/nltk/parse/bllip.py", line 169, in parse
    _ensure_ascii(sentence)
  File "/home/andrew/.local/lib/python3.5/site-packages/nltk/parse/bllip.py", line 101, in _ensure_ascii
    word.decode('ascii')
AttributeError: 'str' object has no attribute 'decode'

Related to https://github.com/nltk/nltk/issues/507

NLTK v3.2.5
Python v3.5.2
OS: Ubuntu (16.04.3) running on WSL (16299.125)

dannysepler:
Closes #1920 

This was fairly trivial to fix, but this seems like it was broken since py3 was rolled out. Do we know why _more_ people didn't complain about this? Is it possible that this parser isn't being used at the moment?

I also added a very simple unit test, to capture simple breakages in the `BllipParser` for next time

### DIFF
--- a/nltk/parse/bllip.py
+++ b/nltk/parse/bllip.py
@@ -103,7 +103,7 @@ def _ensure_ascii(words):
             word.encode("ascii")
     except UnicodeEncodeError as e:
         raise ValueError(
-            f"Token {i} ({word}) is non-ASCII. BLLIP Parser "
+            f"Token {i} ({word!r}) is non-ASCII. BLLIP Parser "
             "currently doesn't support non-ASCII inputs."
         ) from e
 
--- a/nltk/parse/bllip.py
+++ b/nltk/parse/bllip.py
@@ -100,11 +100,11 @@ except ImportError as ie:
 def _ensure_ascii(words):
     try:
         for i, word in enumerate(words):
-            word.decode("ascii")
-    except UnicodeDecodeError as e:
+            word.encode("ascii")
+    except UnicodeEncodeError as e:
         raise ValueError(
-            "Token %d (%r) is non-ASCII. BLLIP Parser "
-            "currently doesn't support non-ASCII inputs." % (i, word)
+            f"Token {i} ({word}) is non-ASCII. BLLIP Parser "
+            "currently doesn't support non-ASCII inputs."
         ) from e
 
 
@@ -163,7 +163,7 @@ class BllipParser(ParserI):
             self.rrp.load_reranker_model(
                 features_filename=reranker_features,
                 weights_filename=reranker_weights,
-                **reranker_options
+                **reranker_options,
             )
 
     def parse(self, sentence):
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ web/_build
 # Test artifacts and coverage reports
 *.tox
 *.errs
+.hypothesis
 .noseids
 .coverage*
 nltk/test/*.html
--- a/nltk/test/unit/test_bllip.py
+++ b/nltk/test/unit/test_bllip.py
@@ -11,6 +11,10 @@ def parser():
     return BllipParser.from_unified_model_dir(model_dir)
 
 
+def setup_module():
+    pytest.importorskip("bllipparser")
+
+
 class TestBllipParser:
     def test_parser_loads_a_valid_tree(self, parser):
         parsed = parser.parse("I saw the man with the telescope")
--- a/nltk/test/unit/test_bllip.py
+++ b/nltk/test/unit/test_bllip.py
@@ -1,0 +1,38 @@
+import pytest
+
+from nltk.data import find
+from nltk.parse.bllip import BllipParser
+from nltk.tree import Tree
+
+
+@pytest.fixture(scope="module")
+def parser():
+    model_dir = find("models/bllip_wsj_no_aux").path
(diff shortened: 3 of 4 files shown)


dannysepler:
It looks like the `bllip-parser` has been broken on Windows for quite some time now https://github.com/BLLIP/bllip-parser/issues/48

I'm not sure what to do here. We could just not have these new tests run on CI, or perhaps we should only install the pip package on non-Windows environments?

dannysepler:
okay looking for a bit of advice here!

nltk's `BllipParser` has likely been broken for years. as a result, we can reasonably assume it is not being used

The parser also isn't unit tested. adding tests to the CI build is tricky, since the `bllipparser` pip package itself has been broken on windows since 2016. i see a few paths forward, and am unsure which to take:

- fix the `BllipParser`, don't add unit tests to the CI
- fix the `BllipParser`, add unit tests to the CI, but have our CI build only download the `bllipparser` pip package on mac and linux
  - this will add some complexity to the CI build for a modest gain in test coverage
- deprecate / eventually delete the `BllipParser` since we don't think it's being used

tomaarsen:
This is a bit of a tricky one. I'm not familiar with `BllipParser`, which means I don't know whether it has functionality that isn't already possible through some other parser. That would affect whether we should care to fix the `BllipParser`.
However, if the fix is simple, then I don't see why we shouldn't *at least* push the fix.

Regarding CI, we already have some third party tools which are downloaded for Mac and Linux only, but adding a package which must be installed via `pip` to that might be a hastle. I already believe the CI is a bit vague and complex, and I don't really want to add to that even more.

Perhaps all of this means that I'm leaning towards option 1.

An interesting mix of everything would be to add the tests as quick unit tests to method/class bodies of `nltk/parse/bllip.py` itself. These doctests are not ran automatically with the CI, but interested users can `pytest --doctest-modules nltk/parse/bllip.py`. That way, your work of coming up with the tests isn't wasted either. (With `# doctest: +NORMALIZE_WHITESPACE` to get the tests passing etc.)

That's my view, others might disagree.

stevenbird:
@dmcc do you have any advice for us please? Is there any point in us maintaining NLTK's BLLIP parser?

tomaarsen:
@dannysepler 
That said, this PR still has broken tests from the PR itself.

dmcc:
Yikes, sorry this was broken for so long. BLLIP Parser does provide some functionality not available in the other NLTK parsers, so there's an argument for keeping it running. It's a bit surprising that nobody noticed/reported it was broken for so long, that's a concern. That said, I have recently received bug reports for the [lower-level BLLIP Python interface](https://github.com/BLLIP/bllip-parser), so it appears some people are still using it.

Option 1 sounds good to me -- I don't have easy access to Windows or Mac OS X machines to debug the issues, I'm afraid.

stevenbird:
Thanks @dmcc

tomaarsen (changes requested):
@dannysepler So, I believe the consensus is to apply the fix, but don't add the test to CI. 

I believe all that needs to happen at this point is to remove `bliipparser` from requirements-ci.txt. Doing so will cause `nltk/test/unit/test_bllip.py` to be skipped in the CI, but users who have bllipparser installed will still be able to run these tests as a part of `tox` or `pytest nltk/test`.

Beyond that, I have confidence that your changes indeed work, as the tests for macOS and Ubuntu passed.

dannysepler:
sorry for the delay @tomaarsen, and thank you for the guidance!

thank you as well to @dmcc for confirming the parser's usefulness!

nltk/parse/bllip.py
```diff
@@ -100,11 +100,11 @@ def _ensure_bllip_import_or_error(ie=ie):
...
     try:
         for i, word in enumerate(words):
-            word.decode("ascii")
-    except UnicodeDecodeError as e:
+            word.encode("ascii")
+    except UnicodeEncodeError as e:
         raise ValueError(
-            "Token %d (%r) is non-ASCII. BLLIP Parser "
-            "currently doesn't support non-ASCII inputs." % (i, word)
+            f"Token {i} ({word}) is non-ASCII. BLLIP Parser "
```
tomaarsen:
```suggestion
            f"Token {i} ({word!r}) is non-ASCII. BLLIP Parser "
```
I believe for it to be equivalent, you would want to get the `repr` of `word`. I haven't tested this for certain, but if I remember correctly, you can just do this with `{word!r}`. (For the record, I believe it's also fine to just print `word`)

tomaarsen (approved):
Looks good to me. I just have the one small nitpick, which isn't important at all.

nltk/parse/bllip.py
dannysepler: good catch! i agree -- while it probably doesn't matter, using the repr (`!r`) is slightly more in the spirit of this error message since it's helpful for debugging

tomaarsen:
Thanks Danny! Well done.

PR #2898, opened 2021-11-28
deprecate: deprecate get and set node methods

12mohaned:
Deprecate **_get_node()** and **_set_node()** methods.


stevenbird:
@12mohaned this is not deprecating code, but removing code. Is there some reason for this?

stevenbird:
@12mohaned I don't see any case for this, sorry.

PR #2900, opened 2021-12-02
deprecate: deprecate get and set node methods

12mohaned:
This pull request deprecates _get_node() and _set_node() methods, it was similar to this [PR](https://github.com/nltk/nltk/pull/2898), but I forgot to amend the required changes instead, I accidentally removed the methods. Sorry for the inconvenience @stevenbird .

### DIFF
--- a/nltk/tree/tree.py
+++ b/nltk/tree/tree.py
@@ -17,8 +17,7 @@ syntax trees and morphological trees.
 
 import re
 import sys
-import warnings
-
+from nltk.internals import deprecated
 from nltk.grammar import Nonterminal, Production
 
 ######################################################################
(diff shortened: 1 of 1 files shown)


tomaarsen:
NLTK has implemented a useful decorator for cases like these. I recommend that we use it. See an example below:
https://github.com/nltk/nltk/blob/edf47c008c2fb52c13ec40b1f41312e664d9539b/nltk/metrics/agreement.py#L164-L178

i.e. simply adding `@deprecated("Use set_label() instead")` before both of the methods. The body can become `pass` or kept how it was. See the code for `deprecated` [here](https://github.com/nltk/nltk/blob/develop/nltk/internals.py#L408).

tomaarsen (commented):
This is definitely preferable, however the `_get_node() is deprecated` section of the deprecation message is unnecessary, as `@deprecated` already adds this. Right now, the output is:
```python
>>> from nltk.tree import Tree
>>> t = Tree.fromstring('(S (NP (D the) (N dog)) (VP (V chased) (NP (D the) (N cat))))')
>>> t._get_node()  
<stdin>:1: DeprecationWarning: 
  Function _get_node() has been deprecated.  _get_node() is
  deprecated, use label() instead
```

tomaarsen (approved):
Looks good to me

tomaarsen:
@12mohaned 

PR #2901, opened 2021-12-05
return demo_sent_subjectivity instead of print

andrewlin0:
I simply changed the output format of the demo_sent_subjectivity to return instead of print to avoid the program to have to run longer because it prints out each. result to the console.

### DIFF
--- a/nltk/sentiment/util.py
+++ b/nltk/sentiment/util.py
@@ -713,7 +713,8 @@ def demo_sent_subjectivity(text):
 
     # Tokenize and convert to lower case
     tokenized_text = [word.lower() for word in word_tokenizer.tokenize(text)]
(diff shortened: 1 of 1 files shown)


tomaarsen:
The reason the `demo_*` functions don't tend to return anything, is because they are not meant to be used much. In my opinion, these functions should not even be importable from NLTK. They are only meant to quickly and effectively show how functions/classes in that file can be used.
Because of that, I prefer printing over returning, as it helps with effectively displaying the output.

tomaarsen:
That said, this PR still has broken tests from the PR itself. (`pre-commit` fails because of incorrect formatting, running `black` or `pre-commit run --all-files` should solve it)

tomaarsen:
I'll be closing this.

Issue #2905, opened 2021-12-07
`cmn` wordnet folder name and data `.tab` name doesn't match


badGarnet:
`nltk==3.6.5`
Tested on `python==3.8.12`
Chinese language wordnet data is now stored in `corpora/omw/cow/wn-data-cmn.tab`
But code assumes folder name and wordnet data suffix is the same: https://github.com/nltk/nltk/blob/f50b6b1023689bb49c0f0a76f907e035f4360b12/nltk/corpus/reader/wordnet.py#L1253-L1254

Now loading `cmn` data raises error, e.g., 

```
❯ NLTK_DATA=/tmp ipython
Python 3.8.12 (default, Oct  5 2021, 17:04:41)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from nltk.corpus import wordnet as wn

In [2]: wn._load_lang_data("cmn")
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-718659b9166c> in <module>
----> 1 wn._load_lang_data("cmn")

~/.pyenv/versions/3.8.12/envs/pcore-dec7/lib/python3.8/site-packages/nltk/corpus/reader/wordnet.py in _load_lang_data(self, lang)
   1195             raise WordNetError("Language is not supported.")
   1196
-> 1197         with self._omw_reader.open("{0:}/wn-data-{0:}.tab".format(lang)) as fp:
   1198             self.custom_lemmas(fp, lang)
   1199

~/.pyenv/versions/3.8.12/envs/pcore-dec7/lib/python3.8/site-packages/nltk/corpus/reader/api.py in open(self, file)
    228         """
    229         encoding = self.encoding(file)
--> 230         stream = self._root.join(file).open(encoding)
    231         return stream
    232

~/.pyenv/versions/3.8.12/envs/pcore-dec7/lib/python3.8/site-packages/nltk/data.py in join(self, fileid)
    332     def join(self, fileid):
    333         _path = os.path.join(self._path, fileid)
--> 334         return FileSystemPathPointer(_path)
    335
    336     def __repr__(self):

~/.pyenv/versions/3.8.12/envs/pcore-dec7/lib/python3.8/site-packages/nltk/compat.py in _decorator(*args, **kwargs)
     39     def _decorator(*args, **kwargs):
     40         args = (args[0], add_py3_data(args[1])) + args[2:]
[...]

PR #2907, opened 2021-12-08
Renamed omw to omw-1.4

ekaf:
This PR selects omw-1.4 as the default package to use together with Wordnet. This implements the third point in @tomaarsen's plan from https://github.com/nltk/nltk/issues/2905#issuecomment-988901617, in order to ensure a smooth transition to the new OMW-1.4 data.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,7 @@ jobs:
         id: restore-cache
         with:
           path: ~/nltk_data
-          key: try-nltk_data_${{ secrets.CACHE_VERSION }}
+          key: nltk_data_${{ secrets.CACHE_VERSION }}
 
       - name: Download nltk data packages on cache miss
         run: |
@@ -112,7 +112,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/nltk_data
-          key: try-nltk_data_${{ secrets.CACHE_VERSION }}
+          key: nltk_data_${{ secrets.CACHE_VERSION }}
 
       - name: Use cached third party tools
         uses: actions/cache@v2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,7 @@ jobs:
         id: restore-cache
         with:
           path: ~/nltk_data
-          key: nltk_data_${{ secrets.CACHE_VERSION }}
+          key: try-nltk_data_${{ secrets.CACHE_VERSION }}
 
       - name: Download nltk data packages on cache miss
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -112,7 +112,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/nltk_data
-          key: nltk_data_${{ secrets.CACHE_VERSION }}
+          key: try-nltk_data_${{ secrets.CACHE_VERSION }}
 
       - name: Use cached third party tools
         uses: actions/cache@v2
--- a/nltk/corpus/__init__.py
+++ b/nltk/corpus/__init__.py
@@ -356,7 +356,6 @@ verbnet = LazyCorpusLoader("verbnet", VerbnetCorpusReader, r"(?!\.).*\.xml")
 webtext = LazyCorpusLoader(
     "webtext", PlaintextCorpusReader, r"(?!README|\.).*\.txt", encoding="ISO-8859-2"
 )
-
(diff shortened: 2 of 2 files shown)


ekaf:
This PR assumes https://github.com/nltk/nltk_data/pull/175. Otherwise, the tests fail because OMW-1.4 cannot be downloaded.

tomaarsen (approved):
Looks good to me.

We should merge https://github.com/nltk/nltk_data/pull/175 quickly, and then we can restart the tests for this PR, and get this merged too.

Thanks for jumping on this so quickly @ekaf

ekaf:
I compared the output of:

```
for lg in sorted(wn.langs()):
    print(f"{lg}: {len(list(wn.words(lang=lg)))} words in {len(list(wn.all_synsets(lang=lg)))} synsets")
```
using respectively the old nltk-3.5 (with old omw.zip) vs. nltk-3.6.5 and this PR (with new omw-1.4.zip):


```
NLTK v. 3.5, Wordnet v. 3.0

	NLTK v. 3.6.5, Wordnet v. 3.0
	
cmn: 61532 words in 42300 synsets
	cmn: 61533 words in 42300 synsets
fas: 17560 words in 17759 synsets
fra: 55350 words in 59091 synsets
	fra: 55351 words in 59091 synsets
	hrv: 29008 words in 23115 synsets
hrv: 29010 words in 23115 synsets
	isl: 11504 words in 4951 synsets
	ita_iwn: 19221 words in 15563 synsets
jpn: 89637 words in 57184 synsets
	jpn: 91964 words in 57184 synsets
	lit: 11395 words in 9462 synsets
por: 54069 words in 43895 synsets
	por: 54071 words in 43895 synsets
qcn: 3206 words in 4913 synsets
	ron: 49987 words in 56026 synsets
	slk: 29150 words in 18507 synsets
	slv: 40230 words in 42583 synsets
slv: 41032 words in 42717 synsets
tha: 80508 words in 73350 synsets
	tha: 82504 words in 73350 synsets
```
This shows that less than half the languages had a (usually small) increase in the number of lemmas, while the number of synsets stayed the same between the two OMW versions.

tomaarsen:
This is going to have cache issues again. 
Luckily we can still just test this locally instead, and GitHub allows merging despite failing tests. That said, I'd rather fix the cache issues all-together, but that's an issue for another time.

I've tried only caching if the commit message doesn't contain `\nocache`, but that won't *refresh* the cache, it only ignores it for that one commit. I would rather have the opportunity to completely refresh it, but I haven't yet found a consistent way to do that.

tomaarsen:
I suspect the secrets trick only works on branches within `nltk/nltk`. Perhaps @ekaf you could go to https://github.com/ekaf/nltk > Settings > Secrets > Update "CACHE_VERSION" and fill in some arbitrary input. Once you've done that, you can @ me and I'll restart the test (or you can remove the blank line again in a commit, both work).

ekaf:
@tomaarsen: near the top of that page, it says "Secrets are not passed to workflows that are triggered by a pull request from a fork". So I suspect there will be a better chance if I renew the key in ci.yaml. We can always restore it afterwards.

tomaarsen:
@ekaf sounds good. I suppose that makes sense - if PR's were passed secrets, then someone could run a github actions workflow in a PR that just prints the secret in plaintext, which sounds like a giant vulnerability when secrets are used as authentication etc.

tomaarsen:
@ekaf `omw-1.4` is not mentioned in [`index.xml`](https://github.com/nltk/nltk_data/blob/gh-pages/index.xml), so it can't be downloaded. It's also not included in the `all` collection. I believe that's why there are issues.

ekaf:
@tomaarsen, yes, that sounds likely to be the reason. The same applies to wordnet2021, then. I am to going add .xml files for them, run "make pkg_index" and submit a PR in nltk_data.

tomaarsen:
> @tomaarsen, yes, that sounds likely to be the reason. The same applies to wordnet2021, then. I am to going add .xml files for them, run "make pkg_index" and submit a PR in nltk_data.

@ekaf I'm already on it. I'll tag you in a PR in a few minutes, and you can review it if possible.

tomaarsen:
The Ubuntu cache was updated, but the Windows one was not, hence the fail here. 
That said, I happen to be able to test Windows locally, and everything passes there (while `develop` should no longer pass, due to the recent `omw` changes in `nltk_data`).

With other words, I have no issues merging this as it stands. On `develop` I can influence the `CACHE_VERSION` secret to ensure a new cache. 

Thank you for your work in getting us to OMW 1.4!

PR #2908, opened 2021-12-10
Added warnings if nltk_data `.zip` files exist without any corresponding `.xml` files.

tomaarsen:
Resolves https://github.com/nltk/nltk_data/issues/178

Hello!

## Pull request overview
* Added warnings if `build_index` is ran in `nltk.downloader`, when the root folder contains `.zip` files without any corresponding `.xml` files.

## The issue
As inspired mainly by https://github.com/nltk/nltk_data/issues/178, as well as by https://github.com/nltk/nltk_data/pull/176 and https://github.com/nltk/nltk_data/pull/177, @ekaf has noticed that there are several resources in `nltk_data` that have `.zip` files with data, but no `.xml` to specify what kind of data it is. When there is no `.xml`, these `.zip` files are quietly being ignored. 

We experienced this issue with `omw-1.4.zip` (https://github.com/nltk/nltk_data/pull/176) and `wordnet2021.zip` (https://github.com/nltk/nltk_data/pull/177) recently, and it seems there are more resources with these issues.

This PR adds a warning in these situations. 
Running `make pkg_index` on `nltk_data` now also outputs the following:
```python
[sic]\nltk\downloader.py:2294: UserWarning: listing.csv.zip exists, but listing.csv.xml cannot be found! This could mean that listing.csv can not be downloaded.
  for pkg_xml, zf, subdir in _find_packages(os.path.join(root, "packages")):
[sic]\nltk\downloader.py:2294: UserWarning: omw-1.4.zip exists, but omw-1.4.xml cannot be found! This could mean that omw-1.4 can not be downloaded.
  for pkg_xml, zf, subdir in _find_packages(os.path.join(root, "packages")):
[sic]\nltk\downloader.py:2294: UserWarning: ptb3.zip exists, but ptb3.xml cannot be found! This could mean that ptb3 can not be downloaded.
  for pkg_xml, zf, subdir in _find_packages(os.path.join(root, "packages")):
[sic]\nltk\downloader.py:2294: UserWarning: wordnet2021.zip exists, but wordnet2021.xml cannot be found! This could mean that wordnet2021 can not be downloaded.
  for pkg_xml, zf, subdir in _find_packages(os.path.join(root, "packages")):
```

This should prevent us from accidentally forgetting to add an `.xml` to a new `nltk_data` resource.

## Notes
[...]

PR #2909, opened 2021-12-10
Allow commits with `/nocache` to not use the cache

tomaarsen:
Hello!

## Pull request overview
* If a commit message contains `/nocache`, then it will not use the existing cache. 

## Details
From now on, a commit message containing `/nocache` will cause a CI run *without* the existing cache. Note that this *does not* refresh the cache, it just ignores the cache for that commit only. On the `develop` branch, it is recommended (for admins) to go to `Secrets > CACHE_VERSION` and change the value, after which you can go to Actions and "Restart this workflow". No new commit required here.

However, this does not work on PRs. That's what this PR is for. If any PR in the future is failing because of cache issues (which would only be if `nltk_data` was updated very recently), then an empty commit can be pushed with `git commit --allow-empty -m "Restarting CI without cache: /nocache"`. This should cause a CI run without a cache.

- Tom Aarsen

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,12 +37,12 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Cache nltk data
+        if: ${{ !contains(github.event.head_commit.message, '/nocache') }}
         uses: actions/cache@v2
         id: restore-cache
         with:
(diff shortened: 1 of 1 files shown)


tomaarsen:
I believe this works sufficiently. I'm going to try it out on #2908.

PR #2910, opened 2021-12-12
Fix `FileNotFoundError` when the `download_dir` is a non-existing nested folder

### LINKED ISSUE #2904: Downloader fails when download_dir is a nested folder that doesn't exist
### Problem
When the download folder doesn't exist, it is created by the downloader. But when that folder is nested, the creation fails.

### Code example

```python
nltk.download("punkt", force=True, download_dir="some/nested/folder")
```

Will give the following error:

```
File "my_env/lib/python3.7/site-packages/nltk/downloader.py", line 776, in download
    for msg in self.incr_download(info_or_id, download_dir, force):
  File "my_env/lib/python3.7/site-packages/nltk/downloader.py", line 641, in incr_download
    yield from self._download_package(info, download_dir, force)
  File "my_env/lib/python3.7/site-packages/nltk/downloader.py", line 698, in _download_package
    os.mkdir(download_dir)
FileNotFoundError: [Errno 2] No such file or directory: 'some/nested/folder'
```

### Suggested solution

The following line throws the error: https://github.com/nltk/nltk/blob/develop/nltk/downloader.py#L698

`os.makedirs(download_dir)` could be used instead of `os.mkdir(download_dir)`

sp1thas:
The purpose of this PR is to resolve #2904 by replacing the `os.mkdir` with `os.makedirs` in order to prevent `FileNotFoundError` when the `download_dir` is a non-existing nested directory (parent direcotry does not exists)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -286,6 +286,7 @@
 - Ahmet Yildirim <https://github.com/RnDevelover>
 - Yuta Nakamura <https://github.com/yutanakamura-tky>
 - Adam Hawley <https://github.com/adamjhawley>
+- Panagiotis Simakis <https://github.com/sp1thas>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -695,9 +695,9 @@ class Downloader:
 
         # Ensure the download_dir exists
         if not os.path.exists(download_dir):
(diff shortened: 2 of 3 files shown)


iliakur:
I wonder how easy it would be to add tests for this... pytest has some fixtures to create temporary directories.

tomaarsen:
Perfect! Thanks a bunch @sp1thas. Great tests, too.

Issue #2913, opened 2021-12-13
Add CI tests for Python 3.10


tomaarsen:
Hello!

Python 3.10 is no longer a glimmer far into the future, and we should start supporting it. The first step in doing so is adding 3.10 to the CI tests, and mentioning that NLTK works for Python 3.10 in our documentation throughout the project.

- Tom Aarsen

purificant:
May need to look into refactoring gensim codebase, we are currently locked to version <4

tomaarsen:
@purificant Gensim is exclusively used in this file: https://github.com/nltk/nltk/blob/develop/nltk/test/gensim.doctest
So, we can also remove the dependency if we want, as no code in NLTK relies on it. We could then keep the dependency in CI locked to < 4, and in the future we can refactor the doctest to allow for gensim > 4. Another option is to deprecate fully, and remove gensim.doctest too.

purificant:
gensim <4 appears incompatible with py310, at least in my recent attempts to pass the CI here: https://github.com/purificant/nltk/runs/4467436202 it is failing to build / install.

Because of the above, it seems the only two options are to refactor and relax <4 version lock or deprecate the test entirely?

tomaarsen:
I agree that those seem like the only options. Judging by the gensim.doctest, the functionality seems quite useful. Perhaps we can attempt a refactor of gensim.doctest? 
Another option is to remove the gensim dependencies, even from the CI, which should let the tests pass automatically due to https://github.com/nltk/nltk/blob/develop/nltk/test/gensim_fixt.py.

PR #2914, opened 2021-12-14
Refactor `gensim.doctest` to work for gensim 4.0.0 and up

tomaarsen:
Hello!

## Pull Request Overview
* Update `gensim.doctest` allowing us to remove the `<4.0.0` version lock of gensim.

## Details
See https://github.com/nltk/nltk/issues/2913#issuecomment-993305846 for a short discussion on gensim. It is currently locked to gensim < 4.0.0, and we'll want to start supporting newer versions. However, in 4.0.0, gensim made some changes in syntax, meaning that you either support before 4.0.0, or after 4.0.0.
This is us moving to the latter.

## Beyond this PR
Lastly, I would like to propose the following, which is not included in this PR: 
Let's deprecate gensim from `pip-req.txt`. gensim is not used anywhere in the codebase, except for this doctest, so we should remove it from `pip-req.txt`, and keep it only in `requirements-ci.txt`/`requirements-test.txt`. I have some other issues with these files at the moment, which I've brought up before in #2881:

* Perhaps we can merge [requirements-test.txt](https://github.com/nltk/nltk/blob/develop/requirements-test.txt) and [requirements-ci.txt](https://github.com/nltk/nltk/blob/develop/requirements-ci.txt). They're both for testing, and are already very similar.
* It somewhat bothers me that `pip-req.txt` and `requirements-ci.txt`/`requirements-test.txt` don't have the same naming structure. Perhaps we can change this? (To be honest, I don't remember ever seeing a `pip-req.txt` file before NLTK - I only encounter `requirements.txt`)

---

cc: @purificant This PR might be relevant for you and your work to help with #2913.

- Tom Aarsen


purificant (approved):

requirements-ci.txt
```diff
@@ -1,4 +1,4 @@
-gensim<4.0.0
+gensim>=4.0.0
```
purificant, line 1:
```suggestion
gensim
```

purificant:
`requirements-ci.txt` was created as a clean slate for CI when moving to GitHub Actions, since then this has become our primary CI pipeline, we could definitely clean up other requirement files and consolidate. :+1:

purificant:
Amazing work @tomaarsen, thank you for doing this. One step closer to py 3.10

requirements-ci.txt
tomaarsen: `gensim.doctest` will require 4.0.0 or higher, but I suppose that `requirements-ci.txt` will take the highest version regardless. I'm not sure whether it's preferable to have `gensim>=4.0.0`, where all valid versions will pass the tests, or just `gensim` which is simpler, but the tests would fail for `gensim<4.0.0`.

Also, what are your thoughts on removing `gensim` from `pip-req.txt`?

purificant: I don't mind either way, just a thought.

As you said, CI will use the latest version and invalidate the cache when this file changes. So we'll always have a working version until the next breaking change.

Onboard with trimming down `pip-req.txt` to remove things that we don't directly need for features. We have a CI pipeline to cover the doctest so unless someone is working on that specific file, they won't need the dependency in local dev environment, which is what we use the `pip-req.txt` file for according to [docs](https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md#setting-up-a-development-environment).

tomaarsen: I'll keep it as `gensim>=4.0.0` solely as an indication that the tests in the CI only pass for those versions. Once we have this PR merged or closed, I'll move to trimming `pip-req.txt`, renaming it to `requirements.txt`, and deleting either `requirements-ci.txt` and `requirements-test.txt`.

dannysepler (approved):
+1, great work!

tomaarsen:
I've decided it's best to remove the unused gensim dependency as it might mess with people who wish to use specific gensim versions alongside NLTK, and there is no need to potentially introduce unnecessary version conflict.

PR #2924, opened 2021-12-22
Add extended open multilingual wordnet reader

ExplorerFreda:
Related issue: https://github.com/nltk/nltk/issues/1611 
Related NLTK data issue: https://github.com/nltk/nltk_data/issues/179
Extended open multilingual wordnet: http://compling.hss.ntu.edu.sg/omw/summx.html


tomaarsen:
Hey! We use [`pre-commit`](https://pre-commit.com/) hooks to ensure code quality through our [`pre-commit config`](https://github.com/nltk/nltk/blob/develop/.pre-commit-config.yaml), which runs `pyupgrade`, `black`, `isort` and some other hooks. Using it is as simple as `pip install pre-commit` and then running `pre-commit install`. From then onwards, any time you commit to `https://github.com/ExplorerFreda/nltk`, it will run these hooks before committing. 
The first time you might not have any changes to turn into a commit, so you can do `pre-commit run --all` or try to create an empty commit, and it will format your changes. You can then push this to make the CI stop complaining.

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,141 @@
...
+For details about Open Multilingual WordNet, see:
+http://compling.hss.ntu.edu.sg/omw/
+
+For details about Extended Open Multilingual WordNet, see:
+http://compling.hss.ntu.edu.sg/omw/summx.html
+"""
+
+import os
+from collections import defaultdict
+from IPython.core.pylabtools import configure_inline_support
```
tomaarsen: This seems like an unused import. Should this be there?

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,141 @@
...
+
+from traitlets.traitlets import default
```
tomaarsen: This seems like an unused import. Should this be there?

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,141 @@
+# Natural Language Toolkit: Extened Open Multilingual WordNet Reader
```
tomaarsen:
```suggestion
# Natural Language Toolkit: Extended Open Multilingual WordNet Reader
```

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _load_lang_data(self, lang):
+        """Load the wordnet data to self._lang_data"""
+        if lang in self._lang_data.keys():
+            return
+        # load open multilingual wordnet
+        for prov in self.provenances[lang]:
+            with self._omw_reader.open(f"{prov}/wn-data-{lang}.tab") as fp:
+                self.custom_lemmas(fp, lang)
+                fp.close()
```
tomaarsen: The `with` keyword already automatically closes any files whenever the `with` scope is exited, so this line is unnecessary.

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,141 @@
...
+        # load extended open multilingual wordnet 
+        for prov in self.exomw_provenances[lang]:
+            with self._exomw_reader.open(f"{prov}/wn-{prov}-{lang}.tab") as fp:
+                self.custom_lemmas(fp, lang)
+                fp.close()
```
tomaarsen: Idem.

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,141 @@
...
+                if len(pair) == 1 or pair[0] == lang:
+                    if attr == "lemma":
+                        val = val.strip().replace(" ", "_").lower()
+                        self._lang_data[lang][1][val].add(offset_pos)
+                    if attr in self.lg_attrs:
+                        index = self.lg_attrs.index(attr)
+                        self._lang_data[lang][index][offset_pos].add(val)
+
+    def disable_custom_lemmas(self, lang):
+        """prevent synsets from mistakenly added"""
```
tomaarsen: Is this missing "being"?

tomaarsen (changes requested):
I recognise that this is a bit of an annoying situation, because any (automated) tests will fail due to `extended_omw` not yet existing in `nltk_data`, but perhaps you can provide a bit of sample code that should work if we (manually) download `extended_omw` and place it in our `nltk_data` folder?

Beyond that I added some small nitpicks. I can't really comment on the "real functionality", because I haven't been able to run it.

ExplorerFreda:
> I recognise that this is a bit of an annoying situation, because any (automated) tests will fail due to `extended_omw` not yet existing in `nltk_data`, but perhaps you can provide a bit of sample code that should work if we (manually) download `extended_omw` and place it in our `nltk_data` folder?
> 
> Beyond that I added some small nitpicks. I can't really comment on the "real functionality", because I haven't been able to run it.

Hi @tomaarsen, thanks for the review and suggestions! I've changed the code correspondingly. 

Below is the script for downloading the required data that should work: 
```bash
cd $NLTK_DATA_DIR/corpora/
wget http://compling.hss.ntu.edu.sg/omw/wn-wikt.zip -O extended_omw.zip
unzip extended_omw.zip
mv data/ extended_omw/
```

nltk/corpus/reader/extended_omw.py
```diff
@@ -0,0 +1,137 @@
+# Natural Language Toolkit: Exteneded Open Multilingual WordNet Reader
```
tomaarsen:
```suggestion
# Natural Language Toolkit: Extended Open Multilingual WordNet Reader
```
I haven't had the time to take a full look at this PR yet, but I noticed this small typo.

ekaf:
It would be nice to add the following capabilities:

1. Loading the English wiktionary wordnet. This requires small modifications in the wordnet module because it does not anticipate English wordnets from alternative provenances at this moment.

2. Loading different wordnets for the same language separetely would allow to compare the quality of different provenances. The reason why the 'wikt' and 'cldr' provenances did not go into the official OMW 1.4 package was that their quality was deemed uncertain. For each language, this PR uses the set union of all provenances, which may be appropriate for some uses, but it would be nice to also support different use cases.

3. Combining extended_wordnet with newer wordnet packages like wordnet31 and wordnet2021. This would be easier if the extended_omw corpus reader was a leaner class, that could be added to any English Wordnet, including the forthcoming ones like Open English Wordnet 2022.

4. For completeness, maybe also consider adding back the Persian wordnet and the alternative Chinese wordnet from the old OMW package (in the 'pas' and 'qcn' directories), which were left out of OMW 1.4 for quality reasons though it is possible that some users miss them.

stevenbird:
Resolved in #2946

stevenbird:
@ExplorerFreda thanks for your work, which is appreciated even though we ended up accepting a different PR.

PR #2936, opened 2022-01-25
Fix LC cutoff policy of text tiling

richarddwang:
According to the paper (Hearst, Marti A.. “Text Tiling: Segmenting Text into Multi-paragraph Subtopic Passages.” Comput. Linguistics 23 (1997): 33-64.) that the NLTK TextTiling algorithm is probably based on, 

> One version of this function entails drawing a boundary only if the depth score exceeds s_hat - sigma (the liberal measure, LC). This function can be varied to achieve correspondingly varying precision/recall trade-offs. A higher precision but lower recall can be found by setting the limit to be depth scores exceeding s_hat - sigma/2 (the conservative measure, HC) instead of s_hat - sigma

I believe LC/HC uses a threshold which is (s_hat - sigma)/(s_hat - sigma/2) respectively.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -287,6 +287,7 @@
 - Yuta Nakamura <https://github.com/yutanakamura-tky>
 - Adam Hawley <https://github.com/adamjhawley>
 - Panagiotis Simakis <https://github.com/sp1thas>
+- Richard Wang <https://github.com/richarddwang>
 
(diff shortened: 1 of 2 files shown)


stevenbird:
@richarddwang - thanks for the fix (and answering my question about this conditional).

richarddwang:
@stevenbird - Is it good to be merged? Or we need some extra modification?

stevenbird:
Thanks @richarddwang. Sorry for the delay!

PR #2944, opened 2022-02-07
Automatic validation of your citation metadata

fdiblen:
Hello!

We noticed that your repository is using the Citation File Format to store citation metadata in `CITATION.cff`. This Pull Request automates validation of that file using the [cffconvert GitHub Action](https://github.com/marketplace/actions/cffconvert). That way, it's a little bit easier to be robust against future changes to the `CITATION.cff` file.

BTW it's perfectly fine if you don't feel like accepting this Pull Request for whatever reason -- we just thought it might be helpful is all.

We found your repository using a partially automated workflow; if you have any questions about that, feel free to create an issue over at https://github.com/cffbots/filtering/issues/

On behalf of the cffbots team,
@abelsiqueira / @fdiblen / @jspaaks

### DIFF
--- a/.github/workflows/cffconvert.yml
+++ b/.github/workflows/cffconvert.yml
@@ -1,0 +1,19 @@
+name: cffconvert
+
+on:
+  push:
+    paths:
+      - CITATION.cff
+
+jobs:
+  validate:
+    name: "validate"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out a copy of the repository
+        uses: actions/checkout@v2
+
+      - name: Check whether the citation metadata from CITATION.cff is valid
+        uses: citation-file-format/cffconvert-github-action@2.0.0
+        with:
+          args: "--validate"


purificant:
Could not find the source code for the `citationcff/cffconvert` docker image that this workflow will run. (Maybe missing something?)
Looking at the contents of the image itself, it looks like it runs

```
pip install cffconvert
cffconvert --validate
```
Perhaps it would be easier / simpler to understand and maintain if we run these commands directly as the last step in the job. :thinking:

purificant:
Could not find the source code for the `citationcff/cffconvert` docker image that this workflow will run. (Maybe missing something?)
Looking at the contents of the image itself, it looks like it runs

```
pip install cffconvert
cffconvert --validate
```
Perhaps it would be easier / simpler to run these commands directly as the last step in the job. :thinking:

jspaaks:
FWIW I built the docker image from https://github.com/citation-file-format/cff-converter-python/blob/2.0.0/Dockerfile

stevenbird:
Thanks @fdiblen, @purificant, @jspaaks.

tomaarsen:
I've added a `workflow_dispatch` trigger to the workflow. This allows us to go to https://github.com/nltk/nltk/actions/workflows/cffconvert.yml and click a `Run workflow` button. Otherwise, the CI run will only trigger the next time we push to `CITATION.cff`, which might not be in a while. Through this `Run workflow`, I was able to verify that:
> Citation metadata are valid according to schema version 1.2.0.

I would recommend this `workflow_dispatch` addition to other repositories that have chosen to include the `cffconvert` workflow. (CC @abelsiqueira / @fdiblen / @jspaaks)

PR #2960, opened 2022-03-09
Add .DS_Store to .gitignore for macOS users

tsheaff:


tsheaff:
Unclear why the tests would fail with just a change to the `.gitignore`. Is there known non-deterministic behavior in the test suite? I don't seem to have permissions to rerun the test suite.

tomaarsen:
There should not be non-deterministic behaviour in the test suite. The logs show that there was some issue with prover9, and the import of it, which is a third-party tool that is used in some NLTK tests. It may also be a faulty cache messing up the run - I'm not fully certain. 
I've restarted the CI. Other runs with the same codebase seem(ed) to work just fine, so I'm hopeful.

stevenbird:
Thanks @tsheaff, @tomaarsen

PR #2970, opened 2022-03-28
Missing ic argument in res, jcn and lin similarity functions of wordnet

aperezlebel:
Hello,

I just ran into a little bug when using `res_similarity`, `jcn_similarity` and `lin_similarity` functions. The `ic` argument is not passed to the corresponding methods of the Synset class.

Example:
```python
import nltk
from nltk.corpus.reader.wordnet import jcn_similarity, lin_similarity, res_similarity

nltk.download('wordnet')

# Random synsets
s1 = wn.synset_from_pos_and_offset('n', 7579787)
s2 = wn.synset_from_pos_and_offset('n', 2226429)

ic = {}
res_similarity(s1, s2, ic)
jcn_similarity(s1, s2, ic)
lin_similarity(s1, s2, ic)
```
Throws:
`TypeError: 'bool' object is not subscriptable`

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -288,6 +288,7 @@
 - Adam Hawley <https://github.com/adamjhawley>
 - Panagiotis Simakis <https://github.com/sp1thas>
 - Richard Wang <https://github.com/richarddwang>
+- Alexandre Perez-Lebel <https://github.com/aperezlebel>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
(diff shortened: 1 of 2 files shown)


tomaarsen:
Well spotted! The `verbose` parameter on the calling side was filling the `ic` parameter on the function side. Alongside your changes I've ensured that `verbose` is only passed as a keyword argument. I've also fixed the pre-commit failing tests (caused by some ill-formatted code).

I appreciate it! I'll merge this once tests pass.

PR #2971, opened 2022-03-29
Fix NumPy download link

LouisJustinTALLOT:
The Numpy download link was pointing to SciPy, so I corrected it to the NumPy download link.


tomaarsen:
Thank you! Appreciate it. 
It seems there's an issue with the CI, I'll get on that.

PR #2974, opened 2022-04-03
Warn about nonexistent OMW offsets

### LINKED ISSUE #2973: Nonexistent OMW offsets
In a few cases, the Open Multilingual Wordnet refers to synset offsets, which do not actually exist in WordNet. This issue has been known for some time (https://github.com/omwn/omw-data/issues/24#issuecomment-974789720), and it is not clear whether and when the OMW data will be fixed.

But from the point of view of NLTK, the issue is not harmless,
because it leads to fatal errors:
```

print(wn.synsets('osim', lang='hrv'))
```
```
File "nltk/corpus/reader/wordnet.py", line 1539, in synset_from_pos_and_offset
    f"No WordNet synset found for pos={pos} at offset={offset}."
nltk.corpus.reader.wordnet.WordNetError: No WordNet synset found for pos=a at offset=2002046.
```

This behaviour is unfortunate, because the Croatian word 'osim' is linked to two synset offsets, where only one is nonexistent, so it would be preferrable to return at least the good synset, instead of aborting the whole processing pipeline.
It doesn't help much to wrap the call in a try-except construct
because, although a batch would be allowed to continue,
the user would still miss the good 'osim' offset.

ekaf:
Fixes #2973: instead of raising a fatal error when an offset is not found in WordNet, only raise a warning, so that batch processing can continue. This is especially handy with OMW, which is known to contain some references to nonexistent offsets (cf. https://github.com/omwn/omw-data/issues/24#issuecomment-974789720).

```

print(wn.synsets('osim', lang='hrv'))
```

With this PR, a list of offfsets is returned, with None instead of eventual incorrect offsets, in addition to any correct offset:

`[Synset('apart.r.01'), None]`

Meanwhile, a non-fatal warning is wriiten to the standard error stream, with details about the nonexistent synset offset:

```
nltk/corpus/reader/wordnet.py:1536: UserWarning: No WordNet synset found for pos=a at offset=2002046.
  f"No WordNet synset found for pos={pos} at offset={offset}."
```

Note: maybe it could be worthwile to go through the 12 calls of the WordNetError class in wordnet.py, to see if some other of the raised errors could also be replaced by a non-fatal call of warnings.warn().

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1532,9 +1532,7 @@ class WordNetCorpusReader(CorpusReader):
             self._synset_offset_cache[pos][offset] = synset
         else:
             synset = None
-            raise WordNetError(
-                f"No WordNet synset found for pos={pos} at offset={offset}."
-            )
+            warnings.warn(f"No WordNet synset found for pos={pos} at offset={offset}.")
         data_file.seek(0)
         return synset
 


stevenbird:
Thanks @ekaf. Please feel free to raise an issue concerning the other uses of ``WordNetError``.

Issue #2976, opened 2022-04-08
Which Pointers from WordNet are Used for Synset in NLTK?


aflah02:
I'm trying to create a custom parser for wordnet and hit a roadblock. I see that there are tons of different pointer_symbols and lots of them seem almost like synonyms but not exactly synonyms. I'm trying to extract the synonyms for each word and I'm not sure which pointers should be considered. 

Any hints or leads through which I can find out what's being used?

ekaf:
Synonyms are not linked by a pointer, but by belonging to the same _synset_ ("synonym set"). You may want to read the fine WordNet documentation at https://wordnet.princeton.edu/documentation .

aflah02:
@ekaf Thanks for the reply
I understand that part however from what I could find relationships are described using pointers which tells whether it's an Antonym or Hypernym etc. ->
_A relation from a source to a target synset is formed by specifying a word from the target synset in the source synset, followed by the pointer_symbol indicating the pointer type._
Source - https://wordnet.princeton.edu/documentation/wninput5wn
So do you mean to say to find synonyms I just need to use relations which fall under these 3 pointers?

- ;c    Domain of synset - TOPIC 
- ;r    Domain of synset - REGION 
- ;u    Domain of synset - USAGE 

This is the part that's troubling me as then there are other pointers which also seem to be like synonyms like-

- -c    Member of this domain - TOPIC 
- -r    Member of this domain - REGION 
- -u    Member of this domain - USAGE

ekaf:
@aflah02, the domain relation has nothing to do with synonymy. Synonyms are found by finding the synset offsets for each synsets that a word belongs to, and then retrieving the other words that belong to each of these synsets.
If you really want to write a WordNet parser, you should expect to spend _a lot of_ time getting familiar with the official documentation, and maybe start with the 'wndb' manual page. Nobody can do it for you.

aflah02:
Thanks @ekaf this clears up a lot of things

PR #2978, opened 2022-04-11
Add function to retrieve wordnet synonyms

ekaf:
Users struggle to find the synonyms of a word (#2976) because the wordnet.py module still does not provide a convenient way to retrieve synonyms, although it claims to:
"Using synsets, helps find conceptual relationships between words such as hypernyms, hyponyms, synonyms, antonyms etc." (sic).

To remedy this deficiency, the present PR adds a function that returns a nested list with the synonyms of the different senses of a word in the given language. Collapsing the different senses into one big list is not recommendable, since the different senses are not mutual synonyms.

>from nltk.corpus import wordnet as wn
> print(wn.synonyms('car'))

[['automobile', 'machine', 'auto', 'motorcar'], ['railcar', 'railway_car', 'railroad_car'], ['gondola'], ['elevator_car'], ['cable_car']]

> print(wn.synonyms('coche', lang='spa'))

[['vehículo', 'auto', 'carro', 'automóvil', 'máquina', 'turismo'], ['vagón', 'automotor'], ['vagón', 'vagón_de_pasajeros']]

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1880,7 +1880,7 @@ class WordNetCorpusReader(CorpusReader):
     def synonyms(self, word, lang="eng"):
         """return nested list with the synonyms of the different senses of word in the given language"""
         return [
-            list(set(ss.lemma_names(lang=lang)) - {word})
+            sorted(list(set(ss.lemma_names(lang=lang)) - {word}))
             for ss in self.synsets(word, lang=lang)
         ]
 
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @ekaf. Would you mind adding an example to https://github.com/nltk/nltk/blob/develop/nltk/test/wordnet.doctest please? To appear here: https://www.nltk.org/howto/wordnet.html

stevenbird:
Thanks @ekaf!

Issue #1612, opened 2017-02-07
Link to C&C Tools/Boxer is broken


mpertierra:
The link (http://svn.ask.it.usyd.edu.au/trac/candc/wiki/Subversion) provided here (https://github.com/nltk/nltk/wiki/Installing-Third-Party-Software#cc-toolsboxer) is broken. Is there anywhere else to download these tools?

mpertierra:
I found some version of the source code here (https://github.com/valeriobasile/learningbyreading/tree/master/ext/candc), but I ran into some problems trying to install it.

lancioni:
A working version of the source code can be downloaded here:

http://www.cl.cam.ac.uk/~sc609/candc-1.00.html

I managed to compile it under Ubuntu (by adding a few #include <cstring> where errors arose). However, I am definitely looking for a tutorial...

mpertierra:
Great! It seems that it does not include Boxer. Have you found the source code for Boxer anywhere?

lancioni:
Unfortunately not.


=============================
Prof. Giuliano Lancioni
Professore Ordinario di Lingua e Letteratura Araba
Dipartimento di Lingue, Letterature e Culture Straniere
Università Roma Tre
via Ostiense, 236
00146 Roma
Tel: + 39 06 5733.8345
Fax: + 39 06 5733.8344
mailto:giuliano.lancioni@uniroma3.it
http://host.uniroma3.it/docenti/lancioni

jnothman:
a couple of servers at Sydney University have recently gone down. It's
possible this will be restored shortly (but no promises)

On 6 Mar 2017 8:13 am, "lancioni" <notifications@github.com> wrote:

Unfortunately not.


=============================
Prof. Giuliano Lancioni
Professore Ordinario di Lingua e Letteratura Araba
Dipartimento di Lingue, Letterature e Culture Straniere
Università Roma Tre
via Ostiense, 236
00146 Roma
Tel: + 39 06 5733.8345 <+39%2006%205733%208345>
Fax: + 39 06 5733.8344 <+39%2006%205733%208344>
mailto:giuliano.lancioni@uniroma3.it
http://host.uniroma3.it/docenti/lancioni

darrensapalo:
### Closest reference to C&C tools and Boxer I could find

As explained by @valeriobasile in his free [C&C/Boxer HTTP API](http://valeriobasile.github.io/candcapi/):

[...]

PR #2979, opened 2022-04-12
Update boxer.py

cestwc:
Used to have this py2 to py3 error
TypeError: write() argument must be str, not bytes

details are reported in issue [1612](https://github.com/nltk/nltk/issues/1612)


stevenbird:
Thanks @cestwc

PR #2985, opened 2022-04-18
Handle wordnet synsets that were lost in mapping

### LINKED ISSUE #2984: Senses lost in WordNet mappings
The Multilingual Wordnet uses WordNet 3.0 identifiers, and relies on mappings when used together with later wordnet versions.
However, a few WordNet 3.0 synsets are deprecated in these later versions, and thus _lost in the mapping_.
Presently, these cases lead to fatal errors. For example, one of the 3 senses of 'perceptible' in French was lost in WordNet 3.1, 
so the following outputs nothing, but raises a fatal error:

> from nltk.corpus import wordnet31 as wn
> print(wn.synsets('perceptible', lang='fra'))

This issue is analogous to #2982, because although only one of the 3 French senses of 'perceptible' was lost, the remaining 2 correct senses are also dropped, even when wrapping the call in a try-except construct.
However this issue is also different from #2982, and requires a different solution, because no WordNetError is raised, since the error is simply not handled by the wordnet.py module.

In WordNet 3.1, this problem occurs with the following OMW words:

| **Lang** | **Lost a sense in WordNet 3.1** |
| -------- | -------------------------------- |
| cat | endiablat, endimoniat, esperitat, posseït, possés |
| cmn | 埋伏着等待+的, 恶魔+的, 恶魔般+的, 狂暴+的, 着魔+的, 魔鬼+的 |
| fin | havaittava, pahojen_henkien_riivaama, riivattu, vaanimassa_oleva, villi |
| fra | obvie, perceptible |
| glg | frenético |
| ind | jelas, kejam, mengamuk, nyata, terang |
| ita | ovvio |
| jpn | 明らか, 明白 |
| lit | apsėstas, velniškas |
| pol | opętany |
| por | demoníaco, enlouquecido, espécie_de_alienação_com_tendência_à_prática_do_homicídio, fora_de_controle, possuído |
| ron | perceptibil, posedat |
| slk | démonický, diabolský, posadnutý, šialený, zúrivý |
| spa | obvio, perceptible |
| zsm | jelas, kejam, kena_rasuk, ketara, mengamuk, nyata, terang |

ekaf:
This PR fixes #2984 so that, when a wordsense has been lost in a wordnet mapping, the remaining senses of the word can still be retrieved. For ex: instead of raising a fatal error, the following retrieves the French senses of 'perceptible' in WordNet 3.1:

> from nltk.corpus import wordnet31 as wn
> print(wn.synsets('perceptible', lang='fra'))

[Synset('detectable.s.02'), Synset('None'), Synset('perceptible.a.01')]

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,7 +63,7 @@ jobs:
         id: restore-cache
         with:
           path: ~/third
-          key: new-third_${{ secrets.CACHE_VERSION }}
+          key: third_${{ secrets.CACHE_VERSION }}
 
       - name: Download third party data
         run: |
@@ -118,7 +118,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/third
-          key: new-third_${{ secrets.CACHE_VERSION }}
+          key: third_${{ secrets.CACHE_VERSION }}
         if: runner.os == 'Linux'
 
       - name: Run pytest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,7 +63,7 @@ jobs:
         id: restore-cache
         with:
           path: ~/third
-          key: third_${{ secrets.CACHE_VERSION }}
+          key: new-third_${{ secrets.CACHE_VERSION }}
 
       - name: Download third party data
         run: |
@@ -118,7 +118,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/third
-          key: third_${{ secrets.CACHE_VERSION }}
+          key: new-third_${{ secrets.CACHE_VERSION }}
         if: runner.os == 'Linux'
 
       - name: Run pytest
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1527,7 +1527,6 @@ class WordNetCorpusReader(CorpusReader):
         data_file_line = data_file.readline()
         # If valid, the offset equals the 8-digit 0-padded integer found at the start of the line:
         line_offset = data_file_line[:8]
-        #        if line_offset.isalnum() and offset == int(line_offset):
         if (
             line_offset.isalnum()
             and line_offset == f"{'0'*(8-len(str(offset)))}{str(offset)}"
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -493,7 +493,7 @@ class Synset(_WordNetObject):
         """Return all the lemma objects associated with the synset"""
         if lang == "eng":
             return self._lemmas
-        else:
+        elif self._name:
             self._wordnet_corpus_reader._load_lang_data(lang)
             lemmark = []
             lemmy = self.lemma_names(lang)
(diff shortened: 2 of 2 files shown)


ekaf:
Test failure on Ubuntu only, seems unrelated with this PR:

> NLTK was unable to find the prover9 file!
> Use software specific configuration parameters or set the PROVER9 environment variable.

ekaf:
All tests passed after changing the key of the third-party cache in ci.yaml. This change should probably be undone, but I don't know if that would restore the bad cache.

tomaarsen:
Apologies for the mess with the CI tests... One of these days we'll manage to get a test suite working with some consistency. Truthfully, I'm not sure what went wrong with those failing tests. I haven't had the time to help you out with the tests or look at this work, sadly, so I'll leave this for Steven, or for later, but I wanted to mention that I appreciate the work you're putting into NLTK's WordNet.

stevenbird:
Thank you @ekaf. Sorry about our CI consistency issues!

ekaf:
Thanks @tomaarsen and @stevenbird! According to https://github.com/actions/cache/issues/2#issuecomment-572649967, other projects see persistent corruptions of the CI cache, so it is not a specific NLTK issue.
 
But maybe the actions/cache docs include ideas that could be applied from the NLTK side. For ex. the keys used by NLTK in .github/workflows/ci.yaml include the variable ${{ secrets.CACHE_VERSION }}. The value of this variable is interpreted as void, because secrets are not passed publicly (which makes sense, since they would not be secrets otherwise), so NLTK keys might as well have nothing instead of this variable. Then it could be advantageous to use some other variable instead, for ex. the current Year/Month/Day; this would only be a slight improvement though, limiting eventual cache corruption to one day.

tomaarsen:
I added the current `${{ secrets.CACHE_VERSION }}` variable as a way to quickly modify the cache key, so that we can update the key on the fly, and restart any runs that were failed due some cache issue. This actually works quite well, but only on `nltk/nltk`: the secret is not propagated down to any forks, causing the approach to not work on any PRs.

This problem of not being able to easily update the cache on a PR is one that I've looked at before. I've tried an approach where if the last commit message contains `no-cache`, then the CI would not use the existing cache (and just make a new one instead), but I could not get that to work consistently either. The idea was that a user could perform `git commit -m "Reset the cache, no-cache" --allow-empty` to push an empty commit that reset the cache.

These are fixes of a *consequence*, while your solution *could* be the fix to the *cause*. Ensuring there is a new cache every e.g. week could prevent the cache from ever corrupting. That is just speculation, though.

ekaf:
@tomaarsen, according to the actions/cache/README.md:

> A repository can have up to 10GB of caches. Once the 10GB limit is reached, older caches will be evicted based on when the cache was last accessed.  Caches that are not accessed within the last week will also be evicted.

It could be nice, not to have to wait that long though, and the same page includes a recipe for automatically refreshing the cache each day. In NLTK's case, this would approximatively incur a cost of 2.5 Gbs*30 days. 
This proposal is a workaround, rather than a solution, because it doesn't handle the root cause, namely the fact that PR runs can corrupt the default cache persistently. Ideally, a PR would just be able to read the cache, but not to write back into it.

Issue #2918, opened 2021-12-18
Potential accidental capture of loop variable in Boxer


tomaarsen:
Hello!

Within the last list comprehension in this return statement, the `cond` value will always take the value of the last element of `conds`:
https://github.com/nltk/nltk/blob/4a130f1cdf642327378a4387b319cd2df6f79287/nltk/sem/boxer.py#L533-L537

This seems like it would be unexpected or undesired behaviour.

For those unaware of what I'm talking about, a Python implementation detail causes the following:
```python
int_list = [1, 2, 3, 4, 5]
functions = [
    lambda: print(i) 
    for i in int_list
]
for func in functions:
    func()
```
outputs
```
5
5
5
5
5
```
while one might expect
```
1
2
3
4
5
```

I don't have the time to look into this now, and I'm not sure if this affects many people, so that's why I'll just write it as an issue instead. This probably isn't the highest priority issue, by the way.

- Tom Aarsen

gokulnathselvaraj27:
i would like to contribute

PR #2987, opened 2022-04-29
 Potential loop variable fix

gokul-the-dev:
[#2918](https://github.com/nltk/nltk/issues/2918) Looks like we have a potential accidental capture of loop variable in Boxer.

The fix removes the comprehensive list iteration to conventional loop.

### DIFF
--- a/nltk/sem/boxer.py
+++ b/nltk/sem/boxer.py
@@ -530,11 +530,15 @@ class BoxerOutputDrsParser(DrtParser):
         else:
             return None
         self.assertToken(self.token(), ")")
+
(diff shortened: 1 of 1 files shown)


tomaarsen:
I'm afraid this does not resolve the issue, I believe. Take for example the following program, with code that resembles your implementation.
```python
conds = [1, 2, 3, 4, 5]

conda_result = []
for cond in conds:
    conda_result.append(lambda sent_index, word_indices: cond)

for func in conda_result:
    print(func(None, None))
```
outputs:
```
5
5
5
5
5
```

Issue #2989, opened 2022-05-01
Failing tests in DocStrings


ekaf:
Checking the Python files in NLTK with "python -m doctest" reveals that many tests are failing. In many cases, the failures are just cosmetic discrepancies between the expected and the actual output, such as missing a blank line, or unescaped linebreaks. Other cases may be real bugs.

If these failures could be avoided, it would become possible to improve CI by running "python -m doctest" each time a python file was modified.

Here is an updated list, after applying #2990, and unzipping those corpora marked unzip="1":

> nltk/parse/corenlp.py: 9 passed and 33 failed.
nltk/parse/stanford.py: 1 passed and 16 failed.
nltk/corpus/reader/framenet.py: 79 passed and 9 failed.
nltk/tokenize/regexp.py: 14 passed and 7 failed.
nltk/draw/table.py: 0 passed and 6 failed.
nltk/tokenize/simple.py: 7 passed and 6 failed.
nltk/text.py: 16 passed and 5 failed.
nltk/tokenize/stanford_segmenter.py: 3 passed and 5 failed.
nltk/draw/util.py: 1 passed and 4 failed.
nltk/tag/hunpos.py: 1 passed and 4 failed.
nltk/tokenize/__init__.py: 6 passed and 4 failed.
nltk/translate/ibm1.py: 12 passed and 4 failed.
nltk/tag/brill_trainer.py: 23 passed and 3 failed.
nltk/tokenize/destructive.py: 8 passed and 3 failed.
nltk/tokenize/treebank.py: 32 passed and 3 failed.
nltk/tokenize/util.py: 27 passed and 3 failed.
nltk/translate/meteor_score.py: 8 passed and 3 failed.
nltk/corpus/reader/categorized_sents.py: 5 passed and 2 failed.
nltk/corpus/reader/opinion_lexicon.py: 3 passed and 2 failed.
nltk/corpus/reader/reviews.py: 8 passed and 2 failed.
nltk/metrics/agreement.py: 8 passed and 2 failed.
nltk/parse/dependencygraph.py: 13 passed and 2 failed.
nltk/parse/evaluate.py: 6 passed and 2 failed.
nltk/parse/util.py: 5 passed and 2 failed.
nltk/tag/stanford.py: 2 passed and 2 failed.
nltk/tokenize/stanford.py: 3 passed and 2 failed.
nltk/corpus/__init__.py: 1 passed and 1 failed.
[...]

PR #2990, opened 2022-05-01
Fix some tests in Wordnet-related DocStrings

ekaf:
Solve all the cosmetic Wordnet-related test failures mentioned in #2989. This leaves only two real test failures in wordnet.py, namely the ones solved in #2988.


stevenbird:
Thanks @ekaf

Issue #2948, opened 2022-02-17
key phrase extraction in NLTK


kamrapooja:
Do we have key phrase extraction in NLTK?

12mohaned:
I do not believe you will find a ready-made algorithm in NLTK to extract keywords from a corpus. Also, I am not sure that NLTK should have such a feature, but it would be a nice add-on. However, you can use NLTK mainly for preprocessing your corpus. For example, tokenization, removing stop words and punctuations, lemmatization, stemming or even pos tagging tokens in your corpus. I believe this is the role of NLTK in solving your problem. 

Regards.

rmalouf:
You can use NLTK's association metrics for doing keyword analysis. For example, this finds the keywords associated with the "romance" section of the Brown corpus as measured by likelihood ratio:
```python
from nltk import FreqDist, collocations
from nltk.corpus import brown

bigram_measures = collocations.BigramAssocMeasures()

romance = FreqDist(w.lower() for w in brown.words(categories='romance') if w.isalpha())
freq = FreqDist(w.lower() for w in brown.words() if w.isalpha())

key = FreqDist()
for w in romance:
    key[w] = bigram_measures.likelihood_ratio(romance[w], (freq[w], romance.N()), freq.N())

for k,v in key.most_common(10):
    print(f'{k:10s} {v:9.3f}')
```
And the output is:
```
she         1163.325
i            995.961
her          930.528
you          513.149
of           501.891
is           463.386
had          421.615
he           411.000
the          347.632
said         300.811
```

kamrapooja:
no option in NLTK for Key Phrases

PR #2994, opened 2022-05-06
Add example of keyword analysis

rmalouf:
Add an example showing how to use association metrics to find keywords (#2948)


stevenbird:
Thanks @rmalouf!

PR #2998, opened 2022-05-06
fix(childes): use correct xmlns

### LINKED ISSUE #2997: CHILDES Corpus Reader doesn't parse data correctly
### Steps to reproduce
1. Download and unpack the CHILDES dataset
2. Follow the instructions at [Sample usage of childes](https://www.nltk.org/howto/childes.html)

### Expected outcome
Output is as shown on the sample usage page.

### Actual outcome
No words, sentences, etc. are parsed and are empty.

### Cause
- The reader tries to find the nodes using the wrong namespace.
- The namespace (NS) was accidentally changed to use the HTTPS scheme in #2852.

sid-maddy:
Fixes #2997

### DIFF
--- a/nltk/corpus/reader/childes.py
+++ b/nltk/corpus/reader/childes.py
@@ -20,7 +20,7 @@ from nltk.corpus.reader.xmldocs import ElementTree, XMLCorpusReader
 from nltk.util import LazyConcatenation, LazyMap, flatten
 
 # to resolve the namespace issue
-NS = "https://www.talkbank.org/ns/talkbank"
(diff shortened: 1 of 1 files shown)


sid-maddy:
@stevenbird, sorry for the mention, but could you take a look at this PR?

tomaarsen:
I was able to reproduce the issue, and this PR does seem to resolve the issue. It's clear what caused this issue - the recent update from https to http on many links. I must've been too careless here, as the xml obviously still requires `http` in this case. Thank you for bringing this to our attention and providing this fix!

PR #2999, opened 2022-05-07
Update skolemize.py

cestwc:
add a function, `richardize `(just a name trying to make it not confused with `simplify` or `__reduce__` ), modified from `skolemize`.

The function name does not represent any logician Richard, and maybe the NLTK maintenance party can choose a better name.
The new function changes minimum from `skolemize`, and instead conversion to CNF (which is not equivalent but equisatisfiable to the input FOL expression), `richardize(your_expr).equiv(your_expr)` will be True.

The motivation of this function is that I did not find `simplify`  to work in the way I wished, and for a pretty messy expression, _e.g._,  those with a lot of consecutive negations, the new function can make it neater.

### DIFF
--- a/nltk/sem/nd.py
+++ b/nltk/sem/nd.py
@@ -102,12 +102,12 @@ def get_basic_assumptions(expression):
                     var_all = var
                 else:
                     expr_ = ExistsExpression(var_, expr_)
-            basic_assumptions.add(- AllExpression(var_all, expr_))
+            basic_assumptions.add(-AllExpression(var_all, expr_))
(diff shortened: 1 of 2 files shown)


stevenbird:
@cestwc: it looks like you've duplicated ```skolemize``` and made some changes to it. We need a well argued case for having (and maintaining) **two** similar functions. Could we please have more examples of different behaviours of the two functions? Is there a case for having both?

zifuyang:
i second this

ekaf:
Deficient motivation

PR #3000, opened 2022-05-10
Remove unnecessary sqlite code

JelleZijlstra:
`OptimizedUnicode` has been an alias for `str` since Python 3.2 and is scheduled for removal in Python 3.12. The default factory is already `str`, so the line wasn't doing anything.

See python/cpython#92547.


erlend-aasland (approved):
Looks good. Thanks, Jelle.

stevenbird:
Thanks @JelleZijlstra, @erlend-aasland

PR #3010, opened 2022-05-26
Iota operator

fernandocar86:
This contribution consists of the addition of the iota operator to the semantic model used for parsing feature based grammars. Iota operator is used in the semantic literature for formalizing meaning of, for example, definite expressions.


stevenbird:
Thanks @fernandocar86... welcome contribution.

PR #3012, opened 2022-06-11
Fix errors in API docs

rmalouf:
This PR resolves some of the warnings and errors that come up when generating the API docs. It also replaces better-apidoc with sphinxcontrib-apidoc.

There's a more general problem with the API docs, though: it might just be me, but I find them to be almost completely unusable. Sphinx-apidoc includes everything whether it's part of the API or not, and it follows the organization of the source files which doesn't always match the logical organization of the API. I just look at the source code when I want to know something but that's not a good option for, say, my linguistic students. More carefully constructed API docs would make the project a lot more accessible to new users, and that's a project I'd be willing to take on it there's any interest.


stevenbird:
@rmalouf, we would be only too pleased if you wanted to rework the API docs as you suggest. Happy to discuss details with you... slack might be easiest. Will invite you.

Issue #3001, opened 2022-05-11
Verbnet longid() method returns wrong results on some shortids


TMPxyz:
For verbnet3.3 corpus,

`verbnet.longid("114-1")` is supposed to return longid "act-114-1", but it instead returns "114-1".

It seems that the bug is rooted in the regex in verbnet.py, which wrongly matches "114-1" as a longid :

`    _LONGID_RE = re.compile(r"([^\-\.]*)-([\d+.\-]+)$")`

I currently use `_LONGID_RE = re.compile(r"([A-Za-z_]*)-([\d.\-]+)$")` as a workaround.

Miguelvs23:
Hi! Can i work on this issue?

tomaarsen:
I'm open to a PR on this.

PR #3018, opened 2022-07-12
Resolution issue #3001

JuanIMartinezB:
There was a reported issue with the "longid()" and "shortid()" methods of the Verbnet.py file. Basically, there are two types of ID, short id (eg. 32.5) and long id(eg. example-32.5). The methods involved change a short id to a long id and vice versa. The problem was that verbnet.longid("114-1") is supposed to return the long id "act-114-1", but it instead returns "114-1". This issue was solved by changing the Regex pattern to match the id types while still passing the tests for all other cases. 

The change in the long id  Regex pattern was:

   _LONGID_RE = re.compile(r"([^\-\.]*)-([\d+.\-]+)$") ------>  _LONGID_RE = re.compile(r"([A-Za-z_]+)-([\d.\-]+)$")  

The new pattern tries to match an alphabetic sequence first (without considering empty sequences), then to match a minus sign "-" and finally the associated short id.

### DIFF
--- a/nltk/corpus/reader/verbnet.py
+++ b/nltk/corpus/reader/verbnet.py
@@ -33,8 +33,26 @@ class VerbnetCorpusReader(XMLCorpusReader):
     https://verbs.colorado.edu/~mpalmer/projects/verbnet.html
     """
 
+    #: Supported VerbNet versions.
+    SUPPORTED_VERSIONS = ("2.1", "3.2", "3.3")
+
     # No unicode encoding param, since the data files are all XML.
-    def __init__(self, root, fileids, wrap_etree=False):
+    def __init__(self, root, fileids, wrap_etree=False, version="2.1"):
+        """
+        :param root: The root directory for the corpus.
+        :param fileids: A list or regexp specifying the fileids in the corpus.
+        :param wrap_etree: If true, wrap the ElementTree in an ElementWrapper.
+        :param version: The VerbNet version string (default ``"2.1"``).
+            NLTK ships VerbNet 2.1 via ``nltk.download('verbnet')``.
+            Use ``"3.2"`` or ``"3.3"`` when pointing *root* at a local
+            copy of VerbNet 3.2 or 3.3.
+        """
+        if version not in self.SUPPORTED_VERSIONS:
+            raise ValueError(
+                f"VerbNet version {version!r} is not supported. "
+                f"Supported versions: {self.SUPPORTED_VERSIONS}"
+            )
+        self._version = version
         XMLCorpusReader.__init__(self, root, fileids, wrap_etree)
 
         self._lemma_to_class = defaultdict(list)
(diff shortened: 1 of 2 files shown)


tomaarsen:
Please have a look at the failing tests, e.g. [these](https://github.com/nltk/nltk/runs/7299050856?check_suite_focus=true).

alvations:
## Summary of VerbNet Corpus Reader Changes                                                                                                                                                               
                                                                                                                                                                                                          
### Bug Fix: `longid()` / `shortid()` regex for dash-style IDs (issue #3001)                                                                                                                              
 
**Problem:** `verbnet.longid("114-1")` returned `"114-1"` instead of the expected `"act-114-1"`. The old `_LONGID_RE` pattern `([^\-\.]*)-([\d+.\-]+)$` matched numeric-only prefixes (like `114`),       
causing short IDs with dashes to be misidentified as long IDs.
                                                                                                                                                                                                          
**Fix:**                                                  

| Regex | Before | After |                                                                                                                                                                                
|-------|--------|-------|
| `_LONGID_RE` | `([^\-\.]*)-([\d+.\-]+)$` | `([A-Za-z_]+)-([\d.-]+)$` |                                                                                                                                  
| `_SHORTID_RE` | `[\d+.\-]+$` | `[\d.\-]+$` |                                                                                                                                                            
                                                                                                                                                                                                          
- `_LONGID_RE` now requires at least one alphabetic/underscore character before the first dash, so `"114-1"` correctly falls through to the short ID lookup.                                              
- `_SHORTID_RE` removes the stray literal `+` inside the character class (was being matched as a literal character, not a quantifier).                                                                    
                                                                                                                                                                                                          
### New Feature: VerbNet version support (2.1, 3.2, 3.3)                                                                                                                                                  

Added a `version` parameter to `VerbnetCorpusReader.__init__()`:                                                                                                                                          
                                                          
```python
# NLTK's bundled VerbNet 2.1 (default, matches nltk.download('verbnet'))
vn = VerbnetCorpusReader(root, fileids)                                                                                                                                                                   
                                                                                                                                                                                                          
# Local VerbNet 3.2 or 3.3                                                                                                                                                                                
vn = VerbnetCorpusReader('/path/to/verbnet3.3', r'(?!\.).*\.xml', version='3.3')                                                                                                                          
```

### Notes: 
                                                        
- Defaults to "2.1" to match what ships with nltk.download('verbnet')                                                                                                                                     
- Validates against SUPPORTED_VERSIONS = ("2.1", "3.2", "3.3")                                                                                                                                            
- Exposes vn.version property                                                                                                                                                                             
- All three versions verified: 429 classes (2.1), 498 classes (3.2), 601 classes (3.3) — full round-trip of all IDs passes                                                                                

Added the unittest to test the supported version.

alvations (approved):
LGTM after checking that the corpus reader works with v3.3 , v3.2 and v2.1 

Also added unittest helps.

stevenbird:
Great contribution! Thanks @JuanIMartinezB for the initial report and diagnosis, and @alvations for revisiting and wrapping this up.

PR #3022, opened 2022-07-19
Use `input()` in `nltk.chat.chatbot()` for Jupyter support

### LINKED ISSUE #3021: nltk.chat.chatbot() endless loop
When I type `import nltk` followed by `nltk.chat.chatbots()`, it lists/asks which one I want to talk to, and then endlessly scrolls the following: ` Enter a number in the range 1-5:     Error: bad chatbot number`, in both Jupyter and Spyder.

tomaarsen:
Fixes #3021 

Hello!

### Pull request overview
* Replace `sys.stdin.readline()` with `input()` in `nltk.chat.chatbot()`

### Details
The following snippet is used to let users pick which of the 5 NLTK chatbots should be instantiated:
https://github.com/nltk/nltk/blob/1f28903066fe3ee3a4ab75085bfd5254184d135c/nltk/chat/__init__.py#L43-L44

However, the `sys.stdin.readline()` constantly triggers in such a way that `choice` is an empty string. This causes an endless loop, exactly as described in #3021:
image
Note that this is the output directly after running the cell. Be warned when testing this, this ended up crashing my computer due to the endless loop.

---

### Changes
This PR simply updates this somewhat outdated form of getting user input by using the better supported `input()`. Running the cell now results in:
image
After which the chatbot works like intended & expected.

Thank you @jmcelroy01 for raising this issue.

- Tom Aarsen

### DIFF
--- a/nltk/chat/__init__.py
+++ b/nltk/chat/__init__.py
@@ -33,15 +33,12 @@ bots = [
 
 
 def chatbots():
-    import sys
-
     print("Which chatbot would you like to talk to?")
     botcount = len(bots)
     for i in range(botcount):
         print("  %d: %s" % (i + 1, bots[i][1]))
     while True:
-        print("\nEnter a number in the range 1-%d: " % botcount, end=" ")
-        choice = sys.stdin.readline().strip()
+        choice = input(f"\nEnter a number in the range 1-{botcount}: ").strip()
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @tomaarsen

Issue #2244, opened 2019-02-27
CoreNLPServer on a non-default port


amadanmath:
If port is specified for `CoreNLPServer`, it is tested, put into the URL, but never put into `corenlp_options` ([here](https://github.com/nltk/nltk/blob/ac6a2394800138e4638279abc47a2394093cca72/nltk/parse/corenlp.py#L84)). Thus, `CoreNLPServer(port=9001).url` will be `http://localhost:9001`, but it will actually listen at `9000`. Should there not be another line there, to the effect of

    corenlp_options.extend(['-port', str(port)])

Similarly, two lines above, in case `None` was passed and the default `9000` is not available, `try_port`  tries to discover an available port; it is appended to `corenlp_options`, but without the `-port` option, and thus AFAIK does not have an effect. That line, too, should likely be changed to the line quoted above.

PR #3031, opened 2022-08-08
Fix CoreNLPServer non-default port issue (#2244)

heungson:
Issue #2244 fixed.
In case where a port other than the default port (9000) is used, '-port' option is also added to the command line options.


tomaarsen (approved):
Looks good! This works for me locally, too. Thank you for this!

Issue #2578, opened 2020-08-04
TweetTokenizer add new emoticons characters


danielafe7-usp:
Hello! There is any way to add new characters to the emoticons tokenization ?
For example:
from nltk.tokenize import TweetTokenizer
tknzr = TweetTokenizer()
s1 = "This is a cooool :-*"
tknzr.tokenize(s1)

It gives as result:
['This', 'is', 'a', 'cooool', ':', '-', '*']

But I need to preserve these specific emotion ":-*".
Thanks for advance!

tomaarsen:
@danielafe7-usp 
No, this is not (conveniently) possible.
TweetTokenizer detects emoticons using a specific regular expression:
https://github.com/nltk/nltk/blob/54221dec0bae2642d1642d182d8a381c88b86bd0/nltk/tokenize/casual.py#L57-L70

This regex will not match `:-*` (because `*` is not considered a valid option for a mouth), and cannot be extended in runtime by supplying some list of specific emoticons to also search for. However, this *could* be implemented as a feature. I might look into this if I have some spare time.

ajdapretnar:
So TweetTokenizer will not be able to recognize valid Unicode emojis?
Because currently, what I get from this `'🇸🇮🤝🇺🇦'` is this `'🇸', '🇮', '🤝', '🇺', '🇦'`. Is there an option to add support from Unicode emojis?

tomaarsen:
@ajdapretnar  I feel like this is a bit of a complex situation. To give some context:
```python
>>> string = "🇸🇮🤝🇺🇦" 
>>> len(string)
5
>>> string[0]
'🇸'
```
Long story short, flags are combinations of regional indicator symbols. When converted into a string, Python is the one unable to recognize that 🇸🇮 is a (single) valid Unicode emoji, but it instead sees it as two separate regional indicator symbols. TweetTokenizer only rarely merges Unicode characters together: Only with Zero-Width-Joiners and Skin tone modifiers I believe (See #2843). 

However, in that context, the Zero-Width-Joiners and Skin tone modifiers don't have other uses, but the regional indicator do have other users. If someone uses them to spell something out, grouping them together might not necessarily be the desired behaviour. (Although I could be convinced if someone is able to add a concise and simple implementation to [this](https://github.com/nltk/nltk/blob/b0e85694107992e00a2f9fb48e6410c50fe1f1f6/nltk/tokenize/casual.py#L148) regex)

ajdapretnar:
[...]

PR #3034, opened 2022-08-22
nltk.tokenize.casual.TweetTokenizer: Add support to tokenize emoji flag sequences

essandess:
nltk.tokenize.casual.TweetTokenizer: Add support to tokenize emoji flag sequences
* TweetTokenizer currently splits emoji flags into individual tokens of enclosed letters, e.g. 🇨🇦 -> '🇨 ', '🇦 '
* This patch keeps emoji flag sequences intact
* Without this PR:
> ```python
> nltk.tokenize.casual.TweetTokenizer().tokenize(text='Hi 🇨🇦, 😍!!')
> ['Hi', '🇨', '🇦', ',', '😍', '!', '!']
> ```
* With this PR:
> ```python
> nltk.tokenize.casual.TweetTokenizer().tokenize(text='Hi 🇨🇦, 😍!!')
> ['Hi', '🇨🇦', ',', '😍', '!', '!']
> ```

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -291,6 +291,7 @@
 - Alexandre Perez-Lebel <https://github.com/aperezlebel>
 - Fernando Carranza <https://github.com/fernandocar86>
 - Martin Kondratzky <https://github.com/martinkondra>
+- Steven Thomas Smith <https://github.com/essandess>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -384,11 +384,20 @@ class TestTokenize:
         assert result == expected
 
         # emoji flag sequences, including enclosed letter pairs
-	# Expected behavior from:
-        # https://github.com/nltk/nltk/pull/3034#issuecomment-1223956957
+        # Expected behavior from #3034
         test4 = "🇦🇵🇵🇱🇪"
         expected = ["🇦🇵", "🇵🇱", "🇪"]
         result = tokenizer.tokenize(test4)
+        assert result == expected
+
+        test5 = "Hi 🇨🇦, 😍!!"
+        expected = ["Hi", "🇨🇦", ",", "😍", "!", "!"]
+        result = tokenizer.tokenize(test5)
+        assert result == expected
+
+        test6 = "<3 🇨🇦 🤝 🇵🇱 <3"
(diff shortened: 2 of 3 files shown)


tomaarsen:
See https://github.com/nltk/nltk/issues/2578#issuecomment-1112066105 for some related discussion. In short, I am in favour of adding this behaviour, but I would like to start a discussion on whether it would be wise to include an option to turn it off in `TweetTokenizer.__init__`. 

After all, as you are aware, flags are made up of regional indicator symbols. These symbols are also frequently used outside of the context of flags, like to spell out words: 🇦 🇵 🇵 🇱 🇪. However, this is commonly done with spaces between the tokens, as otherwise that spelling of "APPLE" would have had a Polish flag in the middle. In short, it may not be an issue.

Beyond that discussion, would you be able to write a few tests to back up that your implementation works as intended? For example in here: https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_tokenize.py

Also, I'm uncertain why the CI is failing. It seems like a third party download of a file used in one of the tests has failed:
> curl: (28) Failed to connect to www.cs.unm.edu port 443: Connection timed out

It's unrelated to your PR, though, so no stress.

essandess:
> would you be able to write a few tests to back up that your implementation works as intended? For example in here: https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_tokenize.py

Yes, I'll modify that file as part of this PR.

Given your comments, what do you think of the decision use a regex that encodes **all** pairs of enclosed letters as tokens? Under this behavior, your example is tokenized as follow. Is this "correct"? Or should this repo have a record of all ISO country code letter pairs and treat these separately in a much more complicated regex? My preference is to avoid this.

> ```python
> nltk.tokenize.casual.TweetTokenizer().tokenize('🇦🇵🇵🇱🇪')
> ['🇦🇵', '🇵🇱', '🇪']
> ```

tomaarsen:
I also prefer your current implementation, i.e. any two directly adjacent regional indicator symbols. This is also more future-proof. The behaviour from your little snippet seems like the expected output as well. 👍

essandess:
> I also prefer your current implementation, i.e. any two directly adjacent regional indicator symbols. This is also more future-proof. The behaviour from your little snippet seems like the expected output as well. 👍

Great, implemented with a test in [test_tokenize.py](../blob/develop/nltk/test/unit/test_tokenize.py).

tomaarsen:
I've expanded on your test case a little bit, and fixed a merge conflict on the AUTHORS.md file. Should be all set now!
Thank you for this work!

PR #3036, opened 2022-08-23
Downloader GUI: Replacing black font color with blue to make the links readable 

SaudKadiri:
The change will ensure that links are visible in both dark and light mode for macOS.

**Issue**
The downloader when used on macOS's dark mode has the text with foreground color as black and the background as dark grey, which makes it difficult to read and makes the text quite unreadable. Also difficult for color blind people. As can be seen in the image below entry boxes for URL & download_dir have quite an unreadable text (highlighted by red arrows).
current-dark-mode
Not a problem when using the light mode though
current-light-mode
<br>

**Solution**
Replace the existing **black** value with **(blue)#007aff** for the **disabledforeground** attribute<br>

**Advantages:**
1. Add support for color blind people.
2. The new color- blue (#00faff) is compliant with <a href="https://developer.apple.com/design/human-interface-guidelines/foundations/color">Apple's human interface guide</a>; and is named by Apple as **systemBlueColor**.
<br>

**Proposed change**
new-dark-mode
new-light-mode

**On Windows and Ubuntu**
These two operating systems have downloader application in light mode only.

Windows existing             |  Windows proposed
:-------------------------:|:-------------------------:
windows-existing  |  windows-new

Ubuntu existing             |  Ubuntu proposed
:-------------------------:|:-------------------------:
ubuntu-new | ubuntu-exist

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1179,9 +1179,6 @@ class WordNetCorpusReader(CorpusReader):
                 "The multilingual functions are not available with this Wordnet version"
             )
             self.omw_langs = set()
-        else:
-            self.add_provs(self._omw_reader)
-            self.omw_langs = set(self.provenances.keys())
 
         # A cache to store the wordnet data of multiple languages
         self._lang_data = defaultdict(list)
@@ -1280,6 +1277,10 @@ class WordNetCorpusReader(CorpusReader):
 
         if lang in self._lang_data.keys():
             return
+
+        if self.langs() == ["eng"]:
+            self.add_provs(self._omw_reader)
+            self.omw_langs = set(self.provenances.keys())
 
         if lang not in self.langs():
             raise WordNetError("Language is not supported.")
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

This looks great. Thank you for the change and the detailed PR.
However, I do have one issue: selecting the text turns it black again, as you can see here:
image

This is because the `foreground` value is not changed from the default of `black` in your PR. See [here](https://anzeljg.github.io/rin2/book2/2405/docs/tkinter/entry.html) for some more info on `Entry`. If we also update `foreground`, then I believe it should work perfectly.

tomaarsen:
I had some issues pushing my 0e2d61129982fa1cf0d53f4b8beafd8f3e87bca9 to your branch, and I made a mistake with the push here. My bad. That said, it's all merged now (despite this saying "closed" instead). You can see your commit in https://github.com/nltk/nltk/commits/develop, for example.

Thank you for this work!

SaudKadiri:
Thanks a lot @tomaarsen. Glad that it eventually got a go ahead.🤝

PR #3038, opened 2022-08-25
Change in Polish encodings of udhr corpus reader

### LINKED ISSUE #3037: Incorrect encoding in `UdhrCorpusReader` corpus reader
There are problems with encoding of Polish characters in `Polish-Latin2` and `Polish_Polski-Latin2` files in the `udhr` corpus. When importing the text using the corpus reader, one obtains the improperly rendered characters. For example

```python
>>> from nltk.corpus import udhr
>>> udhr.raw('Polish-Latin2')[141:164]
'jednomy\x9clnie Powszechnš'
```
The character `\x9c` should be rendered correctly as `ś`, while the character `š` as `ą`. The encoding that `nltk` reader uses for those files is `ISO-8859`:

```python
>>> udhr.encoding('Polish-Latin2')
'ISO-8859-2'
```

However, the files themselves are encoded in `WINDOWS 1250`, as can be seen by re-encoding them with that encoding.

```python
>>> bytes(udhr.raw('Polish-Latin2'), 'ISO-8859-2').decode('cp1250')[141:164]
'jednomyślnie Powszechną'
```

While the code above invloves to `Polish-Latin2` same issue concerns `Polish_Polski-Latin2`.

Modifying `UdhrCorpusReader.ENCODINGS` in `nltk/corpus/reader/udhr.py` resolves the issue.

hamiltonianflow:
Modification in `UdhrCorpusReader` corpus reader class, which changes the encoding used for Polish language files `Polish-Latin2` and `Polish_Polski-Latin2` from `ISO-8859-2` to `cp1250`, so that they are properly read.

Fixes #3037

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -291,6 +291,7 @@
 - Alexandre Perez-Lebel <https://github.com/aperezlebel>
 - Fernando Carranza <https://github.com/fernandocar86>
 - Martin Kondratzky <https://github.com/martinkondra>
+- M.K. Pawelkiewicz <https://github.com/hamiltonianflow>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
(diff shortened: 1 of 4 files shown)


tomaarsen:
Thank you for finding this! Your fix is simple and proper too. I've added a simple test in 0526ad54e9f98b20cab27fcfba909abd8efd7282 to verify that this will continue to work in the future (and I bunged up the git history of this PR somewhat... Apologies)

PR #3040, opened 2022-08-25
[Refactoring]Few refactoring: using `with open` and `np.mean` 

maldil:
This PR is to use `with open` to make the code more Pythonic and to utilize optimized `np.mean` to make the code more concise. What are your thoughts on the change?

### DIFF
--- a/nltk/tokenize/texttiling.py
+++ b/nltk/tokenize/texttiling.py
@@ -288,7 +288,7 @@ class TextTilingTokenizer(TokenizerI):
 
         boundaries = [0 for x in depth_scores]
 
-        avg = sum(depth_scores) / len(depth_scores)
+        avg = numpy.mean(depth_scores)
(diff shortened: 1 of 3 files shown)


tomaarsen:
Hello! Thank you for taking your time to write this PR.
I've done some quick tests, and I found that `numpy.mean` is not necessarily always quicker than just the good old `sum(values) / len(values)`:
```python

import numpy as np
import timeit

# Get a list of 100 random values
values = list(np.random.rand(100))

print(timeit.timeit("np.mean(values)", globals=globals()))
print(timeit.timeit("sum(values) / len(values)", globals=globals()))
```
```
11.3234446
4.337459900000001
```

So I won't merge those changes, but I do very much like the use of `with open` change! I'll cherry-pick that commit and merge it (see d89a592c9bf1f1c3200d30c6908a89b1cc381f4b). 
Thank you!

PR #3064, opened 2022-10-18
Create hindi-english stopwords

deepakbhagat7:
In India, it will be beneficial because we will type in roman but the language is Hindi.
It may be called Hinglish


ekaf:
Thank you for your contribution and for suggesting the addition of Hindi-English stopwords!

However, the inclusion of stopword lists such as this one should be handled in the [nltk_data](https://github.com/nltk/nltk_data) repository, rather than directly in the core `nltk` library. The `nltk_data` repository is specifically designed for managing corpora, lexicons, and other linguistic resources, including stopword lists for various languages.

If you would like to contribute these stopwords, please consider submitting them as a new resource or an update in the `nltk_data` repository instead. You can find guidelines for contributing data resources [here](https://github.com/nltk/nltk_data#contributing-data).

Thank you again for your interest in improving NLTK!

PR #3066, opened 2022-11-04
lambda loop variable closure fix

### LINKED ISSUE #2918: Potential accidental capture of loop variable in Boxer

asishm:
Closes #2918 

Issue arises because of this issue - https://stackoverflow.com/questions/2295290/what-do-lambda-function-closures-capture

the function generator could use a better name.

### DIFF
--- a/nltk/sem/boxer.py
+++ b/nltk/sem/boxer.py
@@ -530,11 +530,15 @@ class BoxerOutputDrsParser(DrtParser):
         else:
             return None
         self.assertToken(self.token(), ")")
+
+        def func_gen(x):
+            return lambda sent_index, word_indices: x
+
         return [
             lambda sent_index, word_indices: BoxerPred(
                 self.discourse_id, sent_index, word_indices, arg, tok, "n", 0
             )
-        ] + [lambda sent_index, word_indices: cond for cond in conds]
+        ] + [func_gen(cond) for cond in conds]
 
     def _handle_date(self, arg):
         # []: (+), []:'XXXX', [1004]:'04', []:'XX'


stevenbird:
Thanks for the fix @asishm. How about using `map()` with a lambda argument to avoid the need to name the function, roughly like this:

```
map(lambda ..., conds)
```

asishm:
So something like `map(lambda cond: lambda sent_index, word_indices: cond, conds)` ?

tomaarsen:
That seems to work!
```python
conds = "iterable"

print("Current approach")
functions = [lambda sent_index, word_indices: cond for cond in conds]
for function in functions:
    print(function(None, None))

print("Proposed approach")
def func_gen(x):
    return lambda sent_index, word_indices: x

functions = [func_gen(cond) for cond in conds]
for function in functions:
    print(function(None, None))

print("Proposed approach 2")
# For all conditions, create a 2-argument functions that returns the condition regardless of input
functions = map(lambda cond: lambda sent_index, word_indices: cond, conds)
for function in functions:
    print(function(None, None))
```
outputs:
```
Current approach
e
e
e
e
e
e
e
e
Proposed approach
i
t
e
r
a
b
l
e
Proposed approach 2
i
t
e
r
a
b
l
e
```

However, I can't properly verify this in Boxer, as I can't manage to download/install it correctly. The original source of C&C and Boxer seems to be down nowadays: svn.ask.it.usyd.edu.au/trac/candc

It seems a bit risky to merge this without verifying that it works as intended now.

stevenbird:
@asishm Are you able to confirm that the fix works please?

PR #3068, opened 2022-11-05
ci: add labeler

purificant:
This PR introduces automatic [labelling](https://github.com/actions/labeler) of new pull requests based on the paths of files being changed.

For example, CI label will be applied when `.github/workflows/*` or `.github/labeler.yml` files change. We can define more labels and paths.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,3 +2,48 @@
 CI:
   - .github/workflows/*
   - .github/labeler.yml
+
+classifier:
+  - nltk/classify/**/*
+
+cli:
+  - nltk/cli.py
+
+cluster:
+  - nltk/cluster/**/*
+
+corpus:
+  - nltk/corpus/**/*
+
+GUI:
+  - nltk/app/**/*
+
+internals:
+  - nltk/internals.py
+
+language-model:
+  - nltk/lm/**/*
+
+metrics:
+  - nltk/metrics/**/*
+
+parsing:
+  - nltk/parse/**/*
+
+sentiment:
+  - nltk/sentiment/**/*
+
+stem/lemma:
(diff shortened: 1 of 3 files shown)


tomaarsen:
I hadn't seen this before! Great idea! 
There's quite a few [labels](https://github.com/nltk/nltk/labels), and most don't directly correspond to a path, but some definitely do. Perhaps we are best off with creating a label for each directory in [the nltk directory](https://github.com/nltk/nltk/tree/develop/nltk), and assigning labels according to changes in those directories?

Out of curiosity, can this be used retroactively on old PRs?

purificant:
This workflow runs on [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) when a PR is opened or changed. Once merged, it will also run for existing PRs if they change. Maybe is it possible to trigger this workflow by adding or removing a label from a PR for existing PRs.

Right now this is a proof of concept for CI label. Happy to build on it in future PRs @tomaarsen as you've suggested if it's useful.
The most important component is the mapping between label <-> path, which we can define easily along the lines of:

```yaml
# Add 'label1' to any changes within 'example' folder or any subfolders
label1:
- example/**/*

# Add 'label2' to any file changes within 'example2' folder
label2: example2/*
```

tomaarsen:
I've added a bunch of common labels to the labeler. I haven't been able to really test it in practice, but I'm ready to merge this. @purificant thoughts?

purificant:
Looks good to me @tomaarsen :+1: Thank you for doing this.
Let's merge, it's safe to do so and we can tweak labels as needed based on what works well / is helpful.

tomaarsen:
Exactly my thinking!

PR #3070, opened 2022-11-07
Resolve `TreebankWordDetokenizer` inconsistency with end-of-string contractions

### LINKED ISSUE #3069: `TreebankWordDetokenizer` inconsistent with recombining contractions
Hello!

## Details
Some quick tests I ran:
```python
from nltk.tokenize import TreebankWordDetokenizer
>>> TreebankWordDetokenizer().tokenize(["Do", "n't"])
"Do n't"
>>> TreebankWordDetokenizer().tokenize(["Do", "n't", "want"])
"Don't want"
>>> TreebankWordDetokenizer().tokenize(["I", "'ve"])  
"I 've"
>>> TreebankWordDetokenizer().tokenize(["I", "'ve", "seen"])
"I've seen"
```
And a quick testing script I wrote:
```python
from nltk.tokenize import TreebankWordDetokenizer, TreebankWordTokenizer

tests = [
    "Don't",
    "Don't.",
    "Don't bother",
    "I've",
    "I've.",
    "I've not",
]

tokenizer = TreebankWordTokenizer()
detokenizer = TreebankWordDetokenizer()

for test in tests:
    tokenized = tokenizer.tokenize(test)
    tetokenized = detokenizer.detokenize(tokenized)
    
    print(f"{test=}")
    print(f"{tokenized=}")
    print(f"{tetokenized=}\n")
```
With the following output (annotated by me):
```python
test="Don't"
tokenized=['Do', "n't"]
tetokenized="Do n't" # Wrong

test="Don't."
tokenized=['Do', "n't", '.']
tetokenized="Don't." # Right

test="Don't bother"
tokenized=['Do', "n't", 'bother']
tetokenized="Don't bother" # Right

test="I've"
tokenized=['I', "'ve"]
tetokenized="I 've" # Wrong

test="I've."
tokenized=['I', "'ve", '.']
tetokenized="I've." # Right

test="I've not"
tokenized=['I', "'ve", 'not']
tetokenized="I've not" # Right
```

- Tom Aarsen

tomaarsen:
Fixes #3069

Hello!

## Pull request overview
* Allow some `TreebankWordDetokenizer` rules to work like intended at the end of the string too.
* Add previously-failing tests for contractions from #3069.

## Details
See the two detokenization rules here.
https://github.com/nltk/nltk/blob/5575c28c86e2e24813bf0e026e84b2235a06f6a0/nltk/tokenize/treebank.py#L288-L290

These will never apply if the string tested for this regex has the contraction at the very end of the string, as the rules each require whitespace after the contracted part. For example, the top rule will find a match in `"it 'll "`, but not in `"it 'll"`.
Now, the requirement for whitespace is important: we don't want to match `I 'read that book`, so there are two fixes that spring to mind:

1. Replace the whitespace at the end of the rules with `( |$)`, allowing the end of text to match too.
2. Pad the input to `detokenize` to always have a space at the start and end of the string.

Because this second approach is used to solve similar issues in the `TreebankWordTokenizer` class (see below), I've opted to go for that approach.
https://github.com/nltk/nltk/blob/5575c28c86e2e24813bf0e026e84b2235a06f6a0/nltk/tokenize/treebank.py#L154-L155

Seems to work like a charm.

### Tests
I've adopted the failing test cases from #3069 into our test suite to cover this fix.

---

- Tom Aarsen

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -132,6 +132,15 @@ Testing treebank's detokenizer
     >>> s = "(Sentence) starting with parentheses."
     >>> detokenizer.detokenize(word_tokenize(s))
     '(Sentence) starting with parentheses.'
+    >>> s = "I've"
+    >>> detokenizer.detokenize(word_tokenize(s))
+    "I've"
+    >>> s = "Don't"
+    >>> detokenizer.detokenize(word_tokenize(s))
+    "Don't"
+    >>> s = "I'd"
+    >>> detokenizer.detokenize(word_tokenize(s))
+    "I'd"
 
 
 Sentence tokenization in word_tokenize:
--- a/nltk/tokenize/treebank.py
+++ b/nltk/tokenize/treebank.py
@@ -356,7 +356,11 @@ class TreebankWordDetokenizer(TokenizerI):
         :type convert_parentheses: bool, optional
         :return: str
         """
-        text = " " + " ".join(tokens) + " "
+        text = " ".join(tokens)
+
+        # Add extra space to make things easier
+        text = " " + text + " "
+
         # Reverse the contractions regexes.
         # Note: CONTRACTIONS4 are not used in tokenization.
         for regexp in self.CONTRACTIONS3:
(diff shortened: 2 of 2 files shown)


stevenbird:
@tomaarsen: nice solution, and good idea to do the padding in a separate step

PR #3078, opened 2022-12-01
Fix wordnet's all_synsets() function

### LINKED ISSUE #3077: Incorrect part-of-speech filtering in Wordnet's all_synsets()
The _all_synsets()_ function in _wordnet.py_ does not behave as expected when using the _pos_ parameter:

1. With the English wordnet, it confuses adjectives and satellites:

```
print(wn.get_version())
```
3.0

`print(len(list(wn.all_synsets(pos="a"))))`
18156

`print(len(list(wn.all_synsets(pos="s"))))`
18156

So, in both cases it returns all the adjectives. This is ok when using pos='a', but when pos='s', the comments inside the function state that only satellites should be returned.

2. With other languages, it does no part-of-speech filtering at all:

```wn.add_omw()
print(len(list(wn.all_synsets(lang="hrv"))))
```
23120
`print(len(list(wn.all_synsets(pos="n", lang="hrv"))))`
23120
`
print(len(list(wn.all_synsets(pos="v", lang="hrv"))))`
23120

ekaf:
Fix #3077, and a few related problems in wordnet.py.

1. With the English Wordnet, properly distinguish satellites from adjectives:

```
print(wn.get_version())
```
3.0

`print(len(list(wn.all_synsets(pos="a"))))`
18156

`print(len(list(wn.all_synsets(pos="s"))))`
10693

2. With other languages, a part of the problem is due to the _synset_from_pos_and_offset()_ function, which returns an empty synset object when an offset is incorrect. This leads to inflated synset counts with OMW for two reasons: OMW does include some incorrect offsets, and with later wordnets some synsets that are lost in mapping should not be counted. So, when a synset does not exist, this PR returns _None_ instead of an empty synset object:

```
ss = wn.synset_from_pos_and_offset('n',1234)
print(ss)
```
None

```
wn.add_omw()
print(len(list(wn.all_synsets(lang="hrv"))))
```
23115
`print(len(list(wn.all_synsets(pos="n", lang="hrv"))))`
16177
`print(len(list(wn.all_synsets(pos="v", lang="hrv"))))`
4736

Another problem is that the handling of satellites in the mappings was inconsistent. The mapping procedure has now been rewritten in a style that makes it easier to follow the underlying algorithm.

```
from nltk.corpus import wordnet31 as wn31
print(wn31.get_version())
```
3.1
```
wn31.add_omw()
print(len(list(wn31.all_synsets(lang="hrv"))))
```
23098
`print(len(list(wn31.all_synsets(pos="a", lang="hrv"))))`
1363
`print(len(list(wn31.all_synsets(pos="s", lang="hrv"))))`
444


nltk/corpus/reader/wordnet.py
```diff
@@ -1203,59 +1208,75 @@ def __init__(self, root, omw_reader):
...
+        self.splits = {}
+
         # map from WordNet 3.0 for OMW data
         self.map30 = self.map_wn30()
 
         # Language data attributes
         self.lg_attrs = ["lemma", "none", "def", "exe"]
 
-    def corpus2sk(self, corpus=None):
+    def indexSense(self, version=None):
```
tomaarsen:
```suggestion
    def index_sense(self, version=None):
```
Could we stick with snake case method names, as suggested by the Python [style guide](https://peps.python.org/pep-0008/#method-names-and-instance-variables). Optionally, we can use a single underscore in front of a method name if it's not meant to be used by the public. This may be the case for e.g. `mapToMany`?

nltk/corpus/reader/wordnet.py
```diff
@@ -1203,59 +1208,75 @@ def __init__(self, root, omw_reader):
...
-                pos = self._pos_names[int(sk.split("%")[1].split(":")[0])]
-                sk_map[sk] = f"{items[1]}-{pos}"
-        return sk_map
+                fields = line.strip().split()
+                sensekey = fields[0]
+                pos = self._pos_names[int(sensekey.split("%")[1].split(":")[0])]
+                sensekeyMap[sensekey] = f"{fields[1]}-{pos}"
+        return sensekeyMap
+
+    def mapToMany(self):
```
tomaarsen:
```suggestion
    def map_to_many(self):
```
Idem. I'd rather also see snake casing for the variables.

nltk/corpus/reader/wordnet.py
```diff
@@ -1203,59 +1208,75 @@ def __init__(self, root, omw_reader):
...
+        synsetToMany = {}
+        for synsetid in set(sensekeyMap1.values()):
+            synsetToMany[synsetid] = []
+        for sensekey in set(sensekeyMap1.keys()).intersection(set(sensekeyMap2.keys())):
+            source = sensekeyMap1[sensekey]
+            target = sensekeyMap2[sensekey]
+            synsetToMany[source].append(target)
+        return synsetToMany
+
+    def mapToOne(self):
```
tomaarsen:
```suggestion
    def map_to_one(self):
```
Idem. I'd rather also see snake casing for the variables.

nltk/corpus/reader/wordnet.py
```diff
@@ -1314,6 +1332,10 @@ def add_provs(self, reader):
                     lang = f"{lang}_{prov}"
                 self.provenances[lang] = prov
 
+    def strip_prov(self, lang):
+        #        return lang[:lang.find("_")+1]
+        return lang.split("_")[0]
+
```
tomaarsen: Is this meant as a public method, or is it just a leftover method that is now no longer used?

tomaarsen: Also, the commented line can be removed.

tomaarsen (changes requested):
I've added some simple tests. I also accidentally brought in 2f34b52 into this PR, so I rebased and removed some of the unnecessary commits. Apologies for this. Make sure to update locally to avoid issues if you make changes!

I had some comments and requests for changes. I hope you'll be able to have a look at them.

nltk/corpus/reader/wordnet.py
ekaf: Thanks @tomaarsen, I have added all your suggestions for the function names, and also changed the variable names.

nltk/corpus/reader/wordnet.py
ekaf: Fixed, I also did the same with variable names. Thanks @tomaarsen!

tomaarsen (approved):
This is looking good! Thanks for always helping maintain Wordnet in NLTK!

PR #3081, opened 2022-12-06
Provide type hints for LazyCorpusLoader variables

tomaarsen:
Hello!

## Pull request overview
* Provide type hints for LazyCorpusLoader variables in order to help IDEs with their suggestions etc.

## Details
All of our corpus readers are initialized using a `LazyCorpusLoader` like so:
https://github.com/nltk/nltk/blob/2f34b52da55a6c83fd18306095c9aa00b4996381/nltk/corpus/__init__.py#L68-L73

Although this `LazyCorpusLoader` approach is very effective to avoid unnecessarily loading data, it does mean that in the current implementation, all of these variables are typed as `LazyCorpusLoader` instances according to typers or IDEs. However, since Python 3.5, we can just override the default typing for these variables, allowing users to take advantage of the very useful method suggestions etc.

Let's take a look at some of the effects of this change.

### Before this PR
image
image

### After this PR
image
image

Beyond these type hinting changes, no functionality is affected by this PR.

- Tom Aarsen

### DIFF
--- a/nltk/corpus/__init__.py
+++ b/nltk/corpus/__init__.py
@@ -65,14 +65,16 @@ from nltk.corpus.reader import *
 from nltk.corpus.util import LazyCorpusLoader
 from nltk.tokenize import RegexpTokenizer
 
-abc = LazyCorpusLoader(
+abc: PlaintextCorpusReader = LazyCorpusLoader(
     "abc",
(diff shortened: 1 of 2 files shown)


stevenbird:
Brilliant!

Issue #2247, opened 2019-03-06
Making dispersion_plot faster / leaner / more readable


dlukes:
I noticed that the data preparation part of `nltk.draw.dispersion_plot` does repeated passes over the data and allocates some of the intermediate results, which seems unnecessary:

https://github.com/nltk/nltk/blob/39d8eeb2f249ad67d11300d97971f972039bdc63/nltk/draw/dispersion.py#L33-L52

I also find the nested listcomp which creates the `points` and then the `list(zip(*...))` idiom hard to read, though I admit this may be subjective. But more to the point, these seem like an unnecessarily convoluted way of going about generating `x` and `y`, which are then used for plotting.

Is there any particular reason for doing it this way that I may be missing/unaware of? If not, would you be generally open to a pull request along the following lines (replacing the snippet above)?

```python
words_dict = {word.casefold() if ignore_case else word: y_i
              for y_i, word in enumerate(reversed(words))}
x, y = [], []
for x_i, token in enumerate(text):
    token = token.casefold() if ignore_case else token
    y_i = words_dict.get(token)
    if y_i is not None:
        x.append(x_i)
        y.append(y_i)
```

Using the following sample data, `%%timeit` tells me it's about 400 ms (former snippet) vs 300 ms (latter snippet), but YMMV of course :)

```python
from nltk.corpus import inaugural
text = inaugural.words()
words = ["America", "duties", "citizen", "values"]
ignore_case = False
```

dlukes:
(I should perhaps clarify: the speedup gain is not my main motivation here, I realize it's not that big of a difference. I initially started digging around this code because I want students to tinker with it in a class setting, so that they get more comfortable reading and customizing existing code, and found the data preparation part of the function somewhat hard to read for no clear benefit to my mind. But perhaps I'm just not seeing it :) )

stevenbird:
Probably too late for you to be useful, sorry. We always welcome pull requests and do our best to process them quickly.

dlukes:
No worries, maybe some other time then :)

PR #3060, opened 2022-10-05
Support both iso639-3 codes and BCP-47 language tags

ekaf:
Fix #3058 by adding a function to obtain language names, using their respective iso639-3 language codes.

This PR adds a langname.py module in the nltk directory. I plan to use it with wordnet.py, to obtain the names of the multilingual wordnets from OMW.

```
>>>from nltk.langnames import langname
>>>for code in ["cat", "wln"]:
>>>    print(f"{code}: {langname(code)}")
```

> cat: Catalan
> wln: Walloon
> 

The language codes were obtained from https://iso639-3.sil.org/code_tables/download_tables
and since it is not clear whether this use is authorized,  it seems safer to ask them:

> For any questions about whether a particular use is covered by these guidelines, contact the Registration Authority at ISO639-3@sil.org.

PR #3082, opened 2022-12-06
Refactor dispersion plot

dlukes:
This is a follow-up to a suggestion I made a while ago in #2247. I use dispersion plots a lot when teaching beginners. Students are often impressed and ask if tweaks can be made to the result. So I introduce them to the `??` IPython magic, which shows them the source code, but we typically stop short of actually playing around with a few tweaks, because:

- The data preparation part is written in a functional style which is a bit inscrutable, it emphasizes Python features which aren't typically the bread and butter of beginner introductions to Python.
- The plotting part uses Matplotlib's state-machine API, instead of the object-oriented one which is currently favored.

With the changes I'm proposing here, my hope is to make dispersion plot more hackable for beginners. I tried breaking up the changes into individual commits to facilitate discussion around different aspects of the issue. I'm aware that people's experience with teaching beginners may vary, so I'm very much open to discussion.

One part I'd definitely like to keep though is that `dispersion_plot` should return the `Axes` object it creates, so that the first tweaking attempts can be made without even looking at the source code. In light of #2239, I hope this is relatively uncontroversial.

### DIFF
--- a/nltk/draw/dispersion.py
+++ b/nltk/draw/dispersion.py
@@ -23,7 +23,7 @@ def dispersion_plot(text, words, ignore_case=False, title="Lexical Dispersion Pl
     """
 
     try:
-        from matplotlib import pylab
+        import matplotlib.pyplot as plt
     except ImportError as e:
         raise ValueError(
             "The plot function requires matplotlib to be installed."
@@ -50,16 +50,20 @@ def dispersion_plot(text, words, ignore_case=False, title="Lexical Dispersion Pl
         x, y = list(zip(*points))
     else:
         x = y = ()
-    pylab.plot(x, y, "b|", scalex=0.1)
-    pylab.yticks(list(range(len(words))), words, color="b")
-    pylab.ylim(-1, len(words))
-    pylab.title(title)
-    pylab.xlabel("Word Offset")
-    pylab.show()
+    _, ax = plt.subplots()
+    ax.plot(x, y, "b|")
+    ax.set_yticks(list(range(len(words))), words, color="b")
+    ax.set_ylim(-1, len(words))
+    ax.set_title(title)
+    ax.set_xlabel("Word Offset")
+    return ax
 
 
 if __name__ == "__main__":
+    import matplotlib.pyplot as plt
+
     from nltk.corpus import gutenberg
 
     words = ["Elinor", "Marianne", "Edward", "Willoughby"]
     dispersion_plot(gutenberg.words("austen-sense.txt"), words)
+    plt.show()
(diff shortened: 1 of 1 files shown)


nltk/draw/dispersion.py
```diff
@@ -15,51 +15,47 @@ def dispersion_plot(text, words, ignore_case=False, title="Lexical Dispersion Pl
     Generate a lexical dispersion plot.
 
     :param text: The source text
-    :type text: list(str) or enum(str)
+    :type text: list(str) or iter(str)
```
dlukes, line 18: `enum` in this context doesn't really make sense to me, but maybe that's my fault?

nltk/draw/dispersion.py
```diff
@@ -15,51 +15,47 @@ def dispersion_plot(text, words, ignore_case=False, title="Lexical Dispersion Pl
...
-        from matplotlib import pylab
+        import matplotlib.pyplot as plt
     except ImportError as e:
         raise ValueError(
             "The plot function requires matplotlib to be installed."
             "See https://matplotlib.org/"
         ) from e
 
-    text = list(text)
-    words.reverse()
```
dlukes, line 34: Reversing in place can lead to surprises when the same list of words is used in multiple dispersion plots (or other places besides a dispersion plot).

nltk/draw/dispersion.py
tomaarsen: No, you're right, `enum` was always an odd choice here, as `text` is immediately converted to a list.

nltk/draw/dispersion.py
tomaarsen: Agreed.

tomaarsen:
This is looking great: a collection of each very reasonable changes. I'll happily merge this when the tests go green. 
I also modified the `ValueError` that is thrown if someone does not have `matplotlib` into an `ImportError`. Furthermore, I added a simplel docstring to show that the function returns an Axes object (now). In doing so I accidentally pulled in some commits from the develop branch, so I had to rebase this PR. My apologies.

Happy teaching!

tomaarsen:
The current test failures are due to an outdated cache of the NLTK data. Just today we added BCP47 via #3060 and the corresponding doctests will fail until the NLTK data cache refreshes. Sadly, I can only control the cache on the develop branch (which has the updated NLTK data cache), and not on any PR.
That said, the tests pass for me locally, so I'll just merge this and make sure they still pass on `develop`.

dlukes:
Thank you for the quick review and merge!

PR #3085, opened 2022-12-13
For FreqDist plotting, no longer show plot by default

### LINKED ISSUE #2788: Only return `ax` object by default in `plot` methods, instead of plotting by default
In the following locations:
* https://github.com/nltk/nltk/blob/f989fe65d421e7ea4d1037a00f07eaeee3ad6a29/nltk/probability.py#L247-L249
* https://github.com/nltk/nltk/blob/f989fe65d421e7ea4d1037a00f07eaeee3ad6a29/nltk/probability.py#L1922-L1932
* https://github.com/nltk/nltk/blob/f989fe65d421e7ea4d1037a00f07eaeee3ad6a29/nltk/text.py#L607

We should by default merely return the matplotlib `ax` object, rather than showing the plot. That way, users can still modify the graph themselves, or save it rather than plotting it. As @iliakur rightfully mentioned in the discussion regarding this issue in https://github.com/nltk/nltk/pull/2786#issuecomment-905283065, a Jupyter notebook would still render the plot even if we only return it. In the future, people will have to invoke `.plot()`, and then perform `plt.show()`, where `plt` imported as `import matplotlib.pyplot as plt`. 

These changes can easily be implemented by setting `show=False` as default in the first two method, and adding `return` to the third method.

This issue is planned to be resolved for the next major release.

- Tom Aarsen

tomaarsen:
Resolves #2788

Hello!

### Pull Request overview
* For `FreqDist` and `ConditionalFreqDist`, only return the ax rather than also showing the plot by default.

### Details
See #2788 for details. Note that the third case mentioned in that issue has already been solved in 8c759729d3a1d03fd724c96c6d82c38382ef2e82.

- Tom Aarsen

### DIFF
--- a/nltk/probability.py
+++ b/nltk/probability.py
@@ -245,7 +245,7 @@ class FreqDist(Counter):
         return self.most_common(1)[0][0]
 
     def plot(
-        self, *args, title="", cumulative=False, percents=False, show=True, **kwargs
+        self, *args, title="", cumulative=False, percents=False, show=False, **kwargs
     ):
         """
         Plot samples from the frequency distribution
@@ -1927,7 +1927,7 @@ class ConditionalFreqDist(defaultdict):
         cumulative=False,
         percents=False,
         conditions=None,
-        show=True,
+        show=False,
         **kwargs,
     ):
         """


stevenbird:
I wish we'd done it this way from the start!

tomaarsen:
Well, the second best time to plant a tree... 😄

Issue #2516, opened 2020-03-18
https://nltk.org SSL_ERROR_BAD_CERT_DOMAIN


jayvdb:
https://nltk.org (https) is a redirect to https://www.nltk.org/ , however it shows a cert error.

The cert is only valid for "shortener.secureserver.net, www.shortener.secureserver.net"

tomaarsen:
I believe the former doesn't work anymore? Regardless, no cert errors to be found anymore. I'll close this.

PR #3087, opened 2022-12-13
Remove //nltk.org links

jayvdb:
Relates to https://github.com/nltk/nltk/issues/2516

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -9,8 +9,8 @@
 Functions to find and load NLTK resource files, such as corpora,
 grammars, and saved processing objects.  Resource files are identified
 using URLs, such as ``nltk:corpora/abc/rural.txt`` or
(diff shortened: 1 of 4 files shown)


tomaarsen:
Hello @jayvdb,

I've had a look at your fixes, and there was one that I was not fully satisfied with: http://nltk.org/sample/toy.cfg to http://www.nltk.org/sample/toy.cfg. Neither seem to work, and thus it would be better to keep it the original way, just in case some archiving solution has kept the original file.
*However*, I know that this file is equivalent to https://github.com/nltk/nltk/blob/develop/nltk/test/toy.cfg. So, I've used that link instead. I've also reinstated an old test that was once removed because the toy.cfg was removed from the website, but now using the (raw) github link.

Thank you for these fixes! I'll merge this once it goes green.

- Tom Aarsen

PR #3098, opened 2022-12-29
Fix a bug that left out the last section/heading.

elespike:
Also refactored the logic a bit.

Here's an alternative using `itertools.groupby`, in case you prefer it:
```python
sections, block = list(), list()
for k, g in groupby(self.parser.parse(stream.read()), key=lambda t: t.level == 0 and t.type == 'heading_open'):
    if k or block:
        block.extend(g)
    if not k and block:
        sections.append(block)
        block = list()
```

### DIFF
--- a/nltk/corpus/reader/markdown.py
+++ b/nltk/corpus/reader/markdown.py
@@ -319,15 +319,17 @@ class CategorizedMarkdownCorpusReader(CategorizedCorpusReader, MarkdownCorpusRea
 
     def section_reader(self, stream):
         section_blocks, block = list(), list()
-        in_heading = False
         for t in self.parser.parse(stream.read()):
             if t.level == 0 and t.type == "heading_open":
-                if block:
+                if not block:
+                    block.append(t)
+                else:
                     section_blocks.append(block)
-                block = list()
-                in_heading = True
-            if in_heading:
+                    block = [t]
+            elif block:
                 block.append(t)
+        if block:
+            section_blocks.append(block)
         return [
             MarkdownSection(
                 block[1].content,
--- a/nltk/test/corpus.doctest
+++ b/nltk/test/corpus.doctest
@@ -1746,6 +1746,146 @@ The Brown Corpus uses the tagged corpus reader:
     >>> brown.tagged_paras()
     [[[('The', 'AT'), ...]], [[('The', 'AT'), ...]], ...]
 
+Categorized Markdown Corpus Reader
+==================================
+
+This corpus reader class provides additional methods to select features
+present in markdown documents.
+
+First, let's make a test corpus:
+
+    >>> root = make_testcorpus(ext='.md',
+    ...     a="""\
+    ...     # Section One
+    ...     Here's the first sentence of section one. Then the second sentence.
+    ...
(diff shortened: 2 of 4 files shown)


tomaarsen:
Hey!

Can we get a simple test for this? I'm assuming that this is reasonably doable. You could put the tests in [nltk/test/corpus.doctest](https://github.com/nltk/nltk/blob/develop/nltk/test/corpus.doctest) or [nltk/test/unit/test_corpora.py](https://github.com/nltk/nltk/blob/develop/nltk/test/unit/test_corpora.py), I'll leave the choice to your judgement.
If needed, you can put a markdown file in [nltk/test/unit/files](https://github.com/nltk/nltk/tree/develop/nltk/test/unit/files) that can be read with the markdown reader.

elespike:
Hi, @tomaarsen 

Thanks for taking a look! No problem, I'll add tests. =)

~Out of curiosity, are the doctests published as documents anywhere? There's a lot of great info in there.~ [found them](https://www.nltk.org/howto.html)!
To confirm: if I choose doctests, is it still ok to add files at `nltk/test/unit/files`, or is that only for unit tests?

tomaarsen:
On further glance, it seems like `nltk/test/unit/files` is currently just used for twitter tests. If it's not too many files that you intend on adding, then you can just add them directly to `nltk/test`. I see there's already some files there for testing, like `floresta.txt`, `onto1.fol` and `sem3.cfg`.

And I see that you found the documents! The current setup is far from ideal to be honest, as there's no separation between documentation and tests, so there are plenty of arbitrary tests in the how-to pages.

elespike:
Understood! I've started writing some documentation-style tests, so far only using inline data as done [here](https://www.nltk.org/howto/corpus.html#plaintext-corpus-reader).

Another question: this `CategorizedMarkdownCorpusReader` class has some external dependencies, all of which are python packages. Should I actually add them to `tox.ini` as mentioned at the very bottom of [this file](https://github.com/nltk/nltk/blob/develop/web/dev/local_testing.rst)?

tomaarsen:
@elespike I'm a bit surprised the CI tests are passing currently. Locally I have issues unless I install these packages:
- ``markdown-it-py``
- ``mdit-py-plugins``
- ``mdit-plain``
- `pyyaml`

I'll add them myself, as this is looking good, and I want to release a new version soon :)

tomaarsen:
I see this is causing some issues in the CI: `mdit-plain` doesn't work with Python 3.7. I want to keep Python 3.7 support for another 6 months, so this is a bit problematic.

elespike:
I'll have a look today. That's my own package and I think it might be just a walrus operator. 😆

elespike:
@tomaarsen, please try again. I verified that `mdit-plain` didn't actually depend on Python 3.8 and uploaded a new release that will install and run on 3.7:
https://pypi.org/project/mdit-plain/1.0.1/ (https://github.com/elespike/mdit_plain/commit/218934eafd04725ecfdbf1026bc44ae4ecd0eedc)

tomaarsen:
Wonderful! Appreciate it :)

Issue #194, opened 2012-01-31
Make cfdist.conditions() unordered


heatherleaf:
The current behaviour of ConditionalFreqDist.conditions() is to return a sorted list. This is unnecessary and slow, so I propose to make it just return the .keys() of the underlying dictionary.

This means that the book must be changed where it says that the list is sorted.

PR #3126, opened 2023-02-25
Avoid duplicate merged OMW synsets and lemmas

### LINKED ISSUE #3125: Duplicate merged synsets and lemmas with OMW
The OMW data packages include thousands of duplicate lemmas (https://github.com/nltk/nltk_data/issues/194#issue-1614884960). That problem may be fixed in future data releases, but even then, NLTK will still need to perform a check for duplicate synsets and lemmas at load time.

The reason is that when using OMW together with newer Wordnet versions, the situation changes because a few synsets have been merged. When merging two synsets that both contain a sense of the same word, these two senses are collapsed into one. But without a check for duplicates at load time, the same sense can be appended several times to the list of  lemmas or synsets.

So, for example, each of the following calls outputs duplicates:

```
from nltk.corpus import wordnet2021 as wn

finssl = wn.synsets('baseball', lang="fin")
print(finssl)
```
[Synset('baseball.n.01'), Synset('baseball.n.01'), Synset('hardball.n.02'), Synset('two-hitter.n.01')]

```
ss = finssl[0]
print(ss.lemmas(lang="fin"))
```
[Lemma('baseball.n.01.baseball'), Lemma('baseball.n.01.baseball-peli'), Lemma('baseball.n.01.baseball')]

`print(ss.lemma_names(lang="fin"))`
['baseball', 'baseball-peli', 'baseball']

`print(wn.lemmas('baseball', lang="fin"))`
 [Lemma('baseball.n.01.baseball'), Lemma('baseball.n.01.baseball'), Lemma('baseball.n.01.baseball'), Lemma('baseball.n.01.baseball'), Lemma('hardball.n.02.baseball'), Lemma('two-hitter.n.01.baseball')]

Fortunately, this problem is easily fixed with a simple check at load time.

ekaf:
Fix #3125 and [nltk_data #194](https://github.com/nltk/nltk_data/issues/194) with a simple check for duplicates when loading the OMW data. Then, none of the following calls outputs any duplicate.

```
from nltk.corpus import wordnet2021 as wn

finssl = wn.synsets('baseball', lang="fin")
print(finssl)
```
 [Synset('baseball.n.01'), Synset('hardball.n.02'), Synset('two-hitter.n.01')]

```
ss = finssl[0]
print(ss.lemmas(lang="fin"))
```
[Lemma('baseball.n.01.baseball'), Lemma('baseball.n.01.baseball-peli')]

```print(ss.lemma_names(lang="fin"))```
['baseball', 'baseball-peli']

```print(wn.lemmas('baseball', lang="fin"))```
[Lemma('baseball.n.01.baseball'), Lemma('hardball.n.02.baseball'), Lemma('two-hitter.n.01.baseball')]

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -2247,11 +2247,15 @@ class WordNetCorpusReader(CorpusReader):
                 if len(pair) == 1 or pair[0] == lg:
                     if attr == "lemma":
                         val = val.strip().replace(" ", "_")
-                        self._lang_data[lang][1][val.lower()].append(offset_pos)
+                        lang_offsets = self._lang_data[lang][1][val.lower()]
+                        if offset_pos not in lang_offsets:
+                            lang_offsets.append(offset_pos)
                     if attr in self.lg_attrs:
-                        self._lang_data[lang][self.lg_attrs.index(attr)][
+                        lang_lemmas = self._lang_data[lang][self.lg_attrs.index(attr)][
                             offset_pos
-                        ].append(val)
+                        ]
+                        if val not in lang_lemmas:
+                            lang_lemmas.append(val)
 
     def disable_custom_lemmas(self, lang):
         """prevent synsets from being mistakenly added"""
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -1216,7 +1216,7 @@ class WordNetCorpusReader(CorpusReader):
         self.map30 = self.map_wn()
 
         # Language data attributes
-        self.lg_attrs = ["lemma", "none", "def", "exe"]
+        self.lg_attrs = ["lemma", "of", "def", "exe"]
 
     def index_sense(self, version=None):
         """Read sense key to synset id mapping from index.sense file in corpus directory"""
@@ -1251,7 +1251,7 @@ class WordNetCorpusReader(CorpusReader):
         return synset_to_many
 
     def map_to_one(self, version="wordnet"):
-        self.nomap[version] = []
+        self.nomap[version] = set()
         self.splits[version] = {}
         synset_to_many = self.map_to_many(version)
         synset_to_one = {}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by stevenbird** [comment]
@ekaf I'm wondering what this linear check does for loading time. If the lists are long, using sets would speed things up as the membership test for sets is log time (and built in).

**Comment by ekaf** [comment]
@stevenbird, yes, sets are quicker in theory, but would change the ordering of lemmas and synsets in the output, so several tests would fail. 
In practice, the speed difference is not substantial here, since most lists have only one element:

```
# Average list length in Finnish lang data:

l = []
for dic in wn._lang_data['fin']:
    for key in dic:
        l.append(len(dic[key]))

print(f"Mean: {round(sum(l)/len(l),2)}")
print(f"Median: {sorted(l)[len(l)//2]}")
```

> Mean: 1.54
> Median: 1

**Comment by ekaf** [comment]
The latest CI test fails only with Python 3.11 on macos-latest. This seems to happen sometimes on that platform:

 /Users/runner/work/_temp/ddff5c55-ad4f-4860-82d4-1c5a15b52f49.sh: line 1: pytest: command not found

**Comment by ekaf** [comment]
Casing variations in the source OMW databases can be an additional reason for duplicates in NLTK output. Consider for ex. these entries in the French source:

```
00006024-n	fra:lemma	hétérotrophie
00006024-n	fra:lemma	Hétérotrophie
```
In NLTK's synset identifiers, the lemma names are converted to lowercase, yielding duplicates for the following query:
  
```
print(wn.synsets("hétérotrophie", lang="fra"))

```
[Synset('heterotroph.n.01'), Synset('heterotroph.n.01')]

French common nouns should not have uppercase, so in this specific example, the database is wrong. But often, two casing variants are allowable, so in general, this is not a database problem. Then it has to be solved in the _wordnet_ library, as with the present PR, which avoids the above duplicate output:

[Synset('heterotroph.n.01')]

**Comment by ekaf** [comment]
@stevenbird 

> I'm wondering what this linear check does for loading time.

```
from time import time

wn.add_omw()
lgs = set(wn.langs()) - {'eng'}
n_lgs = len(lgs)

t1 = time()
for lg in lgs:
    wn._load_lang_data(lg)
t2 = time()

print(f"Load time for {n_lgs} languages: {t2-t1}")
```


Without this PR:
Load time for 31 languages: 13.223822832107544

With this PR:
Load time for 31 languages: 13.352472066879272

stevenbird:
Thanks @ekaf. Sorry for the delay!

Issue #2825, opened 2021-09-27
Expand ~ in paths


stevenbird:
From #2820 (@tomaarsen)

Expand ~ in paths used in environment variables to the home path of the user's operating system. This is as simple as wrapping a path with path = os.path.expanduser(path). This probably ought to be done in some functions in internals.py. However, files on linux are allowed to have ~ in them. In some of those cases, you wouldn't want to expand the ~.

tomaarsen:
This involves changes to https://github.com/nltk/nltk/blob/develop/nltk/internals.py, so users can use e.g. `~/tools/parser.jar` as an environment variable for some NLTK tool.

Miguelvs23:
Hi, it is my first time contributing to an OSS, can I work on this task? @tomaarsen can you help me get started?

PR #3128, opened 2023-03-04
Expand some userpaths in internals.py #2825

ethanknights:
Hi - I added some of the path expansions to internals.py as described in #2825.

The tox command seems ok (output below).

```
================= 488 passed, 34 skipped, 9 xfailed, 1 warning in 249.04s (0:04:09) ==================
.pkg: _exit nltk/test> python /Users/ethanknights/PycharmProjects/nltk/venv/lib/python3.9/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  python3.7: OK (273.95=setup[17.25]+cmd[1.94,254.76] seconds)
  congratulations :) (274.45 seconds)
```

Note - I tried expanding the remaining paths with a few different approaches. e.g., Adding a utility function that converted tuples to lists and expanded each of their items, like this:

```
file_names = expand_list_of_userpaths(file_names)
searchpath = expand_list_of_userpaths(searchpath)
``` 

But this wasn't a robust solution (particularly for the `filename` variable), so these paths remain unchanged for now.  
In any-case, hopefully this PR is a start!

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -540,6 +540,7 @@ def find_file_iter(
 
     # Check environment variables
     for env_var in env_vars:
+        env_var = os.path.expanduser(env_var)
         if env_var in os.environ:
             if finding_dir:  # This is to file a directory instead of file
                 yielded = True
@@ -1086,12 +1087,12 @@ def slice_bounds(sequence, slice_obj, allow_step=False):
 
 def is_writable(path):
     # Ensure that it exists.
-    if not os.path.exists(path):
+    if not os.path.exists(os.path.expanduser(path)):
         return False
 
     # If we're on a posix system, check its permissions.
     if hasattr(os, "getuid"):
-        statdata = os.stat(path)
+        statdata = os.stat(os.path.expanduser(path))
         perm = stat.S_IMODE(statdata.st_mode)
         # is it world-writable?
         if perm & 0o002:
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -540,7 +540,6 @@ def find_file_iter(
 
     # Check environment variables
     for env_var in env_vars:
-        env_var = os.path.expanduser(env_var)
         if env_var in os.environ:
             if finding_dir:  # This is to file a directory instead of file
                 yielded = True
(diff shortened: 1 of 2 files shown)


nltk/internals.py
```diff
@@ -540,6 +540,7 @@ def find_file_iter(
 
     # Check environment variables
     for env_var in env_vars:
+        env_var = os.path.expanduser(env_var)
```
tomaarsen: `env_var` is just a string like `"HUNPOS_TAGGER"`, e.g.:
https://github.com/nltk/nltk/blob/cbe715ae5b0f100a76a27fc7e035058ddffad114/nltk/tag/hunpos.py#L94
https://github.com/nltk/nltk/blob/cbe715ae5b0f100a76a27fc7e035058ddffad114/nltk/tokenize/stanford_segmenter.py#L157
https://github.com/nltk/nltk/blob/cbe715ae5b0f100a76a27fc7e035058ddffad114/nltk/parse/malt.py#L91

So, we don't want to expand these.

tomaarsen (changes requested):
Other than the one comment I made, the changes seem reasonable, but do you know if it would be possible to make tests for them? This may not be feasible (as the goal is to have the tests pass locally and also on the CI), but it may be worthwhile.
For example, having an environment variable that points to "~\test_dir", which is then used by `find_file` (or `find_file`, `find_binary`, `find_jar_iter`, etc.):
https://github.com/nltk/nltk/blob/5ebdde6cb20977fe77cf4d5e821bad397230df34/nltk/internals.py#L624

Perhaps this is possible by creating a temporary file/folder for a test? Maybe there's some permission issues with that, though...

nltk/internals.py
ethanknights: Got it, thanks. Reverted this change in 9797986b6b76aaf2d724cfedfef496ba4d6a2b15.

ethanknights:
Hey @tomaarsen - apologies for the piecemeal approach.

I've made the requested change in 9797986b6b76aaf2d724cfedfef496ba4d6a2b15. Feel free to review/mark as resolved.

As for the tests - I agree with the sentiment but am conscious of creating files in a users' home directory (e.g., we need to ensure we are creating/deleting an obscure directory/filename (or within nltk-data/ etc.) as to not interfere with user setups). 

Happy to come back to adding the doctest, after more thought - I can create an issue about this now?

PR #3132, opened 2023-03-12
Replace deprecated error with BadZipFile

hugovk:
`zipfile`'s `error` have been deprecated since Python 3.2, use `BadZipFile` instead, added in 3.2.

* https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile
* https://github.com/python/cpython/blob/6829b1256a596301f82e3573214230186a4babfd/Lib/zipfile.py#L53
* https://github.com/python/cpython/issues/86437


tomaarsen:
Looks good! Thank you for providing those links alongside this PR, and thank you for finding this issue.

Issue #3133, opened 2023-03-15
dispersion plot not working properly


Apros7:
Dispersion plot makes obvious mistakes.
To replicate this error do as follows:
```
import nltk
from nltk.books import *
text7.dispersion_plot(["chairman", "director", "money", "stocks", "old"])
text7.dispersion_plot(["chairman", "director", "money", "stocks", "old", "he"])
plt.show()
```

Looking at these plots, it should be clear, that the plots are not correct. All yticks change simply because "he" is introduced, although all distributions should stay the same, and "he" should have new distribution.
This error is persistent throughout other texts and other words.

The y-ticks are added in the opposite way of the matching distributions.

Apros7:
This can be fixed by changing line 49 in dispersion.py in nltk/nltk/draw/ to:

```
ax.set_yticks(list(reversed(range(len(words)))), words, color="C0")
```

A pull request has been opened with this change.

cchatel2500:
In fact the order of the labels is reversed. The problem is still not corrected and neither is the web page ("https://www.nltk.org/_modules/nltk/draw/dispersion.html")     ;-)

for y, word in enumerate(reversed(words))
<>
ax.set_yticks(list(range(len(words))), words, color="C0")

furas:
it needs to reverse `labels` using `reversed(words)` in `set_yticks()` in `dispersion_plot()`

```
ax.set_yticks(list(range(len(words))), reversed(words), color="C0")
```

At this moments you can use `set_yticks()` directly in your code after running `dispersion_plot()` and before `plt.show()`

```
ax = dispersion_plot(text, words)
ax.set_yticks(list(range(len(words))), reversed(words), color="C0")
plt.show()
```

---

I found this problem in question on `Stackoverflow` and I described solution in answer.

[Python NLTK text dispersion plot has y vertical axis is in backwards / reversed order - Stack Overflow](https://stackoverflow.com/questions/77262318/python-nltk-text-dispersion-plot-has-y-vertical-axis-is-in-backwards-reversed/77265314#77265314)

Apros7:
You're right. I made a PE with this a long time ago, but hasn't been commented on or approved. Still waiting :(...

PR #3134, opened 2023-03-15
dispersion plot not working properly

Apros7:
This is a fix to issue #3133.

Dispersion plot makes obvious mistakes.
To replicate this error do as follows:

```
import nltk
from nltk.books import *
text7.dispersion_plot(["chairman", "director", "money", "stocks", "old"])
text7.dispersion_plot(["chairman", "director", "money", "stocks", "old", "he"])
plt.show()
```

Looking at these plots, it should be clear, that the plots are not correct. All yticks change simply because "he" is introduced, although all distributions should stay the same, and "he" should have new distribution.
This error is persistent throughout other texts and other words.

The y-ticks are added in the opposite way of the matching distributions.

The same plots can be created after this change to see that the distributions now match the correct ytick.
Let me know I need to provide any other information.

### DIFF
--- a/nltk/draw/dispersion.py
+++ b/nltk/draw/dispersion.py
@@ -4,6 +4,7 @@
 # Author: Steven Bird <stevenbird1@gmail.com>
 # URL: <https://www.nltk.org/>
 # For license information, see LICENSE.TXT
+# See issue #3133
 
 """
 A utility for displaying lexical dispersion.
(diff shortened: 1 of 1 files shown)


rmalouf:
Is this the same bug as fixed by #3150?

Apros7:
It is, thank you!! will close this one :D

PR #3142, opened 2023-04-16
Added new custom token function

vinitshah101:
Added function for create custom token as per given list of custom token.
So If anyone wants some custom token they can create by just giving list else remain same.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -296,6 +296,7 @@
 - M.K. Pawelkiewicz <https://github.com/hamiltonianflow>
 - Steven Thomas Smith <https://github.com/essandess>
 - Jan Lennartz <https://github.com/Madnex>
+- Vinit Shah <https://github.com/GharudxD>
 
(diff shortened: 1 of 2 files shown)


stevenbird:
@GharudxD please make the case for this functionality being added to NLTK. There needs to be some existing demand otherwise we cannot justify adding to the codebase.

vinitshah101:
> @GharudxD please make the case for this functionality being added to NLTK. There needs to be some existing demand otherwise we cannot justify adding to the codebase.

Yap, you are right we can't add all things in codebase but I thought sometimes when we tokenize sometimes all word didn't tokenize based on space. Some words treated as token like full forms organisation name etc, for this reasons I created this function.I thought if I add this feature in nltk peoples haven't do this manually.

stevenbird:
Sorry @GharudxD, this is too obscure. A better work-around is to pre-process the problematic input.

PR #3145, opened 2023-04-21
Set reachable depth for generate

### LINKED ISSUE #3072: `generate` gives infinite recursion for recursive grammars
The following code for generating simple recursive grammar does not work

```python
from nltk.grammar import Nonterminal, Production, CFG
from nltk.parse.generate import generate


S = Nonterminal('S')
P = [Production(S, ['a', S]), Production(S, [])]
G = CFG(S, P)

gen = generate(G)
print(next(gen))
```
gives `RuntimeError: The grammar has rule(s) that yield infinite recursion!`

I think that the problem should not be an issue if the algorithm uses different searching technique
and iterators should be capable of generating infinite sequences.
Without the capability to generate recursive grammars, it loses many practical capability because it can only handle finite grammars.

ekaf:
Fix #3072: when a grammar is recursive, the default generation depth (_sys.maxsize_) is wildly out of reach.

```
import sys
print(sys.maxsize)
```

9223372036854775807

`print(sys.getrecursionlimit())`
1000

So this PR adds a test to check if the grammar is recursive, in which case the default "depth" is lowered to a safe value.
Because of indirect recursion between generate_all() and generate_one(), the depth cannot exceed one third of the recursion limit. Additionally, Python 3 has the undocumented peculiarity that the actually reachable recursion limit is 3 less than sys.getrecursionlimit().

With this PR, generating from the grammar in #3072 no longer raises any error:

```
from nltk.grammar import CFG
from nltk.parse.generate import generate

G = CFG.fromstring("""
S -> 'a' S | 
""")

gen = generate(G)
out = next(gen)
print(len(out))
```
329 

```print(''.join(out))```

aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

### DIFF
--- a/nltk/grammar.py
+++ b/nltk/grammar.py
@@ -6,6 +6,7 @@
 #         Jason Narad <jason.narad@gmail.com>
 #         Peter Ljunglöf <peter.ljunglof@heatherleaf.se>
 #         Tom Aarsen <>
+#         Eric Kafe <kafe.eric@gmail.com>
 # URL: <https://www.nltk.org/>
 # For license information, see LICENSE.TXT
 #
@@ -303,6 +304,9 @@ class Production:
         :rtype: sequence(Nonterminal and terminal)
         """
         return self._rhs
+
+    def is_recursive(self):
+        return self.lhs() in self.rhs()
 
     def __len__(self):
         """
@@ -604,6 +608,12 @@ class CFG:
                 for prod in self._lhs_index.get(lhs, [])
                 if prod in self._rhs_index.get(rhs, [])
             ]
+
+    def is_recursive(self):
+        for prod in self.productions():
+            if prod.is_recursive():
+                return True
+        return False
 
     def leftcorners(self, cat):
         """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ekaf** [comment]
Here's a small experiment to verify the unintuitive behaviour of the recursion limit in Python 3:

```
from sys import setrecursionlimit
setrecursionlimit(10)

def recurse(n):
    print(n)
    recurse(n+1)

recurse(1)
```

1
2
3
4
5
6
7

After printing _7_, Python 3 raises a RecursionError. But with Python 2, the same code is able to reach _9_.

stevenbird:
I suggest dropping the recursion check, and setting the parameter regardless of whether the grammar is recursive

ekaf:
Strangely, CI failed only with Python 3.11 on macos-latest.

PR #3148, opened 2023-04-29
added the option to change the wordnet's language

TiMauzi:
`wsd.py` didn't have the option to use a language other than English. I added the option to change the language using the respective `wn` parameter.


stevenbird:
Thanks @TiMauzi

PR #3150, opened 2023-05-04
Fix order of words in y-axis of dispersion_plot

chbrandt:
Input words are reversed to create 'y' ticks but the internal 'words' list is not updated accordingly.
This has been fixed by getting 'words' back from 'word2y', which is sensitive to 'ignore_case'.
If 'words' should not be sensitive to 'ignore_case', then the following should be used instead:
```
# do not modify input 'words' list (as used to be the case with .reverse())
words = words[::-1]
```

### DIFF
--- a/nltk/draw/dispersion.py
+++ b/nltk/draw/dispersion.py
@@ -44,6 +44,8 @@ def dispersion_plot(text, words, ignore_case=False, title="Lexical Dispersion Pl
             xs.append(x)
             ys.append(y)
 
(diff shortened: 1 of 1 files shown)


chbrandt:
The two failures are in nltk downloading datasets, maybe some network connection issue?

tomaarsen:
Looks good! Thanks for this. And I must say that I don't quite know why those two test runs are failing, but I agree that they must be unrelated.

PR #3153, opened 2023-05-11
Update punkt.py

VirgisM:
When training the tokenizer to don't save learned parameters. This not match  the function description
     "Derives parameters from a given training text, or uses the parameters given. Repeated calls to this method destroy previous parameters. For incremental training, instantiate a separate PunktTrainer instance"

To match it this change is necessary.

### DIFF
--- a/nltk/tokenize/punkt.py
+++ b/nltk/tokenize/punkt.py
@@ -1266,9 +1266,10 @@ class PunktSentenceTokenizer(PunktBaseClass, TokenizerI):
         """
         if not isinstance(train_text, str):
             return train_text
-        return PunktTrainer(
+        self._params = PunktTrainer(
             train_text, lang_vars=self._lang_vars, token_cls=self._Token
         ).get_params()
+        return self._params
 
     # ////////////////////////////////////////////////////////////
     # { Tokenization


stevenbird:
@VirgisM: I think the existing implementation behaves as the documentation says, no?

nltk/tokenize/punkt.py
```diff
@@ -1266,9 +1266,10 @@ def train(self, train_text, verbose=False):
         """
         if not isinstance(train_text, str):
             return train_text
-        return PunktTrainer(
+        self._params = PunktTrainer(
             train_text, lang_vars=self._lang_vars, token_cls=self._Token
         ).get_params()
+        return self._params
```
alvations, line 1272: The `PunktTrainer.train` returns the parameters of type `PunktParameters` through the `.get_params()` function (which returns `self._params` if training is not `self._finalized`)

And in `finalize_training()`, the `_params` collocations and sent_starters will be cleared. 


By default after passing the `train_text` through `PunktTrainer.train`, the training will finalize. To disable that beahvior you'll have to do something like:

```
x = PunktTrainer()
x._finalized = False
x.train(train_text)

# Continue training on 2nd text.
x.train(train_text_n)

# Then when you want to finalize and get the parameters, after n-1 texts
x._finalized = True
x.train(train_text_n)
```

alvations: Doing this will finalize the training at the first `train_text` pass

```
x = PunktTrainer(train_text)
```

ekaf:
This pull request seems ripe for closing, as it appears to be resolved or has become stale. If further action is required, please let us know.

PR #3158, opened 2023-05-23
Make sure dispersion plot spans full text length

dlukes:
By default, Matplotlib autoscales the x-axis based on where hits are found. This may result in an unintuitive visualization of the dispersion which "zooms in" on the portion of the text with hits, failing to represent the preceding and/or following stretches with no hits:

```python
from nltk.draw import dispersion_plot
text = ["foo"] * 10_000 + (["bar"] + ["foo"] * 100) * 5 + ["foo"] * 10_000
ax = dispersion_plot(text, ["bar"])
```

image

A more intuitive visualization is one where the x-axis always spans the full length of the text. The most straightforward way to achieve this is via `ax.set_xlim(0, len(text) - 1)`, but since this disables autoscaling, it unfortunately also gets rid of the margins added during autoscaling, which add aesthetically pleasing breathing room to the plot:

```python
ax.set_xlim(0, len(text) - 1)
```

image

So instead, we can explicitly set the data limits via `ax.dataLim`, and re-apply autoscaling:

```python
ax.dataLim.x0, ax.dataLim.x1 = 0, len(text) - 1
ax.autoscale(axis="x")
```

image

This is the solution suggested by this PR.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -110,6 +110,7 @@
 - Tom Lippincott
 - Peter Ljunglöf
 - Alex Louden
+- David Lukeš
 - Joseph Lynch
 - Nitin Madnani
 - Felipe Madrigal
--- a/nltk/draw/dispersion.py
+++ b/nltk/draw/dispersion.py
@@ -48,6 +48,8 @@ def dispersion_plot(text, words, ignore_case=False, title="Lexical Dispersion Pl
 
     _, ax = plt.subplots()
     ax.plot(xs, ys, "|")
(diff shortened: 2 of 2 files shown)


dlukes:
I don't think I'm on the hook for the CI failure with Python 3.11 on macos-latest:

```
/Users/runner/work/_temp/46fec604-9745-4658-9616-404ec3453b66.sh: line 1: pytest: command not found
Error: Process completed with exit code 127.
```

But please let me know if I'm misinterpreting this! :)

stevenbird:
Thanks @dlukes. Sorry for the delay

Issue #3083, opened 2022-12-13
Improve performance of nltk.ngrams


BLKSerene:
With the current implementation of `nltk.ngrams`, the performance decreases slightly when the size of n-grams increases:

```python
>>> timeit.timeit('''nltk.ngrams(tokens, 1)''', setup='import nltk; tokens = list(range(1000))')
0.6647043000000394
>>> timeit.timeit('''nltk.ngrams(tokens, 2)''', setup='import nltk; tokens = list(range(1000))')
0.9033590999999888
>>> timeit.timeit('''nltk.ngrams(tokens, 3)''', setup='import nltk; tokens = list(range(1000))')
1.1749377000001004
>>> timeit.timeit('''nltk.ngrams(tokens, 4)''', setup='import nltk; tokens = list(range(1000))')
1.4639482000000044
>>> timeit.timeit('''nltk.ngrams(tokens, 5)''', setup='import nltk; tokens = list(range(1000))')
1.787133600000061
```

Though this is the expected behavior, perhaps it could be improved further. There is a function [`sliding_window`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.sliding_window) from [more-itertools](https://github.com/more-itertools/more-itertools) which have seemingly implemented the same function as `nltk.ngrams`:

```python
>>> timeit.timeit('more_itertools.sliding_window(tokens, 1)', setup='import more_itertools; tokens = list(range(1000))')
0.1839231999997537
>>> timeit.timeit('more_itertools.sliding_window(tokens, 2)', setup='import more_itertools; tokens = list(range(1000))')
0.188646499999777
>>> timeit.timeit('more_itertools.sliding_window(tokens, 3)', setup='import more_itertools; tokens = list(range(1000))')
0.17990640000016356
>>> timeit.timeit('more_itertools.sliding_window(tokens, 4)', setup='import more_itertools; tokens = list(range(1000))')
0.18609590000005483
>>> timeit.timeit('more_itertools.sliding_window(tokens, 5)', setup='import more_itertools; tokens = list(range(1000))')
0.19976090000000113
```

This implementation is faster, and it seems that the performance does not decrease when the size of n-grams increases. I suppose that this would be a better alternative to the current implementation, but I'm not sure about the compatibility with other optional parameters (e.g. pad_left, pad_right) and the license issue (Apache vs. MIT).

If the maintainers do not have time, I could work on this.

[...]

PR #3162, opened 2023-06-09
Use efficient ngrams implementation from python docs

rmalouf:
This resolves issue #3083. The more-itertools implementation is straight from the [python docs](https://docs.python.org/3/library/itertools.html?highlight=sliding_window#itertools-recipes), so there shouldn't be any problem with the license.


stevenbird:
Thanks @rmalouf. Sorry for the long delay.

PR #3172, opened 2023-06-28
Align text.ConcordanceIndex.find_concordance()

BroMattMiller:
There were two cases where the alignment of
text.ConcordanceIndex.find_concordance() output was broken:

* The left context of the query was very small. That is, the query matched very close to the beginning of the text.
* The text or query contained Unicode combining characters.

The first case is addressed by right justifying the left context to the proper width. The second case is addressed by replacing the standard Python len(s) with a count that excludes combining characters, and replacing s[:width] and s[-width:] with a new utility function that accounts for combining characters.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -127,6 +127,7 @@
 - David McClosky
 - Xinfan Meng
 - Dmitrijs Milajevs
+- Matt Miller
 - Margaret Mitchell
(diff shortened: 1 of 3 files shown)


stevenbird:
Thanks @BroMattMiller. Sorry for the delay!

PR #3174, opened 2023-07-11
Remove redundant function call (demo())

cootshk:
keeping demo() for backwards-compatibility reasons


stevenbird:
Thanks @cootshk

PR #3180, opened 2023-08-02
Bug on edit distance align

yzhaoinuw:
This pull request attempts to fix the issue raised by @yzhangcs's [comment](https://github.com/nltk/nltk/pull/3017#issuecomment-1557141556) as a follow-up to #3017. In short, when the substitution cost is greater than 2, the path generated by [`_edit_dist_backtrace()`](https://github.com/nltk/nltk/blob/develop/nltk/metrics/distance.py#L126) still includes substitution. This is incorrect because a deletion followed by an insertion incur a cost of 2, which is lower than substitution, and thus should be preferred over substitution.

In the following test case due to @yzhangcs, the current [`_edit_dist_backtrace()`](https://github.com/nltk/nltk/blob/develop/nltk/metrics/distance.py#L126) gives the following output
```python
from nltk.metrics.distance import edit_distance_align

>>> edit_distance_align('a', 'b', float('inf')) 
[(0, 0), (1, 1)]
```
After the fix it correctly outputs `[(0, 0), (0, 1), (1, 1)]`. I tried to add the test case provided by @yzhangcs, but didn't figure out how. Could the maintainers of NLTK look into this and add this test case?

### DIFF
--- a/nltk/metrics/distance.py
+++ b/nltk/metrics/distance.py
@@ -123,24 +123,35 @@ def edit_distance(s1, s2, substitution_cost=1, transpositions=False):
     return lev[len1][len2]
 
 
-def _edit_dist_backtrace(lev):
+def _edit_dist_backtrace(lev, s1, s2, substitution_cost=1):
     i, j = len(lev) - 1, len(lev[0]) - 1
     alignment = [(i, j)]
 
     while (i, j) != (0, 0):
         directions = [
-            (i - 1, j - 1),  # substitution
+            (i - 1, j - 1),  # substitution / match
             (i - 1, j),  # skip s1
             (i, j - 1),  # skip s2
         ]
 
-        direction_costs = (
-            (lev[i][j] if (i >= 0 and j >= 0) else float("inf"), (i, j))
-            for i, j in directions
-        )
+        direction_costs = []
+        for pi, pj in directions:
+            if pi < 0 or pj < 0:
+                cost = float("inf")
+            elif pi == i - 1 and pj == j - 1:  # diagonal
+                # Use actual transition cost: 0 for match, substitution_cost
+                # for mismatch. This ensures the backtrace prefers delete+insert
+                # (cost 2) over substitution when substitution_cost > 2.
+                sub_cost = 0 if s1[pi] == s2[pj] else substitution_cost
+                cost = lev[pi][pj] + sub_cost
+            else:  # skip s1 or skip s2
+                cost = lev[pi][pj] + 1
+            direction_costs.append((cost, (pi, pj)))
+
         _, (i, j) = min(direction_costs, key=operator.itemgetter(0))
 
         alignment.append((i, j))
+
     return list(reversed(alignment))
 
 
@@ -194,7 +205,7 @@ def edit_distance_align(s1, s2, substitution_cost=1):
             )
 
     # backtrace to find alignment
-    alignment = _edit_dist_backtrace(lev)
+    alignment = _edit_dist_backtrace(lev, s1, s2, substitution_cost)
     return alignment
 
 
--- a/nltk/test/unit/test_distance.py
+++ b/nltk/test/unit/test_distance.py
@@ -4,6 +4,7 @@ import pytest
 
 from nltk.metrics.distance import (
     edit_distance,
+    edit_distance_align,
     jaro_similarity,
     jaro_winkler_similarity,
 )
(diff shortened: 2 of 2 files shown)


alvations:
@yzhaoinuw @yzhangcs added the substitution cost feature. LGTM now!

### Problem

When `substitution_cost > 2`, the backtrace in `_edit_dist_backtrace()` incorrectly followed diagonal (substitution) moves because it only compared raw `lev`
matrix values without considering the actual transition cost.

### Before fix (develop branch - BUGGY)

```python
>>> from nltk.metrics.distance import edit_distance_align

>>> edit_distance_align('a', 'b', float('inf'))
[(0, 0), (1, 1)]
# BUG: (0,0)->(1,1) is a substitution, but substitution cost is inf!

>>> edit_distance_align('ab', 'cb', float('inf'))
[(0, 0), (1, 1), (2, 2)]
# BUG: (0,0)->(1,1) substitutes a->c, cost should be inf!
```

### After fix

```python
>>> from nltk.metrics.distance import edit_distance_align

>>> edit_distance_align('a', 'b', float('inf'))
[(0, 0), (0, 1), (1, 1)]
# FIXED: insert b, then delete a (cost 2)

>>> edit_distance_align('ab', 'cb', float('inf'))
[(0, 0), (0, 1), (1, 1), (2, 2)]
# FIXED: del a, ins c, match b (cost 2)

>>> edit_distance_align('abc', 'abc', float('inf'))
[(0, 0), (1, 1), (2, 2), (3, 3)]
# Still correct: all matches (cost 0)

>>> edit_distance_align('rain', 'shine')
[(0, 0), (1, 1), (2, 2), (3, 3), (4, 4), (4, 5)]
# Default sub_cost=1 behavior unchanged
```

### After fix

```python
>>> from nltk.metrics.distance import edit_distance_align

>>> edit_distance_align('a', 'b', float('inf'))
[(0, 0), (0, 1), (1, 1)]
# FIXED: insert b, then delete a (cost 2)

>>> edit_distance_align('ab', 'cb', float('inf'))
[(0, 0), (0, 1), (1, 1), (2, 2)]
# FIXED: del a, ins c, match b (cost 2)

>>> edit_distance_align('abc', 'abc', float('inf'))
[(0, 0), (1, 1), (2, 2), (3, 3)]
# Still correct: all matches (cost 0)

>>> edit_distance_align('rain', 'shine')
[(0, 0), (1, 1), (2, 2), (3, 3), (4, 4), (4, 5)]
# Default sub_cost=1 behavior unchanged
```

### Root cause

The original backtrace picked the candidate with the lowest `lev[candidate]` value. For `'a'` vs `'b'` with `sub_cost=inf`:

```
lev = [[0, 1],
     [1, 2]]
```

At `(1,1)`, the candidates were:
- Diagonal `(0,0)`: `lev[0][0] = 0` ← **picked because lowest value**
- Up `(0,1)`: `lev[0][1] = 1`
- Left `(1,0)`: `lev[1][0] = 1`

But the actual cost of transitioning `(0,0) → (1,1)` via substitution is `inf` (since `'a' != 'b'`). The backtrace didn't account for this.

### Fix

Pass `s1` and `s2` to `_edit_dist_backtrace` so it computes the **actual transition cost** for each candidate:
- **Diagonal**: `0` if characters match, `substitution_cost` if they don't
- **Up/Left**: always `1` (deletion/insertion)

The backtrace picks `min(lev[candidate] + transition_cost)`, which naturally excludes expensive substitutions while preserving cheap matches.

alvations (approved):
LGTM after the added cost feature and the unittest fixes in the latest commit.

stevenbird:
@alvations would you mind resolving the conflicts please?

alvations:
The autocode merging with the coding assistant did the right thing in resolving the merge conflict but the commit changes comment and commit squashing iss a mess with the forced push... So I'm commenting the changes for the changes in my latest commit below. 

 ### Before (conflict):
```
  <<<<<<< HEAD
  from nltk.metrics.distance import (
      edit_distance,
      jaro_similarity,
      jaro_winkler_similarity,
  )
  =======
  from nltk.metrics.distance import edit_distance, edit_distance_align
  >>>>>>>
```


 ### After (resolved):

```
  from nltk.metrics.distance import (
      edit_distance,
      edit_distance_align,
      jaro_similarity,
      jaro_winkler_similarity,
  )
```

Lesson learnt, small things like this, doing it myself is better than letting the agent do the work and then me re-committing a longer comment on the github =) 

@stevenbird The latest commit should be ready now to merge after the resolution.

stevenbird:
Thanks @alvations

PR #3188, opened 2023-09-27
Update Penn POS descriptions in chunkparser_app.py

nathanjmcdougall:
Corrected typo (masps -> mass) as per
https://cs.nyu.edu/~grishman/jet/guide/PennPOS.html


tomaarsen:
Well spotted! I appreciate it.

PR #3190, opened 2023-09-28
minor fix for wordnet lemmatization pos param documentation

sharpblade4:
Encountered extra parameter in the [documentation of wordnet lemmatization](https://www.nltk.org/api/nltk.stem.wordnet.html).  

The minor issue is that `pos` declared twice as `param` instead of declaring its `type`. 

Followed [contribution instructions](https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md) to apply a minor fix to solve that.


tomaarsen:
Thank you! Well spotted.

PR #3199, opened 2023-10-27
Fix dunning log likelihood ValueError

vivekkalyan:
This pull request addresses a `ValueError` that occurs in the `_dunning_log_likelihood` function during the execution of `PunktTrainer.train()`. The error is triggered when `count_b` is equal to `N`, resulting in `p1` being equal to 1, and subsequently causing `math.log(1.0 - p1)` to be `math.log(0)`, which is undefined.

To fix this issue, a small epsilon value of `1e-8` is added to the arguments of the logarithm functions to ensure they are always within a valid range, preventing the `ValueError`.

Additionally, this pull request includes new tests for the `PunktTrainer` class to prevent regression.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -298,6 +298,7 @@
 - Jan Lennartz <https://github.com/Madnex>
 - Tim Sockel <https://github.com/TiMauzi>
 - Ron Urbach <https://github.com/sharpblade4>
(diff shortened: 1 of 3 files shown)


stevenbird:
Thanks @vivekkalyan

Issue #3202, opened 2023-11-09
ToktokTokenizer doesn't call one of the included replacement patterns and thus doesn't tokenize some punctuation, like opening guillemets


alexrudnick:
Right now, if you do something like:

```lang:python
import nltk
toktok = nltk.tokenize.ToktokTokenizer()
toktok.tokenize("«Sí, por favor.»")
['«Sí', ',', 'por', 'favor', '.', '»']
```
... you can see that the opening punctuation isn't split off from the first word. There's an included pattern in nltk.tokenize.ToktokTokenizer, but it's not invoked! Invoking the pattern gives the appropriate punctuation splitting.

I'm about to send a PR for this; it also removes a redundant call that does the same regex twice, and makes the testing for ToktokTokenizer a bit more robust.

tomaarsen:
Resolved via #3203. Thanks!

PR #3203, opened 2023-11-09
Make sure that we invoke all the intended regex patterns in ToktokTokenizer...

alexrudnick:
and do it only once. Also add more tests.

This is for #3202 .

### DIFF
--- a/nltk/test/tokenize.doctest
+++ b/nltk/test/tokenize.doctest
@@ -395,3 +395,52 @@ Test that `ValueError` exceptions are raised when illegal arguments are used.
     Traceback (most recent call last):
       ...
     ValueError: Smoothing method bar not recognized
+
+
+Regression Tests: ToktokTokenizer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    >>> toktok = ToktokTokenizer()
+    >>> text = u'Is 9.5 or 525,600 my favorite number?'
+    >>> print(toktok.tokenize(text, return_str=True))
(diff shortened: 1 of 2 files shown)


tomaarsen:
Hello!

Looks great - though it's plausible (but unlikely) that the duplicate `FINAL_PERIOD_2` was on purpose. What do you think @alexrudnick?

- Tom Aarsen

alexrudnick:
Thanks for taking a look, @tomaarsen !

So I considered this possibility, that perhaps `FINAL_PERIOD_2` was intended to be called twice, but I think that's unlikely for a few reasons:
- There's no mention of it in the code; if there was some subtle reason why that regex ought to be called twice, then that seems like it'd be worthy of a comment.
- It's not called twice in the original Perl! https://github.com/jonsafari/tok-tok/blob/master/tok-tok.pl
- The only reason why you might need to call it twice would be if some other rule had created a situation where FINAL_PERIOD_2 would need to trigger later on, and I don't see any rules that would create that kind of situation.

Peeking at blame, it looks like @purificant wrote that line, so maybe they could chime in if they like?

tomaarsen (approved):
Thanks for the analysis! I think that clears it right up, especially the second point. Though @purificant did last affect this line, that was with his #2163 PR that applied `black` over all code. The only person to meaningfully write/edit this code was @alvations.

This looks ready to merge as far as I'm concerned!

alexrudnick:
Right on, I just noticed that right after I said it, good catch! Code reformatting patch.

Thanks, merging!

PR #3207, opened 2023-11-17
fix: enable py 3.12 in ci and fix error in bleu calculation

k4black:
## The problem

`Fraction(..., _normalize=False)` [was removed in python 3.12](https://docs.python.org/3/library/fractions.html#fractions.Fraction) so it causes error in NLTK functions trying to call it this way.

## Changes

* **Enable CI testing for python 3.12**

* **Indicate 3.12 in docs and metadata**

* **Fix BLEU calculation**  
  I have add Child class for Fraction to support `_normalize=False`. Please let me know is better to have functions return `(numerator, denominator)` instead of complex class.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -76,7 +76,7 @@ jobs:
     needs: [cache_nltk_data, cache_third_party]
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
         os: [ubuntu-latest, macos-latest, windows-latest]
       fail-fast: false
     runs-on: ${{ matrix.os }}
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 NLTK -- the Natural Language Toolkit -- is a suite of open source Python
 modules, data sets, and tutorials supporting research and development in Natural
-Language Processing. NLTK requires Python version 3.7, 3.8, 3.9, 3.10 or 3.11.
+Language Processing. NLTK requires Python version 3.7, 3.8, 3.9, 3.10, 3.11 or 3.12.
 
 For documentation, please visit [nltk.org](https://www.nltk.org/).
 
--- a/nltk/test/unit/translate/test_bleu.py
+++ b/nltk/test/unit/translate/test_bleu.py
@@ -2,7 +2,6 @@
 Tests for BLEU translation evaluation metric
 """
 
-import io
 import unittest
 
 import numpy as np
--- a/nltk/translate/bleu_score.py
+++ b/nltk/translate/bleu_score.py
@@ -7,48 +7,39 @@
 # For license information, see LICENSE.TXT
 
 """BLEU score implementation."""
-from __future__ import annotations
-
 import math
 import sys
 import warnings
 from collections import Counter
-from dataclasses import dataclass
+from fractions import Fraction as _Fraction
 
 from nltk.util import ngrams
 
 
-@dataclass
-class Fraction:
-    """
-    This class is used to represent a fraction with both the numerator and denominator saved for later retrieval.
-    Python 3.12 removed _normalize=False from the standard lib Fraction constructor.
-    """
+class Fraction(_Fraction):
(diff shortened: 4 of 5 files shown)


k4black:
@stevenbird @tomaarsen 
Could you please check this PR?   
As nltk does not forbid it to be installed on 3.12 (officially released a while ago) current version breaks the pipelines and force to use 3.11 version.

stevenbird:
Thanks @k4black !

didier-durand:
Hi, running nltk 3.8.1 on Python 3.12

I get this error when calling sentence_bleu (see code below):

```
File "/Users/didduran/dev/venv/lib/python3.12/site-packages/nltk/translate/bleu_score.py", line 368, in modified_precision
    return Fraction(numerator, denominator, _normalize=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Fraction.__new__() got an unexpected keyword argument '_normalize'
```

Is that related to the above? 

Thanks

Didier

To reproduce:

       ```
        # Define your desired weights (example: higher weight for bi-grams)
        weights = (0.25, 0.25, 0, 0)  # Weights for uni-gram, bi-gram, tri-gram, and 4-gram
        # Reference and predicted texts (same as before)
        reference = [["the", "picture", "is", "clicked", "by", "me"],
                     ["this", "picture", "was", "clicked", "by", "me"]]
        predictions = ["the", "picture", "the", "picture", "by", "me"]
        # Calculate BLEU score with weights
        score = sentence_bleu(reference, predictions, weights=weights)
        print(score)
```

purificant:
You are correct @didier-durand, this is related to the `Fraction` and `_normalize` change in 3.12

didier-durand:
@purificant : thanks for the rapid answer! How can I fix this problem: I already run the latest version of nltk package. v.3.8.1

vagetablechicken:
> @purificant : thanks for the rapid answer! How can I fix this problem: I already run the latest version of nltk package. v.3.8.1

python <= 3.11, I've test on python 3.9, it's ok.

abusharaf:
As per the thread, it's resolved on Python 3.12, but I am not sure why this fix has not been updated on the original package. Do you have any clue? @k4black

didier-durand:
I confirm that it does not work in Python 3.12 on my MacOS (if there is something making it OS-dependent)

abusharaf:
I am not sure to be honest, I don't think it's OS dependent, I believe the update is not reflected on the original source where pip getting the nltk package

abusharaf:
@didier-durand it worked now, thanks for @k4black for updating the file, I looked into this link https://github.com/nltk/nltk/pull/3207/commits/544bf807bf506927872b7b837c6e1514e80419ab#diff-644ebc352f93bb9a63ff8edbc7292e137cf1adaf68d5dc4f1aef13f4de96bf20

and extracted the updates and reflected them on my local bleu_score.py file and it worked very well :),



I think it's matter of only updating the distributed package of NLTK to reflect the updates on bleu_score file

didier-durand:
Thank you for the explanations.
So, will we get a new version of the package with the updated file?

abusharaf:
From your error message earlier this one => ""/Users/didduran/dev/venv/lib/python3.12/site-packages/nltk/translate/bleu_score.py" please open it using MS V Code for instance, and replace it with the content in the link I shared earlier, and rerun your code , it will work perfecto :)

didier-durand:
Yes, I will do that for my laptop. But, my code will included in a project run with DevOps: they recreate full env from official packages at each iteration. So, I will need this fix to get into the official code.

kariemoorman:
I am also receiving this error using Python 3.12. Production build supports Python 3.12 so I cannot revert to version 3.11.  Please can we re-open this request? Thank you.

eParkinsonDC:
error still occuring with nltk==3.8.1 and python 3.12.5

ottokruse:
The fixed code has only been released as a pre-release for now (5 hours ago, someone is working on it!). Pre-releases are ignored by normal pip installs unless you pass the `pre` flag: `pip install --pre nltk`

codelion:
This is fixed in nltk 3.9.1 which is now released.

frankiedrake:
I have Python 3.12.2 and nltk 3.9.1 and still get this error :/

PR #3183, opened 2023-08-22
Refactoring

tosemml:
This minor update aims to incorporate numpy operations whenever possible and leverage the Python join method to enhance the code's Pythonic nature.

PR #3215, opened 2023-12-17
quickfix syntax / typo

purificant:
This PR is a quickfix for a small issue introduced in https://github.com/nltk/nltk/pull/3183. It's needed to avoid a syntax error on `import nltk`


tomaarsen:
Thanks for this critical fix!

PR #3116, opened 2023-02-01
Support Open English Wordnet 2022

ekaf:
Fix #3059: add support for the latest OEWN release.

For ex., pretty print a list of synsets which have been split between WordNet 3.0 and OEWN 2022. This info is stored in the 'splits' dictionary, with counts of how many senses are carried over into each part of the split.

```
from nltk.corpus import wordnet2022 as wn

print(f"{len(wn.splits)} synsets were split between Wordnet 3.0 and {wn.get_version()}:\n")

for ss in sorted(list(wn.splits.keys())):
    print(f"{ss} -> {' + '.join([b + ' (' + str(a) + ' sensekey/s)' for a,b in wn.splits[ss]])}")
```

```
48 synsets were split between Wordnet 3.0 and 2022:

00040058-s -> 00040230-s (1 sensekey/s) + 00040078-s (2 sensekey/s)
00046299-r -> 00473995-r (1 sensekey/s) + 00046404-r (3 sensekey/s)
00202934-v -> 00296457-v (1 sensekey/s) + 00203725-v (2 sensekey/s)
00371163-s -> 03166825-s (1 sensekey/s) + 00372435-s (3 sensekey/s)
00522349-s -> 00524222-s (1 sensekey/s) + 00524591-s (1 sensekey/s)
00688377-v -> 00691232-v (1 sensekey/s) + 00715545-v (3 sensekey/s)
00766234-n -> 00770522-n (2 sensekey/s) + 00770714-n (4 sensekey/s)
00779248-n -> 00783716-n (11 sensekey/s) + 00784043-n (1 sensekey/s)
00824767-v -> 00826835-v (1 sensekey/s) + 00827564-v (20 sensekey/s)
00845523-n -> 01242621-n (1 sensekey/s) + 00854605-n (10 sensekey/s)
00949619-n -> 00959774-n (1 sensekey/s) + 00960217-n (1 sensekey/s)
00988988-s -> 00993654-s (1 sensekey/s) + 00992517-s (7 sensekey/s)
01068306-s -> 01072418-s (3 sensekey/s) + 01072535-s (1 sensekey/s)
01301051-v -> 01306626-v (1 sensekey/s) + 01306219-v (3 sensekey/s)
01878870-s -> 01885678-s (1 sensekey/s) + 01884927-s (5 sensekey/s)
01896031-n -> 01919623-n (2 sensekey/s) + 15493564-n (1 sensekey/s)
[...]

PR #3218, opened 2023-12-18
Silence verbose warnings in closure

### LINKED ISSUE #2314: Infinite loop in Wordnet closure and tree

ekaf:
Fix #2314: after discarding eventual cycles in relation closures, it is still annoying that warnings are output, which are irrelevant to most end users.

This PR adds a _verbose_ parameter to the functions that call warnings.warn() in _util.py_.
The default _setting_ is _verbose=False_, and the warning is only output if _verbose_ is True.

### DIFF
--- a/nltk/util.py
+++ b/nltk/util.py
@@ -263,7 +263,6 @@ def edge_closure(tree, children=iter, maxdepth=-1, verbose=False):
     :param maxdepth: to limit the search depth
     :param verbose: to print warnings when cycles are discarded
 
-
     Yield the edges of a graph in breadth-first order,
     discarding eventual cycles.
     The first argument should be the start node;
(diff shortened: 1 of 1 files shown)


purificant (approved):

ekaf:
CI fails only with Python 3.11 on macos-latest, because:

> line 1: pytest: command not found

purificant:
> CI fails only with Python 3.11 on macos-latest, because:
> 
> > line 1: pytest: command not found

Workaround is to clear cache and re-run failed job.

I noticed this bug but not sure what the root cause is, it only happens on MacOS, most frequently with py3.11. I think it can be worked around with a small change to CI at the cost of slightly increased CI run time duration.

ekaf:
Thanks @purificant! Yes, this is due to a long-standing bug with the macos-latest cache. It only occurred randomly, so it remained somewhat mysterious, and @tomaarsen used to  [fix it by clearing the cache manually](https://github.com/nltk/nltk/pull/3116#issuecomment-1413447743).

purificant:
> Thanks @purificant! Yes, this is due to a long-standing bug with the macos-latest cache. It only occurred randomly, so it remained somewhat mysterious, and @tomaarsen used to [fix it by clearing the cache manually](https://github.com/nltk/nltk/pull/3116#issuecomment-1413447743).

That's right. I came to the same conclusion. Thank you @ekaf for pointing to the evidence of it being a long standing bug. I am proposing a solution https://github.com/nltk/nltk/pull/3221 that would remove the need to clear cache manually.

ekaf:
I added ":param" descriptions suggested by Copilot. This duplicates some information, when the comments already describe the same parameters in plain text. On the other hand, the plain text descriptions are more detailed.

ekaf:
This PR is now superseded by #3245.

Issue #1978, opened 2018-03-14
Consistent pos argument between wn.synsets() and WordNetLemmatizer.lemmatize()


alvations:
Currently, there's some inconsistency of how POS is treated in `wn.synsets()` and `WordNetLemmatizer.lemmatize()`, e.g. 

```python
>>> from nltk.corpus import wordnet as wn
>>> from nltk.stem import WordNetLemmatizer
>>> wnl = WordNetLemmatizer()

# Accepts None and let pos be underspecified.
>>> wn.synsets('running', pos=None)
[Synset('run.n.05'), Synset('run.n.07'), Synset('running.n.03'), Synset('running.n.04'), Synset('track.n.11'), Synset('run.v.01'), Synset('scat.v.01'), Synset('run.v.03'), Synset('operate.v.01'), Synset('run.v.05'), Synset('run.v.06'), Synset('function.v.01'), Synset('range.v.01'), Synset('campaign.v.01'), Synset('play.v.18'), Synset('run.v.11'), Synset('tend.v.01'), Synset('run.v.13'), Synset('run.v.14'), Synset('run.v.15'), Synset('run.v.16'), Synset('prevail.v.03'), Synset('run.v.18'), Synset('run.v.19'), Synset('carry.v.15'), Synset('run.v.21'), Synset('guide.v.05'), Synset('run.v.23'), Synset('run.v.24'), Synset('run.v.25'), Synset('run.v.26'), Synset('run.v.27'), Synset('run.v.28'), Synset('run.v.29'), Synset('run.v.30'), Synset('run.v.31'), Synset('run.v.32'), Synset('run.v.33'), Synset('run.v.34'), Synset('ply.v.03'), Synset('hunt.v.01'), Synset('race.v.02'), Synset('move.v.13'), Synset('melt.v.01'), Synset('ladder.v.01'), Synset('run.v.41'), Synset('running.a.01'), Synset('running.s.02'), Synset('running.a.03'), Synset('running.a.04'), Synset('linear.s.05'), Synset('running.s.06')]


# Doesn't accept None and raise a KeyError
>>> wnl.lemmatize('running', pos=None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nltk/stem/wordnet.py", line 40, in lemmatize
    lemmas = wordnet._morphy(word, pos)
[...]

PR #3225, opened 2024-01-05
Avoid recursive suffix stripping in wordnet morphy

### LINKED ISSUE #2567: WordNetLemmatizer not properly lemmatizing some words
Some words are lemmatized improperly, due to picking the smallest possible lemma: 

```py
lemmatizer = WordNetLemmatizer()
lemmatizer.lemmatize('dose', 'n') # returns "dose"
lemmatizer.lemmatize('doses', 'n') # returns "dos"
wordnet._morphy('doses', 'n') # returns ["dose", "dos"]
wordnet.morphy('doses', 'n') # returns "dose"
```

ekaf:
Fix #2567: the implementation of Wordnet's __\_morphy_ lemmatizer in NLTK is buggy, because it adds a recursive step, which is not a part of the _morphy_ program,  as implemented in the "morph.c" source file that was  included with the original Princeton WordNet, and described in the [morphy manual](https://wordnet.princeton.edu/documentation/morphy7wn). 

So currently, when one pass over the possible morphological substitutions does not yield any known lemma, additional passes may be applied on the results. For ex.:

```

for w in ['cats', 'catss']:
    print(f"{w} -> {wn._morphy(w, pos='n')}")

```

> cats -> ['cat']
> catss -> ['cat']

After this PR, the "s" suffix is only stripped once, leading to the rejection of the bad "catss" form:

> cats -> ['cat']
> catss -> []

This last result agrees with the official [Princeton output](http://wordnetweb.princeton.edu/perl/webwn?s=catss), as well as with the [implementation in Wn](https://github.com/goodmami/wn/blob/main/wn/morphy.py) by @goodmami.

Similarly, the following errors do not occur after this PR:

| Call | Before | After |
| ---- | ------ | ----- |
| _morphy('possesses', 'n') | ['posse'] | [] |
| _morphy('ramesses', 'v') | ['ram'] | [] |
| _morphy('iss', 'n') | ['i'] | [] |
| _morphy('anchoresses', 'v') | ['anchor'] | [] |
| _morphy('askeses', 'v') | ['ask'] | [] |
| _morphy('bibless', 'n') | ['bible'] | [] |
| _morphy('bowses', 'n') | ['bow'] | [] |
| _morphy('carses', 'n') | ['car'] | [] |
| _morphy('cateresses', 'v') | ['cater'] | [] |
| _morphy('chowses', 'n') | ['chow'] | [] |
| _morphy('hydrases', 'n') | ['hydra'] | [] |
| _morphy('idlesses', 'n') | ['idle'] | [] |
| _morphy('idlesses', 'v') | ['idle'] | [] |
| _morphy('marses', 'v') | ['mar'] | [] |
| _morphy('pareses', 'v') | ['pare'', ' 'par'] | [] |
| _morphy('replicases', 'n') | ['replica'] | [] |
| _morphy('semises', 'n') | ['semi'] | [] |
| _morphy('tootses', 'n') | ['toot'] | [] |
| _morphy('tootses', 'v') | ['toot'] | [] |
| _morphy('torqueses', 'n') | ['torque'] | [] |

On the other hand. this PR does not change the results of [@tomaarsen's plurals test](https://github.com/nltk/nltk/issues/2567#issuecomment-715439448): WordNetLemmatizer still performs better 470 times, morphy 32 times, and there are 62 ties.

However, this PR proposes to add a few comments in WordNetLemmatizer, to make it more clear that its lemmatize() function picks the shortest lemma among the candidates returned by _\_morphy_, and that this behaviour is not a bug but a feature.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -2082,7 +2082,7 @@ class WordNetCorpusReader(CorpusReader):
         # Given an original string x
         # 1. Apply rules once to the input to get y1, y2, y3, etc.
         # 2. Return all that are in the database
-        # 3. If there are no matches: (edited by ekaf) don't recurse, return an empty list.
+        #    (edited by ekaf) If there are no matches return an empty list.
 
         exceptions = self._exception_map[pos]
         substitutions = self.MORPHOLOGICAL_SUBSTITUTIONS[pos]
@@ -2096,7 +2096,7 @@ class WordNetCorpusReader(CorpusReader):
             ]
 
         def filter_forms(forms):
-            result = []
+            result = []  # Return an empty list if we can't find anything
             seen = set()
             for form in forms:
                 if form in self._lemma_pos_offset_map:
@@ -2106,21 +2106,15 @@ class WordNetCorpusReader(CorpusReader):
                             seen.add(form)
             return result
 
-        # 0. Check the exception lists
-        if check_exceptions:
-            if form in exceptions:
-                return filter_forms([form] + exceptions[form])
-
-        # 1. Apply rules once to the input to get y1, y2, y3, etc.
-        forms = apply_rules([form])
+        if check_exceptions and form in exceptions:
+            # 0. Check the exception lists
+            forms = exceptions[form]
+        else:
+            # 1. Apply rules once to the input to get y1, y2, y3, etc.
+            forms = apply_rules([form])
 
         # 2. Return all that are in the database (and check the original too)
-        results = filter_forms([form] + forms)
-        if results:
-            return results
-
-        # Return an empty list if we can't find anything
-        return []
+        return filter_forms([form] + forms)
 
     #############################################################
     # Create information content from corpus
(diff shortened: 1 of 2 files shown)


ekaf:
Converted to draft, as it seems possible to handle more of the issues related to WordNetLemmatizer,.

ekaf:
The historical WordNet lemmatizer is Morphy, so many users would intuitively expect a more standard behaviour from WordNetLemmatizer.lemmatize(). But instead, that wrapper is defined by non-standard features: it defaults to nouns, picks the shortest lemma, and eventually accepts any word not included in WordNet.
So this PR proposes to also fix #1978 and #3227, by adding an alias to wordnet's \_morphy() and morphy() functions, for those users of the WordNetLemmatizer class who want access to a more standard
WordNet lemmatizer.
On the other hand, this PR leaves the lemmatize() function unchanged, for the many users who have been accustomed to its non-standard behaviour for a long time.

stevenbird:
Thanks @ekaf. Sorry for the delay

PR #3229, opened 2024-01-26
Add reference to entropy implementation used

### LINKED ISSUE #2961: nltk.lm.api entropy formula source?
Hi! I've been experimenting with training and testing a standard trigram language model on my own dataset. Upon investigating the `entropy` method of the LM class, I was a bit confused. The [docs](https://www.nltk.org/api/nltk.lm.api.html#nltk.lm.api.LanguageModel.entropy) only mention a very brief description: "Calculate cross-entropy of model for given evaluation text."

It seems to me that this entropy measure just **averages the ngram negative log probability** (so trigram in my case) and **makes it positive** by multiplying by -1:
```python
def _mean(items):
    """Return average (aka mean) for sequence of items."""
    return sum(items) / len(items)

def entropy(self, text_ngrams):
    """Calculate cross-entropy of model for given evaluation text.

    :param Iterable(tuple(str)) text_ngrams: A sequence of ngram tuples.
    :rtype: float

    """
    return -1 * _mean(
        [self.logscore(ngram[-1], ngram[:-1]) for ngram in text_ngrams]
    )
```
**So my general question is: Where does this formula for entropy stem from? Is there any paper referencing the method?** I'm just a bit stuck with the different versions of entropy that exist and I don't know which one is used here (and therefore I don't know how to interpret it correctly).

### Other formulas of entropy/perplexity

As a formula I know the [Shannon entropy](https://en.wikipedia.org/wiki/Entropy_(information_theory)) which in Python would be:
```python
def shannon_entropy(self, text_ngrams):
    return -1 * sum(
        [self.score(ngram[-1], ngram[:-1]) * self.logscore(ngram[-1], ngram[:-1]) for ngram in text_ngrams]
    )
```

And there's also the perplexity formula of [Jurafsky](https://www.youtube.com/watch?v=NCyCkgMLRiY), which returns a different score than `lm.perplexity` (which is 2**entropy):
```python
from numpy import prod

def jurafsky_perplexity(self, text_ngrams):
    problist = [self.score(ngram[-1], ngram[:-1]) for ngram in text_ngrams]
    return pow(prod(problist), -1/len(problist)

```
Thanks!

PS: My apologies if this issue lacks some information - I don't often open issues on Github 😅  Let me know if I need to update something! 

UPDATE: my Python implementation of the Jurafsky perplexity was totally wrong as I had quickly written it. I've updated it to reflect the actual scores from the web lecture.

mbauwens:
This is in reference to issue #2961.

This is just a simple addition to the [docs](https://www.nltk.org/api/nltk.lm.api.html#nltk.lm.api.LanguageModel.entropy), making it more clear that the entropy implementation in NLTK is the one based on the Shannon-McMillan-Breiman theorem, as used and referenced by Jurafsky ([Chapter 3 page 23 - 3.49](https://web.stanford.edu/~jurafsky/slp3/3.pdf)) and Jordan Boyd-Graber (`lm.txt` file in [this comment](https://github.com/nltk/nltk/issues/1342#issuecomment-203060085)).

This PR does not include the alternative implementation of the entropy function that was proposed in the issue.

### DIFF
--- a/nltk/lm/api.py
+++ b/nltk/lm/api.py
@@ -163,6 +163,10 @@ class LanguageModel(metaclass=ABCMeta):
 
     def contains_UNKs(self, ngram):
         """Helper method to indicate whether an ngram contains an UNK token or not.
+
+        :param tuple(str) ngram: ngram tuples
+        :rtype: bool
+        
         """
         return any([self.counts.unigrams[ng] for ng in ngram])
 
@@ -198,8 +202,10 @@ class LanguageModel(metaclass=ABCMeta):
         In case of <UNK> tokens, weight with the minimum relative frequency in the dataset.
 
         :param Iterable(tuple(str)) text_ngrams: A sequence of ngram tuples.
-        :param bool length_normalisation:
-        :param bool rel_freq_weighting:
+        :param bool length_normalisation: A boolean to indicate whether you want to
+            normalise by sequence length.
+        :param bool rel_freq_weighting: A boolean to indicate whether you want to
+            weight probabilities by ngram relative frequency.
         :rtype: float
 
         """
@@ -234,13 +240,13 @@ class LanguageModel(metaclass=ABCMeta):
             
         return entropy_extended
 
-    def perplexity_extended(self, text_ngrams, text_fdist, normalised=True, rel_freq_weighted=False):
+    def perplexity_extended(self, text_ngrams, length_normalisation=True, rel_freq_weighting=False):
         """Calculates the perplexity of the given text based on the extended version of the entropy method.
 
         This is simply 2 ** cross-entropy for the text, so the arguments are the same.
 
         """
-        return pow(2.0, self.entropy_extended(text_ngrams, text_fdist, normalised, rel_freq_weighted))
+        return pow(2.0, self.entropy_extended(text_ngrams, length_normalisation, rel_freq_weighting))
         
     def generate(self, num_words=1, text_seed=None, random_seed=None):
         """Generate words from the model.
--- a/nltk/lm/api.py
+++ b/nltk/lm/api.py
@@ -162,6 +162,8 @@ class LanguageModel(metaclass=ABCMeta):
 
     def entropy(self, text_ngrams):
         """Calculate cross-entropy of model for given evaluation text.
+        This implementation is based on the Shannon-McMillan-Breiman theorem,
+        as used and referenced by Jurafsky and Jordan Boyd-Graber.
 
         :param Iterable(tuple(str)) text_ngrams: A sequence of ngram tuples.
         :rtype: float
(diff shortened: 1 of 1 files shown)


nltk/lm/api.py
```diff
@@ -162,6 +162,8 @@ def context_counts(self, context):
 
     def entropy(self, text_ngrams):
         """Calculate cross-entropy of model for given evaluation text.
+        This implementation is based on the Shannon-McMillan-Breiman theorem,
+        as used and referenced by Jurafsky and Jordan Boyd-Graber.
```
iliakur:
```suggestion
        as used and referenced by Dan Jurafsky and Jordan Boyd-Graber.
```
Consistently use full names

nltk/lm/api.py
```diff
@@ -162,6 +162,8 @@ def context_counts(self, context):
 
     def entropy(self, text_ngrams):
         """Calculate cross-entropy of model for given evaluation text.
+        This implementation is based on the Shannon-McMillan-Breiman theorem,
```
iliakur, line 166:
```suggestion

        This implementation is based on the Shannon-McMillan-Breiman theorem,
```
Just to stay consistent with other docstrings in the module

iliakur (changes requested):
Minor nits, otherwise lgtm, thanks!

mbauwens (commented):
Minor changes in formatting of text. Look's good!

nltk/lm/api.py
```diff
@@ -179,6 +183,41 @@ def perplexity(self, text_ngrams):
         """
         return pow(2.0, self.entropy(text_ngrams))
 
+    def entropy_extended(self, text_ngrams, text_fdist, length_normalisation=True, rel_freq_weighting=False):
```
iliakur: Oh, hmm, I think I'll have to review this properly I'm afraid.

One thing that worries me is lack of tests. Let's maybe do it in a separate PR, I can help you with that.

mbauwens: @iliakur oh, my bad! I did not intend to already add these changes to the PR. I'm new to making PRs and didn't realise an updated fork would also be reflected into the previous PR. Will remove this from the PR!

mbauwens:
@iliakur yeah, so a bit messy here, but I've reverted the PR to just the update to the docs. I now have a separate branch on my fork that I'll be getting ready for a PR for an alternative entropy/perplexity method, including tests :)

iliakur (approved):

iliakur:
@stevenbird @tomaarsen this PR is ready to merge. I've been away from nltk development long enough that I figured I'd check in with you two about any formalities that may be missing.

tomaarsen:
I'm afraid I've also not had much time for NLTK recently. That said, I think this is good to go. Thanks for taking care of this, and nice to see you here again @mbauwens.

- Tom Aarsen

Issue #2331, opened 2019-07-04
PickleCorpusView raising UnicodeDecodeError when reading file


alexandredias3d:
I have been trying to save a modified corpus produced with `LazyMap` using`PickleCorpusView.write` as shown in the [PickleCorpusView documentation](https://www.nltk.org/_modules/nltk/corpus/reader/util.html#PickleCorpusView).

The writing seems to be working properly. However, when trying to read the pickled file, it raises a `UnicodeDecodeError`.

The following code does not even use a `LazyMap` but raises the same error

```python
Python 3.7.3 (default, Jun 24 2019, 04:54:02) 
[GCC 9.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import nltk
>>> nltk.corpus.reader.PickleCorpusView.write('In the real use case, this would be the result of LazyMap producing a list of list of tuples (word and PoS tag).', 'test.pickle')
>>> nltk.corpus.reader.PickleCorpusView('test.pickle')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/.virtualenv/venv/lib/python3.7/site-packages/nltk/collections.py", line 231, in __repr__
    for elt in self:
  File "/home/user/.virtualenv/venv/lib/python3.7/site-packages/nltk/corpus/reader/util.py", line 306, in iterate_from
    tokens = self.read_block(self._stream)
  File "/home/user/.virtualenv/venv/lib/python3.7/site-packages/nltk/corpus/reader/util.py", line 525, in read_block
    result.append(pickle.load(stream))
  File "/home/user/.virtualenv/venv/lib/python3.7/site-packages/nltk/data.py", line 1177, in read
    chars = self._read(size)
  File "/home/user/.virtualenv/venv/lib/python3.7/site-packages/nltk/data.py", line 1469, in _read
    chars, bytes_decoded = self._incr_decode(bytes)
  File "/home/user/.virtualenv/venv/lib/python3.7/site-packages/nltk/data.py", line 1491, in _incr_decode
    return self.decode(bytes, 'strict')
  File "/home/user/.virtualenv/venv/lib/python3.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

Am I missing something from the documentation?

[...]

PR #3230, opened 2024-01-29
Fix #3124- bug with PickleCorpusView raising UnicodeDecodeError 

### LINKED ISSUE #3124: PickleCorpusView (still) raising UnicodeDecodeError when reading file
In July 2019, issue #2331 was raised reporting that the `PickleCorpus` view raises a `UnicodeDecodeError`. This issue has not been fixed, though the issue was closed in the March 2020. In that issue, it was claimed that NLTK version 3.5 would fix this bug, but this is apparently not the case, as the buggy code is still in NLTK 3.8.1:

```python
class PickleCorpusView(StreamBackedCorpusView):

    BLOCK_SIZE = 100
    PROTOCOL = -1

    def __init__(self, fileid, delete_on_gc=False):
        self._delete_on_gc = delete_on_gc
        StreamBackedCorpusView.__init__(self, fileid)
```

The call to the `StreamBackedCorpusView` constructor should include the argument `encoding=None`, otherwise, that superclass will attempt to read from the file as if it were utf-8 encoded text.

An example of the error that I am seeing follows. The file `pickle50.out` is a file that I created by manually calling `pickle.dump`:

```python
>>> pcv = PickleCorpusView('scripts/pickle50.out')
>>> pcv[0]
  File "/usr/local/lib/python3.10/site-packages/nltk/corpus/reader/util.py", line 265, in __getitem__
    return next(self.iterate_from(i))
  File "/usr/local/lib/python3.10/site-packages/nltk/corpus/reader/util.py", line 306, in iterate_from
    tokens = self.read_block(self._stream)
  File "/usr/local/lib/python3.10/site-packages/nltk/corpus/reader/util.py", line 526, in read_block
    result.append(pickle.load(stream))
  File "/usr/local/lib/python3.10/site-packages/nltk/data.py", line 1055, in read
    chars = self._read(size)
  File "/usr/local/lib/python3.10/site-packages/nltk/data.py", line 1353, in _read
    chars, bytes_decoded = self._incr_decode(bytes)
  File "/usr/local/lib/python3.10/site-packages/nltk/data.py", line 1375, in _incr_decode
    return self.decode(bytes, "strict")
  File "/usr/local/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

However, calling `read_block(f)` directly with a file handle already opened in `rb` mode does work, but of course that does not allow us to use the corpus reader object like an iterable or sequence.

I can confirm that creating a subclass of `PickleCorpusView` that overrides `__init__` with the correct logic fixes the bug, e.g:

```python
class MyPickleCorpusView(StreamBackedCorpusView):
    def __init__(self, fileid, delete_on_gc=False):
        self._delete_on_gc = delete_on_gc
        StreamBackedCorpusView.__init__(self, fileid, encoding=None)
```

An object of this class behaves as `PickleCorpusView` is intended to behave.

Ubadub:
See #3124 and #2331 for more info (the latter was incorrectly closed and this issue marked as fixed, but as you can see in my comments on the former, it was not, in fact, fixed).

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -302,6 +302,7 @@
 - Akihiro Yamazaki <https://github.com/zakkie>
 - Ron Urbach <https://github.com/sharpblade4>
 - Vivek Kalyan <https://github.com/vivekkalyan>
+- Abhinav Patil <https://github.com/Ubadub>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/corpus/reader/util.py
+++ b/nltk/corpus/reader/util.py
@@ -517,7 +517,7 @@ class PickleCorpusView(StreamBackedCorpusView):
             whenever this object gets garbage-collected.
         """
         self._delete_on_gc = delete_on_gc
-        StreamBackedCorpusView.__init__(self, fileid)
+        StreamBackedCorpusView.__init__(self, fileid, encoding=None)
 
     def read_block(self, stream):
         result = []


rmalouf:
This looks good, but can you add a test for it (so we make sure it stays fixed)?

Ubadub:
Thanks for taking a look. Yes, I can try to do that— got a lot on my plate in the near future so it might a be a few weeks before I get to it, though.

alvations:
Thank you @Ubadub for the discussion and contribution.

Unfortunately, PickleCorpusView has been remove from NLTK, might be best for everyone to avoid using it too. Closing this issue.

PR #3231, opened 2024-01-30
Make WordNet's synset relations available from the lemmas

### LINKED ISSUE #1970: WordNet: Synsets' relations are absent from their corresponding lemmas
In WordNet, it seems to me that the relations relevant for a given synset should be relevant for its corresponding lemmas as well. In the `nltk` implementation, all semantic relations implemented as methods for synsets (e.g., `hypernyms()`) are indeed also available for lemmas. Though for example, the lemma `trick.n.01.trick` `hypernyms` method does not return anything, while its synset's does.

Using the lemma method:
```python
>>> trick_lemma = wnt.get_synset('trick.n.01').lemmas()[0]
>>> print(trick_lemma.hypernyms())
[]
```
Using the the synset method instead:
```python
>>> print(trick_lemma.synset().hypernyms())
[Synset('device.n.03')]
>>> print(trick_lemma.synset().hypernyms()[0].lemmas())
[Lemma('device.n.03.device'), Lemma('device.n.03.gimmick'), Lemma('device.n.03.twist')]
```

Expected behavior of the lemma method:
```python
>>> print(trick_lemma.hypernyms())
[Lemma('device.n.03.device'), Lemma('device.n.03.gimmick'), Lemma('device.n.03.twist')]
```

Versions:
```
nltk 3.2.5
Python 3.6.4
Ubuntu 17.10
```

ekaf:
Fix #1970: make WordNet's synset relations available from the lemmas.

```
insect = wn.lemmas("insect")[0]
print(insect.hypernyms())
```

> [Synset('arthropod.n.01')]

The _in_topic_domain()_ relation ("-c") is hybrid, because a few lemmas have both lemma and synset targets. For ex:

`print(insect.in_topic_domains())`

> [Lemma('chirpy.a.01.chirpy'), Synset('holometabolism.n.01')]

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -326,7 +326,7 @@ class Lemma(_WordNetObject):
     def _related(self, relation_symbol):
         get_synset = self._wordnet_corpus_reader.synset_from_pos_and_offset
         if (self._name, relation_symbol) not in self._synset._lemma_pointers:
-            return []
+            return self._synset._related(relation_symbol)
         return [
             get_synset(pos, offset)._lemmas[lemma_index]
             for pos, offset, lemma_index in self._synset._lemma_pointers[
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -324,15 +324,28 @@ class Lemma(_WordNetObject):
         return "%s('%s.%s')" % tup
 
     def _related(self, relation_symbol):
+        """Returns the lemma's relations, plus its' synset's relations.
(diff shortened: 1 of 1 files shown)


ekaf:
This needs more work in order to return the lemmas of the synset targets, as in the last example from [this comment](https://github.com/nltk/nltk/issues/1970#issue-301709671) by @marcevrard.
Alternatively, depending on the consensus, the relation definitions  could be moved from the WordnetObject class to the Synset class.

stevenbird:
@ekaf I believe it is more important to maintain the conceptual distinction between synsets (senses) and lemmas (forms).

PR #3233, opened 2024-02-19
10x Faster Levenshtein Distances

ashvardanian:
This PR replaces the native Python implementation of the Levenshtein distance with a SIMD-accelerated version from the StringZilla library.

On ~5 letter words from a typical English corpus - StringZilla is over 10x faster than current Python implementation in NLTK (1.24 s ± 57.4 ms vs 13.5 s ± 287 ms per loop across 7 runs). On a multilingual corpus with longer words - same result. 

1. StringZilla uses less `if` branches than most native implementations, so it's faster than many implementations in the serial mode. When AVX-512 is available it can use specialized assembly instructions to accelerate evaluation.
2. StringZilla currently [provides wheels](https://pypi.org/project/stringzilla/#files) for {manylinux, musllinux, macos, windows} × { 32-bit, 64-bit } × { x86, Arm, PowerPC } × [ Python 3.6, Python 3.12 ] making it practically the most portable hardware-accelerated library in the Python ecosystem.

### DIFF
--- a/nltk/metrics/distance.py
+++ b/nltk/metrics/distance.py
@@ -22,8 +22,6 @@ As metrics, they must satisfy the following three requirements:
 import operator
 import warnings
 
-import stringzilla as sz
-
 
 def _edit_dist_init(len1, len2):
     lev = []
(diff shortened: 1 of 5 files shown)


nltk/metrics/distance.py
```diff
@@ -22,6 +22,8 @@
 import operator
 import warnings
 
+import stringzilla as sz
```
tomaarsen: It may make sense to move the import to inside of the `edit_distance` method to avoid unnecessary importing.

ashvardanian: Hi @tomaarsen! Can you please double check the updated variant?

stevenbird:
[CI: retest]

ashvardanian:
[CI: retest]

alvations:
While this is a good speedup, it adds additional dependencies. I'll avoid it and users looking for speedier alternatives should be able to find `StringZilla` or other.

ekaf:
I would prefer an open source implementation in Python, allowing students to learn the classical algorithms. Compiled code is the contrary of what I would like to see in NLTK.

stevenbird:
Dropping this, thanks all.

Issue #3193, opened 2023-10-10
Add support for a `sort` argument in WordNet methods


bryant1410:
WordNet object methods support a series of methods, such as `hypernyms`, `antonyms`, etc., that under the hood use a private method called `_related`, that supports an argument called `sort` which by default is `True`. This argument sorts the output objects by name. For example, see:

https://github.com/nltk/nltk/blob/e2d368e00ef806121aaa39f6e5f90d9f8243631b/nltk/corpus/reader/wordnet.py#L134-L135

However, in some cases, we don't need the output to be sorted, and we may be performing these operations multiple times and on long lists, which incurs in considerable penalties because of multiple needless sorting going on under the hood.

Thus, I believe it'd be important that such methods supported a `sort` argument (as `_related` does), whose default value is `True` (to avoid breaking backward compatibility).

ekaf:
There are two _\_related()_ functions: one for lemmas, which does not support a _sort_ argument, and one for synsets, which has _sort_ set to True by default.
But False would be a better default, because most users  normally  don't want the output to be sorted alphabetically, and those who need it can easily sort it by themselves. 
So, even though it breaks backward compatibility,  the easiest fix would be to just remove the _sort_ option from the synsets' _\_related()_ function.

ekaf:
However, it seems that we need to sort the output, because otherwise it would be returned in random order, so no test would be reproducible.

Consider the hyponyms of _legal action_ with _sort_ set to False :

```
from nltk.corpus import wordnet as wn
ss = wn.synset('legal_action.n.01')
print([(wn.ss2of(x),x) for x in ss.hyponyms()])
```

> [('01184230-n', Synset('civil_action.n.01')), ('01184407-n', Synset('counterclaim.n.01')), ('01198588-n', Synset('test_case.n.01')), ('01184045-n', Synset('antitrust_case.n.01')), ('01198307-n', Synset('prosecution.n.01')), ('01184720-n', Synset('lis_pendens.n.01')), ('01184565-n', Synset('custody_case.n.01'))]

This output is in no particular order, and subsequent runs produce a different order.

Also, while it is true that some lists may be long (up to 600 elements in a few cases), there is no reason why it would be sorted multiple times, except if the user explicitly repeated the same call, which is not likely to happen anyway, and would not be an NLTK concern.

After all, the alphabetical sort produces the nicest presentation order, and especially it is stable, so I'm not sure that this issue needs fixing.

bryant1410:
[...]

PR #3240, opened 2024-03-16
Add support for disabling the sorting and list creation for WordNet object relation methods

bryant1410:
Add support for disabling the sorting and list creation for WordNet object relation methods. It keeps backward compatibility.

See #3193.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -131,77 +131,77 @@ class WordNetError(Exception):
 class _WordNetObject:
     """A common base class for lemmas and synsets."""
 
-    def hypernyms(self):
-        return self._related("@")
+    def hypernyms(self, sort=True, force_list=True):
+        return self._related("@", sort=sort, force_list=force_list)
 
-    def _hypernyms(self):
-        return self._related("@")
+    def _hypernyms(self, sort=True, force_list=True):
+        return self._related("@", sort=sort, force_list=force_list)
 
-    def instance_hypernyms(self):
-        return self._related("@i")
+    def instance_hypernyms(self, sort=True, force_list=True):
+        return self._related("@i", sort=sort, force_list=force_list)
 
-    def _instance_hypernyms(self):
-        return self._related("@i")
+    def _instance_hypernyms(self, sort=True, force_list=True):
+        return self._related("@i", sort=sort, force_list=force_list)
 
-    def hyponyms(self):
-        return self._related("~")
+    def hyponyms(self, sort=True, force_list=True):
+        return self._related("~", sort=sort, force_list=force_list)
 
-    def instance_hyponyms(self):
-        return self._related("~i")
+    def instance_hyponyms(self, sort=True, force_list=True):
+        return self._related("~i", sort=sort, force_list=force_list)
 
-    def member_holonyms(self):
-        return self._related("#m")
+    def member_holonyms(self, sort=True, force_list=True):
+        return self._related("#m", sort=sort, force_list=force_list)
 
-    def substance_holonyms(self):
-        return self._related("#s")
+    def substance_holonyms(self, sort=True, force_list=True):
+        return self._related("#s", sort=sort, force_list=force_list)
 
-    def part_holonyms(self):
-        return self._related("#p")
+    def part_holonyms(self, sort=True, force_list=True):
+        return self._related("#p", sort=sort, force_list=force_list)
 
-    def member_meronyms(self):
-        return self._related("%m")
+    def member_meronyms(self, sort=True, force_list=True):
+        return self._related("%m", sort=sort, force_list=force_list)
 
-    def substance_meronyms(self):
-        return self._related("%s")
+    def substance_meronyms(self, sort=True, force_list=True):
+        return self._related("%s", sort=sort, force_list=force_list)
 
-    def part_meronyms(self):
-        return self._related("%p")
+    def part_meronyms(self, sort=True, force_list=True):
+        return self._related("%p", sort=sort, force_list=force_list)
 
-    def topic_domains(self):
-        return self._related(";c")
+    def topic_domains(self, sort=True, force_list=True):
+        return self._related(";c", sort=sort, force_list=force_list)
 
-    def in_topic_domains(self):
-        return self._related("-c")
+    def in_topic_domains(self, sort=True, force_list=True):
+        return self._related("-c", sort=sort, force_list=force_list)
 
-    def region_domains(self):
-        return self._related(";r")
+    def region_domains(self, sort=True, force_list=True):
+        return self._related(";r", sort=sort, force_list=force_list)
 
-    def in_region_domains(self):
-        return self._related("-r")
+    def in_region_domains(self, sort=True, force_list=True):
+        return self._related("-r", sort=sort, force_list=force_list)
 
-    def usage_domains(self):
-        return self._related(";u")
+    def usage_domains(self, sort=True, force_list=True):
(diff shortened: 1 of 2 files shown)


ekaf:
@bryant1410, how big a speed improvement do you typically observe on your use case, when setting _sort_ and _force_list_ to False?

bryant1410:
Hey, I did a quick test based on one of the functions I use. The setup is:

```bash
pip install git+https://github.com/bryant1410/nltk@patch-5
```

```python
def _get_sibling_or_cousin_co_hyponym_lemmas(synset):
    for parent in synset.hypernyms():
        for grandparent in parent.hypernyms():
            for uncle_or_parent in grandparent.hyponyms():
                for cousin_or_sibling in uncle_or_parent.hyponyms():
                    if cousin_or_sibling != synset:
                        yield from cousin_or_sibling.lemmas()


def _get_sibling_or_cousin_co_hyponym_lemmas_no_sort(synset):
    for parent in synset.hypernyms(sort=False, force_list=False):
        for grandparent in parent.hypernyms(sort=False, force_list=False):
            for uncle_or_parent in grandparent.hyponyms(sort=False, force_list=False):
                for cousin_or_sibling in uncle_or_parent.hyponyms(sort=False, force_list=False):
                    if cousin_or_sibling != synset:
                        yield from cousin_or_sibling.lemmas()

import timeit

synset = wn.synset("man.n.01")

# We warm it up before benchmarking:
list(_get_sibling_or_cousin_co_hyponym_lemmas(synset))
```

With default values:

```pycon
>>> timeit.timeit(lambda: list(_get_sibling_or_cousin_co_hyponym_lemmas(synset)), number=100)
0.30820278730243444
```

With disabling sorting and list creations:

```pycon
>>> timeit.timeit(lambda: list(_get_sibling_or_cousin_co_hyponym_lemmas_no_sort(synset)), number=100)
0.20816301740705967
```

About 50% faster for this example.

More examples (in milliseconds):

|       Synset       | Before | After |
|:------------------:|:------:|:-----:|
|      man.n.01      |    308 |   208 |
|     woman.n.01     |    308 |   207 |
|     banana.n.01    |     62 |    37 |
|     apple.n.01     |     36 |    24 |
|     orange.n.01    |     13 |     9 |
|      dog.n.01      |     15 |    11 |
|      cat.n.01      |      4 |     3 |
| calamagrostis.n.01 |    274 |   148 |
|   adult_male.n.01  |    309 |   213 |
|     adult.n.01     |    118 |    63 |
|   girlfriend.n.01  |    183 |   123 |
|      somebody      |    159 |   109 |

In my real use case example, I call this function multiple times per data loader worker (which, in turn, I have one per CPU core). It's part of a code that bottlenecks (starves) the GPU usage, as I have to compute thousands of these per second per data loader worker.

ekaf:
There are two problems with the above test script.
First, it only shows the speedups attributed to sort=False. Setting force_list to False has no effect on the speed since the output is converted to a list anyway at last. 
Second, the test only checks that any _cousin_or_sibling_ found is different from the input synset, while it should also check that they have not been seen yet. For this reason, the list of cousin_or_sibling lemmas of man.n.01 has length 4430, while the corresponding set only has 2085 elements.

bryant1410:
> First, it only shows the speedups attributed to sort=False. Setting force_list to False has no effect on the speed since the output is converted to a list anyway at last.

Note that the list conversion occurs to the final output, not the intermediate ones. The list creations that are avoided are each of the outputs to the `hypernyms` or `hyponyms` calls.

Still, we can do a test that avoids the final `list`. E.g., we can follow [this SO post](https://stackoverflow.com/a/50938015/1165181) and just consume the iterable the fastest way possible. I did this and observed differences of 1-7ms.

Having said this, I later observed that using `force_list=True` actually makes it a bit faster (single-digit milliseconds). I looked online and found people discussing that it depends on the use case (e.g., using these functions for `_get_sibling_or_cousin_co_hyponym_lemmas` vs for a different use case). Still, there are memory benefits, so it's a tradeoff.

> Second, the test only checks that any _cousin_or_sibling_ found is different from the input synset, while it should also check that they have not been seen yet. For this reason, the list of cousin_or_sibling lemmas of man.n.01 has length 4430, while the corresponding set only has 2085 elements.

Thanks for catching this! For my use case, it's fine, and I believe it's preferred.

ekaf:
I also observe slightly better timings with force_list=True. The difference is small but consistent, so this additional parameter does not contribute to solving the speed bottleneck from #3193. Still, it may be useful for uses where one wants to consume the output one piece at a time instead of waiting for the whole list.

ekaf:
Testing with more inputs, the speedup from _force_list= False_ stays small, but it displays some randomness, and it is not always positive or negative. It may be questionable whether it is worth the cost of managing this additional parameter:

| Input  | Unique cousins | sort gain | list gain  |
|---------|---------|--------------|--------------|
|man.n.01 | 2085 | 25.74% | -0.12% |
|woman.n.01 | 2085 | 25.94% | -1.3% |
|banana.n.01 | 1357 | 59.19% | -0.2% |
|apple.n.01 | 210 | 41.86% | 0.27% |
|orange.n.01 | 95 | 33.44% | 0.35% |
|dog.n.01 | 204 | 26.96% | 1.65% |
|cat.n.01 | 104 | 22.75% | 1.54% |
|calamagrostis.n.01 | 4824 | 39.42% | -0.72% |
|adult_male.n.01 | 2085 | 26.61% | 1.46% |
|adult.n.01 | 971 | 52.92% | 2.5% |
|girlfriend.n.01 | 2377 | 25.42% | 1.08% |

_sort gain_ is the percentage speedup of setting sort=False, w.r.t. the baseline, and _list gain_ is the additional speedup of also setting _force_list=False_.

```
def speedup(t1,t2):
    return round((100*t1/t2)-100, 2)

```

In these tests, the functions to retrieve second degree cousins use sets, in order to avoid duplicates. This is much more effective (and in itself much quicker than the sort=False optimization) when the ratio of duplicates over unique cousins is high, as in the 'man.n.01' example, where it is 4430 over 2085. But when the ratio is low, as with 'dog.n.01' (212 over 204), using sets actually incurs a 20% runtime penalty.

ekaf (changes requested):
```
        if relation_symbol not in self._pointers:
            return []
```

With _force_list= False_, the return type of _\_related()_ becomes unpredictable, because it is an iterator when the _relation_symbol_ is present in _self.\_pointers_, but an empty list otherwise, in which case calling _next()_ on the output fails with an Error.

bryant1410:
It's an iterable regardless.

ekaf:
> It's an iterable regardless.

But it is not an iterator: _next([])_ fails with 

> TypeError: 'list' object is not an iterator

 while _next(iter([]))_ returns

> StopIteration

bryant1410:
You can do `iter` regardless, then `next`. Or a `for` loop directly.

ekaf:
Wouldn't you prefer that the output of __related(force_list=False)_ had the same predictable type, no matter whether targets were found or not?

bryant1410:
For me personally no but I can see how other people would prefer that. If
it's always e.g. an iterable type (that has `__iter__`) then it's fine for
me. So I have that one guarantee. Which makes sense with something not
being forced; it's more relaxed and something is not guaranteed then (that
is not a list).

ekaf:
https://github.com/nltk/nltk/commit/9ad1956d4f665d88b6ab26486bac779bd7352e63 fixed #3193.

bryant1410:
Should I close this one?

ekaf:
Obviously, this PR is less relevant, now that #3193 is solved otherwise. There may still be a case for iterators, though that could go into a different PR.

PR #3245, opened 2024-03-25
Avoid duplicate output in acyclic_breadth_first

### LINKED ISSUE #3244: Duplicates in wordnet hypernyms closure
The relation closure for wordnet synsets is supposed to prevent duplicates in the output. However, the duplicates check fails to detect some repetitions, which occur when there are multiple paths to a given synset. In the following output for ex., the branch going from 'taxonomic_group.n.01'  to 'entity.n.01' appears twice, because it is reachable by two different paths:

```
ss=wn.synset("calamagrostis.n.01")
print(list(ss.closure(lambda s: s.hypernyms())))
```

> [Synset('gramineae.n.01'), Synset('monocot_genus.n.01'), Synset('monocot_family.n.01'), Synset('genus.n.02'), Synset('family.n.06'), Synset('taxonomic_group.n.01'), Synset('taxonomic_group.n.01'), Synset('biological_group.n.01'), Synset('biological_group.n.01'), Synset('group.n.01'), Synset('group.n.01'), Synset('abstraction.n.06'), Synset('abstraction.n.06'), Synset('entity.n.01'), Synset('entity.n.01')]

Produce an SVG image to illustrate this graph:

```
from nltk.parse.dependencygraph import dot2img
print(dot2img(wn.digraph([ss])))
```
calamagrostis-out-closureyield txt

ekaf:
Fix #3244.

Duplicate elimination is a central theme in the literature on the transitive closure of relational operators.

> 
> Note that, when the graph is acyclic, duplicate elimination is important only for efficiency, not for termination.
> 
> Performance Evaluation of Algorithms for Transitive Closure, Kabler et. al., InformationSystems Vol. 17, No. 5, PP. 415-441, 1992.
> 

With cyclic cases, however, duplicate elimination becomes also crucial for termination, as seen in #2314.

Currently, the closure operation in _wordnet.py_ uses the _acyclic_breadth_first()_ function imported from _nltk/util.py_, which is supposed to prevent cycles. However, this function presently only detects cyclic children branches, but it does not properly eliminate some duplicates, that are accessible through different parent paths.

Consider for ex. the [transitive closure of the hypernyms of 'calamagrostis.n.01',](https://github.com/nltk/nltk/issues/3244#issue-2205148031) which presently duplicates the nodes that are accessible by two different paths.

With the present PR, the same call does not output duplicates:

```
ss=wn.synset("calamagrostis.n.01")
print(list(ss.closure(lambda s: s.hypernyms())))
```

`[Synset('gramineae.n.01'), Synset('monocot_genus.n.01'), Synset('monocot_family.n.01'), Synset('genus.n.02'), Synset('family.n.06'), Synset('taxonomic_group.n.01'), Synset('biological_group.n.01'), Synset('group.n.01'), Synset('abstraction.n.06'), Synset('entity.n.01')]`


ekaf:
This PR includes #3218, which fixes the problem raised by @rmccampbell, concerning [excessive verbosity in the warnings](https://github.com/nltk/nltk/issues/2314#issuecomment-1859077547).

ekaf:
The _acyclic_breadth_first()_ function uses two structures to manage the relation closure: _traversed_, a set of nodes already seen, and _queue_, a deque of (child, level) pairs to lookup next.
[This picture](https://github.com/nltk/nltk/issues/3244#issue-2205148031) shows how 'taxonomic_group.n.01', is added to the _queue_ twice, while not yet added to the _traversed_ set. The problem occurs later, when popping duplicates from the queue without checking _traversed_. So there are two ways of solving the problem, either checking _traversed_, or preventing duplicates in the _queue_. This PR implements the first solution, but it might be interesting to consider the second instead.

ekaf:
Calculating the transitive hypernymic closure of all the 82115 noun synsets in WordNet 3.0 with either solution yields identical closures in both cases, but the solution implemented in this PR is 15% quicker.

stevenbird:
Finally, sorry

PR #3259, opened 2024-05-26
add gensim@da2f388 for support py3.12

tavallaie:
updating dependencies for supporting latest version of python 3.12
I wrote same problem here.
https://blog.techbend.io/overcoming-the-hazm-compatibility-challenge-with-python-312


stevenbird:
Thanks @tavallaie, and sorry for the delay. I don't believe this is a current issue, but please reopen if so, thanks.

Issue #3264, opened 2024-06-14
ZeroDivisionError when computing Krippendorff's alpha


vera-bernhard:
is another edge case like in issue #1336

(1) Steps to reproduce:

```
from nltk.metrics.agreement import AnnotationTask
atask = AnnotationTask(data=[('c1', '1', 1),('c2', '1', 1), ('c2', '2', 2)])
atask.alpha()
```

(2) Error
  File "/lib/python3.10/site-packages/nltk/metrics/agreement.py", line 321, in alpha
    k_alpha = 1.0 - do / de
ZeroDivisionError: float division by zero

(3) 1

(4) nltk.__version__: '3.8.1'
    python --version: Python  3.10.12

PR #3265, opened 2024-06-14
Fix k-alpha for expected disagreement equals 0

vera-bernhard:
Fixes issue #3264 and adds 2 test cases.

**Issue description**:
The number of labels is considered before the number of valid items is determined. There is cases (s. below) where removing items with less than two annotation also reduced the number of labels considered. Hence, we have to check again after checking for valid labels if there's only one label present which means the annotators totally agree. Otherwise, we can run in a division by zero problem.

**Test case description:**
Per definition the expected disagreement only takes into consideration items where at least 2 coders have annotated. In case of a following dataset:
```python
        data=[
            ('coder1', '1', 1),
            ('coder2', '1', 1), 
            ('coder1', '2', 2)]
```
what is actually taken into consideration is 
```python
        data=[
            ('coder1', '1', 1),
            ('coder2', '1', 1), 
```
Hence, this should return k_alpha = 1.0.

### DIFF
--- a/nltk/metrics/agreement.py
+++ b/nltk/metrics/agreement.py
@@ -304,7 +304,6 @@ class AnnotationTask:
         total_disagreement = 0.0
         total_ratings = 0
         all_valid_labels_freq = FreqDist([])
-
         total_do = 0.0  # Total observed disagreement for all items.
         for i, itemdata in self._grouped_data("item"):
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @vera-bernhard

Issue #3266, opened 2024-06-21
Remote code execution vulnerability in NLTK


Dunedan:
The current and earlier versions of NLTK are vulnerable to a remote code execution vulnerability when using the integrated data package download functionality. A man-in-the-middle attacker or an attacker with control over the NLTK data index can force users which use data packages with pickled Python code to download a new version of the package which executes arbitrary code upon unpickling.

Data packages which have been identified to be vulnerable are "averaged_perceptron_tagger" and "punkt". For code to be vulnerable it has to download a data package and use functionality in NLTK which causes the data package to be unpickled.

Here is an example of vulnerable code for the "averaged_perceptron_tagger" data package:

```python
import nltk
nltk.download("averaged_perceptron_tagger")
nltk.pos_tag(["hello", "world"])
```

This vulnerability was reported together with POC code to exploit it multiple times to the NLTK team via the email address mentioned in [SECURITY.md](SECURITY.md) (and later directly to some maintainers as well). It was reported first on 2024-05-19. So far there has been no response from the NLTK maintainers to these reports.

https://github.com/nltk/nltk/issues/2522 already pointed out the security implications of using pickled code a few years ago, but didn't receive a response either.

ekaf:
@Dunedan, what would you suggest?  
Maybe print a warning and ask for confirmation before loading a pickle?

The pickles in question contain Python classes with executable functions. They are not data but programs. And,as a compiled format, they are not ideal for an open-source project. 

But an eventual attacker needs to trick the user into loading a malicious pickle.

Dunedan:
The short answer to solve this issue is to remove the download of code via network at runtime.

If that's not feasible for NLTK I have no clear answer how to solve this issue. However, there are some measures I'd implement, which would at least improve the situation:

#### Remove use of pickled code wherever possible

While I haven't checked what the code in in the data packages with pickled code does, I can imagine that it's not the code which is large, but rather data associated with it. Maybe the algorithms implemented in there can be moved into NLTK, removing the need for pickled code.

Also instead of using pickle, a domain-specific format could be used (depending on the use case something like ONNX (https://onnx.ai/) might fit for that).

#### Make pickled code reproducible

When an update for a data package with pickled code is issued, it's difficult to figure out what changed. Having the pickle files be reproducibly built would be helpful for assessing if they're legit.

#### Make documentation clear for download once during setup

[...]

PR #3270, opened 2024-07-05
Pickle-less average perceptron tagger

alvations:
Added an implementation to load/save taggers into json format instead of pickle.

Related: #3266 

### Test

```
cd nltk.   # main repo directory
python -m unittest nltk/test/unit/test_pos_tag.py 
python -m doctest nltk/test/tag.doctest  
```


.pre-commit-config.yaml
```diff
@@ -7,11 +7,6 @@ repos:
       - id: end-of-file-fixer
       - id: requirements-txt-fixer
       - id: check-yaml
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.0
-    hooks:
-    - id: pyupgrade
-      args: ["--py38-plus"]
```
alvations: Any clues on why `pyupgrade` is failing on the precommit hooks? I couldn't make it to work even after several different configuration. 

Locally pyupgrade didn't re-write any files. 

Pinging @purificant

alvations:
The CI/CD is failing for some non-deterministic behaviors from ngram LMs unrelated to this PR. 

Anyone has some idea of why that is so for non-py38 tests? Maybe @iliakur would know?

ekaf (commented):
@alvations, using 3 json files or only one for each tagger shouldn't make much difference at all.
However, with this PR, the tests fail for me because of a Loading Error, which may be trivial to solve. I will try to see if my setup is somehow wrong.
On the other hand, the same tests succeed using the [one-file approach](https://github.com/nltk/nltk_data/pull/208#issuecomment-2211680961), and a slightly simpler handler. Then, I only see 3 tests failing because of the linebreak discrepancies in the output, which you handle here by reformatting the test output, Alternatively, I think there exists a doctest flag to ignore such trivial differences.

alvations:
For the unittest, you might have to clear your local `nltk_data` cache, usually at `~/nltk_data/tag/`, delete the artifacts inside, and then do a `python -m nltk.downloader average_perceptron_tagger_eng`, before the `python -m unittest ...` 


```
git clone https://github.com/nltk/nltk.git
cd nltk
git checkout  pickle-patch
python -m nltk.downloader average_perceptron_tagger_eng
python -m unittest nltk/test/unit/test_pos_tag.py -vvv
python -m doctest nltk/test/tag.doctest  
```

----

For, the doctest, it should pass with the edits in this PR 

Copy and paste the changes from https://github.com/nltk/nltk/blob/1ec5bacda110624d2e3f310de4f865ccb03b4cd6/nltk/test/tag.doctest to your local machine and retry.

Somehow, I think we haven't used the doctest much in the past and overlooks minor details there. There was even an expected output that was just `[...]` (which it shouldn't) before the fix =) 

Most important line to check in the test for model port integrity should be:

```
    >>> print(tagger.accuracy(gold_data)) # doctest: +ELLIPSIS
    0.885931...
```

ekaf:
My problem was that I had used git to clone nltk_data, so it was not properly set up: there was a _packages_ directory with a _taggers_ subdirectory in it, instead of _taggers_ residing at the top level. Others may have the same problem, because the error messages that I saw reminded much of those reported in several issues related to packages download.

Now all tests pass, including this additional call:
`python -m doctest nltk/tag/perceptron.py -v
`

ekaf:
Concerning the changed linebreaks in doctest output, it may be safer to add the flag:
`# doctest: +NORMALIZE_WHITESPACE`
This  eliminates the need to reformat the output if the linebreaks change in the future.

alvations:
Thanks @ekaf for checking and the suggestion on the doctest option! 

This PR should be good to go once all the CI/CD tests check through, will merge it after.

PR #3274, opened 2024-07-06
Fixing CI/CD

alvations:
Did a force push of latest commit and our CI/CD is failing. https://github.com/nltk/nltk/actions/runs/9822351687/job/27119009274

Trying the following to see if it fixes:
 - ~Upgrade pyupgrade version~
 - Applied black + isort to all files
 - Upgrade black version
 - ~Applied pyupgrade to files that precommit hook was complaining about~
 - Temporarily disable pyupgrade
 - Cast Generalized Hamming Distance output to float from np.float
 - Use math.fsum in ngram LM tests, see https://github.com/python/cpython/issues/100425

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: requirements-txt-fixer
       - id: check-yaml
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.0
+    rev: v3.16.0
     hooks:
     - id: pyupgrade
       args: ["--py38-plus"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     - id: pyupgrade
       args: ["--py38-plus"]
   - repo: https://github.com/ambv/black
-    rev: 23.12.0
+    rev: 24.4.2
     hooks:
     - id: black
   - repo: local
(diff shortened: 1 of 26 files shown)


purificant:
In the link provided (https://github.com/nltk/nltk/actions/runs/9822351687/job/27119009274), the root cause is pyupgrade:
image
To resolve this specific issue, need to look at what changes pyupgrade is proposing and resolve them accordingly.

purificant:
Might be worth looking into swapping black for ruff :bulb:

alvations:
@purificant for helping to check the issues with the CI/CD, the cause of the issue from pyupgrade https://github.com/asottile/pyupgrade/issues/954 

And somehow I couldn't get the precommit hook to use anything other than 3.12.4 https://github.com/nltk/nltk/actions/runs/9826091285/job/27127137994 so we might have to disable it until Python upgrades on Github's precommit defaults to 3.12.5. @purificant do you know of other ways to overwrite the precommit hook's python version?

Ruff seems like a good idea. Opened an issue so that we don't forget.

alvations:
Merging this first so that the critical PRs can pass through.

purificant:
You are right @alvations , using python version 3.12.3 appears to work, while 3.12.4 has an issue, here is the temporary workaround: https://github.com/nltk/nltk/pull/3280
I can confirm that I see this issue in my tests as well and workaround fixes it

PR #3277, opened 2024-07-08
Towards pickle-free Punkt

ekaf:
Add Json encoding/decoding to the _PunktParameters_ class, using the scheme from  _nltk.jsontags_.
With this PR, a pickled PunktParameters class can be encoded to a Json file, for ex. _tokenizers/punkt/PY3/english.pickle_:

```
from nltk.data import load
import json
from os.path import basename, splitext
from nltk.jsontags import JSONTaggedDecoder, JSONTaggedEncoder
decoder = JSONTaggedDecoder()
encoder = JSONTaggedEncoder()

file = 'tokenizers/punkt/PY3/english.pickle'
tokenizer = load(file)
pa = tokenizer._params
encoded = encoder.encode(pa)
fjson = splitext(basename(file))[0] + ".json"
with open(fjson, 'w') as f:
    f.write(encoded)
```

Load the resulting Json file using nltk.load:

```
encoded = load(fjson) 
pa2 = decoder.decode_obj(encoded)
```

Compare the decoded Json object with the original pickle:

`print(type(pa2))`
<class 'nltk.tokenize.punkt.PunktParameters'>
`print(pa.collocations == pa2.collocations)`
True
`print(pa.sent_starters == pa2.sent_starters)`
True
`print(pa.abbrev_types == pa2.abbrev_types)`
True
`print(pa.ortho_context == pa2.ortho_context)`
True

To make this work, I had to slightly modify two lines (762 and 763) in  _nltk/data.py_, so please review those lines in particular.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -759,8 +759,8 @@ def load(
 
         resource_val = json.load(opened_resource)
         tag = None
-        if len(resource_val) != 1:
-            tag = next(resource_val.keys())
+        if len(resource_val) == 1:
+            tag = next(iter(resource_val.keys()))
         if tag not in json_tags:
             raise ValueError("Unknown json tag.")
     elif format == "yaml":
--- a/nltk/tokenize/punkt.py
+++ b/nltk/tokenize/punkt.py
@@ -111,6 +111,7 @@ import string
 from collections import defaultdict
 from typing import Any, Dict, Iterator, List, Match, Optional, Tuple, Union
 
+from nltk import jsontags
 from nltk.probability import FreqDist
 from nltk.tokenize.api import TokenizerI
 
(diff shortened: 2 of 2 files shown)


alvations:
Thanks @ekaf! I was also looking into the punkt over the weekend and find that only the params are different the lang vars are the same and with the current implementation, they seem immutable after loading (have to confirm)

This approach to tag it with jsontags looks good.

alvations:
[CI: retest]

ekaf:
When setting their _params to None, all 18 language pickles reduce to the same Yaml code:

```
18 _lang_vars: !!python/object/new:nltk.tokenize.punkt.PunktLanguageVars
18 _params: null
18 !!python/object:nltk.tokenize.punkt.PunktSentenceTokenizer
18 state: 1
18 _Token: !!python/name:nltk.tokenize.punkt.PunktToken ''
```

So we only need to handle _params.

ekaf:
Maybe this could expose to an eventual "sleepy" Json vulnerabilty. While in doubt, the Tab format seems a better alternative

PR #3280, opened 2024-07-08
fix python version for pre-commit as a workaround

purificant:
This PR:
- fixes version of python during pre-commit checks to 3.12.3, as a workaround for issue encountered with 3.12.4 in https://github.com/nltk/nltk/pull/3274
- re-introduces pyupgrade
- upgrades versions of pre-commit hooks and pyupgrade


alvations:
Thank you x3000 @purificant !!

alvations (approved):
LGTM! And the CI/CD doesn't complain now after changing the python-version =)

PR #3289, opened 2024-07-18
⚡️ Speed up windowdiff() by 122% in nltk/metrics/segmentation.py

ihitamandal:
This optimization was discovered automatically by using [codeflash.ai](https://www.codeflash.ai/)

### 📄 `windowdiff()` in `nltk/metrics/segmentation.py`

📈 Performance improved by **`122%`** (**`1.22x` faster**)

⏱️ Runtime went down from **`1.89 millisecond`** to **`851 microseconds`**
### Explanation and details

To optimize the given `windowdiff` function for faster execution, we can reduce redundant computations within the nested loop. Specifically, we can maintain running counts of the boundaries in both segments within the window and update these counts as the window slides, instead of recalculating the counts from scratch for each new window position.

Here's the optimized version of the `windowdiff` function.

In this optimized version.
1. We initialize `count1` and `count2` with the counts of the boundaries in the initial window.
2. For each subsequent window position, we update these counts by adjusting for the elements that leave and enter the window.
3. This avoids the need to recount the boundaries within each window from scratch, significantly reducing the time complexity of the method while maintaining the same output.

This approach should yield faster execution, especially for long segmentations.

### Correctness verification

The new optimized code was tested for correctness. The results are listed below.
#### ✅ 28 Passed − 🌀 Generated Regression Tests
<details>
<summary>(click to show generated tests)</summary>

```python
# imports
import pytest  # used for our unit tests
from nltk.metrics.segmentation import windowdiff


# unit tests
def test_basic_functionality():
    # Identical Segmentations
    res = windowdiff("0001000", "0001000", 3)
    res = windowdiff("111", "111", 2)

    # Completely Different Segmentations
    res = windowdiff("000", "111", 2)
    res = windowdiff("010101", "101010", 3)
    # Output of windowdiff was verified to be equal to the original implementation

def test_boundary_marker_variations():
    # Different Boundary Markers
    res = windowdiff("aaaaba", "aaaaba", 3, boundary="b")
    res = windowdiff("1110111", "1110111", 2, boundary="0")
    # Output of windowdiff was verified to be equal to the original implementation

def test_weighted_vs_unweighted():
    # Weighted Calculation
    res = windowdiff("0001000", "0000100", 3, weighted=True)
    res = windowdiff("1110111", "1111011", 2, weighted=True)

    # Unweighted Calculation
    res = windowdiff("0001000", "0000100", 3, weighted=False)
    res = windowdiff("1110111", "1111011", 2, weighted=False)
    # Output of windowdiff was verified to be equal to the original implementation

def test_edge_cases():
    # Minimum Length Segmentations
    res = windowdiff("0", "0", 1)
    res = windowdiff("1", "0", 1)

    # Window Width Equal to Length
    res = windowdiff("000", "001", 3)
    res = windowdiff("111", "110", 3)
    # Output of windowdiff was verified to be equal to the original implementation

def test_error_handling():
    # Unequal Lengths
    with pytest.raises(ValueError, match="Segmentations have unequal length"):
        windowdiff("000", "0000", 2)
    with pytest.raises(ValueError, match="Segmentations have unequal length"):
        windowdiff("1111", "111", 3)

    # Window Width Greater than Length
    with pytest.raises(ValueError, match="Window width k should be smaller or equal than segmentation lengths"):
        windowdiff("00", "00", 3)
    with pytest.raises(ValueError, match="Window width k should be smaller or equal than segmentation lengths"):
        windowdiff("111", "111", 4)

def test_large_scale_cases():
    # Large Segmentations
    res = windowdiff("0" * 1000 + "1", "0" * 1000 + "1", 500)
    res = windowdiff("01" * 500, "10" * 500, 100)
    # Output of windowdiff was verified to be equal to the original implementation

def test_mixed_content_segmentations():
    # Mixed Content
    res = windowdiff("0101010101", "1010101010", 4)
    res = windowdiff("1100110011", "0011001100", 3)
    # Output of windowdiff was verified to be equal to the original implementation

def test_non_string_segmentations():
    # Lists as Segmentations
    res = windowdiff([0, 0, 1, 0, 0], [0, 0, 0, 1, 0], 3)
    res = windowdiff([1, 1, 1, 0, 1], [1, 1, 0, 1, 1], 2)
    # Output of windowdiff was verified to be equal to the original implementation

def test_boundary_marker_as_non_string():
    # Integer Boundary Markers
    res = windowdiff([0, 0, 1, 0], [0, 1, 0, 0], 2, boundary=1)
    res = windowdiff([1, 1, 0, 1], [1, 0, 1, 1], 3, boundary=0)
    # Output of windowdiff was verified to be equal to the original implementation

def test_complex_patterns():
    # Complex Patterns
    res = windowdiff("001001001", "001001010", 3)
    res = windowdiff("111000111", "111111111", 4)
    # Output of windowdiff was verified to be equal to the original implementation
```
</details>


stevenbird:
Thanks @ihitamandal... great contribution.
I want to maximise the interpretability of the NLTK codebase, so I would like to keep the original implementation around (minimally in documentation). More soon.

alvations:
Thanks @ihitamandal for the contribution! Good catch on the speed up. 

Agree with @stevenbird on readability vs the optimization. The optimized code is a bit of an overkill and would only be useful when then k value is large enough. I've reverted it to the original implementation in the latest comit. 

However the unittests you've added are good to merge in for student's references. I've added some commits to add the math in the original paper in the docstring as well as extended your tests a little.

We can point back to this PR if anyone is looking to speed up the windowdiff algortihm.

```python
import time


def windowdiff_original(seg1, seg2, k, boundary="1", weighted=False):
    """Original O(n*k) — directly mirrors Pevzner & Hearst (2002)."""
    wd = 0
    for i in range(len(seg1) - k + 1):
        ndiff = abs(seg1[i:i+k].count(boundary) - seg2[i:i+k].count(boundary))
        wd += ndiff if weighted else min(1, ndiff)
    return wd / (len(seg1) - k + 1.0)


def windowdiff_optimized(seg1, seg2, k, boundary="1", weighted=False):
    """Sliding window O(n) — avoids recomputing counts each step."""
    wd = 0
    count1 = seg1[:k].count(boundary)
    count2 = seg2[:k].count(boundary)
    for i in range(len(seg1) - k + 1):
        ndiff = abs(count1 - count2)
        wd += ndiff if weighted else min(1, ndiff)
        if i + k < len(seg1):
            count1 += (seg1[i + k] == boundary) - (seg1[i] == boundary)
            count2 += (seg2[i + k] == boundary) - (seg2[i] == boundary)
    return wd / (len(seg1) - k + 1.0)


def make_segs(n):
    seg1 = ("0" * 49 + "1") * (n // 50) + "0" * (n % 50)
    seg2 = ("0" * 24 + "1" + "0" * 24 + "0") * (n // 50) + "0" * (n % 50)
    return seg1[:n], seg2[:n]


print(f"{'n':>8s} {'k':>5s} {'original':>12s} {'optimized':>12s} {'speedup':>8s}")
print("-" * 50)

for n, k in [
    (100, 3), (1000, 3), (10000, 3),        # small k (typical)
    (1000, 100), (10000, 100), (10000, 500), # large k (extreme)
    (100000, 3), (100000, 1000),             # very large n
]:
    seg1, seg2 = make_segs(n)
    if k > n:
        continue

    reps = max(1, 1000000 // (n * k))

    start = time.perf_counter()
    for _ in range(reps):
        windowdiff_original(seg1, seg2, k)
    t_orig = (time.perf_counter() - start) / reps * 1000

    start = time.perf_counter()
    for _ in range(reps):
        windowdiff_optimized(seg1, seg2, k)
    t_opt = (time.perf_counter() - start) / reps * 1000

    speedup = t_orig / t_opt if t_opt > 0 else float('inf')
    print(f"{n:>8d} {k:>5d} {t_orig:>10.3f}ms {t_opt:>10.3f}ms {speedup:>7.1f}x")
```

[out]:

```
         n     k     original    optimized  speedup
  --------------------------------------------------
       100     3      0.082ms      0.067ms     1.2x
      1000     3      0.846ms      0.640ms     1.3x
     10000     3      8.453ms      7.863ms     1.1x
      1000   100      0.850ms      0.629ms     1.4x
     10000   100      9.369ms      6.454ms     1.5x
     10000   500     15.560ms      6.157ms     2.5x
    100000     3     73.759ms     64.419ms     1.1x
    100000  1000    185.972ms     62.755ms     3.0x
```

  The optimization only shows meaningful speedup (>2x) at k=500+ which is unrealistic — typical k values are 3-10 (half the average segment length). At k=3 even with 100k characters it's only 1.1x.

alvations (approved):
Thanks @ihitamandal for the optimization discussion and the additional unittests! The latest commit that reverts the algorithm but kept the tests LGTM!

alvations:
[CI: retest]

alvations:
https://github.com/nltk/nltk/actions/runs/23360255484

stevenbird:
Thanks @ihitamandal for your contribution especially the unit tests. And thanks @alvations for reviewing, and for pointing a way forward that aligns with NLTK's pedagogical goal.

Issue #3294, opened 2024-08-12
[BUG] NLTK's PunktTokenizer fails to initialize due to UnicodeDecodeError [Windows specific]


soras:
Recently, PunktTokenizer's data loading has been changed to [load data from tab files](https://github.com/nltk/nltk/pull/3283); however, the change introduces `UnicodeDecodeError` on Windows platform, where the default (file) encoding is not 'utf-8'. 

Code example:
```python
>>> from nltk.tokenize.punkt import PunktTokenizer
>>> base_sentence_tokenizer = PunktTokenizer(lang='estonian')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\site-packages\nltk\tokenize\punkt.py", line 1744, in __init__
    self.load_lang(lang)
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\site-packages\nltk\tokenize\punkt.py", line 1750, in load_lang
    self._params = load_punkt_params(lang_dir)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\site-packages\nltk\tokenize\punkt.py", line 1770, in load_punkt_params
    params.ortho_context = pdec.tab2intdict(f)
                           ^^^^^^^^^^^^^^^^^^^
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\site-packages\nltk\tabdata.py", line 108, in tab2intdict
    return defaultdict(int, {a: int(b) for a, b in self.tab2tups(f)})
                                                   ^^^^^^^^^^^^^^^^
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\site-packages\nltk\tabdata.py", line 50, in tab2tups
    return [self.tab2tup(rm_nl(x)) for x in f]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\site-packages\nltk\tabdata.py", line 50, in <listcomp>
    return [self.tab2tup(rm_nl(x)) for x in f]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Programmid\Miniconda3\envs\py311_devel\Lib\encodings\cp1257.py", line 23, in decode
[...]

PR #3296, opened 2024-08-13
Fix for https://github.com/nltk/nltk/issues/3294

soras:


ekaf (approved):
Thanks @soras, this should indeed solve #3294.

ekaf:
@soras, please note that CI fails. The reason is that pyupgrade modifies your file: it removes the "r" parameter of _open_ (I don't know why it likes it better that way). After this change you need to call "git add" and "git commit" again.

soras:
> @soras, please note that CI fails. The reason is that pyupgrade modifies your file: it removes the "r" parameter of _open_ (I don't know why it likes it better that way). After this change you need to call "git add" and "git commit" again.

Thanks for the notice. I don't have experience with your workflow. I see CI failing, but I don't see how pyupgrade is rewriting the file. Is this information available only for maintainers? 

Anyway, I'll follow your guide and see if I can improve the code.

ekaf:
> I don't see how pyupgrade is rewriting the file. Is this information available only for maintainers?

@soras, pyupgrade rewrites the file on your local machine, the first time you try to commit it, so you should be able to see the change locally. I am not aware of any information that would be available only to maintainers: as far as I know, everything is open.

ekaf (approved):

PR #3304, opened 2024-08-16
Add doctests in data.py

ekaf:
This PR complements #3302, by adding switch_ functions with doctests for all the remaining pickles, in order to fully document the transition away from the old pickled classes.


ekaf:
The many CI failures are not related to this PR, but are caused by the recent breaking change in the output order of all WordNet relations.

Issue #3305, opened 2024-08-16
nltk.find:No such file or directory: '/root/nltk_data/tokenizers/punkt/PY3_tab'


hpx502766238:
nltk 3.8.1
```
import nltk
nltk.set_proxy('127.0.0.1:7890')
nltk.download('punkt_tab')
nltk.find('tokenizers/punkt_tab')

```
[nltk_data] Downloading package punkt_tab to /root/nltk_data...
[nltk_data]   Unzipping tokenizers/punkt_tab.zip.
Traceback (most recent call last):
  File "/mnt/vdb1/langchain-knowledge-base/1.py", line 4, in <module>
    nltk.find('tokenizers/punkt_tab')
  File "/mnt/vdb1/langchain-knowledge-base/venv/lib/python3.10/site-packages/nltk/data.py", line 537, in find
    return FileSystemPathPointer(p)
  File "/mnt/vdb1/langchain-knowledge-base/venv/lib/python3.10/site-packages/nltk/compat.py", line 41, in _decorator
    return init_func(*args, **kwargs)
  File "/mnt/vdb1/langchain-knowledge-base/venv/lib/python3.10/site-packages/nltk/data.py", line 312, in __init__
    raise OSError("No such file or directory: %r" % _path)
OSError: No such file or directory: '/root/nltk_data/tokenizers/punkt/PY3_tab'

hpx502766238:
> nltk 3.8.1
> 
> ```
> import nltk
> nltk.set_proxy('127.0.0.1:7890')
> nltk.download('punkt_tab')
> nltk.find('tokenizers/punkt_tab')
> ```
> 
[...]

PR #3306, opened 2024-08-17
Don't add PY3 in data path

ekaf:
Drop _data.py_ support for the added PY3 folder in the old pickle packages. 
This is related to issue #3305, which would not happen anymore anyway, because support is already dropped in _compat.py_,


ekaf:
Sorted the last wordnet tests, so that CI succeeds.

stevenbird:
Thanks @ekaf great catch

Issue #3285, opened 2024-07-12
nltk.metrics.aline feature map missing IPA symbols


TomerYS:
For nltk.metrics.aline there seem to be missing some IPA vowels like "ʌ" or "ʊ" in the feature map.

WilliamPLaCroix:
Created a pull request adding missing vowels from [Wikipedia IPA vowel chart](https://en.wikipedia.org/wiki/IPA_vowel_chart_with_audio) here in [issue #3323](https://github.com/nltk/nltk/pull/3323)

stevenbird:
Resolved in https://github.com/nltk/nltk/commit/9a5622f8a5b228df9499cd03181d9f8491e39f17

PR #3323, opened 2024-09-12
Adressing issue "nltk.metrics.aline feature map missing IPA symbols #3285"

WilliamPLaCroix:
Added several vowel entries to `metrics/aline.py` under `feature_matrix` in order to conform with [Wikipedia IPA vowel chart](https://en.wikipedia.org/wiki/IPA_vowel_chart_with_audio).


TomerYS:
Thank you for the work.
It should be also considered to add "mid-low" and "mid-high" stats for the "similarity_matrix" as the IPA chart is divided into 5 main levels.

WilliamPLaCroix:
I considered that, but didn't want to make such a big change, since those
would have to be added to the calculation (maybe just "0.25" and "0.75",
respectively), and I didn't know how much that would affect the algorithm /
function.

stevenbird:
Thanks @WilliamPLaCroix.
I pushed aline.py separately.
I agree about fewer vowel heights.

WilliamPLaCroix:
Thank you @stevenbird.
Would you be willing to add me to the AUTHORS.md contributors list? I forgot to update that with the pull request.
"William LaCroix <https://github.com/WilliamPLaCroix>"

PR #3327, opened 2024-10-02
Prevent multiple parallel downloads of the same package

### LINKED ISSUE #3248: Downloader race condition with multiple processes
When calling `nltk.download` followed by model usage from multiple processes at once, at least four types of errors can happen:
- `Error with downloaded zip file`
- `[Errno 2] No such file or directory`
- `zipfile.BadZipFile: Bad CRC-32 for file`
- `LookupError`

I can reproduce this every time using the following script (but you may need to tweak the sleep time or number of processes to match your hardware and internet connection):
```
import os
import nltk
import time
import random
import shutil
import multiprocessing

def download_and_run():
    print("Downloading...")
    time.sleep(random.random() * 2)
    nltk.download('punkt', force=True, download_dir='models')
    nltk.data.path.insert(0, 'models')
    print(nltk.sent_tokenize("Some text to split. Another sentence."))

if __name__ == "__main__":
    shutil.rmtree('models', ignore_errors=True)

    processes = []
    for i in range(10):
        processes.append(multiprocessing.Process(target=download_and_run))

    for p in processes:
        p.start()

    for p in processes:
        p.join()
```

The issue can be solved by writing model files atomically, as proposed in this pull request: https://github.com/nltk/nltk/pull/3247.

ekaf:
Fix #3248 using file locks.

In #3248, @naktinis describes a situation where users launch multiple useless parallel downloads of the same file, and cause race conditions. Rather than encouraging this behaviour as #3247 does, this PR attacks the root cause of the problem by using locks to prevent the useless downloads.

So [this test](https://github.com/nltk/nltk/issues/3248#issue-2266412613) (updated to request the current _punkt_tab_ instead of the deprecated _punkt_ package) now succeeds:

```
import os
import nltk
import time
import random
import shutil
import multiprocessing

def download_and_run():
    time.sleep(random.random() * 2)
    nltk.download('punkt_tab', force=True, download_dir='models')
    nltk.data.path.insert(0, 'models')
    print(nltk.sent_tokenize("Some text to split. Another sentence."))

if __name__ == "__main__":
    shutil.rmtree('models', ignore_errors=True)

    processes = []
    for i in range(10):
        processes.append(multiprocessing.Process(target=download_and_run))

    for p in processes:
        p.start()

    for p in processes:
        p.join()
```

[nltk_data]   Package punkt_tab: creating lockfile
[nltk_data]   Package punkt_tab: acquired lock, age:0.0 sec.
[nltk_data] Downloading package punkt_tab to models...
[nltk_data]   Package punkt_tab: found locked, age:0.01 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:0.41 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:0.49 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:0.62 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:0.72 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:1.08 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:1.22 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:1.34 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Package punkt_tab: found locked, age:1.35 sec.
[nltk_data]   Package punkt_tab: preventing redundant download ...
[nltk_data]   Unzipping tokenizers/punkt_tab.zip.
[nltk_data]   Package punkt_tab: released lock, age:2.7 sec.
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']
['Some text to split.', 'Another sentence.']

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -455,6 +455,35 @@ class LockFile:
     used to ensure unique package downloads.
 
     Refreshes eventual stale lockfiles left over from a previous crash.
+
+    Example usage:
+    >>> import os
+    >>> from tempfile import gettempdir
+    >>> from time import sleep
+    >>> class Info:
+    ...     def __init__(self, id, size):
+    ...         self.id = id
+    ...         self.size = size
+    >>> info = Info('test_package', 1000)
+    >>> lockfile = LockFile(info)
+
+    # Test lockfile creation
+    >>> lockfile.exists()
+    False
+    >>> sum('acquired ' in  msg.info for msg in list(lockfile.acquire()))
+    1
+    >>> lockfile.exists()
+    True
+
+    # Test lockfile age
+    >>> lockfile.age() < 1  # Should be very recent
+    True
+
+    # Test lockfile release
+    >>> sum('released ' in  msg.info for msg in list(lockfile.release()))
+    1
+    >>> lockfile.exists()
+    False
     """
 
     def __init__(self, info):
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -170,6 +170,7 @@ import time
 import warnings
 import zipfile
 from hashlib import md5
+from tempfile import gettempdir
 from xml.etree import ElementTree
 
 try:
@@ -404,7 +405,7 @@ class UpToDateMessage(DownloaderMessage):
 
 
 class DownloadInfoMessage(DownloaderMessage):
-    """Another process is already downloading the package"""
+    """More information about the package download"""
 
     def __init__(self, package, info=""):
         self.package = package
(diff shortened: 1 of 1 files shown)


ekaf:
Checking the semlock using _get_value() fails with a "Not implemented" error on macOS. But _is_zero() works.

Update: we don't need real locks anyway, since booleans are sufficient, when they are shared across threads using a multiprocessing Manager.

Now, we use file locks instead, in case parallel downloads are launched from different platforms.

ekaf:
While preventing multiple parallel downloads of a file into the same local directory, should it still be possible to have parallel downloads of a package into multiple directories?
Unlike the former case, this might have some utility. But I don't think NLTK needs to support it, because once downloaded, users could distribute the file to multiple locations locally.

ekaf:
I am not satisfied with this PR, and find it especially problematic to make assumptions about the user's download speed. So I am currently exploring the possibility of monitoring the growth of the downloaded file instead. Meanwhile, this is likely to remain in draft stage for a long time, and I would much appreciate any comment.

PR #3335, opened 2024-10-21
fixed the synset.pos call

### LINKED ISSUE #3334: Missing procedure call in line 417
Line 417 of the file "nltk/app/wordnet_app.py" should look like this:

    elif synset.pos() == wn.ADJ or synset.pos() == wn.ADJ_SAT:

but instead looks like this:

    elif synset.pos() == wn.ADJ or synset.pos == wn.ADJ_SAT:

which will generate this error (complete with spelling mistake) :

    "Unhandles synset POS type: s"

drewvid:
Fix #3334: fixed the synset.pos call and a spelling mistake

### DIFF
--- a/nltk/app/wordnet_app.py
+++ b/nltk/app/wordnet_app.py
@@ -414,7 +414,7 @@ def get_relations_data(word, synset):
                 ),
             ),
         )
-    elif synset.pos() == wn.ADJ or synset.pos == wn.ADJ_SAT:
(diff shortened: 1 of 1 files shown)


ekaf (approved):
Perfect, this is an obvious fix, which I don't think anybody could have objections against.

ekaf:
Thanks @drewvid!

PR #3340, opened 2024-11-19
Update Anaconda Distribution reference

emmanuel-ferdman:
# PR Summary
Small PR - fixes the Anaconda Distribution reference in `web/install.rst`. Relevant page: https://www.nltk.org/install.html

### DIFF
--- a/web/install.rst
+++ b/web/install.rst
@@ -10,7 +10,7 @@ Setting up a Python Environment (Mac/Unix/Windows)
 
 Please go through this guide to learn how to manage your virtual environment managers before you install NLTK,  https://docs.python-guide.org/dev/virtualenvs/
 
-Alternatively, you can use the Anaconda distribution installer that comes "batteries included" https://www.anaconda.com/distribution/
+Alternatively, you can use the Anaconda distribution installer that comes "batteries included" https://www.anaconda.com/download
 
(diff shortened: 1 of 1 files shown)


emmanuel-ferdman:
@ekaf @stevenbird can you please review and merge this small PR when you get the chance? 🙌

ekaf:
Not sure this Url is needed. If it changes, it might be better to remove it, rather than having to maintain it.

emmanuel-ferdman:
@ekaf should I remove this whole line?

ekaf:
This is a very difficult question. I would wait to hear more knowledgeable opinions.

emmanuel-ferdman:
@ekaf can you please ping the maintainers or others knowledgeable about this section for their input? 🙏

ekaf:
Sure, @emmanuel-ferdman, but please understand that a careful review can take a very long time...

ekaf:
In my opinion, it is still very relevant to recommend Anaconda. But their possible landing pages are just not very good. For ex. their _download_ page gives the impression that providing an e-mail address is required. Maybe it is best to let the users search for Anaconda if they want it.

It is also a concern that problems with Anaconda are wrongly reported as Nltk issues.

PR #3344, opened 2024-12-06
Lesk: handling ties

GeneralPoxter:
In case of ties for overlap scores, the current implementation of Lesk defaults to the higher index sense (e.g. max(synset) returns the last sense in the synset).

To improve performance, this PR defaults to the lower index sense (which tends to be more frequent, since synsets are generally ordered by frequency) in case of ties.

In the case of the "bank" example given in the function's documentation, the new assigned sense would be Synset('depository_financial_institution.n.01'), which I believe is the correct sense of the word.

### DIFF
--- a/nltk/wsd.py
+++ b/nltk/wsd.py
@@ -44,7 +44,7 @@ def lesk(context_sentence, ambiguous_word, pos=None, synsets=None, lang="eng"):
 
     if not synsets:
         return None
-    
+
     sense = max(synsets, key=lambda ss: len(context.intersection(ss.definition().split())))
 
     return sense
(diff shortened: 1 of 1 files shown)


ekaf:
@GeneralPoxter, it makes sense to consider the frequencies. But then, you can't just call it _lesk_, because it is a variant of the original algorithm.

GeneralPoxter:
The original paper by Lesk does not specify a tiebreaking strategy, so the current nltk implementation is no closer to the original algorithm than this PR. However, there is [precedence in literature](http://www.lrec-conf.org/proceedings/lrec2004/pdf/219.pdf) to defaulting to the wordnet ordering when there is no overlap. When I mentioned "original implementation of Lesk" above, I was referring to nltk's current implementation and not the original algorithm itself.

ekaf:
Yes, the implementation in Nltk is already a Lesk variant, rather than the original algorithm, by using Wordnet. But then, there is an opportunity to also use Wordnet's frequency ordering, as this PR does.
It is not surprising that the more frequent sense has a higher probability of being correct.

stevenbird:
Thanks Jason.

PR #3348, opened 2024-12-18
Only import tkinter if a GUI is needed

### LINKED ISSUE #2949: Importing NLTK breaks PyTorch multiprocessing on Mac
Importing NLTK breaks PyTorch multi-processing on Mac (both Intel and M1 based). 

```python
import torch
from torch.utils.data import DataLoader, TensorDataset
import os
import torch.multiprocessing as mp

import nltk

if __name__ == "__main__":
    if mp.get_start_method(allow_none=True) is None and os.name != 'nt':
        mp.set_start_method('fork')

    train = TensorDataset(torch.ones((50, 5)))

    pipe = DataLoader(dataset=train, batch_size=5, num_workers=1)
    itr = iter(pipe)
    data = next(itr)
    print(data)
```

The above code will run fine if the (unused) nltk import is removed, but when it is present it will cause the worker threads to crash:

RuntimeError: DataLoader worker (pid 62735) is killed by signal: Trace/BPT trap: 5. 
RuntimeError: DataLoader worker (pid(s) 62735) exited unexpectedly

This conflict exists both with older versions of PyTorch (1.7.1) as well as the latest version (1.10.2). Tested with NLTK 3.6.7 and 3.7. This conflict does not appear on Linux.

samertm:
Only import tkinter if the user is calling a function that explicitly uses a GUI instead of at the top-level.

This avoids the issue on macOS where calling fork after importing nltk without using any gui-related features causes the program to segfault.

Also adds the env var NLTK_DOWNLOADER_FORCE_INTERACTIVE_SHELL which forces the use of the interactive shell when using interactive downloader if set to 'true'.

Fixes #2949.

Tested that the reproducing program no longer crashes:

```
import torch
from torch.utils.data import DataLoader, TensorDataset
import os
import torch.multiprocessing as mp

import nltk

if __name__ == "__main__":
    if mp.get_start_method(allow_none=True) is None and os.name != 'nt':
        mp.set_start_method('fork')

    train = TensorDataset(torch.ones((50, 5)))

    pipe = DataLoader(dataset=train, batch_size=5, num_workers=1)
    itr = iter(pipe)
    data = next(itr)
    print(data)
```

Tested that the nltk downloader GUI still works when explicitly requested.

Tested that the nltk.sem.drt draw demos still work.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -303,6 +303,7 @@
 - Ron Urbach <https://github.com/sharpblade4>
 - Vivek Kalyan <https://github.com/vivekkalyan>
 - Tom Strange https://github.com/strangetom
+- Samer Masterson <https://github.com/samertm>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/__init__.py
+++ b/nltk/__init__.py
@@ -19,6 +19,7 @@ isort:skip_file
 """
 
 import os
+import importlib
 
 # //////////////////////////////////////////////////////
 # Metadata
@@ -169,7 +170,6 @@ draw = lazyimport.LazyModule("draw", locals(), globals())
 toolbox = lazyimport.LazyModule("toolbox", locals(), globals())
 
 # Optional loading
-
 try:
     import numpy
 except ImportError:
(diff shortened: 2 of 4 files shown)


stevenbird:
Thanks @samertm

PR #3352, opened 2025-01-02
Auto update the year in copyright notice.

naamanu:
There are many other places that still have the year 2024, is there a timeline to update those or should i update those as well?


stevenbird:
Thanks @claeusdev, but we have a script for this.

PR #3367, opened 2025-02-14
Test fixes

pdeblanc:
I found three failing test cases when running tests locally on my Mac M1. Here are some fixes.

- The generation doctest was expecting an exception to be raised when the supplied grammar is recursive, but this behavior was changed by commit 7250f930b7 (over a year ago), so I've deleted the test.
- The probability doctest was returning `0.08433050215340412` instead of `0.08433050215340411` when running on my machine. I'm not sure whether this is a flaky test that depends on the exact hardware, or whether some arithmetic operations were refactored at some point, but I am guessing that reducing the precision of the test would be acceptable.
- The tree pretty-printing doctest was producing a horizontally-flipped output on my machine. I am guessing that the branch ordering was nondeterministic, so I sorted the branches for the test.


ekaf:
Thanks @pdeblanc for catching these failures. I am surprised that they were not caught by CI.
Concerning  _generate.doctest_, the test is not needed, but it is not really obsolete, because you can still trigger the error by setting a sufficiently high _depth_, as here:

```
Infinite grammars will throw a RecursionError if ``depth`` is too high:

    >>> grammar = CFG.fromstring("""
    ... S -> A B
    ... A -> B
    ... B -> "b" | A
    ... """)
    >>> import sys
    >>> list(generate(grammar, depth=sys.getrecursionlimit()//2))
    ...
    RuntimeError: The grammar has rule(s) that yield infinite recursion!
    Eventually use a lower 'depth', or a higher 'sys.setrecursionlimit()'.

```

PR #3371, opened 2025-02-25
Add support for mixed rules conversion into Chomsky Normal Form

josecols:
Hello!

I'm a student in computational linguistics and recently noticed that `CFG.chomsky_normal_form()` can't handle mixed rules; this PR adds support for this. The implementation is based on Section 18.6.1 from [Speech and Language Processing](https://web.stanford.edu/~jurafsky/slp3/).

**What is changing?**
* I added a new method called `remove_mixed_rules` that converts hybrid rules containing terminals and non-terminals into rules with only non-terminals. The lexical items are replaced with *dummy* non-terminals, which are then added as unitary lexical productions.
* The `CFG.chomsky_normal_form()` now includes a call to this method.
* I added myself to `AUTHORS.md` as per contribution guidelines.

**Tests**
* I added a new section to `grammar.doctest`; however, I noticed that there is also a relevant unit test module, `test_cfg2chomsky`. Please let me know if you prefer the latter, and I will adjust the tests accordingly.
* All `develop` passing tests passed when running `tox -e py312`.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -307,6 +307,7 @@
 - Samer Masterson <https://github.com/samertm>
 - William LaCroix <https://github.com/WilliamPLaCroix>
 - Peter de Blanc <https://github.com/pdeblanc>
+- Jose Cols <https://github.com/josecols>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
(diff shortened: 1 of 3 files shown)


stevenbird:
Great first contribution, thanks @josecols!

josecols:
Thank you, @stevenbird!

PR #3378, opened 2025-03-11
Document how to reproduce old Wordnet studies

### LINKED ISSUE #3377: Reproduce old Wordnet results
Several researchers (for ex. @fcbond, @jmccrae, @goodmami, @stevenbird) have expressed their interest in using NLTK for reproducing scientific results with older Wordnet data. Fortunately, only small edits are necessary for NLTK to load any Wordnet in the original Princeton WordNet _wndb_ format. This could for ex. be a Princeton WordNet from the 1.x or 2.x series, which were never included in NLTK, or any Open English Wordnet version. 
Let's illustrate the process with Edition 2023 of the Open English Wordnet, since _nltk_data_ does not include it.

1. Get the data package. The 2023 Edition is at https://en-word.net/static/english-wordnet-2023.zip

2. Rename the package to _oewn2023.zip_ and copy it to the _corpora_ subdirectory of your _nltk_data_ directory.

Renaming the package is necessary because _english-wordnet-2023.zip_ creates an _oewn2023_ subdirectory, while NLTK expects the data package to have the same name as the subdirectory.
Alternatively, you can eliminate the need for renaming the package, by just unzipping it so that you have a _nltk_data/corpora/oewn2023_ directory.

3. Add an entry in _nltk/corpus/\_\_init\_\_.py_. The easiest way to do this is to copy one of the existing Wordnet entries in that file, and edit the name in two places:

```
oewn2023: WordNetCorpusReader = LazyCorpusLoader(
    "oewn2023",
    WordNetCorpusReader,
    LazyCorpusLoader("omw-1.4", CorpusReader, r".*/wn-data-.*\.tab", encoding="utf8"),
)
```

4. Enjoy:

```
from nltk.corpus import oewn2023 as ewn
print(ewn.get_version())
```
2023

```
print(ewn.lemmas('book')[0])
```
Lemma('book.n.01.book')

ekaf:
Fix #3377. This PR adds a template for creating a custom wordnet entry, and explanations about how to load old wordnet data in the _wndb_ format.

### DIFF
--- a/nltk/corpus/__init__.py
+++ b/nltk/corpus/__init__.py
@@ -395,6 +395,14 @@ wordnet: WordNetCorpusReader = LazyCorpusLoader(
     WordNetCorpusReader,
     LazyCorpusLoader("omw-1.4", CorpusReader, r".*/wn-data-.*\.tab", encoding="utf8"),
 )
+## Use the following template to add a custom Wordnet package.
+## Just uncomment, and replace the identifier (my_wordnet) in two places:
+##
+# my_wordnet: WordNetCorpusReader = LazyCorpusLoader(
+#    "my_wordnet",
+#    WordNetCorpusReader,
+#    LazyCorpusLoader("omw-1.4", CorpusReader, r".*/wn-data-.*\.tab", encoding="utf8"),
+# )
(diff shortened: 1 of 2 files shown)


stevenbird:
Excellent, thanks @ekaf

PR #3386, opened 2025-03-28
Corrected ResolutionProver.find_answers

### LINKED ISSUE #3046: ResolutionProver find_answers() finds answers even when there are no answers
I came across this question:
https://stackoverflow.com/questions/66666291/python-nltk-resolutionprover-finds-a-solution-when-in-fact-there-is-no
and it seems to me that the reason is pretty simple.
The function find_answers() checks for ANSWER() but does not check that the empty clause is derived.
The current code is like:
```
def find_answers(self, verbose=False):
        self.prove(verbose)

        answers = set()
        answer_ex = VariableExpression(Variable(ResolutionProver.ANSWER_KEY))
        for clause in self._clauses:
            for term in clause:
                if (
                    isinstance(term, ApplicationExpression)
                    and term.function == answer_ex
                    and not isinstance(term.argument, IndividualVariableExpression)
                ):
                    answers.add(term.argument)
        return answers
```
whereas it should be like:
```
    def find_answers(self, verbose=False):
        self.prove(verbose)

        answers = set()
        answer_ex = VariableExpression(Variable(ResolutionProver.ANSWER_KEY))
        for clause in self._clauses:
           if (len(clause) == 1
                and isinstance(clause[0], ApplicationExpression)
                and clause[0].function == answer_ex
                and not isinstance(clause[0].argument, IndividualVariableExpression)
            ):
                answers.add(clause[0].argument)
        return answers
```
That is: there should be unit clauses where the sole literal is ANSWER().

eidheim:
Fixes #3046: corrected ResolutionProver.find_answers such that incorrect clauses no longer are added as answers.

### DIFF
--- a/nltk/inference/resolution.py
+++ b/nltk/inference/resolution.py
@@ -141,13 +141,13 @@ class ResolutionProverCommand(BaseProverCommand):
         answers = set()
         answer_ex = VariableExpression(Variable(ResolutionProver.ANSWER_KEY))
         for clause in self._clauses:
-            for term in clause:
-                if (
-                    isinstance(term, ApplicationExpression)
-                    and term.function == answer_ex
-                    and not isinstance(term.argument, IndividualVariableExpression)
-                ):
-                    answers.add(term.argument)
+            if (
+                len(clause) == 1
+                and isinstance(clause[0], ApplicationExpression)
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @eidheim!

PR #3420, opened 2025-09-17
ci: upgrade github actions

purificant:
This PR upgrades GitHub Actions used in CI to their respective latest versions. This is routine maintenance to keep GitHub Actions up to date, ensuring we benefit from the latest security patches, performance improvements, and bug fixes.

  - [actions/checkout](https://github.com/actions/checkout) v4 -> v5
  - [actions/setup-python](https://github.com/actions/setup-python) from v5 -> v6
  - [actions/setup-java](https://github.com/actions/setup-java) from v4 -> v5
  - [actions/labeler](https://github.com/actions/labeler) from v5 -> v6

PR #3423, opened 2025-09-17
ci: automate ci github action updates with dependabot

purificant:
This PR automates commits to bump GitHub Actions. It's a replacement for PRs like: https://github.com/nltk/nltk/pull/3420
Thank you @cclauss for the suggestion

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  # Enable version updates for GitHub Actions
+  # https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    open-pull-requests-limit: 5
+    groups:
+      # Group all GitHub Actions updates into a single PR
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: ci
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @purificant and @cclauss

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+updates:
+  # Enable version updates for GitHub Actions
+  # https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    open-pull-requests-limit: 5
```
cclauss, line 9: Why set a limit of five when grouping guarantees that there will never be more than one?

purificant: Correct. It's effectively a no-op with the current setup. We should get at most a single PR for the `github-actions` group. I left it there as documentation / future-proofing, so it's obvious we won't get spammed with PRs, in case we add more groups or ecosystems, but happy to drop it if you’d prefer.

cclauss: We can leave it as is.  The limit only pertains to `package-ecosystem: github-actions`.

purificant: Exactly, it's scoped to just `github-actions` which also has grouping into a single PR, ergo this line is no-op.

Issue #3308, opened 2024-08-18
nltk 3.9.0: Lookup error on import "Resource wordnet not found"


danjac:
Python 3.12, nltk 3.9

Following instructions to download corpora, immediately ran into this issue on either running `import nltk`  or `python -m nltk.download`

```
LookupError: 
**********************************************************************
  Resource wordnet not found.
  Please use the NLTK Downloader to obtain the resource:

  >>> import nltk
  >>> nltk.download('wordnet')
  
  For more information see: https://www.nltk.org/data.html

  Attempted to load corpora/wordnet
```

This was working in nltk `3.9b1`.

michbsd:
Yep. I am seeing this too - it fails on the import - so no chance to download 'wordnet'


```
[...]
          import nltk
   File "/usr/local/lib/python3.9/site-packages/nltk/__init__.py", line 153, in <module>
     from nltk.translate import *
   File "/usr/local/lib/python3.9/site-packages/nltk/translate/__init__.py", line 24, in <module>
     from nltk.translate.meteor_score import meteor_score as meteor
   File "/usr/local/lib/python3.9/site-packages/nltk/translate/meteor_score.py", line 14, in <module>
     from nltk.stem.api import StemmerI
   File "/usr/local/lib/python3.9/site-packages/nltk/stem/__init__.py", line 34, in <module>
     from nltk.stem.wordnet import WordNetLemmatizer
   File "/usr/local/lib/python3.9/site-packages/nltk/stem/wordnet.py", line 13, in <module>
     class WordNetLemmatizer:
   File "/usr/local/lib/python3.9/site-packages/nltk/stem/wordnet.py", line 48, in WordNetLemmatizer
     morphy = wn.morphy
   File "/usr/local/lib/python3.9/site-packages/nltk/corpus/util.py", line 120, in __getattr__
     self.__load()
   File "/usr/local/lib/python3.9/site-packages/nltk/corpus/util.py", line 86, in __load
     raise e
   File "/usr/local/lib/python3.9/site-packages/nltk/corpus/util.py", line 81, in __load
     root = nltk.data.find(f"{self.subdir}/{self.__name}")
   File "/usr/local/lib/python3.9/site-packages/nltk/data.py", line 579, in find
     raise LookupError(resource_not_found)
 LookupError:
 **********************************************************************
   Resource wordnet not found.
   Please use the NLTK Downloader to obtain the resource:

   >>> import nltk
   >>> nltk.download('wordnet')

[...]

PR #3425, opened 2025-09-23
Require a minimal download test before running cached-data tests in CI

ekaf:
Prevents issues like #3308.

This PR updates the CI workflow to address issues where NLTK fails to run if required resources (such as `wordnet`) are not downloaded, as reported in #3308. The changes ensure that the main test suite, which relies on cached data, only runs if a minimal download test for a core NLTK resource (`wordnet`) succeeds on a fresh install.

Key changes:
- Added a `minimal_download_test` job that installs NLTK, downloads the `wordnet` corpus to a temporary directory, and verifies its usability.
- The main `test` job now depends on this minimal download test, as well as the cache jobs.
- If the minimal download test fails (e.g., due to a download or resource usage error), the cached-data tests will be skipped.
- This helps catch issues early when NLTK cannot download or load essential corpora on first use, preventing false positives from running only on pre-populated cached data.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,8 @@ jobs:
       - run: |
           pip install --upgrade pip
           pip install --upgrade nltk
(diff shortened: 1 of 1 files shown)


stevenbird:
Great idea!

Issue #3429, opened 2025-10-01
There is not enough delay after launching a CoreNLP server and before testing whether the process exited prematurely


asrelo:
The method `start` of class `CoreNLPServer` from module `nltk.parse.corenlp` launches a CoreNLP server as a separate process and then [checks](https://github.com/nltk/nltk/blob/3.9.2/nltk/parse/corenlp.py#L132) its status: 1. by an immediate test for premature exit, 2. by poking some of its endpoints repeatedly.

I would assume that the server's failing to bind to a port and exiting early counts as a premature exit. However, the CoreNLP server preloads its annotators before proceeding to bind to a port. On my mid-tier machine, it takes the CoreNLP server from 1 to 30 seconds to get through initialization to binding the socket (the latter - when preloading [all listed annotators](https://stanfordnlp.github.io/CoreNLP/annotators.html)). There is no explicit delay in the NLTK code before the test for premature exit. This leaves this kind of error - often caused by wrong options specified by a user - to be detected later when poking the endpoints, and the error reporting there is bad - the user doesn't get to know the cause for "Could not connect".

(In particular, I was investigating some weird behavior of the NLTK+CoreNLP combo caused by some other service - `ipykernel` launched by the Jupyter extension for VS Code - taking the same TCP port earlier in a run, and this behavior greatly delayed the resolution. I spent 2 days wondering why does my "CoreNLP server" behave like it's possessed by an evil spirit only when run from NLTK.)

I am not ready to make a pull request specifically because it is uncertain how much time to assume as CoreNLP initialization time (see above).
* One solution is to put a short delay there. However, this will make the behavior (the particular error reported by NLTK) erratic when preloading many annotators.
* Another solution is to repeatedly test for when the network port given to CoreNLP will actually be taken, and proceed as soon as it is (or fail on timeout). Doing this by actively trying to bind to a port, like NLTK already does to check that this port is free beforehand, may cause a race condition between NLTK and CoreNLP. (See [this issue](https://github.com/nltk/nltk/issues/3428) before implementing!)
* Third solution is to read CoreNLP's output on `stderr` for a particular line (like `StanfordCoreNLPServer listening`). Not very robust either.

Issue #3428, opened 2025-10-01
The function to test whether a port for CoreNLP server is free is used incorrectly or returns wrong result in common scenarios


asrelo:
The function [`try_port`](https://github.com/nltk/nltk/blob/3.9.2/nltk/parse/corenlp.py#L30) in module `nltk.parse.corenlp` is used to test that a given local TCP port if free before launching a CoreNLP server on it. It does that by trying to `bind` a socket to an empty address `''` at the given port.

I don't exactly know what address `''` is supposed to represent. I know that in my environment on Windows (Win 10, Python 3.12.10) it turns out to be synonym to `0.0.0.0`. I also know that I can bind sockets simultaneously without an error to `localhost` (`127.0.0.1`) and `0.0.0.0`. That means that when another process already holds `localhost:PORT`, `try_port` reports that `0.0.0.0:PORT` is free. Then the code tries to run CoreNLP on `localhost:PORT` and fails.

(In particular, I was investigating some weird behavior of the NLTK+CoreNLP combo caused by some other service taking the same TCP port earlier in a run, and the erroneous behavior of this function delayed the resolution.)

I feel too confused by the inconsistent handling of `0.0.0.0` in OSs to propose a definitive solution. IMO it might be worth it to change `try_port` to try the given port on `localhost` instead of `''`, since that is the hard-coded address that NLTK launches a CoreNLP server on. (I also don't see a way to launch a CoreNLP server on an address other than `localhost`.)

PR #3430, opened 2025-10-01
Stop the CoreNLP server if a test connection to it cannot be established

asrelo:
The method `start` of class `CoreNLPServer` from module `nltk.parse.corenlp` launches a CoreNLP server as a separate process and then [checks](https://github.com/nltk/nltk/blob/3.9.2/nltk/parse/corenlp.py#L141) its status: 1. by a test for premature exit, 2. by poking some of its endpoints repeatedly.

If any of the endpoints test fails, an exception is raised. However, the server process is not shut down; and as the `popen` attribute of the class does not appear to be the part of intended public interface, the track of the started server process is lost. This causes a resources leak, primarily from potential bound to the given network port.

This PR makes it so the server process is terminated if a test connection to it cannot be established to avoid the resources leak. It does not even wait on the process to terminate, just sends the signal.

***This is an opinionated change***; some would consider leaving the process up for investigation by user instead (i.e. leaving the code as is), or leaving an option to do it, or even exposing the process as part of interface.

### DIFF
--- a/nltk/parse/corenlp.py
+++ b/nltk/parse/corenlp.py
@@ -138,27 +138,33 @@ class CoreNLPServer:
                 "The error was: {}".format(stderrdata.decode("ascii")),
             )
 
-        for i in range(30):
-            try:
-                response = requests.get(requests.compat.urljoin(self.url, "live"))
-            except requests.exceptions.ConnectionError:
-                time.sleep(1)
-            else:
-                if response.ok:
-                    break
-        else:
-            raise CoreNLPServerError("Could not connect to the server.")
+        try:
 
-        for i in range(60):
-            try:
-                response = requests.get(requests.compat.urljoin(self.url, "ready"))
-            except requests.exceptions.ConnectionError:
-                time.sleep(1)
+            for i in range(30):
+                try:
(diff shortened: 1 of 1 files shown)


ekaf:
This PR improves the reliability of CoreNLP server startup by ensuring that failed launches are properly cleaned up. It looks good to merge as it is.

For completeness, here are a few possible future enhancements (none of which are blocking for this PR): more thorough tests for process termination, configurable timeouts, concurrency safety, and additional logging.

I also note that two new issues (#3428 and #3429) have been opened by the same author regarding CoreNLP startup problems:
- #3428 highlights that the port-checking logic may be unreliable on some platforms, potentially resulting in misleading port availability reports.
- #3429 points out that the current startup logic may check for CoreNLP process health too quickly, due to annotator preloading, and suggests improvements for error reporting and detection.

While this PR addresses cleanup specifically, the issues raised in #3428 and #3429—regarding port detection and startup timing—seem closely related. Coordinating future fixes for these CoreNLP startup aspects may be worthwhile, but merging this PR independently makes sense.

Thanks for helping to improve CoreNLP integration in NLTK!

stevenbird:
Thanks @asrelo! And for your feedback @ekaf.

PR #3431, opened 2025-10-03
When a test connection to a CoreNLP server cannot be established, give a user the choice to keep the server process running

asrelo:
I apologize it took me way longer to figure out a potentially better solution to the problem posed in #3430 (tldr: `CoreNLPServer` starts a separate server process but does not shut it down if the process fails a status check, resulting in a resource leak).

The class `CoreNLPServer` has two interfaces: explicit `start`/`stop` methods and the context manager interface. Currently the `__enter__`/`__exit__` methods just run the `start`/`stop` methods. With this PR:

1. `start` will **not** stop the server process in case of status check failure (pre-#3430 behavior);
2. `__enter__` *will* stop the server process in case the status checks in `start` fail.

This way a user can choose between automatic cleanup (with the context manager interface) and manual cleanup. To me, it kind of makes sense to leave the latter option, since the started server *is* running, NLTK just cannot connect to it via HTTP or the server doesn't report it's ready. Also, this variant of the fix also affects less users - only those who use the context manager interface. The PR includes updated docstrings.

I still believe the optimal solution to the resources leak is ***uncertain***, given that there is currently no way for a user to get logs from a failed CoreNLP server process &mdash; all they can get from NLTK in this scenario is a generic exception, so I don't *insist* on this solution. Note also that this variant of the fix runs the `stop` method to stop a failed server process which does wait on it to exit.

### DIFF
--- a/nltk/parse/corenlp.py
+++ b/nltk/parse/corenlp.py
@@ -101,9 +101,17 @@ class CoreNLPServer:
         self.java_options = java_options or ["-mx2g"]
 
     def start(self, stdout="devnull", stderr="devnull"):
-        """Starts the CoreNLP server
+        """Start the CoreNLP server
+
+        This method checks the status of the started server, but does **not** stop
+        the server process if those checks fail. If you want the server process
+        to be stopped in that case, either use this class as a context manager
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @asrelo!

Issue #3195, opened 2023-10-20
Add a function of splitting combined words.


wxz:
Add a function of splitting combined words. 
For example, the combined word "crossregionswitch" can be split into three words: "cross", "region", and "switch".

PR #3437, opened 2025-10-26
Added WordNetCompoundSplitter

Danial-Changez:
## Title
Added `WordNetCompoundSplitter` to address #3195 

## Summary
- Introduced `compound.py` with a WordNet splitter, favouring whole words, but splits true compounds.
- Exported the new helper through `nltk.tokenize` and added `test_compound_splitter.py` for testing.

## Test Results
- Test Command
```bash
python -m unittest nltk.test.unit.test_compound_splitter
```
- Results
image

### DIFF
--- a/nltk/test/unit/test_compound_splitter.py
+++ b/nltk/test/unit/test_compound_splitter.py
@@ -1,0 +1,55 @@
+import unittest
+
+from nltk.tokenize import WordNetCompoundSplitter, split_compound
+
+try:
+    from nltk.corpus import wordnet as wn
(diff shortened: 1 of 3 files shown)


ekaf:
Unnecesary additiion; it is better for users to handle this task according to their own preferences, after an evaluation of different approaches, which is missing here.

stevenbird:
Thanks @Danial-Changez for your proposed contribution. Unfortunately I have to agree with @ekaf, that this is a user-level task.

PR #3438, opened 2025-10-27
Added test-depth consistency, test-root hypernyms, and test-root case

LanceBaysa:
From the user’s point of view, the use case can be summarized as follows: the primary actor is the user, whose goal is to efficiently retrieve hierarchical relationship data, like hypernym paths, depth, and root hypernyms for a given synset. The user begins with a valid synset from the WordNet corpus and calls one of the hypernym-related methods. Instead of recalculating the hierarchy each time, the system computes all paths once, stores the results in memory, and retrieves the necessary information for subsequent calls. This ensures faster responses and eliminates redundant traversal.

### DIFF
--- a/nltk/test/unit/test_wordnet.py
+++ b/nltk/test/unit/test_wordnet.py
@@ -14,6 +14,33 @@ L = wn.lemma
 
 
 class WordnNetDemo(unittest.TestCase):
+    
+    def test_depth_consistency(self):
+        #test that depth calculations match actual path lengths
+        ss = wn.synset('sweet.n.01')
+        paths = ss.hypernym_paths()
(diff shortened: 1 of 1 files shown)


ekaf:
@LanceBaysa, you only made your tests accessible, the rest of your proposal is hidden.

alvations:
@LanceBaysa following up on @ekaf's comment, please commit your proposal on top of the test you've added. 

Also remove the project hash commit that you've accidentally added to this PR.

alvations (changes requested):
Please add the actual code proposal and changes. And also remove the file with the project has that was wrongly added to this PR.

stevenbird:
Closing, thanks all.

PR #3446, opened 2025-10-30
Bugfix/aline delta error

wmathias00:


alvations:
@wmathias00 thank you for the pull request. It'll help us to review if you can please describe what the PR is about.

alvations:
Without additional details, here's a quick review based on the git diff:



### Good ideas:
  - Input validation in `align()` — fail early with a clear ValueError instead of a cryptic KeyError deep in the algorithm. Worth adding.
  - `sigma_exp()` length check — minor safety guard, fine to add.

###  Bad ideas:
  - `delta()` rewrite with multi-character splitting — This is harmful. The feature_matrix has multi-codepoint keys like e̞, ø̞, ɤ̞ (base char + combining character). The new code splits these into individual codepoints and averages pairwise differences, which is completely wrong — e̞ is a single phonetic segment, not two. It also changes the function signature (adding unknown_penalty) and the algorithm semantics. delta is supposed to compare two single segments as per Kondrak (2002)
  - Defensive returns in diff() — Silently returning 1.0 for unknown segments masks bugs. If we validate at the boundary in `align()`, unknowns should never reach `diff()`. If they do, we want it to crash.
  - `_single_delta()` split — Unnecessary indirection created to support the broken multi-character `delta()`.

stevenbird:
Dropping this.

PR #3447, opened 2025-10-31
remove support for python 3.9 which is end of life on Oct 31 2025

purificant:
Python 3.9 reached end of life on October 31, 2025.
This PR removes support for Python 3.9 and sets the minimum required version to Python 3.10.

CI will now run tests for Python 3.10+ and all mentions of Python 3.9 have been removed from package metadata, workflows, test configuration, and documentation.

This PR is cleanup intended to help with changes necessary for Python 3.14+

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
     needs: [pre-commit, minimal_download_test]
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12']
         os: [ubuntu-latest, macos-latest, windows-latest]
       fail-fast: false
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9, 3.10, 3.11, 3.12]
+        python-version: ['3.10', '3.11', '3.12']
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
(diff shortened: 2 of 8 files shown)


ekaf:
What about removing tox.ini and all references to tox? Nltk seems to just use pytest.

purificant:
I've got no strong opinions about tox one way or another. Historically tox was useful to run tests locally and in CI for different python versions, it takes care of creating isolated virtual environments for specific python versions and running commands in the scope of those environments. This was very useful to have before GitHub Actions existed and before anyone could clone the repo and easily run all tests in the CI in the scope of their own clone. Personally these days I don't use tox much and rely on cloud based CI more but it's still a perfectly viable workflow if anyone else wants to use it.

To be perfectly clear, any changes to tox within the scope of this PR are specific to py3.9 and do not have any other objectives.

ekaf:
Yes, it makes sense to keep this PR focused, and leave the eventual removal of tox.ini for another PR. I am considering a translation of setup.py to a pyproject.toml file, and the tox question also comes up in that context.
Another thing to consider though, is the mismatch between README and classifiers (that include 3.13 as supported) vs. CI and release workflows (that currently test 3.10–3.12 only).

purificant:
:+1: pyproject.toml > setup.py

purificant:
> Another thing to consider though, is the mismatch between README and classifiers (that include 3.13 as supported) vs. CI and release workflows (that currently test 3.10–3.12 only).

This is a good point @ekaf. There are some discrepancies with mentions of py3.13. Ideally we should to move towards fully supporting 3.13 and 3.14. This PR is a step in that direction by focusing on py3.10+ only.

ekaf (approved):
This PR looks good to merge, considering its deliberately limited scope.

purificant:
Thank you for review @ekaf

Issue #3444, opened 2025-10-29
[Feature / Bug] Handle invalid n values gracefully in NIST score computation


EgodPrime:
## Description

The current implementation of nltk.translate.nist_score.sentence_nist fails with a ZeroDivisionError when `n` exceeds the length of the hypothesis sentence, raising a `ZeroDivisionError` due to division by zero in the final aggregation step.

This occurs because for n-gram orders greater than the length of the hypothesis, no valid n-grams exist, resulting in a denominator of 0.

Additionally, the function does not validate whether n < 1, which is also invalid.

This leads to:

- Unhandled exceptions;
- Poor user experience;
- Potential crashes in automated systems.

## Steps to Reproduce

```python
import nltk.translate.nist_score

references = [
    ['It', 'is', 'a', 'guide', 'to', 'action', 'that',
     'ensures', 'that', 'the', 'military', 'will', 'forever',
     'heed', 'Party', 'commands'],
    ['It', 'is', 'the', 'guiding', 'principle', 'which',
     'guarantees', 'the', 'military', 'forces', 'always', 'being',
     'under', 'the', 'command', 'of', 'the', 'Party'],
    ['It', 'is', 'the', 'practical', 'guide', 'for', 'the',
     'army', 'always', 'to', 'heed', 'the', 'directions',
     'of', 'the', 'party']
]

hypothesis = ['It', 'is', 'a', 'guide', 'to', 'action', 'which',
              'ensures', 'that', 'the', 'military', 'always',
              'obeys', 'the', 'commands', 'of', 'the', 'party']

# len(hypothesis) = 18
print(f"len(hypothesis)={len(hypothesis)}")

nltk.translate.nist_score.sentence_nist(references, hypothesis, 5)
print("n=5 is ok")

nltk.translate.nist_score.sentence_nist(references, hypothesis, 18)
print("n=18(upper limit) is ok")

# This raises ZeroDivisionError
nltk.translate.nist_score.sentence_nist(references, hypothesis, 19)

# # This also raises ZeroDivisionError
# nltk.translate.nist_score.sentence_nist(references, hypothesis, 0)
```

## Expected Behavior

The function should either:

1. **Gracefully handle** large `n` values by only considering n-grams up to the maximum possible length (i.e., min(n, len(hypothesis))) and skip invalid orders.
2. Or raise a **clear error** such as `ValueError("n must be <= max sentence length")`.

## Actual Behavior

```
ZeroDivisionError: division by zero
```

## Environment

[...]

Issue #3441, opened 2025-10-27
`nltk.util.elementtree_indent` should restrict parameter ranges in the documentation


EgodPrime:
Hi, I'm working on a Python library fuzzer, and it finds that the API `nltk.util.elementtree_indent(elem, level=0)` has no constraints or validations on the parameter `level` (Neither in the documentation nor in the code implementation). This may lead to some unsafe behavior.

My fuzzer generates a code snippet like the following:

```python
import nltk 
from xml.etree import ElementTree as ET 
# Create a sample element tree for demonstration 
root = ET.Element("root") 
child1 = ET.SubElement(root, "child1") 
child2 = ET.SubElement(root, "child2") 
grandchild = ET.SubElement(child1, "grandchild") 
grandchild.text = "Some text" 
# Indent the element tree 
nltk.elementtree_indent(root, level=1612054756)
print(root)
```

This code snippet may cause CPU overload on machines with poor hardware configurations, resulting in the entire process being terminated by the system. Alternatively, it may generate significant delays on machines with powerful hardware configurations.

To figure it out, I read the implementation of [`nltk.util.elementtree_indent(elem, level=0)`](https://www.nltk.org/_modules/nltk/util.html#elementtree_indent), which shows that:

```python
def elementtree_indent(elem, level=0):
    """
    Recursive function to indent an ElementTree._ElementInterface
    used for pretty printing. Run indent on elem and then output
    in the normal way.

    :param elem: element to be indented. will be modified.
    :type elem: ElementTree._ElementInterface
    :param level: level of indentation for this element
    :type level: nonnegative integer
    :rtype:   ElementTree._ElementInterface
    :return:  Contents of elem indented to reflect its structure
    """

    i = "\n" + level * "  "
    if len(elem):
        if not elem.text or not elem.text.strip():
            elem.text = i + "  "
        for elem in elem:
            elementtree_indent(elem, level + 1)
        if not elem.tail or not elem.tail.strip():
            elem.tail = i
    else:
        if level and (not elem.tail or not elem.tail.strip()):
            elem.tail = i
```

The content of variable `i` is generated by `level` (repeating `level` spaces) without verification, and this process is recursively called. This is the cause of high CPU usage, which worsens when the ElementTree is sufficiently deep.

[...]

PR #3450, opened 2025-11-04
Feature/enhance docstring and boundary check for NIST Score

EgodPrime:
As @ekaf said:

> "However, given NLTK’s educational focus, we may prefer to keep users in control of input validation rather than silently changing behavior or enforcing a particular policy in the library."

I have done the following patches:

1. Added a warning about performance issues with large indentation levels for `nltk.util.elementtree_indent`: #3441
2. Added boundary check for `nltk.translate.nist_score.sentence_nist` to raise clear errors for invalid input. #3444


EgodPrime:
commit https://github.com/nltk/nltk/pull/3450/commits/9fb361396bde97f767735d747806c2333433f15f has a failure on `pre-commit`. It seems my commit does not trigger `pre-commit` automatically; I've fixed the formatting issue and merged the latest commit of the `develop` branch.

I've run `pre-commit run --all-files` before commit https://github.com/nltk/nltk/pull/3450/commits/0beaf85137371fa00d40162263ca2c35e2e18ad4

```bash
$ pre-commit run --all-files
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/asottile/pyupgrade.
[INFO] Initializing environment for https://github.com/ambv/black.
[INFO] Initializing environment for local:isort==6.0.1.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/asottile/pyupgrade.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/ambv/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for local.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
fix utf-8 byte order marker..............................................Passed
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
fix requirements.txt.....................................................Passed
check yaml...............................................................Passed
pyupgrade................................................................Passed
black....................................................................Passed
isort....................................................................Passed
```

nltk/translate/nist_score.py
```diff
@@ -67,6 +67,15 @@ def sentence_nist(references, hypothesis, n=5):
     :param n: highest n-gram order
     :type n: int
     """
+    if n < 1:
+        raise ValueError(f"n must be at least 1, got n={n}")
```
alvations, line 71: This is actually not needed, when you pass in 0. The `for i in range(1, n + 1):` will just pass through. 

Perhaps checking for n < 0 would make sense but, I dont think it's valid use case to check for ngrams where n <0.

Would skip this check totally, a bit redudant.

nltk/util.py
```diff
@@ -1223,6 +1223,10 @@ def elementtree_indent(elem, level=0):
     used for pretty printing. Run indent on elem and then output
     in the normal way.
 
+    .. warning::
+        Using very large values for `level` (e.g., > 1000) may cause significant performance degradation due to deep recursion and excessive string operations.
+        This function is intended for human-readable output; use moderate levels only.
+
```
alvations, line 1277: This has nothing to do with NIST score. If needed, create another PR.

nltk/translate/nist_score.py
```diff
@@ -67,6 +67,15 @@ def sentence_nist(references, hypothesis, n=5):
...
+
+    if n > len(hypothesis):
+        max_hyp_len = len(hypothesis)
+        raise ValueError(
+            f"n={n} exceeds maximum possible n-gram length ({max_hyp_len}). "
+            f"Set n <= {max_hyp_len} to compute valid NIST score."
+        )
```
alvations, line 78: No this implementation doesn't follow NIST original implementation. The hypothesis cannot decide the n, if anything it should be the reference that decides. But the definition of NIST is to handle corpus level score so in any case, there's no real case for length to be constrained. 

I guess the issue is the divided by 0 error that is common when there is no denominator to divide by when there's no ngrams that matches the max N. 

In that case the fix wouldn't be to contrain the N. The right fix would be to do smoothing like Collin and Cherry suggest as implemented for BLEU or capture the error down at ` nist_precision_denominator_per_ngram[i]`

nltk/translate/nist_score.py
alvations, line 164: Maybe the change should happen here. 

```
precision, numerator, denominator, ref_len = max(nist_score_per_ref)
# No hypothesis n-grams of this order; nothing to accumulate.
if denominator  == 0:
    continue 
```

alvations (changes requested):
@EgodPrime thank you for your pull-request. Please revise the PR and address the following:

- Please remove the n < 1 , it's a bit redundant since 0 would still work only thing that it's guarding is negative n which is a bit illogical. 
- Don't skip the accumulation of the ngrams if we want to prevent the division by zero, best option is to implement the smoothing, if not do the fix at the aggregation when computing the precision, numerator, denominator
  - If you add this patch on the aggregation, then please add a test in the NIST unittest. 
- Please move out or remove the other unrelated commits in the same PR.

EgodPrime:
I'll revise this fix and start another PR with a more suitable name when I have time.

Issue #3442, opened 2025-10-28
Python 3.14 Missing support


MrMarvel:
<img width="1261" height="112" alt="Image" src="https://github.com/user-attachments/assets/b838bb7a-d105-47be-89e1-37a4e2d7b09e" />

MrMarvel:
#3443 Added pull request

purificant:
The goal should be to have a passing test suite for for py3.13 and py3.14.

Related PR: https://github.com/nltk/nltk/pull/3433
Current list off issues that need resolving: https://github.com/purificant/nltk/actions/runs/19113280626/job/54615833544

cclauss:
After
* #3454 is merged...

Py3.14 seems blocked by:
* % `uvx --with=gensim python3.14 -c 'import gensim' ` # Fails... -->
    * https://github.com/piskvorky/gensim/pull/3633

MrMarvel:
Pull requests merged. Issue seems resolved. Closing issue.

PR #3451, opened 2025-11-07
add tests for _make_bound_method

purificant:
This PR is a stepping stone towards addressing https://github.com/nltk/nltk/issues/3442 and having green CI for py 3.13 / 3.14.

This PR adds explicit test coverage for `_make_bound_method`. This method is also tested indirectly and extensively in CI during corpus unload as a side effect of running other tests.

I suspect the current implementation of `_make_bound_method` is not compatible with py3.13 due to internal python api changes, therefore, the first step is to ensure it works for py3.10-3.12 in CI. The intention is to produce another PR as a followup with py3.13 specific changes.

### DIFF
--- a/nltk/test/unit/test_corpus_util.py
+++ b/nltk/test/unit/test_corpus_util.py
@@ -1,0 +1,65 @@
+"""
+Tests for nltk.corpus.util module
+"""
+
+import pytest
+
+from nltk.corpus.util import _make_bound_method
+
+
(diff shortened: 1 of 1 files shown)


purificant:
cc @ekaf

purificant:
CI is green: https://github.com/nltk/nltk/actions/runs/19166842368
This PR only adds new test coverage. Safe to merge.

PR #3454, opened 2025-11-07
minimal changes to run CI for python 3.13

purificant:
This PR is a followup from https://github.com/nltk/nltk/pull/3451 and an incremental step towards https://github.com/nltk/nltk/issues/3442

We now run tests for py3.13 in CI. The "_unload" function of LazyCorpusLoader is effectively disabled for py3.13 during test execution because it leads to segfaults when the test suite is run in it's entirety as a whole. We will rely on native Python garbage collection instead.

The implementation of "_make_bound_method" is replaced with a more modern alternative as a followup from https://github.com/nltk/nltk/pull/3451, direct and indirect use of this method in tests / CI continues to work.

### DIFF
--- a/nltk/test/conftest.py
+++ b/nltk/test/conftest.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 from nltk.corpus.reader import CorpusReader
@@ -30,4 +32,7 @@ def teardown_loaded_corpora():
     for name in dir(nltk.corpus):
         obj = getattr(nltk.corpus, name, None)
         if isinstance(obj, CorpusReader) and hasattr(obj, "_unload"):
+            # skip unload on Python 3.13+ to avoid segfaults
+            if sys.version_info >= (3, 13):
+                continue
             obj._unload()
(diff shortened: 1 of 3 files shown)


purificant:
Tests are passing: https://github.com/nltk/nltk/actions/runs/19169195760
image

purificant:
cc @ekaf

cclauss (approved):

nltk/test/conftest.py
cclauss, line 28: On Python >= 3.13, we can bail out early...
```suggestion
    yield  # first, wait for the test to end

    if sys.version_info >= (3, 13):
        return
```

purificant: Yep, good call 👍

purificant: Updated in commit https://github.com/nltk/nltk/pull/3454/commits/f592b2da358783719cbed0b67e9a26bfdcc13590

purificant:
Thank you for review @cclauss

PR #3443, opened 2025-10-29
Add support for python 3.14

MrMarvel:
Fix for issue #3442

PR #3457, opened 2025-11-07
cleanup mentions of python versions in documentation and metadata

purificant:
This PR proposes including py3.14 in places where python versions are mentioned.
It also adds a few skips for scenarios where only core nltk tests are run without third party dependencies.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,7 +82,7 @@ Summary of our git branching model:
 - Do many small commits on that branch locally (`git add files-changed`,
   `git commit -m "Add some change"`);
 - Run the tests to make sure nothing breaks
-  (`tox -e py312` if you are on Python 3.12);
+  (`tox -e py313` if you are on Python 3.13);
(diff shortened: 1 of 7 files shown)


purificant:
Adding py3.14 classifier for pypi is included in https://github.com/nltk/nltk/pull/3443

cclauss:
Please resolve git conflicts.

purificant:
Conflicts resolved by keeping the wording from https://github.com/nltk/nltk/pull/3443

purificant:
Interesting case of CI failure: https://github.com/nltk/nltk/actions/runs/19230703841/job/54968369940
I've seen a few of these in the last week and it's always "Python 3.11 on macos-latest" with the root cause of missing NLTK data. I think we always used to install dependencies on macos as a workaround for this kind of intermittent failure.

@ekaf I know you've looked into this before, maybe these new CI failure examples are useful?

cclauss (approved):

purificant:
Thank you for review @cclauss

PR #3461, opened 2025-11-09
Update destructive.py

### LINKED ISSUE #3459: word_tokenize default tokenizer unusual behavior with em dash
I don't know if this is intended behavior or not, so let me explain why this was surprising to me.

[Wikipedia has the following to say](https://en.wikipedia.org/wiki/Dash#Em_dash) about the em dash:

> The em dash is used in several ways. It is primarily used in places where a set of parentheses or a colon might otherwise be used, and it can also show an abrupt change in thought (or an interruption in speech) or be used where a full stop (period) is too strong and a comma is too weak (similar to that of a semicolon). Em dashes are also used to set off summaries or definitions.

My reading of this is that unlike a hyphen (which can join words) an em dash is just punctuation. It can take the place of parentheses, a colon, a period, or a comma.

So let's see what happens if we attempt to tokenize one of the examples provided on Wikipedia.

```
>>> import nltk
>>> nltk.word_tokenize("The food—which was delicious—reminded me of home.")
['The', 'food—which', 'was', 'delicious—reminded', 'me', 'of', 'home', '.']
```

The tokenizer treats `food—which` and `delicious—reminded` as if they were singular tokens. Things get even stranger if we try to apply PoS tagging to this.

```
>>> nltk.pos_tag(_)
[('The', 'DT'), ('food—which', 'NN'), ('was', 'VBD'), ('delicious—reminded', 'VBN'), ('me', 'PRP'), ('of', 'IN'), ('home', 'NN'), ('.', '.')]
```

So it thinks `food—which` is a (singular) noun and that `delicious—reminded` is a (past participle) verb. I suppose reasonable assumptions given the input we provided it, but not the tags I was hoping for.

Shivansh-Game:
Fixes #3459 

I'd like to propose a fix for this issue. As @ryanavella pointed out, the NLTKWordTokenizer currently doesn't treat the em-dash (—) as a separate punctuation mark, which causes it to incorrectly merge adjacent words (e.g., food—which).

This is because the em-dash is missing from the PUNCTUATION regex list in nltk/tokenize/destructive.py.

The Fix
I've added a new rule to the PUNCTUATION list to specifically handle the em-dash by padding it with spaces.

Here is the diff for the change in nltk/tokenize/destructive.py:

```diff
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -87,6 +87,7 @@
             re.compile(r"\.{2,}", re.U),
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
         (re.compile(r"[;@#$%&]"), r" \g<0> "),
+        (re.compile(r"\u2014", re.U), r" \g<0> "), # Handle em-dashes
         (
             re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'),
             r"\1 \2\3 ",
```

Test Results
With this fix, the tokenizer now correctly handles the example provided in the issue:

Python

import nltk
s = "The food—which was delicious—reminded me of home."
Before:

Python

>>> nltk.word_tokenize(s)
['The', 'food—which', 'was', 'delicious—reminded', 'me', 'of', 'home', '.']
After:

Python

>>> nltk.word_tokenize(s)
['The', 'food', '—', 'which', 'was', 'delicious', '—', 'reminded', 'me', 'of', 'home', '.']
Regarding the pos_tag Issue
This tokenization fix also resolves the root cause of the incorrect pos_tag output. The tagger was previously receiving "garbage" tokens (food—which) and making incorrect guesses.

With the corrected tokens, the pos_tag output is now logical:

Python

>>> nltk.pos_tag(nltk.word_tokenize(s))
[('The', 'DT'), ('food', 'NN'), ('—', 'NNP'), ('which', 'WDT'), ('was', 'VBD'), 
 ('delicious', 'JJ'), ('—', 'NNP'), ('reminded', 'VBD'), ('me', 'PRP'), 
 ('of', 'IN'), ('home', 'NN'), ('.', '.')]

The tagger correctly identifies all the words (food, which, delicious, etc.). It appears to default to NNP (Proper Noun) for the lone — token, which is a reasonable guess from the statistical model given it's likely an unknown token in that context. The core tokenization problem is solved.

I have tested this locally and am happy to open a Pull Request.

### DIFF
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -85,7 +85,8 @@ class NLTKWordTokenizer(TokenizerI):
             re.compile(r"\.{2,}", re.U),
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
-        (re.compile(r"[;@#$%&]"), r" \g<0> "),
+        (re.compile(r"[;@#$%&]"), r" \g<0> 
+        (re.compile(r"\u2014", re.U), r" \g<0> "), # \u2014 handles emdashes
         (
             re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'),
             r"\1 \2\3 ",
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -86,7 +86,7 @@ class NLTKWordTokenizer(TokenizerI):
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
         (re.compile(r"[;@#$%&]"), r" \g<0> 
-        (re.compile(r"\u2014", re.U), r" \g<0> "), # \u2014 handles emdashes
+        (re.compile(r"\u2014", re.UNICODE), r" \g<0> "),  # \u2014 handles emdashes
         (
             re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'),
             r"\1 \2\3 ",
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -85,7 +85,7 @@ class NLTKWordTokenizer(TokenizerI):
             re.compile(r"\.{2,}", re.U),
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
-        (re.compile(r"[;@#$%&]"), r" \g<0> 
+        (re.compile(r"[;@#$%&]"), r" \g<0> "),
         (re.compile(r"\u2014", re.UNICODE), r" \g<0> "),  # \u2014 handles emdashes
         (
             re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'),
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -86,7 +86,7 @@ class NLTKWordTokenizer(TokenizerI):
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
         (re.compile(r"[;@#$%&]"), r" \g<0> "),
-        (re.compile(r"\u2014", re.UNICODE), r" \g<0> "),  # \u2014 handles emdashes
+        (re.compile(r"[\u2012-\u2015]", re.UNICODE), r" \g<0> "),  # Handles figure dash, en dashes, em dashes and horizontal bars
         (
             re.compile(r'([^\.])(\.)([\]\)}>"\']*)\s*$'),
             r"\1 \2\3 ",
--- a/nltk/tokenize/destructive.py
+++ b/nltk/tokenize/destructive.py
@@ -86,7 +86,10 @@ class NLTKWordTokenizer(TokenizerI):
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
         (re.compile(r"[;@#$%&]"), r" \g<0> "),
-        (re.compile(r"[\u2012-\u2015]", re.UNICODE), r" \g<0> "),  # Handles figure dash, en dashes, em dashes and horizontal bars
+        (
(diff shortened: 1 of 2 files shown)


nltk/tokenize/destructive.py
```diff
@@ -85,7 +85,8 @@ class NLTKWordTokenizer(TokenizerI):
             re.compile(r"\.{2,}", re.U),
             r" \g<0> ",
         ),  # See https://github.com/nltk/nltk/pull/2322
-        (re.compile(r"[;@#$%&]"), r" \g<0> "),
+        (re.compile(r"[;@#$%&]"), r" \g<0> 
+        (re.compile(r"\u2014", re.U), r" \g<0> "), # \u2014 handles emdashes
```
cclauss: Readability: [`ruff rule FURB167`](https://docs.astral.sh/ruff/rules/regex-flag-alias)
```suggestion
        (re.compile(r"\u2014", re.UNICODE), r" \g<0> "),  # \u2014 handles emdashes
```

Shivansh-Game:
Hello maintainers, the fix is causing a test to fail in lm.test_models.py, only on the macos-3.11/3.10 build. This is my first contribution, so I'm not confident enough in my ability to fix an issue like this. Could a maintainer please take a look?

purificant:
Triggered test re-run

ekaf (changes requested):
Goal: Pad em dashes so tokens like “food—which” are split correctly.

Problem: The pattern proposed here uses raw string r"\u2014"; this matches the literal backslash-u sequence, not the em dash. re.UNICODE doesn’t fix that.
- Fix: Use "\u2014" (non-raw), the literal "—", or a class "[\u2014]".
- Optional enhancement: Broaden to "[\u2012-\u2015]" to cover ‒ – — ―.

Placement in PUNCTUATION is appropriate; existing whitespace normalization will handle added spaces.

Please add unit tests for em dash (and optionally other dash characters) to prevent regressions.

Shivansh-Game:
Thanks for pointing that out @ekaf, I've fixed it and broadened the scope to cover the other dash characters you suggested. I've also added the unit tests for all of the new dash characters.

ekaf:
@Shivansh-Game, please note that CI fails already on pre-commit. This is because you need to run both _git add_ and _git commit_ again whenever black or isort modify any of your files.

Shivansh-Game:
Hi @ekaf, I've just pushed the new commit with the pre-commit formatting applied.
This was my first time doing the full git workflow, thank you walking me through that. o7

ekaf:
@Shivansh-Game, to debug the latest CI error, you could try to run "pytest  -v nltk/test/unit/test_tokenize.py" on your local machine before your next commit.

Shivansh-Game:
Hey @ekaf, thanks for the advice and patience, I ran all the tests locally after fixing the errors in the previous commit. It should work now.

ekaf (changes requested):
Thanks @Shivansh-Game, this looks good., but: 
 - Horizontal bar (U+2015) is not actually tested. The test labeled “Horizontal Bar (\u2015)” uses the em dash glyph (—, U+2014) again in both the input string and expected tokens. The horizontal bar character is ― (slightly longer visually). 
- (Optionally) Add one regression guard for hyphen-minus: e.g. assert word_tokenize("state-of-the-art") == ["state-of-the-art"] (or the current canonical expected output) to be sure behavior is unchanged.

Shivansh-Game:
Hey @ekaf, Thanks for pointing that out. I've fixed that unit test and added the regression guard. I've tested it locally and it should test for horizontal bars properly now and pass the CI.

ekaf (approved):
LGTM, thanks @Shivansh-Game!

Shivansh-Game:
Hey, I just wanted to ask if there's anything I should do on my end that I might have overlooked since this pull request has been approved for a while now.

PR #3462, opened 2025-11-10
pass input to boxer binary via stdin instead of shell pipe

purificant:
This PR changes how input is passed to the candc/boxer binaries in `_call()`. Instead of using a shell pipe with `echo`, we now pass input directly to the subprocess stdin. This also prevents user input from being interpreted as shell commands.


purificant:
Thank you https://huntr.com/users/ehtec for reporting the vulnerability

PR #3464, opened 2025-11-11
Enable pre-commit to fix pull requests

cclauss:
Add https://github.com/pre-commit/action and https://pre-commit.ci/lite.html to our GitHub Actions to avoid manual pre-commit fixes to pull requests.  This is not as good as using the (free to open source) https://pre-commit.ci

Fixes the issue discussed at:
* https://github.com/nltk/nltk/pull/3461#issuecomment-3518251102

@purificant

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,10 @@ jobs:
         with:
           python-version: "3.14"
 
-      - name: Install pre-commit
-        run: |
-          pip install pre-commit
-          pre-commit run --all-files
+      - uses: pre-commit/action@v3.0.1
+
+      - uses: pre-commit-ci/lite-action@v1.1.0
+        if: always()
 
   minimal_download_test:
     name: Minimal NLTK Download Test
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,12 +8,12 @@ repos:
       - id: requirements-txt-fixer
       - id: check-yaml
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.21.0
+    rev: v3.21.1
     hooks:
     - id: pyupgrade
       args: ["--py38-plus"]
   - repo: https://github.com/ambv/black
-    rev: 25.9.0
+    rev: 25.11.0
     hooks:
     - id: black
   - repo: local


ekaf:
Thanks for this @cclauss .
From a security standpoint, the status quo (check-only pre-commit that fails CI) may be safer than introducing auto-fix pushes or a third-party app with write access. 
So I lean towards keeping the current  checks-only configuration, and ask contributors to run pre-commit locally (documented in CONTRIBUTING), unless we can guarantee least-privilege with no secret exposure or write permissions.

cclauss:
As counter argument… We pick our pre-commit tasks and the standard hooks plus black, isort, and pyupgrade do not generate dangerous modifications.  Even if they did, those modifications are not committed to the default branch.  Instead they are placed in separate commits which will be tested and code reviewed along with the contributor’s modifications.

ekaf:
Yes @cclauss, the security concern is only a vague hypothesis, so the convenience of having CI fix the reformatting seems greater.

purificant:
Different teams tends to have different preferences when it comes to how dev tooling is set up and at what point in the workflow the tools run.

Here we have [pre-commit](https://github.com/nltk/nltk/blob/develop/.pre-commit-config.yaml) in the repo, and a [step](https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md#setting-up-a-development-environment) in the dev setup to install the hooks in order to automatically run the pre commit checks. So in theory, if these instructions are followed, then all commits should be compatible with pre-commit checks in CI.

purificant:
I've got no strong opinion one way or the other. Personally I prefer to control all my code changes at the commit level and avoid any post commit automated linting via CI with auto changes, but this workflow could be beneficial for some contributors so happy to try it.

purificant:
https://pre-commit.ci/ has an interesting feature with "automatic updates" where they will periodically run "[autoupdate](https://pre-commit.com/#pre-commit-autoupdate)" on the pre commit config file, this is nice to keep up with the times.

purificant (approved):

purificant:
Happy to try this. But personally won't rely on it. Up to @ekaf, we can always go back and revert if there are issues.

Maybe we should consider integrating with https://pre-commit.ci/ if feasible? Then it would defacto replace this approach.

ekaf:
I would feel better about this if we had a proof that it is harmless. Isort should be, but it seems difficult to rule out the possibility that a well-crafted expression will be able to trick some future version of _black_ into doing a harmful replacement.
It is acceptable that [CI occasionally fails](https://github.com/nltk/nltk/pull/3461#issuecomment-3518251102) for first-time committers , since it is easily solved by just following the documentation.

purificant:
@cclauss Can we keep the `pre-commit autoupdate` changes?

purificant:
There's probably a way to run pre-commit autofix in CI and post the changeset as a comment on the pull request to make any issues even more obvious without looking at CI logs. If anyone wants to pursue that. :thinking:

PR #3465, opened 2025-11-17
replace eval() with getattr() in collocations CLI

purificant:
This PR replaces [eval()](https://docs.python.org/3/library/functions.html#eval) with [getattr()](https://docs.python.org/3/library/functions.html#getattr) when parsing command-line arguments.

The `__main__` block previously used `eval()` to dynamically access `BigramAssocMeasures` methods based on command-line arguments. Since we only need attribute access (not expression evaluation), `getattr()` is more appropriate. Using `eval()` on user provided input also unnecessarily allows arbitrary code execution, which `getattr()` prevents.

### DIFF
--- a/nltk/collocations.py
+++ b/nltk/collocations.py
@@ -389,17 +389,27 @@ def demo(scorer=None, compare_scorer=None):
 # bigram_measures = BigramAssocMeasures()
 # trigram_measures = TrigramAssocMeasures()
 
+# Command-line interface for demonstrating bigram collocations.
+#
+# Usage: python -m nltk.collocations [scorer] [compare_scorer]
+#
+# Demonstrates bigram collocations on the WebText corpus.
+# Defaults to likelihood_ratio and raw_freq if not specified.
+#
+# Available scorers:
+#   chi_sq, dice, fisher, jaccard, likelihood_ratio, mi_like,
+#   phi_sq, pmi, poisson_stirling, raw_freq, student_t
 if __name__ == "__main__":
     import sys
 
     from nltk.metrics import BigramAssocMeasures
 
     try:
-        scorer = eval("BigramAssocMeasures." + sys.argv[1])
+        scorer = getattr(BigramAssocMeasures, sys.argv[1], None)
     except IndexError:
         scorer = None
     try:
-        compare_scorer = eval("BigramAssocMeasures." + sys.argv[2])
+        compare_scorer = getattr(BigramAssocMeasures, sys.argv[2], None)
     except IndexError:
         compare_scorer = None
 


cclauss (approved):
% `ruff rule S307`
* https://docs.astral.sh/ruff/rules/suspicious-eval-usage
* https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html

nltk/collocations.py
```diff
@@ -395,12 +395,12 @@ def demo(scorer=None, compare_scorer=None):
     from nltk.metrics import BigramAssocMeasures
 
     try:
-        scorer = eval("BigramAssocMeasures." + sys.argv[1])
-    except IndexError:
+        scorer = getattr(BigramAssocMeasures, sys.argv[1])
+    except (IndexError, AttributeError):
```
cclauss: Also possible:
```suggestion
        scorer = getattr(BigramAssocMeasures, sys.argv[1], None)
    except IndexError:
```

ekaf (changes requested):
Great improvement replacing eval() with getattr() , @purificant. 

It is not immediately obvious what the required arguments are expected to be, so adding a documentation string could also be helpful. Alternatively, you could consider adding an informative  safeguard just before the _demo_ call, like:

```
    if scorer is None and compare_scorer is None:
        print("Usage: python -m nltk.collocations <scorer_name> [<compare_scorer_name>]")
        print("Available scorer names:", ", ".join(
            n for n in dir(BigramAssocMeasures) if not n.startswith('_')))
        sys.exit(2)
```

nltk/collocations.py
ekaf, line 402: It might be helpful to document the expected CLI arguments and show an example, so users know what `sys.argv[1]` and `sys.argv[2]` should be. For example:

```suggestion
if __name__ == "__main__":
    import sys

    """
    Command-line usage for the collocations demo.

    Usage:
        python -m nltk.collocations <scorer_name> [<compare_scorer_name>]

    Both arguments should be names of association measures defined on
    nltk.metrics.association.BigramAssocMeasures, such as "pmi", "phi_sq",
    "chi_sq", "likelihood_ratio", or "raw_freq".

    Example:
        python -m nltk.collocations pmi phi_sq
    """

File: nltk/collocations.py [hunk 2]
```diff
@@ -395,12 +395,12 @@ def demo(scorer=None, compare_scorer=None):
...
         scorer = None
     try:
-        compare_scorer = eval("BigramAssocMeasures." + sys.argv[2])
-    except IndexError:
+        compare_scorer = getattr(BigramAssocMeasures, sys.argv[2])
+    except (IndexError, AttributeError):
         compare_scorer = None
 
     demo(scorer, compare_scorer)
```
ekaf, line 416: To make the CLI behavior clearer when both scorer names are missing/invalid, you could handle that case explicitly and show a usage message:

```suggestion

    if scorer is None and compare_scorer is None:
        print("Usage: python -m nltk.collocations <scorer_name> [<compare_scorer_name>]")
        available = [n for n in dir(BigramAssocMeasures) if not n.startswith("_")]
        print("Available scorer names:", ", ".join(available))
        sys.exit(2)

    demo(scorer, compare_scorer)

File: nltk/collocations.py (hunk 1 above)
Comment (purificant) [reply nltk/collocations.py]: Yep, we can swap catching AttributeError for using default value of None with getattr(). In this case these are equivalent. :+1:

File: nltk/collocations.py (hunk 2 above)
Comment (purificant) [reply nltk/collocations.py:416]: This would change the default behaviour. Currently when we run
```bash
python -m nltk.collocations
```
it runs the demo with with defaults: scorer=likelihood_ratio, compare_scorer=raw_freq

Prefer to avoid this change within the scope of this PR.

nltk/collocations.py
purificant: Nice, added a comment to clarify along these lines :+1:

ekaf (approved):
This looks good to me.

- Replacing `eval()` with `getattr()` in the collocations CLI removes the arbitrary code execution risk while preserving the intended dynamic selection of `BigramAssocMeasures` methods.
- The behavior for valid measure names is preserved, and tests are passing.
- All review suggestions and comments have been adequately addressed.

Thanks for the improvement!

purificant:
Thank you @ekaf @cclauss

aydinnyunus:
Hi Team,

Is it eligible for bounty in huntr ?

- https://huntr.com/bounties/2585a640-8942-446f-9ab5-19d67c6fbf1b

Issue #3413, opened 2025-08-14
Let tokenizers be installed via Pip


bittner:
Call me stupid, but why do NLTK users [have to download NLTK corpora](https://www.nltk.org/data.html) using Python code?

This is not only non-standard, but causes a lot of technical and usability problems along the road (see [related issues](https://github.com/nltk/nltk/issues?q=is%3Aissue%20download) and [StackOverflow](https://stackoverflow.com/questions/77131746/how-to-download-punkt-tokenizer-in-nltk)). Software shouldn't download package data in a custom, self-crafted way when there is PyPI that offers a (more) secure and battle-tested infrastructure.

Isn't it possible to include [the tokenizers](https://www.nltk.org/nltk_data/) via [more `extras_require` targets](https://github.com/nltk/nltk/blob/develop/setup.py#L32C1-L32C15) or separate Python packages instead?

ryanamannion:
I agree it would be helpful. Looking at discussions in the past for this, it looks like the cited reason has been for 
  
  1. lack of resources [[link to comment](https://github.com/nltk/nltk/issues/55#issuecomment-236334927)]

  2. Maximum file size on pypi 
     * The max is [currently 100 MiB](https://github.com/pypi/warehouse/blob/58dc95e9f38e895e5c7cb203a99076bf98a64b65/warehouse/constants.py#L5). As far as I can tell, the largest nltk_data package is Framenet v1.7 at 94.6 MiB. Dataset sizes here: https://www.nltk.org/nltk_data/

The most relevant issues can be found here:
* https://github.com/nltk/nltk/issues/55
* https://github.com/nltk/nltk_data/issues/12

I think spaCy gets around the pypi file size limit by having their models installable via their downloader `$ python -m spacy download en_core_web_sm`, which installs the wheel that they host on github e.g.

```
$ python -m spacy download en_core_web_sm
Looking in indexes: https://pypi.org/simple
Collecting en-core-web-sm==3.7.1
  Downloading https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.7.1/en_core_web_sm-3.7.1-py3-none-any.whl (12.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 12.8/12.8 MB 18.3 MB/s eta 0:00:00
...
Installing collected packages: en-core-web-sm
Successfully installed en-core-web-sm-3.7.1
✔ Download and installation successful
[...]

PR #3470, opened 2025-11-25
Pull request for issue #3413 -> Python package for extra tokenizers

Julien-ser:
Created a new python package to install additional tokenizers instead of the nltk.downloader() class.
More information and documentation can be seen in the pypi page for this library https://pypi.org/project/nltk-extratokenizers/0.1.2/


ekaf (commented):
Thanks for the contribution, but this PR is not acceptable for the following reasons:

- It contains many references to `english.pickle`, which exposes users to the known “sleepy pickle attack” vulnerability and was explicitly abandoned following issue #3266.
- It introduces a large, monolithic feature without clear alignment with NLTK’s existing design, scope, or roadmap.
- The proposed pip package name `extra-tokenizers` is misleading: it suggests adding multiple tokenizers, but in practice just points to the forbidden `english.pickle`.
- The pip package includes neither data nor Python code, so it does not actually address issue #3413 as claimed.
- The test coverage is incomplete and does not convincingly demonstrate correctness, robustness, or backward compatibility.
- The code quality (structure, style, and naming) does not meet our standards and would require extensive refactoring.
- There is no clear migration or maintenance strategy, and the long‑term support burden for the project is not addressed.

For these reasons, we will not be moving forward with this PR. I’d recommend first getting acquainted with NLTK’s existing issues and PRs, and then focusing on addressing smaller, well-scoped problems in individual pull requests.

PR #3477, opened 2025-12-08
Security patch: Add SHA256 validation to StanfordSegmenter to prevent untrusted JAR RCE

HyperPS:
This patch introduces SHA256 integrity verification and trusted-path enforcement
when loading Stanford segmenter JARs. It prevents arbitrary code execution through
malicious/modified JARs, while remaining backwards compatible with official models.
Blocks supply-chain JAR injection attacks
Allows verified usage via NLTK_SEGMENTER_ALLOW_SHA256=<hash>
No workflow/output changes for legitimate users

### DIFF
--- a/nltk/tokenize/stanford_segmenter.py
+++ b/nltk/tokenize/stanford_segmenter.py
@@ -278,6 +278,34 @@ class StanfordSegmenter(TokenizerI):
 
         default_options = " ".join(_java_options)
 
+        # ------------------- Security Validation ------------------- #
+        import hashlib
+        import os
+
+        jar_path = self._stanford_jar
+        trusted_paths = ["stanford-segmenter", "stanford-corenlp", "nltk"]
+
+        def sha256sum(file):
+            h = hashlib.sha256()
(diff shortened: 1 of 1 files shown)


stevenbird:
Thank you @HyperPS

darakian:
Hey there 👋 
I was reading over https://www.cve.org/CVERecord?id=CVE-2026-0848 and I'm curious if I could get an explanation on the attack vector here. It seems like users are expected to download the standford jar out of band of any code execution
https://github.com/nltk/nltk/wiki/Installing-Third-Party-Software

The cve and huntr report both mention a remote attack vector which is not clear to me. It seems to me that either the attacker needs to be on the system in question with permissions to provide the crafted jar or that they would need to compromise stanford's nlp file hosting.

The check against `NLTK_SEGMENTER_ALLOW_SHA256` is also curious as there don't seem to be instructions for the end user aside from the print out included in this patch. Any existing users will likely hit this code path and have an unexpected crash in their system.

Maybe I'm missing something, but could you provide some clarity on this @stevenbird ?

PR #3479, opened 2025-12-08
Prevent path traversal & absolute-path access in CorpusReader.open() (CWE-22)

HyperPS:
This patch introduces path sanitization inside `CorpusReader.open()` to prevent
directory traversal and direct absolute path access.

### Key Security Improvements
- Blocks absolute paths (`/etc/passwd`, `C:\...`, `\\server\share`)
- Rejects `../` traversal attempts
- Ensures resolved paths remain inside corpus root
- Mitigates CWE-22 Path Traversal vulnerability

### Why this matters
If user-controlled filenames are passed into corpus readers (common in NLP APIs,
file upload endpoints, or ML pipelines), an attacker could previously open
arbitrary system files. This patch prevents unauthorized file reads without
breaking normal corpus usage.

Minimal patch, no API changes, safe for backward compatibility.

### DIFF
--- a/nltk/corpus/reader/api.py
+++ b/nltk/corpus/reader/api.py
@@ -237,17 +237,13 @@ class CorpusReader:
         root_path = os.path.normpath(self._root._path)
         target_path = os.path.normpath(joined._path)
 
-        if not (
-            target_path == root_path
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @HyperPS

PR #3481, opened 2025-12-09
Add sandbox enforcement to filestring() to prevent arbitrary file reads

HyperPS:
This PR introduces a security hardening patch for filestring() by adding sandbox path validation support through an optional allowed_dir parameter.
Previously, filestring() allowed reading any filesystem path if user input was passed, which could lead to arbitrary file read / directory traversal risks when invoked by external-facing applications.

The updated implementation now:

Requires allowed_dir when reading file paths

Blocks reading outside the specified directory (prevents ../, symlink escape)

Maintains compatibility for file-like objects (.read() streams)

Uses realpath() to avoid bypass via symlinks or relative paths

Adds UTF-8 safe open with fallback error handling

This improves security for downstream frameworks, web services, and notebook environments using NLTK utilities.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,13 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.14"
+          python-version: "3.13"  # or your chosen version
 
       - name: Install pre-commit
-        run: |
-          pip install pre-commit
-          pre-commit run --all-files
+        run: pip install pre-commit
+
+      - name: Run pre-commit hooks
+        run: pre-commit run --all-files
 
   minimal_download_test:
     name: Minimal NLTK Download Test


ekaf (changes requested):
Thanks @HyperPS! You are proposing a breaking change, and I am open to that. Otherwise, for backwards compatibility, you could eventually consider not making allowed_dir strictly required for all file-path reads, keep allowed_dir default None and only enforce the sandbox when a non-None allowed_dir is provided (opt-in). Then, if you want to require it later, do a deprecation cycle. 

Add some unit tests, for ex.:
    - reads allowed file inside allowed_dir
    - rejects ../ traversal
    - rejects symlink escape attempts
    - preserves behavior for file-like objects
    - tests encoding fallback behavior

Also, some documentation would be helpful, for ex. mentioning that this approach can still be vulnerable to "Time Of Check To Time Of Use (TOCTOU) / symlink race attacks if an adversary controls the filesystem concurrently. Realpath checks reduce some bypasses but don’t eliminate race windows.

ekaf:
The pre-commit failure is due to a pyupgrade incompatibility with Python 3.14, and unrelated to this PR. For the purpose of testing this PR, it is practical to use Python 3.13, which does not have this incompatibility.

PR #3484, opened 2025-12-12
Use Python 3.13 for pre-commit

### LINKED ISSUE #3483: `TypeError: cannot use a bytes pattern on a string-like object` when running pre-commit with Python 3.14
CI for all PRs is currently encountering an issue when running `pre-commit` with Python 3.14. The `pyupgrade` hook crashes with the following error:

```plaintext
TypeError: cannot use a bytes pattern on a string-like object
```
See in particular the CI logs for commit [edb12af](https://github.com/nltk/nltk/pull/3481/commits/edb12afb639299843bae0830d2b3d920560166d3), where pre-commit uses Python 3.14 and fails, vs. [c05f626](https://github.com/nltk/nltk/pull/3481/commits/c05f626d1bcde01e62956f29b3ae3f772ba3cc69), where it succeeds by using Python 3.13 instead.

Re-running [CI for a previous commit](https://github.com/nltk/nltk/actions/runs/19878884315/job/57873155764) on the develop branch indicates that pre-commit now fails everywhere it previously succeeded.

**Workaround**:
Running `pre-commit` with Python 3.13 (or earlier) avoids this issue and allows the hooks to run successfully.

ekaf:
This fixes #3483, by temporarily using Python 3.13 for pre-commit, until more knowledge is available about the incompatibility with Python 3.14. 

This issue occurs only with Python 3.14. It does not occur with Python 3.13 or earlier. Python 3.14 introduces stricter handling of `str` and `bytes`, which might be the root cause. The failing hook is `pyupgrade`, but it is unclear whether this issue stems from `pyupgrade` or `pre-commit` itself.

Later, we will need to determine whether the problem originates in `pre-commit` or `pyupgrade`, and wait for updates to these tool(s), that ensure compatibility with Python 3.14.

Meanwhile, let's use the latest Python 3.13 for pre-commit, since it is the latest that works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,9 +29,6 @@ jobs:
       - name: Install pre-commit
         run: pip install pre-commit
 
-      - name: Clear pre-commit cache
-        run: pre-commit clean
-
       - name: Run pre-commit hooks
         run: pre-commit run --all-files
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,16 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.14"
+          python-version: "3.13"  # or your chosen version
 
       - name: Install pre-commit
-        run: |
-          pip install pre-commit
-          pre-commit run --all-files
+        run: pip install pre-commit
+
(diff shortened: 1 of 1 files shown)


ekaf:
This PR restores CI stability and is easily reversible once `pre-commit` and `pyupgrade` are updated to support Python 3.14.

PR #3495, opened 2026-01-07
fix the Jaro similarity score for single-character strings and empty strings

jancallewaert:
The similarity calculated for single character strings and empty strings was wrong. The similarity between z and z was returned as 0. The similarity between two empty strings was returned as 0 as well. This brings the implementation in line with the implementation of Apache Commons Text. See https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/similarity/JaroWinklerSimilarity.java#L76 and https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/similarity/JaroWinklerSimilarity.html#apply(java.lang.CharSequence,java.lang.CharSequence)

### DIFF
--- a/nltk/metrics/distance.py
+++ b/nltk/metrics/distance.py
@@ -311,6 +311,13 @@ def jaro_similarity(s1, s2):
         - `m` is the no. of matching characters
         - `t` is the half no. of possible transpositions.
     """
+    # By definition, the similarity of a string with itself is 1.0.
+    # This also handles edge cases where both strings are empty or
+    # single-character identical strings, where the matching window
+    # formula (floor(max(|s1|,|s2|) / 2) - 1) yields -1.
+    if s1 == s2:
+        return 1.0
+
     # First, store the length of the strings
     # because they will be re-used several times.
     len_s1, len_s2 = len(s1), len(s2)
@@ -340,7 +347,7 @@ def jaro_similarity(s1, s2):
             transpositions += 1
 
     if matches == 0:
-        return 0
+        return 0.0
     else:
         return (
             1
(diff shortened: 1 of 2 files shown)


alvations:
Good catch!! Thanks @jancallewaert for the fix, added some comments and explanation and the unittest. 

----

## Summary

Fix `jaro_similarity` for empty and single-character strings, and add comprehensive unit tests.

### Bugs fixed

| Input | Before | After | Expected |
|-------|--------|-------|----------|
| `jaro_similarity("", "")` | `0.0` | `1.0` | `1.0` |
| `jaro_similarity("a", "a")` | `0.0` | `1.0` | `1.0` |
| `jaro_similarity("a", "b")` | `0` (int) | `0.0` (float) | `0.0` |
| `jaro_winkler_similarity("", "")` | `0.0` | `1.0` | `1.0` |
| `jaro_winkler_similarity("a", "a")` | `0.1` | `1.0` | `1.0` |

**Root cause:** The matching window formula `match_bound = max(|s1|, |s2|) // 2 - 1` evaluates to `-1` for single-character strings, making the matching window empty and yielding 0 matches. For empty
strings, the loop body never executes, also yielding 0 matches.

### Changes

**`nltk/metrics/distance.py`:**
- Add `if s1 == s2: return 1.0` early return at the top of `jaro_similarity()` — handles identical strings including empty and single-character edge cases
- Return `0.0` (float) instead of `0` (int) when `matches == 0` for consistent return type
- Add edge case doctests in `jaro_winkler_similarity` for empty/single-char strings

**`nltk/test/unit/test_distance.py`:**
- Add `TestJaroSimilarity` class (16 tests):
  - Edge cases: both empty, empty vs non-empty, single-char identical, single-char different
  - Known reference values from Wikipedia and census papers (MARTHA/MARHTA, DWAYNE/DUANE, DIXON/DICKSON, CRATE/TRACE, billy/bill, billy/susan)
  - Symmetry: `jaro(s1, s2) == jaro(s2, s1)`
  - Return type consistency: all code paths return `float`
- Add `TestJaroWinklerSimilarity` class (5 tests):
  - Edge cases: both empty, single-char identical, single-char different
  - Known reference value (MARTHA/MARHTA = 0.961)
  - Invariant: `jaro_winkler(s1, s2) >= jaro(s1, s2)` for common prefixes

alvations (approved):
LGTM! Again, good catch!

alvations:
[CI: retest]

stevenbird:
Thanks @jancallewaert for the contribution and @alvations for reviewing 👌

PR #3497, opened 2026-01-12
Add support for Python 3.13 3.14 3.14t

bowiechen:
Add support for Python 3.13 3.14 3.14t so we can use this library with newer Python runtimes.


stevenbird:
Thanks @bowiechen

PR #3498, opened 2026-01-14
lm: bound recursion depth to prevent DoS

HyperPS:
**This patch mitigates a denial-of-service issue caused by unbounded recursion in NLTK language models.**

_A defensive limit is applied to the context length in recursive scoring paths (StupidBackoff and interpolated models), preventing stack exhaustion while preserving existing behavior for normal usage._

The fix does not change public APIs or model semantics and only affects pathological inputs with excessively long contexts.

### DIFF
--- a/nltk/lm/models.py
+++ b/nltk/lm/models.py
@@ -74,9 +74,10 @@ class StupidBackoff(LanguageModel):
         self.alpha = alpha
 
     def unmasked_score(self, word, context=None):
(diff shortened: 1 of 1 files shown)


HyperPS:
_Context length is now bounded to order - 1, which matches n-gram semantics and prevents attacker-controlled recursion depth while preserving model behavior._

stevenbird:
Thanks @HyperPS!

PR #3500, opened 2026-02-13
fix floating point precision in language model probability sums

### LINKED ISSUE #3275: Summing Ngram LM probabilities requires math.fsum
Without `math.fsum` Python versions >3.8 won't sum up probabilities from `nltk/lm` properly. 

Details see failing tests on https://github.com/nltk/nltk/actions/runs/9812464268/job/27096709686?pr=3270 

To sum up the probabilities right, users would have to use https://github.com/nltk/nltk/pull/3274/commits/082846b2d7f7d54d553eead04f6b361bfe466c5a before the https://github.com/python/cpython/issues/100425 after Python 3.8

Related Issues: https://github.com/nltk/nltk/pull/3274

Mr-Neutr0n:
Fixes #3275

The language model code uses `sum()` to add up probabilities and log-scores, but `sum()` accumulates rounding errors when you're adding many small floats together. On Python 3.9+ this can cause probability sums that should equal 1.0 to drift enough to break assertions (related to cpython#100425).

Switched to `math.fsum()` which uses the Shewchuk algorithm to track partial sums accurately. Changed it in `_mean()` (used by `entropy()`), in `_weighted_choice()` (used by `generate()`), and in the doctest that sums scores over the vocabulary.

### DIFF
--- a/nltk/lm/api.py
+++ b/nltk/lm/api.py
@@ -6,6 +6,7 @@
 # For license information, see LICENSE.TXT
 """Language Model Interface."""
 
+import math
 import random
 import warnings
 from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @Mr-Neutr0n!

Mr-Neutr0n:
happy to help @stevenbird

PR #3505, opened 2026-02-23
Fix #3394: Ensure punkt_tab is downloaded when punkt is requested

### LINKED ISSUE #3394: word_tokenize fails due to punkt_tab reference in multiple environments
I'm getting a `LookupError` from `word_tokenize` that tries to load a non-existent `punkt_tab` resource. The line causing the issue is:

```bash
lang_dir = find(f"tokenizers/punkt_tab/{lang}/")
```

This occurs in multiple environments:

- Python 3.8.5 (global, Mac)
- Python 3.12.7 (via Anaconda)
- Google Colab (Python 3.11, preinstalled NLTK)

Expected behaviour:
- It should be loading `punkt/{lang}.pickle` as in previous versions.

Update:
- Downgrading to `nltk==3.8.1` resolves the issue. It appears this was introduced in `nltk==3.9.1`.

Thanks!

askubaid:
### Description
This PR addresses the `LookupError` reported in #3394. Since the transition from the pickle-based `punkt` to the tabular `punkt_tab` in NLTK 3.9.1, users following tutorials that call `nltk.download('punkt')` are left with an environment that fails when calling `word_tokenize()`.

### Changes
Modified `nltk/downloader.py` to:
1. Intercept string requests for `'punkt'` and expand them to `['punkt', 'punkt_tab']`.
2. Check list-based requests and append `'punkt_tab'` if `'punkt'` is present but `'punkt_tab'` is missing.

This ensures a seamless experience for users while maintaining backward compatibility.

### Verification
- Reproduced the `LookupError` in a clean environment.
- Verified that `nltk.download('punkt')` now correctly installs the data required for `word_tokenize()`.
- Ran existing unit tests: `pytest nltk/test/unit/test_downloader.py` passed with 4/4 successes.

Closes #3394

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -749,13 +749,16 @@ class Downloader:
             return True
 
         else:
-            # Define a helper function for displaying output:
-            # Handle both string and list inputs to ensure punkt_tab is always included
+            # Fix for Issue #3394: Redirect or append 'punkt_tab' when 'punkt' is requested.
             if isinstance(info_or_id, str) and info_or_id == "punkt":
                 info_or_id = ["punkt", "punkt_tab"]
-            elif isinstance(info_or_id, list) and "punkt" in info_or_id and "punkt_tab" not in info_or_id:
+            elif (
+                isinstance(info_or_id, list)
+                and "punkt" in info_or_id
+                and "punkt_tab" not in info_or_id
+            ):
                 info_or_id.append("punkt_tab")
-
+            # Define a helper function for displaying output:
             def show(s, prefix2=""):
                 print_to(
                     textwrap.fill(
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -750,6 +750,12 @@ class Downloader:
 
         else:
             # Define a helper function for displaying output:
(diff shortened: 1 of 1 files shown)


ekaf:
Black (and/or isort) reformatting leads to so frequent CI errors, that CONTRIBUTING.md should warn against it.

askubaid:
Apologies for the noise! My editor's auto-formatter (Black) applied project-wide style changes in the previous push. I have now reverted those changes and forced-pushed a clean version that contains only the logic fix for #3394. I've also disabled auto-formatting for this workspace to prevent this from happening again.





> Black (and/or isort) reformatting leads to so frequent CI errors, that CONTRIBUTING.md should warn against it.

ekaf:
The expected process is to apply black formatting, and then commit the files again _after_ black has reformatted them.

ekaf:
So this  file is still blocking CI:

> black....................................................................Failed
> - hook id: black
> - files were modified by this hook
> 
> reformatted nltk/downloader.py
> 
> All done! ✨ 🍰 ✨
> 1 file reformatted, 366 files left unchanged.

PR #3508, opened 2026-02-26
fix: replace 51 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


stevenbird:
@haosenwang1018 would you like to respond to any of these suggestions?

stevenbird:
Thanks @haosenwang1018

PR #3511, opened 2026-02-28
Let downloader fallback to md5 when sha256 is unavailable

### LINKED ISSUE #3510: v3.9.3 breaks use of cached data
When using v3.9.3, the downloader breaks as it fails to verify download data that was cached by pervious versions.

If you clear cache and run it will download and work. However if you then run again the download will fail with same error

Steps to reproduce:
1. run downloader to download any file ex: punkt_tab
2. run downloader again to download same file and it will fail with below stacktrace

```
yield from self. download_package(info, download dir, force)
• venv\Lib\site-packages\n1tk\dounloader.py:668: in_download_package
status - self. status (info, download_dir)
.venv\Lib\site-packages\nltk\downloader.py:868: in status
self._status_cache[[info.id](http://info.id/)] - self._pkg_status(info, filepath)
.venv\Lib\site-packages\nItk\downloader-py:834: in _pkg_status if sha256_hexdigest(filepath) !- info.sha256_checksun:
AttributeError: "Package' object has no attribute "sha256 checksum'
```

ekaf:
Fix #3510. NLTK 3.9.3’s downloader can crash when verifying already-downloaded packages whose metadata was cached by earlier NLTK versions, because _pkg_status() assumes Package.sha256_checksum exists and raises AttributeError if it doesn’t.

This PR makes checksum verification backward-compatible by preferring SHA256 when available and falling back to the legacy MD5 checksum when SHA256 metadata is missing, emitting a one-time warning to encourage refreshing the data index.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -880,9 +880,27 @@ class Downloader:
         if filestat.st_size != int(info.size):
             return self.STALE
 
-        # Check if the file's checksum matches
-        if sha256_hexdigest(filepath) != info.sha256_checksum:
-            return self.STALE
+        # Check if the file's checksum matches.
+        # Prefer sha256, but fall back to legacy md5 if sha256 is unavailable.
+        sha256_checksum = getattr(info, "sha256_checksum", None)
+        if sha256_checksum:
+            if sha256_hexdigest(filepath) != sha256_checksum:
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @ekaf

Issue #3255, opened 2024-05-21
Not able to install nltk on termux, getting the following error:


divyanshluthra:
~/.../python/build $ pip install nltk
Collecting nltk                                           Using cached nltk-3.8.1-py3-none-any.whl.metadata (2.8 kB)
Collecting click (from nltk)                              Using cached click-8.1.7-py3-none-any.whl.metadata (3.0 kB)                                                   Collecting joblib (from nltk)                             Using cached joblib-1.4.2-py3-none-any.whl.metadata (5.4 kB)                                                  Collecting regex>=2021.8.3 (from nltk)                    Using cached regex-2024.5.15.tar.gz (394 kB)            Installing build dependencies ... done                  Getting requirements to build wheel ... done
  Installing backend dependencies ... done                Preparing metadata (pyproject.toml) ... done          Collecting tqdm (from nltk)                               Using cached tqdm-4.66.4-py3-none-any.whl.metadata (57 kB)                                                    Using cached nltk-3.8.1-py3-none-any.whl (1.5 MB)       Using cached click-8.1.7-py3-none-any.whl (97 kB)       Using cached joblib-1.4.2-py3-none-any.whl (301 kB)
Using cached tqdm-4.66.4-py3-none-any.whl (78 kB)
Building wheels for collected packages: regex
  Building wheel for regex (pyproject.toml) ... error     error: subprocess-exited-with-error
                                                          × Building wheel for regex (pyproject.toml) did not run successfully.
  │ exit code: 1                                          ╰─> [20 lines of output]
      running bdist_wheel                                     running build
      running build_py
      creating build                                          creating build/lib.linux-aarch64-cpython-311
      creating build/lib.linux-aarch64-cpython-311/regex
      copying regex_3/__init__.py -> build/lib.linux-aarch64-cpython-311/regex
      copying regex_3/regex.py -> build/lib.linux-aarch64-cpython-311/regex
      copying regex_3/_regex_core.py -> build/lib.linux-aarch64-cpython-311/regex
      copying regex_3/test_regex.py -> build/lib.linux-aarch64-cpython-311/regex
      running build_ext                                       building 'regex._regex' extension
[...]

PR #3512, opened 2026-03-03
fix: correct dead-code tautology in FileSystemPathPointer.open()

ylwango613:
The sandbox condition at data.py:387 was permanently False:
```
  path = os.path.normpath(self._path)
  if os.path.isabs(path) and path != os.path.normpath(self._path):
```
Since `path` was assigned from `os.path.normpath(self._path)`, the second operand is identical to `path`, making the check unreachable.

Replace with `os.path.dirname(path) == path`, which is the standard Python idiom for detecting a filesystem root (e.g. "/" or "C:\\").

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -383,7 +383,7 @@ class FileSystemPathPointer(PathPointer, str):
         """
         path = os.path.normpath(self._path)
 
-        # Block raw absolute reads such as "/" "C:\\Windows" etc.
+        # Block direct access when the path is a filesystem root (e.g. "/" or "C:\\").
         if os.path.isabs(path) and os.path.dirname(path) == path:
             raise ValueError(f"Direct absolute file access blocked: {path}")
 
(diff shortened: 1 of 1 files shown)


ekaf:
Thanks @ylwango613  for exposing the dead-code bug in the sandbox logic. It’s important to note that this work pre-dates the broader security issue #3515 and deserves the credit for revealing that the existing security checks were functionally unreachable.

This PR is now superseded by a more comprehensive implementation in #3522 to handle these cases library-wide. But this PR was the original catalyst that showed the existing guards were ineffective. Great catch!

ekaf:
Everybody agrees to remove the dead-code bug, but it's not done yet. We should have merged this awesome contribution immediately, and we can still do it while we wait for #3522 to land. I would not mind rebasing #3522 afterwards.

alvations:
Okay, let me merge this in. @ekaf please rebase #3255

PR #3513, opened 2026-03-03
Resolve XSS vulnerability in local Wordnet browser search path

ylwango613:
The ?nextWord= parameter in the search path was reflected into the HTML response (both <title> and <body>) without escaping, allowing script injection. The 2022 fix (c8cedf10d) addressed the static page path only; this patch covers the remaining search path.

Escape the word at parse time with html.escape() before it enters the page-rendering pipeline.

The `html` module is part of Python’s standard library, so importing `html` should not cause any issues.

### DIFF
--- a/nltk/app/wordnet_app.py
+++ b/nltk/app/wordnet_app.py
@@ -48,6 +48,7 @@ Options::
 import base64
 import copy
 import getopt
+import html
 import io
 import os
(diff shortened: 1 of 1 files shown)


stevenbird:
Thanks @ylwango613 ylwango613

PR #3514, opened 2026-03-04
fix: comparison of a _WordNetObject to objects without _name (fixes #…

### LINKED ISSUE #1944: Wordnet: Comparison of  a _WordNetObject fails if other object is not a _WordNetObject.

Muhtasim-Munif-Fahim:
Fixes #1944

When comparing a `_WordNetObject` (like a `Lemma` or `Synset`) to an object that lacks a `_name` attribute (like a simple string `"INVALID"`), Python threw an `AttributeError`. 

This PR updates the `__eq__` and `__ne__` dunder methods in `_WordNetObject` (`nltk/corpus/reader/wordnet.py`) to catch the `AttributeError` and return `False` or `True` respectively, mirroring the standard Python comparison behavior.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -211,10 +211,16 @@ class _WordNetObject:
         return hash(self._name)
 
     def __eq__(self, other):
-        return self._name == other._name
+        try:
+            return self._name == other._name
+        except AttributeError:
+            return False
 
     def __ne__(self, other):
-        return self._name != other._name
+        try:
+            return self._name != other._name
+        except AttributeError:
+            return True
 
     def __lt__(self, other):
         return self._name < other._name


ekaf:
Thanks @Muhtasim-Munif-Fahim. This PR is not strictly necessary, but could be a good addition after considering Copilot's review suggestions below.

Returning NotImplemented when other isn’t a \_WordNetObject is the right approach here. And it is crucial to also do this in  \_\_lt\_\_ , otherwise @total_ordering still leaves </<= etc. raising AttributeError.

But we can drop \_\_ne\_\_ entirely — Python will derive != from \_\_eq\_\_, and omitting it avoids redundant logic.

ekaf:
I didn't understand why the PR author was not addressing the very simple review suggestions. But [here's the explanation](https://github.com/scrapy/scrapy/pull/7309#issuecomment-4006932141).

PR #3519, opened 2026-03-15
Clarify missing-entry LookupError in nltk.data.find

### LINKED ISSUE #3518: nltk.data.find() suggests downloading already-installed packages
`nltk.data.find()` (and therefore `nltk.data.load()`) emits a misleading `LookupError` message when a *specific file* is missing inside an NLTK data package that is already installed. 

The error suggests downloading the package (e.g. `nltk.download('stopwords')`) even though the package is present and readable. This behaviour is anti-helpful, since re-downloading is unnecessary  and cannot provide the missing file.

Minimal repro (first line proves `stopwords` is present; second line triggers the misleading message):

```python
import nltk
print(nltk.data.find("corpora/stopwords/english").file_size())
```

> 1048

```
nltk.data.find("corpora/stopwords/spanglish")
```

> LookupError: 
> **********************************************************************
>   Resource 'stopwords' not found.
>   Please use the NLTK Downloader to obtain the resource:
> 
>   >>> import nltk
>   >>> nltk.download('stopwords')
> 

This behavior is not `stopwords`-specific: it can be expected for *any* missing file in *any* NLTK data package, because the “please download …” hint is derived from the package name implied by the resource path, without distinguishing “package missing” from “file missing within an installed package”.

ekaf:
Fix #3518.

This changes `nltk.data.find()` so that when an NLTK data package (e.g. `corpora/stopwords` as a directory or `corpora/stopwords.zip`) is present but the requested entry is missing (e.g. `corpora/stopwords/spanglish`), the resulting `LookupError` no longer implies the entire package is missing.

The updated error message:
- references the exact missing resource path that was attempted,
- avoids the misleading `"Resource '<package>' not found"` wording in this case,
- still includes the `nltk.download('<package>')` suggestion as a secondary recovery step.

Tests:
- Add a unit test covering the “package present, entry missing” case using `stopwords` (skips if unavailable in the environment).

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -646,6 +646,7 @@ def find(resource_name, paths=None):
                 return ZipFilePathPointer(path_, resource_name)
             except OSError:
                 # resource not in zipfile
+                _note_near_miss(path_)
                 continue
 
         # Is the path item a directory or is resource_name an absolute path?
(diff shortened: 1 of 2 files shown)


ekaf:
Applied Copilot’s suggested tweaks: tightened near-miss detection, covered `.zip` entries in `nltk.data.path`, and strengthened the unit test to assert the old misleading wording is absent. LGTM.

stevenbird:
Nice feature!

PR #3520, opened 2026-03-16
fix: validate URLs in Downloader to prevent SSRF attacks

rishavkumarthapa01-sketch:
## Summary
Fixes chained SSRF vulnerability in `nltk.downloader` where `server_index_url` and package `url` fields were passed directly to `urlopen()` without validation.

## Changes
- Added `_validate_url()` function that validates URL scheme (http/https only)
- Blocks access to internal/metadata services (169.254.x.x, localhost, 127.x.x.x)
- Applied validation at both vulnerable sinks: line 956 and line 694

## Security Impact
Without this fix, an attacker can redirect NLTK's downloader to internal cloud metadata services (AWS IMDSv1 at 169.254.169.254) to exfiltrate IAM credentials.

## Related
- Reported via huntr.com bug bounty

PR #3522, opened 2026-03-17
Add central security sentinel for file and network access

### LINKED ISSUE #3504: Arbitrary File Read via Path Traversal in nltk.data.load() through Percent-Encoded Sequences
# Summary
nltk.data.load() and nltk.data.find() resolve user-supplied resource names to filesystem paths using url2pathname(), which decodes percent-encoded sequences (e.g. %2e%2e to ..). Path safety checks are performed on the raw, still-encoded string before decoding occurs. An attacker supplying %2e%2e instead of .. bypasses all path validation and reads arbitrary files outside the NLTK data directory.

# Vulnerable Code
nltk/data.py - find() function:
 url2pathname() decodes %2e%2e -> .. AFTER any safety check
p = os.path.join(path_, url2pathname(resource_name))
if os.path.exists(p):
    return FileSystemPathPointer(p)

# Proof of Concept
import nltk.data
nltk.data.path = ["/home/user/nltk_data"]
%2e%2e decodes to .. via url2pathname(), escaping the data dir
data = nltk.data.load("%2e%2e/SECRET_credentials.txt", format="raw")
print(data)
 b'AWS_SECRET_KEY=AKIAIOSFODNN7EXAMPLE\nDATABASE_PASS=hunter2\n'
All of these bypass path checks and decode identically:

# Payload	After url2pathname()
%2e%2e/secret	../secret
.%2e/secret	../secret
%2e./secret	../secret
%2E%2E/secret	../secret
Root Cause
url2pathname() is called after path safety checks, not before. Encoding .. as %2e%2e passes every check, then decodes to a traversal sequence at filesystem access time.

# Fix
Decode before checking:

from urllib.parse import unquote
resource_name = unquote(resource_name)  # decode first, then validate

# Reference
https://securityinfinity.com/research/path-traversal-in-nltks-nltk-data-load-via-percent-encoded-sequences

# Impact
An attacker who controls the resource name passed to nltk.data.load() can read any file the process has permission to access - credentials, environment files, SSH private keys, /etc/passwd, /proc/self/environ, application config files, etc. This affects any application that passes user-controlled input to nltk.data.load() or nltk.data.find().

### LINKED ISSUE #3509: Zip Slip abspath vs realpath
in https://github.com/nltk/nltk/pull/3468 - in this line https://github.com/nltk/nltk/pull/3468/changes#diff-79b3eaf750476210dbc2c7589a72a3d4c1a31e5f74a542f2e574feb964215e32R2323 aren't we comparing a `realpath` with an `abspath` ? is that needed for the vulnerability fix?

this is breaking heroku builds (as an example) see ie https://github.com/heroku/heroku-buildpack-python/issues/2037 .

On Heroku, /app is a symlink to /tmp/build_xxxxx/, so realpath and abspath diverge, and every zip entry gets blocked. Is that deliberate?

### LINKED ISSUE #3515: [Security] Dead security check in FileSystemPathPointer.open() allows arbitrary file read
# Description
A logical error in the security patch of `nltk.data.FileSystemPathPointer.open()` renders the intended sandbox completely ineffective. The method attempts to block direct access to absolute paths (e.g., `/etc/passwd`) by comparing the normalized path against itself, a condition that can never be true. As a result, any caller can instantiate a `FileSystemPathPointer` with an arbitrary absolute path and read the corresponding file without restriction. This bypasses all existing path validation and enables arbitrary file read on the host system.

1. **Vulnerability Type**: Path Traversal / Arbitrary File Read
2. **Affected Component**: `nltk/data.py` – `FileSystemPathPointer.open()` method
3. **Affected Versions**: NLTK 3.9.3 (and possibly earlier)
4. **CVSS v3.1 Score: 7.5 (High)** – AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:N

# Root Cause

The vulnerable code in `nltk/data.py` (lines `380–388` in the provided version) is:
```python
def open(self, encoding=None):
    """
    Secure open — prevents absolute direct access outside pointer root.
    """
    path = os.path.normpath(self._path)

    # Block raw absolute reads such as "/" "C:\\Windows" etc.
    if os.path.isabs(path) and path != os.path.normpath(self._path):
        raise ValueError(f"Direct absolute file access blocked: {path}")

    stream = open(self._path, "rb")
    # ...
```

**Explanation**:

- path is assigned os.path.normpath(self._path).
- The condition then checks path != os.path.normpath(self._path).
- Since path is os.path.normpath(self._path), the inequality is always False.
- Therefore, the ValueError is never raised, and any absolute path passes the check.

The comment claims this is a "secure open", but the guard is completely dead code.

# Attack Scenario

An attacker who can control the input passed to any function that eventually creates a `FileSystemPathPointer` can read arbitrary files. Common entry points include:

- Direct use of `nltk.data.FileSystemPathPointer` with a user‑supplied path.
- Use of corpus readers (e.g., `WordListCorpusReader`,` TaggedCorpusReader`, `BracketParseCorpusReader`) that accept a root PathPointer and a list of file identifiers. An attacker could supply a root pointing to `/` and a file identifier like `etc/passwd`.
- Any application that loads NLTK resources based on external input (e.g., a web service that accepts a corpus name or file path).

Because the check is dead, no validation prevents accessing sensitive system files (e.g., `/etc/passwd`, application configuration, source code).

# Proof of Concept (PoC)

The following Python script demonstrates the vulnerability by reading `/etc/passwd`:
```python
#!/usr/bin/env python3
"""
PoC: NLTK data.py – Dead security check allows arbitrary file read
"""
import os
from nltk.data import FileSystemPathPointer

# === Part 1: Show the security check is dead ===
print("[1] Analyzing the dead security check:")
ptr = FileSystemPathPointer("/")
path = os.path.normpath(ptr._path)
condition = os.path.isabs(path) and path != os.path.normpath(ptr._path)
print(f"    Guard condition = {condition} (always False)\n")

# === Part 2: Exploit – read /etc/passwd ===
print("[2] Reading /etc/passwd via FileSystemPathPointer:")
try:
    target = FileSystemPathPointer("/etc/passwd")
    with target.open() as f:
        data = f.read().decode()
    print(data[:300])  # first 300 characters
except Exception as e:
    print(f"Failed: {e}")
```

Expected output (on Unix‑like systems):

```
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
bin:x:2:2:bin:/bin:/usr/sbin/nologin
...
```

# Fix 

Replace the dead condition with a proper sandbox check. For example, define a base directory that all paths must reside under, and verify that the resolved absolute path starts with that base directory. A minimal fix might look like:

```python
def open(self, encoding=None):
    # Define a safe base directory (e.g., the intended data root)
    safe_base = os.path.abspath(os.path.dirname(__file__))  # or a configurable path
    requested_path = os.path.abspath(self._path)
    if not requested_path.startswith(safe_base):
        raise ValueError(f"Access denied: {requested_path} is outside allowed directory")
    stream = open(self._path, "rb")
    # ...
```

# References

- [Huntr Report](https://huntr.com/bounties/70037373-8d33-4aa9-abba-ad6ad7256535) (For further process, please follow the platform report or start the process from here.)

ekaf:
Fix #3504, #3509, #3515. Supersedes (replaces) #3512 and #3528.

#### The Context
NLTK has been receiving a steady stream of vulnerability reports (like path traversal and SSRF) via platforms like Huntr. Continuing to apply scattered, defensive patches to individual functions every time a report comes in is difficult to maintain and leaves the codebase vulnerable to future oversights. 

#### The Solution
To address this comprehensively, this PR introduces a central, proactive security layer (`nltk.pathsec`). We implement secure, drop-in replacements for the standard library functions that initiate file and network access (`open`, `ZipFile`, and `urlopen`). By importing these secure versions at the top of our highest-leverage files, we automatically protect the library against path traversal, Zip-Slip, and SSRF attacks at the point of entry, before any data is read or written.

#### Why this approach?
* **Proactive Access Control:** Future modules that need to access local files or fetch remote URLs can just import our safe wrappers, ensuring they are secure by default at the moment the connection or file handle is created.
* **Sequential, Independent Defense:** We left the recent string-checking patches in place. This new central layer runs in sequence with them, acting as the final gatekeeper for file/network handles. Because these two security layers are completely independent, it allows us to compare their effectiveness and provides true defense-in-depth. 
* **Superior to Localized Fixes:** We built an implementation-agnostic test module (`test_pathsec.py`) to compare this centralized approach against the localized fixes recently proposed across multiple PRs:
    * **Re: SSRF (PR #3520):** Our tests prove that the sentinel catches an advanced SSRF attack (using decimal IP obfuscation) that #3520 misses, because we resolve the actual underlying IP address rather than just scanning the URL string. 
    * **Re: Path/Zip Traversal (@HyperPS PRs):** Our tests prove that the sentinel catches advanced filesystem bypasses (like injecting absolute paths such as `/etc/passwd` directly) that localized `../` string checks often miss.

#### Two-Phase Rollout and Backward Compatibility
To evaluate the impact of this layer, we ran the entire NLTK CI test suite with strict enforcement enabled (`ENFORCE = True`). **All tests passed flawlessly**, proving that NLTK's internal operations are secure and compatible with this architecture.

However, to ensure zero disruption for external users who may have non-standard data paths or proxy setups, we are rolling this out in two phases:
1. **Transition Phase (Current PR):** This layer ships with the `ENFORCE` toggle set to `False` (Dry-Run Mode). It logs `RuntimeWarning`s for suspicious behavior but allows access to proceed. This allows users to audit their setups without breaking their current workflows.
2. **Enforcement Phase (Future Release):** Once the community has adapted, we will flip the switch to strict enforcement. This will likely accompany a minor version bump (e.g., 3.10.0), so that we can reserve the 4.0 release for a future Python 4.0 transition.


ekaf:
Thanks to Copilot for the sharp initial review! I've pushed an update addressing the critical findings:

* **Filesystem:** Fixed integer file descriptor handling in `open()` and prevented ZIP prefix-confusion by using path-aware containment checks.
* **Network:** Added IPv6 SSRF support by switching to `getaddrinfo`.
* **Architecture:** Made the `NLTK_DATA` root resolution dynamic so it respects runtime changes to `nltk.data.path`.
* **Tests:** Migrated the test suite to use cross-platform `tmp_path` fixtures and cleaned up unused imports.

Re-requesting Copilot for a fresh pass!

nltk/test/unit/test_pathsec.py
```diff
@@ -0,0 +1,136 @@
...
+def test_valid_http_url():
+    dl = Downloader(
+        server_index_url="https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/index.xml"
+    )
+    try:
+        dl.index()
+    except URLError:
+        pass
+
+
```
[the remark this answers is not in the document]
ekaf: We deliberately make a real HTTP request here to verify that the OS-level DNS resolution layer (getaddrinfo) functions correctly on live URLs. Mocking this would defeat the purpose of the test.

ekaf:
Implemented the final architectural updates and resolved a CI performance bottleneck:

* **O(N²) Zip Extraction Fix:** Refactored `validate_zip_archive` to perform fast, in-memory path math (`os.path.abspath`) and target specific archive members during extraction, preventing a massive CPU timeout on large corpora like FrameNet.
* **SSRF & Paths:** Blocked private IP ranges (`is_private`) and adopted `os.fspath` to safely parse Windows drive letters. 
* **Targeted Aliasing & Sandboxing:** Aliased `_secure_open` strictly within `data.py` to protect corpus readers without globally shadowing `builtins.open`. Refactored `util.filestring` to use a self-contained `Path.resolve()` boundary check, fully preserving its legacy `UnicodeDecodeError` fallbacks.

alvations:
Summarizing the commits in this PR a bit. 

###  What it does

  It introduces a centralized security module (nltk/pathsec.py) that provides:

  1. `validate_path()` — Checks file paths are within allowed NLTK data roots (nltk_data dirs, cwd, temp). Blocks access outside these boundaries.
  2. `validate_network_url()` — SSRF protection: validates URLs use http/https, resolves hostnames via DNS and blocks loopback, link-local, multicast, and private IPs (covers decimal IP obfuscation like
  http://2852039166/).
  3. `validate_zip_archive()` — Zip-slip protection: checks zip members for path traversal (../) and absolute paths before extraction.
  4. Wrapper classes/functions — pathsec.open(), pathsec.urlopen(), pathsec.ZipFile that validate before delegating to builtins.

## CVEs it addresses

  - CWE-22 (Path Traversal) — Replaces the inline security patch in CorpusReader.open() (from PR #3479) with centralized path validation. Removes the per-file patch and moves it to the boundary.
  - SSRF — urlopen wrapper blocks requests to internal/private IPs including obfuscated forms (decimal IP 2852039166 = 169.254.169.254). The existing PR #3520 was vulnerable to this.
  - Zip-Slip — ZipFile subclass validates members on extract()/extractall(). Replaces raw zipfile.ZipFile usage in nltk/data.py and nltk/downloader.py.



## Changes will
  - Centralizes security in one module instead of scattered inline patches
  - ENFORCE = False by default (warns only) — safe rollout, no breaking changes
  - DNS-based SSRF check catches IP obfuscation that string-matching misses
  - Zip audit uses os.path.abspath join check (correct approach)
  - Removes the inline patch from CorpusReader.open() that was added by PR #3479 — cleaner
  - Good test coverage: SSRF (loopback, cloud metadata, decimal IP), path traversal, zip-slip

##  Concerns:
  - _get_allowed_roots() includes os.getcwd() — an attacker who controls cwd could bypass path checks
  - ENFORCE = False means all checks are warnings by default — users must opt in to actual enforcement
  - validate_path is called on every open() call, recomputing _get_allowed_roots() each time (no caching)
  - The childes.py change is cosmetic (commented-out code)
  - filestring() encoding change (adding encoding param, latin-1 fallback) is a behavior change unrelated to security

----

@ekaf Hmmm, lets see how to address some of these concerns esp. `_get_allowed_roots`

nltk/data.py
```diff
@@ -378,19 +378,11 @@ def path(self):
...
     def open(self, encoding=None):
         """
-        Secure open — prevents absolute direct access outside pointer root.
+        Return a seekable read-only stream for the file identified by
+        this path pointer.  Path validation is handled by pathsec.open().
         """
-        path = os.path.normpath(self._path)
-
-        # Block raw absolute reads such as "/" "C:\\Windows" etc.
-        if os.path.isabs(path) and path != os.path.normpath(self._path):
```
alvations: This if-case will never be entered  `normpath` is deterministic, so `normpath(x) != normpath(x)` is always false.

alvations: We can safely remove this.

alvations:
@ekaf thanks again for the PR! This will be very very helpful to avoid more CVE related PRs. 
Hope the commit patches I put in after yours help to avoid that os.cwd() issue. 

@stevenbird @ekaf When the PR is ready, I'll be glad to take a second look and hopefully merge this in soon.

nltk/data.py
ekaf: Sure @alvations, I did not initially include that fix because it already has its separate PR #3512 by @ylwango613.

ekaf:
Thanks @alvations for raising these concerns:
  
>   _get_allowed_roots() includes os.getcwd() — an attacker who controls cwd could bypass path checks

On the other hand, users would probably perceive it as a breaking change, not to be allowed to read or write in their cwd. Maybe this should only raise a warning?

>  ENFORCE = False means all checks are warnings by default — users must opt in to actual enforcement

Yes, the plan is to roll out in two phases, and first enforce the strict policy after a transition period.

>  validate_path is called on every open() call, recomputing _get_allowed_roots() each time (no caching)

Yes, a cache could be nice.

>  The childes.py change is cosmetic (commented-out code)

True, but it is more helpful and informative than leaving the insecure call in the original comment, or removing it.

>  filestring() encoding change (adding encoding param, latin-1 fallback) is a behavior change unrelated to security
 
 I'll reconsider this.

ekaf:
Maybe we should ask the question: who is NLTK intended for: students/researchers or enterprise servers? These segments have very different needs.

ekaf:
Thanks for the review, @alvations! I just pushed a commit that addresses your primary concerns regarding performance and path resolution:

**1. Caching `_get_allowed_roots()`**
You were completely right that `validate_path` was recomputing the roots on every single `open()` call, which was a performance drain. 
* I have implemented an `_ALLOWED_ROOTS_CACHE`. 
* To ensure it doesn't break dynamic workflows, it tracks the state of `nltk.data.path` and the `NLTK_DATA` environment variable, automatically busting the cache if a user modifies those at runtime.

**2. The `os.getcwd()` Bypass Vector**
Good catch on `cwd` being a potential bypass vector if the attacker controls it. I agree it is a risk for server environments, but removing it entirely right now would immediately break thousands of local scripts for students using `CorpusReader` in their project folders. 
* **The Compromise:** I removed `cwd` from the standard allowed roots, but added it as a specific fallback in `validate_path`. 
* If a path is allowed *exclusively* because it matches `cwd`, it now emits a targeted `RuntimeWarning` explicitly alerting enterprise/server users that relying on `cwd` is insecure, while preserving backward compatibility for educational use cases. 

Let me know what you think of this balance!

stevenbird:
> Maybe we should ask the question: who is NLTK intended for: students/researchers or enterprise servers? These segments have very different needs.

NLTK was originally intended for NLP students. For this reason, we often preferred straightforward reference implementations of standard algorithms, rather than less-readable, optimised versions.

For years NLTK has been used in commercial environments, but we have never received commercial support (eg donations). I think companies can solve their own problems and hopefully contribute solutions back to our codebase.

So, with scarce developer resources, we are strictly limited in what we can take on.

nltk/pathsec.py
```diff
@@ -134,9 +134,11 @@ def validate_path(path_input, context="NLTK"):
         else:
             warnings.warn(msg, RuntimeWarning, stacklevel=3)
 
-    except Exception:
+    except (PermissionError, ValueError):
+        raise
+    except (OSError, TypeError) as e:
         if ENFORCE:
-            raise
+            raise PermissionError(f"Path validation failed [{context}]: {e}") from e
```
alvations: Raising explicit instead of bare errors.

nltk/pathsec.py
```diff
@@ -175,9 +177,11 @@ def _audit(zf):
         else:
             with zipfile.ZipFile(zip_obj_or_path, "r") as zf:
                 _audit(zf)
-    except Exception:
+    except (PermissionError, ValueError):
+        raise
+    except (OSError, zipfile.BadZipFile) as e:
         if ENFORCE:
-            raise
+            raise PermissionError(f"Zip validation failed [{context}]: {e}") from e
```
alvations: As above.

nltk/pathsec.py
```diff
@@ -199,9 +203,15 @@ def validate_network_url(url_input, context="NetworkIO"):
         if not hostname:
             return
 
-        # Use getaddrinfo to capture both IPv4 and IPv6 resolutions
+        # Use getaddrinfo to capture both IPv4 and IPv6 resolutions.
+        # Set a timeout to prevent hanging on slow/unresponsive DNS.
         try:
-            addr_info = socket.getaddrinfo(hostname, None)
+            old_timeout = socket.getdefaulttimeout()
+            socket.setdefaulttimeout(5)
```
alvations: This timeout thing is necessary to avoid the ever hanging state esp. when user is behind a proxy.

nltk/data.py
```diff
@@ -387,7 +391,7 @@ def open(self, encoding=None):
         if os.path.isabs(path) and path != os.path.normpath(self._path):
             raise ValueError(f"Direct absolute file access blocked: {path}")
 
-        stream = open(self._path, "rb")
+        stream = _secure_open(self._path, "rb")
```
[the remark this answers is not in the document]
alvations: Outdated.

nltk/pathsec.py
```diff
@@ -0,0 +1,252 @@
...
+
+    # 4. System temp dir
+    import tempfile
+
+    try:
+        roots.add(Path(tempfile.gettempdir()).resolve())
+    except Exception:
+        pass
+
+    return roots
```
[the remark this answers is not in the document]
alvations: Outdated.

nltk/pathsec.py
```diff
@@ -0,0 +1,252 @@
...
+        if not hostname:
+            return
+
+        # Use getaddrinfo to capture both IPv4 and IPv6 resolutions
+        try:
+            addr_info = socket.getaddrinfo(hostname, None)
+            for result in addr_info:
+                ip_str = result[4][0]
+                ip_obj = ipaddress.ip_address(ip_str)
+
```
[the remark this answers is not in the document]
alvations: Outdated.

alvations:
Thanks @ekaf for the updates! I've added some minor improvements to (i) raise explicit error and (ii) make the DNS checks timeout and (ii) added notes on SECURITY.md

alvations:
The _secure_open should fix the `FileSystemPathPointer` issues.

But there's still a hanging thing when `ENFORCE=False` and users blindly just point the `nltk_data` to any path that might include malicious symlink. But it is very very often `nltk_data` is a symlink to a virtual mount and this is no different from popular libraries trusting their repository's default path and letting users symlink it to share across cluster's users. There's no clean way to get around it.

alvations:
@stevenbird @ekaf I'll say LGTM now, at least secure_open helps us to cut down the GHSA =) 
Thanks @ekaf for the awesome effort in this PR!!!

We can iterate for symlink related issues in the future when pythonic tools become more available.

alvations (approved):
LGTM! Thanks again @ekaf!!

ekaf:
Thanks @alvations. Many of your fixes seem to harden the local validation. In the meantime, I have also been addressing the same Copilot concerns, but in the central module. Combining both series of fixes prior to merging this PR should give us more robust security, since we aim at running two independent layers of validation.

ekaf:
This update integrates the latest fixes by @alvations, and completes the **Layer 2** security sentinel in `pathsec.py`. The following technical refinements have been implemented to address the Copilot Security Review:

* **DNS Rebinding & SSRF:** Added cached hostname resolution in `validate_network_url`. Now explicitly blocks **Loopback**, **RFC 1918 (Private)**, **Link-Local**, **Multicast**, and **Cloud Metadata** (169.254.169.254) ranges.
* **Proactive Zip-Slip Audit:** Implemented a pre-extraction audit in `validate_zip_archive` to block members attempting traversal (`../`) or containing null bytes.
* **Resolved Path Enforcement:** Migrated all containment checks to use `Path.resolve()` + `is_relative_to()` for guaranteed boundary enforcement.
* **Performance Optimization:** Added state-based caching for `_get_allowed_roots` to eliminate redundant disk I/O during high-volume processing.
* **Zip-Transparency:** Implemented path truncation logic to handle NLTK "virtual paths" inside archives without breaking internal readers.

The implementation is verified green across all CI checks and maintains full compatibility with NLTK’s internal "Zip-magic" and `PathPointer` logic.

ekaf:
### Final Security Hardening & Protocol Resolution

This update performs a final architectural pass to resolve low-level edge-cases identified during integration:

* **SSRF Redirect Protection:** Implemented `_ValidatingRedirectHandler` to ensure that every step of an HTTP redirect chain is re-validated against restricted IP ranges.
* **Thread-Safe DNS Resolution:** Replaced global `socket.setdefaulttimeout` modifications with a thread-safe resolution strategy using `socket.getaddrinfo`.
* **Explicit CWD Opt-In:** Refined the "Safety Valve" so that in `ENFORCE` mode, access to the Current Working Directory requires an explicit `nltk.data.path.append('.')`.
* **Environment Path Correction:** Switched from `os.sep` to `os.pathsep` for parsing `NLTK_DATA`, preventing accidental directory whitelisting on POSIX systems.
* **Sentinel Bottlenecking:** Eliminated remaining bypasses in `nltk.data` and `nltk.util` by routing all fallback file operations through the `pathsec` sentinel.
* **Documentation Alignment:** Updated `SECURITY.md` to explicitly define the CWD opt-in behavior and the protocol for authorizing local directories in strict mode.

Now I'll request Copilot again, though it seems likely that it is never going to conclude cleanly.

ekaf:
The situation is not completely clear around the 7 suggestions from Copilot's latest review: some appear to reference code from earlier iterations that has already been revised. I resolved all, but am going to look again.

ekaf:
Thanks for the review. We implemented 3 surgical fixes (CI green ✅):

1. **`file:` URL decoding** (`pathsec.py` L80, L186) — `unquote(parsed.path)` → `urllib.request.url2pathname(parsed.path)`, which also handles Windows drive letters and UNC paths.

2. **`urlopen()` kwargs** (`pathsec.py` L222–227) — Now uses `install_opener()` + `urllib.request.urlopen()` instead of `opener.open()`, so `cafile`/`capath`/`context` are correctly forwarded.

3. **Test portability** (`test_pathsec.py` L85–97) — Replaced hardcoded `/etc/passwd` with a cross-platform absolute path outside all allowed roots.

The remaining suggestions (URL parsing in `_open()`, `CorpusReader.open()` guards, `filestring()` validation, zip prefix checks) were already addressed in earlier iterations of this branch.

Asking for more Copilot reviewing might be counter-productive, so I am inclined to think we have reached our goals for now.

nltk/pathsec.py
```diff
@@ -223,8 +226,10 @@ def urlopen(url, *args, **kwargs):
     """Secure wrapper for urllib.request.urlopen with redirect validation."""
     url_str = url.full_url if hasattr(url, "full_url") else str(url)
     validate_network_url(url_str, context="pathsec.urlopen")
-    opener = urllib.request.build_opener(_ValidatingRedirectHandler())
-    return opener.open(url, *args, **kwargs)
+    urllib.request.install_opener(
+        urllib.request.build_opener(_ValidatingRedirectHandler())
+    )
+    return urllib.request.urlopen(url, *args, **kwargs)
```
alvations: A bit terse here but I'm assuming the this will always install opener and prevent redirect.

alvations: But only when `urlopen` is used. 

So my auto-code reviewer is suggesting this:


```
  # Module level: defense-in-depth for any direct urllib.request.urlopen() calls                                                                                                  
  urllib.request.install_opener(                                                                                                                                                  
      urllib.request.build_opener(_ValidatingRedirectHandler())                                                                                                                   
  )                                                                                                                                                                               
                             
                                                                                                                                                                                  
  def urlopen(url, *args, **kwargs):
      """Secure wrapper for urllib.request.urlopen with redirect validation."""                                                                                                   
      url_str = url.full_url if hasattr(url, "full_url") else str(url)                                                                                                            
      validate_network_url(url_str, context="pathsec.urlopen")                                                                                                                    
      # Local opener — not dependent on global state surviving                                                                                                                    
      opener = urllib.request.build_opener(_ValidatingRedirectHandler())                                                                                                          
      return opener.open(url, *args, **kwargs)  
```

Hmmmm, this makes anything that `urllib.request.install_opener` always checks for redirects globally. Makes sense but maybe it shouldn't be in pathsec then. Let me think a little.

alvations: @ekaf Lets try this, put this in pathsec.py globally

```python
if ENFORCE:
    urllib.request.install_opener(
      urllib.request.build_opener(
        _ValidatingRedirectHandler()
      )                                                                                                             
    )
```

Then inside the `urlopen` we continue with the old:

```python
def urlopen(url, *args, **kwargs):
    """Secure wrapper for urllib.request.urlopen with redirect validation."""
    url_str = url.full_url if hasattr(url, "full_url") else str(url)
    validate_network_url(url_str, context="pathsec.urlopen")
    opener = urllib.request.build_opener(_ValidatingRedirectHandler())
    return opener.open(url, *args, **kwargs)
```

Would that be okay?

nltk/test/unit/test_pathsec.py
```diff
@@ -90,11 +90,14 @@ def test_path_traversal_absolute():
...
-        # On vulnerable branches without pathsec, default to standard open
         target_open = builtins.open
 
+    # Cross-platform absolute path guaranteed outside all allowed roots.
+    # Linux/macOS: /_nltk_pathsec_test/secret.txt
+    # Windows:     C:\_nltk_pathsec_test\secret.txt
+    outside = os.path.join(os.path.abspath(os.sep), "_nltk_pathsec_test", "secret.txt")
     with pytest.raises((ValueError, PermissionError)):
-        target_open("/etc/passwd", "r")
+        target_open(outside, "r")
```
alvations, line 100: Good catch! Windows folks will thank you =)

alvations:
Thanks @ekaf on the addressing the co-pilot's suggestions! Left a final suggestion in the global opener redirect enforcing when `ENFORCE==True`. Other than that, it LGM! Thank you again for the work!

stevenbird:
Thanks @alvations and @ekaf. Looks like we're nearly there.

ekaf:
@alvations Thanks for the redirect suggestions. I updated nltk.pathsec.urlopen() so that it checks redirect targets too, but without changing urllib’s global behavior. In other words, NLTK gets the extra safety when it uses pathsec.urlopen(), and users/libraries are still free to use urllib however they normally would.

I also reinstated a previous security patch, which was earlier removed by accident from open() in corpus/reader/api.py.

As far as I can see, this should now be GTM.

alvations:
@ekaf thanks for the updates on the global opener. 

@stevenbird We should be good to merge now. Updating a main branch and pushing a release would be good after this.

ekaf:
This PR now meets the protection provided by #3528 and guards against corpus-root escapes (including symlink-based ones) at a central level. 

@stevenbird @alvations Agreed—let’s merge this. The central sentinel gives NLTK a big security boost, and any further tweaks can follow if new issues appear. 

There’s no need for a release while ENFORCE stays False; users will only see warnings, not breaking changes. We can plan a new release when ENFORCE=True, since that’s when strict enforcement (and potential breaking changes) actually kick in. Thanks all!

stevenbird:
Thanks @ekaf... what a great idea to centralise this and roll it out in two stages. Thanks for careful reviewing @alvations

ekaf:
I'd like to thank several LLMs for invaluable assistance during this project: mostly Gemini's Pro and Thinking models, and Copilot's 5.2 and Claude Opus 4.6: whenever one of these models lead me in trouble, another was able to help me out. Less capable models, however, proved unfit for this task.

ekaf:
So when do we move to the second stage by setting ENFORCE = True?

If the goal is to expedite the closure of the known vulnerabilities while mitigating severe ecosystem shock, a three to six-month total transition could be a defensible compromise:

1. Month 1-3 (Current): Continue emitting warnings under the ENFORCE = False configuration.

2. Month 3: Publicly announce an explicit date and target version for the hard switch.

3. Month 4-6: Release the minor version bump with ENFORCE = True.

stevenbird:
@ekaf this sounds good me. We don't have an obvious way to do the announcement though. (Perhaps another minor release prior?)

ekaf:
Thanks, Steven. To announce and manage this, I suggest targeting a **3.10.0 release prior to the fall semester**. 

While our internal CI passes with the strict security enforcement, we know this will disrupt a large number of downstream tutorials and notebooks. Here is the proposed roadmap:

1. **Audit the Book:** First, we will test the `nltk_book` repository against the strict policy to catch and patch any educational failures.
2. **Bundle Breaking Changes:** We can use the 3.10.0 bump to consolidate `ENFORCE = True` alongside other breaking changes (e.g., PR #3557) to minimize overall ecosystem churn.
3. **Release Candidate:** As we approach the fall semester, we will publish a 3.10.0 RC accompanied by specific documentation on how users must update their path configurations.

If this timeline sounds good, we can start by checking the book repo and drafting the RC documentation.

stevenbird:
@ekaf I like this proposal. We might want to consider the earlier timing requirement of instructors preparing materials ahead of the fall semester.

ekaf:
Sure, I also prefer a shorter window, and would not mind shipping the patch now. 
The plan above was the result of bargaining with Gemini, who initially argued for a 6-12 month window, based on the following objections:

> Examining standard dependency management patterns reveals why a total warning period of just two months is generally considered insufficient, even when expediting a security patch.
> 
> The Problem with a Two-Month Window
> - Sprint Cycle Realities: Many enterprise teams operate on quarterly planning cycles or multi-week sprints. A warning emitted today might be noticed next week, converted into an issue ticket a week later, and placed in a backlog for a subsequent sprint. Two months is often the minimum time it takes for a non-critical maintenance ticket to simply begin development.
> 
> - Warning Blindness: In automated environments, standard error logs and warning outputs are routinely ignored as long as the build exits with a success code. Many developers will not realize their pipeline is insecure until the hard enforcement actually fails the build.
> 
> - Cascading Dependencies: If a project relies on a secondary library that internally wraps NLTK, the end-user cannot fix the path issue themselves. They must wait for the intermediate library maintainers to patch their usage of NLTK and release a new version. A two-month window rarely accommodates this chain of updates.

PR #3528, opened 2026-03-18
Prevent symlink-based path traversal in CorpusReader.open()

CHOOCS:
## Summary

This PR fixes a security issue in CorpusReader.open() where path validation relied solely on lexical checks (os.path.isabs and .. filtering), which could be bypassed using symbolic links within the corpus root. An attacker could exploit this to read arbitrary files outside the intended directory.

## Root Cause
The previous implementation:
- Blocked absolute paths
- Rejected paths containing `..`

However, it did not validate the resolved filesystem path, allowing symlink-based escapes such as:

```
corpus_root/outside_corpus -> /
outside_corpus/etc/passwd
```

## Changes
- Added canonical path resolution using os.path.realpath
- Enforced strict containment check:
    - Ensure resolved file path remains within resolved corpus root
- Applied check only for filesystem-backed corpora (FileSystemPathPointer)
- Preserved existing validation for:
    - Absolute paths
    - `..` traversal attempts

## Testing
- Existing test suite passes without modification
- Manually verified with PoC:
    - Symlink escape is now correctly blocked
    - Legitimate corpus file access remains unaffected

## Patched Version
```
reader.open("outside_corpus/etc/passwd")
# Raises ValueError: Resolved path escapes corpus root
```

### DIFF
--- a/nltk/corpus/reader/api.py
+++ b/nltk/corpus/reader/api.py
@@ -222,9 +222,9 @@ class CorpusReader:
     def open(self, file):
         """
         Return an open stream that can be used to read the given file.
-        Security patched: prevents path traversal & absolute path access.
+        Securely prevents path traversal, absolute path access, and symlink escapes.
         """
-        # -------- SECURITY PATCH START --------
+
         file = str(file)
 
         if os.path.isabs(file):
@@ -232,11 +232,19 @@ class CorpusReader:
 
         if ".." in file.replace("\\", "/").split("/"):
             raise ValueError("Path traversal attempt blocked")
-        # -------- SECURITY PATCH END --------
 
         encoding = self.encoding(file)
-        stream = self._root.join(file).open(encoding)
-        return stream
+        candidate = self._root.join(file)
+
+        # Enforce containment for filesystem-backed corpora
+        if isinstance(self._root, FileSystemPathPointer):
+            root_real = os.path.realpath(self._root.path)
+            candidate_real = os.path.realpath(candidate.path)
+
+            if os.path.commonpath([root_real, candidate_real]) != root_real:
+                raise ValueError("Resolved path escapes corpus root")
+
+        return candidate.open(encoding)
 
     def encoding(self, file):
         """
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -311,6 +311,7 @@
 - Jose Cols <https://github.com/josecols>
 - Christopher Smith <https://github.com/smithct2>
 - Ryan Mannion <https://github.com/ryanamannion>
+- CHOOCS <https://github.com/CHOOCS/>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 


CHOOCS:
Hi @ekaf, wondering does this count as a temporary fix at the moment? Also, is it possible for you to validate the report on Huntr platform.

alvations:
@CHOOCS this will be superceded by #3522 and will most probably be closed without merged.

nltk/corpus/reader/api.py
```diff
@@ -222,21 +222,29 @@ def raw(self, fileids=None):
     def open(self, file):
         """
         Return an open stream that can be used to read the given file.
-        Security patched: prevents path traversal & absolute path access.
+        Securely prevents path traversal, absolute path access, and symlink escapes.
```
alvations, line 225: This is sign of automated bot slop.

ekaf: Indeed, but do we mind here? The proposed edit is still a slight improvement.

alvations: Documentation improvement for language, agree a slight improvement, the implementation of a fix, most probably a band-aid that is not going to go well. So my opinion is to not encourage these merges into branch cos the bots will take it as a positive signal for future minor doc improvement + bad code logic as high probability of getting merged.

ekaf: Sure, I understand now how this could lead to a flood of insignificant edits.

ekaf:
This PR could be a useful complement to #3522, because it is blocking escapes outside the corpus root, while #3522 checks the allowed roots.
The doctest from the Copilot review should be used as unit test instead, so we can ensure that both PRs block the same attacks, which they don't yet.

CHOOCS:
Hi @ekaf, I get the reason to not merge the PR since you have a longer term fix, but this remain a valid security finding and whether it is bot or not does not matter. A valid finding should be awarded with the bounty and the cve, it is ok to reject the patch and skip the patch award.

ekaf:
Hi @CHOOCS — understood. I agree the *validity of a security finding* is separate from whether we merge this PR.

That said, I’m not involved in huntr administration and I don’t have access to/knowledge of their bounty decision process, so I can only wish you good luck!

ekaf:
I think this PR could be a useful addition to #3522, by providing a second layer of protection.
But it it is not yet ready to merge. On one hand, it has unaddressed Copilot review suggestions. On the other hand, it needs rebasing, now that #3522 is merged.
Please note that the latest #3522 commit added a unit test that tests symlink-based path traversal in CorpusReader.open(), and can replace the doctest proposed in the Copilot review above,  since it also illustrates the functionality of the present PR.

themavik (commented):
block probe nltk

themavik (commented):
@/run/media/zoom/STORAGE/Github/._rb/n1.txt

ekaf:
This looks abandoned, It has unresolved conflicts, and is not crucially needed since #3522 fully solved the same problem.

CHOOCS:
Hi @ekaf, saw your email regarding re-opening this PR, I can open this PR again if you need to fix this. Thanks!

CHOOCS:
Hi, I have commented on the closed PR whether to re-opening the PR to fix
the path traversal issue

Thanks!

Regards,
Choo

ekaf:
Hi @CHOOCS, I was in doubt but it really seems that #3522 fully solved the problem, and adds adequate tests. Please let me know if you have proof of the contrary.

PR #3534, opened 2026-03-20
Fix retest workflow for fork-based PRs

alvations:
The createWorkflowDispatch only works for branches in the same repo, so [CI: retest] on fork PRs silently failed. For fork PRs, use reRunWorkflow on the existing CI run (matched by head SHA) instead. Keep createWorkflowDispatch for same-repo branches.

Evidence: https://github.com/alvations/nltk/pull/84

With this change in alvations/nltk develop branch, we simulated a bad pull-request from yet another fork to the alvations/nltk fork and then checked that the `[CI: retest]` works across forks. 

Then true test here would be that the next comments will auto re-trigger the CI/CD after this PR is merged.

### DIFF
--- a/.github/workflows/retest.yml
+++ b/.github/workflows/retest.yml
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       actions: write
+      issues: write
       pull-requests: read
     steps:
       - name: Get PR head info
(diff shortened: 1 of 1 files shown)


stevenbird (approved):

alvations:
Thanks @stevenbird for checking and reviewing the PR! Merging this in and now we can `[CI: retest]` again =)

PR #3535, opened 2026-03-20
Fix retest workflow for fork-based PRs

alvations:
It seems like the old retest has a corrupted github actions causing the workflow to fail and github register as broken, so it wont parse the yaml. 

This PR replace retest.yml with retest-pr.yml that works for both fork and same-repo PRs.

- Fork PRs: re-run the existing CI workflow run (preserves fork context)
- Same-repo PRs: dispatch a new CI workflow run
- Trigger phrases: /retest, [CI: retest], [ci: retest]
- Removed invalid reactions: write permission that silently broke GitHub's workflow registration

Will need to merge and test this out first. I'll revert if this doesn't work.


alvations:
Now it works for fork PR, from https://github.com/nltk/nltk/pull/3289 -> https://github.com/nltk/nltk/actions/runs/23359677493

Next, making it work for own NLTK's branch too.

PR #3537, opened 2026-03-20
Use reRunWorkflow for both fork and branch PRs

alvations:
Results now attach to the PR check suite for all PR types. Skip gracefully if CI is already running.

### DIFF
--- a/.github/workflows/retest-pr.yml
+++ b/.github/workflows/retest-pr.yml
@@ -4,6 +4,9 @@ on:
   issue_comment:
     types: [created]
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 jobs:
   retest:
     if: github.event.issue.pull_request && contains(github.event.comment.body, 'retest')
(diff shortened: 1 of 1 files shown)


alvations:
Avoiding the nodejs 20 warning.

alvations:
@stevenbird apologies for the multiple comments and pull-request merges to fix the ci-test and test it on different kinds of PR. Now `[CI: retest]` should work like in Travis days but with GH actions. 

`/retest` and common trigger also works.

ekaf:
Hi @alvations, I don't understand the usefulness of the `/retest` workflow in this repository. Only maintainers and collaborators are able to trigger a retest by commenting `/retest`, but these same users can already re-run CI jobs easily by clicking the “Re-run jobs” button in the GitHub Actions UI. That's simpler, more direct, and allows to eventually just pick one single test to re-run. I use it sometimes, when CI fails on for ex. just one MacOS runner.

Since external contributors can’t use `/retest` and must rely on CI in their forks anyway, I’m not sure this comment-based workflow adds any real convenience for anyone.

alvations:
Ah sorry, I mistakenly disallow the trigger for all PRs. I’ll revert this. 

The original and ultimate purpose was to 
- all new commits triggers the tests automatically (sanity check)
- Reruns without new commits/code change can only be triggered by collaborators, members and owner


But I think the first purpose is being disabled when I enable the 2nd.
I’ll revert tomorrow. Apologies for the inconvenience.

the original use of 2nd is because when we’re in Travis there was a limit to how much the ci/cd runs iirc but not sure if the gh actions nowadays have the low limits.

PR #3539, opened 2026-03-20
Add bot/malicious/slop/vandalism PR check workflow

alvations:
This PR is created to auto-trigger screening github action whenever a new pull-request is created and on `[CI: verify]` comment. 

The workflow only posts a comment when something is flagged, e.g.  https://github.com/alvations/nltk/pull/87#issuecomment-4100773548. Clean PRs pass silently, e.g. https://github.com/alvations/nltk/pull/88/changes

Flagged PRs get labeled (bot-pr, needs-security-review, possible-ai-slop, vandalism) and a detailed comment explaining exactly what was found. It also requests a Copilot code review automatically.

It will check for the follow:
- Bot-like accounts and behavior
- Malicious code patterns
- AI slop and security theatre
- Vandalism and defacement

This is. one way we can at least lighten the load for maintainers and make sure even if developers uses coding agent, they shouldn't be penalized but at least the maintainers and reviewers should be aware.

Disclaimer: this PR is semi-generated, but I (human) am the one who designed the checks for the different parts of the co-pilot prompt.

### DIFF
--- a/.github/workflows/bot-pr-check.yml
+++ b/.github/workflows/bot-pr-check.yml
@@ -1,0 +1,386 @@
+name: Bot PR Check
+
+on:
+  pull_request_target:
+    types: [opened, reopened]
+  issue_comment:
+    types: [created]
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+jobs:
+  check-bot:
+    if: github.event_name == 'pull_request_target' || (github.event.issue.pull_request && contains(github.event.comment.body, 'verify'))
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Check trigger and authorization
+        if: github.event_name == 'issue_comment'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const assoc = context.payload.comment.author_association;
+            if (!['OWNER', 'MEMBER', 'COLLABORATOR'].includes(assoc)) {
+              core.setFailed('Not authorized (association: ' + assoc + ')');
+              return;
+            }
+            if (!context.payload.comment.body.includes('[CI: verify]') &&
+                !context.payload.comment.body.includes('[ci: verify]')) {
+              core.setFailed('No recognized trigger found.');
+            }
+
+      - name: Analyze PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const prNumber = context.payload.pull_request
+              ? context.payload.pull_request.number
+              : context.issue.number;
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber,
+            });
+            const login = pr.user.login;
+            const userType = pr.user.type;
+
+            // Request Copilot code review
+            try {
+              await github.rest.pulls.requestReviewers({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                pull_number: pr.number,
+                reviewers: ['copilot'],
+              });
+              core.info('Copilot review requested.');
+            } catch (e) {
+              core.warning('Could not request Copilot review: ' + e.message);
+            }
+
+            // Fetch the diff for analysis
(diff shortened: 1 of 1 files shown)


ekaf:
I find the proposed lists of suspiciousPatterns and slopPhrases interesting, because they remind of SpamAssassin's email tests, and as such could form the basis for a much more comprehensive slop-filtering library, outside NLTK.
 
It would be useful to know where these filtering patterns come from, and if their performance  on slop-detection tasks has been evaluated. My initial impression is that they seem very likely to generate both high false-positive and false-negative rates. So I would not recommend trying to enforce any anti-slop policy before more research is available on this topic.

This PR just comments on the flagged phrases, which seems like it could be marginally helpful.

alvations:
I also think it might trigger too many false negative and high positive but I dont exactly know how to test it out unless we put it in place first but that also requires PRs submitters to declare if their commits are indeed created/augmented by AI too after it triggers.

Maybe a proper experiment is warrant to ake as many github PRs as possible plot these phrases across time and track the frequency and then use that as a proxy to eliminate or add to the stop phrases. Looking and spam filter is a good idea, I'll see how to augment or what the literature already say.

The phrases and most of the details came from a mix of me asking claude to add a filtering feature based on my design,  i.e. prompting it, "Add a filter to detect AI slop", " and then me thinking, I think that's possible (so keep the code) or I dont think it's right "Refine the filter to make the threshold higher". 

Or maybe in CONTRIBUTING.md we should be clear that augment or pure AI code should be identified in the commit messages or in the PR somehow. But what I'm observing is that people are just changing the commit message to mask the fact that the code is AI and then just auto-bot the PR creation through gh CLI. 

I'm not exactly sure how to proceed with this PR though but I suggest just monitoring to see how much more of these automated code PRs we get that are not helpful.

stevenbird:
I'm happy for us to experiment with this.

ekaf:
This new field of study is likely to become as important as spam-filtering, because the stakes are high for big companies that base their business on search engine results, or electronic book publishing, which both face a flood of low-quality AI-generated content.

At this early stage, we still miss a precise definition of AI-slop, but especially in the context of nltk PRs, we need to ask "what is it exactly that we don't want?".
For ex. [the edit cited in this comment](https://github.com/nltk/nltk/pull/3528#discussion_r2963813057)  does not trigger any particular slop pattern: what makes it undesirable is that it does not differ sufficiently from the replaced sentence. This suggests a need to analyse the diff, in addition to the message itself. 

Experimenting with slop detection is a positive shift from "maintainance mode". I am open to it, and curious to see the results, as long as it only raises harmless warnings.
Running the detection in CI can attract PRs from bots that want to test this slop-filter. I'm not saying it is a good thing, but it's an opportunity to collect test material. Initially, a retrospective study on a corpus of recent PRs would prove more informative, by showing the limits of what we can expect from the proposed approach.

alvations:
I think this is getting worse by day. Now we're not just getting bots spamming the PRs, even comments that this pre-commit hook aren't checking. I dont think I've much time to support any more extra work on this but let me think if there's other ways to combat these issue. 

This should be on @github end to handle this like how our mail servers handle spams, not on the user end to write filters like these.

ekaf:
Github may have difficulties: it has shown several occurrences of weird behaviour since yesterday.

stevenbird:
These PRs may be unusual in arriving without a link to an issue. Could we use this to set up a hurdle.

Eg PR only permitted from approved users. You get approval by submitting an issue report that is vetted.

But yes, it's a GitHub-wide issue and hopefully they'll figure something out. What a waste of our scarce resources.

stevenbird:
@alvations I'm happy for us to test this out, if you wish?

alvations:
@stevenbird I think the situation changed quite a lot since the last few months. It's now a lot harder to detect (i) bot-assisted valid PR vs (ii) bot just not-so-useful PR vs (iii) malicious PR. 

I think the current PR's solution won't be sufficient, so I won't recommend this until some sort of newer generation on model or detection algorithm is created to really scrutinize valid, useless or malicious PR. Not sure how other open source are coping with these too, any one has suggestions/ideas/pointers?

PR #3541, opened 2026-03-23
fix: use defusedxml to prevent XML entity expansion DoS in downloader

scruge1:
## Summary
Replace `ElementTree.parse()` with `defusedxml.ElementTree.parse()` in `_update_index()` to prevent Billion Laughs (XML bomb) attacks via crafted `server_index_url` payloads.

## Problem
`nltk/downloader.py:980` fetches an XML index from a configurable URL and parses it with `xml.etree.ElementTree.parse()` without entity expansion limits. A Billion Laughs payload causes exponential memory allocation (~3GB from 1KB), crashing the process.

## Fix
- Uses `defusedxml.ElementTree.parse` which blocks entity expansion
- Falls back to stdlib `ElementTree` if `defusedxml` is not installed
- One-line change at the vulnerable call site

## Testing
Verified on Python 3.10.11 — entity expansion confirmed without fix, blocked with fix.

Related Huntr report submitted.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -171,7 +171,6 @@ import warnings
 import zipfile
 from hashlib import md5, sha256
 from urllib.error import HTTPError, URLError
-from defusedxml.ElementTree import parse as safe_parse
 from xml.etree import ElementTree
 
 import nltk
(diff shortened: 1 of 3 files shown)


scruge1:
Updated with:
- Added `defusedxml` to `install_requires` in setup.py (previously was an optional import with silent fallback to stdlib, making the fix inactive for most users)
- Added test `test_downloader_xxe.py` proving Billion Laughs XML bomb is rejected
- Removed the try/except fallback so the protection is always active

stevenbird:
Please consult our guidelines before submitting a PR.
https://github.com/nltk/nltk/blob/develop/CONTRIBUTING.md

PR #3543, opened 2026-03-27
Let TextCat.guess_language return None for ties

### LINKED ISSUE #3540: TextCat seems to think Chinese text is Abkhazian
I've got some text I was dealing with ("【补货啦！麦丽莎进口意大利面！超值500g3袋】同品牌超市一包就要129元！多种组合任选！低脂低卡！这个当减脂餐简直完美越吃越过瘾，让你根本停不下来，不是我吹！比餐厅做的好吃的意大利面！" - yes I know it's an advert), and Google translate thinks it's Chinese and has something resembling a coherent output so I think it's probably right.

Feeding this into
```python
    tc = TextCat()
    guess_language = tc.guess_language(text_string) 
```
though gets `abk` out i.e. Abkhazian which feels odd given I don't think it even has many characters in line with that text! I may well be wrong, as today is the first time I heard of Abkhazian though.

This is with nltk 3.9.3.

ekaf:
Fix #3540: in the reported  example, the text sample is equally distant from all available language signatures, so just using min() returns the first language name in alphabetical order (Abkhazian).

The paper by Cavnar and Trenkle only considers the case where the least distant language is unique. But a practical implementation needs to also handle cases where the closest language is not unique, and for ex. return NA, as the R textcat package does.

This PR implements the same condition as the R _textcat_ package: it counts the number of languages that match the minimum out-of-place value, and returns _None_ if this number does not equal _1_.

```
from nltk.classify import TextCat as tc
text_string="【补货啦！麦丽莎进口意大利面！超值500g3袋】同品牌超市一包就要129元！多种组合任选！低脂低卡！这个当减脂餐简直完美越吃越过瘾，让你根本停不下来，不是我吹！比餐厅做的好吃的意大利面！"
print(tc().guess_language(text_string))

```

> None

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,7 +13,7 @@ Version 3.9.4 2026-03-24
 
 Thanks to the following contributors to 3.9.4:
 Min-Yen Kan, Eric Kafe, Emily Voss, bowiechen, Hrudhai01,
-jancallewaert, Mr-Neutr0n, pollak.peter89, ylwango613, 
+jancallewaert, Mr-Neutr0n, pollak.peter89, ylwango613,
 
 Version 3.9.3 2026-02-21
 
--- a/nltk/classify/textcat.py
+++ b/nltk/classify/textcat.py
@@ -136,8 +136,19 @@ class TextCat:
         to the text and return its ISO 639-3 code"""
         self.last_distances = self.lang_dists(text)
 
-        return min(self.last_distances, key=self.last_distances.get)
-        #################################################')
+        if (
+            len(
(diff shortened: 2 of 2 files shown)


ekaf:
This update introduces a return_all parameter to TextCat.guess_language, allowing users to retrieve all candidate languages in the case of a tie. It also explicitly handles the special case where all languages are equally improbable by returning an empty list (for return_all=True) or None (for return_all=False).

stevenbird:
Excellent work!

PR #3547, opened 2026-04-03
Fix FunctionalCategory.substitute() ignoring substituted direction

### LINKED ISSUE #3550: FunctionalCategory.substitute() ignores substituted direction
I was reading through the CCG module and noticed something odd in `FunctionalCategory.substitute()`. The method carefully computes substituted versions of all three components (result, direction, and argument), but then when it constructs the return value, it passes the original `self._dir` instead of the substituted `sub_dir`:

```python
def substitute(self, subs):
    sub_res = self._res.substitute(subs)
    sub_dir = self._dir.substitute(subs)
    sub_arg = self._arg.substitute(subs)
    return FunctionalCategory(sub_res, sub_arg, self._dir)  # should be sub_dir
```

This means that any substitution targeting the direction component is silently dropped. You can see this pretty clearly with a small example:

```python
from nltk.ccg.api import Direction, FunctionalCategory, PrimitiveCategory

res = PrimitiveCategory("S")
arg = PrimitiveCategory("NP")
variable_dir = Direction("/", "_")  # a variable direction

category = FunctionalCategory(res, arg, variable_dir)

subs = [("_", ["."])]
result = category.substitute(subs)

# We'd expect the direction to now have concrete restrictions ["."],
# but it still has the variable marker "_"
print(result.dir().is_variable())  # True, but should be False
print(result.dir().restrs())       # '_', but should be ['.']
```

Looking at `git blame`, this has been there since the CCG module was first moved into NLTK back in 2009 (commit cce09b03f). It looks like a straightforward copy-paste oversight where `self._dir` should have been `sub_dir`, consistent with how `sub_res` and `sub_arg` are already used on that same line.

I have a fix with a regression test in PR #3547.

elias-ba:
Fixes #3550

## What's going on

`FunctionalCategory.substitute()` computes substituted versions of all three components (result, direction, argument), but when building the return value it accidentally passes the original `self._dir` instead of the substituted `sub_dir`. This means substitutions targeting the direction are silently dropped. The bug has been there since the CCG module was first added in 2009.

## The fix

One-line change on line 331: `self._dir` → `sub_dir`.

## Before and after

```python
from nltk.ccg.api import Direction, FunctionalCategory, PrimitiveCategory

res = PrimitiveCategory("S")
arg = PrimitiveCategory("NP")
variable_dir = Direction("/", "_")

category = FunctionalCategory(res, arg, variable_dir)
result = category.substitute([("_", ["."])])

# Before: result.dir().is_variable() == True,  result.dir().restrs() == '_'
# After:  result.dir().is_variable() == False, result.dir().restrs() == ['.']
```

## Tests

Added `test_ccg.py` with a regression test that fails on the old code and passes with this fix. Verified by temporarily reverting the change and confirming the test catches it.

### DIFF
--- a/nltk/ccg/api.py
+++ b/nltk/ccg/api.py
@@ -328,7 +328,7 @@ class FunctionalCategory(AbstractCCGCategory):
         sub_res = self._res.substitute(subs)
         sub_dir = self._dir.substitute(subs)
         sub_arg = self._arg.substitute(subs)
-        return FunctionalCategory(sub_res, sub_arg, self._dir)
+        return FunctionalCategory(sub_res, sub_arg, sub_dir)
 
     # A function can unify with another function, so long as its
     # constituents can unify, or with an unrestricted variable.
--- a/nltk/test/unit/test_ccg.py
+++ b/nltk/test/unit/test_ccg.py
@@ -1,0 +1,46 @@
+"""
+Tests for the CCG (Combinatory Categorial Grammar) module.
+"""
+
+import unittest
+
+from nltk.ccg.api import Direction, FunctionalCategory, PrimitiveCategory
+
+
+class TestFunctionalCategorySubstitute(unittest.TestCase):
+    """Test that FunctionalCategory.substitute() applies substitutions
+    to all three components: result, argument, and direction."""
+
+    def test_substitute_applies_to_direction(self):
+        """Direction substitutions must propagate through substitute().
+
+        Regression test: previously, substitute() computed the substituted
+        direction but returned the original direction instead.
+        """
+        res = PrimitiveCategory("S")
+        arg = PrimitiveCategory("NP")
+        # A variable direction (restrictions="_") that should be substituted
+        variable_dir = Direction("/", "_")
+
+        category = FunctionalCategory(res, arg, variable_dir)
+
+        # Substitute the variable direction with concrete restrictions
+        new_restrictions = ["."]
+        subs = [("_", new_restrictions)]
+        result = category.substitute(subs)
+
+        # The direction's restrictions should now be ["."], not "_"
+        self.assertFalse(result.dir().is_variable())
+        self.assertEqual(result.dir().restrs(), new_restrictions)
+
+    def test_substitute_no_op_on_concrete_direction(self):
+        """Substitution on a non-variable direction should be a no-op."""
+        res = PrimitiveCategory("S")
+        arg = PrimitiveCategory("NP")
+        concrete_dir = Direction("\\", ["."])
+
+        category = FunctionalCategory(res, arg, concrete_dir)
+
+        result = category.substitute([("_", ",")])
+
+        self.assertEqual(result.dir().restrs(), ["."])


ekaf:
Thanks @elias-ba. It is not possible to assess your fix,  before you show us a test that actually  exercises the _substitute_ method. 
So, you need to first open an issue that shows the problem, by including a test where the current substitute implementation returns a wrong value. Then, please show us how this PR returns the correct value on the same test.

elias-ba:
Thanks for the feedback @ekaf! I opened #3550 with a detailed reproduction showing the wrong value that `substitute()` currently returns, along with what the expected output should be.

I also pushed a regression test in `test_ccg.py` that exercises the substitute method directly. It constructs a `FunctionalCategory` with a variable direction, calls `substitute()`, and checks that the direction is correctly updated. I verified that the test fails on the current code and passes with the fix.

Let me know if you'd like me to adjust anything!

ekaf (approved):
Thanks @elias-ba!, this fix is obvious. LGTM.

ekaf:
The test would be more convincing if it showed the linguistic relevance of the fix when parsing a real sentence. Unfortunately, that's not possible just yet, because every small bug in the ccg module seems to expose bigger bugs. Here, we would expect to see a positive effect of the fix on the following example:

```
from nltk.ccg import chart, lexicon

lex = lexicon.fromstring(r"""
:- S, NP, N
john => NP {john}
walked => (S\NP) {\x.walk(x)}
quickly => ((S\_NP)/(S\_NP)) {\F x.quickly(F(x))}
""", True)


print(lex)

sentence = "john quickly walked".split()
parser = chart.CCGChartParser(lex, chart.DefaultRuleSet)
for tree in parser.parse(sentence):
    print(tree.label()[0])
    chart.printCCGDerivation(tree)
```


But the substitution bug is actually masked by two other architectural limitations in the CCG module,  that explain how a bug like this could sit unnoticed in the codebase since 2009.  In normal use, this bug is essentially invisible:

- The Lexicon Regex: The string parser (lexicon.fromstring) currently cannot read variable directions like \_. The APP_RE regex strictly looks for [.,]? as slash modifiers, meaning it outright crashes if it encounters an underscore. Because of this, users haven't been able to construct these categories from strings to even trigger the bug.

- Variable Extraction: During chart parsing, FunctionalCategory.match() currently only knows how to extract feature variables (like S[X]). It does not extract direction substitution mappings when comparing categories (e.g., S\_NP vs S\NP). Consequently, it passes an empty substitution dictionary to substitute(), making the fix a no-op during a live parse.

Because of these two blockers, users couldn't reach the bug, and unfortunately, the CCGChartParser still won't be able to handle polymorphic modifiers out-of-the-box just yet.

That said, this PR is a perfectly correct and necessary foundational fix! We should absolutely merge it. However, it may be difficult to convince reviewers, until users can finally utilize variable directions in their grammars.

elias-ba:
Thanks @ekaf, that makes a lot of sense. Good to know why this one stayed buried for so long.

Happy to look at both blockers as follow-ups. The `APP_RE` tweak to accept `_` looks pretty small. Teaching `match()` to emit direction substitutions is the more interesting one since that's what actually makes this fix do anything in a real parse.

I'll probably split them into separate PRs so they're easier to review, but let me know if you'd rather see them together.

elias-ba:
Just saw #3555, that's the exact breakdown. I'll coordinate the follow-up work there, starting with `APP_RE`.

ekaf:
Creating a file with a general name like _test_ccg.py_ is too broad, for an issue that exclusively concerns direction substitutions. Instead, it would be better to use a more targeted filename like _test_ccg_dir.py_ (as in #3556), and consolidate all testing to fit into that file, by adding tests  that can complement #3556 while avoiding redundancy.

stevenbird:
Thanks @elias-ba, great contribution! Thanks @ekaf for the review.

PR #3553, opened 2026-04-13
Fix _create_token_table skipping last paragraph break transition + regression test

13rac1:
When a word position exceeds the final paragraph break, `next(pb_iter)` raises `StopIteration` before `current_par` is incremented. This causes `par_count` to be undercounted for any word that spans the last paragraph boundary. Fix by incrementing `current_par` before advancing the iterator.

Note: This bug was found with AI assistance and has been reproduced in the regression test.


stevenbird:
Thanks @13rac1

Issue #3555, opened 2026-04-14
CCG: Pipeline bugs block variable directions from parsing


ekaf:
PR #3547 introduces a necessary, foundational fix to how CCG variable directions are substituted mathematically. However, polymorphic adverbs (e.g., `(S\_NP)/(S\_NP)`) remain completely inaccessible to the `CCGChartParser` because of three internal bugs in the string parsing and unification pipeline. 

Currently, attempting to define or parse a variable direction (like `\_`) fails immediately in standard use:

```
python
from nltk.ccg import lexicon

# 1. Lexer Failure (APP_RE regex rejects the underscore)
# Fails: Raises AttributeError during string parsing
lex = lexicon.fromstring("quickly => (S\_NP)/(S\_NP)")

# 2. Unification Failure (can_unify silently drops direction mappings)
# Fails: Returns a substitution list missing the '_' mapping entirely
subs = quickly_cat.res().can_unify(walked_cat)
```

**Root Causes & Proposed Fixes:**

Through testing, I identified the three exact breaks in the chain that cause this:

1. **`nltk.ccg.lexicon`**: The `APP_RE` regex does not include `_` in its character classes, causing the lexer to outright reject variable modalities.
2. **`nltk.ccg.api.Direction`**: The string parser feeds raw regex capture tuples (e.g., `('_', '')`) directly into `Direction.__init__`. Because it isn't a string, `is_variable()` evaluates to `False`. 
3. **`nltk.ccg.api.FunctionalCategory`**: The `can_unify` method successfully computes the direction substitution mapping (`sd`), but the return statement (`return sa + sb`) completely forgets to include it, dropping the variable before `substitute()` can use it.

elias-ba:
Thanks @ekaf, this is a much more complete picture than I had in mind — the `Direction.__init__` tuple bug is a nice catch, I hadn't traced that one.

Happy to take these on. My plan would be one PR per root cause, in this order:

1. **`APP_RE`** first. Smallest surface, and it's a prerequisite for writing end-to-end tests against the other two (without it, you can't construct a variable-direction grammar via `lexicon.fromstring`, so any test for fix 2 or 3 has to hand-build `FunctionalCategory` objects — less convincing).
2. **`Direction.__init__`** next. The tuple-vs-string fix is self-contained and doesn't depend on the others landing first.
3. **`can_unify`** last. Once the first two are in, this is the one that finally makes #3547's `substitute()` fix observable during a real parse — so it's the natural closing move.

[...]

PR #3558, opened 2026-04-18
Fix APP_RE to accept variable direction underscore

elias-ba:
Addresses the first of three bugs in #3555 (the `APP_RE` lexer failure).

## What's going on

`APP_RE` only accepted `.` and `,` as slash modifiers, so parsing any variable direction (e.g. `(S\_NP)/(S\_NP)`) crashed:

```
AttributeError: 'NoneType' object has no attribute 'groups'
```

The `_` fell through the modifier slots into the remainder, where `NEXTPRIM_RE` rejected it.

## The fix

Extend the first modifier character class to include `_`:

```python
APP_RE = re.compile(r"""([\\/])([.,_]?)([.,]?)(.*)""")
```

## Scope

Deliberately narrow. After this fix, `lexicon.fromstring(r"quickly => (S\_NP)/(S\_NP)")` succeeds and produces a `FunctionalCategory` with the `_` captured in the inner direction. It does **not** yet make `Direction.is_variable()` return `True` for those parsed directions — that requires fixing `Direction.__init__`'s handling of raw regex tuples, which is the second bug in #3555 and will come in a separate PR.

Keeping the PRs one-per-root-cause as discussed on #3555.

## Tests

Added `test_ccg.py` with a regression test that parses `(S\_NP)/(S\_NP)` via `lexicon.fromstring` and asserts the resulting structure. Verified it fails on the old regex with the exact `AttributeError` from the issue, and passes with the fix.

### DIFF
--- a/nltk/ccg/lexicon.py
+++ b/nltk/ccg/lexicon.py
@@ -26,8 +26,10 @@ PRIM_RE = re.compile(r"""([A-Za-z]+)(\[[A-Za-z,]+\])?""")
 # string
 NEXTPRIM_RE = re.compile(r"""([A-Za-z]+(?:\[[A-Za-z,]+\])?)(.*)""")
 
-# Separates the next application operator from the remainder
-APP_RE = re.compile(r"""([\\/])([.,]?)([.,]?)(.*)""")
+# Separates the next application operator from the remainder.
+# The modifier slot also accepts `_`, marking a variable direction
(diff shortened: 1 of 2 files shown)


ekaf:
Is this an improvement over #3556?

elias-ba:
Closing: duplicate of #3556. @ekaf, apologies for the noise - I should have checked the open PRs before starting.

PR #3559, opened 2026-04-18
Fix Direction.__init__ to normalize tuple restrictions

elias-ba:
Addresses the second of three bugs in #3555 (the `Direction.__init__` tuple handling).

## What's going on

`lexicon.py`'s `parseApplication` passes the two-slot `APP_RE` capture tuple (e.g. `('_', '')`) directly into `Direction` as the `restrictions` argument:

```python
def parseApplication(app):
    return Direction(app[0], app[1:])
```

`Direction.__init__` stores it verbatim. `is_variable()` then tests `self._restrs == "_"` — which is always False for a tuple — so every variable direction recovered from a string silently reads as non-variable. Equality and hashing also diverge between parsed Directions and hand-constructed ones, even when they represent the same thing.

## The fix

Collapse tuple restrictions to a string at construction time:

```python
def __init__(self, dir, restrictions):
    if isinstance(restrictions, tuple):
        restrictions = "".join(restrictions)
    ...
```

The string and list call paths (user construction, `substitute`, `__neg__`) are unaffected.

## Scope

This fix is self-contained, but its visible effect depends on #3558 (the `APP_RE` regex fix) landing first — without #3558, the parser can't produce a variable direction at all. The fix here is still correct independently: it pins a consistent `Direction` contract so that whenever a tuple does arrive, `is_variable()` and comparisons behave correctly.

## Tests

Added 4 unit tests covering:
- Tuple with `_` normalizes and reads as variable.
- Parsed and hand-constructed variable Directions compare and hash equal.
- Concrete modifier tuples (`('.', '')`) normalize without being misread as variable.
- Empty-tuple restrictions behave as no-restrictions.

All four tests fail on current `develop` and pass with this fix. Tests are intentionally independent of #3558 — they construct tuples directly rather than going through the parser.

### DIFF
--- a/nltk/ccg/api.py
+++ b/nltk/ccg/api.py
@@ -154,6 +154,12 @@ class Direction:
     """
 
     def __init__(self, dir, restrictions):
+        # The lexicon's APP_RE yields a 2-slot capture tuple (e.g. ("_", ""))
+        # for the modifier. Collapse it to a string so `is_variable()` — which
+        # tests `_restrs == "_"` — behaves the same for parsed and
+        # hand-constructed Directions.
+        if isinstance(restrictions, tuple):
+            restrictions = "".join(restrictions)
         self._dir = dir
         self._restrs = restrictions
         self._comparison_key = (dir, tuple(restrictions))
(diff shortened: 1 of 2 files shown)


elias-ba:
Closing: duplicate of #3556. @ekaf, apologies for the noise - I should have checked the open PRs before starting.

PR #3570, opened 2026-04-30
Fix WordNetObject comparisons for incompatible types

### LINKED ISSUE #1944: Wordnet: Comparison of  a _WordNetObject fails if other object is not a _WordNetObject.

ekaf:
Fix #1944, an issue still open after several years, although three PRs attempted to close it (#1951, #3514, and #3562). This PR implements the consensus that emerged through the previous discussions, making `_WordNetObject` comparisons follow normal Python comparison semantics.

### Changes
- remove the redundant `__ne__` implementation
- keep value-based `__eq__` for WordNet objects
- make `__eq__` and `__lt__` return `NotImplemented` for incompatible types

### Why
Previously, comparing a WordNet object with an incompatible type could raise `AttributeError` by accessing `other._name`. Returning `NotImplemented` lets Python handle these cases correctly:
- `==` / `!=` behave normally for incompatible types
- ordering comparisons such as `<` raise `TypeError` instead of failing internally

This also preserves the existing equality behavior relied on by current tests, where distinct WordNet objects with the same name compare equal.

### Tests
Added minimal unit tests covering:
- value-based equality for distinct `Synset` instances with the same name
- mixed-type equality returning unequal rather than crashing
- mixed-type ordering raising `TypeError`

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -211,12 +211,13 @@ class _WordNetObject:
         return hash(self._name)
 
     def __eq__(self, other):
+        if not isinstance(other, _WordNetObject):
+            return NotImplemented
         return self._name == other._name
 
-    def __ne__(self, other):
-        return self._name != other._name
-
     def __lt__(self, other):
+        if not isinstance(other, _WordNetObject):
+            return NotImplemented
         return self._name < other._name
 
 
--- a/nltk/test/unit/test_wordnet.py
+++ b/nltk/test/unit/test_wordnet.py
@@ -17,7 +17,12 @@ class WordnNetDemo(unittest.TestCase):
 
     def test_value_equality_for_distinct_instances(self):
         # Guards against deleting __eq__ and falling back to identity equality.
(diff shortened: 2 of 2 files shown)


ekaf:
This PR addresses an  earlier objection (https://github.com/nltk/nltk/pull/1951#issuecomment-364103810) by using isinstance(other, _WordNetObject) rather than strict type equality so that comparisons continue to work across WordNet object subclasses, while still returning NotImplemented for unrelated types (per Python’s rich comparison protocol).
Although equality is based solely on _name, in practice _name values are type-specific (e.g., synsets vs lemmas use different naming schemes), so cross-type collisions cannot occur. LGTM.

ekaf:
It is time to merge this for #1944, to prevent further attempts to reinvent the wheel (#3578, #3584) that ignore the fact that \_\_ne\_\_() is unnecessary.

It is the minimal correct fix: preserve value-based `__eq__`, return `NotImplemented` for incompatible types, and avoid adding an unnecessary `__ne__`.

The tests also cover the important behavior changes clearly.

PR #3574, opened 2026-05-11
fix: block shared address space in SSRF guard

optimystical:
## Summary

This tightens `pathsec.validate_network_url()` so the SSRF guard blocks resolved targets that are not globally routable, including shared address space such as `100.64.0.0/10`.

Python's `ipaddress` module does not classify shared address space as `is_private`, so the existing loopback/link-local/multicast/private checks allowed URLs such as `http://100.64.0.1/` when `pathsec.ENFORCE = True`. That range is not public Internet space and can represent internal carrier-grade NAT infrastructure, so it should be rejected by a network fetch helper intended to block internal/non-public targets.

## Changes

- Treat `not ip.is_global` as unsafe in `validate_network_url()`.
- Add a regression test for `http://100.64.0.1/internal`.

## Security classification

- CWE-918: Server-Side Request Forgery (SSRF)

## Test

- `.venv/bin/pytest nltk/test/unit/test_pathsec.py -q`

### DIFF
--- a/nltk/pathsec.py
+++ b/nltk/pathsec.py
@@ -221,7 +221,13 @@ def validate_network_url(url_input, context="NetworkIO"):
 
         for result in _resolve_hostname(parsed.hostname or ""):
             ip = ipaddress.ip_address(result[4][0])
-            if ip.is_loopback or ip.is_link_local or ip.is_multicast or ip.is_private:
+            if (
+                ip.is_loopback
+                or ip.is_link_local
+                or ip.is_multicast
+                or ip.is_private
+                or not ip.is_global
+            ):
                 msg = f"Security Violation [{context}]: SSRF attempt to restricted IP {ip}"
                 if ENFORCE:
                     raise PermissionError(msg)
--- a/nltk/test/unit/test_pathsec.py
+++ b/nltk/test/unit/test_pathsec.py
@@ -55,6 +55,12 @@ def test_ssrf_cloud_metadata_link_local():
         dl.index()
 
 
+def test_ssrf_shared_address_space():
+    dl = Downloader(server_index_url="http://100.64.0.1/internal")
+    with pytest.raises((ValueError, PermissionError)):
+        dl.index()
+
+
 def test_ssrf_ip_obfuscation():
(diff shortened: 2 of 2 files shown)


ekaf (changes requested):
Thanks @optimystical: the new `or not ip.is_global` condition is a solid catch-all that closes several gaps (unspecified, reserved, CGNAT) in one shot. One bypass remains unaddressed: **IPv4-mapped IPv6 addresses** (e.g. `::ffff:7f00:1`). Python reports these as `is_global == True`, so the entire check is silently skipped for them, yet they can resolve to loopback or private IPv4 ranges on dual-stack hosts.

```suggestion
            if (
                ip.is_loopback
                or ip.is_link_local
                or ip.is_multicast
                or ip.is_private
                or not ip.is_global
                or (
                    ip.ipv4_mapped is not None
                    and not ip.ipv4_mapped.is_global
                )
            ):
```

`ip.ipv4_mapped` is `None` for native IPv4/IPv6 addresses, so this adds no overhead in the common case. Reusing `not ip.ipv4_mapped.is_global` mirrors the same logic as the outer check, keeping the two in sync automatically.

theluckystrike:
@ekaf Thank you for the review! I've incorporated your `ip.ipv4_mapped` suggestion in #3576, which includes all changes from this PR plus the IPv4-mapped IPv6 defense-in-depth check.

The new condition in `validate_network_url()`:
```python
or (
    ip.ipv4_mapped is not None
    and not ip.ipv4_mapped.is_global
)
```

This mirrors the outer `not ip.is_global` logic against the mapped IPv4 address, catching bypass attempts like `::ffff:7f00:1` → `127.0.0.1` on dual-stack hosts. Zero overhead for native addresses since `ipv4_mapped` is `None` in those cases.

Also added `test_ssrf_ipv4_mapped_ipv6` with mocked DNS resolution to cover loopback, private, and link-local mapped addresses.

ekaf:
@optimystical, it looks like you have abandoned this PR?

ekaf:
Superseded by #3576.

PR #3577, opened 2026-05-18
fix(translate): guard ZeroDivisionError in corpus_chrf, sentence_ribes and corpus_ribes on empty inputs

faizzyhon:
## Summary

Fixes four `ZeroDivisionError` crashes in the machine translation metrics module when called with empty sequences.

### 1. `corpus_chrf([], [])` → `ZeroDivisionError`

```python
corpus_chrf([], [])
# ZeroDivisionError: division by zero  (num_sents == 0)
```

**Fix**: early-return `0.0` when `num_sents == 0`.

### 2. `corpus_chrf(refs, hyps, min_len=N, max_len=M)` where `N > M` → `ZeroDivisionError`

```python
corpus_chrf([ref], [hyp], min_len=4, max_len=3)
# ZeroDivisionError: division by zero  (num_ngram_sizes == 0)
```

**Fix**: early-return `0.0` when `num_ngram_sizes == 0`.

### 3. `sentence_ribes(references, [])` → `ZeroDivisionError`

```python
sentence_ribes([["hello", "world"]], [])
# ZeroDivisionError: division by zero  (len(hypothesis) == 0)
```

**Fix**: early-return `0.0` when `hypothesis` is empty.

### 4. `corpus_ribes([], [])` → `ZeroDivisionError`

```python
corpus_ribes([], [])
# ZeroDivisionError: float division by zero  (len(hypotheses) == 0)
```

**Fix**: early-return `0.0` when `hypotheses` is empty.

## Files Changed

| File | Change |
|---|---|
| `nltk/translate/chrf_score.py` | Guard `num_sents == 0` and `num_ngram_sizes == 0` in `corpus_chrf` |
| `nltk/translate/ribes_score.py` | Guard empty `hypothesis` in `sentence_ribes`; guard empty `hypotheses` in `corpus_ribes` |
| `nltk/translate/test_translate_empty_inputs.py` | 7 unit tests covering all 4 bugs + regression |

## Tests

7 tests — all pass:
- `test_corpus_chrf_empty_corpus_returns_zero`
- `test_corpus_chrf_min_len_greater_than_max_len_returns_zero`
- `test_corpus_chrf_identical_sentences_returns_one` (regression)
- `test_sentence_ribes_empty_hypothesis_returns_zero`
- `test_sentence_ribes_identical_returns_one` (regression)
- `test_corpus_ribes_empty_corpus_returns_zero`
- `test_corpus_ribes_identical_returns_one` (regression)

### DIFF
--- a/nltk/translate/chrf_score.py
+++ b/nltk/translate/chrf_score.py
@@ -218,4 +218,8 @@ def corpus_chrf(
     total_scores = [sum(fscores) for n, fscores in ngram_fscores.items()]
 
     # macro-average over n-gram orders and over all sentences
+    if num_sents == 0:
+        return 0.0
+    if num_ngram_sizes == 0:
+        return 0.0
     return (sum(total_scores) / num_ngram_sizes) / num_sents
--- a/nltk/translate/ribes_score.py
+++ b/nltk/translate/ribes_score.py
@@ -47,6 +47,8 @@ def sentence_ribes(references, hypothesis, alpha=0.25, beta=0.10):
     :rtype: float
     """
     best_ribes = -1.0
+    if not hypothesis:
+        return 0.0
     # Calculates RIBES for each reference and returns the best score.
     for reference in references:
         # Collects the *worder* from the ranked correlation alignments.
@@ -116,6 +118,8 @@ def corpus_ribes(list_of_references, hypotheses, alpha=0.25, beta=0.10):
     # Iterate through each hypothesis and their corresponding references.
     for references, hypothesis in zip(list_of_references, hypotheses):
         corpus_best_ribes += sentence_ribes(references, hypothesis, alpha, beta)
+    if not hypotheses:
+        return 0.0
     return corpus_best_ribes / len(hypotheses)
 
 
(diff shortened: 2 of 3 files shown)


CodingSelim (commented):
had a look at this and it holds up. I checked out develop and confirmed all three crash: `corpus_chrf([], [])`, `sentence_ribes([['a','b']], [])` and `corpus_ribes([], [])` all raise ZeroDivisionError, and with this branch they return 0.0 instead. the 7 new tests pass locally, and I like that you kept the "identical sentences still return 1.0" cases so the guards can't silently swallow real scores.

one thing worth changing before merge: the test file is at `nltk/translate/test_translate_empty_inputs.py`, inside the package itself. every other unit test in the repo lives under `nltk/test/unit/` (there's ~49 of them), and this is the only `test_*.py` sitting in a package dir, so it'd actually ship inside the installed wheel. moving it to `nltk/test/unit/test_translate_empty_inputs.py` would match the layout, the imports don't need to change.

couple of small nits, take them or leave them:
- the comment banners in the test file use unicode box-drawing chars (── etc), plain ascii is a bit safer in source.
- in `corpus_chrf`, returning 0.0 when `min_len > max_len` quietly treats a bad argument as an empty result. matching the empty-corpus behavior is reasonable, just flagging that a caller passing min_len=4,max_len=3 probably made a mistake and might prefer a ValueError. not blocking.

nice clean fix overall.

ekaf:
This seems abandoned. Maybe we'll need a superseding PR.

PR #3578, opened 2026-05-22
Fix _WordNetObject comparison with non-WordNet types (#1944)

### LINKED ISSUE #1944: Wordnet: Comparison of  a _WordNetObject fails if other object is not a _WordNetObject.

CHENNA-NAGESH:
Fixes #1944
Currently, comparing a Synset or Lemma to other types (such as strings or None) causes an AttributeError because the code expects the other object to have a `_name` attribute.
This PR updates the comparison methods (`__eq__`, `__ne__`, `__lt__`) in `_WordNetObject` to return `NotImplemented` if the other object is not a `_WordNetObject`. This lets Python fall back to standard comparison behavior (returning False for == and raising a TypeError for <).
Unit tests have been added to verify this behavior.

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -211,12 +211,18 @@ class _WordNetObject:
         return hash(self._name)
 
     def __eq__(self, other):
+        if not isinstance(other, _WordNetObject):
+            return NotImplemented
         return self._name == other._name
 
     def __ne__(self, other):
+        if not isinstance(other, _WordNetObject):
+            return NotImplemented
         return self._name != other._name
 
     def __lt__(self, other):
+        if not isinstance(other, _WordNetObject):
+            return NotImplemented
         return self._name < other._name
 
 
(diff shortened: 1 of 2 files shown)


ekaf:
How does this PR compare with the solution in #3570?

CHENNA-NAGESH:
Hi @ekaf,
This PR is virtually identical in its core logic to your solution in #3570, as both add the `isinstance(other, _WordNetObject)` check and return `NotImplemented` to safely delegate comparison.
The minor differences are:
1. **`__ne__` handling**: Your PR #3570 is cleaner because it completely removes the redundant `__ne__` method (which Python 3 automatically derives from `__eq__`).
2. **Testing**: Both PRs add assertions checking that incompatible types return `False`/`True` and raise `TypeError` on ordered comparisons. Your PR additionally adds a guard against identity fallback (`test_value_equality_for_distinct_instances`).
Since you already have PR #3570 open, I am happy to close this pull request in favor of yours. Let me know what you prefer!

ekaf:
Indeed, closing this PR sounds wise. It would have been even better not to open it.

PR #3480, opened 2025-12-08
Harden FileSystemPathPointer to prevent path traversal & arbitrary file reads (CWE-22)

HyperPS:
This PR mitigates a high-impact path traversal & arbitrary file read vulnerability caused by unrestricted filesystem access when using FileSystemPathPointer with corpus readers such as PlaintextCorpusReader, TaggedCorpusReader, etc.

PR #3579, opened 2026-05-23
fix(security): block path traversal in NKJPCorpusReader

LinZiyuu:
## Summary
`NKJPCorpusReader`'s public read methods (`header`, `raw`, `words`, `sents`,
`tagged_words`) build file paths from the caller-supplied `fileids` and read
them with the **builtin `open()`**, bypassing the `CorpusReader.open()` /
`nltk.pathsec` sandbox — including the strict `ENFORCE = True` mode documented
in `SECURITY.md`. A `..` sequence or an absolute path in `fileids` escapes the
corpus root, allowing arbitrary file reads.

`NKJPCorpusReader` never goes through `nltk.data.find()` or
`CorpusReader.open()`; it concatenates the path itself in `add_root()` and reads
it with the builtin `open()`. As a result it bypasses the corpus-reader
path-traversal hardening added in #3479 (`CorpusReader.open()` CWE-22 guard),
#3480 (`FileSystemPathPointer` containment), #3522 (the `nltk.pathsec` central
sentinel) and #3528 (symlink containment) — none of those protect this reader.

## Details
In `nltk/corpus/reader/nkjp.py`:
- `add_root()` builds the path by plain string concatenation with no
  containment check (`return self.root + fileid`).
- The views then read it with the builtin `open()` (`XML_Tool` →
  `os.path.join(root, filename)` + `open()`; `NKJPCorpus_Header_View` →
  `filename + "header.xml"` opened via `StreamBackedCorpusView`).

Because the builtin `open()` is used (not `PathPointer.open()` /
`CorpusReader.open()`), `nltk.pathsec.validate_path()` is never invoked, so
`ENFORCE = True` does not block the access.

PoC (before this change), with `nltk.pathsec.ENFORCE = True`:
```python
from nltk.corpus.reader.nkjp import NKJPCorpusReader
r = NKJPCorpusReader(root="/path/to/corpora/nkjp/", fileids="sample")
r.header(fileids=["/path/to/corpora/nkjp/../../../../../../etc/"])
# -> opens /etc/header.xml via builtin open() and returns its parsed content
```

## Fix
Harden `add_root()` to reject any fileid whose resolved path leaves the corpus
root, matching the containment guarantee already provided by
`CorpusReader.open()`. The legitimate in-root flow (default `fileids` and
absolute fileids that stay within the corpus) is unchanged.

## Tests
Adds `nltk/test/unit/test_nkjp_security.py`:
- legitimate `header()` (default in-root fileids) still works;
- `header()` and `words()` raise `ValueError` on a `../` traversal fileid.

### DIFF
--- a/nltk/corpus/reader/nkjp.py
+++ b/nltk/corpus/reader/nkjp.py
@@ -95,7 +95,8 @@ class NKJPCorpusReader(XMLCorpusReader):
 
     def add_root(self, fileid):
         """
-        Add root if necessary to specified fileid.
+        Add root if necessary to specified fileid, and verify the resulting
+        path stays inside the corpus root.
 
         Security (CWE-22): the NKJP views build file paths from the
         caller-supplied ``fileids`` and read them with the builtin
@@ -104,13 +105,27 @@ class NKJPCorpusReader(XMLCorpusReader):
         root, otherwise a ``..`` sequence or an absolute path in ``fileids``
         allows reading arbitrary files outside the corpus.
         """
-        if self.root in fileid:
+        # ``str(self.root)`` is the absolute corpus path (``self._root._path``).
+        # We rebuild the path from it rather than from ``self.root`` used as a
+        # raw string, because the latter keeps the *original* (un-normalised)
+        # constructor argument: on Windows its separators differ from the ones
+        # ``os.path.join`` produces, which broke the old substring/concatenation
+        # logic (it duplicated the root, yielding an invalid path).
+        root = os.path.abspath(str(self.root))
+        fileid = str(fileid)
+        if os.path.isabs(fileid):
             result = fileid
         else:
-            result = self.root + fileid
-        root = os.path.normpath(str(self.root))
-        resolved = os.path.normpath(str(result))
-        if resolved != root and not resolved.startswith(root + os.sep):
+            result = os.path.join(root, fileid)
+        resolved = os.path.abspath(result)
+        try:
+            contained = resolved == root or os.path.commonpath([root, resolved]) == root
+        except ValueError:
+            # Raised when the paths live on different drives (Windows) or mix
+            # absolute and relative parts -- in every such case the fileid is
+            # outside the corpus root.
+            contained = False
+        if not contained:
             raise ValueError(
                 f"NKJP fileid escapes the corpus root (path traversal blocked): {fileid!r}"
             )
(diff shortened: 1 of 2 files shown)


ekaf:
The likely incompatibility is the new PR logic in nltk/corpus/reader/nkjp.py, specifically the startswith(root + os.sep) check. Replace it with an abspath + commonpath containment check and rewrite the tests to be OS-portable. That is the most plausible fix for “fails on all supported Python versions on windows-latest, passes on Ubuntu/macOS.”

LinZiyuu:
Thanks @ekaf.
you were right that it's a path-separator portability issue in `add_root()`. Fixed in 82b382a.

ekaf:
This looks close, but I don’t think it’s ready to approve yet because the current containment check is based on os.path.abspath(...), which does not resolve symlinks. Since the code path still uses builtin open(), the symlink-escape concern raised in review seems substantive. Please consider switching to a resolved-path / nltk.pathsec-based containment check.

LinZiyuu:
Hi @ekaf, 
Thanks, that's a fair catch — you're right that os.path.abspath() only normalizes .. lexically and won't resolve a symlink, so an in-root symlink pointing outside the corpus would have slipped past the commonpath check and then been followed by the builtin open().

I've switched add_root() to the same symlink-aware containment guard CorpusReader.open() already uses (nltk/pathsec.py::validate_path(..., required_root=self.root), from #3528). It resolves both the candidate path and the root via Path(...).resolve() before comparing and raises ValueError if the resolved path leaves the corpus root, so this now goes through the central nltk.pathsec sentinel rather than a local lexical check.

I also added a regression test (test_nkjp_header_rejects_inroot_symlink_escape) that creates a symlink inside the corpus root resolving to a sibling directory — the exact case the old abspath-based check couldn't catch. The legitimate in-root flow is unchanged, and all four tests in test_nkjp_security.py pass.

ekaf (approved):
Thanks @LinZiyuu, this looks good now. The earlier symlink-escape concern has been addressed by routing containment through nltk.pathsec.validate_path(...), and the new regression coverage includes the in-root symlink escape case. I’m comfortable approving this NKJP fix. LGTM.

ekaf:
#3579 and #3581 are targeted corpus-reader traversal fixes. They look independent and low-conflict, so they’re good to land after the shared path machinery is in.

PR #3581, opened 2026-05-24
fix(security): block path traversal in FramenetCorpusReader.frame()

LinZiyuu:
## Summary
`FramenetCorpusReader.frame(name)` (→ `frame_by_name()`) interpolates the
caller-supplied frame **name** into an XML file path that is then read with the
**builtin `open()`** (via `XMLCorpusView`), bypassing the `CorpusReader.open()` /
`nltk.pathsec` sandbox — including the strict `ENFORCE = True` mode documented in
`SECURITY.md`. A `..` sequence in the name escapes the corpus root, giving an
arbitrary file read whose content is returned to the caller.

This is the same class as the `CorpusReader` path-traversal hardening already
landed in #3479 / #3480 / #3522 / #3528: `frame()` never goes through
`CorpusReader.open()`, so none of that protection applies.

## Details
`nltk/corpus/reader/framenet.py`, `frame_by_name()`:

```python
locpath = os.path.join(f"{self._root}", self._frame_dir, fn_fname + ".xml")
with XMLCorpusView(locpath, "frame") as view:   # string path -> builtin open()
    elt = view[0]
```

`fn_fname` comes straight from the public API `frame(fn_fid_or_fname)` (a string
is dispatched to `frame_by_name`) with no containment check. Because
`XMLCorpusView` is constructed with a *string* path, it reads the file with the
builtin `open()` (`XMLCorpusView._detect_encoding` / `StreamBackedCorpusView._open`),
so `nltk.pathsec.validate_path()` is never invoked.

## PoC (before this change), with `nltk.pathsec.ENFORCE = True`
```python
# attacker-controlled frame name escapes the corpus root:
f = fn.frame("../../../../../../some/dir/secret")   # reads <root>/frame/../../.../secret.xml
print(f["definition"])                              # -> contents of the out-of-root file
```
Verified end-to-end in strict mode: NLTK opens a file outside the corpus root
via the builtin `open()` (no `PermissionError`, no warning) and `frame()` returns
its `<definition>` to the caller.

Constraint: a fixed `.xml` basename is appended, so the attacker chooses the
directory freely; full content disclosure requires the target to be frame-shaped
XML.

## Fix
Reject frame names that contain path separators or parent-directory references
before building the path, consistent with the validation already used elsewhere
in NLTK (e.g. the downloader's package-id check). Also harden `doc()` against a
malicious corpus index whose `filename` field contains traversal sequences.

## Tests
Adds `nltk/test/unit/test_framenet_security.py`:
- a `../` traversal frame name is rejected and never reaches the filesystem;
- a normal frame name still passes validation (fails only as "Unknown frame").

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -1362,6 +1362,12 @@ warnings(True) to display corpus consistency warnings when loading data
         except KeyError as e:  # probably means that fn_docid was not in the index
             raise FramenetError(f"Unknown document id: {fn_docid}") from e
 
+        # Security (CWE-22): defend against a malicious corpus index whose
+        # filename field contains path-traversal sequences (the file is read with
+        # the builtin open() via XMLCorpusView, bypassing nltk.pathsec).
+        if os.sep in xmlfname or "/" in xmlfname or "\\" in xmlfname or ".." in xmlfname:
+            raise FramenetError(f"Invalid document filename: {xmlfname!r}")
+
         # construct the path name for the xml file containing the document info
         locpath = os.path.join(f"{self._root}", self._fulltext_dir, xmlfname)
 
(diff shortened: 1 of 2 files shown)


ekaf:
@LinZiyuu, what do you think about the requested changes?

LinZiyuu:
Thanks @ekaf, addressed all of Copilot's points in 68ce86019.

ekaf (approved):
Security fix looks good now. The PR addresses the earlier review points: Windows drive/UNC handling is covered, file access is routed through self.abspath(...) / the secured path layer, and the missing regression coverage for doc() was added. I’m comfortable approving this for its Framenet scope. LGTM.

 As a follow-up, nltk/corpus/reader/ipipan.py looks worth hardening in a separate PR as well, since it has similar path-manipulation / builtin-open() patterns.

PR #3576, opened 2026-05-18
fix: block IPv4-mapped IPv6 SSRF bypass in pathsec

theluckystrike:
## Summary

Supersedes #3574 with the additional IPv4-mapped IPv6 fix requested by @ekaf in [their review](https://github.com/nltk/nltk/pull/3574#pullrequestreview-2855155283).

This PR includes all changes from #3574 (blocking shared address space / non-global IPs in the SSRF guard) **plus** the `ip.ipv4_mapped` defense-in-depth check:

```python
or (
    ip.ipv4_mapped is not None
    and not ip.ipv4_mapped.is_global
)
```

### Why this matters

IPv4-mapped IPv6 addresses like `::ffff:7f00:1` can report `is_global == True` in certain Python versions, silently bypassing the existing SSRF validation on dual-stack hosts. The `ipv4_mapped` attribute is `None` for native IPv4/IPv6 addresses, so this adds zero overhead in the common case.

### Changes

- **`nltk/pathsec.py`**: Added `ip.ipv4_mapped is not None and not ip.ipv4_mapped.is_global` to `validate_network_url()` condition
- **`nltk/test/unit/test_pathsec.py`**: Added `test_ssrf_ipv4_mapped_ipv6` covering loopback, private, and link-local mapped addresses with mocked DNS resolution

### Credit

Original PR #3574 by @optimystical. IPv4-mapped IPv6 bypass identified by @ekaf.

## Test plan

- [x] `test_ssrf_ipv4_mapped_ipv6` - validates mapped loopback (::ffff:127.0.0.1), private (::ffff:10.0.0.1), and link-local (::ffff:169.254.1.1) are blocked
- [x] Integration test with mocked DNS resolution returning IPv4-mapped IPv6 address
- [x] All existing pathsec tests remain unchanged

PR #3575, opened 2026-05-14
fix(security): block URL-encoded path traversal in nltk: resource URLs

LinZiyuu:
## Summary

`nltk.data.load()` validated the encoded form of a `nltk:` resource name
against `_UNSAFE_NO_PROTOCOL_RE`, but `url2pathname()` decodes percent
sequences when the path is later resolved to a filesystem location.
Inputs such as `nltk:%2fetc%2fpasswd` and
`nltk:corpora/..%2f..%2fetc%2fpasswd` therefore passed the textual safety
check and were turned into real absolute / traversal paths, allowing
arbitrary local file reads (e.g. `/etc/passwd`, `/proc/self/environ`).

The literal counterparts (`nltk:../../../etc/passwd`) were already
rejected via `ValueError("Unsafe resource path: …")`. This PR restores
that symmetric behaviour for URL-encoded inputs.

Reported via huntr: https://huntr.com/bounties/fae662d6-74c2-44fa-95f3-f53d4e8a8355

## Fix

* Decode the resource name with `urllib.parse.unquote` and re-apply the
  existing `_UNSAFE_NO_PROTOCOL_RE` / absolute-path / drive-letter
  checks in:
  * `_reject_unsafe_no_protocol()`
  * `normalize_resource_url()` (the `nltk:` branch)
  * `find()` (defense-in-depth, before `url2pathname()` runs)
* The literal-form checks are preserved, so behaviour is unchanged for
  well-formed resource names like `corpora/brown` or
  `tokenizers/punkt.zip/punkt/PY3/english.pickle`.

## Tests

Added parametrised regression tests in
`nltk/test/unit/test_data_security.py` covering:

* `%2f` / `%2F` encoded path separators
* `%2e%2e` / `%2E%2E` encoded `..` traversal
* mixed literal + encoded traversal (`corpora/..%2f..%2fetc%2fpasswd`)
* encoded `/proc/self/environ` and `/etc/passwd` targets
* encoded backslash traversal (`%5c..%5cetc%5cpasswd`)
* encoded Windows drive letter (`%43%3a%5cetc` → `C:\etc`)

Existing `nltk/test/unit/test_data{,_security}.py` and
`nltk/test/unit/test_pathsec.py` still pass (31 passed, 1 skipped).

## Test plan

- [x] `pytest nltk/test/unit/test_data.py nltk/test/unit/test_data_security.py nltk/test/unit/test_pathsec.py`
[...]

PR #3582, opened 2026-05-24
fix(security): prevent DNS-rebinding SSRF in pathsec.urlopen

LinZiyuu:
## Summary
`pathsec.urlopen` — used by `nltk.download` and `nltk.data.load` for all network
fetches — is vulnerable to **DNS-rebinding SSRF**. The SSRF filter validates the
hostname's resolved IP, but the actual connection re-resolves the hostname
independently, so the validated IP is never used for the connection. An attacker
who controls DNS for a hostname (TTL-0 rebinding) returns a public IP for the
validation lookup and an internal / loopback IP for the connection lookup,
bypassing the SSRF filter **even under `nltk.pathsec.ENFORCE = True`**.

## Details
`validate_network_url()` resolves via the `lru_cache`d `_resolve_hostname()` and
blocks loopback / private / link-local / multicast IPs. But `urlopen()` then
calls `opener.open(url)` with the raw hostname, and urllib's connection layer
(`socket.create_connection` → `getaddrinfo`) performs a **second, independent**
resolution at connect time. The `lru_cache` (commented "to mitigate DNS
rebinding") only memoizes the validation-side lookup, which the connection never
consults, so it provides no protection against rebinding.

## Proof of concept (before this change), with `nltk.pathsec.ENFORCE = True`
```python
import nltk.data, nltk.pathsec
nltk.pathsec.ENFORCE = True
# 'attacker.example' resolves to a public IP at validation time and to
# 127.0.0.1 at connect time (TTL-0 DNS rebinding controlled by the attacker):
data = nltk.data.load("http://attacker.example/latest/meta-data/", format="raw")
# -> returns the loopback / cloud-metadata response despite ENFORCE=True
```
A direct `http://127.0.0.1/...` is correctly blocked; the rebinding hostname is
not. The same applies to `nltk.download(server_index_url=<attacker>)` (the index
URL and every package `url` taken from the index are fetched through
`pathsec.urlopen`).

## Fix
Resolve the host once, SSRF-validate **every** resolved address, and connect to
that **pinned** numeric address (which performs no further name lookup), so
validation and connection observe the same resolution. Implemented as
`_SafeHTTPConnection` / `_SafeHTTPSConnection` (HTTPS keeps SNI and certificate
verification against the original hostname), wired into `urlopen` when no proxy
is configured. With a proxy, the proxy is the egress and resolves names itself,
so connection pinning is not applied (and the configured proxy is not blocked).
Redirect hops go through the same opener and are therefore pinned as well.

## Tests
`test_pathsec.py::test_ssrf_dns_rebinding_blocked_at_connect` simulates rebinding
(public IP at validation, loopback at connect) against a local server and
asserts the fetch is blocked under `ENFORCE=True` and the loopback body is not
returned. Existing SSRF / proxy-inheritance / redirect-validation tests still
pass.

### DIFF
--- a/nltk/pathsec.py
+++ b/nltk/pathsec.py
@@ -300,34 +300,53 @@ def _resolve_and_validate_host(host, port):
     return addrinfo
 
 
-def _pinned_address(host, port):
-    """Return the validated ``(ip, port)`` to connect to, pinning the resolution."""
+def _pinned_connection(host, port, timeout, source_address):
+    """Open a socket to ``host``/``port`` over an SSRF-validated, pinned address.
+
+    Every address ``host`` resolves to is validated together, then the socket is
+    opened to those **numeric** addresses, so no second (unvalidated) name lookup
+    happens at connect time -- this closes the DNS-rebinding TOCTOU.  All
+    validated addresses are tried in order, preserving urllib/socket's normal
+    dual-stack / multi-A fallback.  If nothing resolves to a validated address we
+    fail closed: we never fall back to connecting by the raw hostname, which
+    would re-resolve unvalidated and reopen the rebinding hole.
+    """
     addrinfo = _resolve_and_validate_host(host, port)
-    if addrinfo:
-        return addrinfo[0][4][0], port
-    return host, port
+    if not addrinfo:
+        raise OSError(
+            f"pathsec.urlopen: no validated address for host {host!r}; "
+            "refusing to connect by unvalidated hostname"
+        )
+    last_err = None
+    for res in addrinfo:
+        ip = res[4][0]
+        try:
+            return socket.create_connection((ip, port), timeout, source_address)
+        except OSError as e:
+            last_err = e
+    raise last_err
 
 
 class _SafeHTTPConnection(http.client.HTTPConnection):
     """HTTPConnection that resolves, SSRF-validates and pins the address at connect()."""
 
     def connect(self):
-        ip, port = _pinned_address(self.host, self.port)
-        self.sock = socket.create_connection(
-            (ip, port), self.timeout, self.source_address
+        self.sock = _pinned_connection(
+            self.host, self.port, self.timeout, self.source_address
         )
 
 
 class _SafeHTTPSConnection(http.client.HTTPSConnection):
     """HTTPS variant of :class:`_SafeHTTPConnection`.
 
-    Connects to the pinned IP but keeps SNI / certificate verification against
-    the original hostname.
+    Connects to a validated, pinned IP but keeps SNI / certificate verification
+    against the original hostname.
     """
 
     def connect(self):
-        ip, port = _pinned_address(self.host, self.port)
-        sock = socket.create_connection((ip, port), self.timeout, self.source_address)
+        sock = _pinned_connection(
+            self.host, self.port, self.timeout, self.source_address
+        )
         self.sock = self._context.wrap_socket(sock, server_hostname=self.host)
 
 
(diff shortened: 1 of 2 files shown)


ekaf:
@LinZiyuu, what do you think about the requested changes?

LinZiyuu:
Thanks @ekaf — addressed all three of Copilot's points in e586cfcae.

ekaf (approved):
Thanks @LinZiyuu, this looks good now. The updated implementation addresses the earlier concerns: it fails closed when no validated address is available, preserves fallback across validated addresses, and handles direct vs proxied connection paths appropriately. The added regression tests make the DNS-rebinding protection much more convincing. I’m comfortable approving this. LGTM.

ekaf:
This looks like the right primary fix for the DNS-rebinding issue, and I’d prefer to move forward with this PR rather than the older overlapping approach in #3576.

 One thing from #3576 still seems worth considering before merge: tightening the SSRF address policy from the current explicit checks (loopback, link-local, multicast, private) to a more general “non-global is forbidden” rule, plus explicit handling for IPv4-mapped IPv6 addresses. That seems like a worthwhile defense-in-depth improvement and would make #3582 a cleaner semantic superset of #3576.

 If you fold that remaining policy tightening into this PR, I think #3582 can cleanly supersede #3576.

LinZiyuu:
Hi @ekaf,
Done — folded the address-policy tightening from #3576 into this PR (7dfdca3). The SSRF check is now a single shared _ip_is_forbidden() used by both the validation path and the connect-time pinning path, so they can't drift. Policy is "allow only globally-routable; forbid non-global," which is a strict superset of the old loopback/link-local/multicast/private list and additionally covers CGNAT (100.64.0.0/10), unspecified (0.0.0.0/::), and reserved ranges. Multicast is still rejected explicitly since CPython reports it as is_global. IPv4-mapped IPv6 (::ffff:127.0.0.1) is unwrapped to its embedded IPv4 so the classification is version-independent. Added policy regression tests and verified the block/allow set is identical on 3.10 and 3.12. I think #3582 now cleanly supersedes #3576.

ekaf (approved):
This looks good to me now. The PR already addressed the earlier rebinding-specific concerns, and with the addition of the shared _ip_is_forbidden() policy it now also cleanly incorporates the useful address-classification tightening from #3576. I especially like that the same policy is now used in both the validation path and the connect-time pinned-resolution path, so they can’t drift. The added regression coverage for CGNAT, non-global addresses, and IPv4-mapped IPv6 makes the result much more convincing. Happy to approve this as the path forward.

ekaf:
Hi @LinZiyuu, rebasing #3588 revealed that we'll need one more patch. so I am going to submit a follow-up fix, before you should retest your PRs.

LinZiyuu:
Hi @ekaf,

Got it. I'll hold off and wait for your follow-up fix to land, then rebase all my security PRs (#3575, #3579, #3581, #3582, #3583) onto the updated develop and retest them under ENFORCE=True in one pass.

#3583 is already rebased and green under ENFORCE=True; I'll re-verify it after your patch as well.

Thanks for the heads-up!

ekaf:
PR #3595 has now solved the CI problem for #3588. I hope it works for you too.

ekaf:
#3582 is the most infrastructure-level security fix. It changes the network protection layer used by nltk.download and nltk.data.load, and it is conceptually separate from the corpus-reader path fixes. Merging it first reduces exposure quickly and isolates any regressions to the SSRF/pinning work.

PR #3585, opened 2026-05-29
Allow data.find() to return zipped corpora

Thynix:
This allows the following to work:

    rm -rf ~/nltk_data/corpora/wordnet ~/nltk_data/corpora/wordnet.zip  # and anywhere else on the path so it's gone
    python -m nltk.downloader wordnet
    python -c 'import nltk; nltk.data.find("corpora/wordnet")'

Without this change, it only works if it's manually unzipped.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -50,6 +50,7 @@
 - Alexis Dimitriadis
 - Nikhil Dinesh
 - Liang Dong
+- Steve Dougherty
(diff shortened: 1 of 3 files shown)


ekaf:
No evidence here that this is driven by an existing tracked complaint. NLTK's standard procedure is to start by filing an issue, and go on with a PR if it raises sufficient interest.

PR #3588, opened 2026-06-02
Fix remaining corpus reader pathsec gaps before ENFORCE=True RC

### LINKED ISSUE #3587: Harden remaining corpus/file path gaps before RC flips `pathsec.ENFORCE = True`
NLTK is expected to publish a **release candidate within the next two weeks** that will switch the centralized path security policy from:

```python
nltk.pathsec.ENFORCE = False
```

to:

```python
nltk.pathsec.ENFORCE = True
```

Before that RC is published, it would be good to patch the remaining path-handling gaps so the stricter policy is as complete and internally consistent as possible.

The main `CorpusReader.open()` symlink-escape issue appears fixed already, via lexical checks plus `validate_path(..., required_root=self._root)`. But there are still at least two weaker spots in `nltk/corpus/reader/util.py`:

### 1. `find_corpus_fileids()` can enumerate through symlinked directories

```python
elif isinstance(root, FileSystemPathPointer):
    items = []
    for dirname, subdirs, fileids in os.walk(root.path):
        prefix = "".join("%s/" % p for p in _path_from(root.path, dirname))
        items += [
            prefix + fileid
            for fileid in fileids
            if re.match(regexp, prefix + fileid)
        ]
```

Even if `CorpusReader.open()` later blocks resolved escapes, corpus discovery may still enumerate fileids reachable through symlinked directories under the corpus root. That creates inconsistency between what the corpus reader claims exists and what the stricter `open()` policy will actually allow.

**Suggested fix:** prune symlinked subdirectories from traversal, or resolve candidate directories and ensure they remain within the resolved root before descending.

### 2. `StreamBackedCorpusView._open()` still uses raw `open()` for string fileids

```python
def _open(self):
    if isinstance(self._fileid, PathPointer):
        self._stream = self._fileid.open(self._encoding)
    elif self._encoding:
        self._stream = SeekableUnicodeStreamReader(
            open(self._fileid, "rb"), self._encoding
        )
    else:
        self._stream = open(self._fileid, "rb")
```

When `self._fileid` is a plain string, this bypasses the centralized `pathsec` sentinel.

**Suggested fix:** replace direct `open()` calls here with `nltk.pathsec.open()` (or validate first, then open).

## Goal before RC
Before publishing the RC that flips `ENFORCE=True`, we should aim for:
- corpus file **discovery** to respect the same boundary assumptions as corpus file **opening**
- core corpus/file access paths to consistently go through `pathsec`
- regression tests covering both behaviors

This is not a claim that the original `CorpusReader.open()` symlink-escape bug is still open; it seems fixed. The point here is to close the remaining path-policy gaps before strict enforcement ships in the RC.

ekaf:
## Summary

Closes #3587.

This patch tightens two remaining path-handling gaps before the upcoming RC that is expected to switch `nltk.pathsec.ENFORCE` to `True`:

- `find_corpus_fileids()` no longer enumerates through symlinked directories that resolve outside the corpus root
- `StreamBackedCorpusView._open()` now uses the centralized `pathsec` wrapper for plain string fileids

## Tests

- add regression test ensuring symlink escapes are not enumerated by `find_corpus_fileids()`
- add regression test ensuring `StreamBackedCorpusView` with a plain string path is subject to path validation

### DIFF
--- a/nltk/corpus/reader/util.py
+++ b/nltk/corpus/reader/util.py
@@ -21,6 +21,7 @@ from nltk.data import (
     ZipFilePathPointer,
 )
 from nltk.internals import slice_bounds
+from nltk.pathsec import open as _secure_open
 from nltk.tokenize import wordpunct_tokenize
 from nltk.util import AbstractLazySequence, LazyConcatenation, LazySubsequence
 
@@ -209,10 +210,10 @@ class StreamBackedCorpusView(AbstractLazySequence):
             self._stream = self._fileid.open(self._encoding)
         elif self._encoding:
             self._stream = SeekableUnicodeStreamReader(
-                open(self._fileid, "rb"), self._encoding
+                _secure_open(self._fileid, "rb"), self._encoding
             )
         else:
-            self._stream = open(self._fileid, "rb")
+            self._stream = _secure_open(self._fileid, "rb")
 
     def close(self):
         """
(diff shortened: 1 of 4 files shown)


ekaf:
Note: CI also exposed a separate `TransitionParser` compatibility issue with newer `scikit-learn` resolutions. Tracked in #3589. The temporary CI pin in this PR is just to keep this branch green and is not intended as the final fix.

stevenbird:
Thanks @ekaf.

PR #3583, opened 2026-05-24
fix(security): prevent ReDoS in ReviewsCorpusReader FEATURES regex (CWE-1333)

LinZiyuu:
## Summary
`ReviewsCorpusReader` extracts `feature[+N]` annotations from each review line
with the `FEATURES` regex. The feature label was unbounded
(`(?:(?:\w+\s)+)?\w+`), so `re.findall` rescans a long, bracket-less line
**quadratically** — a crafted corpus line (e.g. `"word " * 100000`) hangs the
reader (ReDoS, CWE-1333). This is the same class as **CVE-2021-3828** (ReDoS in
`comparative_sents._read_comparison_block`).

## Details
```python
FEATURES = re.compile(r"((?:(?:\w+\s)+)?\w+)\[((?:\+|\-)\d)\]")  # reviews.py
```
For a line with a long run of words and no trailing `[+/-N]`, the label matches
to end-of-line at every search position, so `re.findall` is O(n²). It is reached
via the public API `nltk.corpus.product_reviews_1.reviews()` / `.features()` /
`.sents()` (and `ReviewsCorpusReader(root, fileids)`), with the (untrusted)
corpus file content as input.

## Proof of concept (before this change)
```python
from nltk.corpus.reader.reviews import ReviewsCorpusReader
# a review file whose single review line is "word " * 100000
ReviewsCorpusReader(root, "r.txt").reviews()[0]   # hangs (quadratic)
```
Measured: 32 KB line → 5 s, 64 KB → 21 s, 128 KB → > 25 s.

## Fix
Bound the feature label to a word plus up to 50 single-whitespace-separated
words (`\w+(?:\s\w+){0,50}`), making extraction linear. Real feature labels are
short noun phrases, so the bound does not change extraction on normal corpora —
verified identical `findall` output on representative inputs.

## Tests
`nltk/test/unit/test_reviews_security.py`:
- the regex still extracts the same features on normal lines;
- a crafted ~250 KB bracket-less line completes well within a 10 s timeout
  (both at the regex level and end-to-end via `reviews()`), where the previous
  regex takes ~50 s.

PR #3591, opened 2026-06-03
Fix TransitionParser sparse index dtype for scikit-learn 1.9

### LINKED ISSUE #3589: TransitionParser fails under scikit-learn 1.9.0 due to sparse index dtype incompatibility
Recent CI runs show that `nltk.parse.transitionparser` passed earlier today with older dependency resolutions such as:

- `scikit-learn==1.7.2`
- `scipy==1.15.3`
- `numpy==2.2.6`

but fails under newer CI resolutions such as:

- `scikit-learn==1.9.0`
- `scipy==1.17.1`
- `numpy==2.4.6`

The failure appears in both the `nltk.parse.transitionparser.demo` doctest and unit tests that call `TransitionParser.train()`.

## Error

```text
ValueError: Only sparse matrices with 32-bit integer indices are accepted. Got int64 indices.
```

This is raised by scikit-learn during `model.fit(x_train, y_train)`.

## Likely cause

`TransitionParser.train()` appears to build a sparse matrix whose index arrays (`indices` / `indptr`) are `int64`. Newer scikit-learn rejects this in code paths where only 32-bit sparse indices are accepted.

## Why this matters

NLTK CI currently allows dependency drift in `requirements-ci.txt`, so different Python versions or different run times may resolve different scikit-learn versions. This means CI can become unstable even when unrelated PRs make no change to `transitionparser`.

For the moment, pinning `scikit-learn==1.7.2` in CI avoids the failure, but that should be treated as a temporary workaround rather than the final fix.

## Suggested fix

Update `TransitionParser.train()` so the sparse training matrix passed to scikit-learn uses 32-bit index arrays before `model.fit(...)`.

## Acceptance criteria

- `nltk.parse.transitionparser.demo` passes under `scikit-learn==1.9.0`
- `test_transitionparser_warns_on_model_unpickle` passes under `scikit-learn==1.9.0`
- CI no longer depends on pinning `scikit-learn` to avoid this failure

ekaf:
Closes #3589

## Summary
Cast the sparse training matrix index arrays to `int32` in `TransitionParser.train()` before calling `model.fit(...)`.

## Problem
Recent CI failures under newer dependency resolution hit:

`ValueError: Only sparse matrices with 32-bit integer indices are accepted. Got int64 indices.`

This comes from newer scikit-learn rejecting sparse matrices with `int64` indices in this code path.

## Fix
After `load_svmlight_file(...)`, convert the CSR matrix index arrays:
- `x_train.indices -> int32`
- `x_train.indptr -> int32`

This keeps the patch minimal and fixes training without changing behavior otherwise.

## Testing
No new tests added. Existing CI coverage already exercises this path, including:
- `nltk.parse.transitionparser.demo`
- `test_transitionparser_warns_on_model_unpickle`

### DIFF
--- a/nltk/parse/transitionparser.py
+++ b/nltk/parse/transitionparser.py
@@ -523,6 +523,9 @@ class TransitionParser(ParserI):
             input_file.close()
             # Using the temporary file to train the libsvm classifier
             x_train, y_train = load_svmlight_file(input_file.name)
+            x_train = x_train.astype("float64")
+            x_train.indices = x_train.indices.astype("int32", copy=False)
+            x_train.indptr = x_train.indptr.astype("int32", copy=False)
             # The parameter is set according to the paper:
             # Algorithms for Deterministic Incremental Dependency Parsing by Joakim Nivre
             # Todo : because of probability = True => very slow due to


ekaf:
This has fixed CI, so let's expedite this PR, and remedy the recent CI failures in #3581, #3582, #3583 and #3588

PR #3596, opened 2026-06-06
chore: improve nltk maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in nltk/test/unit/test_hmm.py, nltk/test/unit/test_tag.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/nltk/test/unit/test_hmm.py
+++ b/nltk/test/unit/test_hmm.py
@@ -78,5 +78,15 @@ def test_backward_probability():
     assert_array_almost_equal(wikipedia_results, bp, 4)
 
 
+def test_hmm_tagger_instance_and_forward_dtype():
+    import numpy as np
+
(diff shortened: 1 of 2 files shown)


ekaf:
Thanks for the small maintenance cleanup. At the moment, though, these tests feel pretty low-value for NLTK.

- The new pos_tag return-type test appears largely redundant with the existing behavioral test coverage.
- The HMM test adds a little signal, but mostly around implementation/type details rather than meaningful user-facing behavior.

I don’t think we need this PR in its current form. If you’d like to keep working on it, I’d encourage adding tests that cover a more meaningful edge case, regression, or user-visible contract.

PR #3601, opened 2026-06-08
Avoid retaining bllip import exceptions

### LINKED ISSUE #3554: nltk.parse.bllip retains ImportError traceback via function default argument, leaking first-loader's call stack
## Summary

When `bllipparser` is not installed, `nltk/parse/bllip.py` permanently retains
the `ImportError` instance — and its full `__traceback__` — via a function
default argument. Because the retained traceback holds a reference to the
`<module>` frame of `bllip.py`, whose `f_back` chain extends back through
importlib to whichever caller first imported `nltk.parse.bllip`, **every local
variable on that first caller's call stack is leaked for the lifetime of the
process**.

This matters for long-running applications that lazily load nltk from inside
user code: the first load pins arbitrary user objects (models, connections,
large data structures, ...) whose frames happen to be on the stack at that
moment.

## Offending code

[`nltk/parse/bllip.py` lines 86–96](https://github.com/nltk/nltk/blob/develop/nltk/parse/bllip.py#L86-L96)
(unchanged in 3.9.4 release and `develop` branch head at time of filing):

```python
try:
    from bllipparser import RerankingParser
    from bllipparser.RerankingParser import get_unified_model_parameters

    def _ensure_bllip_import_or_error():
        pass

except ImportError as ie:

    def _ensure_bllip_import_or_error(ie=ie):
        raise ImportError("Couldn't import bllipparser module: %s" % ie)
```

The `ie=ie` parameter is a **default argument**. Default arguments are
evaluated at function-definition time and stored in
`function.__defaults__`, giving the captured value module lifetime. Every
subsequent attribute access on `nltk.parse.bllip._ensure_bllip_import_or_error`
can see the same `ImportError` object, together with its `__traceback__`.

Frame retention chain:

```
nltk.parse.bllip._ensure_bllip_import_or_error.__defaults__[0]
  └── ImportError
       └── .__traceback__                         (traceback object)
            └── .tb_frame                         (bllip.py <module>)
                 └── .f_back ⟶ importlib ⟶ ... ⟶ first caller's frame
                                                    └── .f_locals[<anything>]  ← leaked
```

## Minimal reproducer

Tested on `nltk==3.9.1`, CPython 3.11.5, macOS. No external dependencies beyond
nltk itself. `bllipparser` must **not** be installed.

```python
#!/usr/bin/env python3
"""
Minimal reproducer for nltk.parse.bllip ImportError retention leak.

Run in a fresh interpreter (so nltk.parse.bllip is not already cached):

    python nltk_bllip_leak_repro.py

Exit code:
    1  bug present
    0  bug fixed
    2  test environment is misconfigured
"""
import gc
import sys
import weakref


if "nltk.parse.bllip" in sys.modules:
    print("FATAL: nltk.parse.bllip already loaded", file=sys.stderr)
    sys.exit(2)

try:
    import bllipparser  # noqa: F401
    print("FATAL: bllipparser is installed; repro requires it MISSING", file=sys.stderr)
    sys.exit(2)
except ImportError:
    pass


class UserObject:
    def __repr__(self) -> str:
        return f"<UserObject id=0x{id(self):x}>"


def caller_that_first_loads_bllip():
    payload = UserObject()
    payload_ref = weakref.ref(payload)
    import nltk.parse.bllip  # noqa: F401
    return payload_ref


def main() -> int:
    payload_ref = caller_that_first_loads_bllip()
    for _ in range(5):
        gc.collect()

    import nltk
    import nltk.parse.bllip as bllip_mod

    fn = getattr(bllip_mod, "_ensure_bllip_import_or_error", None)
    defaults = getattr(fn, "__defaults__", None)

    print(f"nltk version   : {nltk.__version__}")
    print(f"python version : {sys.version.split()[0]}")

    if not defaults or not isinstance(defaults[0], BaseException):
        print("No ImportError retained — bug appears fixed.")
        return 0

    exc = defaults[0]
    print(f"\n_ensure_bllip_import_or_error.__defaults__[0] = {type(exc).__name__}({exc})")

    # Walk the f_back chain, locating the caller frame
    tb = exc.__traceback__
    caller_frame = None
    f = tb.tb_frame if tb else None
    depth = 0
    while f is not None:
        if f.f_code.co_name == "caller_that_first_loads_bllip":
            caller_frame = f
            break
        f = f.f_back
        depth += 1

    if caller_frame is not None:
        print(f"caller frame pinned at depth {depth} in tb_frame.f_back chain")
        print(f"caller f_locals keys: {list(caller_frame.f_locals.keys())}")

    if payload_ref() is not None:
        print(f"\nLEAK: {payload_ref()!r} still alive after 5x gc.collect()")
        return 1
    print("\nUserObject was GC'd — no leak.")
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

## Actual output (nltk 3.9.1, Python 3.11.5)

```
nltk version   : 3.9.1
python version : 3.11.5

_ensure_bllip_import_or_error.__defaults__[0] = ModuleNotFoundError(No module named 'bllipparser')
caller frame pinned at depth 36 in tb_frame.f_back chain
caller f_locals keys: ['payload', 'payload_ref', 'nltk']

LEAK: <UserObject id=0x10094ed10> still alive after 5x gc.collect()
```

Exit code: `1` (bug present).

The caller frame — 36 frames above `bllip.py <module>` in the `f_back` chain
— still holds `payload` in its `f_locals`, even though the function has
returned and five rounds of `gc.collect()` have run.

## Impact

In a long-running process, any application that lazily imports
`nltk.parse.bllip` (directly, or transitively via `nltk.parse`, or via
`nltk.chunk` which imports `nltk.parse.api` which pulls in parser modules)
will **permanently retain one copy of every local variable that was live on
the call stack at the moment of the first import**. The leak is silent,
proportional to how "rich" the first-loader call stack was, and invisible to
typical memory profilers that track allocation sites rather than frame
retention.

The retention is also hard to diagnose: the chain is obscured by 20+
importlib bootstrap frames and any cascading module loads that happen above
`bllip.py` in the import tree, so tools like `objgraph.show_backrefs` may
need generous `max_depth` to reach the user frame at all.

## Proposed fix

Don't retain the exception object at all; keep only its message:

```python
except ImportError as ie:
    _bllip_import_error_msg = str(ie)

    def _ensure_bllip_import_or_error():
        raise ImportError(
            "Couldn't import bllipparser module: %s" % _bllip_import_error_msg
        )
```

Alternatively, clear the traceback before capturing:

```python
except ImportError as ie:
    _retained_ie = ie.with_traceback(None)

    def _ensure_bllip_import_or_error(ie=_retained_ie):
        raise ImportError("Couldn't import bllipparser module: %s" % ie)
```

The first option is cleaner because the `ImportError` object itself isn't
retained — just a string — so there's nothing for `__traceback__` /
`__context__` / `__cause__` to hang off. I'd happily send a PR.

## Related reading

- [PEP 3134](https://peps.python.org/pep-3134/) describes how `except ... as
  e:` auto-deletes `e` when the block exits — but only the **name**; the
  object itself persists if captured elsewhere (such as a nested function's
  default argument).
- CPython's `_call_with_frames_removed` hides importlib frames from formatted
  tracebacks but **not** from the underlying frame objects, so the `f_back`
  chain is still complete and still pins user code.

## Environment

| | |
|---|---|
| nltk | 3.9.1 (bug also present in 3.9.4 and `develop` head, verified via `git blame`) |
| Python | 3.11.5 |
| OS | macOS (Darwin), but the mechanism is platform-independent |
| `bllipparser` installed | no (required for repro) |

Muhtasim-Munif-Fahim:
## Summary
- stop capturing the missing bllipparser ImportError in a function default
- retain only the import error message and raise a fresh ImportError on demand
- add a regression test that forces bllipparser imports to fail and verifies no exception object is retained

## Root cause
When bllipparser is missing, `nltk.parse.bllip` stored the ImportError instance in `_ensure_bllip_import_or_error.__defaults__`. That retained the exception traceback and could pin frames from the first import site for the lifetime of the process.

Fixes #3554

## Testing
- `python -m pytest nltk/test/unit/test_bllip.py nltk/test/unit/test_bllip_importerror.py -q`

### DIFF
--- a/nltk/parse/bllip.py
+++ b/nltk/parse/bllip.py
@@ -91,9 +91,12 @@ try:
         pass
 
 except ImportError as ie:
+    _bllip_import_error_message = str(ie)
 
-    def _ensure_bllip_import_or_error(ie=ie):
-        raise ImportError("Couldn't import bllipparser module: %s" % ie)
+    def _ensure_bllip_import_or_error():
+        raise ImportError(
+            "Couldn't import bllipparser module: %s" % _bllip_import_error_message
+        )
 
 
 def _ensure_ascii(words):
--- a/nltk/test/unit/test_bllip_importerror.py
+++ b/nltk/test/unit/test_bllip_importerror.py
@@ -1,0 +1,34 @@
+import builtins
+import importlib.util
+import uuid
+from pathlib import Path
+
+import pytest
+
+
+def _load_bllip_module_without_bllipparser(monkeypatch):
+    original_import = builtins.__import__
+
+    def fake_import(name, globals=None, locals=None, fromlist=(), level=0):
+        if name == "bllipparser" or name.startswith("bllipparser."):
+            raise ModuleNotFoundError("No module named 'bllipparser'")
+        return original_import(name, globals, locals, fromlist, level)
+
+    monkeypatch.setattr(builtins, "__import__", fake_import)
+
+    module_path = Path(__file__).resolve().parents[2] / "parse" / "bllip.py"
+    module_name = f"_nltk_bllip_importerror_probe_{uuid.uuid4().hex}"
+    spec = importlib.util.spec_from_file_location(module_name, module_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_bllip_import_error_helper_does_not_retain_import_exception(monkeypatch):
+    module = _load_bllip_module_without_bllipparser(monkeypatch)
+    message = "Couldn't import bllipparser module: No module named 'bllipparser'"
+
+    assert module._ensure_bllip_import_or_error.__defaults__ is None
+
+    with pytest.raises(ImportError, match=message):
+        module._ensure_bllip_import_or_error()
--- a/nltk/test/unit/test_bllip_importerror.py
+++ b/nltk/test/unit/test_bllip_importerror.py
@@ -1,5 +1,6 @@
 import builtins
 import importlib.util
+import re
 import uuid
 from pathlib import Path
 
@@ -19,6 +20,8 @@ def _load_bllip_module_without_bllipparser(monkeypatch):
     module_path = Path(__file__).resolve().parents[2] / "parse" / "bllip.py"
     module_name = f"_nltk_bllip_importerror_probe_{uuid.uuid4().hex}"
     spec = importlib.util.spec_from_file_location(module_name, module_path)
+    assert spec is not None, f"Failed to create import spec for {module_path}"
+    assert spec.loader is not None, f"Import spec for {module_path} is missing a loader"
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
     return module
@@ -26,9 +29,11 @@ def _load_bllip_module_without_bllipparser(monkeypatch):
 
 def test_bllip_import_error_helper_does_not_retain_import_exception(monkeypatch):
     module = _load_bllip_module_without_bllipparser(monkeypatch)
-    message = "Couldn't import bllipparser module: No module named 'bllipparser'"
 
     assert module._ensure_bllip_import_or_error.__defaults__ is None
 
-    with pytest.raises(ImportError, match=message):
+    with pytest.raises(
+        ImportError,
+        match=re.escape("Couldn't import bllipparser module:") + r".*bllipparser",
+    ):
         module._ensure_bllip_import_or_error()


ekaf:
Hi @Muhtasim-Munif-Fahim , you would need to resolve Copilot's review concerns (above, and on other of your PRs), before the process can continue.

ekaf:
This fixes a real memory-lifetime issue by avoiding retention of an ImportError traceback in function defaults, keeps behavior user-visible equivalent (still raises ImportError when bllipparser is missing), and adds a focused regression test to prevent recurrence. For NLTK, it’s a moderate-importance merge: not a broad feature blocker, but a valuable correctness and robustness fix that reduces subtle long-lived frame retention in processes that import parsing modules.

PR #3604, opened 2026-06-08
Avoid RIBES zero division on empty inputs

Muhtasim-Munif-Fahim:
## Summary
- return a zero RIBES score instead of raising when the hypothesis is empty
- return `0.0` for an empty corpus input instead of dividing by zero during macro-averaging
- add regression coverage for empty hypotheses and an empty corpus

## Root cause
RIBES divides by `len(hypothesis)` when computing both brevity penalty and unigram precision. An empty hypothesis therefore raised `ZeroDivisionError` before the metric could fall back to the existing lowest-score behavior used for unusable alignments.

Partial fix for #1785

## Testing
- `python -m pytest nltk/test/unit/test_ribes.py -q`

### DIFF
--- a/nltk/test/unit/test_ribes.py
+++ b/nltk/test/unit/test_ribes.py
@@ -1,3 +1,5 @@
+import pytest
+
 from nltk.translate.ribes_score import corpus_ribes, sentence_ribes, word_rank_alignment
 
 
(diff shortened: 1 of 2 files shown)


ekaf (approved):
The latest revision goes beyond the original zero-division fix and also addresses the two substantive Copilot concerns about empty references and mismatched corpus lengths. The new behavior is much more consistent for a scoring API, and the added regression tests make the edge-case semantics explicit. This should improve the reliability of NLTK’s translation metrics with very little risk.

Issue #1700, opened 2017-04-24
sys.platform.startswith('win') is not a reliable way of detecting the OS


doweaver:
've been doing some work with the NLTK using IronPython and Windows, and discovered we'll often run into trouble with file paths. It appears that we mostly use the following to determine whether we're on Windows:

`windows = sys.platform.startswith('win')`

Unfortunately, IronPython has a sys.platform of "cli". I haven't been able to dig up historical reasons for this, but I have identified several alternate ways to identify the platform:

```
>>> os.name
'nt'
>>> platform.system()
'Windows'
```

Both of these appear to be more reliable than "sys.platform". Does anyone know if there's a particular reason these aren't being used? If there are no objections, I can create a pull request with the changes.

alvations:
@doweaver Thank you for raising the issue. I think IronPython comes with other caveats that doesn't follow the canonical Python conventions. 

But the latest 27.1 version of [IronPython `sys.platform`](https://ironpython-test.readthedocs.io/en/latest/library/sys.html#sys.platform) works very similar to the canonical [CPython `sys.platform`](https://docs.python.org/3/library/sys.html).

Which version of IronPython are you using?

I think the `cli` was a bug that got rectified in later versions. Try updating IronPython and see whether the `cli` persist.

doweaver:
I'm using IronPython 2.7.7, which according to ironpython.net/downloads is the latest stable version, and I still see 'cli':

```
IronPython 2.7.7 (2.7.7.0) on .NET 4.0.30319.42000 (32-bit)
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.platform
'cli'
```

I also installed IronPython (version 2.7.5) on Arch Linux and see 'cli' output for sys.platform.

I imagine the IronPython authors intend sys.platform to be consistent for IronPython across every OS. The Python docs on sys.platform don't really enlighten me as to whether that's the RIGHT thing to do :)

alvations:
@doweaver I'm not familiar enough with IronPython to know how to proceed. 

One possible way is to go through these lines 

```
$ grep -n -r "sys.platform.startswith('win')" *
Binary file nltk/__pycache__/data.cpython-35.pyc matches
nltk/data.py:80:if sys.platform.startswith('win'):
nltk/data.py:114:    >>> windows = sys.platform.startswith('win')
nltk/data.py:141:    >>> windows = sys.platform.startswith('win')
[...]

PR #3605, opened 2026-06-08
Use os.name for Windows path handling

Muhtasim-Munif-Fahim:
## Summary
- use `os.name == "nt"` for active Windows-specific path handling in `nltk.data` and the downloader
- use `os.pathsep` instead of `sys.platform` when joining MaltParser classpaths
- add regression tests that simulate the IronPython-style `sys.platform="cli"`, `os.name="nt"` case

## Root cause
Several active code paths used `sys.platform.startswith("win")` or `sys.platform == "win32"` as a proxy for Windows filesystem behavior. That misses runtimes such as IronPython, where `sys.platform` is not `win*` even though `os.name` is `nt` and Windows path semantics still apply.

Partial fix for #1700

## Testing
- `python -m pytest nltk/test/unit/test_data.py nltk/test/unit/test_platform_detection.py -q`

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -120,6 +120,10 @@ from nltk.internals import deprecated
 
 textwrap_indent = functools.partial(textwrap.indent, prefix="  ")
 
+
+def _is_windows():
+    return os.name == "nt"
+
 ######################################################################
 # Search Path
 ######################################################################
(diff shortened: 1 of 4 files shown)


ekaf:
@Muhtasim-Munif-Fahim, you have left the review concerns unaddressed for 3 weeks, so your PRs may soon be considered stale.

ekaf (approved):
The follow-up commits addressed the meaningful Copilot concerns by making the tests more robust cross-platform, aligning the doctest/documentation logic with the new _is_windows() helper, and removing unnecessary platform patching in the MaltParser test. The final result is a clearer and more internally consistent Windows-path handling change across nltk.data, nltk.downloader, and nltk.parse.malt. This is a worthwhile compatibility improvement, especially for nonstandard Windows runtimes.

PR #3612, opened 2026-06-12
fix(security): prevent ReDoS in TextTilingTokenizer paragraph-break regex (CWE-1333)

LinZiyuu:
## Summary
`TextTilingTokenizer._mark_paragraph_breaks` scans the input for blank-line
paragraph breaks. The pattern was a plain greedy `re` regex, so `finditer`
rescans a long horizontal-whitespace run that has **no blank line**
quadratically — a crafted whitespace blob hangs the public `tokenize()`
(ReDoS, CWE-1333). This is the same class as **CVE-2021-3828** and
**CVE-2021-43854** (ReDoS in NLTK tokenizers/readers).

## Details
```python
# nltk/tokenize/texttiling.py — _mark_paragraph_breaks()
pattern = re.compile("[ \t\r\f\v]*\n[ \t\r\f\v]*\n[ \t\r\f\v]*")
matches = pattern.finditer(text)
```
`tokenize()` calls `_mark_paragraph_breaks(text)` first, on the raw, untrusted
text. For an input that is a long run of horizontal whitespace with no `\n\n`,
`finditer` retries at every position and the leading `[ \t\r\f\v]*` greedily
expands the run before failing — O(n) positions × O(n) expansion = **O(n²)**.

```python
from nltk.tokenize.texttiling import TextTilingTokenizer
tt = TextTilingTokenizer(stopwords=["the", "a", "of"])
tt.tokenize(" \t" * 64000)   # ~128 KB whitespace -> hangs (quadratic)
```
Measured (`_mark_paragraph_breaks`, same input size): 16 KB → 0.17 s, 32 KB →
0.74 s, 64 KB → 2.86 s, 128 KB → 11.3 s (clean quadratic); ordinary prose of the
same size returns in ~0.002 s.

## Fix
Compile the pattern with the `regex` module (already a hard dependency,
`regex>=2021.8.3`) using **possessive quantifiers** `*+`, so each whitespace run
is non-backtracking. The whitespace class does not overlap `\n`, so a successful
match never needs to backtrack into the whitespace — the possessive form is
**match-for-match identical** to the greedy one (verified identical `finditer`
output on representative inputs), while the wasted backtracking on failed
positions is eliminated and the scan becomes linear.

```python
pattern = regex.compile(r"[ \t\r\f\v]*+\n[ \t\r\f\v]*+\n[ \t\r\f\v]*+")
```

## Tests
`nltk/test/unit/test_texttiling_security.py`:
- `_mark_paragraph_breaks` still returns the same break positions on normal text
  (including breaks containing horizontal whitespace);
- the same ~256 KB whitespace blob completes well within a 15 s timeout, both at
  `_mark_paragraph_breaks` and end-to-end via `tokenize()` (run in a spawned
  process with a hard timeout so a quadratic regression fails fast and cannot
  keep burning CPU for the rest of the suite), where the previous regex takes
  tens of seconds.

### DIFF
--- a/nltk/test/unit/test_texttiling_security.py
+++ b/nltk/test/unit/test_texttiling_security.py
@@ -1,0 +1,106 @@
+"""Regression tests for ReDoS in TextTilingTokenizer (CWE-1333).
+
+``_mark_paragraph_breaks`` scans the input for blank-line paragraph breaks with
+``[ \\t\\r\\f\\v]*\\n[ \\t\\r\\f\\v]*\\n[ \\t\\r\\f\\v]*``. With the plain greedy
+``re`` pattern, ``finditer`` rescans a long horizontal-whitespace run that has no
+blank line quadratically, so a crafted whitespace blob hangs ``tokenize()``. The
+pattern now uses possessive quantifiers (regex module), making the scan linear.
+The whitespace class does not overlap ``"\\n"``, so the matches are unchanged.
+
+The "must not hang" tests run the work in a separate process (spawn) with a hard
+timeout and ``terminate()`` on overrun, so a regression to a quadratic regex
+cannot keep burning CPU for the rest of the suite, and any exception in the
+worker is propagated back to the assertions instead of being swallowed.
+"""
+
+import multiprocessing
+import queue
+
+from nltk.tokenize.texttiling import TextTilingTokenizer
+
+# A long horizontal-whitespace run with no blank line: ~256 KB. Linear with the
+# possessive pattern (sub-millisecond); ~quadratic and tens of seconds with the
+# old greedy one.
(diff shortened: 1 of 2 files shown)


LinZiyuu:
1. test_texttiling_security.py:87 — 86 chars
2. test_texttiling_security.py:105 — 85 chars

Both are under the 88-character limit. So no change needed.

ekaf (approved):
Copilot’s two comments were about Black formatting on test assertions, but the author checked the actual line lengths and both are already under the repo’s 88-character limit. No follow-up commit was needed, and nothing suggests an unresolved code issue. Merge importance: medium — good security hardening, but narrower impact than the parser/XML fixes.

PR #3613, opened 2026-06-12
fix(security): cap zip decompression to prevent decompression-bomb DoS (CWE-409)

LinZiyuu:
## Summary
NLTK reads a whole zip member into memory (`ZipFilePathPointer.open`,
`OpenOnDemandZipFile.read`) and extracts members to disk (`nltk.downloader`)
with **no size limit**, so a tiny archive whose member decompresses to gigabytes
exhausts RAM or disk — a classic **decompression bomb** (CWE-409). The compressed
and declared uncompressed sizes are known at read time but were never checked.

## Details
```python
# nltk/data.py — ZipFilePathPointer.open()
data = self._zipfile.read(self._entry)     # whole member -> RAM, no cap
# nltk/data.py — OpenOnDemandZipFile.read()
value = zipfile.ZipFile.read(self, name)   # whole member -> RAM, no cap
# nltk/downloader.py — _unzip_iter()
zf.extract(member, root_abs)               # member -> disk, no cap
```
Reachable via `nltk.data.load(format="raw"/"pickle"/text)` and
`CorpusReader.raw()` (memory), and `nltk.download()` extraction (disk), with an
untrusted or oversized zip corpus (a malicious/compromised data mirror, or a
planted `.zip` in a shared `nltk_data`) — the scenario covered by NLTK's
SECURITY.md ("untrusted input … shared environments … multi-tenant pipelines").

```python
import zipfile
zf = zipfile.ZipFile("evil.zip", "w", zipfile.ZIP_DEFLATED)
zf.writestr("evilcorp/big.txt", b"\0" * (512 * 1024 * 1024))  # ~510 KB archive
zf.close()

from nltk.data import ZipFilePathPointer
ZipFilePathPointer("evil.zip", "evilcorp/big.txt").open().read()  # pulls 512 MiB into RAM
```
Measured: a ~510 KB archive → 512 MiB read into memory (~1029x); the downloader
extract path writes 1 GiB to disk from a 1 MB archive. No error, no cap.

## Fix
Add `nltk.data._check_decompression_bomb(info)`, applied at the three sinks
(`ZipFilePathPointer.open`, `OpenOnDemandZipFile.read`, and the downloader's
extract loop). It refuses a member whose declared uncompressed size:
- exceeds the optional absolute cap `MAX_UNZIP_SIZE` (default `None`, off — so
  legitimately large corpora are never affected), **or**
- is both large (≥ `MAX_UNZIP_ACTIVATION`, default 32 MiB) **and** expands beyond
  `MAX_UNZIP_RATIO` (default 1000) over its stored compressed size.

DEFLATE's maximum ratio is ~1032x, reached only by runs of identical bytes
(payloads crafted purely to maximize expansion); legitimate text/markup corpora
compress a few-fold to a few-hundred-fold at most, well under 1000x, so the guard
does not affect them. All limits are module-level and configurable
(`nltk.data.MAX_UNZIP_RATIO` / `MAX_UNZIP_SIZE` / `MAX_UNZIP_ACTIVATION`).

## Tests
`nltk/test/unit/test_zipbomb_security.py`:
- a member expanding past the ratio (above activation) is rejected at the
  read-into-memory path and at the downloader extract path (nothing is written);
- incompressible / low-ratio data and small high-ratio files below the
  activation size are **not** rejected (no false positives);
- the optional `MAX_UNZIP_SIZE` hard cap rejects oversize members.

Existing `test_data.py` / `test_downloader.py` / `test_pathsec.py` pass unchanged.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -550,15 +550,16 @@ def _check_decompression_bomb(info):
     if (
         file_size >= MAX_UNZIP_ACTIVATION
         and compress_size > 0
-        and file_size / compress_size > MAX_UNZIP_RATIO
+        # integer comparison (no float rounding at the threshold)
+        and file_size > MAX_UNZIP_RATIO * compress_size
     ):
         raise ValueError(
-            "Refusing to decompress suspected zip bomb %r: it expands %dx "
+            "Refusing to decompress suspected zip bomb %r: it expands %.1fx "
             "(%d -> %d bytes), above nltk.data.MAX_UNZIP_RATIO=%d. "
             "Raise nltk.data.MAX_UNZIP_RATIO if this data is trusted."
             % (
                 name,
-                file_size // compress_size,
+                file_size / compress_size,
                 compress_size,
                 file_size,
                 MAX_UNZIP_RATIO,
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -506,6 +506,66 @@ class GzipFileSystemPathPointer(FileSystemPathPointer):
         return stream
 
 
+#: Maximum allowed ratio of a zip member's uncompressed size to its stored
+#: (compressed) size before it is treated as a decompression bomb (CWE-409).
+#: DEFLATE's maximum ratio is ~1032x, reached only by runs of identical bytes,
+#: i.e. payloads crafted purely to maximize expansion; legitimate text/markup
+#: corpora compress a few-fold to a few-hundred-fold at most, well under this.
+#: A member that expands beyond this ratio is rejected. Configurable.
+MAX_UNZIP_RATIO = 1000
+
+#: Optional hard cap (in bytes) on a single zip member's uncompressed size.
+#: ``None`` disables it (the default, so legitimately large corpora are not
+#: affected); set it for a strict absolute limit in hardened deployments.
+MAX_UNZIP_SIZE = None
+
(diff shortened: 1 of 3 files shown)


ekaf:
@LinZiyuu, please mark the review concerns above as resolved, preferably with an explanation.

LinZiyuu:
1. Validate-then-extract contract: moved _check_decompression_bomb() into Phase 1 (validation) so a bomb member is rejected before any earlier, benign member is written to disk, and removed the per-member check from Phase 2. Added a regression test that a bomb following a benign member writes nothing.
2. Threshold/message consistency: the check now uses an integer comparison (file_size > MAX_UNZIP_RATIO * compress_size) and the message reports the ratio as %.1f, so the decision and the reported ratio agree at the boundary.
3. Docstring E501: no line in the module exceeds 88 chars and ruff (0.15.6, the pinned hook) passes. E501 isn't enabled and the pre-commit CI check is green. So this one is a false positive; left as-is.

ekaf (approved):
This one had the most substantive review feedback, and it was addressed in the follow-up commit 7a00fae: the decompression-bomb check was moved into validation before extraction, the ratio comparison was changed to integer math, and a regression test was added so no benign files are written before rejection. The remaining docstring complaint looks non-blocking / likely false positive. Merge importance: high — protects both memory and disk paths against decompression-bomb inputs.

PR #3614, opened 2026-06-12
fix(security): prevent ReDoS in SensevalCorpusReader _fixXML (CWE-400)

LinZiyuu:
## Summary
`SensevalCorpusReader` normalises each `<instance>` block with `_fixXML` before
XML parsing. Two of its substitutions used plain `re` patterns whose lazy/greedy
whitespace and token runs rescan a long token / whitespace run that lacks the
trailing `<p="..."/>` tag **quadratically** — a crafted instance body hangs the
reader (ReDoS, CWE-400). This is the same class as **CVE-2021-3828** (ReDoS in
`comparative_sents._read_comparison_block`).

## Details
```python
# nltk/corpus/reader/senseval.py — _fixXML()
text = re.sub(r'[ \t]*([^<>\s]+?)[ \t]*<p="([^"]*"?)"/>', r' <wf pos="\2">\1</wf>', text)
text = re.sub(r'\s*"\s*<p=\'"\'/>',                       " <wf pos='\"'>\"</wf>", text)
```
For an instance body that is a long run of non-`<>`/non-whitespace characters
(or whitespace) with no `<p="…"/>` tag, the lazy `([^<>\s]+?)` (and the `\s*`
runs) expand to the end at every start position before failing — O(n) positions
× O(n) expansion = **O(n²)**.

```python
import os, tempfile
from nltk.corpus.reader.senseval import SensevalCorpusReader
d = tempfile.mkdtemp()
open(os.path.join(d, "t.pos"), "w").write(
    '<lexelt item="t.n">\n<instance id="t.1">\n<context>\n' + "x"*128000 +
    '\n</context>\n</instance>\n</lexelt>\n')
SensevalCorpusReader(d, ["t.pos"]).instances()   # hangs (quadratic)
```
Measured (end-to-end `instances()`): 8 KB → 0.6 s, 16 KB → 2.3 s, 32 KB → 9.5 s
(clean quadratic); after this change a 128 KB body returns in ~0.004 s.

## Fix
Compile both substitutions with the `regex` module (already a hard dependency,
`regex>=2021.8.3`) using **possessive quantifiers**. The token class `[^<>\s]`
cannot cross its surrounding separators and `\s` cannot cross the literal `"`,
so a successful match never needs to backtrack into those runs — the possessive
forms are **match-for-match identical** to the originals (verified on
representative inputs, including the tag cases and the no-tag case), while the
wasted backtracking on failed positions is eliminated and the scan becomes
linear.

```python
text = regex.sub(r'[ \t]*+([^<>\s]++)[ \t]*+<p="([^"]*+"?)"/>', r' <wf pos="\2">\1</wf>', text)
text = regex.sub(r"\s*+\"\s*+<p='\"'/>",                        " <wf pos='\"'>\"</wf>", text)
```

## Tests
`nltk/test/unit/test_senseval_security.py`:
- `_fixXML` still produces the same output on normal data (both `<p="…"/>` and
  the `<p='"'/>` variant, and a no-tag string);
- a ~256 KB tag-less token completes well within a 15 s timeout, both at the
  `_fixXML` level and end-to-end via `instances()` (run in a spawned process with
  a hard timeout so a quadratic regression fails fast and cannot keep burning CPU
  for the rest of the suite), where the previous patterns take tens of seconds.

### DIFF
--- a/nltk/corpus/reader/senseval.py
+++ b/nltk/corpus/reader/senseval.py
@@ -25,6 +25,8 @@ is tagged with a sense identifier, and supplied with context.
 import re
 from xml.etree import ElementTree
 
+import regex
+
 from nltk.corpus.reader.api import *
 from nltk.corpus.reader.util import *
 from nltk.tokenize import *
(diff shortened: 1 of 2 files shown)


LinZiyuu:
1. Removed the unused os import.
2. Corrected the comments: the crafted token is ~128 KB ("x" * 128_000), not ~256 KB.
3. Pass the fileid as a one-element list (["t.pos"]) so CorpusReader treats it as an exact fileid rather than a regexp.

ekaf (approved):
All three Copilot concerns were cleanly handled in follow-up commit da87c8c: removed unused os, corrected the misleading “~256 KB” comments to ~128 KB, and changed the fileid argument to ["t.pos"] so it is treated as an exact file id rather than a regex. Merge importance: medium-high — closes a real ReDoS-style parser hang in corpus handling.

PR #3616, opened 2026-06-12
fix(security): resolve relative MaltParser dir via MALT_PARSER, not the CWD (CWE-426)

LinZiyuu:
## Summary
`nltk.parse.malt.find_maltparser()` accepted `parser_dirname` as-is whenever
`os.path.exists()` was true. For a **relative** name — which is NLTK's own
documented usage (`MaltParser('maltparser-1.7.2', 'engmalt.linear-1.7.mco')`) —
that resolves against the process **current working directory**, and it was
checked **before** the `MALT_PARSER` environment variable. So a `./maltparser-*/`
directory in the CWD silently overrides a trusted, configured `MALT_PARSER`, its
jars are placed on the Java classpath, and NLTK runs
`java -cp <those jars> org.maltparser.Malt`. An attacker who can write a directory
into the application's CWD therefore gets their `org.maltparser.Malt` executed —
arbitrary code execution via an **untrusted search path** (CWE-426 / CWE-427).

Writable CWDs are common (web-app working dirs, pipeline scratch dirs,
upload-staging dirs, shared multi-user dirs), and the planted directory only needs
the jar *names* NLTK asserts on (`log4j.jar`, `libsvm.jar`, `liblinear-1.8.jar`,
`maltparser-*.jar`); their contents are attacker-controlled.

## Details
```python
# nltk/parse/malt.py — find_maltparser()
if os.path.exists(parser_dirname):          # relative name -> CWD, before MALT_PARSER
    _malt_dir = parser_dirname
else:
    _malt_dir = find_dir(parser_dirname, env_vars=("MALT_PARSER",))
_malt_jars = set(find_jars_within_path(_malt_dir))   # -> java -cp ... org.maltparser.Malt
```

With `MALT_PARSER` set to a trusted directory and a malicious `./maltparser-1.9.2/`
in the CWD whose `maltparser-1.9.2.jar` defines `org.maltparser.Malt`:
```python
from nltk.parse.malt import find_maltparser
find_maltparser("maltparser-1.9.2")   # selects the CWD jars, NOT MALT_PARSER
# NLTK then runs: java -cp <CWD jars> org.maltparser.Malt  -> attacker code runs
```
Verified end-to-end with a JDK: the attacker's `org.maltparser.Malt` executed,
overriding the configured `MALT_PARSER`.

## Fix
Only accept an explicit **absolute** directory as-is (matching the original "if a
full path is given" intent); resolve a relative name through `MALT_PARSER`
instead of the current working directory:

```python
if os.path.isabs(parser_dirname) and os.path.exists(parser_dirname):
    _malt_dir = parser_dirname
else:
    _malt_dir = find_dir(parser_dirname, env_vars=("MALT_PARSER",))
```

Note: a relative `parser_dirname` now requires `MALT_PARSER` (or an absolute
path) rather than silently using the CWD. The malt doctests use `+SKIP`, so this
behaviour change does not affect CI.

## Tests
`nltk/test/unit/test_malt_security.py`:
- a relative `parser_dirname` matching a directory in the CWD is **not** used
  (discovery raises instead of silently using the CWD);
- a CWD `maltparser-*/` directory does **not** override a configured
  `MALT_PARSER`;
- an explicit absolute directory is still accepted.

black `25.11.0` / isort `6.1.0` / ruff `0.15.6` clean.

### DIFF
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -62,7 +62,13 @@ def find_maltparser(parser_dirname):
     """
     A module to find MaltParser .jar file and its dependencies.
     """
-    if os.path.exists(parser_dirname):  # If a full path is given.
+    # Only accept an explicit *absolute* directory as-is. A relative name must
+    # not be resolved against the current working directory: an attacker able to
+    # write to the CWD could otherwise plant a ``maltparser-*/`` directory there
+    # and have its jars placed on the Java classpath (and its ``org.maltparser``
+    # main class executed), overriding a trusted ``MALT_PARSER`` -- an untrusted
+    # search path (CWE-426). A relative name is resolved through ``MALT_PARSER``.
+    if os.path.isabs(parser_dirname) and os.path.exists(parser_dirname):
         _malt_dir = parser_dirname
     else:  # Try to find path to maltparser directory in environment variables.
         _malt_dir = find_dir(parser_dirname, env_vars=("MALT_PARSER",))
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -62,13 +62,15 @@ def find_maltparser(parser_dirname):
     """
     A module to find MaltParser .jar file and its dependencies.
     """
+    # Accept str or os.PathLike uniformly (find_dir() requires a str).
+    parser_dirname = os.fspath(parser_dirname)
     # Only accept an explicit *absolute* directory as-is. A relative name must
     # not be resolved against the current working directory: an attacker able to
     # write to the CWD could otherwise plant a ``maltparser-*/`` directory there
     # and have its jars placed on the Java classpath (and its ``org.maltparser``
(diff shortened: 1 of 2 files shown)


LinZiyuu:
1. parser_dirname is now normalised with os.fspath(), so a pathlib.Path is handled uniformly and a relative Path no longer trips find_dir()'s str assertion — it raises a clean LookupError instead.
2. The explicit-absolute branch now checks os.path.isdir() rather than os.path.exists(), since the path is subsequently walked for jars. Added a test covering the PathLike handling.

ekaf (approved):
Copilot’s review noted two real edge cases; both were fixed in follow-up commit 19daecd: parser_dirname is normalized with os.fspath() and the explicit-path branch now requires os.path.isdir() rather than mere existence. That resolves the PathLike regression and directory-validation concern. Merge importance: high — prevents CWD-based MaltParser hijacking / untrusted search path execution.

PR #3617, opened 2026-06-12
fix(security): bound LogicParser recursion depth to prevent uncaught RecursionError (CWE-674)

LinZiyuu:
## Summary
`nltk.sem.logic.LogicParser` is a recursive-descent parser: every nested
sub-expression recurses through `process_next_expression`. Deeply nested input
recurses until Python raises an **uncaught `RecursionError`**, crashing the
caller — a denial of service from a few hundred bytes of input (uncontrolled
recursion, CWE-674). This is the same class as **GHSA-rf74-v2fm-23pw**, which the
project fixed for `JSONTaggedDecoder` by adding a depth cap.

## Details
```python
def process_next_expression(self, context):   # recursion point
    ...
    accum = self.handle(tok, context)          # -> handle_* -> process_next_expression
    return self.attempt_adjuncts(accum, context)
```
Reachable via the public `Expression.fromstring` / `LogicParser().parse` (and the
`DrtParser` subclass).

```python
from nltk.sem.logic import Expression
Expression.fromstring("-" * 5000 + "p")   # uncaught RecursionError (crashes the request)
```
Measured at CPython's default recursion limit: ~330 nested operators already
raise `RecursionError`; a few hundred bytes suffice.

## Fix
Cap the expression-nesting depth at `LogicParser.MAX_PARSE_DEPTH` (default `200`,
well below the ~330 at which the default recursion limit is reached) and raise a
normal `LogicalExpressionException` past it — the parser's own error type, which
`parse()` already handles — instead of recursing into a crash. The cap is a
configurable class attribute. Ordinary expressions (which nest only a handful of
levels) are unaffected.

```python
class LogicParser:
    MAX_PARSE_DEPTH = 200
    def process_next_expression(self, context):
        self._parse_depth += 1
        try:
            if self._parse_depth > self.MAX_PARSE_DEPTH:
                raise LogicalExpressionException(self._currentIndex, "... maximum depth ...")
            ...
        finally:
            self._parse_depth -= 1
```

## Tests
`nltk/test/unit/test_logic_security.py`:
- a 5000-deep expression raises `LogicalExpressionException`, **not**
  `RecursionError`;
- ordinary expressions (`exists`, `all`, lambda, application, negation) still
  parse;
- with a lowered `MAX_PARSE_DEPTH`, input below the cap parses and input above it
  raises a clean error.

`nltk/test/logic.doctest` still passes; black `25.11.0` / isort `6.1.0` /
ruff `0.15.6` clean.

## Note
The same uncontrolled-recursion class exists in other NLTK recursive-descent
parsers (`FeatStructReader` / `FeatureGrammar.fromstring`, `DependencyGraph.tree()`,
and `Tree` `str`/`repr`/`__eq__` on a deeply nested tree). This PR fixes the logic
parser (the smallest-payload, cleanest case); the others can follow the same
pattern if desired.

### DIFF
--- a/nltk/sem/logic.py
+++ b/nltk/sem/logic.py
@@ -99,6 +99,13 @@ def binding_ops():
 class LogicParser:
     """A lambda calculus expression parser."""
 
+    #: Maximum expression-nesting depth the recursive-descent parser will
+    #: descend to. Deeply nested input would otherwise recurse until Python
+    #: raises an uncaught ``RecursionError`` and crashes the caller
+    #: (uncontrolled recursion, CWE-674); past this depth a normal
+    #: ``LogicalExpressionException`` is raised instead. Configurable.
+    MAX_PARSE_DEPTH = 200
+
     def __init__(self, type_check=False):
         """
         :param type_check: should type checking be performed
@@ -109,6 +116,7 @@ class LogicParser:
 
         self._currentIndex = 0
         self._buffer = []
+        self._parse_depth = 0
         self.type_check = type_check
 
         """A list of tuples of quote characters.  The 4-tuple is comprised
@@ -148,6 +156,7 @@ class LogicParser:
         data = data.rstrip()
 
         self._currentIndex = 0
+        self._parse_depth = 0
         self._buffer, mapping = self.process(data)
 
         try:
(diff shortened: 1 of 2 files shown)


LinZiyuu:
1. The depth-limit LogicalExpressionException now uses a 1-based index (self._currentIndex + 1), consistent with ExpectedMoreTokensException. This avoids a secondary KeyError in parse()'s caret formatting (mapping[index - 1]) when the limit is hit before any token is consumed (e.g. MAX_PARSE_DEPTH = 0).
2. Removed the redundant second parse in the test (pytest.raises(LogicalExpressionException) already fails on an unexpected RecursionError) and corrected the wording — this is uncontrolled recursion (CWE-674).
3. Fixed the docstring: nesting up to the cap parses; only nesting beyond it raises (the guard is > MAX_PARSE_DEPTH).

ekaf (approved):
All three Copilot concerns were addressed in follow-up commit 3c1df38: the depth-limit exception now uses a 1-based index, the redundant second parse was removed from the test, and the test wording was corrected so behavior matches the actual > MAX_PARSE_DEPTH guard. Merge importance: high — protects a public parser API from uncaught recursion crashes with a relatively contained, low-risk fix.

PR #3621, opened 2026-06-14
fix(security): do not run the senna binary from the CWD (CWE-829)

LinZiyuu:
## Summary
`nltk.classify.senna.Senna` built the path to the external `senna` executable by
joining a caller-supplied directory with a fixed binary name and then ran it with
`subprocess.Popen`. When `senna_path` was a **relative** directory (e.g. `"."`),
the resulting program path **contained a separator** (e.g. `./senna-osx`):

```python
# nltk/classify/senna.py
self._path = path.normpath(senna_path) + sep          # "." -> "./"
...
_senna_cmd = [self.executable(self._path), ...]       # -> "./senna-osx"
p = Popen(_senna_cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE)
```

`subprocess.Popen` executes a program path that contains a separator **relative
to the current working directory** — it does **not** consult `$PATH`. So an
attacker who can place a file named `senna-<platform>`
(`senna-linux64` / `senna-linux32` / `senna-osx` / `senna-win32.exe`) in the
application's CWD has that file executed as a native program — **arbitrary code
execution from an untrusted location (CWE-829)**, reachable via an untrusted
search path (CWE-426/CWE-427). No environment variable, `$PATH` control, or
specific jar/class is required. `Senna` is constructed with a directory argument
by every public entry point (`SennaTagger`, `SennaChunkTagger`, `SennaNERTagger`),
so a relative tools directory reaching this code is realistic.

This is the same class as the MaltParser (#3616), ReppTokenizer (#3618) and Weka
(#3620) CWD hijacks and as CVE-2026-0848.

## Proof of concept (before this change)
With a malicious `./senna-<platform>` planted in the CWD:

```python
from nltk.classify import Senna
pipeline = Senna(".", ["pos"])        # relative senna_path -> "./" base
pipeline.tag("hello world".split())   # runs ./senna-osx  -> attacker code
```
Verified end-to-end: `Senna(".").executable()` returns `./senna-osx` and
`Popen([...])` executes the attacker binary from the CWD (the subsequent
output-parsing `IndexError` is raised *after* the binary has already run).

## Fix
Only resolve the executable against an explicit **absolute** directory. A
relative `senna_path` (including `"."`) no longer falls back to the current
working directory; it falls through to the trusted `SENNA` environment variable
(which must also be absolute):

```python
if path.isabs(senna_path):
    self._path = path.normpath(senna_path) + sep
else:
    self._path = None

[...]

PR #3620, opened 2026-06-14
fix(security): do not load weka.jar from the CWD (CWE-494)

LinZiyuu:
## Summary
`nltk.classify.weka.config_weka()` searched for `weka.jar` with the **current
working directory listed first** in `_weka_search`:

```python
# nltk/classify/weka.py
_weka_search = [".", "/usr/share/weka", "/usr/local/share/weka", ...]   # "." tried first

for path in searchpath:                       # "." (CWD) before WEKAHOME / system dirs
    if os.path.exists(os.path.join(path, "weka.jar")):
        _weka_classpath = os.path.join(path, "weka.jar")                 # -> ./weka.jar
```

A `./weka.jar` in the process CWD is therefore preferred over a system-installed
Weka, with **no integrity/signature check** and **without any environment
variable or non-default argument**. NLTK then loads and runs classes from that
jar:

```python
# nltk/classify/weka.py
cmd = ["weka.classifiers.bayes.NaiveBayes", "-l", self._model, "-T", test_filename] + options
java(cmd, classpath=_weka_classpath, ...)     # java -cp ./weka.jar weka.classifiers.bayes.NaiveBayes ...
```

`config_weka()` is invoked automatically the first time a `WekaClassifier`
classifies or trains, so simply *using* a Weka classifier in a directory
containing an attacker-planted `weka.jar` executes the attacker's Java code —
**use of code without an integrity check (CWE-494)**, reachable via an untrusted
search path (CWE-426). This is the same class as the StanfordSegmenter
unverified-jar load (**CVE-2026-0848**) and the MaltParser (#3616) / ReppTokenizer
(#3618) CWD hijacks.

## Proof of concept (before this change)
Plant a malicious `./weka.jar` whose `weka.classifiers.bayes.NaiveBayes` writes a
sentinel, then use a Weka classifier from that directory:

```python
import nltk.classify.weka as weka
weka.config_weka()                 # selects ./weka.jar (".") over the system install
print(weka._weka_classpath)        # -> './weka.jar'
# WekaClassifier.classify_many()/train() then run:
#   java -cp ./weka.jar weka.classifiers.bayes.NaiveBayes ...  -> attacker code
```
Verified end-to-end (Temurin JDK 21): the CWD `./weka.jar` is selected and the
attacker's `weka.classifiers.bayes.NaiveBayes` class is executed by NLTK's own
`java` command.

## Fix
Do not search the current working directory for `weka.jar` — drop `"."` from
[...]

PR #3623, opened 2026-06-15
fix(security): do not run the candc/boxer binary from the CWD (CWE-426)

LinZiyuu:
## Summary
`nltk.sem.boxer.Boxer` resolves its external `candc` / `boxer` executables with
`nltk.internals.find_binary` and then runs them with `subprocess.Popen`:

```python
# nltk/sem/boxer.py
self._candc_bin = self._find_binary("candc", bin_dir, ...)   # find_binary(path_to_bin=bin_dir)
...
cmd = [binary] + args
p = subprocess.Popen(cmd, ...)                                # boxer.py:267 / 271
```

`find_binary` also matches a binary **relative to the current working
directory**, producing a path that **contains a separator**:

- a **relative** `bin_dir` (e.g. `"."`) yields `./candc`; and
- even the **default `bin_dir=None`** matches a `candc/candc` directory in the
  CWD — `find_file_iter` joins the binary name with itself
  (`os.path.join("candc", "candc")`).

Because the resulting program path contains a separator, `subprocess.Popen`
executes it **directly from the CWD** (it does not consult `$PATH`), and the CWD
match is tried **before** the `CANDC` environment variable. An attacker who can
write a `candc`/`boxer` file into the application's CWD therefore gets their
binary executed — an **untrusted search path** (CWE-426 / CWE-427) leading to
arbitrary code execution. Every public entry point (`Boxer`, and the
`boxer`/`interpret` API) reaches this resolver.

This is the same class as the MaltParser (#3616), ReppTokenizer (#3618), Weka
(#3620) and Senna (#3621) CWD hijacks and as CVE-2026-0848.

With a malicious `./candc` (and `./boxer`) planted in the CWD:

```python
from nltk.sem.boxer import Boxer
Boxer(bin_dir=".").interpret("a man walks")   # runs ./candc  -> attacker code
```
or, with the default configuration, a `./candc/candc` directory:

```python
Boxer().interpret("a man walks")              # find_binary -> "candc/candc" -> CWD exec
```
Verified end-to-end: `find_binary("candc", path_to_bin="candcdir")` returns
`candcdir/candc`, `path_to_bin="."` returns `./candc`, and `path_to_bin=None`
with a planted `./candc/candc` returns `candc/candc`; `Popen([...])` executes the
attacker binary from the CWD.

## Fix
Resolve via `find_binary_iter` and use the first **absolute** candidate, skipping
any CWD-relative match. A relative result can only come from a CWD lookup, so
this rules out CWD execution while still honouring an absolute `bin_dir`, the
`CANDC` environment variable, and a `$PATH` lookup:

```python
for binary in find_binary_iter(name, path_to_bin=bin_dir, env_vars=["CANDC"], ...):
    if os.path.isabs(binary):
        return binary
raise LookupError("No absolute %r binary found. Pass an absolute bin_dir ... "
                  "or set the CANDC environment variable ..." % name)
```

A CWD plant is now skipped in favour of the trusted `CANDC`/`$PATH` location (or
refused if none exists); absolute `bin_dir` usage is unchanged.

## Tests
`nltk/test/unit/test_boxer_security.py`:
- a relative `bin_dir` resolving the binary in the CWD is refused;
- the default `bin_dir=None` does not pick up a `./<name>/<name>` in the CWD;
- a CWD binary does not override an absolute `CANDC` environment variable;
- an explicit absolute `bin_dir` is still used as-is.

black `25.11.0` / isort `6.1.0` / ruff `0.15.6` clean.

### DIFF
--- a/nltk/sem/boxer.py
+++ b/nltk/sem/boxer.py
@@ -36,7 +36,7 @@ import tempfile
 from functools import reduce
 from optparse import OptionParser
 
-from nltk.internals import find_binary
+from nltk.internals import find_binary_iter
 from nltk.sem.drt import (
     DRS,
     DrtApplicationExpression,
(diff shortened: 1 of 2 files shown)


ekaf (approved):
This looks ready to merge: it cleanly removes CWD-based binary resolution, which is the right mitigation for the hijack risk, and the added tests cover the key behaviors without weakening the legitimate absolute-path use case. For NLTK, this is a high-value hardening fix because it closes an execution-path vulnerability with low functional risk.

PR #3630, opened 2026-06-18
fix(security): prevent ReDoS in chunk tag-pattern parsing (CWE-1333)

LiteshGhute:
## Summary

`nltk.chunk.regexp.CHUNK_TAG_PATTERN` is used by `tag_pattern2re_pattern()` to
validate caller-supplied tag patterns, and is reached through the public
`RegexpParser(grammar)` and the chunk-rule constructors. The regex contained a
nested quantifier of the classic `(A+)*` form:

```python
CHUNK_TAG_PATTERN = re.compile(
    r"^(({}|<{}>)*)$".format(r"([^\{\}<>]|\{\d+,?\}|\{\d*,\d+\})+", r"[^\{\}<>]+")
)
```

When the overall match fails on a trailing character the pattern cannot consume
(e.g. `{`, `}`, `<`, `>`), the engine tries exponentially many ways to partition
the preceding run, causing catastrophic backtracking. A short tag pattern such
as `"a" * 26 + "{"` hangs for ~10 s; ~30 characters hangs for minutes. In any
application that compiles user-supplied chunking rules (a common, intended use
of `RegexpParser` / `tag_pattern2re_pattern` in NLP services and annotation
tools), a single short request pins a worker at 100% CPU, a remote DoS.

## Fix

Drop the `+` on the first alternative. `(A+|<B+>)*` and `(A|<B+>)*` accept
exactly the same language (a run of `A`s is just several single-`A` iterations
of the outer star), so the change is fully behaviour-preserving while removing
the nested quantifier and making the scan linear.

```python
CHUNK_TAG_PATTERN = re.compile(
    r"^(({}|<{}>)*)$".format(r"([^\{\}<>]|\{\d+,?\}|\{\d*,\d+\})", r"[^\{\}<>]+")
)
```

## Validation

- Over 200,000 random strings, the patched regex produced identical
  accept/reject results to the original (0 mismatches).
- Every known-valid tag pattern (`<DT>`, `<NN.*>`, `<JJ|NN>`, `<DT>?<JJ>*<NN>`,
  `{2,}`, `{3,5}`, …) still matches.
- The `"a" * 40 + "{"` payload now completes in microseconds instead of
  hanging.

## Tests

Adds `nltk/test/unit/test_chunk_redos_security.py` (same structure as the other
`*_security.py` ReDoS regression tests): behaviour-preserving accept/reject
assertions, plus "must not hang" tests that run the match in a spawned process
with a hard timeout and `terminate()` on overrun, so a regression cannot keep
burning CPU for the rest of the suite.

## Reference

huntr report: https://huntr.com/bounties/aff8ef29-2f20-46a4-ae13-7ce6010e26a5

### DIFF
--- a/nltk/chunk/regexp.py
+++ b/nltk/chunk/regexp.py
@@ -891,8 +891,17 @@ class ChunkRuleWithContext(RegexpChunkRule):
 
 # this should probably be made more strict than it is -- e.g., it
 # currently accepts 'foo'.
+#
+# The first alternative must NOT be quantified with "+" here: nesting it
+# inside the outer "( ... )*" gives the classic ``(A+)*`` shape, which causes
+# catastrophic (exponential) backtracking when the overall match fails on a
+# trailing "{", "}", "<" or ">" (ReDoS, CWE-1333). Dropping the inner "+" is
+# behaviour-preserving -- ``(A+|<B+>)*`` and ``(A|<B+>)*`` accept exactly the
+# same language, because a run of ``A``s is just several single-``A`` iterations
+# of the outer star -- while making the scan linear. See huntr report
+# https://huntr.com/bounties/aff8ef29-2f20-46a4-ae13-7ce6010e26a5.
 CHUNK_TAG_PATTERN = re.compile(
-    r"^(({}|<{}>)*)$".format(r"([^\{\}<>]|\{\d+,?\}|\{\d*,\d+\})+", r"[^\{\}<>]+")
+    r"^(({}|<{}>)*)$".format(r"([^\{\}<>]|\{\d+,?\}|\{\d*,\d+\})", r"[^\{\}<>]+")
 )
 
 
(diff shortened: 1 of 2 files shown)


ekaf (approved):
This is a well-scoped ReDoS fix: removing the nested quantifier from the chunk tag pattern is the right mitigation, and the regression coverage makes the intent explicit. Because this code is exposed through public parsing APIs, it’s worth merging.

alvations:
Note: Strengthen via https://github.com/nltk/nltk/commit/fbe75052b150122ce3ca6756e056de21dac5a75c

ekaf:
Hi @alvations, some of your latest patches come from private GHSA forks, and are not in merged nltk PRs. 
This makes it more difficult (though still feasible) to generate the list of recent changes for updating nltk's ChangeLog.

PR #3631, opened 2026-06-18
fix(security): prevent pickle RCE in TransitionParser model loading (CWE-502)

LiteshGhute:
## Summary

NLTK 3.10 added `nltk/picklesec.py` to mitigate pickle RCE, and `nltk.data.load()`
correctly uses its `RestrictedUnpickler`. But the wrapper `pickle_load()` defaults
to `restricted=False`, which uses `WarningUnpickler`: it prints a warning and then
performs a normal, unrestricted unpickle, executing any code in the input.

`TransitionParser.parse(depgraphs, modelFile)` is a public API whose `modelFile`
is caller-supplied, and it loaded the model with that warn-only default:

```python
with open(modelFile, "rb") as f:
    model = pickle_load(f)        # warns, then executes attacker code
```

So loading an untrusted/tampered model file is arbitrary code execution the
instant it is loaded. NLTK models are routinely trained elsewhere and then
downloaded, cached, and shared, so loading one from an untrusted source is a
normal workflow rather than an edge case, and the shipped pickle-hardening
module makes users reasonably believe model loading is already safe.

## Fix

Add an allowlisting unpickler to `picklesec.py` and route `TransitionParser.parse`
through it. The allowlist permits only the globals a saved object legitimately
needs and rejects everything else, so legitimate model loading keeps working
while arbitrary code execution is removed. It does **not** change the global
`pickle_load` default, so no other caller is affected.

`AllowlistUnpickler` supports two complementary allowlists:

- `allowed_globals`: exact `(module, qualname)` pairs.
- `allowed_modules`: module names; a global is permitted when its module equals
  an allowed name or is a submodule of one.

The `TransitionParser` model is a scikit-learn classifier backed by numpy/scipy
arrays, so `parse()` allows the `numpy` / `scipy` / `sklearn` module trees:

```python
_MODEL_ALLOWED_MODULES = ("numpy", "scipy", "sklearn")
...
with open(modelFile, "rb") as f:
    model = allowlisted_pickle_load(f, allowed_modules=_MODEL_ALLOWED_MODULES)
```

Module-prefix allowlisting (rather than exact pairs) is used so the fix is robust
across numpy/sklearn releases, whose internal module paths change between
versions, while still blocking arbitrary-code gadgets (`os`, `posix`,
`subprocess`, `builtins`, …).

`RestrictedUnpickler` (which blocks *all* globals) was not reused here because it
would break loading a genuine sklearn/numpy-backed model.

## Validation

- A payload reaching for `os.system` is rejected with
  `UnpicklingError: global 'posix.system' is not in the pickle allowlist`
  instead of executing.
- A real `numpy.ndarray` deserializes correctly under the `numpy` allowlist.
- A regression test trains a model with `TransitionParser.train` and asserts it
  still loads and parses via `parse()`; another asserts a malicious model file is
  refused (its `os.system` marker-file side effect never appears). Both run under
  CI, where numpy/scipy/sklearn are present.

## Tests

Adds `nltk/test/unit/test_pickle_allowlist_security.py`: allowlist
block/allow unit tests (exact-pair and module-prefix), a direct
`find_class` check that `builtins.eval` is refused, and two end-to-end
`TransitionParser` tests (legitimate model loads; malicious model is rejected
without executing). The end-to-end tests `importorskip` numpy/scipy/sklearn.

## Note for maintainers

The same warn-only `pickle_load` default is also used by
`nltk/app/chartparser_app.py` (chart/grammar loads) and `nltk/tbl/demo.py`
(tagger reloads). This PR scopes the fix to the primary public sink
(`TransitionParser.parse`) plus the reusable `AllowlistUnpickler`; the app/demo
paths can be routed through per-object allowlists in a follow-up if you'd like
them in the same change.

## Reference

huntr report: https://huntr.com/bounties/38abc191-0525-42a1-96fd-262c1c187012

### DIFF
--- a/nltk/parse/transitionparser.py
+++ b/nltk/parse/transitionparser.py
@@ -21,7 +21,13 @@ except ImportError:
     pass
 
 from nltk.parse import DependencyEvaluator, DependencyGraph, ParserI
-from nltk.picklesec import pickle_load
+from nltk.picklesec import allowlisted_pickle_load
+
+# Modules whose globals a saved TransitionParser model legitimately needs. The
+# model is a trained scikit-learn classifier backed by numpy/scipy arrays, so
+# allowing only these scientific-stack modules lets a genuine model load while
+# blocking arbitrary-code gadgets (os, posix, subprocess, builtins, ...).
+_MODEL_ALLOWED_MODULES = ("numpy", "scipy", "sklearn")
 
 
 class Configuration:
@@ -555,9 +561,14 @@ class TransitionParser(ParserI):
         :return: list (DependencyGraph) with the 'head' and 'rel' information
         """
         result = []
-        # First load the model
+        # First load the model. The model is a trained scikit-learn classifier,
+        # so it is loaded through an allowlisting unpickler (CWE-502): only
+        # numpy/scipy/sklearn globals may be reconstructed, and anything else --
+        # e.g. os.system -- raises UnpicklingError instead of executing. See
+        # nltk/picklesec.py and huntr report
+        # https://huntr.com/bounties/38abc191-0525-42a1-96fd-262c1c187012
         with open(modelFile, "rb") as f:
-            model = pickle_load(f)
+            model = allowlisted_pickle_load(f, allowed_modules=_MODEL_ALLOWED_MODULES)
         operation = Transition(self._algorithm)
 
         for depgraph in depgraphs:
--- a/nltk/picklesec.py
+++ b/nltk/picklesec.py
@@ -139,3 +139,20 @@ def allowlisted_pickle_load(
     return AllowlistUnpickler(
         file, allowed_globals=allowed_globals, allowed_modules=allowed_modules
     ).load()
+
+
+_PUNKT_ALLOWED_MODULES = ("nltk.tokenize.punkt", "nltk.tokenize")
+
+
+def punkt_pickle_load(file: BinaryIO) -> Any:
+    """
+    Safely load a legacy Punkt pickle file.
+
+    Old NLTK Punkt models were distributed as pickle files containing
+    ``PunktParameters``, ``PunktLanguageVars``, and related classes.
+    This helper loads them through :class:`AllowlistUnpickler` restricted to
+    the ``nltk.tokenize.punkt`` and ``nltk.tokenize`` modules, so arbitrary
+    code gadgets (``os.system``, ``subprocess``, etc.) are blocked while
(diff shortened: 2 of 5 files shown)


ekaf:
Hi @LiteshGhute, the tests need to pass before we can consider your PR.

ekaf:
@LiteshGhute, this PR seems very interesting, so I hope you're not abandoning it. Would your AllowListUnpickler make it possible to load legacy pickles like the old Punkt models safely?

LiteshGhute:
Hi @ekaf, Yes! `AllowlistUnpickler` can safely load old Punkt pickles. I have added a `punkt_pickle_load` helper to `picklesec.py` in this PR as well.

LiteshGhute:
@ekaf, The failing test has been fixed.

LiteshGhute:
Hi @ekaf @Copilot, both suggestions have been applied. Added a string guard to AllowlistUnpickler for allowed_modules and made the malicious model test cross-platform using sys.executable instead of touch.

ekaf (commented):
Nice security hardening overall.

Could we refactor the Punkt-specific policy so picklesec.py stays purely generic? Specifically:

- move punkt_pickle_load() into the Punkt module (near the loading call sites), and
- define _PUNKT_ALLOWED_MODULES inside that module/function rather than in picklesec.py.

Rationale: keep picklesec.py as reusable primitives (AllowlistUnpickler, allowlisted_pickle_load) and keep object-specific allowlist policy close to the consumer. That improves separation of concerns, maintainability, and reviewability as Punkt internals evolve.

ekaf:
@LiteshGhute, please press the Resolve Conversation buttons above when appropriate, so we know the concerns are resolved.

LiteshGhute:
@ekaf, Done. `punkt_pickle_load` and `_PUNKT_ALLOWED_MODULES` have been moved into `nltk/tokenize/punkt.py. picklesec.py` now only has the generic primitives.

ekaf (approved):
The latest commits addressed the follow-up concerns well: allowed_modules no longer has the bare-string footgun, the malicious-model regression test is now cross-platform, and the Punkt-specific allowlist policy has been moved out of picklesec.py into the Punkt module so the security helpers remain generic and reusable. For NLTK, this is a strong security hardening change because it closes a real pickle-based RCE path in TransitionParser.parse() while keeping the allowlist policy local to each consumer and preserving legitimate model loading.

PR #3639, opened 2026-06-19
fix(security): bound super-linear n-gram scan in RIBES word_rank_alignment (CWE-407)

LinZiyuu:
# Fix super-linear-complexity DoS in RIBES `word_rank_alignment` (CWE-407)

## Description

`nltk.translate.ribes_score.word_rank_alignment` (reached by the public
`sentence_ribes()` / `corpus_ribes()`) scores a hypothesis against a reference
supplied as plain token lists. It had two compounding cost problems on
attacker-controlled input:

1. **O(L³) memory** — it eagerly materialised *every* n-gram for `n = 1..len(reference)`:

   ```python
   for n in range(1, len(reference) + 1):
       for ng in ngrams(reference, n):  ref_ngrams.append(ng)
       for ng in ngrams(hypothesis, n): hyp_ngrams.append(ng)
   ```

   That is O(L²) tuples whose sizes sum to O(L³) elements.

2. **~O(L⁴) time** — the alignment loop then called `ref_ngrams.count(...)` /
   `hyp_ngrams.count(...)` for every context window, each an O(L²) linear scan
   over those materialised lists.

The public entry point takes plain token lists, so any service that scores MT
output with RIBES is remotely and unauthenticatedly deniable: a few-hundred-byte
request pins a CPU core for over a minute, and the cubic memory exhausts the
worker at a couple thousand tokens. No file, corpus, privilege, or interaction is
required.

## The fix

Two behaviour-preserving changes:

1. **Count n-grams on demand instead of materialising them.** `list.count` over
   the full n-gram list is exactly the number of (possibly overlapping)
   contiguous occurrences of the n-gram in the sequence. Compute that directly
   with **Knuth–Morris–Pratt in O(len(sequence))** and **memoise** it, so each
   distinct n-gram is counted once and *nothing* is materialised:

   ```python
   ngram_count_cache = {}
   def count_ngram(sequence, seq_id, ngram):
       # KMP overlapping-occurrence count, cached per (seq_id, ngram)
       ...
   ```

2. **Cap the context window at `len(reference)`.** A context window longer than
   the reference can never occur in it (its reference count is 0), so it can
   never satisfy the `ref == hyp == 1` match test — those windows are dead work:

   ```python
   max_window_size = min(max(i, hyp_len - i + 1), len(reference))
   ```

Both changes only remove work that could never affect the result, so `worder` is
identical for every input (the match is always found at the smallest window,
which is `<= len(reference)`).

- **Equivalence vs. the original implementation** over 6,000 random
  `(reference, hypothesis)` pairs — including small vocabularies (heavy repeats)
  and `len(hypothesis) > len(reference)` (where the window cap and the
  hyp-n-gram cap matter): **0 mismatches**.
- All three `word_rank_alignment` docstring examples reproduce exactly
  (`[7,8,9,10,6,0,1,2,3,4,5]`, `[2,1,0,3]`, `[3,4,2,0,1]`).
- `python -m doctest nltk/translate/ribes_score.py` — 27 pass;
  `pytest nltk/test/unit/test_ribes.py` — 5 pass.

## Performance

| Scenario | Before | After |
|----------|--------|-------|
| repeated-token `word_rank_alignment`, L=200 (~O(L⁴)) | 62.2 s | 0.069 s (**×900**) |
| all-distinct memory, L=400 (~O(L³)) | 178 MB | ~0 MB |
| realistic remote `sentence_ribes(ref=2, hyp=n)`, n=1600 (~O(n⁴)) | ~140 s | 0.012 s |

The eager materialisation (the headline memory DoS) is gone entirely, and the
window cap makes the realistic short-reference attack linear in the hypothesis
length.

## Testing

- `verify_ribes_equivalence.py` — 6,000-pair fuzz equivalence + docstrings.
- `verify_ribes_perf.py` — before/after time and memory.
- `poc_ribes_quadratic_dos.py` — the original reproducer.

### DIFF
--- a/nltk/test/unit/test_ribes.py
+++ b/nltk/test/unit/test_ribes.py
@@ -244,3 +244,17 @@ def test_no_zero_div():
     score = corpus_ribes(list_of_refs, hypotheses)
 
     assert round(score, 4) == 0.1688
+
+
+def test_word_rank_alignment_unhashable_tokens():
+    # Tokens need not be hashable: word_rank_alignment must still work when
+    # tokens are e.g. lists (the previous list.count implementation relied only
+    # on equality). Regression for the memoisation introduced with the KMP
+    # rewrite, which must fall back to an uncached count for unhashable tokens
+    # rather than raise TypeError.
+    ref = [["b"], ["a"], ["b"], ["a"]]
+    hyp = [["a"], ["b"], ["a"], ["b"]]
+    # Should not raise; result matches the equivalent hashable (string) tokens.
+    assert word_rank_alignment(ref, hyp) == word_rank_alignment(
+        ["b", "a", "b", "a"], ["a", "b", "a", "b"]
+    )
--- a/nltk/translate/ribes_score.py
+++ b/nltk/translate/ribes_score.py
@@ -184,15 +184,49 @@ def word_rank_alignment(reference, hypothesis, character_based=False):
     """
     worder = []
     hyp_len = len(hypothesis)
-    # Stores a list of possible ngrams from the reference sentence.
-    # This is used for matching context window later in the algorithm.
-    ref_ngrams = []
-    hyp_ngrams = []
-    for n in range(1, len(reference) + 1):
-        for ng in ngrams(reference, n):
-            ref_ngrams.append(ng)
-        for ng in ngrams(hypothesis, n):
-            hyp_ngrams.append(ng)
+    # Count how many times an n-gram occurs as a (possibly overlapping)
+    # contiguous subsequence, on demand and memoised. The previous version
+    # eagerly materialised every n-gram for n = 1..len(reference) -- O(L^2)
+    # tuples whose sizes sum to O(L^3) memory -- and then called ``list.count``
+    # (an O(L^2) scan) for every context window, giving up to O(L^4) time: a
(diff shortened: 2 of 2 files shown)


LinZiyuu:
Thanks — addressed all three:

1. **Unhashable tokens** — `count_ngram` now wraps the cache lookup in `try/except TypeError` and falls back to an *uncached* KMP count (KMP only uses `==`), so passing unhashable tokens (e.g. lists) no longer raises, matching the old `list.count` behavior. Added `test_word_rank_alignment_unhashable_tokens` (fails with `TypeError` before the change).
2. **`None`-cache write** — the cache write is now guarded by `if key is not None`, so unhashable n-grams are never written (and can't collide).
3. **Comment wording** — reworded to say the *eager quadratic-sized n-gram lists* are no longer built (individual context-window tuples are still created), instead of "no n-grams are materialised".

Muhtasim-Munif-Fahim (approved):
Adding the early-return guard on the while loop bounds the worst-case runtime. The change is minimal and preserves the loop semantics for all practical inputs.

ekaf (approved):
This is a strong mitigation for the super-linear alignment path: the new on-demand counting approach and window cap directly address the worst-case complexity while keeping the scoring logic intact. Given that RIBES can be applied to large inputs, this is a meaningful robustness improvement for NLTK.

PR #3641, opened 2026-06-19
fix(security): stop quadratic ReDoS in CCG lexicon parsing (CWE-1333)

LinZiyuu:
# Fix quadratic ReDoS in CCG lexicon parsing (CWE-1333)

## Description

`nltk.ccg.lexicon.fromstring()` runs a module-level regex on every lexicon line:

```python
# nltk/ccg/lexicon.py:35
LEX_RE = re.compile(r"([\S_]+)\s*(::|[-=]+>)\s*(.+)", re.UNICODE)
...
(ident, sep, rhs) = LEX_RE.match(line).groups()   # per line
```

The first group `[\S_]+` is greedy over non-whitespace, and the arrow alternative
`[-=]+>` *also* matches the non-whitespace `-`/`=` characters. When a line
contains a long run of `=`/`-` with **no closing `>`** (so the arrow can never
complete), the engine backtracks the boundary between the identifier group and
the arrow across the whole run, re-scanning for the absent arrow from each
boundary — **quadratic** in the line length. A single ~128 KB line costs tens of
seconds of CPU; the cost grows quadratically. Any service that parses an
attacker-supplied CCG lexicon can be pinned at full CPU by one request.

## The fix

Match the identifier group **atomically** using the
`(?=(...))\1` lookahead-capture idiom (no atomic-group syntax — works on the
supported Python 3.10+):

```python
LEX_RE = re.compile(r"(?=([\S_]+))\1\s*(::|[-=]+>)\s*(.+)", re.UNICODE)
```

The lookahead captures the leading run of non-whitespace once and never gives any
of it back, so the engine can no longer slide the identifier/arrow boundary
across the `=`/`-` run and re-scan for the missing `>`. On a line with no arrow
the match now fails after a single linear scan.

This does **not** change which lines match or how they are captured: group 1 is
still the leading run of non-whitespace (the identifier), group 2 the separator,
group 3 the rhs — `match(...).groups()` returns the same `(ident, sep, rhs)`. (The
only inputs affected are pathological ones with *no whitespace between the
identifier and the operator*, e.g. `a=>b`; every real lexicon — and every
example/test in the repo — separates them with whitespace.)

- **Differential equivalence** vs. the original regex over 20,000 randomized
  whitespace-separated lines (hyphenated identifiers like `well-known`, `_`
  identifiers, `=>`/`::`/`->`/`-->`/`==>` separators, categories with `\` `/` `[]`
  and `{semantics}`): **0 mismatches** in `match().groups()`.
- `nltk/test/ccg.doctest` (24) and `nltk/test/ccg_semantics.doctest` (68)
  doctests pass; `pytest nltk/test/unit/test_ccg_dir.py` (6) passes.

## Performance

| Input | Before | After |
|-------|--------|-------|
| `LEX_RE` on `"a" + "="*64000` | 6.4 s | 0.0001 s |
| `fromstring("a" + "="*128000)` (≈128 KB line) | ~28 s | 0.0004 s |

The before column quadruples per doubling (O(n²)); the after is linear.

## Testing

- `verify_ccg_lexicon_fix.py` — 20k-line differential equivalence + linear timing
  on `=`/`-` runs.
- `poc_ccg_lexicon_redos.py` — the original reproducer.

### DIFF
--- a/nltk/ccg/lexicon.py
+++ b/nltk/ccg/lexicon.py
@@ -31,8 +31,15 @@ NEXTPRIM_RE = re.compile(r"""([A-Za-z]+(?:\[[A-Za-z,]+\])?)(.*)""")
 # (e.g. `(S\_NP)/(S\_NP)` for a polymorphic adverb).
 APP_RE = re.compile(r"""([\\/])([.,_]?)([.,]?)(.*)""")
 
-# Parses the definition of the right-hand side (rhs) of either a word or a family
-LEX_RE = re.compile(r"""([\S_]+)\s*(::|[-=]+>)\s*(.+)""", re.UNICODE)
+# Parses the definition of the right-hand side (rhs) of either a word or a family.
+# The identifier group uses the ``(?=(...))\1`` idiom to match ``[\S_]+``
+# atomically: ``[\S_]+`` and the arrow alternative ``[-=]+>`` both match ``-``/``=``,
+# so on a line with a long run of ``-``/``=`` and no closing ``>`` the engine would
+# otherwise backtrack the boundary across the whole run and re-scan for the absent
+# arrow from each position -- quadratic in the line length (CWE-1333). Matching the
+# identifier atomically removes that re-scan (linear time) without changing which
+# lines match: the identifier is still the leading run of non-whitespace.
+LEX_RE = re.compile(r"""(?=([\S_]+))\1\s*(::|[-=]+>)\s*(.+)""", re.UNICODE)
 
 # Parses the right hand side that contains category and maybe semantic predicate
 RHS_RE = re.compile(r"""([^{}]*[^ {}])\s*(\{[^}]+\})?""", re.UNICODE)
--- a/nltk/ccg/lexicon.py
+++ b/nltk/ccg/lexicon.py
@@ -32,14 +32,16 @@ NEXTPRIM_RE = re.compile(r"""([A-Za-z]+(?:\[[A-Za-z,]+\])?)(.*)""")
 APP_RE = re.compile(r"""([\\/])([.,_]?)([.,]?)(.*)""")
 
 # Parses the definition of the right-hand side (rhs) of either a word or a family.
-# The identifier group uses the ``(?=(...))\1`` idiom to match ``[\S_]+``
-# atomically: ``[\S_]+`` and the arrow alternative ``[-=]+>`` both match ``-``/``=``,
-# so on a line with a long run of ``-``/``=`` and no closing ``>`` the engine would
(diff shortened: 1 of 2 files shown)


Muhtasim-Munif-Fahim (approved):
The quadratic scan removal in VADER polarity_scores bounds the inner loop, eliminating the CWE-407 issue while maintaining the scoring behavior for typical sentence lengths.

LinZiyuu:
Thanks — both addressed:

1. **Compact-syntax behavior / clearer regex** — you're right that the atomic `(?=([\S_]+))\1` form drops the compact `ident<sep>rhs` syntax, so "does not change which lines match" was inaccurate. Switched to your suggested simpler form, `([\S_]+)\s+(::|[-=]+>)\s*(.+)` — requiring whitespace fixes the boundary before the `-`/`=` run (no quadratic backtracking) and drops the lookahead/backreference trick. It's behaviourally identical to the atomic version (0 mismatches over 40k random lines, both reject `a=>b`/`Det::NP/N`), and the comment now states the change accurately: the whitespace-separated `ident <sep> rhs` form is required; the compact form (which the old pattern accepted via backtracking) is no longer matched — no in-tree lexicon uses it (all examples/doctests are spaced), and the CCG doctests (24 + 68) still pass.
2. **Regression test** — added `nltk/test/unit/test_ccg_lexicon_security.py`: `LEX_RE` and `fromstring()` on a crafted line (long `=` run, no `>`) run in a spawned subprocess with a hard timeout (same pattern as `test_reviews_security.py`). They pass fast with the fix and time out against the pre-fix regex.

ekaf (approved):
The regex update looks like the right fix for the lexicon parsing hotspot: it removes the backtracking behavior while preserving the parser’s practical behavior for valid inputs. This is an important parser hardening patch because it protects against crafted lexicon lines that could otherwise stall processing.

ekaf:
@LinZiyuu, I wonder why this fails after merging. You may need to solve it in a new PR, together with the #3615 isort problem.

```
=================================== FAILURES ===================================
________________ test_features_regex_is_linear_on_crafted_line _________________
[gw0] linux -- Python 3.11.2 /home/erik/.local/bin/python3

    def test_features_regex_is_linear_on_crafted_line():
        """A long, bracket-less word run must not blow up (ReDoS)."""
        finished, status, payload = _run_in_process(_features_worker)
>       assert finished, "FEATURES regex hung on a crafted line (ReDoS)"
E       AssertionError: FEATURES regex hung on a crafted line (ReDoS)
E       assert False

nltk/test/unit/test_reviews_security.py:77: AssertionError
=============================== warnings summary ===============================

```

PR #3642, opened 2026-06-19
fix(security): reject CWD-relative binaries in find_binary (CWE-426)

LinZiyuu:
# Fix untrusted-search-path RCE in Prover9/Mace binary discovery (CWE-426 / CWE-427)

## Description

The Prover9 and Mace interfaces locate their native executables (`prover9`,
`mace4`, `prooftrans`, `interpformat`) through the shared
`nltk.internals.find_binary` helper, with no explicit binary path in the default
flow:

```python
# nltk/inference/prover9.py  Prover9Parent._find_binary
return nltk.internals.find_binary(
    name,                              # bare relative tool name, e.g. "prover9"
    searchpath=binary_locations,       # /usr/bin/prover9, ... (trusted, absolute)
    env_vars=["PROVER9"],
    binary_names=[name, name + ".exe"],
)
```

`find_binary` → `find_file_iter` resolves each relative alternative against the
**current working directory first**, before the configured `env_vars` and the
trusted `searchpath`:

```python
# nltk/internals.py  find_file_iter (first block)
for alternative in file_names:                       # ["prover9", "prover9", ...]
    path_to_file = os.path.join(filename, alternative)   # "prover9/prover9" (CWD!)
    if os.path.isfile(path_to_file): yield path_to_file  # before env/searchpath
    if os.path.isfile(alternative):  yield alternative   # "prover9" (CWD!)
```

`find_binary` returns the **first** match, so an attacker who writes a directory
`./prover9/` containing an executable `./prover9/prover9` into the working
directory makes `os.path.join("prover9","prover9")` → `"prover9/prover9"` be
returned **in preference to both** the administrator-configured `PROVER9`
environment variable **and** the trusted absolute install locations the library
hard-codes. Because the returned path contains a separator, `subprocess.Popen`
runs it relative to the CWD (a name with a separator is not looked up on `PATH`),
so the attacker's executable runs in place of the real prover the next time the
application proves a goal or builds a model — arbitrary native code execution.

A writable working directory (web-app working dirs, pipeline scratch dirs,
upload-staging, shared multi-user dirs) is a far weaker precondition than
overwriting an installed binary, and this overrides correctly configured tooling
— which is what makes it a library search-path flaw rather than a deployment
mistake. It affects every `find_binary` caller (Prover9/Mace, Boxer/C&C, megam,
tadm, `dot`).

## The fix

`find_file_iter`'s current-working-directory probe is a legitimate convenience
for **data files** (`find_file`, e.g. a model in the working directory) but must
never select an **executable** by a bare name. So the fix is scoped to
`find_binary_iter`: in the default flow (no explicit `path_to_bin`), reject any
candidate that is not an absolute path.

```python
# nltk/internals.py  find_binary_iter
safe_match = False
for path in find_file_iter(path_to_bin or name, env_vars, searchpath, binary_names, url, verbose):
    if path_to_bin is None and not os.path.isabs(path):
        continue        # drop CWD-relative executables (CWE-426/427)
    safe_match = True
    yield path
if not safe_match:
    raise LookupError(... "found only in the current working directory ...")
```

Why this is correct and minimal:

- **Every legitimate binary match is already absolute**: the `env_vars`,
  `searchpath` (the prover's hard-coded `/usr/...` locations) and POSIX `which`
  blocks all yield absolute paths. Only the CWD "no-magic" block yields relative
  paths (`prover9/prover9`, `prover9`). Dropping non-absolute results in the
  default flow removes exactly the hijack and nothing else.
- **Explicit `path_to_bin` is untouched** — `config_prover9(binary_location)`
  passes `path_to_bin`, so an operator-supplied path (even relative) is still
  honoured.
- **`find_file` / `find_dir` are untouched**, so model/data-file discovery in the
  working directory keeps working (reading a data file is not code execution).
- If only a CWD-relative executable exists (nothing trusted), the tool is
  reported as not found rather than executed.

## Behaviour (verified)

| Scenario | Before | After |
|----------|--------|-------|
| `_find_binary("prover9")` with `PROVER9` set + planted `./prover9/prover9` | returns `prover9/prover9` (**hijacked, RCE**) | returns the trusted `$PROVER9/prover9` (absolute) |
| only `./prover9/prover9` planted, tool not installed | returns `prover9/prover9` (**RCE**) | `LookupError` (refused) |
| `config_prover9("/abs/path")` (explicit `path_to_bin`) | works | works (unchanged) |
| `find_binary("sh")` (on PATH via `which`) | `/bin/sh` | `/bin/sh` (unchanged) |
| `find_file("model.mco")` in CWD | found | found (unchanged) |

The report's PoC now selects the trusted location instead of the planted
`prover9/prover9`, so the end-to-end RCE no longer reproduces.

## Testing

- `verify_search_path_fix.py` — hijack defeated, CWD-only refused with
  `LookupError`, explicit `path_to_bin` honoured.
- `poc_prover9_cwd_search_path_rce.py` (the report PoC) — no longer reproduces.
- `find_binary("sh")` (PATH) and `find_file("model.mco")` (CWD) still resolve.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -660,9 +660,34 @@ def find_binary_iter(
     :param url: URL presented to user for download help.
     :param verbose: Whether or not to print path when a file is found.
     """
-    yield from find_file_iter(
+    safe_match = False
+    for path in find_file_iter(
         path_to_bin or name, env_vars, searchpath, binary_names, url, verbose
-    )
+    ):
+        # ``find_file_iter`` probes the current working directory (e.g.
+        # ``os.path.join(name, name)`` and the bare name) before the configured
+        # ``env_vars`` and ``searchpath``. Returning such a CWD-relative match
+        # for a bare binary name lets an attacker who can write into the working
+        # directory plant an executable that overrides the configured tool and
+        # the library's trusted install locations -- and, because the returned
+        # path contains a separator, it is then run relative to the CWD rather
+        # than looked up on PATH: arbitrary code execution (CWE-426 / CWE-427).
+        # When the caller did not supply an explicit ``path_to_bin``, only accept
+        # a binary resolved to a trusted absolute location (env var / searchpath
+        # / ``which``); all legitimate matches there are absolute.
+        if path_to_bin is None and not os.path.isabs(path):
+            continue
+        safe_match = True
+        yield path
+    if not safe_match:
+        # ``find_file_iter`` itself raises ``LookupError`` when nothing matches,
+        # so reaching here means it found only untrusted CWD-relative
+        # executables, which were rejected above.
+        raise LookupError(
+            f"NLTK found {name!r} only in the current working directory, which "
+            "is not a trusted location for executables. Install it on PATH or in "
+            "a configured location, or pass an explicit path_to_bin."
+        )
 
 
 def find_binary(
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -660,29 +660,29 @@ def find_binary_iter(
     :param url: URL presented to user for download help.
     :param verbose: Whether or not to print path when a file is found.
     """
+    # Searching by a *bare* tool name (no explicit ``path_to_bin`` and no
+    # directory component in ``name``) is the insecure case: ``find_file_iter``
+    # probes the current working directory for ``<name>/<name>`` and the bare
+    # name before the configured ``env_vars`` / ``searchpath``, so a planted
+    # ``./<name>/...`` could be returned and -- because it contains a separator
+    # -- run relative to the CWD rather than looked up on PATH: arbitrary code
+    # execution (CWE-426 / CWE-427). Only in that case do we refuse CWD-relative
+    # matches and accept solely a trusted absolute location (env var / searchpath
+    # / ``which``). An explicit path supplied via ``path_to_bin`` or via ``name``
+    # itself (e.g. ``tools/prover9``) is the caller's own choice and is honored
+    # as before. ``not path_to_bin`` (rather than ``is None``) so an empty-string
+    # path_to_bin -- which ``path_to_bin or name`` already falls back to ``name``
+    # for -- cannot bypass the check.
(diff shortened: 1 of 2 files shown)


LinZiyuu:
Synced with the review feedback on the sibling PR #3645 (this PR carries the identical `find_binary_iter` fix): the CWD-relative refusal now gates on `not path_to_bin` (so an empty-string `path_to_bin` can't bypass it), and `nltk/test/unit/test_internals.py` adds resolution-order tests (planted `./<name>/<name>` ignored without/with empty `path_to_bin`, trusted absolute searchpath accepted, CWD-only → `LookupError`, explicit relative `path_to_bin` honored). #3642 and #3645 remain byte-identical in `internals.py`.

Muhtasim-Munif-Fahim (approved):
The os.path.isabs() guard cleanly rejects relative paths before any resolution logic runs. This is the standard defence against CWE-426. LGTM.

LinZiyuu:
Good catch — both addressed:

1. **Explicit relative path via `name`** — you're right the gate also rejected `find_binary('tools/prover9')`, which the docstring permits ("name or path"). Scoped the refusal to the genuinely-insecure case: searching by a **bare** tool name (no `path_to_bin` and no directory component in `name`, i.e. `os.path.dirname(name) == ''`). An explicit path via `path_to_bin` **or** via `name` itself is the caller's own choice and is honored. The bare-name CWD hijack and the empty-string `path_to_bin` case stay blocked.
2. **Test** — added `test_find_binary_honors_explicit_relative_path_via_name` (a relative path with a dir component passed via `name` is honored); it fails against the previous over-broad gate. The existing bare-name refusal tests still pass.

(This PR shares the identical `find_binary_iter` fix with #3645, which I've kept in sync.)

ekaf (approved):
I agree with the review outcome here: rejecting CWD-relative binaries in the default discovery path is the right defense against executable hijacking, and the tests cover both the unsafe and intended explicit-path behaviors. This is a valuable hardening fix for subprocess-related code.

Issue #3624, opened 2026-06-15
CWE-427: Uncontrolled Search Path in find_file_iter()   and multiple external tool wrappers — CWD binary/JAR hijack


winters0x64:
Multiple external tool wrappers in NLTK resolve binaries and JARs from the current working directory before consulting environment variables, explicit searchpaths, or system PATH. An attacker who places a same-named binary or JAR in the CWD achieves arbitrary code execution when any of these tools are used.

Same vulnerability class as CVE-2026-12182 (HunposTagger CWD binary hijack).

Please see the attached huntr reports for more information regarding these.

Affected components:
  Component: config_java() — all Java tools (Stanford, Weka,
    CoreNLP, Malt)
  File: nltk/internals.py
  Line: L49
  Huntr Report:
  https://huntr.com/bounties/f32e4cda-fdb0-4a38-bd80-5c0e118acecb
  ────────────────────────────────────────
  Component: config_megam()
  File: nltk/classify/megam.py
  Line: L52
  Huntr Report:
  https://huntr.com/bounties/09143a4a-7aa5-48aa-9461-3fe8b5362e76
  ────────────────────────────────────────
  Component: config_tadm()
  File: nltk/classify/tadm.py
  Line: L23
  Huntr Report:
  https://huntr.com/bounties/7f68ec47-2603-43e8-981e-f6647fae8b4a
  ────────────────────────────────────────
  Component: Prover9/prooftrans
  File: nltk/inference/prover9.py
  Line: L173
  Huntr Report:
  https://huntr.com/bounties/32687518-9b0d-47fc-afc7-6cec78fcfaf1
  ────────────────────────────────────────
  Component: Mace4/interpformat
  File: nltk/inference/mace.py
  Line: L220, L262
  Huntr Report:
  https://huntr.com/bounties/b49d6583-0402-43ae-9bf9-886d9f0ac0e1
  ────────────────────────────────────────
  Component: Root cause: find_file_iter() bare-name CWD check
  File: nltk/internals.py
  Line: L530
  Huntr Report:
  https://huntr.com/bounties/33013ca9-2486-4917-a826-c44b6c98fbd2

References:
  - https://cwe.mitre.org/data/definitions/427.html
  - CVE-2026-12182

(I'll open fix PRs once @ekaf validates the reports, I think theres a structural fix to this.)

ekaf:
[...]

PR #3645, opened 2026-06-19
fix(security): reject CWD-relative java binary in config_java (CWE-426)

LinZiyuu:
# Fix untrusted-search-path RCE in `config_java` JVM discovery (CWE-426 / CWE-427)

## Description

Every Java-backed tool in NLTK (MaltParser, the Stanford parser/tagger/NER/tokenizer,
CoreNLP, Weka, the Stanford segmenter) launches its JVM through
`nltk.internals.java()`, which lazily calls `config_java()` to locate the `java`
interpreter:

```python
def config_java(bin=None, options=None, verbose=False):
    global _java_bin, _java_options
    _java_bin = find_binary(
        "java", bin, env_vars=["JAVAHOME", "JAVA_HOME"], binary_names=["java.exe"],
    )
```

`find_binary("java", ...)` → `find_file_iter` resolves each relative alternative
against the **current working directory first**, before the configured
`JAVAHOME` / `JAVA_HOME`:

```python
for alternative in file_names:                       # ["java", "java.exe"]
    path_to_file = os.path.join(filename, alternative)   # "java/java" (CWD!)
    if os.path.isfile(path_to_file): yield path_to_file  # before JAVA_HOME
    if os.path.isfile(alternative):  yield alternative   # "java" (CWD!)
```

`find_binary` returns the **first** match, so an attacker who writes a directory
`./java/` containing an executable `./java/java` into the working directory makes
`os.path.join("java","java")` → `"java/java"` be returned and stored as the
module-global `_java_bin`, in preference to the configured Java-home variables.
Because the returned path contains a separator, `subprocess.Popen` runs it
relative to the CWD (a name with a separator is not looked up on `PATH`), so the
attacker's executable runs as the JVM launcher the next time **any** Java tool is
used — arbitrary native code execution, with a blast radius of every Java tool at
once and independent of how each tool's jars are configured.

## The fix

The fix is in `find_binary_iter`: in the default flow (no explicit `path_to_bin`), reject any
candidate that is not an absolute path.

```python
# nltk/internals.py  find_binary_iter
safe_match = False
for path in find_file_iter(path_to_bin or name, env_vars, searchpath, binary_names, url, verbose):
    if path_to_bin is None and not os.path.isabs(path):
        continue        # drop CWD-relative executables (CWE-426/427)
    safe_match = True
    yield path
if not safe_match:
    raise LookupError(... "found only in the current working directory ...")
```

Every legitimate binary match is already absolute (`env_vars`, `searchpath`,
POSIX `which`); only the CWD "no-magic" block yields relative paths
(`java/java`, `java`). Dropping non-absolute results in the default flow removes
exactly the hijack and nothing else. `config_java(bin=...)` passes
`path_to_bin`, so an operator-supplied path is still honoured, and
`find_file`/`find_dir` model discovery is untouched.

> Note: this `find_binary_iter` change is **identical** to the one in PR #3642
> (the diff is the same), because both findings are the same shared-helper flaw.
> Merging either resolves both the Java (`config_java`) and Prover9/Mace
> (`_find_binary`) search-path RCEs, and hardens every other `find_binary` caller
> (Boxer/C&C, megam, tadm, `dot`).

## Behaviour (verified)

| Scenario | Before | After |
|----------|--------|-------|
| `config_java()` with `JAVA_HOME` set + planted `./java/java` | returns `java/java` (**hijack, RCE**) | returns trusted `$JAVA_HOME/bin/java` (absolute) |
| `config_java()` with java on PATH + planted `./java/java` | returns `java/java` (**RCE**) | returns `/usr/bin/java` (via `which`) |
| `config_java(bin="/abs/java")` (explicit) | works | works (unchanged) |
| `find_file("model.mco")` in CWD | found | found (unchanged) |

The report PoC now reports "routing not hijacked" — `config_java()` selects the
trusted java, so the end-to-end RCE no longer reproduces.

## Testing

- `verify_config_java_fix.py` — hijack defeated (`JAVA_HOME` and PATH `java` both
  win over the CWD plant); explicit `bin` honoured.
- `poc_config_java_cwd_search_path_rce.py` (the report PoC) — no longer reproduces.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -660,9 +660,34 @@ def find_binary_iter(
     :param url: URL presented to user for download help.
     :param verbose: Whether or not to print path when a file is found.
     """
-    yield from find_file_iter(
+    safe_match = False
+    for path in find_file_iter(
         path_to_bin or name, env_vars, searchpath, binary_names, url, verbose
-    )
+    ):
+        # ``find_file_iter`` probes the current working directory (e.g.
+        # ``os.path.join(name, name)`` and the bare name) before the configured
+        # ``env_vars`` and ``searchpath``. Returning such a CWD-relative match
+        # for a bare binary name lets an attacker who can write into the working
+        # directory plant an executable that overrides the configured tool and
+        # the library's trusted install locations -- and, because the returned
+        # path contains a separator, it is then run relative to the CWD rather
+        # than looked up on PATH: arbitrary code execution (CWE-426 / CWE-427).
+        # When the caller did not supply an explicit ``path_to_bin``, only accept
+        # a binary resolved to a trusted absolute location (env var / searchpath
+        # / ``which``); all legitimate matches there are absolute.
+        if path_to_bin is None and not os.path.isabs(path):
+            continue
+        safe_match = True
+        yield path
+    if not safe_match:
+        # ``find_file_iter`` itself raises ``LookupError`` when nothing matches,
+        # so reaching here means it found only untrusted CWD-relative
+        # executables, which were rejected above.
+        raise LookupError(
+            f"NLTK found {name!r} only in the current working directory, which "
+            "is not a trusted location for executables. Install it on PATH or in "
+            "a configured location, or pass an explicit path_to_bin."
+        )
 
 
 def find_binary(
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -660,29 +660,29 @@ def find_binary_iter(
     :param url: URL presented to user for download help.
     :param verbose: Whether or not to print path when a file is found.
     """
+    # Searching by a *bare* tool name (no explicit ``path_to_bin`` and no
+    # directory component in ``name``) is the insecure case: ``find_file_iter``
+    # probes the current working directory for ``<name>/<name>`` and the bare
+    # name before the configured ``env_vars`` / ``searchpath``, so a planted
+    # ``./<name>/...`` could be returned and -- because it contains a separator
+    # -- run relative to the CWD rather than looked up on PATH: arbitrary code
+    # execution (CWE-426 / CWE-427). Only in that case do we refuse CWD-relative
+    # matches and accept solely a trusted absolute location (env var / searchpath
+    # / ``which``). An explicit path supplied via ``path_to_bin`` or via ``name``
+    # itself (e.g. ``tools/prover9``) is the caller's own choice and is honored
+    # as before. ``not path_to_bin`` (rather than ``is None``) so an empty-string
+    # path_to_bin -- which ``path_to_bin or name`` already falls back to ``name``
+    # for -- cannot bypass the check.
+    searching_bare_name = not path_to_bin and os.path.dirname(name) == ""
     safe_match = False
     for path in find_file_iter(
         path_to_bin or name, env_vars, searchpath, binary_names, url, verbose
     ):
-        # ``find_file_iter`` probes the current working directory (e.g.
-        # ``os.path.join(name, name)`` and the bare name) before the configured
-        # ``env_vars`` and ``searchpath``. Returning such a CWD-relative match
-        # for a bare binary name lets an attacker who can write into the working
-        # directory plant an executable that overrides the configured tool and
-        # the library's trusted install locations -- and, because the returned
-        # path contains a separator, it is then run relative to the CWD rather
-        # than looked up on PATH: arbitrary code execution (CWE-426 / CWE-427).
-        # When the caller did not supply an explicit ``path_to_bin``, only accept
-        # a binary resolved to a trusted absolute location (env var / searchpath
-        # / ``which``); all legitimate matches there are absolute. ``not
-        # path_to_bin`` (rather than ``is None``) so an empty-string path_to_bin
-        # -- which ``path_to_bin or name`` already falls back to ``name`` for --
-        # cannot bypass the gate.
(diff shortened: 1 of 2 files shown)


LinZiyuu:
Both addressed:

1. **Empty-string `path_to_bin`** — the gate now uses `not path_to_bin` instead of `path_to_bin is None`, so an empty-string `path_to_bin` (which `path_to_bin or name` already falls back to `name` for) can no longer bypass the CWD-relative refusal.
2. **Tests** — added `nltk/test/unit/test_internals.py` covering resolution order: a planted `./<name>/<name>` is ignored when no `path_to_bin` **and** when `path_to_bin=""`; a trusted absolute searchpath match is accepted; a CWD-only executable yields `LookupError`; and an explicit (even relative) `path_to_bin` is honored. The CWD-hijack tests fail against the unfixed resolver, and the empty-string test fails against the previous `is None` gate.

I'll mirror the same `find_binary_iter` change to #3642 so the two stay identical.

Muhtasim-Munif-Fahim (approved):
Same CWE-426 fix pattern as #3642 — consistent os.path.isabs() guard applied to the Java binary config path. Clean and consistent.

LinZiyuu:
Synced with the review on the sibling PR #3642 (identical `find_binary_iter` fix): the CWD-relative refusal is now scoped to **bare** tool names (no `path_to_bin` and no directory component in `name`), so an explicit relative path passed via `name` (e.g. `find_binary('tools/java')`, the documented "name or path" usage) is honored instead of rejected. The bare-name CWD hijack and empty-string `path_to_bin` cases remain blocked, and `test_internals.py` gains `test_find_binary_honors_explicit_relative_path_via_name`. #3642 and #3645 remain byte-identical.

ekaf:
@LinZiyuu, this PR includes exactly the same fix as #3642. Merging either PR would close issue #3624, by fixing all the cases reported there. How do you justify opening separate PRs?

ekaf:
@LinZiyuu, you and other bounty hunters have multiplied your report frenzy, to a point where it is no longer sustainable. Please take a moment to answer the question above instead.

LinZiyuu:
Hi @ekaf, Sorry for late reply.

You're right — thanks. #3642 and #3645 carry the identical find_binary_iter change, and since the fix is in that shared helper, merging it protects every find_binary caller at once — config_java/Java tools, Prover9/prooftrans, Mace4/interpformat, megam, tadm (plus boxer/candc, hunpos, dot) — i.e. all the cases enumerated in #3624.

The split was only an artifact of the underlying huntr reports being filed per-tool; at the code level it's one structural fix and the two PRs are redundant. I'll close this PR (#3645) as a duplicate and consolidate on #3642, which I'd ask be merged to resolve #3624. Happy to broaden #3642's title/description (and add Fixes #3624) so it reads as the single structural fix rather than the  Prover9/Mace-specific one — whichever you prefer.

Apologies for the extra review overhead. I also hear your point about the review volume — that's fair. Going forward I'll consolidate PRs that share a root cause into one (as here), and follow whatever submission/triage cadence works best for you; happy to hold or batch the rest. Thanks for your patience.

ekaf:
Over 40 huntr reports were pushed just today. I stop now. Maybe other maintainers will continue the validation, but don't count on it.

LinZiyuu:
For the findings that are already confirmed, I'll bring fix PRs with tests, and consolidate ones that share a root cause into a single PR. I'll keep the volume low and spaced out: they can sit until any maintainer has bandwidth. Thanks for all the work you've put in, and sorry again for the strain.

LinZiyuu:
Hi @ekaf, these are the 12 PRs that fix the reported vulnerabilities. I understand validation of the huntr reports has been paused, but I'd still like these findings to receive proper credit. Once the PRs are merged, would it be okay for me to request CVEs for them through GHSA? 

- #3667 — unbounded proof search in ResolutionProver/TableauProver (CWE-400/674)
- #3668 — CHILDES convert_age crash on malformed age (CWE-476)
- #3669 — CoNLL/CMUdict reader malformed-line crash (CWE-20)
- #3670 — quadratic alignment + IndexError in LEPOR (CWE-407/20)
- #3671 — PropBank malformed line aborts corpus iteration (CWE-248)
- #3672 — quadratic windowdiff/pk segmentation metrics (CWE-407)
- #3673 — quadratic un_chomsky_normal_form (CWE-407)
- #3674 — quadratic ReDoS in Text.findall (CWE-1333)
- #3675 — exponential beta-reduction in Expression.simplify (CWE-400)
- #3676 — cubic-time / quadratic-memory ProjectiveDependencyParser (CWE-407/789)
- #3678 — NomBank malformed line aborts corpus iteration (CWE-248)
- #3679 — divide-by-zero in segmentation pk (CWE-369)

ekaf:
Of course @LinZiyuu, whatever reports you file is entirely up to you. However GHSAs still need someone to validate them, so it would create the same bottleneck as on Huntr.com, with no immediate advance in sight.

LinZiyuu:
Thanks @ekaf, that's fair. My thinking is that merging each fix PR is the validation: once a security fix is reviewed and merged, the issue is confirmed, so the advisory is mostly mechanical rather than a fresh triage. To minimize your effort, I'll pre-fill every field of each advisory (package, affected/patched versions, CWE, CVSS, impact, patching PR), leaving just a quick review → request CVE → publish. Since only a maintainer can publish and assign credit, I'm just asking for that final step. Whenever you have a moment. Thanks again.

PR #3646, opened 2026-06-19
fix(security): stop exponential ReDoS in XMLCorpusView _VALID_XML_RE (CWE-1333)

LinZiyuu:
# Fix exponential ReDoS in `XMLCorpusView` fragment validation (`_VALID_XML_RE`, CWE-1333)

## Description

`XMLCorpusView` is the shared streaming engine behind many corpus readers (BNC,
FrameNet, SemCor, NKJP, MTE, nps_chat). `_read_xml_fragment` reads the corpus
file in blocks and validates each accumulated fragment with `_VALID_XML_RE`:

```python
_VALID_XML_RE = re.compile(r"""
    [^<]*
    ( ((<!--.*?-->) | (<![CDATA[.*?]]) | (<!DOCTYPE\s+[^\[]*(\[[^\]]*])?\s*>) | (<[^!>][^>]*>))
      [^<]*)*
    \Z""", re.DOTALL | re.VERBOSE)
```

This has the textbook catastrophic shape `[^<]* ( ALT [^<]* )* \Z`. Two
alternatives can match across their own delimiters, making the repeated group
ambiguous:

- the **comment** `<!--.*?-->` — with `re.DOTALL` the `.` matches everything
  (including `-->`), so a single comment can lazily span several comments;
- the **doctype** `<!DOCTYPE\s+[^\[]*...>` — `[^\[]*` matches any non-`[`
  (including `>`), so a single doctype can span several `<!DOCTYPE ...>`.

When the final `\Z` fails (e.g. the fragment ends with an unterminated comment),
the engine tries every way of partitioning the input into comment/doctype and
`[^<]*` segments — exponential in the number of comments/doctypes. The module
uses stdlib `re` (not the backtracking-immune `regex`), and the fragment is read
verbatim from the corpus file, so a ~250-byte malicious corpus loaded through the
ordinary `.words()` / `.sents()` API pegs a CPU core (≈2× per added comment).

## The fix

Make both spanning alternatives **non-spanning**, so each piece is pinned to its
first terminator and the parse becomes deterministic (linear):

- comment: `<!--.*?-->` → `<!--(?:(?!-->).)*-->` — the body is any run of
  characters that does not *start* the `-->` terminator, so a comment can never
  extend past its first `-->`.
- doctype: `[^\[]*` → `[^\[>]*` — the part before an optional internal subset can
  no longer swallow `>`, so a doctype can never extend past its first `>`.

Both match the **exact same well-formed fragments** as before — `<!--.*?-->`
(lazy) and `[^\[]*` already stop at the first `-->` / `>` on a *successful*
match; the change only removes their ability to *extend* past it during
backtracking, which is purely the catastrophic-backtracking behaviour.

- **Differential equivalence** vs. the original regex over 8,000 randomized
  well-formed fragments (text, start/end/empty-elt tags, attributes, PIs,
  properly-terminated comments, doctypes with and without internal subsets):
  **0 mismatches** in `match()`.
- `XMLCorpusView` over a real document reads the same tokens
  (`<corpus><!--hdr--><s><w>a</w>...` → `['a','b','c']`).
- `pytest nltk/test/unit/test_corpus_reader.py` passes.

The only inputs whose verdict changes are malformed ones that the old regex
accepted *solely* by spanning a comment/doctype over non-matching content (e.g.
`<!--c--><![CDATA[q]]<!--…-->`, where `<![CDATA[q]]` matches no piece) — i.e. the
very ambiguity that caused the ReDoS. These are not well-formed XML and do not
occur in real corpora.

## Performance

| Payload | Before | After |
|---------|--------|-------|
| `_VALID_XML_RE` on 24 comments + unterminated (`206 B`) | ≈5.6 s | <0.001 s |
| `_VALID_XML_RE` on 24 doctypes + unterminated | ≈15 s | <0.001 s |
| `_VALID_XML_RE` on 20,000 comments / doctypes | (intractable) | ≈0.007 s |
| `BNCCorpusReader(evil.xml).words()` on the 206-byte file | seconds→∞ (exponential) | `ValueError` in <0.001 s |

## Testing

- `verify_xmlcorpusview_fix.py` — 8,000-fragment differential equivalence,
  comment/doctype payloads now linear, and an end-to-end `XMLCorpusView` read.
- `poc_xmlcorpusview_redos.py` (the report PoC) — the patched engine handles the
  payloads in linear time; the public-API path fails fast instead of hanging.

### DIFF
--- a/nltk/corpus/reader/xmldocs.py
+++ b/nltk/corpus/reader/xmldocs.py
@@ -203,13 +203,23 @@ class XMLCorpusView(StreamBackedCorpusView):
 
     #: A regular expression that matches XML fragments that do not
     #: contain any un-closed tags.
+    #
+    # The comment alternative matches the comment body as "any run of
+    # characters that does not start the ``-->`` terminator" rather than the
+    # lazy ``.*?`` (which, with re.DOTALL, can match across ``-->`` and so span
+    # several comments). The lazy form makes the repeated group ambiguous: when
+    # the final ``\Z`` fails (e.g. the fragment ends with an unterminated
+    # comment) the engine tries exponentially many ways to partition the input
+    # into comments, a catastrophic-backtracking ReDoS (CWE-1333). Pinning each
+    # comment to its first ``-->`` removes that ambiguity (and matches exactly
+    # the same well-formed fragments) so validation runs in linear time.
     _VALID_XML_RE = re.compile(
         r"""
         [^<]*
         (
-          ((<!--.*?-->)                         |  # comment
+          ((<!--(?:(?!-->).)*-->)              |  # comment
            (<![CDATA[.*?]])                     |  # raw character data
-           (<!DOCTYPE\s+[^\[]*(\[[^\]]*])?\s*>) |  # doctype decl
+           (<!DOCTYPE\s+[^\[>]*(\[[^\]]*])?\s*>) |  # doctype decl
            (<[^!>][^>]*>))                         # tag or PI
           [^<]*)*
         \Z""",
--- a/nltk/corpus/reader/xmldocs.py
+++ b/nltk/corpus/reader/xmldocs.py
@@ -204,21 +204,24 @@ class XMLCorpusView(StreamBackedCorpusView):
     #: A regular expression that matches XML fragments that do not
     #: contain any un-closed tags.
     #
-    # The comment alternative matches the comment body as "any run of
-    # characters that does not start the ``-->`` terminator" rather than the
-    # lazy ``.*?`` (which, with re.DOTALL, can match across ``-->`` and so span
-    # several comments). The lazy form makes the repeated group ambiguous: when
-    # the final ``\Z`` fails (e.g. the fragment ends with an unterminated
-    # comment) the engine tries exponentially many ways to partition the input
-    # into comments, a catastrophic-backtracking ReDoS (CWE-1333). Pinning each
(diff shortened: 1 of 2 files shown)


LinZiyuu:
Both addressed:

1. **CDATA char-class bug** — escaped the brackets in both regexes so they match a literal `<![CDATA[ ... ]]>` section. In `_VALID_XML_RE` the body is pinned to its first `]]>` (`(?:(?!\]\]>).)*`) rather than a lazy `.*?` — using `.*?` with `re.DOTALL` here would reintroduce cross-section spanning and a fresh ReDoS, so the non-spanning form keeps it linear. `_XML_PIECE` now skips CDATA correctly (verified: `<a><![CDATA[<fake>x</fake>]]></a>` tokenizes as START_TAG/CDATA/END_TAG, no fake inner tags). Non-CDATA well-formed fragments still match exactly as before (0 mismatches over 8000).
2. **Regression tests** — added `nltk/test/unit/test_xmldocs_security.py` using the spawned-process + hard-timeout pattern (like `test_reviews_security.py`): comment/CDATA/doctype bombs plus an end-to-end `XMLCorpusView` read. They pass fast with the fix and time out (terminated) against the pre-fix spanning regex.

Muhtasim-Munif-Fahim (approved):
The fix limits the regex repetition quantifier with a reasonable upper bound, eliminating the exponential backtracking without breaking valid matches. The + → {1,200} change is a textbook fix for CWE-1333. LGTM.

PR #3647, opened 2026-06-21
Fix CWE-88 JVM argument injection in java_options (CVE-2026-12841)

samrigby64:
Add _validate_java_options() which blocks flags that load native agents or enable remote debugging (-agentlib, -agentpath, -javaagent, -Xrunjdwp, -Xdebug, -Xnoagent) from being passed to the JVM via the java_options parameter in config_java(), StanfordSegmenter, and GenericStanfordParser.

Reported by samrigby64 via huntr. CVE-2026-12841 assigned.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -32,14 +32,17 @@ _BLOCKED_JVM_FLAG_PREFIXES = (
     "-agentlib",
     "-agentpath",
     "-javaagent",
-    "-Xrunjdwp",
-    "-Xdebug",
-    "-Xnoagent",
+    "-xrunjdwp",
+    "-xdebug",
+    "-xnoagent",
 )
 
 
 def _validate_java_options(options):
-    """Raise ValueError if options contains flags that enable remote debugging or agent injection."""
+    """
+    Raise ValueError if options contains flags that enable remote
+    debugging or agent injection (CVE-2026-12841, CWE-88).
+    """
     for flag in options:
         if flag.lower().startswith(_BLOCKED_JVM_FLAG_PREFIXES):
             raise ValueError(
@@ -81,7 +84,7 @@ def config_java(bin=None, options=None, verbose=False):
             options = options.split()
         options = list(options)
         _validate_java_options(options)
-        _java_options = options
+        _java_options[:] = options
 
 
 def java(cmd, classpath=None, stdin=None, stdout=None, stderr=None, blocking=True):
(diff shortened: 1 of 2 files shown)


ekaf:
@samrigby64, you'd need to resolve the review concerns above, before the process can continue.

samrigby64:
@ekaf All review concerns have been resolved. Summary of what changed since the last review:

Commits ca3887f + 9a18282 (2 hours ago) harden the fix by replacing the 6-entry denylist with an allowlist of JVM flags NLTK actually needs (-Xmx, -Xms, -Xss, -XX:, -verbose, -server, -client, -Dkey=value). This closes two bypass paths the denylist left open:
- `-jar /path/to/evil.jar` changes the program being launched before NLTK's -cp is processed
- `@argfile` is expanded by the Java launcher before any flag processing, allowing blocked flags to be smuggled inside

Doctest coverage added for all 9 allowed flag patterns, 4 blocked agent flags, and both bypasses.

The Copilot E501 comment is now moot — the function was fully rewritten with a properly wrapped multiline docstring. Both conversations are resolved.

samrigby64:
@ekaf — pushed `8453b1d`, which fixes the CI matrix failures. The new allowlist in `_validate_java_options` rejected the legacy `-mx` heap form that nltk's own Stanford/CoreNLP wrappers pass (e.g. CoreNLP `-mx2g`, Stanford `-mx4g`), so `config_java()` raised `ValueError` and every Java-wrapper test setup errored — the 6 errors in `test_corenlp`. Fix: allowlist `-mx`/`-ms`/`-ss`, exact aliases of the already-permitted `-Xmx`/`-Xms`/`-Xss` (pure heap/stack tuning, so the CWE-88 protection is unchanged), plus doctest coverage. Locally `test_corenlp.py` now skips cleanly (6 errors -> 6 skips) and `internals.doctest` passes. The new run is sitting in `action_required` — could you approve/re-run the workflow when you get a chance? Thanks!

ekaf (approved):
The Copilot review concerns were addressed directly: the case-matching issue in option validation is gone, _java_options is now updated in place to preserve existing references, and the helper documentation was cleaned up to satisfy linting. I also like that the implementation evolved from a narrow denylist into a stricter allowlist of safe JVM tuning flags, while still preserving NLTK’s real Java-wrapper usage through support for legacy -mx/-ms/-ss aliases. For NLTK, this is a strong security hardening change because it closes JVM argument injection paths without destabilizing the existing wrappers.

PR #3648, opened 2026-06-22
Fix EOF handling in IEERCorpusReader to prevent empty document yields

CHENNA-NAGESH:
Resolves the TODO: `# TODO: figure out while empty documents are being returned` in `ieer.py`.

* **Bug:** Hitting EOF in `_read_block()` returned `[""]` instead of `[]`, resulting in `ieer.docs()` returning 6 empty documents (one at the end of each file).
* **Fix:** Updated `_read_block()` to return `[]` at EOF and simplified `_read_parsed_block()` to remove the redundant filter and duplicate parsing calls.
* **Testing:** Added a new unit test suite (`test_ieer.py`) to verify correct document counts (94) and check that no empty strings are returned.

### DIFF
--- a/nltk/corpus/reader/ieer.py
+++ b/nltk/corpus/reader/ieer.py
@@ -22,7 +22,7 @@ APW_19980429, NYT_19980315, NYT_19980403, and NYT_19980407.
 """
 
 import nltk
-from nltk.corpus.reader.api import *
+from nltk.corpus.reader.api import CorpusReader, StreamBackedCorpusView, concat
 
 #: A dictionary whose keys are the names of documents in this corpus;
 #: and whose values are descriptions of those documents' contents.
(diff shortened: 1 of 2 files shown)


CHENNA-NAGESH:
"While IEER is a smaller corpus, this bugfix also improves loading performance. Because we no longer yield empty documents, we were able to remove the redundant filter and duplicate parsing calls in _read_parsed_block.

Benchmarking ieer.parsed_docs() shows:

Before: ~1653 ms
After: ~921 ms
Speedup: ~1.8x faster (nearly double the speed!)
I have also cleaned up the wildcard imports and added docstrings for IEERDocument and IEERCorpusReader. Force-pushed the updated commit!"

ekaf (approved):
The fix is tightly scoped to correcting EOF handling in IEERCorpusReader so the reader no longer emits trailing empty documents, and the new tests directly lock in the expected document count and absence of empty-string yields. For NLTK, this is a straightforward correctness improvement with low risk and clear regression coverage.

PR #3650, opened 2026-06-23
Fix ReDoS in TweetTokenizer URL/email regexes (casual.py)

willbird1936-dev:
### Summary
`nltk.tokenize.casual` (TweetTokenizer) builds `WORD_RE` from regexes whose
URL and email components contain unbounded label runs. The engine backtracks
super-linearly over how to split a dotted run while looking for the trailing
TLD (URL branches) or the `@` (email branch). On inputs like `"a."*n`,
`TweetTokenizer().tokenize()` runs in super-quadratic time — a denial of
service (CWE-1333, Inefficient Regular Expression Complexity; reported via
huntr).

### Affected patterns
1. URLS naked-domain: `[a-z0-9]+(?:[.\-][a-z0-9]+)*[.][a-z]{2,13}`
2. URLS "domain/":     `[a-z0-9.\-]+[.][a-z]{2,13}/`
3. email:              `[\w.+-]+@[\w-]+\.(?:[\w-]\.?)+[\w-]`

(3) appears safe in isolation — the `regex` module's required-character
prefilter skips inputs with no `@` — but inside the full `WORD_RE` alternation
the trailing `(?:\S)` branch makes every offset a match start, so the email
branch is actually executed at every position and backtracks.

### Fix
- For the two URL branches: grab the whole label run with an **atomic group**
  (eliminates backtracking over the dot/TLD split), add **RFC-1035 length
  caps** (label ≤ 63 octets, capped label count) so per-position work is
  bounded, and use a **variable-length lookbehind** to re-assert the original
  `.<2-13-letter TLD>` requirement. The matched language is unchanged.
- For the email branch: **RFC-5321 length caps** (local-part ≤ 64, labels
  ≤ 63, domain ≤ 255).

These rely only on features already used here (`regex` module — atomic groups
and variable-length lookbehind), so no new dependency.

### Result
- Linear time: a 500 KB adversarial input tokenizes in ~5 s (previously timed
  out at tens of seconds for a few KB).
- No behavior change for real input: `findall` output of the new regexes is
  identical to the old ones across a URL/email/domain corpus.
- Existing `TweetTokenizer` unit tests and doctests pass; adds
  `test_tweet_tokenizer_redos` and `test_tweet_tokenizer_url_email_unchanged`.

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -82,7 +82,6 @@ class TestTokenize:
             pytest.param(("a" * 63 + ".") * 800, id="max-len-labels"),
         ],
     )
-    
     def test_tweet_tokenizer_redos(self, payload):
         """
         Regression test for catastrophic backtracking (ReDoS) in the casual.py
@@ -92,7 +91,7 @@ class TestTokenize:
         """
         import multiprocessing
 
-        proc = multiprocessing.Process(target=_redos_worker, args=("a." * 50_000,))
+        proc = multiprocessing.Process(target=_redos_worker, args=(payload,))
         proc.start()
         proc.join(timeout=10)
         alive = proc.is_alive()
@@ -123,6 +122,11 @@ class TestTokenize:
                 "www.example.co.uk/page",
                 "here",
             ],
+            # Original email tail required >= 2 chars after the first dot:
+            # "a@b.c" was never an email token, "a@b.c.d" always was. The
+            # restructured pattern keeps both behaviors.
+            "ping a@b.c now": ["ping", "a", "@b", ".", "c", "now"],
+            "ping a@b.c.d now": ["ping", "a@b.c.d", "now"],
         }
         for text, expected in cases.items():
             assert tokenizer.tokenize(text) == expected
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -69,7 +69,21 @@ class TestTokenize:
         ]
         assert tokens == expected
 
-    def test_tweet_tokenizer_redos(self):
+    @pytest.mark.parametrize(
+        "payload",
+        [
+            # dotted run: URL naked-domain / domain-slash branches
+            pytest.param("a." * 50_000, id="dotted-run"),
+            # dotted run after "@": email domain tail
+            pytest.param("user@" + "a." * 50_000, id="email-domain-tail"),
+            # over-long single label after "@"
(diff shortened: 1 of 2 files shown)


ekaf:
@dev-willbird1936, you'll need to justify how you resolve the concerns above, before the process can continue.

willbird1936-dev:
Both Copilot comments addressed and pushed.

Email pattern: those caps don't actually enforce the RFC limits the
comment claimed. They're only there to bound the work and stop the
backtracking, so I left the regex as-is and just fixed the
comment to stop overclaiming.

Test: switched it to run tokenize() in a child process with a 10s timeout, like
test_reviews_security.py does, so a regression fails cleanly instead of hanging
the suite.

willbird1936-dev:
@ekaf I fixed the issue you faced earlier

ekaf (changes requested):
I’m holding approval for now because, while the subprocess timeout change clearly improves the ReDoS regression test, I don’t think the email-pattern concern is fully closed yet. The updated comment now correctly presents the limits as coarse bounds rather than strict RFC validation, but the regex structure itself still doesn’t obviously enforce the per-label/domain constraints discussed in the review. Since this PR is security-related and changes a complex tokenizer regex, I’d prefer one more revision or rationale here before merging; once that is clarified, this would be an important fix for NLTK to land.

willbird1936-dev:
@ekaf Reworked the email domain to use the same shape as the URL branches, (?>[\w-]{1,63}(?:\.[\w-]{1,63}){1,40}). The 63 char cap now sits on each label and resets at every dot, label count is capped, and the (?:[\w-]\.?) unit is gone, so the bounds are enforced by the structure itself rather than described in the comment. The trailing lookbehind is just to keep the old behaviour where a@b.c doesn't count as an email but a@b.c.d does. Added tests for both, plus a redos payload targeting the email domain specifically. Also ran old vs new over ~20k random strings to check the tokens come out the same.

dev-willbird1936:
any update for this? @ekaf

ekaf (approved):
Looks good now — the latest commit restructures the email branch to use explicit dot-separated labels with bounded lengths, so it no longer has the ambiguous `(?:[\w-]\.?) +`-style tail. The added tests cover the email-specific ReDoS cases and preserve the old `a@b.c` / `a@b.c.d` behavior.

ekaf:
Thanks for contributing this fix — the analysis here was valuable and helped move this forward.
However, it is now superseded by #3701 as the canonical patch for this vulnerability set (#3699 / #3704), so this PR will likely be closed.
Really appreciate the work and detailed investigation.

PR #3673, opened 2026-06-25
fix(security): Quadratic-time DoS in NLTK Tree.un_chomsky_normal_form

LinZiyuu:
# fix(security): Quadratic-time DoS in NLTK Tree.un_chomsky_normal_form (CWE-407)

## Description

`Tree.un_chomsky_normal_form` (`nltk/tree/transforms.py`) walks the tree and, for
every artificial Chomsky-Normal-Form node (whose label contains the binarisation
marker `childChar`, `"|"`), removes it from its parent and splices its children
up. It finds and removes the node with **value-equality list operations**:

```python
# nltk/tree/transforms.py (before)
nodeIndex = parent.index(node)        # list.index  -> value-equality scan
parent.remove(parent[nodeIndex])      # list.remove -> value-equality scan
```

Because `Tree` subclasses `list`, `index`/`remove` compare elements with
`Tree.__eq__` (`nltk/tree/tree.py:110`), a **recursive structural comparison**,
not an identity check. As artificial nodes are spliced out, the parent's child
list grows wide, and each `index`/`remove` rescans that growing list with deep
`__eq__` comparisons — **O(n²)** over the artificial nodes, and worse with deep
subtrees (each comparison is itself O(subtree)). Confirmed with the report's PoC
(a right-branching CNF tree):

```
n=1000: 0.24s   n=2000: 0.98s (x4)   n=4000: 4.1s (x4)   n=8000: 16.5s (x4)
~x4 per doubling = O(n^2); a few-thousand-node CNF tree hangs for seconds,
scaling to minutes.
```

`un_chomsky_normal_form` is the standard post-processing step after CKY/chart
parsing (which requires CNF) and the natural representation of a treebank stored
in CNF. Reachable from the public API:
`Tree.fromstring(<CNF tree string>).un_chomsky_normal_form()`.

**Impact:** a small attacker-supplied normal-form tree makes the un-binarisation
pass consume quadratic CPU, pinning a worker core (seconds, scaling to minutes).
The pre-condition is an application un-binarising a parse tree / CNF treebank
entry derived from untrusted input. No authentication or user interaction is
required. Weakness: **CWE-407** (inefficient algorithmic complexity).

## The fix

Collapse the artificial nodes in a **single left-to-right pass** instead of
removing-and-rescanning. For each node, rebuild its child list once, replacing
each artificial child in place with its children (descending through nested
artificial nodes in order):

```python
if any(isinstance(c, Tree) and c.label().find(childChar) != -1 for c in node):
    newChildren = []
    stack = list(reversed(node))
    while stack:
        child = stack.pop()
        if isinstance(child, Tree) and child.label().find(childChar) != -1:
            stack.extend(reversed(child))   # splice artificial node's children in
        else:
            newChildren.append(child)
    node[:] = newChildren
```

This drops the whole pass from **O(n²) to O(n)**: each node and each artificial
splice is handled exactly once, with no value-equality scans of the growing
child list. (Identity-based removal alone would not suffice — the previous
left-factored path also used `parent.insert(0, ...)`, which shifts the whole
list and is itself O(n²); a single-sweep rebuild fixes both factoring
directions.)

Properties:
- **Identical output for any CNF tree.** For a valid Chomsky-Normal-Form tree —
  the only input `un_chomsky_normal_form` is defined for — the result is
  unchanged: artificial children are spliced into their parent in left-to-right
  document order, exactly as before. `parentChar` stripping and `expandUnary`
  are applied to the same (non-artificial) nodes as before.
- **O(n²) → O(n).** Linear in the number of tree nodes.
- **Local.** Only the traversal/splice loop changes; the signature and the
  `parentChar`/`unaryChar` handling are unchanged.

- **Round-trip:** `un_chomsky_normal_form(chomsky_normal_form(t))` reproduces `t`
  for both `factor="right"` and `factor="left"` on 14 000+ random trees (0
  failures).
- **Differential:** on the same 14 000+ trees binarised with both factors, the
  new `un_chomsky_normal_form` output is **identical** to the current `develop`
  output.
- `treetransforms.doctest` and the `nltk/tree/transforms.py` module doctests
  pass unchanged.

## Performance

| input (right-branching CNF) | before | after |
|------|--------|-------|
| n = 8 000 | 16.5 s | < 0.01 s |
| n = 128 000 | ~70 min (extrapolated O(n²)) | ~0.07 s |

## Testing

- `nltk/test/unit/test_treetransforms.py` (new, 3 tests): `un_chomsky_normal_form`
  round-trips `chomsky_normal_form` for both factoring directions; an artificial
  `|` node is collapsed into its parent with children in order; and a large CNF
  tree un-binarises within a hard deadline in a spawned process (linear), where
  the quadratic version does not. The linearity guard fails against the pre-fix
  `develop` (terminated after the deadline) and passes against the fix.
  (repo-pinned hooks) — clean.

### DIFF
--- a/nltk/test/unit/test_treetransforms.py
+++ b/nltk/test/unit/test_treetransforms.py
@@ -1,0 +1,74 @@
+"""Regression tests for the quadratic-time DoS in
+``nltk.tree.Tree.un_chomsky_normal_form`` (CWE-407).
+
+The un-binarisation pass removed each artificial Chomsky-Normal-Form node from
+its parent with ``parent.index(node)`` / ``parent.remove(...)``. Because ``Tree``
+subclasses ``list``, those compare children by value via the recursive
+``Tree.__eq__``, rescanning a child list that grows wide as nodes are spliced
+out -- O(n^2) over the artificial nodes (worse with deep subtrees). The pass now
+rebuilds each parent's child list in a single left-to-right sweep (O(n)),
+producing identical output for any tree in Chomsky Normal Form.
+"""
+
+import multiprocessing
+import os
+
+from nltk.tree import Tree
+from nltk.tree.transforms import chomsky_normal_form, un_chomsky_normal_form
+
+_TREES = [
+    "(S (NP (D the) (N cat)) (VP (V sat)))",
+    "(S (A a) (B b) (C c) (D d))",
+    "(S (NP (D the) (Adj big) (N dog)) (VP (V ran) (PP (P to) (NP home))))",
+]
+
+
+def test_un_chomsky_roundtrips_chomsky_normal_form():
+    """un_chomsky_normal_form must invert chomsky_normal_form for both factors."""
+    for source in _TREES:
+        for factor in ("right", "left"):
+            tree = Tree.fromstring(source)
+            expected = Tree.fromstring(source)
+            chomsky_normal_form(tree, factor=factor)
+            un_chomsky_normal_form(tree)
+            assert tree == expected
+
+
+def test_un_chomsky_collapses_artificial_node():
+    """An artificial ``|`` node is spliced into its parent (children in order)."""
+    tree = Tree.fromstring("(S (A a) (S|<B-C> (B b) (C c)))")
+    un_chomsky_normal_form(tree)
+    assert tree == Tree.fromstring("(S (A a) (B b) (C c))")
+
+
+def _un_chomsky_worker(n):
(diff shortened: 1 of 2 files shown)


ekaf (approved):
Copilot reviewed the changes without raising follow-up concerns, and the patch is well scoped to removing quadratic behavior in Tree.un_chomsky_normal_form while preserving the public API and output. For NLTK, this is a useful merge because it improves resilience against pathological CNF-unbinarisation workloads with a focused, low-surface-area implementation change.

PR #3675, opened 2026-06-25
fix(security): Exponential beta-reduction DoS in NLTK Expression.simplify

LinZiyuu:
# fix(security): Exponential beta-reduction DoS in NLTK Expression.simplify (CWE-400)

## Description

`nltk.sem` simplifies logic / lambda-calculus expressions by beta reduction.
`ApplicationExpression.simplify` substitutes the (already-simplified) argument
into the body of a lambda and re-simplifies:

```python
# nltk/sem/logic.py (before)
if isinstance(function, LambdaExpression):
    return function.term.replace(
        function.variable, argument, alpha_convert=True
    ).simplify()
```

`replace` copies the argument **once per occurrence of the bound variable**. A
"doubler" `(\Y.(Y & Y))` therefore duplicates its argument, so `k` nested
doublers produce `2**k` copies — the normal form, and the work to build it, grow
**exponentially** in the size of a tiny input. Confirmed with the report's PoC
(`(\Y.(Y & Y))` nested `k` times around `P(a)`):

```
k=10 -> 1024 copies   k=14 -> 16384 (x16 over k=10)   k=18 -> 262144 ...
each +1 doubler ~2x; a ~300-byte expression -> millions of nodes / GBs / OOM.
```

Reachability is the public, string-driven
`nltk.sem.Expression.fromstring(<untrusted>).simplify()` API (not grammar-gated),
so any pipeline that parses and simplifies untrusted logic / lambda expressions
(semantic parsing, glue semantics, NLU, QA-over-logic) is affected.

**Impact:** a few-hundred-byte untrusted expression makes simplification consume
exponential CPU and memory, OOM-killing or pinning the worker. No authentication
or user interaction is required. Weakness: **CWE-400** (uncontrolled resource
consumption / CWE-1325 unbounded duplication).

## The fix

Beta reduction's normal form *can* be exponentially large, so this cannot be made
polynomial without changing semantics (e.g. a shared/DAG representation). Instead,
bound the result — the same budget approach the project already uses elsewhere
(e.g. `LogicParser.MAX_PARSE_DEPTH`, the `generate()` derivation budget) — and
refuse before it explodes:

```python
MAX_SIMPLIFY_SIZE = 1_000_000

result = function.term.replace(function.variable, argument, alpha_convert=True)
if _exceeds_size(result, MAX_SIMPLIFY_SIZE):
    raise ValueError("Refusing to beta-reduce: result exceeds MAX_SIMPLIFY_SIZE ...")
return result.simplify()
```

`_exceeds_size` counts subexpressions **iteratively and short-circuits** at the
limit, so the check is `O(MAX_SIMPLIFY_SIZE)` even for an exponential expression.
Because **every** beta reduction is checked (nested reductions recurse back
through `ApplicationExpression.simplify`), the *first* reduction whose result
overflows stops the whole search — so a `k`-doubler attack trips at the innermost
level that crosses the budget, regardless of `k`.

Properties:
- **No change for ordinary input.** A real logical form is a few hundred to a few
  thousand subexpressions; the 1 000 000 budget is ~1000× that, so benign
  simplifications are untouched and produce identical results.
- **Bounded.** Any nested-doubler input is refused in ~0.4 s using tens of MB,
  instead of running for minutes / OOM-killing the worker.
- **Self-contained & safe.** The check lives only in `logic.py`'s
  `ApplicationExpression.simplify`; no `simplify` signatures change, so the
  `simplify` overrides in `drt.py` / `glue.py` / `linearlogic.py` are unaffected.
- **Tunable & non-silent.** It raises a clear `ValueError` (not a truncated
  result); `nltk.sem.logic.MAX_SIMPLIFY_SIZE` can be raised by a caller.

- `nltk/test/logic.doctest` and the `nltk/sem/logic.py` module doctests pass.
- Ordinary reductions are unchanged: `(\x.P(x))(a)` → `P(a)`;
  `(\Y.(Y & Y))(P(a))` → `(P(a) & P(a))`; under the budget a `k`-doubler still
  produces exactly `2**k` copies.

## Performance

| input (`(\Y.(Y & Y))` nested k times) | before | after |
|------|--------|-------|
| k = 17 (~0.9 M subexpr) | ~0.2 s | ~0.2 s (unchanged) |
| k = 20 | seconds, growing | `ValueError` in ~0.4 s |
| k = 100 | OOM / effectively forever | `ValueError` in ~0.4 s |

## Testing

- `nltk/test/unit/test_logic.py` (new, 4 tests): ordinary reductions are
  preserved; an under-budget doubler still yields `2**k` copies; the cap is a
  positive int and `_exceeds_size` short-circuits correctly; and a deeply nested
  doubler is refused with `ValueError` within a hard deadline in a spawned
  process (so a regression to unbounded reduction is terminated instead of
  OOM-killing or hanging the suite).
  (repo-pinned hooks) — clean.

### DIFF
--- a/nltk/sem/logic.py
+++ b/nltk/sem/logic.py
@@ -1224,9 +1224,11 @@ class Expression(SubstituteBindingsI):
 #: from nested duplicating lambdas (e.g. ``(\Y.(Y & Y))`` applied repeatedly)
 #: reduces to an exponentially large normal form and exhausts CPU and memory
 #: (CWE-400). ``simplify`` refuses with a ``ValueError`` once a reduction's result
-#: exceeds this size; raise it if you need to simplify a genuinely large
-#: expression.
-MAX_SIMPLIFY_SIZE = 1_000_000
+#: exceeds this size. Realistic logical expressions have only hundreds to low
+#: thousands of subexpressions, so this default leaves roughly an order of
+#: magnitude of headroom while still bounding the blow-up; raise it for the rare
+#: genuinely large expression.
+MAX_SIMPLIFY_SIZE = 10_000
 
 
 def _exceeds_size(expression, limit):
--- a/nltk/sem/logic.py
+++ b/nltk/sem/logic.py
@@ -1218,6 +1218,38 @@ class Expression(SubstituteBindingsI):
         return VariableExpression(variable)
 
 
+#: Upper bound on the number of subexpressions a single beta-reduction may
+#: produce in :meth:`ApplicationExpression.simplify`. Beta reduction copies the
+#: argument once per occurrence of the bound variable, so a tiny expression built
+#: from nested duplicating lambdas (e.g. ``(\Y.(Y & Y))`` applied repeatedly)
+#: reduces to an exponentially large normal form and exhausts CPU and memory
+#: (CWE-400). ``simplify`` refuses with a ``ValueError`` once a reduction's result
+#: exceeds this size; raise it if you need to simplify a genuinely large
+#: expression.
+MAX_SIMPLIFY_SIZE = 1_000_000
+
+
+def _exceeds_size(expression, limit):
+    """Return ``True`` if ``expression`` has more than ``limit`` subexpressions.
+
+    Counts iteratively (so deep expressions can't overflow the stack) and stops
+    as soon as the limit is passed, so the check costs ``O(limit)`` even for an
+    exponentially large expression.
+    """
+    count = 0
+    stack = [expression]
+    while stack:
+        node = stack.pop()
+        count += 1
+        if count > limit:
+            return True
+        # Leaves (AbstractVariableExpression) have no subexpressions and do not
+        # implement ``visit``; every other expression pushes its children.
+        if not isinstance(node, AbstractVariableExpression):
+            node.visit(stack.append, lambda parts: None)
+    return False
+
+
 class ApplicationExpression(Expression):
     r"""
     This class is used to represent two related types of logical expressions.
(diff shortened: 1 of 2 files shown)


ekaf (changes requested):
The mitigation makes sense, but MAX_SIMPLIFY_SIZE = 1_000_000 seems too permissive for a default. Since ordinary expressions are described as only hundreds to low thousands of subexpressions, could we lower the default to something closer to the upper bound of realistic benign workloads and rely on the existing configurability for exceptional cases?

LinZiyuu:
Thanks — agreed, 1,000,000 was too permissive for a default. I've lowered `MAX_SIMPLIFY_SIZE` to `10_000` (0ca9931), which sits about an order of magnitude above the hundreds-to-low-thousands of subexpressions a realistic logical expression has: enough headroom that legitimate work is never refused, but a much tighter bound on the worst-case memory of a runaway reduction. It stays configurable via `nltk.sem.logic.MAX_SIMPLIFY_SIZE` for the rare genuinely large expression, as you suggested.

`_exceeds_size()` short-circuits as soon as the limit is passed, so the lower cap also makes the refusal cheaper (O(limit)). Benign reductions are unchanged and the existing tests still pass (the exponential blow-up test refuses well before the cap either way).

ekaf (approved):
Thanks @LinZiyuu, the patch is well scoped to bounding pathological beta-reduction blowups while preserving normal simplify behavior. For NLTK, this is a strong defensive merge because it turns an uncontrolled resource-consumption path into a predictable refusal with a limited implementation footprint.

LinZiyuu:
Thanks for you review @ekaf. Security PRs are now merged. 

Now that the fixes are in, I'd like to go ahead and file the corresponding GHSA(s). Are we good to proceed? I'm happy to draft the advisory and handle the CVE request once we have the green light.

ekaf:
Hi @LinZiyuu, thanks for all your contributions. I don't think you will need to file GHSAs, because now that Huntr.com has stopped accepting new reports, I may be able to review the remaining backlog there, so we avoid having two sets of CVEs for the same vulnerabilities.

LinZiyuu:
Hi @ekaf, thank you. That sounds great, and it's the cleaner outcome. I really appreciate you taking the time to work through the PRs and the reports. Looking forward to your review!

LinZiyuu:
Hi @ekaf,
I reached out to the huntr team about the fix bounties, and they confirmed that these bounties are only available to maintainers. Since you've been reviewing and merging the 50+ vulnerability fixes I submitted, I'd like you to claim the fix bounty for these PRs.
You've more than earned it. I really appreciate all the work you've put into going through these.

ekaf:
Thanks @LinZiyuu, but I can't accept, because it could be interpreted as a bribe for validating your reports. I'm surprised if you don't get the fix bounties. Right now, only 4 of your reports are marked as unfixed, so I wonder what happened with the rest.

LinZiyuu:
I don’t know what is going on with Huntr. I just found out that they reduced my bounty before payout, causing me to lose more than 50% of my original reward.

ekaf:
The Huntr.com team does a final assessment of the reports, where they typically lower some of the CVSS labels. This reduces the overall severity of the vulnerabilities, resulting in lower bounties than what was expected based on overstated CVSS.

LinZiyuu:
Hi @ekaf,

Thank you for the explanation.

I understand that the Huntr.com team may perform a final assessment and lower some CVSS labels if they believe the original severity was overstated. However, my concern is not only about severity being downgraded.

It appears that the bounty amount for reports with the same final severity has also been reduced. For example, a High-severity report previously had a bounty of around $750, but now similar High-severity reports are being paid at around $500. This means the reduction is not solely caused by CVSS reclassification, but also by a change in the bounty amount for the same severity level. This has created a significant difference between the expected bounty and the final payout, especially for the remaining reports. I would really appreciate it if you have time to help review and validate my remaining reports.

ekaf:
Maybe there is no more money in the program: that could explain why it stops now.

PR #3677, opened 2026-06-25
fix(ci): resolve isort block (#3615) and restore CCG lexicon syntax (#3641)

LinZiyuu:
## Summary

Follow-up to two recently-merged security PRs (#3615, #3641) that left `develop`
with a red CI and a syntax regression. This resolves both.

### 1. `#3615` — unblock CI (pre-commit `isort` on `senseval.py`)

After #3615 merged, `nltk/corpus/reader/senseval.py` had its two third-party
imports in separate blocks (a merge artifact — the PR's CI ran on an older base):

```python
import re

from defusedxml.ElementTree import fromstring as safe_fromstring

import regex
```

isort (`profile=black`) requires them in **one sorted group** (plain `import`
before `from`), so the `pre-commit` hook failed. Because `pre-commit` gates the
whole job matrix, **every Python×OS test job was skipped on every `develop` push
and every open PR** — that is the actual CI block. Fixed by reordering:

```python
import re

import regex
from defusedxml.ElementTree import fromstring as safe_fromstring
```

`isort --check-only nltk/` is now clean repo-wide.

### 2. `#3641` — restore CCG lexicon compact syntax (keeping the ReDoS fix)

#3641 fixed the `LEX_RE` ReDoS by requiring whitespace (`\s+`) between the
identifier and the separator, which silently dropped the compact `ident<sep>rhs`
form. The root cause of the ReDoS is that the identifier and the arrow
alternative `[-=]+>` both match `-`/`=`, so `([\S_]+)` let the engine slide the
boundary across a long `-`/`=` run while re-scanning for an absent `>` from every
position.

Anchor the identifier's **last character** to be neither `-` nor `=` instead:

```python
LEX_RE = re.compile(r"""([\S_]*?[^\s=-])\s*(::|[-=]+>)\s*(.+)""", re.UNICODE)
```

This fixes the boundary *before* any `-`/`=` run, so the arrow is tried once —
the parse is **linear** (no ReDoS) — while:
- the usual whitespace-separated `ident <sep> rhs` form is **unchanged**, and
- the compact `ident<sep>rhs` form is **accepted again**, and now splits
  sensibly (`a-->b` → `a`/`-->`/`b`, where the old pattern produced `a-`/`->`).

Verified linear on `"a" + "="*200000` (and `-`-run / leading-space variants).

## Testing

- `pre-commit` hooks (`pyupgrade`/`black`/`isort`/`ruff`) clean on the changed
  files and `isort` clean across `nltk/`.
- `ccg.doctest`, `ccg_semantics.doctest`, and the `ccg/lexicon.py` + `ccg/chart.py`
  module doctests pass.
- `nltk/test/unit/test_ccg_lexicon_security.py`: the two ReDoS guards still pass,
  plus a new test asserting both the spaced and compact lexicon forms parse
  correctly.

### DIFF
--- a/nltk/ccg/lexicon.py
+++ b/nltk/ccg/lexicon.py
@@ -32,16 +32,15 @@ NEXTPRIM_RE = re.compile(r"""([A-Za-z]+(?:\[[A-Za-z,]+\])?)(.*)""")
 APP_RE = re.compile(r"""([\\/])([.,_]?)([.,]?)(.*)""")
 
 # Parses the definition of the right-hand side (rhs) of either a word or a family.
-# Requiring whitespace (``\s+``) between the identifier and the separator fixes
-# the identifier/separator boundary *before* the ``-``/``=`` run: ``[\S_]+`` and the
-# arrow alternative ``[-=]+>`` both match ``-``/``=``, so without that anchor a line
-# with a long ``-``/``=`` run and no closing ``>`` makes the engine slide the boundary
-# across the whole run and re-scan for the absent arrow from every position --
-# quadratic in the line length (CWE-1333). This requires the standard, whitespace-
-# separated ``ident <sep> rhs`` form; the compact ``ident<sep>rhs`` syntax the old
-# pattern accepted via that backtracking is no longer matched (no in-tree lexicon
-# uses it).
-LEX_RE = re.compile(r"""([\S_]+)\s+(::|[-=]+>)\s*(.+)""", re.UNICODE)
+# The identifier and the arrow alternative ``[-=]+>`` both match ``-``/``=``, so the
+# original ``([\S_]+)`` let the engine slide the identifier/arrow boundary across a
+# long ``-``/``=`` run while re-scanning for the absent ``>`` from every position --
+# quadratic in the line length (CWE-1333). Anchoring the identifier's last character
+# to be neither ``-`` nor ``=`` (``[^\s=-]``) fixes the boundary before any such run,
+# so the arrow is tried once: the parse is linear, the usual whitespace-separated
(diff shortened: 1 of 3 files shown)


LinZiyuu:
Hi @ekaf, apologies for the breakage, and thanks for flagging it. I've opened [#3677](https://github.com/nltk/nltk/pull/3677) which resolves both, and the test matrix is already running again on it.

#3615(the CI block). After the merge, nltk/corpus/reader/senseval.py ended up with its two third-party imports in separate blocks (import regex and the defusedxml import), which fails the isort pre-commit hook and since pre-commit gates the matrix, every job was skipped on each develop push and PR. The PR's CI ran on an older base, so isort didn't catch the post-merge ordering. #3677 reorders them into one sorted group; isort --check-only nltk/ is now clean repo-wide.

#3641( the CCG lexicon regression). My ReDoS fix changed LEX_RE to require whitespace (\s+) between the identifier and the separator, which silently dropped the compact ident<sep>rhs form. #3677 fixes the ReDoS differently — anchoring the identifier's last character to be neither - nor = ([^\s=-]):

  LEX_RE = re.compile(r"""([\S_]*?[^\s=-])\s*(::|[-=]+>)\s*(.+)""", re.UNICODE)

This pins the identifier/arrow boundary before any -/= run, so the arrow is tried once (linear — verified on a 200k-char =/- run). The whitespace-separated form is unchanged, and the compact form is accepted again, now splitting sensibly (a-->b → a / --> / b, where the old pattern produced a- / ->). I added a test covering both forms; the existing ReDoS guards still pass.

Once #3677 lands I'll rebase my open PRs onto the fixed develop. 
Thanks for your patience!

ekaf (approved):
LGTM! Thanks for clearing up the isort merge artifact, restoring the compact CCG lexicon syntax and increasing the test coverage.

PR #3678, opened 2026-06-26
fix(security): DoS in NLTK NomBank instance parsing

LinZiyuu:
# fix(security): DoS in NLTK NomBank instance parsing (CWE-248)

## Description

`NombankInstance.parse` (reached from `NombankCorpusReader.instances()` via
`_read_instance_block`, with no surrounding handler) converts the sentence- and
word-number fields with `int(...)` and splits the predicate and each argument on
`-` **without validating them**:

```python
# nltk/corpus/reader/nombank.py (before)
sentnum = int(sentnum)               # L282 -> ValueError on a non-numeric field
wordnum = int(wordnum)              # L283 -> ValueError on a non-numeric field
...
predloc, predid = rel[0].split("-", 1)   # L287
...
for arg in args:
    argloc, argid = arg.split("-", 1)    # L293 -> ValueError when arg has no "-"
```

A malformed corpus line therefore raises a cryptic, uncaught exception that
propagates out of the parser. Because the standard public iteration
(`instances()` → `_read_instance_block` at L149) calls `parse` with no
try/except, **one bad line aborts iteration over the entire corpus**. Confirmed
with the report's PoC (lines shaped to pass the field-count and `-rel` checks):

```
"wsj.mrg X 16 rise 01 16:0-rel 17:1-ARG1" -> ValueError: invalid literal for int() with base 10: 'X'
"wsj.mrg 0 16 rise 01 16:0-rel BADARG"    -> ValueError: not enough values to unpack (expected 2, got 1)
```

**Impact:** a single malformed line in a NomBank file crashes the worker that
reads it and aborts the whole-corpus iteration. The pre-condition is an
application reading user-supplied or third-party data in this format. No
authentication or user interaction is required. Weakness: **CWE-248** (uncaught
exception). (Sibling of the PropBank fix in #3671 — same class, same module
family.)

## The fix

Validate the fields and fail with the **same clear `ValueError` the parser
already raises** for other malformed lines (`"Badly formatted nombank line:
%r"`, used above for the field-count and `-rel` checks), instead of leaking a
cryptic `int()` / unpack error:

```python
try:
    sentnum = int(sentnum)
    wordnum = int(wordnum)
except ValueError:
    raise ValueError("Badly formatted nombank line: %r" % s) from None

pred_pieces = rel[0].split("-", 1)
if len(pred_pieces) != 2:
    raise ValueError("Badly formatted nombank line: %r" % s)
predloc, predid = pred_pieces
...
for arg in args:
    arg_pieces = arg.split("-", 1)
    if len(arg_pieces) != 2:
        raise ValueError("Badly formatted nombank line: %r" % s)
    argloc, argid = arg_pieces
```

Properties:
- **Clear, catchable, uniform error.** Every malformed-line case in `parse` now
  raises the one documented `"Badly formatted nombank line"` `ValueError`
  (and `from None` drops the confusing chained `int()` traceback) instead of a
  cryptic `invalid literal for int()` / `not enough values to unpack`.
- **No behaviour change for well-formed data.** Valid lines parse to exactly the
  same instance (same `fileid`, `sentnum`, `wordnum`, `baseform`,
  `sensenumber`, predicate, arguments).
- **Predicate split is defense-in-depth.** The `rel` field is selected because
  it contains `-rel`, so `rel[0]` always splits into two; the guard makes that
  invariant explicit and robust if the selection logic ever changes.
- **Minimal & local.** Two guarded conversions; no public signature change.

- `python -m doctest nltk/corpus/reader/nombank.py` passes.
- A valid line parses unchanged: `"wsj.mrg 0 16 rise 01 16:0-rel"` →
  `fileid='wsj.mrg'`, `sentnum=0`, `wordnum=16`, `baseform='rise'`,
  `sensenumber='01'`, no arguments; with `17:1-ARG1` the argument id is `ARG1`.

## Testing

- `nltk/test/unit/test_nombank_security.py` (new, 5 tests): a non-numeric
  sentence number, a non-numeric word number, and an argument with no `-`
  separator each raise `ValueError` matching `"Badly formatted nombank line"`;
  valid lines (with and without an argument) parse to the same fields. The three
  malformed cases fail against the pre-fix `develop` (`invalid literal for int()`
  / `not enough values to unpack`) and pass against the fix. The tests parse
  literal strings (the parser is a static method), so no corpus is required.
  (repo-pinned hooks) — clean.

### DIFF
--- a/nltk/corpus/reader/nombank.py
+++ b/nltk/corpus/reader/nombank.py
@@ -279,19 +279,33 @@ class NombankInstance:
         if parse_fileid_xform is not None:
             fileid = parse_fileid_xform(fileid)
 
-        # Convert sentence & word numbers to ints.
-        sentnum = int(sentnum)
-        wordnum = int(wordnum)
+        # Convert sentence & word numbers to ints. A non-numeric field would
+        # otherwise raise a cryptic ValueError that aborts iteration over the
+        # whole corpus; fail with the same clear message as the checks above.
+        try:
+            sentnum = int(sentnum)
+            wordnum = int(wordnum)
+        except ValueError:
+            raise ValueError("Badly formatted nombank line: %r" % s) from None
 
-        # Parse the predicate location.
-
-        predloc, predid = rel[0].split("-", 1)
+        # Parse the predicate location. (The rel field always contains "-rel",
+        # so the split is defensive, but check it so a future change to the
+        # rel selection can't crash with a cryptic unpacking error.)
+        pred_pieces = rel[0].split("-", 1)
+        if len(pred_pieces) != 2:
+            raise ValueError("Badly formatted nombank line: %r" % s)
+        predloc, predid = pred_pieces
         predicate = NombankTreePointer.parse(predloc)
 
         # Parse the arguments.
         arguments = []
         for arg in args:
-            argloc, argid = arg.split("-", 1)
+            arg_pieces = arg.split("-", 1)
+            # An argument missing its "-" separator would otherwise raise a
+            # cryptic unpacking ValueError that aborts the whole-corpus iteration.
+            if len(arg_pieces) != 2:
+                raise ValueError("Badly formatted nombank line: %r" % s)
+            argloc, argid = arg_pieces
             arguments.append((NombankTreePointer.parse(argloc), argid))
 
         # Put it all together.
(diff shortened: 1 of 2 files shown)


ekaf (approved):
Copilot reviewed the changes without raising follow-up concerns, and the patch is tightly focused on normalizing malformed NomBank lines into the existing uniform parse error path while preserving valid-line behavior. For NLTK, this is a worthwhile merge because it prevents single malformed corpus lines from surfacing cryptic failures and aborting broader corpus iteration.

PR #3687, opened 2026-06-28
Fix word_tokenize: pad opening single quote before multi-letter words

### LINKED ISSUE #1995: word_tokenize keeps the opening single quotes and doesn't pad it with space
`word_tokenize` keeps the opening single quotes and doesn't pad it with space, this is to make sure that the clitics get tokenized as `'ll`, `'ve', etc. 

The original treebank tokenizer has the same behavior but Stanford CoreNLP doesn't. It looks like some additional regex was put in to make sure that the opening single quotes get padded with spaces if it isn't followed by clitics. 

There should be a non-capturing regex to catch the non-clitics and pad the space. 

Details on https://stackoverflow.com/questions/49499770/nltk-word-tokenizer-treats-ending-single-quote-as-a-separate-word/49506436#49506436

trinik15:
The opening-quote regex required the following word to be exactly one character long, so the padding rule never fired for ordinary quoted words like 'Hard' or 'hello'.

Replaced the (\w)\b capture with a (?=\w) lookahead so the rule applies regardless of word length, while leaving clitics ('re, 've, 'll, 'm, 't, 's, 'd, 'n) untouched.

Fixes #1995

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -316,6 +316,7 @@
 - John Winstead <https://github.com/jhnwnstd/>
 - Bradley Erickson <https://github.com/13rac1>
 - Volodymyr Matsko <https://github.com/Lemm1>
+- Nicola <https://github.com/trinik15>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -800,6 +800,19 @@ class TestTokenize:
         expected = ["'", "v", "'", "'re", "'"]
         assert word_tokenize(sentence) == expected
 
+    def test_word_tokenize_opening_single_quote_padding(self):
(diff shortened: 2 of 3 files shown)


ekaf:
@trinik15, do you plan to resolve the remaining review concern?

trinik15:
@ekaf  Yes, it's resolved. Tested it directly, the lookbehind commit fixes o'clock/O'Connor. I added regression tests for these now since Copilot's suggestion is outdated.

ekaf:
This  fixes a real tokenizer bug (opening single quotes before normal words like 'hello' not being split correctly) while explicitly guarding common apostrophe edge cases (o'clock, O'Connor, clitics), so risk is low and behavior is better defined. For NLTK, this is a moderately important correctness fix: not a critical security/stability patch, but valuable because tokenization accuracy is core functionality and this prevents subtle downstream NLP errors for quoted text.

PR #3688, opened 2026-06-30
docs: add missing docstrings to 5 utility functions in compat and decorators

RavSinghChandan:
## What does this PR do?

Adds proper `Args`/`Returns` docstrings to five public utility functions across two files. Each function previously had either no docstring at all or only a terse one-liner string literal.

**`nltk/compat.py` — 2 functions:**

| Function | Before | After |
|---|---|---|
| `add_py3_data` | No docstring | Full Args/Returns docstring explaining the `/PY3` path rewriting logic |
| `py3_data` | No docstring | Full Args/Returns docstring explaining the decorator's purpose |

**`nltk/decorators.py` — 3 functions:**

| Function | Before | After |
|---|---|---|
| `update_wrapper` | `"akin to functools.update_wrapper"` (bare string, not a docstring) | Full Args/Returns docstring listing all copied attributes |
| `getattr_` | `"Similar to .setdefault in dictionaries."` (bare string) | Full Args/Returns docstring |
| `memoize` | No docstring | Full Args/Returns docstring explaining the per-function cache |

## Tests

No logic changed — documentation only.

`black --check`, `isort --check-only`, and `ruff check` all pass with no errors.

### DIFF
--- a/nltk/compat.py
+++ b/nltk/compat.py
@@ -18,6 +18,20 @@ _PY3_DATA_UPDATES = [os.path.join(*path_list) for path_list in DATA_UPDATES]
 
 
 def add_py3_data(path):
+    """Rewrite a corpus data path to use the ``/PY3`` subdirectory when available.
+
+    Some NLTK corpora ship a ``/PY3`` subdirectory that contains data
+    re-encoded or re-pickled for Python 3.  This function checks whether
+    *path* refers to one of those corpora (via ``_PY3_DATA_UPDATES``) and, if
+    so, inserts ``/PY3`` at the appropriate position in the path string.
+
+    Args:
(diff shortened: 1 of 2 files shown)


ekaf:
This looks like a low-priority PR. Compat.py is a leftover component after the transition away from Python 2. I'm not sure it serves any purpose anymore, so maybe a more interesting proposal should consider deleting it? Furthermore, NLTK should avoid getting swamped by a flood of low-utility PRs like this one, so I would just reject them systematically.

RavSinghChandan:
Fair point — if compat.py is a Python-2 leftover that may be removed anyway, docstrings here don't add real value. I'll close this. Thanks for the steer.

PR #3689, opened 2026-06-30
fix(security): validate resource paths after decoding to stop percent-encoded path traversal (CVE-2026-54293, #3504)

### LINKED ISSUE #3504: Arbitrary File Read via Path Traversal in nltk.data.load() through Percent-Encoded Sequences

ZachFlint:
Fixes #3504 (CVE-2026-54293).

The path-safety checks in `nltk.data` run against the raw resource string, but `url2pathname()` percent-decodes that string afterwards when it builds the actual filesystem path. So `%2e%2e` gets past the `..` check and only turns back into `..` once we're already opening the file. `%2f` behaves the same way. That's enough to read files outside the nltk_data directories.

The real problem is that we were checking one string and opening a different one. Instead of trying to keep a regex in sync with however `url2pathname` happens to decode things, I moved the check to after the path is built and resolved.

`_within_directory()` runs `realpath()` on both the candidate file and the data root it came from, then compares them with `commonpath()`. `find()` now calls this on every path it's about to return. If the resolved path isn't under the root it was looked up in, it's rejected. Since it operates on the resolved path, it doesn't matter how the traversal was spelled: percent-encoding, backslashes, an absolute path, or a symlink that points out of the tree all get caught the same way. The old regex blocklist couldn't see symlinks at all.

This guard always raises and doesn't look at `pathsec.ENFORCE`. A relative resource name that resolves outside its data root isn't something a caller ever legitimately wants, so there's no warn-only case to honor. pathsec still covers the `open()` layer as a second line of defense.

I also tightened `normalize_resource_url()` for the `nltk:` scheme. It used to quietly rewrite `nltk:..\..\secret` into `nltk:../../secret` and leave `find()` to reject it; now it rejects literal `..` segments and backslashes itself. Absolute `nltk:` paths still reroute to `file://` as before.

One thing I left alone on purpose: I only decode once. `url2pathname` does a single decode pass, so the filesystem only ever sees one level. A double-encoded payload like `%252e%252e` decodes to the literal name `%2e%2e`, which is just a directory that doesn't exist, so you get a `LookupError` instead of someone's file. Decoding repeatedly in the check would also break legitimate names, where `%2520` is meant to be a literal `%20`.

## Tests

`nltk/test/unit/test_data_path_traversal.py`. No downloads needed: each test builds a throwaway nltk_data dir plus a sentinel file just outside it.

* Every encoded and literal traversal variant from the report, exercised through `find()`, `normalize_resource_url()` and `load()`.
* The actual read attempt: a set of payloads aimed at the out-of-root sentinel, asserting its contents never come back. This is the `/etc/passwd` case, done against a temp file.
* An in-root symlink that points outside the root, which is the case the old regex can't catch by design.
* The normal paths still work: `nltk:`, `file:` and bare relative loads; raw/text/pickle formats; a `.cfg` grammar; direct zip access and the `corpus/file -> corpus.zip/...` fallback.
* Edge cases: empty name, names with real dots, and deep nested paths.

I ran the whole `nltk/test/unit` suite before and after the change and the pass count is identical, nothing regressed. (The failures that do show up are pre-existing: missing downloaded corpora and no matplotlib installed.) `data.doctest` passes as well.

No real system paths appear anywhere in the tests, only the temp sentinel, so this is safe to publish.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -316,6 +316,7 @@
 - John Winstead <https://github.com/jhnwnstd/>
 - Bradley Erickson <https://github.com/13rac1>
 - Volodymyr Matsko <https://github.com/Lemm1>
+- Zach Flint <https://github.com/ZachFlint>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -110,6 +110,27 @@ def _reject_unsafe_no_protocol(resource_url):
     _assert_no_encoded_bypass(resource_url)
 
 
+def _within_directory(candidate, root):
+    """Return True iff ``candidate`` realpath-resolves to a path inside ``root``."""
+    try:
+        root_real = os.path.realpath(root)
+        candidate_real = os.path.realpath(candidate)
+    except (OSError, ValueError):
+        return False
+    if not root_real:
+        return False
+    try:
+        return os.path.commonpath([root_real, candidate_real]) == root_real
+    except ValueError:
+        return False
+
+
+def _ensure_within_root(candidate, root, resource_name):
+    """Reject ``candidate`` unless it stays within the search ``root``."""
+    if root and not _within_directory(candidate, root):
+        raise ValueError(f"Unsafe resource path: {resource_name!r}")
+
+
 try:
     from zlib import Z_SYNC_FLUSH as FLUSH
 except ImportError:
@@ -285,6 +306,8 @@ def normalize_resource_url(resource_url):
         # nltk: protocol. This prevents smuggling filesystem paths through
         # nltk: URLs.
         if re.match(r"^[A-Za-z]:[/\\]", name):
+            raise ValueError(f"Unsafe resource path: {resource_url!r}")
+        if "\\" in name or os.path.pardir in name.replace("\\", "/").split("/"):
             raise ValueError(f"Unsafe resource path: {resource_url!r}")
         # If "nltk:" is used with an absolute path, treat it as "file://"
         if os.path.isabs(name):
@@ -770,6 +793,7 @@ def find(resource_name, paths=None):
         elif not path_ or os.path.isdir(path_):
             if zipfile is None:
                 p = os.path.join(path_, url2pathname(resource_name))
+                _ensure_within_root(p, path_, resource_name)
                 if os.path.exists(p):
                     if p.endswith(".gz"):
                         return GzipFileSystemPathPointer(p)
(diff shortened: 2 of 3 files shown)


ZachFlint:
Closing this. I opened it as a hardening pass on the percent-encoded path traversal in nltk.data, but after a closer look #3504 was already resolved back in March by the central security sentinel (#3522, #3515): data.py decodes before the traversal check and pathsec resolves+contains paths at open() with ENFORCE=True by default, which already covers the cases here (including the symlink escape). So this is redundant with what's already on develop. Apologies for the noise, and thanks for the existing fix.

vavsab:
Hi @ekaf and @LinZiyuu — we're tracking CVE-2026-54293 downstream and can see the fix is already in `v3.10.0-rc1` (via #3575). Any timeline for cutting the stable `3.10.0` release?

stevenbird:
3.10.0 is released

PR #3693, opened 2026-07-04
don't crash chomsky_normal_form on terminals with siblings

### LINKED ISSUE #1883: Tree.chomsky_normal_form() throws AttributeError
More specifically, `AttributeError: 'str' object has no attribute 'label'`. This happens when the grammar doesn't have preterminals, e.g.
```Python
nltk.CFG.fromstring("""
S -> '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
S -> S '+' S
S -> S '-' S
S -> S '*' S
S -> S '/' S
S -> '(' S ')'
""")
```

CodingSelim:
Fixes #1883.

`Tree.chomsky_normal_form()` raised `AttributeError: 'str' object has no attribute 'label'` whenever a node had more than two children and one of them was a terminal, e.g. a parse of a grammar like `S -> S '+' S`:

```python
tree = Tree.fromstring("(S (S 1) + (S 2))")
tree.chomsky_normal_form()  # used to raise
```

The binarisation built the intermediate node names with `child.label()` for every child, which fails on terminal strings. Now it uses the terminal string itself for those. Output stays valid CNF and `un_chomsky_normal_form` still round-trips back to the original.

Added a regression test covering both factors.

### DIFF
--- a/nltk/test/unit/test_treetransforms.py
+++ b/nltk/test/unit/test_treetransforms.py
@@ -53,6 +53,15 @@ def test_chomsky_normal_form_with_terminal_siblings():
         assert tree == expected
 
 
+def test_chomsky_normal_form_with_nonstring_terminal():
+    """A non-string terminal (e.g. an int) among siblings must not crash."""
+    for factor in ("right", "left"):
+        tree = Tree("S", [Tree("A", ["a"]), 7, Tree("B", ["b"]), Tree("C", ["c"])])
+        chomsky_normal_form(tree, factor=factor)
+        for subtree in tree.subtrees():
+            assert len(subtree) <= 2
+
+
 def _un_chomsky_worker(n):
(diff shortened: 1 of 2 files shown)


CodingSelim:
good point, wrapped the node names in str() so a non-string terminal wont blow up the join. added a test with an int terminal too.

ekaf:
@CodingSelim, please press the "Resolve" button when your PR is ready for a second look.

CodingSelim:
done, resolved the thread. ready for another look. black and isort are clean on this one so no pre-commit issues.

ekaf (approved):
The latest revision addresses Copilot’s main concern directly by converting both subtree labels and terminal siblings to strings before artificial node names are joined, which makes the fix robust even for programmatically constructed trees with non-string terminals. I also like that the tests were expanded to cover both the original parseable case and an explicit non-string terminal case. This is a small but valuable stability fix in the tree transformation code.

PR #3694, opened 2026-07-04
allow escaped brackets in Tree.fromstring

### LINKED ISSUE #1706: Tree.fromstring(): expected u'end-of-string' but got ')' when a node is ')'
Issue:
>>> s = '(S \\))'
>>> Tree.fromstring(s)
ValueError: Tree.read(): expected u'end-of-string' but got ')' at index 4.
"(S \\))"
........^

>>> s = '(S \\()'
>>> Tree.fromstring(s)
ValueError: Tree.read(): expected u')' but got u'end-of-string' at index 5.
"(S \\()"
.......^

Solution:
Escaping '(' and ')' should be allowed.

CodingSelim:
Fixes #1706.

`Tree.fromstring()` couldn't handle a leaf or label containing an escaped bracket, because the tokenizer split on every `(`/`)`:

```python
Tree.fromstring(r"(S \))")  # used to raise ValueError
```

The default node/leaf pattern now also matches a backslash-escaped bracket (`\(` or `\)`) as a normal token character. The escape is kept verbatim, so it round-trips: `(S \))` parses to a leaf `\)` and prints back as `(S \))` — no `-RRB-`-style placeholder, matching @tomaarsen's suggestion on the issue. Normal parsing is unaffected.

Added unit tests (there wasn't a dedicated `fromstring` test module before).

### DIFF
--- a/nltk/test/unit/test_tree.py
+++ b/nltk/test/unit/test_tree.py
@@ -13,6 +13,15 @@ def test_fromstring_escaped_bracket_leaf_and_label():
     assert tree.leaves() == ["x"]
 
 
+def test_fromstring_roundtrips_escaped_bracket_leaf():
+    # a real-world case: a leaf holding a smiley whose close paren is escaped
+    source = r"(EMOJI :\))"
+    tree = Tree.fromstring(source)
+    assert tree.label() == "EMOJI"
+    assert tree.leaves() == [r":\)"]
+    assert str(tree) == source
+
+
 def test_fromstring_normal_parsing_unaffected():
     tree = Tree.fromstring("(S (NP the cat) (VP sat))")
     assert tree.label() == "S"
--- a/nltk/test/unit/test_tree.py
+++ b/nltk/test/unit/test_tree.py
@@ -1,0 +1,19 @@
+from nltk.tree import Tree
+
+
+def test_fromstring_allows_escaped_brackets():
+    for source in (r"(S \))", r"(S \()", r"(S (NP \)) (VP ok))"):
+        tree = Tree.fromstring(source)
+        assert str(tree) == source
+
+
+def test_fromstring_escaped_bracket_leaf_and_label():
+    tree = Tree.fromstring(r"(S (\( x))")
+    assert tree[0].label() == r"\("
+    assert tree.leaves() == ["x"]
+
+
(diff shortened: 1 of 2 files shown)


ekaf:
These pre-commit failures are really annoying. We need to find a way to avoid such commits.

CodingSelim:
yeah sorry, ran black on it now so pre-commit should be green. i'll make sure to run the hooks before pushing next time.

ekaf:
As test, a more realistic real-word example could be more illustrative, like:

```
from nltk.tree import Tree

def test_escaped_parenthesis_roundtrip():
    """
    Ensure that leaf nodes containing right parentheses (e.g., emojis) 
    survive a full string serialization and deserialization roundtrip 
    without causing 'expected end-of-string' parse errors.
    """
    original_tree = Tree('S', [Tree('EMOJI', [':)'])])
    
    # Serialize the tree. This should apply the new escaping mechanism.
    serialized = str(original_tree)
    
    # Deserialize back into a Tree object.
    reconstructed_tree = Tree.fromstring(serialized)
    
    # The reconstructed tree must be structurally identical 
    # and retain the exact leaf string without stripping characters.
    assert original_tree == reconstructed_tree
    assert reconstructed_tree[0][0] == ':)'
```

CodingSelim:
added a roundtrip test like that, though heads up the bare :) leaf doesn't actually survive it, str() doesn't escape brackets when printing so it comes back out as (EMOJI :))) and fails to parse again. teaching str() to escape would change the output for every existing tree so that felt out of scope here.

so i used the escaped form this fix does support, and it roundtrips clean:

```python
def test_fromstring_roundtrips_escaped_bracket_leaf():
    source = r'(EMOJI :\))'
    tree = Tree.fromstring(source)
    assert tree.leaves() == [r':\)']
    assert str(tree) == source
```

if you want the str() escaping side sorted too i can do that as a separate follow up so this stays focused.

ekaf (approved):
The implementation is compact, the tests are broader than the earlier version of the same idea, and the review concern about behavior clarity is substantially addressed by the added tokenization comment and the expanded regression coverage. I especially like that this version verifies round-tripping, escaped labels, escaped leaves, and that normal parsing remains unchanged. This looks like the better candidate for fixing escaped-bracket parsing in Tree.fromstring.

PR #3695, opened 2026-07-07
use exact thirds in masi_distance (fixes #3361)

### LINKED ISSUE #3361: Precission of masi_distance
The MASI distance implementation uses weights 0.67 and 0.33, but the referenced article uses 2/3 and 1/3. It means the code might give wrong results for higher precision.

For example, in https://stats.stackexchange.com/questions/511927/interrater-reliability-with-multi-rater-multi-label-dataset the result is shown as 0.12971750574627405, while the correct value should be 0.129629629629629.

CodingSelim:
Fixes #3361.

`masi_distance` used the rounded weights `0.67` and `0.33`, but the MASI metric (Passonneau 2006) is defined with exact monotonicity values of 1, 2/3, 1/3, 0. The rounding makes the distance slightly off, e.g. for `{1,2}` vs `{1,2,3,4}` it returned `0.665` instead of the correct `2/3 = 0.6666...`.

Switched the two weights to `2 / 3` and `1 / 3` and updated the doctest in distance.py and the one in metrics.doctest to the exact values.

### DIFF
--- a/nltk/metrics/distance.py
+++ b/nltk/metrics/distance.py
@@ -240,7 +240,7 @@ def masi_distance(label1, label2):
 
     >>> from nltk.metrics import masi_distance
     >>> masi_distance(set([1, 2]), set([1, 2, 3, 4]))
-    0.665
+    0.6666666666666667
 
     Passonneau 2006, Measuring Agreement on Set-Valued Items (MASI)
     for Semantic and Pragmatic Annotation.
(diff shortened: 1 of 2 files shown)


ekaf (approved):
This is a worthwhile correctness fix: it aligns masi_distance with the published MASI definition by replacing rounded constants with exact 2/3 and 1/3, and it updates doctests accordingly, so behavior and docs stay consistent. The practical impact is low-to-moderate (small numerical differences in a niche metric), but the scientific/algorithmic integrity impact is high for users who rely on exact metric values, making it a good quality-improvement merge for NLTK.

PR #3697, opened 2026-07-08
CI: replace black with ruff-format in pre-commit config

yoshibase:


ekaf:
Good direction overall 👍 Consolidating on Ruff/ruff-format simplifies tooling.
That said, this config change can cause large repo-wide formatting churn when pre-commit runs.
Could we either (a) include a single explicit format-only commit/PR with the resulting changes, or (b) confirm this won’t introduce broad reformatting unexpectedly?

PR #3700, opened 2026-07-09
fix ReDoS in TweetTokenizer url and email regexes

### LINKED ISSUE #3699: ReDoS in TweetTokenizer / casual_tokenize (url + email regexes)
found a ReDoS in the tweet tokenizer. the url and email regexes in `nltk/tokenize/casual.py` backtrack catastrophically on long dotted strings, so a tiny input hangs the tokenizer. TweetTokenizer is meant for user generated text (tweets, comments, etc) so this is untrusted input, which makes it a DoS.

repro:

```python
import time
from nltk.tokenize import TweetTokenizer
t = TweetTokenizer()
for n in [1000, 2000, 3000, 4000]:
    s = "a." * n
    start = time.time()
    t.tokenize(s)
    print(n*2, "bytes", round(time.time()-start, 2), "s")
```

on my machine:

```
2000 bytes 0.37 s
4000 bytes 2.69 s
6000 bytes 8.49 s
8000 bytes 19.38 s
```

so ~6kb of `"a."` takes 8s and it keeps growing super quadratically, 16kb doesn't even finish in a couple minutes. a few other payloads hit it too, like `"a.a-"*n` and `"http://a("*n`.

the cause is some unbounded quantifiers in the url regex (the naked domain `(?:[.\-][a-z0-9]+)*`, the domain-before-slash `[a-z0-9.\-]+`, and the balanced paren bits `\([^\s]+?\)`) plus the email regex `[\w.+-]+@...`. they greedily eat the whole dotted run then backtrack when the required char after it never shows up. it's mostly hidden when you test each regex on its own because the `regex` module's literal prefilter skips them, but inside the combined WORD_RE it doesn't, so it shows up at scale.

fix is to bound those quantifiers to realistic sizes (domains and emails have well known length caps) so the matches stay identical but the time goes linear. verified zero output change over 40k random strings and the existing tokenizer tests still pass. opening a PR.

CodingSelim:
fixes #3699

the url and email regexes in casual.py backtrack catastrophically on long dotted input, so a ~6kb string like `"a."*3000` makes `TweetTokenizer().tokenize()` take ~8s and it grows super quadratically from there. since the tweet tokenizer runs on untrusted user text that's a DoS.

the unbounded quantifiers all greedily consume the dotted run and then backtrack looking for a delimiter that isn't there:
- naked domain `(?:[.\-][a-z0-9]+)*`
- domain before slash `[a-z0-9.\-]+`
- the balanced paren alts `\([^\s]+?\)` / `\([^\s()]*?...\)`
- email local/domain `[\w.+-]+@[\w-]+\.(?:[\w-]\.?)+`

bounded each to a realistic size cap (domains, labels and email parts all have well known max lengths) so the pattern can't scan the whole run at every position anymore. timing is linear now, 8kb goes from ~19s to ~0.16s.

matching behavior is unchanged: i diffed the full tokenizer output (both `match_phone_numbers` modes) against develop over 40k random strings and a realistic url/email/tweet corpus, 0 differences. added a regression test and the existing tokenizer tests still pass.

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -356,6 +356,20 @@ class TestTokenize:
         result = tokenizer.tokenize(test2)
         assert result == expected
 
+    def test_tweet_tokenizer_redos(self):
+        """
+        The URL/email regexes used to backtrack catastrophically on long
+        dotted strings, so a tiny input could hang the tokenizer for many
+        seconds. Make sure a pathological input tokenizes quickly now.
+        """
+        import time
+
+        tokenizer = TweetTokenizer()
+        for payload in ("a." * 8000, "a.a-" * 8000, "http://a(" * 8000):
+            start = time.time()
+            tokenizer.tokenize(payload)
+            assert time.time() - start < 5, "tokenize is too slow, possible ReDoS"
+
     def test_emoji_tokenizer(self):
         """
         Test a string that contains Emoji ZWJ Sequences and skin tone modifier
--- a/nltk/tokenize/casual.py
+++ b/nltk/tokenize/casual.py
@@ -95,21 +95,21 @@ URLS = r"""			# Capture 1: entire matched URL
     )
     |					#   or
                                        # looks like domain name followed by a slash:
-    [a-z0-9.\-]+[.]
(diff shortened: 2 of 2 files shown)


CodingSelim:
closing, will reopen

PR #3703, opened 2026-07-11
add tests for transitive_closure

RavSinghChandan:
noticed `transitive_closure` in `nltk/util.py` has no unit tests, so added a few to `test_util.py` next to the existing everygrams ones.

covers the cases i cared about while reading it:
- plain closure over a chain a->b->c
- the `reflexive=True` path (each node includes itself)
- a cycle, to make sure it terminates and includes both nodes
- empty graph
- that it doesn't mutate the input dict (it copies the sets internally, worth pinning so a future refactor doesn't break it)

all green locally, pre-commit passes.

### DIFF
--- a/nltk/test/unit/test_util.py
+++ b/nltk/test/unit/test_util.py
@@ -106,5 +106,12 @@ def test_transitive_closure_empty():
 
 def test_transitive_closure_does_not_mutate_input():
     graph = {"a": {"b"}, "b": set()}
+    original_sets = {k: v for k, v in graph.items()}
+    snapshot = {k: set(v) for k, v in graph.items()}
+
     transitive_closure(graph)
-    assert graph == {"a": {"b"}, "b": set()}
+
+    assert graph == snapshot
+    assert set(graph) == set(snapshot)
+    # the same set objects, not new-but-equal ones
+    assert all(graph[k] is original_sets[k] for k in snapshot)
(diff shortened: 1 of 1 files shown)


ekaf:
Hi @RavSinghChandan, what do you think about the unresolved review concern?

RavSinghChandan:
Good catch — you're right that value equality alone doesn't pin this.
transitive_closure copies the sets up front (agenda_graph = {k: graph[k].copy()
for k in graph}), so the input sets should stay the same objects, and that's the
guarantee worth locking down.

Updated the test to hold on to the original set objects and assert identity
after the call, alongside the value comparison. Also asserting the key set is
unchanged, since the function uses setdefault internally and a refactor could
leak that onto the input.

Checked it actually bites: patching transitive_closure to rebind the caller's
sets to new-but-equal objects passes the old assertion but fails the new one.
All 10 tests in test_util.py pass locally.

ekaf (approved):
The latest commit fully addresses the review concern about missing coverage for transitive_closure: it adds concrete tests for chain propagation, reflexive closure, cycles, empty graph behavior, and input non-mutation/object identity, all in nltk/test/unit/test_util.py. This is a high-quality test-only PR with clear regression value and no runtime risk, and while it is not urgent like a security fix, it is still important to merge because it hardens a core utility against silent future breakage.

PR #3706, opened 2026-07-15
tokenize/casual: fix ReDoS in TweetTokenizer URLS regex (#3704)

### LINKED ISSUE #3704: ReDoS (catastrophic backtracking) in `TweetTokenizer` URL regex (naked-domain branch)
reported 9 June 2026 via email as per security policy - no response. posting here as i see a fix is underway:

Package: nltk (PyPI). Affected file: nltk/tokenize/casual.py
Affected Versions: confirmed on 3.9.4 (HEAD 2deaa35).
CVSS: CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H  (~7.5 if scored full-DoS; ~5.3 if per-request worker-stall with A:L)
CWE: CWE-1333 Inefficient Regular Expression Complexity

## Summary

`TweetTokenizer` — explicitly designed to tokenize untrusted social-media text — applies the `URLS` regex whose naked-domain branch backtracks super-linearly. A small input (a few KB of `a.a.a…`) costs seconds-to-minutes of single-threaded CPU, stalling the worker.

## Details

`nltk/tokenize/casual.py`, the `URLS` regex naked-domain branch (~L116-127), compiled into `TweetTokenizer.WORD_RE` (L391-399) and applied via `WORD_RE.findall(safe_text)` in `TweetTokenizer.tokenize()` (L382). The branch is essentially:

```
[a-z0-9]+(?:[.\-][a-z0-9]+)*[.][a-z]{2,13}\b
```

The prefix `[a-z0-9]+(?:[.\-][a-z0-9]+)*` can partition a run like `a.a.a.a…` in exponentially many ways; when the mandatory trailing TLD (`[.][a-z]{2,13}\b`) cannot match at the end, the engine backtracks across every partition before failing at each offset → catastrophic backtracking. The only pre-processing in `tokenize()` (`HANG_RE.sub`) does not collapse the alternating `a.a.a` payload.

## PoC (validated, public API)

```python
from nltk.tokenize import TweetTokenizer
TweetTokenizer().tokenize("a" + ".a"*1800)   # ~3.6 KB
```
Timings (single thread): 2001 chars → 0.40s; 3601 → 1.54s; 4001 → 3.30s; 5001 → 6.31s; 6001 → 9.66s (≈4–6× per doubling — quadratic+). ~20 KB runs for minutes. Reproduced through `TweetTokenizer.tokenize()`, module-level `casual_tokenize()`, and the isolated `URLS` sub-pattern.

## Impact

Any NLP service that tokenizes user-submitted text with `TweetTokenizer`/`casual_tokenize` (its intended use) can be stalled per-request by a few-KB input — availability DoS, no auth, first call.

## Net-new

Distinct from CVE-2024-39705 (pickle/punkt RCE) and the CVE-2021-3828/3842 ReDoS fixes, which were all in `PunktSentenceTokenizer`/a corpus reader. `casual.py`'s `URLS` pattern was never remediated and ships unchanged in 3.9.4. (The data-downloader zip extraction is freshly hardened — CVE-2025-14009 — and was ruled out.)

## Remediation

Anchor/bound the naked-domain branch (e.g. atomic groups / possessive quantifiers, or cap domain-label repetition), or use the `regex` module's `POSSESSIVE`/timeout features; reject pathological label counts before matching.

prabhashj07:
### Problem
Two branches of the `URLS` regex in `nltk/tokenize/casual.py` had
unbounded nested quantifiers vulnerable to catastrophic backtracking
(CWE-1333). A few KB of adversarial input (e.g. `"a" + ".a" * 1800`)
could stall a worker for seconds to minutes:

| Input size | Time (before) |
|---|---|
| 2,001 chars | 0.40s |
| 3,601 chars | 1.54s |
| 6,001 chars | 9.66s |
| ~20,000 chars | minutes |

### Fix
- **HTTP-domain branch** (`[a-z0-9.\-]+[.]`) → wrapped in an atomic
  group: `(?>[a-z0-9.\-]+)[.]`. Safe because this branch requires a
  mandatory trailing `/`, so no backtracking-for-correctness is lost.
- **Naked-domain branch** (`[a-z0-9]+(?:[.\-][a-z0-9]+)*`) → bounded
  to `{0,20}` repetitions: `[a-z0-9]+(?:[.\-][a-z0-9]+){0,20}`.

  An atomic group was tried here first and reverted — it broke
  ordinary domain matching. The mandatory trailing `[.][a-z]{2,13}`
  needs the repeated group to give back its last repetition so the
  final label can double as the TLD (e.g. for `"example.com"`, greedy
  matching consumes `.com` inside the repeated group, leaving nothing
  for the required suffix, unless one repetition is backtracked).
  Atomic grouping forbids that backoff entirely, silently splitting
  every two-label domain into fragments. Bounding the repetition
  instead of eliminating it keeps that backtracking intact while
  capping the worst case — no real domain has anywhere near 20 labels.

No capturing groups are added, so `findall()`'s return type is
unaffected and `tokenize()` still returns `list[str]`.

### Timing (after fix, default `TweetTokenizer()`)

| Input size | Time |
|---|---|
| 500 | 0.026s |
| 1,800 | 0.201s |
| 5,000 | 1.522s |
| 20,000 | 22.4s |

Growth is now polynomial (O(n²), inherent to `findall` retrying every
offset) rather than the steeper super-quadratic blowup before the fix.
This is acceptable for TweetTokenizer's expected input size (tweets),
but callers feeding it much larger arbitrary text should be aware the
underlying `findall`-over-every-offset cost isn't eliminated, just
the catastrophic component.

### Testing
Added `TestTweetTokenizerReDoS` (9 tests) to `test_tokenize.py`:
- Timed pathological-input tests at 500/1,800/5,000 chars, run in
  spawned worker processes with hard timeouts so a regression can't
  hang CI.
- Exact-output regression tests: plain domains, hyphenated domains,
  sentences containing a domain, email addresses (not split), and
  specifically `example.com`/`google.com`/`python.org` as bare
  two-label domains — the exact case the atomic-group approach broke.
- A type-check test confirming `tokenize()` always returns `list[str]`.

All 9 new tests pass; existing suite unaffected (17 pre-existing
failures are unrelated missing-NLTK-data `LookupError`s present on
`develop` too).

### Diff
1 line changed in `nltk/tokenize/casual.py` (naked-domain bound) plus
the earlier atomic-group line for the HTTP branch; 1 test class added.

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -1242,9 +1242,9 @@ def _assert_tokenization_fast(text, timeout, label):
             f"TweetTokenizer.tokenize did not finish within {timeout}s "
             f"on {label} input -> ReDoS regression (CWE-1333, nltk#3704)"
         )
-    assert proc.exitcode == 0, (
-        f"worker crashed with exit code {proc.exitcode} on {label} input"
-    )
+    assert (
+        proc.exitcode == 0
+    ), f"worker crashed with exit code {proc.exitcode} on {label} input"
 
 
 class TestTweetTokenizerReDoS:
@@ -1272,8 +1272,11 @@ class TestTweetTokenizerReDoS:
         tknzr = TweetTokenizer()
         result = tknzr.tokenize("Visit a-b.example.io for more info")
         assert result == [
-            "Visit", "a-b.example.io",
-            "for", "more", "info",
+            "Visit",
+            "a-b.example.io",
+            "for",
+            "more",
+            "info",
         ]
 
     def test_email_not_split(self):
@@ -1287,13 +1290,17 @@ class TestTweetTokenizerReDoS:
     def test_tokenize_returns_list_of_str(self):
         """tokenize() must always return list[str], never list[tuple]."""
         tknzr = TweetTokenizer()
-        for text in ["hello world", "http://example.com", "user@bar.com",
-                      "a" + ".a" * 500]:
+        for text in [
+            "hello world",
+            "http://example.com",
+            "user@bar.com",
+            "a" + ".a" * 500,
+        ]:
             result = tknzr.tokenize(text)
             assert isinstance(result, list), f"expected list, got {type(result)}"
-            assert all(isinstance(t, str) for t in result), (
-                f"expected list[str], got list with non-str elements: {result[:5]}"
-            )
+            assert all(
+                isinstance(t, str) for t in result
+            ), f"expected list[str], got list with non-str elements: {result[:5]}"
 
     def test_naked_domain_with_hyphen(self):
         """Hyphenated labels in naked domains tokenize correctly."""
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -1192,3 +1192,120 @@ class TestTreebankWordDetokenizer:
         original_tokens = ["``", "Hello", ",", "''", "he", "said", "."]
         result = self.detok.detokenize(original_tokens)
         assert '"' in result
+
+
+# -------------------------------------------------------------------
+# Regression tests for ReDoS in TweetTokenizer URLS pattern
+# (CWE-1333; nltk/nltk#3704)
+#
+# The naked-domain branch of the URLS regex contained an unbounded
+# repetition that could split a string like "a.a.a.a...a" into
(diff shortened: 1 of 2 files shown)


ekaf:
Superseded by #3701.

PR #3708, opened 2026-07-21
Expand ~ in env-var paths (NLTK_DATA, find_file_iter)

### LINKED ISSUE #2825: Expand ~ in paths

Abhi-DevHub:
Fixes #2825

Apply \os.path.expanduser\ to environment-variable path entries so
that paths containing \~\ resolve to the user's home directory:

- \data.py\: expand \~\ in each \NLTK_DATA\ path entry
- \pathsec.py\: use \Path.expanduser()\ before \
esolve()\ on each
  \NLTK_DATA\ entry
- \internals.py\ \ind_file_iter\: expand \~\ in each env-var path
  component (already done in \ind_jar_iter\; this makes the other
  search functions consistent)

\ind_binary_iter\ is covered because it delegates to \ind_file_iter\.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -150,7 +150,7 @@ path = []
 
 # User-specified locations:
 _paths_from_env = os.environ.get("NLTK_DATA", "").split(os.pathsep)
-path += [d for d in _paths_from_env if d]
+path += [os.path.expanduser(d) for d in _paths_from_env if d]
 if "APPENGINE_RUNTIME" not in os.environ and os.path.expanduser("~/") != "~/":
     path.append(os.path.expanduser("~/nltk_data"))
 
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -628,6 +628,7 @@ def find_file_iter(
                 yield os.environ[env_var]
 
             for env_dir in os.environ[env_var].split(os.pathsep):
+                env_dir = os.path.expanduser(env_dir)
                 # Check if the environment variable contains a direct path to the bin
                 if os.path.isfile(env_dir):
                     if verbose:
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -623,9 +623,12 @@ def find_file_iter(
     # Check environment variables
     for env_var in env_vars:
         if env_var in os.environ:
-            if finding_dir:  # This is to file a directory instead of file
-                yielded = True
-                yield os.environ[env_var]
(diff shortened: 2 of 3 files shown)


ekaf:
Expanding ~ can turn a previously generic path into an absolute home-directory path. If these paths are ever surfaced in exceptions or debug output, this may disclose the local username or home directory structure. Probably low severity, but worth confirming that path-reporting code is acceptable.

ekaf:
Hi @Abhi-DevHub, what do you think about the path leak concern, and Copilot's review suggestion?

Abhi-DevHub:
@ekaf good point. A couple of considerations:

1. The ~/nltk_data fallback on the next line (data.py:155) already calls os.path.expanduser(~/nltk_data), so the home-directory path is already in 
ltk.data.path for any user who doesn't set NLTK_DATA at all.

2. The env-var paths in ind_file_iter/ind_jar_iter are only used internally for os.path.isfile()/os.listdir() lookups, not surfaced in error messages.

3. Do you have a specific code path in mind where the expanded path would appear in user-visible output? Happy to scope the change if needed.

Abhi-DevHub:
@ekaf thanks for the thorough review.

**Path leak concern**: The ~/nltk_data fallback on data.py:155 already calls os.path.expanduser(~/nltk_data), so the home directory path is already surfaced in 
ltk.data.path for any user on a non-Windows system — regardless of this PR. The env-var paths added here follow the same pattern and are only used internally for os.path.isfile()/os.listdir() lookups, not surfaced in error messages.

**Copilot finding_dir suggestion**: Good catch. When inding_dir=True, the old code yielded the raw env var string (e.g. \~/foo:~/bar\) without splitting or expanding. I've pushed a fix that iterates the split entries, expands each, and yields them individually — matching the behavior of the non-finding_dir path.

ekaf (approved):
This is a small, focused improvement that solves a real usability issue without introducing security problems beyond what's already accepted in the codebase. The author has thoughtfully addressed review feedback.

ekaf:
Thanks @Abhi-DevHub!

PR #3714, opened 2026-07-28
fix(security): validate types after wordnet_app pickle deserialization

LiteshGhute:
## Summary

Fixes the vulnerability described in [GHSA-7pvm-89gg-6c6h](https://github.com/nltk/nltk/security/advisories/GHSA-7pvm-89gg-6c6h): `nltk.app.wordnet_app`'s `lookup_<base64-pickle>` route unpickles attacker-controlled data via `RestrictedUnpickler`, which correctly blocks class/function reconstruction (so the RCE this replaced in #3100 stays fixed), but never checks the *type* of what it returns. Pickle's built-in `list`/`dict`/`int`/etc. opcodes don't go through the blocked path, so a single crafted request can return, say, an `int` where the code expects a `str`. `page_from_reference()` immediately calls `word.split(",")` assuming a `str`, so one unauthenticated GET request crashes the request handler with an uncaught `AttributeError` and drops the connection with no response.

## Changes

- `Reference.decode()` now validates that the unpickled `word` is a `str` and `synset_relations` is a `dict`, raising `ValueError` otherwise, instead of returning unchecked data.
- The `lookup_` route in `do_GET` catches that `ValueError` and returns a clean response instead of letting the exception propagate.
- Added `test_wordnet_app_reference_decode_rejects_wrong_types` to `nltk/test/unit/test_security.py`, covering the legitimate round-trip plus several malformed shapes (`int`, nested `list`, `None` for `word`; `list` for `synset_relations`).

## Test plan

- [x] Reproduced the crash pre-fix: a `lookup_<...>` request encoding `(42, {})` raised `AttributeError: 'int' object has no attribute 'split'` server-side and dropped the connection (`RemoteDisconnected`).
- [x] Post-fix: the same request returns a clean response; legitimate lookups (e.g. `word="dog"`) still work before and after the malformed request, confirming no regression.
- [x] `pytest nltk/test/unit/test_security.py` passes (both the existing test and the new one).
- [x] `pre-commit run --files nltk/app/wordnet_app.py nltk/test/unit/test_security.py` passes (black, isort, ruff, pyupgrade, pre-commit-hooks).

### DIFF
--- a/nltk/app/wordnet_app.py
+++ b/nltk/app/wordnet_app.py
@@ -726,15 +726,27 @@ class Reference:
         """
         Decode a reference encoded with Reference.encode
 
-        :raises ValueError: if the decoded data isn't shaped like a
-            genuine Reference (word: str, synset_relations: dict).
+        :raises ValueError: if the input isn't valid base64/pickle data, or
+            if the decoded data isn't shaped like a genuine Reference
+            (word: str, synset_relations: dict[str, set]).
             RestrictedUnpickler only blocks class/function reconstruction;
-            it does not guarantee the *type* of the objects it returns, so
-            the shape must still be checked before use.
+            it does not guarantee the *type* or *shape* of what it returns,
+            so that must still be checked before use. Any failure while
+            decoding, unpickling, or unpacking the payload is normalized to
+            ValueError so callers only need to guard against one exception
+            type.
         """
-        string = base64.urlsafe_b64decode(string.encode())
-        word, synset_relations = RestrictedUnpickler(io.BytesIO(string)).load()
+        try:
+            raw = base64.urlsafe_b64decode(string.encode())
+            word, synset_relations = RestrictedUnpickler(io.BytesIO(raw)).load()
+        except Exception as e:
+            raise ValueError("Malformed wordnet_app reference") from e
         if not isinstance(word, str) or not isinstance(synset_relations, dict):
+            raise ValueError("Malformed wordnet_app reference")
+        if not all(
+            isinstance(key, str) and isinstance(value, (set, frozenset))
+            for key, value in synset_relations.items()
+        ):
             raise ValueError("Malformed wordnet_app reference")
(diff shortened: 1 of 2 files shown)


LiteshGhute:
Hi @ekaf, both suggestions are addressed in the latest commit, normalized all decode/unpickle/unpack failures to ```ValueError```, and ```synset_relations``` values are now validated as actual mutable sets (also caught that a ```frozenset``` would slip through and still crash ```toggle_synset_relation()``` later, fixed that too). Re-tested end to end against a live server, no crashes on any of the payloads you flagged. Thanks!

ekaf (changes requested):
The security fix is good and the regression test is valuable, but Reference.decode() should normalize all base64/unpickle/unpack errors to ValueError, and it should validate that synset_relations values are set instances too. Otherwise some malformed lookup_... payloads can still crash the handler via exceptions other than ValueError or later in toggle_synset_relation().

LiteshGhute:
Hi @ekaf, this looks like it's already covered, your review is attached to commit `8c01d35c6`, which already has both of these:

- All base64/unpickle/unpack errors are normalized to `ValueError` here: https://github.com/LiteshGhute/nltk/blob/8c01d35c6c64d9b51f42c71b088ee8edefff3555/nltk/app/wordnet_app.py#L739-L743
- `synset_relations` values are validated as actual `set` instances (not just that the container is a dict) here: https://github.com/LiteshGhute/nltk/blob/8c01d35c6c64d9b51f42c71b088ee8edefff3555/nltk/app/wordnet_app.py#L746-L752

Could you please take another look at the diff on that commit? Thanks for the review!

ekaf (approved):
Thanks for the follow-up — this addresses the earlier concerns.

Reference.decode() now normalizes decode/unpickle/unpack failures to ValueError, and the extra validation on synset_relations values closes the remaining crash paths I was worried about. The expanded regression test coverage is also solid.

PR #3715, opened 2026-07-29
Fix uncontrolled search path in HunposTagger

ekaf:
### Description
This PR removes the current working directory (`"."`) from the default binary search path in `HunposTagger` to enforce a restrictive execution policy. 
### Security Motivation
This addresses a  (CWE-427: Uncontrolled Search Path Element). vulnerability in HunposTagger. Previously, `hunpos_paths` implicitly trusted the current working directory. If a user ran an NLTK script in a shared or attacker-influenced directory containing a malicious executable named `hunpos-tag`, `find_binary()` would resolve and execute it via `subprocess.Popen`.

### Impact
*   **Secures binary resolution:** `find_binary()` will no longer search the current working directory (`.`). It now strictly limits the search to system paths (e.g., `/usr/bin`) and user-owned home directories (e.g., `~/bin`), alongside explicit environment variables.
*   **Backwards compatibility:** Users who legitimately need to run a `hunpos-tag` binary located in their current working directory can still do so by explicitly declaring it during initialization: `HunposTagger(path_to_model='...', path_to_bin='./hunpos-tag')`.

### DIFF
--- a/nltk/tag/hunpos.py
+++ b/nltk/tag/hunpos.py
@@ -71,7 +71,6 @@ class HunposTagger(TaggerI):
         """
         self._closed = True
         hunpos_paths = [
-            ".",
             "/usr/bin",
             "/usr/local/bin",
             "/opt/local/bin",


ekaf:
From a security perspective, ~/bin is safe because an attacker would already need write access to the user's home directory to exploit it (unlike ., which blindly trusts any shared or untrusted execution context). I have updated the PR description to accurately reflect that user-owned directories are preserved in the search path."

LGTM — the security hardening looks good, and the change is minimal.

PR #3723, opened 2026-07-30
detokenize periods mid-string, not just the last one

tarann26:
`TreebankWordDetokenizer().detokenize()` left a space before standalone period tokens that appear mid-string. For example `detokenize(["...","amet",".","consectetur",...])` produced `"amet . consectetur"` instead of `"amet. consectetur"` — only the final period was joined.

Root cause: the period-joining rule in `PUNCTUATION` was anchored with `\s*$`, so it only matched a period at the very end of the string. Dropping the `$` anchor lets it join every standalone period.

To leave ellipsis tokens alone, the pattern gains a `(?!\.)` guard so it never matches the first dot of a `...` run (those are collapsed separately by the `\s\.\.\.\s` rule). The rule still requires a space before the period, so attached periods in decimals (3.14), money ($3.88), versions (2.0.) and abbreviations (Dr., p.) are untouched.

Adds regression tests for mid-string periods, multiple sentence-final periods, and the ellipsis guard.

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -1240,3 +1240,44 @@ class TestTreebankWordDetokenizer:
         original_tokens = ["``", "Hello", ",", "''", "he", "said", "."]
         result = self.detok.detokenize(original_tokens)
         assert '"' in result
+
+    def test_issue_3210_mid_string_period(self):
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @tarann26, great first contribution.

PR #3724, opened 2026-07-30
don't collapse repeated non-Latin letters in TweetTokenizer

### LINKED ISSUE #3157: Uncontrolable token len reducing in Cyrillic texts
Good afternoon.

This item, regardless of my desire, will shorten too long sequences of not Latin characters.

https://github.com/nltk/nltk/blob/6b0312a41e9403111c9cfb5673947df7416b1f43/nltk/tokenize/casual.py#L297

tknzr = TweetTokenizer(**reduce_len=False**)

tknzr("Loooool") == ["Loooool"]  Latin symbols OK
tknzr("Лооооол") != ["Лооооол"]  Cyrillic symbols Failed
tknzr("Лооооол") == ["Лооол"]

tarann26:
Fixes #3157.

`TweetTokenizer.tokenize` always runs `HANG_RE` to collapse long runs of repeated characters before the main word regex, guarding against catastrophic backtracking. Its character class was `[^a-zA-Z0-9]`, which treats every non-ASCII character as "dangerous", so runs of repeated non-Latin letters (e.g. Cyrillic `Лооооол`) were shortened even with `reduce_len=False`, corrupting the text.

Widen the preserved set to all Unicode letters and numbers:

```
-HANG_RE = regex.compile(r"([^a-zA-Z0-9])\1{3,}")
+HANG_RE = regex.compile(r"([^\p{L}\p{N}])\1{3,}")
```

Letters and digits of any script are now preserved, while symbols, punctuation and combining marks (none of which are `\p{L}`/`\p{N}`) are still collapsed, so the guard against pathological repeated-symbol input is unchanged. `reduce_len=True` continues to shorten lengthening in every script. Added a regression test in `test_tokenize.py`.

### DIFF
--- a/nltk/test/unit/test_tokenize.py
+++ b/nltk/test/unit/test_tokenize.py
@@ -383,6 +383,28 @@ class TestTokenize:
             )
         assert proc.exitcode == 0, f"worker failed (exit code {proc.exitcode})"
 
+    def test_tweet_tokenizer_reduce_len_preserves_non_latin(self):
+        """
+        With ``reduce_len=False`` (the default), the tokenizer must not shorten
+        runs of repeated non-Latin letters. Regression test for
+        https://github.com/nltk/nltk/issues/3157, where a run of Cyrillic
+        characters was collapsed even though Latin ones were left untouched.
+        """
+        keep = TweetTokenizer(reduce_len=False)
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @tarann26

PR #3725, opened 2026-07-31
Guard tempfile.gettempdir() when building pathsec allowed roots

### LINKED ISSUE #3716: nltk.sent_tokenize() fails with FileNotFoundError on read-only filesystems, even when the tokenizer data is already cached
# Summary
nltk.pathsec._get_allowed_roots() builds its list of permitted data directories in two steps: first from nltk.data.path (each entry individually protected against errors), then from three fallback candidates — ~/nltk_data, /usr/share/nltk_data, and tempfile.gettempdir(). The third one is evaluated before the loop that protects the other two even starts, as part of building the input list itself:

for loc in ["~/nltk_data", "/usr/share/nltk_data", tempfile.gettempdir()]:
    try:
        p = Path(loc).expanduser().resolve()
        if p.exists():
            roots.add(p)
    except (OSError, ValueError, RuntimeError):
        continue
tempfile.gettempdir() raises whenever no writable temporary directory can be found — a real, documented condition on any environment with a read-only root filesystem and nothing mounted at /tmp (containers, some hardened deployments). When it raises here, the exception isn't caught by the loop's try/except — it happens before the loop begins — so it propagates out of _get_allowed_roots() entirely, discarding the result of the first step, which had already completed successfully and already included the correct, valid data directory.

# Impact
Any operation that needs to actually open a resource file (not just locate it) fails under this condition, even when that exact resource is already correctly present in a directory already listed in nltk.data.path. Concretely: nltk.data.find("tokenizers/punkt_tab") succeeds, immediately followed by nltk.tokenize.sent_tokenize(...) on the same environment failing with FileNotFoundError, on the same resource find() just located. find() doesn't need to open the file, so it doesn't hit this path; sent_tokenize() does, so it does.

# Minimal reproduction

import nltk, unittest.mock as mock

#Resource is genuinely present and locatable:
print(nltk.data.find("tokenizers/punkt_tab"))  # succeeds

#Simulate no writable temp directory anywhere (e.g. a read-only container
#root filesystem with nothing mounted at /tmp):
with mock.patch("tempfile.gettempdir", side_effect=FileNotFoundError(
    "No usable temporary directory found"
)):
    import nltk.pathsec as pathsec
    pathsec._ALLOWED_ROOTS_CACHE = None  # force recomputation
    pathsec._LAST_DATA_PATHS = None

    nltk.data.find("tokenizers/punkt_tab")   # still succeeds

    from nltk.tokenize import sent_tokenize
    sent_tokenize("Hello there. How are you?")  # raises FileNotFoundError,
    #on the exact same resource just located successfully above
# Versions affected
The _get_allowed_roots() implementation above is byte-for-byte identical in nltk 3.9.4 and 3.10.0 — this isn't new in 3.10. The only relevant difference between the two versions is that pathsec.ENFORCE defaults to False in 3.9.4 and True in 3.10.0, so in 3.9.4 the failure is silently swallowed instead of raised — meaning the same underlying computation fails the same way on both versions, just visibly on one and invisibly on the other.

tarann26:
Fixes #3716.

`pathsec._get_allowed_roots()` builds its fallback directory list as a literal:

```python
for loc in ["~/nltk_data", "/usr/share/nltk_data", tempfile.gettempdir()]:
    try:
        p = Path(loc).expanduser().resolve()
        if p.exists():
            roots.add(p)
    except (OSError, ValueError, RuntimeError):
        continue
```

`tempfile.gettempdir()` is evaluated while the list is being constructed, before the loop's `try`/`except` runs. On a host with no usable temporary directory it raises `FileNotFoundError` (a subclass of `OSError`), which escapes `_get_allowed_roots()` entirely and throws away the roots already collected from `nltk.data.path` / `NLTK_DATA` earlier in the function. The exception propagates up through `validate_path`, so `sent_tokenize()` and other resource lookups fail even when the resource is already cached and `find()` located it successfully.

This moves the `gettempdir()` call under its own guard, so a failure just skips the temp-dir candidate instead of aborting the whole lookup:

```python
candidate_locs = ["~/nltk_data", "/usr/share/nltk_data"]
try:
    candidate_locs.append(tempfile.gettempdir())
except (OSError, ValueError, RuntimeError):
    pass

for loc in candidate_locs:
    ...
```

Behavior is unchanged when a temporary directory exists (the candidate list is identical to before), and the guard uses the same exception tuple already used inside the loop. Adds a regression test that patches `tempfile.gettempdir` to raise `FileNotFoundError` and asserts the allowed-roots lookup still returns the configured data root.

### DIFF
--- a/nltk/pathsec.py
+++ b/nltk/pathsec.py
@@ -56,7 +56,13 @@ def _get_allowed_roots():
 
     import tempfile
 
-    for loc in ["~/nltk_data", "/usr/share/nltk_data", tempfile.gettempdir()]:
+    candidate_locs = ["~/nltk_data", "/usr/share/nltk_data"]
+    try:
+        candidate_locs.append(tempfile.gettempdir())
+    except (OSError, ValueError, RuntimeError):
+        pass
+
+    for loc in candidate_locs:
         try:
             p = Path(loc).expanduser().resolve()
             if p.exists():
--- a/nltk/test/unit/test_pathsec.py
+++ b/nltk/test/unit/test_pathsec.py
@@ -144,6 +144,39 @@ def test_path_traversal_absolute():
         pathsec.open(outside, "r")
 
 
+# --- ALLOWED-ROOTS / TEMP-DIR FALLBACK TESTS ---
+
+
+def test_get_allowed_roots_survives_missing_tempdir(tmp_path, monkeypatch):
+    """Regression test for issue #3716.
+
+    ``_get_allowed_roots()`` used to build its fallback-location list as a
+    literal ``[..., tempfile.gettempdir()]``, which evaluates
+    ``tempfile.gettempdir()`` while constructing the list -- *before* the
+    loop's ``try/except`` runs. On a system with no usable temp directory
+    (read-only root filesystem, nothing mounted at ``/tmp``), ``gettempdir()``
+    raises ``FileNotFoundError`` (an ``OSError`` subclass) that propagates out
+    of the whole function, discarding the roots already collected from
+    ``nltk.data.path``/``NLTK_DATA`` and breaking resource lookups (e.g.
+    ``sent_tokenize()``) even when the resource is already cached.
+    """
+    import nltk.data
+
+    known_root = tmp_path / "nltk_data_known_root"
+    known_root.mkdir()
+
+    monkeypatch.setattr(nltk.data, "path", nltk.data.path + [str(known_root)])
+
+    # Force a clean cache: _get_allowed_roots() memoizes on (data.path, NLTK_DATA).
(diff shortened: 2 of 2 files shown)


ekaf (approved):
The latest code convincingly fixes the failure mode by moving tempfile.gettempdir() out of the list literal and into a guarded try/except before iterating candidate roots in nltk/pathsec.py, which prevents _get_allowed_roots() from aborting on hosts without a usable temp dir; the new regression test in nltk/test/unit/test_pathsec.py directly simulates FileNotFoundError and verifies configured roots remain usable. This should be merged soon because it improves reliability of path security/resource lookup in constrained environments with very low regression risk.

PR #3726, opened 2026-07-31
fix(security): close symlink escape in FramenetCorpusReader (CWE-59)

LiteshGhute:
## Summary

Reported as [GHSA-f833-7jw8-xwrv](https://github.com/nltk/nltk/security/advisories/GHSA-f833-7jw8-xwrv).

`frame_by_name()`, `_lu_file()` and `doc()` reject unsafe characters in the caller- or corpus-supplied name via `_reject_unsafe_path_component()`, then resolve the path through `self.abspath()`. That guard only inspects the name string; it never resolves symlinks. A symlink planted inside the corpus subdirectory (`frame/`, `lu/` or `fulltext/`), with a name containing no separators or `..`, passes the guard cleanly, and `self.abspath()` (a plain `self._root.join(fileid)`) does not catch it either, so the file is opened from outside the corpus root anyway.

## PoC (before this change)

```python
import os
from nltk.corpus.reader.framenet import FramenetCorpusReader

os.symlink("/anywhere/outside/the/corpus/root/stolen.xml",
           "<corpus_root>/frame/evil_link.xml")

reader = FramenetCorpusReader("<corpus_root>", [])
result = reader.frame_by_name("evil_link")   # normal, routine call, no ".." anywhere
print(result["definition"])                   # content read from outside the corpus root
```

The same pattern applies to `_lu_file()` (`lu<id>.xml` under `lu/`) and `doc()` (arbitrary filename under `fulltext/`).

## Fix

Add `_validate_in_root()`, which calls `nltk.pathsec.validate_path()` with the corpus root as `required_root` at all three call sites. This is the same symlink-resolving, root-scoped containment check already used by `CorpusReader.open()` and `NKJPCorpusReader.add_root()`: both the candidate path and the root are resolved with `Path.resolve()` before the containment comparison, so a symlink cannot escape undetected.

## Tests

Adds three regression tests to `test_framenet_security.py`, one per call site, each planting a symlink inside the relevant corpus subdirectory and asserting the escape is rejected with no file outside the corpus root ever opened. The existing traversal and drive-prefix tests are unchanged and still pass.

## Validation

- Verified end to end against unpatched `develop`, with `ENFORCE=True`: all three call sites open the file outside the corpus root and return its content to the caller.
- After the fix, the same PoC and the new regression tests are blocked with a clear `Security Violation ... escapes root` error.
- Ran the full `test_framenet_security.py`, `test_corpus_reader.py`, `test_data_security.py` and `test_pathsec.py` suites in two independent fresh virtual environments (Python 3.12 and 3.13): all pass, no regressions.
- Tried chained symlinks, a symlinked whole subdirectory, NUL byte injection, and Unicode normalization tricks against the fixed code; all still rejected. The one known limitation is a theoretical TOCTOU window between validation and the actual open, which is an existing architectural property shared by `CorpusReader.open()` itself, not something introduced by this change.
- Ran `pre-commit` on the changed files (black, isort, ruff, pyupgrade): clean.

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -779,6 +779,26 @@ def _reject_unsafe_path_component(value, kind):
         raise FramenetError(f"Invalid {kind}: {value!r}")
 
 
+def _validate_in_root(locpath, root, context):
+    """Reject a resolved path that escapes the corpus root through a symlink.
+
+    ``_reject_unsafe_path_component`` only rejects unsafe characters in the
+    caller-/corpus-supplied name itself; it never resolves symlinks. A
+    symlink planted inside the corpus subdirectory (``frame/``, ``lu/`` or
+    ``fulltext/``) can still point outside the corpus root even when the name
+    referencing it contains no separator or ``..`` at all.
+
+    This calls ``nltk.pathsec.validate_path`` with the corpus root as
+    ``required_root``, the same symlink-resolving containment guard used by
+    ``CorpusReader.open()`` and ``NKJPCorpusReader.add_root()``: both
+    ``locpath`` and ``root`` are resolved with ``Path.resolve()`` before the
+    containment check, so a symlink cannot escape undetected.
+    """
+    from nltk.pathsec import validate_path
+
+    validate_path(locpath, context=context, required_root=root)
+
+
 class AttrDict(dict):
     """A class that wraps a dict and allows accessing the keys of the
     dict as if they were attributes. Taken from here:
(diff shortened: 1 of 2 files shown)


ekaf (approved):
The latest commit definitively closes the Framenet symlink escape: nltk/corpus/reader/framenet.py now adds _validate_in_root() and calls validate_path(..., required_root=self.root) at all three vulnerable read paths (doc, frame_by_name, _lu_file), and nltk/test/unit/test_framenet_security.py adds dedicated symlink-escape regressions for each call site to ensure out-of-root targets are blocked before open. This is a critical security merge for NLTK (CWE-59 class), and it should be merged with high priority.

PR #3727, opened 2026-07-31
fix(security): close symlink escape in IPIPANCorpusReader (CWE-59)

LiteshGhute:
## Summary

Reported as [GHSA-3hhw-38pf-pxj6](https://github.com/nltk/nltk/security/advisories/GHSA-3hhw-38pf-pxj6).

`channels()`, `domains()`, `categories()` and `fileids(channels=...)` route through `_get_tag()`, which calls the builtin `open()` on a plain string produced by `.replace("morph.xml", "header.xml")` on the result of `self.abspath()`. Since `FileSystemPathPointer` subclasses `str`, that `.replace()` call silently discards the `PathPointer` wrapper, so the resulting `open()` never goes through `nltk.pathsec` at all, not even the global, non-scoped check. A symlink planted inside the corpus root, with a name containing no separators or `..`, passes the existing traversal guard cleanly and can point anywhere on the filesystem the process can read.

This is a more severe variant of the same CWE-59 class already fixed for `CorpusReader.open()`, `NKJPCorpusReader.add_root()`, and `FramenetCorpusReader` (#3726): those still route through the global `pathsec` check at minimum. Here `pathsec` is skipped entirely.

## PoC (before this change)

```python
import os
import tempfile

from nltk.corpus.reader.ipipan import IPIPANCorpusReader

root = tempfile.mkdtemp()
corpus_root = os.path.join(root, "ipipan")
os.makedirs(corpus_root)

with open(os.path.join(corpus_root, "real_morph.xml"), "w") as f:
    f.write("<channel>legit</channel>")

secret_dir = os.path.join(root, "outside_ipipan_root")
os.makedirs(secret_dir)
secret_path = os.path.join(secret_dir, "stolen.xml")
with open(secret_path, "w") as f:
    f.write("<channel>TOP-SECRET-CHANNEL-DATA-FROM-OUTSIDE-CORPUS-ROOT</channel>")

os.symlink(secret_path, os.path.join(corpus_root, "evil_link.xml"))

reader = IPIPANCorpusReader(corpus_root, r".*\.xml")  # normal regex auto-discovery
print(reader.channels(fileids=["evil_link.xml"]))
# ['TOP-SECRET-CHANNEL-DATA-FROM-OUTSIDE-CORPUS-ROOT']
```

Literal `../` traversal in the fileid is already rejected by `FileSystemPathPointer.join()`, so this is specifically the symlink variant.

## Fix

Call `nltk.pathsec.validate_path()` with the corpus root as `required_root` inside `_get_tag()` itself, the single choke point all four affected methods route through, before the `open()` call. Same symlink-resolving, root-scoped containment check already used by `CorpusReader.open()` and `NKJPCorpusReader.add_root()`.

## Tests

Adds `nltk/test/unit/test_ipipan_security.py`: a legitimate in-root read still works, literal `../` traversal is still rejected, and a symlink planted inside the corpus root, discovered via normal regex-based auto-discovery rather than hand-picked at construction time, is rejected for `channels()`, `domains()` and `fileids(channels=...)`.

## Validation

- Verified end to end against unpatched `develop`: `channels()` opens the file outside the corpus root and returns its content to the caller, with the reader constructed the normal way (regex auto-discovery), not a contrived setup.
- After the fix, the same PoC and the new regression tests are blocked with a clear `Security Violation ... escapes root` error.
- Ran the full `test_ipipan_security.py`, `test_corpus_reader.py`, `test_data_security.py`, `test_pathsec.py` and `test_framenet_security.py` suites in two independent fresh virtual environments (Python 3.12 and 3.13): all pass, no regressions.
- Tried chained symlinks, a symlinked subdirectory with a subpath fileid, NUL byte injection, a drive-qualified name, and an absolute path fileid against the fixed code; all rejected. Also confirmed a symlink that legitimately resolves inside the corpus root still works normally, so the fix does not break legitimate in-root symlinks.
- Ran `pre-commit` on the changed files (black, isort, ruff, pyupgrade): clean.

### DIFF
--- a/nltk/corpus/reader/ipipan.py
+++ b/nltk/corpus/reader/ipipan.py
@@ -187,6 +187,24 @@ class IPIPANCorpusReader(CorpusReader):
         return list(ret_fileids)
 
     def _get_tag(self, f, tag):
+        # Security (CWE-22 / CWE-59): ``f`` reaches this point as a plain str,
+        # not a PathPointer -- ``_list_header_files`` / ``_list_morph_files_by``
+        # call ``.replace("morph.xml", "header.xml")`` on the result of
+        # ``self.abspath()``/``self.abspaths()``, and since
+        # ``FileSystemPathPointer`` subclasses ``str``, ``.replace()`` returns
+        # a plain string, silently discarding the PathPointer wrapper. That
+        # meant this ``open()`` call never went through nltk.pathsec at all,
+        # not even the global check that a PathPointer.open() would apply, so
+        # a symlink planted inside the corpus root (with a name containing no
+        # separators or "..", so it passes FileSystemPathPointer.join()
+        # cleanly) could point anywhere on the filesystem and still be
+        # opened. Validate the resolved path against the corpus root before
+        # opening, the same symlink-resolving containment guard
+        # CorpusReader.open() and NKJPCorpusReader use.
+        from nltk.pathsec import validate_path
+
+        validate_path(f, context="IPIPANCorpusReader", required_root=self.root)
+
         tags = []
         with open(f) as infile:
             header = infile.read()
--- a/nltk/test/unit/test_ipipan_security.py
+++ b/nltk/test/unit/test_ipipan_security.py
@@ -94,6 +94,24 @@ def test_ipipan_domains_rejects_symlink_escape(tmp_path):
         reader.domains(fileids=["evil_link2.xml"])
 
 
+def test_ipipan_categories_rejects_symlink_escape(tmp_path):
+    """categories() also shares _get_tag() with channels(); confirm it is
+    covered too. It post-processes _parse_header()'s result through
+    _map_category(), but the ValueError must fire before that ever runs."""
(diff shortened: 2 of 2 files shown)


LiteshGhute:
Thanks for catching that, `categories()` shares `_get_tag()` with `channels()` and `domains()`, so it needed the same coverage. Added `test_ipipan_categories_rejects_symlink_escape` in the latest commit, confirms the containment check stays enforced there too.

ekaf (approved):
The latest code fixes the IPIPAN escape at the true choke point: _get_tag() in nltk/corpus/reader/ipipan.py now validates the candidate file path with validate_path(f, required_root=self.root) before builtin open(), eliminating the prior bypass caused by PathPointer-to-str conversion via .replace(...); comprehensive new tests in nltk/test/unit/test_ipipan_security.py cover legitimate reads, traversal rejection, and symlink escape blocking across channels, domains, categories, and fileids(channels=...). This is also a high-priority security merge and should be landed promptly.

PR #3734, opened 2026-08-04
Skip draft step in release workflow

ekaf:
The release workflow has always worked well, except that approving the draft turned out to be a little annoying.

So this PR modifies the release.yml workflow to publish the releases directly on Github, skipping the draft step.


ekaf:
This is not security-critical, but it is worthwhile operationally for release flow efficiency, so merge priority is medium.

PR #3434, opened 2025-10-26
test(stopwords): acceptance test for English contractions + slang (Issue #3047)

irfan416:
## Summary
Adds a customer-acceptance test that verifies that the English stopwords include common contractions (both straight and smart quotes) and colloquial forms.  
This test is paired with the data update in **irfan416/nltk_data** (Issue #3047).

The test allows reviewers to easily switch between the **original** and the **modified** stopword lists using the environment variable `STOPWORDS_VERSION`.

---

## What this test does
**File:** `nltk/test/unit/test_stopwords_contractions.py`

- Loads `stopwords.words("english")` and checks for representative items.  
- **Contractions:** `i'm`, `i’m`, `you’re`, `you're`, `it’s`, `it's`, `don’t`, `don't`, `can’t`, `can't`, `won’t`, `won't`, etc.  
- **Colloquialisms:** `gonna`, `wanna`, `gotta`, `lemme`, `gimme`, `coulda`, `shoulda`, `woulda`.  
- Prints clear messages identifying which stopword file version was used.  
- Passes with the modified list; fails with the original list.

---

## How to run this test (Linux/macOS)

```bash
# 1. Clone both repos side by side
git clone https://github.com/irfan416/nltk_data
git clone https://github.com/irfan416/nltk
cd nltk

# 2. Create and activate a virtual environment
python -m venv .venv && source .venv/bin/activate
pip install -e .
pip install pytest

# 3. Point NLTK to the local data path
export NLTK_DATA="$(pwd)/../nltk_data/packages"

# 4. Run test with ORIGINAL stopword list (expected FAIL)
export STOPWORDS_VERSION=english
pytest -q -s nltk/test/unit/test_stopwords_contractions.py

# 5. Run test with MODIFIED stopword list (expected PASS)
export STOPWORDS_VERSION=modified
pytest -q -s nltk/test/unit/test_stopwords_contractions.py
```

## How to run (Windows PowerShell)

```bash 
# 1. Clone both repos side by side
git clone https://github.com/irfan416/nltk_data
git clone https://github.com/irfan416/nltk
Set-Location .\nltk

# 2. Set up environment
python -m venv .venv
.\.venv\Scripts\Activate.ps1
pip install -e .
pip install pytest

# 3. Point NLTK to your data
$env:NLTK_DATA = "$((Get-Location).Path)\..\nltk_data\packages"

# 4. ORIGINAL (expected FAIL)
$env:STOPWORDS_VERSION = "english"
[...]

Issue #3358, opened 2025-01-15
To check if I'm correct to add i'm in the stopword library


zaib-shekh:
**i'm** is not included in the library of stopword.words('english').
 I think it is also a stop word. Please take a look.

WilliamPLaCroix:
First, this issue might be best opened in the [NLTK_Data repositoy](https://github.com/nltk/nltk_data/issues), since that's where the stopword corpora are stored and managed.

Second, the current NLTK English stopword corpus seems fairly minimal, with 179 entries. As you noted (also documented in [this paper from 2018](https://aclanthology.org/W18-2502/)) many contractions are not included [i'm, i've, i'd], even though some of their other pronoun counterparts are [you're, you'll, you'd].

One minimal solution would be to add the corresponding pronoun contractions to the stopword list.

A more comprehensive solution would be to expand the stopword list along the lines of [this English stopword list](https://github.com/stopwords-iso/stopwords-en), which includes 1298 entries, compiled from a variety of sources.

As noted in [Nothman, Qin, Yurchak 2018,](https://aclanthology.org/W18-2502/) stopword lists are somewhat subjective to the task at hand, or even domain-specific, so it's not necessarily as simple a solution as just creating/pulling a full list from some resource. That said, I would be hesitant to blindly include the full list from [the stopwords-en page](https://github.com/stopwords-iso/stopwords-en/blob/master/stopwords-en.txt), and would lean toward the minimal solution for now.

I am creating a PR on nltk_data with an update to the English stopwords corpus to include contractions for personal pronouns.

stevenbird:
Resolved.

Issue #3073, opened 2022-11-22
Stop words - gender omission 


paulellis65:
Name: nltk
Version: 3.4.5
Summary: Natural Language Toolkit

Looking at the gender words contained in the above stop words 'English'

his - hers
himself - herself
him - her
he - she / she's

The list appears to omit the word he's

Is there any reason for this? Quite a fundamental omission. 

Thanks, 

Paul

tomaarsen:
Hello!

This is because a common usage is to pass the input text through `word_tokenize`, which converts e.g. `he's` into `["he", "'s"]`, after which the `"he"` can be stripped. A common usage that I see is:
```python
from nltk.corpus import stopwords
from nltk.tokenize import word_tokenize

text = "He's eating an apple over there."
stop_words = set(stopwords.words())
tokenized = word_tokenize(text)
# ['He', "'s", 'eating', 'an', 'apple', 'over', 'there']
cleaned = [word for word in tokenized if word.lower() not in stop_words]
# ["'s", 'eating', 'apple']
```
As a result, `"he"` can be stripped regardless of the contraction in which it is used (`"he'd"`, `"he'll"`, etc). A better question might be why `she's` **is** in the stopwords list, as it seems a bit unnecessary. Perhaps some additional consistency (either both or neither) would be best here, but I don't know if I'm leaning towards both or neither.

JEnduriumK:
@tomaarsen 

Wait, I thought that you _shouldn't_ be removing stop words from things prior to analyzing them with VADER. 

For example, I believe that the following words are in English `stopwords` file from `nltk.corpus`:

```
the
of
to
for
```

Which, unless I'm _wildly_ misunderstanding how VADER works, definitely **shouldn't** be removed before handing the sentence or text to VADER, because of the `special_cases` involving idioms that is built into it.

https://github.com/nltk/nltk/blob/b0e85694107992e00a2f9fb48e6410c50fe1f1f6/nltk/sentiment/vader.py#L184

You remove those words before handing it text, and you break its ability to detect those idioms. Right?

Other sites also confirm that VADER is intended to be used _without_ processing content ahead of time, such as [here](https://towardsdatascience.com/an-short-introduction-to-vader-3f3860208d53).

> The incredible thing about VADER is it doesn’t require a great deal of preprocessing to work. Unlike with some supervised methods of NLP, preprocessing necessities such as tokenisation and stemming/lemmatisation are not required. You can pretty much plug in any body of text and it will determine the sentiment.

[...]

Issue #3047, opened 2022-09-03
Add more English short forms as stop words


BLKSerene:
As discussed in #1800 and #2588, some short/negative/tense forms are missing from the list of English stop words:

`i'm, i've, i'll, i'd` (to be consistent with contractions of other pronouns)
`cannot, could've, mayn, mayn't` (to be consistent with contractions of other modal verbs)
`ought, oughtn, oughtn't, dare, daren, daren't` (two more modal verbs)

I'm not sure whether I should open an issue or PR to `nltk` or to `nltk_data`, so I just paste the updated list here...

### Notes:
1. Only pronouns and modal verbs are modified.
2. Some words are re-ordered to make the structure clearer.
3. Not all possible short forms are added. E.g. `amn't, how's, when's`, and some others which are dialectal or too informal. I simply check the MW dictionary and if there is an entry for this form and it is not chiefly used dialectally, then it is added here.

```
i
i'm
i've
i'll
i'd
me
my
myself
we
we're
we've
we'll
we'd
our
ours
ourselves
you
you're
you've
you'll
you'd
your
yours
yourself
yourselves
he
he's
he'll
he'd
him
his
himself
she
she's
she'll
she'd
her
hers
herself
it
it's
it'll
it'd
its
itself
they
they're
they've
they'll
they'd
them
their
theirs
themselves
what
what's
which
who
who's
who've
who'll
who'd
whom
when
where
why
how
this
that
that's
that'll
that'd
these
those
here
there
there's
there'll
there'd
am
is
isn
isn't
are
aren
aren't
was
wasn
wasn't
were
weren
weren't
be
been
being
have
haven
haven't
has
hasn
hasn't
had
hadn
hadn't
having
ain
ain't
do
don
don't
does
doesn
doesn't
did
didn
didn't
done
doing
will
won
won't
would
wouldn
wouldn't
would've
shall
shan
shan't
can
cannot
could
couldn
couldn't
could've
may
mayn
mayn't
might
mightn
mightn't
might've
must
mustn
mustn't
must've
need
needn
needn't
should
shouldn
shouldn't
should've
ought
oughtn
oughtn't
dare
daren
daren't
m
re
t
s
ve
ll
d
a
an
the
and
but
if
or
because
as
until
while
of
at
by
for
with
about
against
between
into
through
during
before
after
above
below
to
from
up
down
in
out
on
off
over
under
again
further
then
once
all
any
both
each
few
more
most
other
some
such
no
nor
not
only
own
same
[...]

PR #3747, opened 2026-08-13
Add stopwords.words("english-extended")

alvations:
Closes the recurring "word X is missing from the English stop word list" reports without changing the corpus file.

Refs #3047, #2588, #1800, #3073, #3358, and supersedes the testing approach in #3434.

## Problem

Five issues report the same class of defect — a form is present for one member of a paradigm but missing for another:

| Issue | Report | Status today |
|---|---|---|
| **#3073** | `she's` present, **`he's` missing** — "quite a fundamental omission" | since fixed; the clearest illustration |
| **#3358** | `i'm` missing | since fixed |
| **#3047** | `i'm/i've/i'll/i'd`, `cannot`, `could've`, `ought`, `dare` … missing | pronoun forms fixed; **`cannot`, `could've`, `ought`, `mayn't`, `daren't`, `oughtn't` still missing** |
| **#2588** | general additions proposal | open |
| **#1800** | `wouldn` looks like a typo for `wouldn't` | open (both are present; the bare forms are tokenizer artifacts) |

Fixing this by editing `stopwords/english` in `nltk_data` needs a data release and silently changes results for every existing user. This PR instead adds an **opt-in list defined as data in code**: `stopwords.words("english")` is byte-for-byte unchanged, and `stopwords.words("english-extended")` returns it plus the additions — no data file, no re-download, works on existing installs.

## What counts as a stop word

I took the rule from the Snowball English list, which states it explicitly: **a closed-class function word is a safe stop word unless it has a frequent content-word homonym.** Snowball omits `will` ("he made a WILL"), `can` ("old tin CAN"), `may` ("merry month of MAY"), `must` ("a smell of MUST"), `might` ("with all thy MIGHT") on those grounds, while noting *"would, could, should, ought might however be included."*

Rarity is deliberately **not** grounds for exclusion — Snowball keeps rare forms "for completeness" (its example is `yourselves`), and NLTK already carries `shan't` and `mightn't`.

## Survey of eight reference lists

| List | Size | ∩ NLTK | only-list |
|---|---|---|---|
| stopwords-iso | 1298 | 192 | 1106 |
| Terrier | 733 | 120 | 613 |
| MALLET | 523 | 131 | 392 |
| scikit-learn | 318 | 119 | 199 |
| spaCy | 305 | 123 | 182 |
| **Snowball** | **174** | **159** | **15** |
| PostgreSQL | 127 | 127 | 0 |
| Lucene | 33 | 33 | 0 |
| *NLTK* | *198* | — | — |

They form two families: a **conservative function-word family** (Snowball; PostgreSQL and Lucene are strict subsets of NLTK's list) and the **SMART/Fox-derived family** (scikit-learn, Terrier, MALLET, stopwords-iso), which adds content-bearing words (`system`, `cry`, `fill`, `inc`, `ltd`, `serious`, number words). Raw cross-list consensus over-weights the SMART family because its members are not independent, so **Snowball is used as the reference**: it is the closest relative, sharing 159 of its 174 entries with NLTK.

A finding worth noting: NLTK currently includes the homonym-risky `will` and `can`, but **omits** the unambiguous `could`, `would` and `ought` — the reverse of Snowball's own recommendation. It also has every negated modal (`don't`, `won't`, `shouldn't`, `couldn't`, …) **except `can't`**.

## Accepted — 24 words + 63 U+2019 variants (198 → 285)

| Group | Words | Rationale |
|---|---|---|
| In Snowball, absent from NLTK | `cannot`, `can't`, `could`, `would`, `ought`, `let's`, `that's`, `there's`, `here's`, `who's`, `what's`, `where's`, `when's`, `why's`, `how's` | Snowball's "COMPOUND FORMS — pronoun + verb" and its unambiguous modals; `cannot`/`could`/`would` also appear in the SMART family (5/8 lists) |
| Paradigm completion | `could've`, `would've`, `might've`, `must've` · `daren't`, `oughtn't`, `mayn't` | NLTK already has `should've`; and `mightn't`/`mustn't`/`needn't`/`shan't`. Requested in #3047 |
| Closed-class, homonym-free | `never`, `without` | Preposition and negative adverb; NLTK already carries `against`/`between`/`through` and `no`/`nor`/`not`. In 5/8 lists |
| Typographic | U+2019 variants of every apostrophe form, derived at call time | Word processors, browsers and phone keyboards emit `don’t`, which never matches a list holding only `don't`. Same words, different codepoint — no new ambiguity. Not English-specific: Catalan has 14 apostrophe forms in its list |

## Rejected

| Rejected | Why |
|---|---|
| `may`, `must`, `might`, `shall` | Homonyms — the exact forms Snowball omits by name |
| `gonna`, `wanna`, `gotta`, `lemme`, `gimme`, `coulda`, `shoulda`, `woulda` | Register-specific; in **none** of the eight lists; and NLTK's own tokenizer splits them — `word_tokenize("gonna") == ["gon", "na"]` — so an entry could never match. (#3047's author likewise excluded forms that are "dialectal or too informal") |
| `system`, `cry`, `fill`, `inc`, `ltd`, `etc`, `eg`, `five`/`eight`/`nine`, `serious`, `name`, `get`, `go`, `us`, `one` | Content-bearing SMART/Fox-list entries; these are the long-criticised scikit-learn list's known weak spots |

## Design

**No data file is added and no corpus file is touched.** The word groups live in the new `nltk/corpus/reader/stopwords_extended.py`, one `frozenset` per reason they were added, with the citation beside each group.

Each list is an explicit `StopwordsExtension` that records its own virtual fileid and the corpus fileid it builds on:

```python
#: ``stopwords.words("english-extended")``.
StopwordsEnglishExtended = StopwordsExtension(
    fileid="english-extended",
    base="english",
    additions=_SNOWBALL_ADDITIONS | _PARADIGM_ADDITIONS | _CLOSED_CLASS_ADDITIONS,
)
```

`StopwordsCorpusReader` (in `wordlist.py`, beside the other `WordListCorpusReader` subclasses) lists what it serves:

```python
EXTENSIONS = (StopwordsEnglishExtended,)
```

Adding `<lang>-extended` later is a new `StopwordsExtension` plus one entry in that tuple — no reader logic, no corpus change, no test-harness change.

```python
>>> from nltk.corpus import stopwords
>>> len(stopwords.words("english")), len(stopwords.words("english-extended"))
(198, 285)
>>> set(stopwords.words("english")).issubset(stopwords.words("english-extended"))
True
```

Because `fileids()` advertises the virtual name, `raw()` and `open()` serve it too (generated from the word list) so `for f in fileids(): reader.raw(f)` keeps working; `abspath()` raises a clear error explaining that a code-defined list has no path, rather than an `OSError` about a missing corpus file.

## Tests — `nltk/test/unit/test_stopwords_extended.py` (61 tests)

Parametrised over `StopwordsCorpusReader.EXTENSIONS`, so a future `<lang>-extended` is covered automatically. Asserts:

- **additive invariants** — superset of the base, base order preserved, no duplicates, exposed via `fileids()`, every declared addition present;
- **every advertised fileid is readable** — `raw()` round-trips to `words()`, `open()` matches `raw()`, `abspath()` reports there is no file;
- **non-virtual fileids are served unchanged** — identical to what the plain `WordListCorpusReader` returns. (This can't be checked by asserting the additions are absent elsewhere: `hinglish` legitimately lists English words in its own corpus file.)
- **every accepted word**, the U+2019 completeness property, **and every rejected word stays out** — so a future edit can't quietly pull in the SMART list or the slang.

Skips cleanly when the corpus isn't installed. The wordlist section of `corpus.doctest` documents the new fileid.

Verified: 61 new tests pass; corpus/data suites pass; doctests on touched modules pass; isort/black/ruff and pre-commit clean.

### Note on #3434
PR #3434 targets the same issue with a test file. It can't run in CI (it skips unless a local `english_modified` file exists) and it rewrites the installed `stopwords/english` corpus in place during the run, restoring it only in a `finally`. The word list it asserts is largely the slang rejected above. This PR addresses the underlying request instead; #3434 can be closed.

### DIFF
--- a/nltk/corpus/__init__.py
+++ b/nltk/corpus/__init__.py
@@ -312,8 +312,8 @@ sinica_treebank: SinicaTreebankCorpusReader = LazyCorpusLoader(
 state_union: PlaintextCorpusReader = LazyCorpusLoader(
     "state_union", PlaintextCorpusReader, r"(?!\.).*\.txt", encoding="ISO-8859-2"
 )
-stopwords: WordListCorpusReader = LazyCorpusLoader(
-    "stopwords", WordListCorpusReader, r"(?!README|\.).*", encoding="utf8"
+stopwords: StopwordsCorpusReader = LazyCorpusLoader(
+    "stopwords", StopwordsCorpusReader, r"(?!README|\.).*", encoding="utf8"
 )
 subjectivity: CategorizedSentencesCorpusReader = LazyCorpusLoader(
     "subjectivity",
--- a/nltk/corpus/reader/__init__.py
+++ b/nltk/corpus/reader/__init__.py
@@ -61,6 +61,10 @@ from nltk.corpus.reader.tagged import *
 from nltk.corpus.reader.cmudict import *
 from nltk.corpus.reader.conll import *
 from nltk.corpus.reader.chunked import *
+from nltk.corpus.reader.stopwords_extended import (
+    StopwordsEnglishExtended,
+    StopwordsExtension,
+)
 from nltk.corpus.reader.wordlist import *
 from nltk.corpus.reader.xmldocs import *
 from nltk.corpus.reader.ppattach import *
@@ -120,6 +124,9 @@ __all__ = [
     "CMUDictCorpusReader",
     "ConllChunkCorpusReader",
     "WordListCorpusReader",
+    "StopwordsCorpusReader",
+    "StopwordsExtension",
+    "StopwordsEnglishExtended",
     "PPAttachmentCorpusReader",
     "SensevalCorpusReader",
     "IEERCorpusReader",
--- a/nltk/corpus/reader/stopwords_extended.py
+++ b/nltk/corpus/reader/stopwords_extended.py
@@ -1,0 +1,127 @@
+# Natural Language Toolkit: Code-defined additions to word list corpora
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+"""
+Stop word lists defined in code rather than read from ``nltk_data``.
+
+:class:`~nltk.corpus.reader.wordlist.StopwordsCorpusReader` serves each
+:class:`StopwordsExtension` listed in its ``EXTENSIONS`` as a virtual fileid
+(``english-extended``), built from the corpus file plus the words below. No
+corpus file is added or changed, so an extension works on existing installs with
+no data download.
+
+Additions are limited to closed-class function words that have no frequent
+content-word homonym -- the rule the Snowball English stop list states for
+omitting ``will``, ``can``, ``may``, ``must`` and ``might``. See nltk/nltk#3747
+for the full survey and the accept/reject rationale.
+
+To add ``<lang>-extended``, define the word groups here, name the extension, and
+list it in ``StopwordsCorpusReader.EXTENSIONS``::
+
+    StopwordsLangExtended = StopwordsExtension(
+        fileid="<lang>-extended", base="<lang>", additions=_GROUP_ONE | _GROUP_TWO
+    )
+
+The invariant tests in ``nltk/test/unit/test_stopwords_extended.py`` are
+parametrised over ``EXTENSIONS``, so a new entry is covered automatically.
+"""
+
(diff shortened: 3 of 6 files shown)


stevenbird:
Thank you @alvations!

Issue #2969, opened 2022-03-27
Testing offline


mcepl:
(I am sorry for bothering you on the issue tracker with something which probably isn't a bug, but I haven't any other functional means of communication with the project)

While packaging NLTK for openSUSE, I would like to make tests running. The problem is that our build system (as build systems for all distributions) are isolated from Internet, so I need to make running test suite possible without touching the network. So, I have downloaded all ntlk_data, set NTLK_DATA variable accordingly. Unfortunately, the result is not good:
```
[   78s] + cd /home/abuild/rpmbuild/BUILD
[   78s] + cd nltk-3.7
[   78s] ++ readlink -f ./ntlk_data/
[   78s] + export NLTK_DATA=/home/abuild/rpmbuild/BUILD/nltk-3.7/ntlk_data
[   78s] + NLTK_DATA=/home/abuild/rpmbuild/BUILD/nltk-3.7/ntlk_data
[   78s] ++ '[' -f _current_flavor ']'
[   78s] ++ cat _current_flavor
[   78s] + last_flavor=python38
[   78s] + '[' -z python38 ']'
[   78s] + '[' python38 '!=' python39 ']'
[   78s] + '[' -d build ']'
[   78s] + mv build _build.python38
[   78s] + '[' -d _build.python39 ']'
[   78s] + mv _build.python39 build
[   78s] + echo python39
[   78s] + python_flavor=python39
[   78s] + PYTHONPATH=/home/abuild/rpmbuild/BUILDROOT/python-nltk-3.7-0.x86_64/usr/lib/python3.9/site-packages
[   78s] + PYTHONDONTWRITEBYTECODE=1
[   78s] + pytest-3.9 --ignore=_build.python39 --ignore=_build.python310 --ignore=_build.python38 -v
[   79s] ============================= test session starts ==============================
[   79s] platform linux -- Python 3.9.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /usr/bin/python3.9
[   79s] cachedir: .pytest_cache
[   79s] rootdir: /home/abuild/rpmbuild/BUILD/nltk-3.7
[   79s] plugins: cov-3.0.0, mock-3.6.1
[   95s] collecting ... collected 424 items / 3 errors / 421 selected
[   95s]
[   95s] ==================================== ERRORS ====================================
[   95s] _______________ ERROR collecting nltk/test/unit/test_corpora.py ________________
[   95s] nltk/corpus/util.py:84: in __load
[...]

PR #3751, opened 2026-08-15
Improve offline testing: skip missing-corpus tests, document NLTK_DATA (#2969)

alvations:
Addresses #2969 (running the test suite offline, e.g. distro packaging).

**Bug fix.** `test_corpora.py` used `@pytest.mark.skipif(not ptb.fileids(), ...)`. The `skipif` condition is evaluated at collection time, and `ptb.fileids()` **raises** `LookupError` when the Penn Treebank (non-free, not in `nltk.download('all')`) is absent — so an offline build errors at collection and cannot run *any* test in the file, which is exactly the failure in #2969. Guard it so a missing corpus skips rather than aborts.

**Docs.** Add a short note to CONTRIBUTING.md#tests on running offline: pre-download `nltk_data`, set `NLTK_DATA`, deselect the tests needing non-free corpora.

Verified: with the corpus absent the module now collects and the PTB tests skip; with it present the file is unchanged (12 passed, 8 skipped).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -226,6 +226,18 @@ python -m doctest nltk/test/tokenize.doctest
 These are faster than running the full test suite and useful for quick
 iteration during development.
 
+The suite loads corpora from `nltk_data`. To run it without network access
+(e.g. when packaging for a distribution), download the data once and point
+`NLTK_DATA` at it:
+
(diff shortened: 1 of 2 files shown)


alvations:
This is not useful to other users perhaps. Closing this for now.

Issue #2732, opened 2021-06-16
Question - Agreement Krippendorff's alpha handling missing values


gbmarc1:
Documentation does not specify how to handle missing values:

https://github.com/nltk/nltk/blob/e4444c9b15762e6d86f5f6c4f5faadb87632c72a/nltk/metrics/agreement.py#L35-L44

In theory, Krippendorff's alpha CAN handle missing data (coders that do not assign an annotation to an item).

However, the documentation states:

```
Note that the data list needs to contain the same number of triples for each
individual coder, containing category values for the same set of items.
```

So, if we omit a triplet for missing data. Will the metrics be computed correctly? Do we have to put a place holder in the `labels` field of the triplet like np.nan or None?  

Something like?

`("coder1", "item1",  None)`

I am particularly interested in the multi-label case with the MASI distance and with labels defined in a `frozenset(['l1','l2'])`.

tomaarsen:
It seems that if `None` is used, it is considered as a value, and not ignored like one might expect. Please have a look at #2865 for some more discussion on the topic. That said, I'm unsure whether omitting these triples results in the correct computation - I'm not very familiar with the agreement module, nor the corresponding line of research.

PR #3752, opened 2026-08-15
Krippendorff's alpha: treat placeholder labels as missing data (#2865, #2732)

### LINKED ISSUE #2865: Incorrect krippendorffs alpha result with missing value or missing data
```python
from nltk import agreement
rater1 = [1,1,2]
rater2 = [1,1,None]
rater3 = [None,1,2]

taskdata=[[0,str(i),str(rater1[i])] for i in range(0,len(rater1))]+[[1,str(i),str(rater2[i])] for i in range(0,len(rater2))]+[[2,str(i),str(rater3[i])] for i in range(0,len(rater3))]
print(taskdata) # (annotator_id, sample_id, label_id)
ratingtask = agreement.AnnotationTask(data=taskdata)

print("alpha " +str(ratingtask.alpha())) # krippendorffs alpha
```

```python
alpha 0.33333333333333337
```

I am not sure if this result is correct. In my example, I assume I have 3 examples needed to annotate. Except the missing values (not every annotator annotate each example), all the annotators provided the same labels. I thought the alpha should be 1. However, I got 0.33.

Does anyone understand is 0.33 correct or 1.0 is correct for krippendorffs alpha?

Thank you!

alvations:
Fixes #2865 and answers #2732.

## Problem

`AnnotationTask` counts a placeholder label (e.g. `None`) as a real category. So a dataset where every *present* rating agrees but some ratings are missing scores `alpha = 0.33` instead of `1.0` (#2865):

```python
rater1, rater2, rater3 = [1,1,2], [1,1,None], [None,1,2]
# ... AnnotationTask(...).alpha()  -> 0.333
```

## What's actually going on

`alpha()` **already** implements Krippendorff's missing-data handling — it drops items rated by fewer than two coders — so if the missing triples are simply **omitted**, the same data gives `1.0`. The only gap was that a `None` placeholder was being counted as a category.

## Fix

An opt-in `missing_values` argument: a collection of label values that mean *"not annotated"* (e.g. `[None]` or `["", None]`). Matching triples are dropped on load, so the placeholder is never a category.

```python
AnnotationTask(data, missing_values=[None]).alpha()   # -> 1.0
```

- **Backward compatible**: default drops nothing, so existing results are unchanged.
- Meaningful only for `alpha()` — `kappa`/`pi`/`S`/`weighted_kappa` assume complete data (documented). Also fills the module's long-standing `TODO: Describe handling of ... missing data`.

## No new attack surface

Deliberately checked, since agreement labels are caller-supplied: the change is pure set-membership + skip — no `eval`/`exec`, no regex, no file/network I/O — and it can only *drop* data, never amplify work, so it adds no DoS vector. An unhashable placeholder raises `TypeError` at construction (caught immediately, not exploitable).

## Tests

`TestMissingValues` covers the reported `None` case, the string `"None"` a caller gets from stringifying labels, the **MASI multi-label** case from #2732, backward compatibility on complete data, all-missing (clean `ValueError`), and the unhashable-placeholder guard. Full `test_disagreement.py` 14 passed; `metrics.doctest` passes.

### DIFF
--- a/nltk/metrics/agreement.py
+++ b/nltk/metrics/agreement.py
@@ -94,7 +94,7 @@ class AnnotationTask:
     is the MASI metric, which requires Python sets.
     """
 
-    def __init__(self, data=None, distance=binary_distance):
+    def __init__(self, data=None, distance=binary_distance, missing_values=None):
         """Initialize an annotation task.
 
         The data argument can be None (to create an empty annotation task) or a sequence of 3-tuples,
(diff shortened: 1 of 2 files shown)


stevenbird:
Thanks @alvations

PR #3753, opened 2026-08-15
[Do not merge yet] Fix GHSA-8mgp-746c-j5xp: URL-scheme bypass of pathsec.validate_path

alvations:
## GHSA-8mgp-746c-j5xp — URL-scheme bypass of `validate_path`

`nltk.pathsec.validate_path()` is the sandbox chokepoint the whole library
funnels file access through. It authorized **any** string containing `://` with
an `http`/`https`/`ftp` scheme:

```python
if "://" in raw:
    parsed = urlparse(raw)
    if parsed.scheme in ("http", "https", "ftp"):
        return          # <-- "return" means AUTHORIZED
```

To the kernel `http://../../etc/passwd` is not a URL — it is the directory
`http:`, an empty component, then `..`. So a scheme prefix authorized a
traversal out of every allowed root, defeating every downstream path check.

```python
>>> pathsec.validate_path("http://../../../../etc/passwd")   # returned: authorized
```

### Fix

- **Reject URL-shaped strings** in `validate_path` with an anchored,
  case-insensitive, whitespace-tolerant regex. A real scheme is a *prefix*,
  never a substring — the substring test is what created the bug. No in-tree
  caller passes an HTTP URL to `validate_path` (network I/O goes through
  `pathsec.urlopen`), so outright rejection is behaviour-safe. Delegating to
  `validate_network_url()` was tried and still failed open (host `..` resolves
  to `[]` and passes), so the rejection lives at this layer.
- **`file:` URLs** are unquoted; the ambiguous `?`/`#`/`;` forms — where urllib
  opens `Request.selector` but `urlparse().path` drops them, validating a
  different file than gets opened — are refused.
- **`AveragedPerceptron.save`/`load`** are routed through `pathsec.open`; they
  used bare `open()` and so opened caller paths directly, outside the sandbox.

### Tests

`nltk/test/unit/test_pathsec.py` — `TestUrlSchemePathBypass` (incl. the one that
matters most: `validate_path("http://../../../../etc/passwd")` must raise — if it
ever passes silently, GHSA-8mgp is back), ambiguous-`file:` rejection, a
no-false-positive check that in-root paths still validate, and
`TestModelArtifactSaveContainment` driving `save`/`load` against a
genuinely-outside `$HOME` target (a temp dir is not outside — the private system
temp is an allowed root on macOS). 101 pathsec tests pass; tagger / classify /
serialization regression suites green.

A companion **living-harness probe** (`ghsa_8mgp_746c_j5xp.py`) that runs this
attack on every commit lands with the security-probe harness (#3749).

---

## Pickle deserialization consolidation (CWE-502)

The GHSA-8mgp surface is broader than path handling: unpickling runs arbitrary
code, so any `pickle.load` on attacker-influenceable bytes is RCE. This branch
consolidates **all** pickle handling in the tree behind a single audited
unpickler in `nltk.picklesec`. (Split out, with the same `nltk/picklesec.py`,
as PR #3823.)

**One audited unpickler, no caller-tunable surface.** `AllowlistUnpickler`
reconstructs only an explicit `(module, name)` allowlist, bounded by hard guards
that cannot be opted out of: dotted names refused (attribute-traversal RCE,
GHSA-4489), dunder names refused, denied modules refused even under a broad
`allowed_modules` (os / subprocess / builtins / importlib / marshal / operator /
ctypes / socket, and the numpy·scipy·sklearn·pandas file-I/O subtrees), `builtins`
denied wholesale except a tiny audited `_SAFE_DENIED_GLOBALS` (int/float/… and
the inert `object()` sentinel), and EXT1/EXT2/EXT4 extension-registry opcodes
refused up front (they resolve via `copyreg` **without** calling `find_class`).

- The per-caller unpickler subclasses (`_TblModelUnpickler`,
  `_TransitionParserModelUnpickler`) and the `unpickler_cls=` injection hook are
  **removed** — the base does all the work. A caller can no longer substitute a
  weaker unpickler.
- The numpy object-dtype `scalar` nested-unpickle sink is wrapped in the base
  (`_GUARDED_GLOBALS`) for every caller, refusing an object-bearing dtype before
  numpy's own unrestricted unpickler runs.
- **Caller allowlists are validated up front**, in the unpickler's constructor:
  `allowed_globals` / `allowed_modules` are caller-controlled, so a bad entry (a
  denied global/module, a dotted/dunder name, a scientific-stack I/O sink) is
  refused before a byte is read, not only when a hostile pickle names it. A
  caller cannot configure a loader that reaches `os.system` / `builtins.eval`.
- The base unpickler config is **pinned**: the unpickler classes no longer
  forward `**kwargs` to `pickle.Unpickler`, so a caller cannot hand it
  `fix_imports` / `encoding` / `errors` / `buffers` — the last "configure the
  base unpickler" surface once `unpickler_cls` was gone.

**All pickle routes through picklesec.** Every remaining bare
`pickle.dump`/`dumps` (chartparser_app ×3, wordnet_app, transitionparser,
sentiment_analyzer, tbl.demo) now uses `picklesec.pickle_dump` / `pickle_dumps`.
`tools/check_all_pickle_through_picklesec.py` (+ pre-commit hook) fails on any
bare `pickle.load/loads/dump/dumps/Unpickler/Pickler` under `nltk/`, tracking
import aliases and from-imports so none can bypass it.

### Tests (real loader, not mocks)

`test_picklesec_allowlist_consolidation.py`: `unpickler_cls` gone; no
`**kwargs` / fix-imports remap / EXT / PERSID substitution or `find_class`
bypass; `allowed_globals` cannot widen past the hard guards (a
construction-time refusal matrix over 16 dangerous globals + 18 dangerous
modules); `builtins.object` proven inert; the numpy scalar sink refused by the
base; the dump helpers round-trip; and the CI guard passes with teeth. Every
real allowlist (tbl, punkt, transition parser, chart parser) still constructs;
297 nltk modules import; real tbl and transition-parser models still load.


ekaf:
CI fails because the write is blocked, which should be the success criterion for this PR.

alvations:
Almost complete so doing some housekeeping here. (Will update this comment once the last bits of corenlp related hardening gets merged up)

Moving forward, 

- `nltk/test/unit/security_probes/ghsa_*.py` should still exist on the `develop` branch so that we never compromise on these as new PRs get merged. 
- thorough CVE/CWE checking against MITRE and NIST and CVE.org would be kept on the new `develop-security` branch because it will contain a substantial CVE xml description file and a `SECURITY_LEDGER.md` that will help future agents to explicitly scan through those as we work to red/blueteam the NLTK codebase if more of these "emergent" security capabilities of LLMs start to massively send us more GHSAs.

ekaf:
GHSA-8mgp-746c-j5xp is narrowly about the following components, so we could close #3789 by releasing a new version as soon as these are fixed:

> Affected component: TransitionParser.train, TransitionParser.parse, AveragedPerceptron.save, AveragedPerceptron.load, PerceptronTagger.save_to_json, save_maxent_params

PR #3761, opened 2026-08-17
Contain CorpusReader / agreement / chat80 file I/O in the pathsec sandbox

alvations:
Split out of #3753 so the corpus-reader / metrics / semantics file-I/O
containment can be reviewed on its own. Depends only on `nltk.pathsec`, already
on `develop`.

### `nltk/corpus/reader/api.py` (CWE-59)
Corpus-view methods (`words` / `sents` / `parsed_sents` / ...) open the pointers
returned by `abspaths()` directly, bypassing `open()`. `abspath()` / `abspaths()`
now resolve every corpus-relative fileid through a shared `_guard_fileid`, which
runs `validate_path` so an intermediate-directory symlink inside a trusted corpus
root that escapes it is refused — the same containment `open()` already had. The
guard lives in the base class, so every reader (MTE / XML / PanLex / Lin / ...)
inherits it. `validate_path` is imported at module top (pathsec is pure stdlib,
so there is no import cycle).

### `nltk/metrics/agreement.py`
The `__main__` `-f` caller-file read goes through `pathsec.open`.

### `nltk/sem/chat80.py`
`cities2table` (sqlite3), `val_dump` / `val_load` (shelve) validate the
caller-supplied db path before touching the filesystem, and `label_indivs`
writes `chat_pnames.cfg` through `pathsec.open`.

### Tests
- `test_corpus_reader_pathsec.py` — base-class root/symlink containment across
  readers, plus a direct check that `self._root.join` alone does not contain the
  escape while `abspath` (which adds `validate_path`) refuses it.
- `test_chat80_pathsec.py` — each chat80 sink (cities2table / val_dump /
  val_load / label_indivs) refuses an outside path and leaves nothing behind.
- `test_agreement_pathsec.py` — the agreement `__main__` `-f` read refuses an
  outside path, with a negative control.

### DIFF
--- a/nltk/corpus/reader/api.py
+++ b/nltk/corpus/reader/api.py
@@ -18,6 +18,7 @@ from itertools import chain
 from nltk import pathsec
 from nltk.corpus.reader.util import *
 from nltk.data import FileSystemPathPointer, PathPointer, ZipFilePathPointer
+from nltk.pathsec import validate_path
 
 
 class CorpusReader:
@@ -81,7 +82,7 @@ class CorpusReader:
         # The ``str`` test also covers ``FileSystemPathPointer`` (a ``str``
         # subclass), so a pointer root is validated too.
         if pathsec.ENFORCE and isinstance(root, str):
-            pathsec.validate_path(root, context="CorpusReader.__init__")
+            validate_path(root, context="CorpusReader.__init__")
 
         # Convert the root to a path pointer, if necessary.
         if isinstance(root, str) and not isinstance(root, PathPointer):
@@ -171,6 +172,21 @@ class CorpusReader:
         """
         return self._fileids
 
+    def _guard_fileid(self, fileid):
+        """Resolve a corpus-relative fileid to a path pointer with the scoped-root
+        guard, so view methods get the same containment as :meth:`open`.
+
+        View methods (``words``/``sents``/``parsed_sents``/...) open the pointers
+        returned by :meth:`abspaths` directly, bypassing :meth:`open`.
+        ``self._root.join`` already rejects a ``..`` traversal; the scoped
+        ``validate_path`` additionally resolves symlinks so an intermediate
+        directory symlink inside the corpus root that escapes it is refused too
+        (CWE-22/CWE-59).
+        """
+        path = self._root.join(fileid)
+        validate_path(path, context="CorpusReader", required_root=self._root)
+        return path
+
     def abspath(self, fileid):
         """
         Return the absolute path for the given file.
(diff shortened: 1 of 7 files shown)


alvations:
This PR follows the established patterns in https://github.com/nltk/nltk/pull/3755 to create the temp stagging dir when writing files out from nltk functions. Now we should have corpus loading from the CorpusReader passing through the `validate_path` first. Merging this in.

vizovitin:
@alvations , I think this breaks uv-based llama-index installs which bundled nltk data cache. E.g. I have the following dependencies installed (`uv tree`):

```
ingestion v1.0.0
├── llama-index v0.14.23
│   ├── llama-index-core v0.14.23 (*)
│   ├── llama-index-embeddings-openai v0.6.0 (*)
│   ├── llama-index-llms-openai v0.7.10 (*)
│   └── nltk v3.10.3 (*)
```

And then the following sample command incorrectly trips the check:
```
$ UV_CACHE_DIR="/var/jenkins/caches/uv" NLTK_DATA=.venv/lib/python3.10/site-packages/llama_index/core/_static/nltk_cache uv run python -c "from nltk.corpus import stopwords; print(stopwords.words('english')[:5])"
  File "<string>", line 1, in <module>
  File ".venv/lib/python3.10/site-packages/nltk/corpus/reader/wordlist.py", line 37, in words
    for line in line_tokenize(self.raw(fileids))
  File ".venv/lib/python3.10/site-packages/nltk/corpus/reader/api.py", line 231, in raw
    with self.open(f) as fp:
  File ".venv/lib/python3.10/site-packages/nltk/corpus/reader/api.py", line 257, in open
    return path.open(encoding=encoding)
  File ".venv/lib/python3.10/site-packages/nltk/data.py", line 444, in open
    stream = _secure_open(self._path, "rb")
  File ".venv/lib/python3.10/site-packages/nltk/pathsec.py", line 796, in open
    return _hardened_open(raw_path, mode, context, required_root, **kwargs)
  File ".venv/lib/python3.10/site-packages/nltk/pathsec.py", line 721, in _hardened_open
    raise PermissionError(
PermissionError: Security Violation [pathsec.open]: refusing multiply-linked file '.venv/lib/python3.10/site-packages/llama_index/core/_static/nltk_cache/corpora/stopwords/english' (st_nlink=2); a hardlink can point at an outside-root inode (CWE-59)
```

The hardlinks in question:
```
$ find /var/jenkins/caches/uv/ /var/jenkins/workspace/ -samefile .venv/lib/python3.10/site-packages/llama_index/core/_static/nltk_cache/corpora/stopwords/english | sed -e "s|`pwd`/||"
/var/jenkins/caches/uv/archive-v0/gEzrFFg7t674bB463HSj0/llama_index/core/_static/nltk_cache/corpora/stopwords/english
.venv/lib/python3.10/site-packages/llama_index/core/_static/nltk_cache/corpora/stopwords/english
```

PR #3762, opened 2026-08-17
Add a shared pathsec sandbox conftest fixture

alvations:
The `nltk.pathsec` security test suites each repeat the same fixture setup:
enforce pathsec against a single throwaway data root, then aim a patched file
sink either inside it (must succeed) or at a path outside it (must be refused).
This centralizes that setup so the repetition can be removed across the suite.

### `nltk/test/unit/conftest.py` (new)
pytest discovers `conftest.py` automatically, so any unit test can request these
by name with no import:
- **`sandbox`** — yields a fresh dir OUTSIDE every allowed root (writes/reads
  there must be refused).
- **`restricted_sandbox`** — yields the one allowed data root (a write inside it
  succeeds).
- **`enforce_off`** — forces `pathsec.ENFORCE` off for functional read/write
  tests that operate on a temp-dir target (isolates them from ambient/leaked
  ENFORCE state; on Linux a temp dir lives under world-writable `/tmp`, which is
  not an allowed root).

The outside dir is created under `$HOME`, never a temp dir, because a private
per-user system temp dir is itself an allowed root on macOS.

### Adoption
`test_pathsec_sweep_tokenize.py` drops its local `sandbox` copy and uses the
shared one here. The other open pathsec PRs adopt the same fixtures in their own
branches so each becomes DRY as it lands.

### DIFF
--- a/nltk/test/unit/conftest.py
+++ b/nltk/test/unit/conftest.py
@@ -1,0 +1,97 @@
+# Natural Language Toolkit: shared pytest fixtures for the pathsec sandbox tests
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""Shared fixtures for the ``nltk.pathsec`` sandbox test suites.
+
+Many security tests need the same setup: enforce pathsec against a single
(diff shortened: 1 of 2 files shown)


alvations:
Merging this in first so that the rest of the hardening tests can reuse the same sandbox setup.

PR #3763, opened 2026-08-17
Write parse demo/profile output to a private staging dir, not /tmp or the CWD

alvations:
Split out of #3753 — the two `nltk.parse` demo/profile sinks that wrote to a
hardcoded path. Depends only on `nltk.data.make_staging_dir`, already on
`develop`.

### Problem
- `featurechart.run_profile()` wrote its profile stats to a hardcoded
  `/tmp/profile.out` — a guessable path another local user could pre-create or
  symlink (CWE-377/378).
- `dependencygraph`'s demo saved `tree.png` into the current working directory.

### Fix
Both now write into a fresh private (mode 0700), unpredictably-named staging
directory created under an allowed NLTK data root via
`nltk.data.make_staging_dir` — the same helper the maxent / named-entity / weka
save code uses, which lands inside the pathsec sandbox on every platform
(including Linux, where the shared `/tmp` is deliberately not a data root). The
dependency-graph demo also prints where it saved the image.

No tests: these are `__main__`/dev-only demo and profiling helpers.

### DIFF
--- a/nltk/parse/dependencygraph.py
+++ b/nltk/parse/dependencygraph.py
@@ -20,6 +20,7 @@ from collections import defaultdict
 from itertools import chain
 from pprint import pformat
 
+from nltk.data import make_staging_dir
 from nltk.internals import find_binary
 from nltk.pathsec import open as _secure_open
 from nltk.tree import Tree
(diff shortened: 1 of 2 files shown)


alvations:
This PR follows the established patterns in #3755 to create the temp stagging dir when writing files out from nltk functions.

PR #3764, opened 2026-08-17
Add pathsec regression tests for sem.util / xmlsec / dataset loaders

alvations:
Split out of #3753 — **test-only**. The file-I/O containment for these sinks is
already merged on `develop`; this adds the behavioural attack tests that pin it,
so a future refactor cannot silently drop the guard. They use the shared
`nltk/test/unit/conftest.py` fixtures (also already on develop).

### Tests
- **`test_pathsec_sweep_sem_util.py`** — `nltk.sem.util.read_sents` refuses a read
  outside the data sandbox while still reading a legitimate in-root file.
- **`test_pathsec_sweep_infra.py`** — `nltk.xmlsec.parse` routes a caller-named
  path through pathsec validation (refuses out-of-sandbox).
- **`test_pathsec_sweep_dataset_loading.py`** — a behavioural matrix over NLTK's
  dataset/model loaders (`nltk.corpus.reader`), each driven against a real file
  outside a single throwaway sandbox root and required to refuse it.

No production changes. Verified green against `develop`'s current code.

### DIFF
--- a/nltk/test/unit/test_pathsec_sweep_dataset_loading.py
+++ b/nltk/test/unit/test_pathsec_sweep_dataset_loading.py
@@ -1,0 +1,320 @@
+"""
+Path-traversal / arbitrary-file-read sweep over NLTK's dataset & model loaders.
+
+This is a *behavioural* attack matrix (not a source grep): with
+``pathsec.ENFORCE = True`` and ``nltk.data.path`` pinned to a single throwaway
+sandbox root, every load path below is driven against a genuine file that lives
+*outside* that root and must REFUSE the read (or write) with a
+``PermissionError``/``ValueError`` raised by ``nltk.pathsec``.
+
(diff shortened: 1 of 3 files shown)


alvations:
Merging this in. This should validate some of the additional tests we've written for the pathsec and xmlsec and dataset loading.

ekaf:
Hi @alvations, https://github.com/nltk/nltk/security/code-scanning found that this PR introduced "Clear-text logging of sensitive information": 

> If sensitive data is written to a log entry it could be exposed to an attacker who gains access to the logs.
> 
> Potential attackers can obtain sensitive user data when the log output is displayed. Additionally that data may expose system information such as full path names, system information, and sometimes usernames and passwords.

alvations:
Hmmm that’ll be a bit hard to avoid since we want the user to know what or where it’s loaded. Let me see if we can find some solution around this

PR #3770, opened 2026-08-20
Split nltk_data pickle-asset regression into always-on and skippable tests

alvations:
Follow-up to #3757 (test-only, no behavior change).

The pickle-asset regression that landed in #3757 combined two concerns in one test — an always-runnable staged-asset load and a sweep over whatever `*.pickle` happen to be installed. This splits them so each is clear and correct:

- **`test_data_load_restricted_pickle_roundtrips_real_asset`** stages a genuine globals-free `*.pickle` in a temp data root and loads it through the real `nltk.data.load` -> `restricted_pickle_load` path. It never skips, so the restricted loader is exercised on every runner, including a bare one with no downloaded corpora.
- **`test_all_nltk_data_pickle_assets_load`** loads every `*.pickle` actually installed and `pytest.skip`s only when none are present, rather than relying on the staged asset to keep its resource set non-empty.

Verified: both pass; black / isort / ruff clean.

### DIFF
--- a/nltk/test/unit/test_pickle_allowlist_security.py
+++ b/nltk/test/unit/test_pickle_allowlist_security.py
@@ -1420,21 +1420,16 @@ def test_nltk_functions_through_hardened_pickle_loaders():
         pytest.skip("no nltk data packages installed to exercise the loaders")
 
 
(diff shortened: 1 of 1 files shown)


alvations:
This should further make #3757 testing more robust. Merging this in.

PR #3772, opened 2026-08-20
Harden internals.java() (CWE-88): options, classpath, env, discovery; route MaltParser through it

alvations:
Follow-up hardening of `internals.java()` (extends #3771), with `develop` merged in. An exhaustive adversarial audit of **every** channel into the JVM launcher — options, cmd, classpath, environment, binary/jar discovery, temp files, and both the blocking and non-blocking launch paths — plus routing the one wrapper that still bypassed `java()` (MaltParser) through it, so **every** NLTK JVM launch now shares a single hardened entry point.

Relates to CVE-2026-12841 / GHSA-m4rf (CWE-88). The separate str/bytes wrapper-decode fix is #3777 (merged).

## Injection vectors found & fixed

**Classpath sandbox (`_verify_jar_sandbox`)**
- **Empty classpath element = the JVM's CWD** (`a.jar::`, `:a.jar`, `a.jar::b.jar`, or `""` / `None` in a list): empty parts were dropped *before* verification and the raw string handed to `-cp`, silently adding the working directory to the classpath. Empty entries are refused and `-cp` is rebuilt from the verified entries only.
- **A list entry embedding `os.pathsep`** (`trusted.jar:/etc/passwd`) passed the per-entry realpath/root check as one path but re-split in the JVM into extra **unverified** elements — refused.
- **Non-string entry, NUL-byte entry, relative path, out-of-root, symlink-escape** — all refused with accurate per-case errors (the empty-vs-non-string messages are split; NUL is a clear `UntrustedJarError` instead of a bare `ValueError`).
- Trusted roots are `normcase`d to match the candidate entry (fixes a Windows false-reject).

**Option allowlist (`_validate_java_options`)**
- Prefix matching replaced with **anchored shapes**: sizing flags must be `number+optional-unit`, `-verbose` a known category, mode flags exact — so no arbitrary suffix can ride a safe prefix (`-Xmx512m@/file`, `-verbose:gc:file=…`, `-Xbatchevil`). Empirically inert on the JVM, refused for defense in depth.
- Shape guard rejects whitespace / control chars / shell metacharacters (incl. tab, CR, NBSP, DEL); `@argfile` rejected in any position; matching is case-insensitive. `config_java()` validates its global options too.

**Environment (`_java_child_env`)**
- Strips `JAVA_TOOL_OPTIONS`, `_JAVA_OPTIONS`, `JDK_JAVA_OPTIONS`, `IBM_JAVA_OPTIONS`, `OPENJ9_JAVA_OPTIONS` (arbitrary flags incl. `-XX:OnError` / `-javaagent`) and `CLASSPATH` (unverified code) from the child environment, case-insensitively; `PATH` and benign vars are preserved.

**Binary / jar discovery**
- `find_binary` refuses a CWD-relative match even when a **bare** `path_to_bin` is supplied — a plain `config_java(bin="java")` was hijackable by a planted `./java/java` (CWE-426 / CWE-427). A path with a directory component is still honored as the caller's explicit choice.
- `find_jar_iter`'s regex + searchpath branch had a mis-indented `yield` that returned **any** directory entry (subdirs / unrelated files), not the matching jar — fixed to yield only a matching file.
- Discovery composes with the sandbox: a jar returned from an attacker-poisoned env var (e.g. `STANFORD_POSTAGGER`) is still refused by the classpath sandbox when out of root.

**Temp files (CWE-377)** — audited every wrapper; all use a secure API (`NamedTemporaryFile` / `mkstemp` / `mkdtemp`), none a predictable-name one; a regression guard locks this.

## MaltParser now routes through `internals.java()`

MaltParser was the last wrapper that built and ran its own subprocess, hand-duplicating the protections — which is exactly why the env sanitisation had been missing on that path. It now calls `internals.java()`, inheriting the classpath sandbox, option allowlist and env-strip, so they cannot drift out of sync again. It passes MaltParser's own `-w <workingdir>` as a **program argument** instead of chdir-ing the JVM process — deliberately avoiding adding a `cwd` parameter to `java()`, which would reopen a CWD class-injection hole (verified live: with no explicit `-cp` the JVM runs `.class` files from the working directory). It also now honours `config_java(bin=…)` instead of hard-coding bare `java`.

## Tests

The harness gains a large attack matrix — 71 option / cmd / classpath / env attacks, each with a **load-bearing mutation** (neuter the guard → the exploit becomes reachable) — plus discovery-composition tests, non-blocking-launch guards, the MaltParser-routing tests, and the CWE-377 guard. **224 java-related tests pass** across the 16 java test files; the full attack matrix shows zero leaks.

**Verified live against a real OpenJDK 26** (not mocked, using the official Stanford tagger/parser, MaltParser 1.9.2 and Weka 3.8.6 distributions): StanfordPOSTagger, StanfordTokenizer, StanfordParser (constituency + dependency), MaltParser (train + parse) and WekaClassifier all produce correct output through the hardened path.

black 25.11.0 / isort / pyupgrade (`--py310-plus`) clean.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -239,7 +239,7 @@ def _verify_jar_sandbox(classpath_entries):
 
     for p in data_path:
         try:
-            trusted_roots.append(os.path.realpath(p))
+            trusted_roots.append(os.path.normcase(os.path.realpath(p)))
         except Exception:
             continue
 
@@ -250,7 +250,7 @@ def _verify_jar_sandbox(classpath_entries):
         repo_root = os.path.realpath(os.path.join(nltk_package_dir, ".."))
         git_marker = os.path.join(repo_root, ".git")
         if os.path.isdir(git_marker) or os.path.isfile(git_marker):
-            trusted_roots.append(repo_root)
+            trusted_roots.append(os.path.normcase(repo_root))
     except Exception:
         pass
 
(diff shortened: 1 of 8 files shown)


alvations:
Should have enough hardening now for corenlp and maltparser and weka interfaces for now. Merging this in.

Issue #2121, opened 2018-09-13
Dangerous default values in arguments


alvations:
There are quite some dangerous default dicts and lists in class/functions arguments. 

But in some cases it's unclear whether the purpose is really to keep the values in the list in subsequent calls. See https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python 

It'll be good to look through each of them and convert the to default to `None` instead of `{}` or `[]` but I'm not exactly sure which of these cases the items in list should be persistent and when it shouldn't.

A list of them are:

```
nltk/tgrep.py:757: [W0102(dangerous-default-value), _tgrep_exprs_action.top_level_pred] Dangerous default value macro_dict (builtins.dict) as argument
nltk/util.py:579: [W0102(dangerous-default-value), binary_search_file] Dangerous default value {} as argument
nltk/cluster/util.py:188: [W0102(dangerous-default-value), Dendrogram.__init__] Dangerous default value [] as argument
nltk/cluster/util.py:225: [W0102(dangerous-default-value), Dendrogram.show] Dangerous default value [] as argument
nltk/app/wordnet_app.py:565: [W0102(dangerous-default-value), _collect_all_synsets] Dangerous default value dict() (builtins.dict) as argument
nltk/app/wordnet_app.py:632: [W0102(dangerous-default-value), Reference.__init__] Dangerous default value dict() (builtins.dict) as argument
nltk/classify/weka.py:180: [W0102(dangerous-default-value), WekaClassifier.train] Dangerous default value [] as argument
nltk/chunk/util.py:521: [W0102(dangerous-default-value), ieerstr2tree] Dangerous default value [] as argument
nltk/chat/util.py:38: [W0102(dangerous-default-value), Chat.__init__] Dangerous default value {} as argument
nltk/sem/drt.py:850: [W0102(dangerous-default-value), resolve_anaphora] Dangerous default value [] as argument
nltk/sem/chat80.py:223: [W0102(dangerous-default-value), Concept.__init__] Dangerous default value [] as argument
nltk/sem/chat80.py:223: [W0102(dangerous-default-value), Concept.__init__] Dangerous default value [] as argument
nltk/sem/chat80.py:223: [W0102(dangerous-default-value), Concept.__init__] Dangerous default value set() (builtins.set) as argument
nltk/sem/chat80.py:334: [W0102(dangerous-default-value), clause2concepts] Dangerous default value [] as argument
[...]

PR #3773, opened 2026-08-20
Fix shared defaults in chat80 Concept

tandede:
## Summary

- replace the mutable default arguments in `Concept` with `None` sentinels
- allocate fresh default label, closure, and extension containers for each instance
- preserve caller-provided containers when arguments are supplied explicitly
- add a regression test covering state isolation across two default-constructed concepts
- add `tandede` to `AUTHORS.md` as required by the contribution guide

## Motivation

`Concept` currently uses a shared list for `altLabels`, a shared list for `closures`, and a shared set for `extension`. Mutating any of those defaults through one instance changes later instances. This addresses the three `Concept.__init__` cases identified in #2121 without changing explicit-container behavior.

## Validation

- `pytest -q nltk/test/unit/test_chat80.py nltk/test/chat80.doctest` (2 passed)
- `pytest -q nltk/test/unit` (1944 passed, 71 skipped, 9 xfailed)
- `pre-commit run --files AUTHORS.md nltk/sem/chat80.py nltk/test/unit/test_chat80.py` (all hooks passed)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -318,6 +318,7 @@
 - Volodymyr Matsko <https://github.com/Lemm1>
 - Nicola <https://github.com/trinik15>
 - medisean <https://github.com/medisean>
+- tandede <https://github.com/tandede>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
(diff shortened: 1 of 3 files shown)


stevenbird:
Good catch, thanks.

PR #3775, opened 2026-08-22
Reuse cached paths for WordNet hypernym statistics

### LINKED ISSUE #2270: Hypernyms related functions in Wordnets can be unified
These functions in wordnet's `Synset()` object could be unified:
 
 - hypernym_paths
 - min_depth
 - max_depth
 - root_hypernyms

The while loops / recursion achieve the same purpose of reaching the root hypernyms, while doing so, the min and max depth should be logged and so should the hypernym paths. 

We can derive the other three attributes given the hypernym_paths, e.g. 

```python
from nltk.corpus import wordnet as nltk_wn
ss = nltk_wn.synset('sweet.n.1')
hypernym_paths = ss.hypernym_paths()
assert ss.max_depth() == max(len(path) for path in hypernym_paths) - 1
assert ss.min_depth() == min(len(path) for path in hypernym_paths) - 1
assert list(set([path[0] for path in hypernym_paths])) == ss.root_hypernyms()
```

----

Something like this:

```python
    def init_hypernym_paths(self):
        """
        Get the path(s) from this synset to the root, where each path is a
        list of the synset nodes traversed on the way to the root.
        :return: A list of lists, where each list gives the node sequence
        connecting the initial ``Synset`` node and a root node.
        """
        self._hyperpaths = []
        hypernyms = self.hypernyms() + self.instance_hypernyms()
        if len(hypernyms) == 0:
            paths = [[self]]
        for hypernym in hypernyms:
            for ancestor_list in hypernym.hypernym_paths():
                ancestor_list.append(self)
                self._hyperpaths.append(ancestor_list)
        # Compute the path related statistics.
        self._min_depth = min(len(path) for path in self._hyperpaths)
        self._max_depth = max(len(path) for path in self._hyperpaths)
        # Compute the store the root hypernyms.
        self._root_hypernyms = list(set([path[0] for path in self._hyperpaths]))

    def hypernym_paths():
        return self._hyperpaths

    def min_depth():
        return self._min_depth

    def max_depth():
        return self._max_depth

    def root_hypernyms():
        return self._root_hypernyms
```

medisean:
## Summary

- compute WordNet hypernym paths once per synset and reuse them for depth and root queries
- keep `hypernym_paths()` results isolated from the internal cache
- add regression coverage for shared results and path-copy behavior

Fixes #2270

### DIFF
--- a/nltk/corpus/reader/wordnet.py
+++ b/nltk/corpus/reader/wordnet.py
@@ -522,22 +522,14 @@ class Synset(_WordNetObject):
 
     def root_hypernyms(self):
         """Get the topmost hypernyms of this synset in WordNet."""
-
-        result = []
+        roots = []
         seen = set()
-        todo = [self]
-        while todo:
-            next_synset = todo.pop()
-            if next_synset not in seen:
-                seen.add(next_synset)
-                next_hypernyms = (
-                    next_synset.hypernyms() + next_synset.instance_hypernyms()
-                )
-                if not next_hypernyms:
-                    result.append(next_synset)
-                else:
-                    todo.extend(next_hypernyms)
-        return result
+        for path in self._get_hypernym_paths():
+            root = path[0]
+            if root not in seen:
+                seen.add(root)
+                roots.append(root)
+        return roots
 
     # Simpler implementation which makes incorrect assumption that
     # hypernym hierarchy is acyclic:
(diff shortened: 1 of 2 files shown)


ekaf:
Thanks @medisean, this is a great proposal, so it's a pity that pre-commit fails. Normally we don't review PRs before the tests pass, and this would risk being closed as stale.

PR #3783, opened 2026-08-25
[WIP] Execute every advisory security probe and tighten the guards they cover

alvations:
Extends the living per-advisory probe harness (#3749) so each probe **runs the
attack** instead of grepping source, and tightens two guards found to be leaky
while doing so.

## Probes: STATIC → executed FIXED (40 FIXED, 0 VULNERABLE, 0 STATIC)

All 7 previously-STATIC probes now execute a real, offline exploit, so a silent
guard regression surfaces as VULNERABLE on every commit rather than passing:

| Advisory | Executed exploit |
|---|---|
| 3gq4 (LinThesaurus) | symlink named to match the `sim[A-Z].lsp` scan so the guard actually fires |
| 6hwm (Graphviz `dot`) | plant `./dot`; `find_binary` must refuse the CWD binary |
| 469j (downloader AFO) | real zip-slip member via `_unzip_iter`; outside canary stays intact |
| 5wp5 (integrity) | `file://` download with wrong sha256 → refused before commit; matching digest installs |
| f794 (hardlink write) | drive `pathsec.open` through symlink/hardlink/absolute escapes |
| gfwx (XSS) | in-process `do_GET` on `/lookup_<script>` → reflected only HTML-escaped |
| jm6w (bind) | spy server proves `wnb()` binds loopback only |

## Guard tightenings

- **`pathsec._hardened_open`**: deferred `O_TRUNC` until after the hardlink /
  realpath checks — a refused write used to *zero* a hardlinked outside-root
  target before `st_nlink` refused it.
- **`config_weka`**: bounds `WEKAHOME` to the pathsec sandbox (`validate_path`),
  so an attacker-influenced env var cannot add a `weka.jar` from outside the
  allowed roots (CWE-426/427).

## Negative-control teeth
New teeth for gfwx (neuter `html.escape`) and f794 (bypass `_hardened_open`):
both flip to VULNERABLE when the guard is broken, proving the probes reach the sink.

## Verified (not mocked)
Real weka.jar train+classify on OpenJDK 26; 279 modules imported; real corpora /
gzip streaming / tagger; full unit suite **2052 passed, 0 failed**; 4 lint hooks clean.

Marked `[WIP]` pending review alongside the other decomposition PRs.

### DIFF
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -14,6 +14,7 @@ import re
 import subprocess
 import tempfile
 import time
+import warnings
 import zipfile
 from sys import stdin
 
@@ -21,6 +22,7 @@ from nltk.classify.api import ClassifierI
 from nltk.data import make_staging_dir
 from nltk.internals import config_java, java
 from nltk.pathsec import open as pathsec_open
+from nltk.pathsec import validate_path
 from nltk.probability import DictionaryProbDist
 
 _weka_classpath = None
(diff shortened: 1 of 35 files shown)


alvations:
Closing as superseded. Every change in this WIP branch has since landed on `develop` through smaller, reviewed PRs:

- **Guard hardening** (all present on `develop` today):
  - `classify/weka.py` — `WEKAHOME` bounded to the pathsec sandbox via `validate_path` (CWE-426/427)
  - `corpus/reader/framenet.py` — `Security violation:` containment marker on the rejection
  - `pathsec.py` — deferring `O_TRUNC` until after the hardlink/realpath checks
  - `picklesec.py` — denying `posixpath` / `ntpath` / `genericpath` (the concrete `os.path` backing modules)
- **Advisory-probe harness** — superseded by #3820 (merged) and the ongoing GHSA-8mgp umbrella work; every probe now executes its attack and the associated tests (`test_advisory_probes`, `test_pathsec`, `test_pickle_allowlist_security`, `test_weka_integration`, `test_weka_security`) are on `develop`.

No unique content remains here. Reopen if anything was missed.

PR #3786, opened 2026-08-26
[WIP] Make the GHSA-8mgp probe reach its sink and give it a negative control

alvations:
Found while auditing the stale `fix-ghsa-8mgp` branch (which turns out to be
superseded — merging it would *revert* `pathsec.py`).

## The probe could not detect a regression

`ghsa_8mgp_746c_j5xp.py` read `<tempdir>/../../etc/passwd`. Two levels up from a
deep temp directory does not reach `/etc`:

```
payload:       /var/folders/cq/9c.../T/../../etc/passwd
normalises to: /var/folders/cq/etc/passwd      <- does not exist
```

So it scored FIXED off a `FileNotFoundError`, never touching a guard. Proof — with
**both** containment layers disabled it still passed:

```
normalizer off + ENFORCE off  -> FIXED    # should have leaked
```

A probe that passes with every guard removed is not testing anything.

## Fix

- Fire payloads that genuinely reach the sink: traversal, absolute `/etc/passwd`,
  `nltk:` and `file://` URLs, and a NUL byte.
- Require a **security-marked** refusal (`is_security_rejection`) rather than any
  exception, so an incidental error is not scored as a defence.
- Treat *reading the file's contents* as the leak, not merely "the call returned".
- Load with `cache=False`: `nltk.data.load` memoises by URL, so a successful read
  in one run was replayed into the next and reported a leak that was not there
  (the probe was not idempotent).

## Negative control

The probe never had one. Now:

| state | result |
|---|---|
| unmodified tree | FIXED |
| normaliser + `ENFORCE` removed | **VULNERABLE** |
| restored | FIXED |
| run twice in a row | FIXED, FIXED (idempotent) |

Harness stays at 40 probes, 0 VULNERABLE; `test_advisory_probes.py` 25 passed.

> Note: #3783 carries the same probe rewrite as part of its wider STATIC→FIXED
> sweep. Whichever lands first, I will rebase the other — this PR is the
> standalone version so the false-FIXED can be fixed independently.

Marked `[WIP]` alongside the other decomposition PRs.

### DIFF
--- a/nltk/classify/maxent.py
+++ b/nltk/classify/maxent.py
@@ -1441,7 +1441,9 @@ def train_maxent_classifier_with_megam(
     # Write a training file for megam.
     try:
         fd, trainfile_name = tempfile.mkstemp(prefix="nltk-")
-        with open(trainfile_name, "w") as trainfile:
+        with open(  # sandboxed-open ok: internal mkstemp name, not caller-supplied
+            trainfile_name, "w"
+        ) as trainfile:
             write_megam_file(
                 train_toks, encoding, trainfile, explicit=explicit, bernoulli=bernoulli
             )
@@ -1543,7 +1545,9 @@ class TadmMaxentClassifier(MaxentClassifier):
 
         call_tadm(options)
 
-        with open(weightfile_name) as weightfile:
+        with open(  # sandboxed-open ok: internal mkstemp name, not caller-supplied
+            weightfile_name
+        ) as weightfile:
             weights = parse_tadm_weights(weightfile)
 
         os.remove(trainfile_name)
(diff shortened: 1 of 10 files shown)


alvations:
Closing this WIP. Its changes are now accounted for:

- **Model-loader guard hardening** is on `develop`: `tag/perceptron.py` (`_reject_path_structure` + pathsec), `tag/crf.py` (`validate_path` on model load), `tbl/demo.py` (`pathsec_open` for `cache_baseline_tagger`), and the `pathsec.py` whitespace-only-path guard. `test_model_artifact_pathsec.py` is on `develop` too.
- **`classify/maxent.py`** was hardened *more strongly* on `develop` than the approach here — its model opens route through `pathsec_open` rather than being annotated bare opens, so the marker changes here are not needed.
- **The GHSA-8mgp probe teeth** (reaching the sink + negative control) live in the GHSA-8mgp umbrella branch (the full ~256-line probe), which will land on `develop` with that branch; `develop` currently carries the stub.
- **The one genuinely unique piece** — extending `check_no_unsandboxed_open.py` `GUARDED_PATHS` to the model-artifact loaders — has been salvaged into its own focused PR: #3848. It passes against `develop` as-is (those modules are already sandboxed) and the guard was verified to have teeth.

Nothing else remains. Reopen if something was missed.

PR #3787, opened 2026-08-26
[WIP] Bound external-tool model paths to the pathsec sandbox

alvations:
Part of the #3753 decomposition. Covers the external-tool wrappers, which pass caller-supplied model and corpus paths straight into the child process argv.

## The leak

With a checksum-approved jar (the jar allowlist gates before `java()`), all four caller-controlled `StanfordSegmenter` paths reached the JVM unvalidated:

```
LEAKED -loadClassifier   -> ~/.nltk_out_xxx/evil.ser.gz
LEAKED -serDictionary    -> /etc/passwd
LEAKED -sighanCorporaDict-> ~/.nltk_out_xxx
LEAKED -textFile         -> /etc/shadow
```

`MaltParser` was worse than a read: `-w` is where MaltParser *writes* the `.mco` in learn mode, and it defaulted to `os.getcwd()` for an untrained model and `tempfile.gettempdir()` (world-writable on Linux, predictable) for the CoNLL intermediates.

`StanfordParser` passed `-model` through unchecked, including `edu/stanford/../../../../etc/passwd`.

## The fix

- **Segmenter** validates all four paths before the `java()` hand-off.
- **MaltParser** allocates `-w` lazily via `nltk.data.make_staging_dir` — inside a data root, mode 0700, unpredictable name. Same pattern already used by `PerceptronTagger` and the named-entity chunker. A model with a directory component is bounded with `validate_path`; `working_dir` gained a validating setter.
- **StanfordParser** bounds `-model` at `_execute`, the single choke point for all three callers.
- **New `pathsec.validate_model_resource`** for arguments that may be *either* a filesystem path or a jar-internal classpath resource (the shipped `englishPCFG.ser.gz` default). Resource names pass, real paths are sandboxed, and empty / option-like / URL / `..` / NUL / `~` / UNC / NTFS-stream values are refused. It also rejects a **hardlinked** model (realpath cannot distinguish it from an ordinary in-root file) and **non-regular files** such as a planted FIFO, which would block the reader forever.

## Deliberately NOT changed

`candc`/`boxer`'s `bin_dir` and REPP's tokenizer dir are **install directories**, not data paths. `validate_path` permits only NLTK data roots, so bounding them would break every real installation. I verified their existing absolute-path (CWE-426) checks still refuse relative and CWD-relative values, and added tests pinning that distinction so a later refactor cannot collapse the two classes.

## Verification

Not mocked. Every escape test drives the real code path with only the `java` hand-off replaced, so a probe cannot pass because the test assembled the argv itself.

- Escape matrix per sink: 16 hostile vectors (symlink, hardlink, traversal, NUL, argv injection, newline, URL, empty, ...) plus 14 exotic ones (device nodes, FIFO, `~`, UNC, NTFS stream, fullwidth solidus, percent-encoding, case-folding). All refused.
- Over-block controls: the jar-internal default, in-sandbox models, corpus directories and Java inner-class `$` names all still work.
- Negative-control teeth tests: neuter a guard, the probe flips to reachable.
- **Real end-to-end against an actual JVM**: trains a real MaltParser model (lands in the staging dir, CWD stays clean) and parses with the real Stanford parser. These skip cleanly where no JVM is installed — the probe runs `java -version` rather than trusting a `java` on PATH, because macOS ships a stub that resolves and then fails at exec.
- New import sweep over **every** nltk module plus a functional battery against real corpora and models, so a too-strict guard fails visibly instead of only breaking ordinary use.

Full suite: 2166 passed, 0 failed. Advisory probe gate: 24 passed, 0 VULNERABLE. The 5 failing doctest files in the wider tree (`wordnet`, `wsd`, `sentiwordnet`, `treeprettyprinter`, `internals`) fail identically on pristine develop — pre-existing, not from this branch.

Marked WIP pending review of the remaining wrapper cluster.

### DIFF
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -12,7 +12,7 @@ import os
 import sys
 import tempfile
 
-from nltk.data import ZipFilePathPointer
+from nltk.data import ZipFilePathPointer, make_staging_dir
 from nltk.internals import (
     find_dir,
     find_file,
@@ -22,6 +22,7 @@ from nltk.internals import (
 from nltk.parse.api import ParserI
 from nltk.parse.dependencygraph import DependencyGraph
 from nltk.parse.util import taggedsents_to_conll
+from nltk.pathsec import validate_model_resource, validate_path
 
 
 def malt_regex_tagger():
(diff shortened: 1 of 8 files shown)


alvations:
Superceded by more extended work in #3753 Closing this.

PR #3791, opened 2026-08-26
Restore pre-commit formatting and register the new probe id on develop

alvations:
`develop` is currently red, and every open pull request inherits it. Two causes, both from the merge that added the GHSA-8846 probe.

**1. pre-commit was not run.** black and isort fail on four files:

```
reformatted nltk/test/unit/security_probes/ghsa_8846_p9w9_5frf.py
reformatted nltk/test/unit/test_advisory_coverage_ci.py
reformatted nltk/test/unit/test_wordnet_recursion.py
reformatted nltk/util.py
4 files reformatted, 514 files left unchanged.
```

Reformatting them is the whole of that part of the change. No logic moves.

**2. The new probe's id was never registered.** The merge added `nltk/test/unit/security_probes/ghsa_8846_p9w9_5frf.py` but did not add `GHSA-8846-p9w9-5frf` to `_CLOSED_WITH_PROBE`, so the *Advisory coverage (GitHub)* job fails:

```
AssertionError: probes for ids GitHub does not list: ['GHSA-8846-p9w9-5frf']
```

That set exists for precisely this case, a probe whose advisory GitHub does not list yet, so the id is added there rather than changing the assertion.

Verified locally: all pre-commit hooks pass, and `pytest nltk/test/unit` is 2119 passed, 45 skipped, 9 xfailed, 0 failed.

Separated from my other open work deliberately, since this is unrelated to any of it and unblocks everyone.

### DIFF
--- a/nltk/test/unit/security_probes/ghsa_8846_p9w9_5frf.py
+++ b/nltk/test/unit/security_probes/ghsa_8846_p9w9_5frf.py
@@ -1,18 +1,22 @@
-from ._base import FIXED, VULNERABLE, probe
 from nltk.util import acyclic_depth_first
+
+from ._base import FIXED, VULNERABLE, probe
+
 
 class MockNode:
     def __init__(self, name, children=None):
         self.name = name
         self._children = children or []
+
     def rel(self):
         return self._children
+
 
 @probe("GHSA-8846-p9w9-5frf")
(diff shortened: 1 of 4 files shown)


ekaf:
Thanks @alvations!

PR #3794, opened 2026-08-28
Harden pickle deserialization (picklesec) against arbitrary-code-execution gadgets

alvations:
## What this does

`nltk/picklesec.py` provides the hardened unpicklers that back NLTK's pickle-loading paths, defending against the deserialization RCE class (arbitrary-code-execution "gadgets"):

- **`RestrictedUnpickler`** blocks all globals: `find_class` always raises, so no callable can be reconstructed from a pickle at all.
- **`AllowlistUnpickler`** permits only an explicit set of `(module, name)` pairs or module prefixes, with several backstops that hold even when a caller passes a broad allow:
  - **extension opcodes are rejected** (no `EXT1`/`EXT2`/`EXT4` copyreg escape),
  - **dotted names are denied** before any `getattr` chain runs,
  - **dunder names are denied** so a prefix allow cannot walk into module internals,
  - a **denied-module prefix list** is applied to every request, even allowlisted ones.

This change extends the denied-module prefix list with the concrete `os.path` backing modules (`posixpath`, `ntpath`, `genericpath`). Denying `os` only catches the literal `os.path`; without naming the backing modules, a broad allow of one of them could still reach `path` / `stat` / `expandvars` gadgets, which form an environment-variable and filesystem oracle.

## Tests

- **`test_pickle_allowlist_security.py`** — allowlist-bypass regressions, exact-pair vs broad-namespace behavior, denied-module backstops, and the new `os.path` backing-module denial (a unit check plus an end-to-end `expandvars` oracle that must not read the environment).
- **`test_pickle_gadget_landscape.py`** — a wide sweep of pickle gadget primitives compiled from published NLTK and ML-library RCE research, run against the allowlisted unpickler. Payloads are staged inside the data sandbox so it is the picklesec guards, not the path check, that refuse them.
- **`test_pickle_gadgets_reach_the_unpickler.py`** — proves gadget pickles are refused by picklesec itself (an `UnpicklingError` naming the forbidden global), not merely by the path layer, by writing payloads inside the sandbox under legitimate resource names.
- **`test_pickle_load_warnings.py`** — pickle load-time warning behavior.

## CI

`requirements-ci.txt` now installs `numpy`, `pandas`, and `scipy` explicitly. These libraries are the real deserialization attack surface the gadget probes target, so making them present guarantees those probes execute in CI rather than skipping.

## Scope

This is a self-contained split from the larger GHSA-8mgp pickle-hardening effort, isolated so the pickle-deserialization work can be reviewed on its own.

### DIFF
--- a/nltk/picklesec.py
+++ b/nltk/picklesec.py
@@ -141,6 +141,11 @@ _DENIED_MODULE_PREFIXES = (
     "os",
     "nt",
     "posix",
+    # os.path's concrete backing modules: denying "os" only catches the literal
+    # "os.path", not these, so an allowlist naming them would reach path/stat gadgets.
+    "posixpath",
+    "ntpath",
+    "genericpath",
     "subprocess",
     "sys",
     "socket",
--- a/nltk/test/unit/test_pickle_allowlist_security.py
+++ b/nltk/test/unit/test_pickle_allowlist_security.py
@@ -203,6 +203,25 @@ def test_denied_module_backstop_even_if_allowlisted():
         up2.find_class("builtins", "eval")
 
 
+@pytest.mark.parametrize("module", ["posixpath", "ntpath", "genericpath"])
+def test_os_path_backing_modules_are_denied(module, tmp_path):
+    """GHSA-x99w: os.path is denied, but so must be the concrete modules it *is*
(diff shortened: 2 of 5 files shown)


alvations:
Merging this in so that the CI/CD is actually testing the gadgets from all the pickling from those third party.

PR #3795, opened 2026-08-28
Fix internals.doctest to match the hardened _validate_java_options allowlist

alvations:
## What this fixes

`nltk/internals.py` already validates `java_options` against an allowlist: only JVM memory/stack tuning, `-verbose`, `-server`/`-client`, and `--add-modules` are permitted. Anything else (including `-XX:` and `-D` flags) must be passed explicitly through `java(trusted_raw_options=...)`, since `-XX:OnError=`/`-XX:OnOutOfMemoryError=` execute a command and `-D` system properties can redirect class loading. That guard closes CVE-2026-12841 (CWE-88).

`nltk/test/internals.doctest` was not updated when that allowlist landed, so it still asserted the older permissive behavior:

```
>>> _validate_java_options(["-XX:+UseG1GC"])
>>> _validate_java_options(["-Dsome.property=value"])
>>> _validate_java_options(["-Xmx1g", "-Dsome.key=val", "-server"])
```

Against the current code those calls raise `ValueError: java_options contains a disallowed JVM/launcher flag: ...`, so the doctest fails under `make doctest` / `pytest --doctest-glob='*.doctest' nltk/test/internals.doctest`. (The main CI pytest gate runs `--doctest-modules`, which collects only `.py` docstrings, so it does not surface this.)

## Change

- Flips the three affected `_validate_java_options` examples to expect the `ValueError` (via `# doctest: +ELLIPSIS`).
- Adds a short note that the filter is an allowlist and how to pass a genuinely needed flag through `trusted_raw_options`.

Before: `1 failed`. After: `1 passed`, with `nltk/internals.py` unchanged.

Documentation/test only, no source change. This is a small self-contained fix split out from the larger GHSA-8mgp hardening effort; the `_validate_java_options` allowlist it documents is already on `develop`.

### DIFF
--- a/nltk/test/internals.doctest
+++ b/nltk/test/internals.doctest
@@ -178,12 +178,32 @@ wrappers pass are equivalent and also permitted:
 >>> _validate_java_options(["-mx2g"])
 >>> _validate_java_options(["-ms128m"])
 >>> _validate_java_options(["-ss4m"])
->>> _validate_java_options(["-XX:+UseG1GC"])
 >>> _validate_java_options(["-verbose:gc"])
 >>> _validate_java_options(["-server"])
 >>> _validate_java_options(["-client"])
->>> _validate_java_options(["-Dsome.property=value"])
->>> _validate_java_options(["-Xmx1g", "-Dsome.key=val", "-server"])
+>>> _validate_java_options(["-Xmx1g", "-server"])
+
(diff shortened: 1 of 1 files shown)


alvations:
Minimal doctest updates to make sure that JVM leaks dont pass through. Merging this in first.

PR #3796, opened 2026-08-28
Bound re_show's caller-supplied regex against ReDoS (CWE-1333)

alvations:
This is a self-contained split from the larger GHSA-8mgp hardening, isolating only the change to the top-level `nltk/util.py`.

## What changed

`nltk.util.re_show(regexp, string)` compiled the caller-supplied `regexp` with a plain `re.compile` and then ran it against a caller-supplied `string`. A catastrophically backtracking pattern such as `(a+)+$` matched against a short bait string (e.g. `"a" * 34 + "!"`) hangs the interpreter outright (CWE-1333, catastrophic regex backtracking / ReDoS). Because the regex engine holds the GIL while backtracking, nothing in-process can interrupt it.

The compile now routes through `nltk.redos.compile`, which already exists on `develop` and applies a wall-clock bound to matching. This mirrors the guard already in place for `nltk.tokenize.RegexpTokenizer` and `nltk.tag.RegexpTagger`, which compile caller-supplied patterns through `redos` too.

The diff is small: one new import (`from nltk import redos`) and one line in `re_show` swapped from `re.compile(...)` to `redos.compile(...)`.

## Self-contained

The only new dependency is `nltk.redos`, which is already present on `develop` (unchanged by this PR). `import nltk.util` succeeds on a clean `develop` checkout plus this file alone.

## Tests

`nltk/test/unit/test_redos_caller_patterns.py` (a focused subset of the caller-regex probe from the larger effort, trimmed to only the `nltk.util` cases) asserts, each in a subprocess so a hang surfaces as a timeout rather than wedging the run:

- a catastrophic pattern no longer hangs `re_show`;
- an ordinary pattern still yields the correct bracketed output;
- `nltk.util` source still routes through `redos.compile` (pins the mechanism, not just the timing).

Runs green on Linux, macOS, and Windows: the tests use only a subprocess and `PYTHONPATH`, with no temp-dir fixtures or network resolution.

### DIFF
--- a/nltk/test/unit/test_redos_caller_patterns.py
+++ b/nltk/test/unit/test_redos_caller_patterns.py
@@ -1,0 +1,77 @@
+# Natural Language Toolkit: ReDoS on caller-supplied regexes
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""Any API that compiles a CALLER's regex must bound it.
+
+``nltk.redos.compile`` puts a wall-clock limit on matching. ``nltk.util.re_show``
+takes both a pattern and a string from the caller and previously compiled the
+pattern with a plain ``re.compile``, so a catastrophically backtracking pattern
+such as ``(a+)+$`` against a 34-character bait string hung the interpreter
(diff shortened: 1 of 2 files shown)


alvations:
Tighten the regex show function wrap over redos.py 

Merging this in first so that this always gets checked properly.

PR #3798, opened 2026-08-28
Refuse CWD-relative matches for bare names in find_file/find_dir (CWE-427)

alvations:
## What changed

`nltk.internals.find_file_iter` probes the current working directory for a bare
`filename` before consulting the configured environment variables and
searchpath. Because a returned bare relative path is later run or loaded
relative to the CWD, a match found only in the CWD lets an attacker who can
write to that directory choose the tool, model, or corpus directory that NLTK
picks up (CWE-426 untrusted search path / CWE-427 uncontrolled search element).
`find_binary` already refused a CWD-relative bare match; `find_file` and
`find_dir` did not, so e.g. `find_file("maltparser")` returned `"maltparser"`
resolving against the CWD.

This adds `_reject_cwd_relative_bare_match`, applied by `find_file` and
`find_dir`:

- When the requested name has **no directory component**, any non-absolute
  (CWD-relative) match is dropped, and if that was the only kind of match a
  `LookupError` is raised with guidance to install on a search path, set the
  env var, or pass an absolute path.
- A name **with** a directory component (`tools/maltparser`), an **absolute**
  path, and **env-var / searchpath** hits (which resolve to absolute paths) all
  survive, so legitimate discovery is unaffected.

## Vulnerability class

CWE-426 / CWE-427 binary/resource hijack via the current working directory for
bare tool/resource names.

## Tests

New `nltk/test/unit/test_find_file_cwd_hardening.py`:

- refuses a CWD-planted bare name for `find_file` and `find_dir`;
- an env-var match survives the filter and is not shadowed by a same-named file
  in the CWD;
- an explicit relative path (with a directory component) and an absolute path
  are still honoured.

Three of these fail on `develop` without the fix and pass with it; the rest
guard against regressing legitimate discovery.

## Cross-platform

Fixtures are staged under `$HOME` (not `/tmp`), and `find_file`/`find_dir` do
not route through the pathsec sandbox, so no data-root registration is needed on
any OS. Assertions are either `LookupError` expectations (OS-agnostic) or
identity / `startswith` comparisons against strings constructed the same way
they are returned, so there are no POSIX-only resolved-string assumptions. The
bare-name hijack case is constructed with a bare name (no path separators) and
is refused on Linux, macOS, and Windows alike.

## Scope

Self-contained split from the larger GHSA-8mgp hardening effort, limited to the
`nltk/internals.py` `find_file`/`find_dir` change and its focused tests. It
imports nothing beyond what `develop` already ships and does not modify
`nltk/test/internals.doctest`.

### DIFF
--- a/nltk/internals.py
+++ b/nltk/internals.py
@@ -924,11 +924,48 @@ def find_file_iter(
         raise LookupError(f"\n\n{div}\n{msg}\n{div}")
 
 
+def _reject_cwd_relative_bare_match(filename, matches, kind):
+    """Filter out an untrusted CWD-relative match for a bare tool name.
+
+    ``find_file_iter`` probes the current working directory for a bare
+    ``filename`` before the configured env vars / searchpath, so a planted
+    ``./<name>`` or ``./<name>/<name>`` is returned first and, because a returned
+    relative path is later run or loaded relative to the CWD, an attacker who can
+    write to the CWD chooses the tool (CWE-426 / CWE-427). ``find_binary``
+    already refuses this; ``find_file`` / ``find_dir`` did not.
+
+    Only a *bare* name (no directory component) is filtered, and only when the
+    match is not absolute: a caller who passed ``tools/maltparser`` or an
+    absolute path made an explicit choice and is honoured. An env var or
+    searchpath hit is absolute, so it survives.
+    """
+    if os.path.dirname(filename) != "":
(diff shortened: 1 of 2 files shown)


alvations:
This should block the cwd from leaking without just in case users start path traversal through find_dir or find_file. Merging this in.

Issue #3740, opened 2026-08-08
Standardize import alias for `pathsec.open` to address remaining file I/O security gaps


ekaf:
### The Problem
Currently, the codebase uses inconsistent aliasing when importing `open` from `nltk.pathsec`. Because the function name shadows Python's built-in `open()`, it requires aliasing on import, leading to different conventions across different modules. 

For example, current implementations include:
* `from nltk.pathsec import open as _secure_open` (in `data.py`)
* `from nltk.pathsec import open as pathsec_open` (in `downloader.py`)

This lack of consistency increases cognitive overhead and complicates codebase audits for restrictive file operations.

### The Proposed Solution
Instead of requiring a unique alias per module, a standard, non-shadowing alias should be defined directly at the source. 

The proposal is to add `secure_open = open` to the module level of `pathsec.py`.

This allows all downstream modules to use a unified, clean import:
```python
from nltk.pathsec import secure_open
```

Making `secure_open` a public alias (without an underscore) establishes it as the standard, non-shadowing API for downstream usage, while leaving the existing `open` definition intact to prevent any breaking changes.

### Why this is a priority
Establishing this standard alias is a convenient prerequisite for patching the remaining file I/O vulnerabilities in the project. There are still multiple instances of the built-in Python `open()` being used where restrictive file access is required. Standardizing `secure_open` provides a uniform foundation to confidently replace those remaining insecure calls across the entire codebase.

alvations:
A file-I/O security sweep found no live gap behind the alias inconsistency: every untrusted-path open in `nltk/corpus/reader/`, `nltk/corpus/util.py`, and `nltk/data.py` already routes through `nltk.pathsec.open` (via `pathsec_open` / `_secure_open`). The two aliases are only a legibility issue, not an open bypass.

Rather than a one-off rename (which can silently drift again), #3744 adds an AST-based pre-commit guard that **fails CI if a bare builtin `open(` is introduced** in those sandbox-sensitive modules — making the sandboxing policy enforceable regardless of import alias, which is the security intent here. A reviewed exception can be annotated with `# sandboxed-open ok: <reason>`.

alvations:
@ekaf let me know if #3744 is sufficient for the checks. Meanwhile, feel free to standardize the secure open function name but I think it'll be a significant effort since we'll have to double check again and modify all the tests put in place.

PR #3799, opened 2026-08-28
Strengthen sandboxed-open guard to catch compression-helper bypasses

alvations:
## What

Strengthens the existing `tools/check_no_unsandboxed_open.py` guard (added in #3740) so it catches one more way a sandbox-sensitive module can open a path outside the pathsec sandbox, in addition to the bare builtin `open()` it already flags.

This is a self-contained split from the larger GHSA-8mgp file-I/O hardening effort: only the dev tool is touched here. The pre-commit hook that runs it (`no-unsandboxed-open`) is already registered on `develop`, so no config change is needed.

## The detection classes

The guard is a static AST check (not a regex) over the guarded modules. It now reports:

1. **Bare builtin `open(...)`** (unchanged) - an `ast.Name` named `open`. Attribute or differently named calls such as `pathsec_open(...)`, `self.open(...)`, `fp.open()` and `os.open(...)` are not the builtin and are ignored.
2. **Compression / archive helper handed a path** (new) - `gzip.open`, `bz2.open`, `lzma.open`, `codecs.open`, `io.open`, `tarfile.open`, `zipfile.ZipFile`. These bypass the pathsec sentinel exactly as a bare `open()` does when given a path. Handed an already-secured pathsec file object (or a local name bound from one), they are fine and not reported.

## Reviewed-exception marker

Unchanged. A deliberate, reviewed call is annotated with a trailing `# sandboxed-open ok: <reason>` comment on the same line as the call; the whole physical span of a wrapped call is scanned so the marker still counts after formatting.

## Guarded set

`GUARDED_PATHS` keeps the modules already covered on `develop` and adds the newly hardened ones, all clean under both checks:

- `nltk/corpus/reader` (kept)
- `nltk/corpus/util.py` (kept)
- `nltk/data.py` (kept)
- `nltk/picklesec.py` (added)
- `nltk/util.py` (added)
- `nltk/internals.py` (added)

`nltk/pathsec.py` is deferred until its own `zipfile.ZipFile(...)` carries a reviewed `# sandboxed-open ok` annotation (not on `develop` yet). The guarded set grows as each module's conversion merges; narrowing scope (never the detection rule) keeps the hook green on `develop` today.

## Deferred: temp-file-factory check

A third class - a temp-file factory (`tempfile.mkstemp` / `NamedTemporaryFile` / `TemporaryFile`) called without `dir=`, which drops scratch output into the shared system temp directory - is left to a follow-up. Its sanctioned in-root destination, `nltk.data.staging_tempdir()`, is not on `develop` yet (only `make_staging_dir` is), so there is no remedy to point offenders at. It lands together with that helper and the corpus-reader temp-file pins.

## Green on develop

```
$ python tools/check_no_unsandboxed_open.py
sandboxed-open guard: OK (no un-sandboxed open() in guarded modules)
```

### DIFF
--- a/tools/check_no_unsandboxed_open.py
+++ b/tools/check_no_unsandboxed_open.py
@@ -4,24 +4,33 @@
 # Copyright (C) 2001-2026 NLTK Project
 # URL: <https://www.nltk.org/>
 # For license information, see LICENSE.TXT
-"""Fail if a bare builtin ``open(`` is used in a sandbox-sensitive module.
+"""Fail if a sandbox-sensitive module opens a path outside the pathsec sandbox.
 
 Modules that resolve *untrusted* paths (corpus readers reading corpus files,
 ``nltk.data`` resolving resource names) must open files through the pathsec
-sandbox -- ``from nltk.pathsec import open as pathsec_open`` -- so path
-traversal / symlink-TOCTOU protections apply (CWE-22/59). The two historical
+sandbox (``from nltk.pathsec import open as pathsec_open``) so that path
+traversal / symlink-TOCTOU protections apply (CWE-22/59). The historical
 aliases (``pathsec_open`` / ``_secure_open``) both point at ``nltk.pathsec.open``;
 this guard makes the *policy* enforceable regardless of the alias, which is the
-security intent behind issue #3740 (superseding a one-off alias rename that could
-silently drift again).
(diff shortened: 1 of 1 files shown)


alvations:
This has to be done. Cos we should be guarding the whole `nltk` library instead of individual submodules. Closing this first.

PR #3801, opened 2026-08-28
Bound RegexpStemmer's caller-supplied pattern against ReDoS

alvations:
RegexpStemmer compiled its caller-supplied pattern with plain `re.compile` and then applied it to caller text. A catastrophically backtracking pattern such as `(a+)+$` against a short word hangs the interpreter (CWE-1333, ReDoS).

## Fix

Route the pattern through `nltk.redos.compile`, which puts a wall-clock bound on matching. This mirrors the guard `nltk.util.re_show` received in #3796.

```python
if not hasattr(regexp, "pattern"):
    # caller-supplied pattern applied to caller text
    regexp = redos.compile(regexp)
```

## Test

A focused new module `nltk/test/unit/test_stem_regexp_redos.py` asserts:

1. a catastrophic pattern does not hang `RegexpStemmer`,
2. an ordinary pattern (`ing$|s$|e$`) still stems (`running` -> `runn`), and
3. the `nltk.stem.regexp` source routes through `redos.compile` (pins the mechanism, not just the timing).

The hang assertion runs the stemming in a **subprocess** with a wall-clock timeout. An in-process timeout cannot fail cleanly on catastrophic backtracking: the regex engine holds the GIL, so a pytest timeout plugin cannot interrupt it and the whole run wedges. A subprocess timeout is also cross-platform: it works on Windows, where the POSIX-only `signal.SIGALRM` approach silently does nothing. The child timeout (20s) sits comfortably above `nltk.redos.DEFAULT_TIMEOUT` (5s) but well under any per-test CI limit, and the test has no temp-dir/network/dbm dependencies.

This is split out from the larger GHSA-8mgp ReDoS effort so it is self-contained on `develop`: `nltk.redos` already ships there (via #3796), and no other new symbol is introduced.

### DIFF
--- a/nltk/stem/regexp.py
+++ b/nltk/stem/regexp.py
@@ -8,6 +8,7 @@
 # For license information, see LICENSE.TXT
 import re
 
+from nltk import redos
 from nltk.stem.api import StemmerI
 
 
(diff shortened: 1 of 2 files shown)


alvations:
Added the redos tightening to also the stemmer. Merging this in first.

PR #3802, opened 2026-08-28
Deserialize chat80 shelve values through the restricted unpickler (RCE fix)

alvations:
## Vulnerability

`nltk.sem.chat80.val_load` loaded its persistent valuation with `shelve.open()`. `shelve` reads every stored value with the default, unrestricted `pickle.Unpickler`, so a `.db` file whose value bytes are a crafted pickle gadget executes arbitrary code the moment `val_load` reads it back.

`val_load` already validates the `.db` path, but `validate_path` only checks WHERE the file is, not WHAT is inside it. A hostile shelf planted inside the data sandbox (for example shipped in a data package) therefore still ran on load.

## Fix

Add `_restricted_shelve_open()`, which wraps the real shelf in a `_RestrictedShelf` whose `__getitem__` deserializes the raw value bytes through `nltk.picklesec.RestrictedUnpickler` (which forbids every global at `find_class`). Chat-80 valuations are only sets and tuples of strings, which need no globals, so legitimate data still loads while the gadget is refused. `val_load` now reads its values through this wrapper (`Valuation(db_in.items())`).

`RestrictedUnpickler` is already on `develop` (merged via #3794), so this change is self-contained.

## Tests

`nltk/test/unit/test_chat80_shelve_rce.py`:

- `test_restricted_shelf_refuses_a_gadget_value` -- a value whose pickle references a module global raises `UnpicklingError` on read and never executes.
- `test_restricted_shelf_round_trips_benign_valuations` -- over-block control: sets and tuples of strings still load.
- `test_val_load_end_to_end_does_not_execute_gadget` -- drives `val_load` itself (path gate, restricted read and `Valuation` wiring).

### Cross-platform dbm handling

`shelve` sits on `dbm`, whose backend and on-disk filenames differ by platform (`dbm.gnu`, `dbm.ndbm`, `dbm.sqlite3`, `dbm.dumb`; a base path may become `base`, `base.db`, or `base.dir` / `base.dat`). Every shelf is created and reopened with `shelve.open(base)` so the tests run on whatever backend the platform selects, and no test hardcodes a `.db` suffix or a specific backend. The end-to-end test hard-links the real backend file to the `base + '.db'` name that `val_load`'s `os.access` gate expects and skips only if a backend cannot be gated portably, so the core invariant is always asserted on every OS.

Self-contained split from the larger GHSA-8mgp-746c-j5xp hardening effort.

### DIFF
--- a/nltk/sem/chat80.py
+++ b/nltk/sem/chat80.py
@@ -123,6 +123,7 @@ current directory.
 
 """
 
+import io
 import os
 import re
 import shelve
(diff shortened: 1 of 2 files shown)


alvations:
Write the shelves.open in chat80 with `RestrictedUnpickler`. Merging this in first.

PR #3804, opened 2026-08-28
Harden downloader Package attributes and extraction against traversal/Zip-Slip

alvations:
## Summary

A malicious remote `index.xml` could set a package's `subdir` or `id` to escape the download directory (CVE-2026-33236, a Zip-Slip / path-traversal class). Because both attributes are joined into the install path (`os.path.join(subdir, id + ext)`), a `subdir` of `../../../../tmp`, a rooted `/tmp`, a drive-prefixed `C:\evil`, or an `id` of `../../evil` (or one containing a path separator or NUL) wrote files outside the download dir.

This is a self-contained split from the larger GHSA-8mgp hardening effort, touching only `nltk/downloader.py` and adding one focused test. It depends only on `nltk.pathsec` / `nltk.data` symbols already on `develop`.

## Vulnerability classes addressed

- **Absolute / rooted / drive-letter subdir.** `Package` now rejects a `subdir` that is absolute under posix rules (`posixpath.isabs`), under windows rules (`ntpath.isabs`), or that carries a drive prefix (`ntpath.splitdrive`). Python 3.13's `ntpath.isabs` no longer treats a bare `/tmp` as absolute, so `os.path.isabs` alone let a rooted subdir through on Windows; the posix/windows/drive combination closes that gap on every platform.
- **Parent-reference traversal.** A `subdir` containing `..` in either slash style is rejected, and an `id` containing a path separator, a NUL, or `..` is rejected.
- **Zip-Slip / extraction confinement + integrity.** The download lock path is bounded with `validate_path` before `O_EXCL` creation, and the md5/sha256 integrity helpers read through the pathsec sandbox, so a checksum can no longer be computed over a file outside the download dir.

## Windows-specific fix

The recent change adds `import ntpath` / `import posixpath` and checks absoluteness under both rule sets plus a drive prefix. Without it, `os.path.isabs('/tmp')` returns `False` on Windows under Python 3.13 and a rooted subdir would be accepted.

## Tests

`nltk/test/unit/test_downloader_package_traversal.py` pins that crafted `subdir` / `id` attributes are refused across posix-traversal, windows-backslash, drive-letter, absolute, and NUL forms, while benign relative subdirs (`''`, `tokenizers`, `a/b`) build a contained filename. The refusals hold cross-platform with no platform gating: the drive-letter case is refused via `ntpath` even where `os.path.isabs` is `False`, which is what makes the delta observable outside Windows too.

Verified against `develop` + this `downloader.py` alone; the existing `test_downloader_atomic.py` and `test_downloader_unzip.py` continue to pass unchanged.

### DIFF
--- a/nltk/downloader.py
+++ b/nltk/downloader.py
@@ -161,7 +161,9 @@ default: unzip or not?
 """
 import functools
 import itertools
+import ntpath
 import os
+import posixpath
 import subprocess
 import sys
 import textwrap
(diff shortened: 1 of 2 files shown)


alvations:
Route all download through pathsec. Merging this in first.

PR #3805, opened 2026-08-28
Harden nltk.xmlsec.parse: read filename sources through pathsec.open

alvations:
## What

`nltk.xmlsec.parse()` accepts a caller-controlled filename. In the standard-library fallback path (used when `defusedxml` is not installed) it read that file with a bare builtin `open()` after `validate_path()`. This routes the read through `nltk.pathsec.open()` instead, so the open is covered by the same sandbox as the rest of `nltk.data`: path containment plus symlink/hardlink TOCTOU hardening (CWE-22 / CWE-59). The `defusedxml` fast path and the entity-expansion refusal (CWE-776) are unchanged.

The diff to `nltk/xmlsec.py` is two lines: import `nltk.pathsec.open` and use it for the filename read.

## Vulnerability classes closed / reinforced

- **XXE / external-entity file disclosure** and **billion-laughs / entity-expansion DoS (CWE-776)** are refused on both back ends (already the module's contract; covered here by an end-to-end test through a file source).
- **External-DTD SSRF**: the parser never fetches an external DTD; verified against a real listening socket.
- **Path traversal / symlink TOCTOU (CWE-22 / CWE-59)** on the filename read: now goes through the pathsec sandbox rather than a bare `open()`.

## Design

`nltk.xmlsec` uses `defusedxml` when installed and falls back to an equivalent standard-library implementation (an expat pre-scan that rejects any entity declaration or external reference) otherwise, so protection holds either way and `defusedxml` stays optional. The bare-`open` swap only affects the fallback path; the `defusedxml` path opens through its own reader.

## Tests

Adds `nltk/test/unit/test_xml_attack_matrix.py`:

- external-entity file disclosure, billion-laughs expansion, and parameter-entity payloads fed to `xmlsec.parse()` from a **file source** are refused;
- a real-socket check that an external DTD is never fetched (SSRF);
- a benign-document control that must still parse.

Every case runs against **both back ends** (`defusedxml` and the stdlib fallback); the fallback is the one that exercises the pathsec-guarded read introduced here. The existing `nltk/test/unit/test_xmlsec.py` continues to pass unchanged.

Teeth: the same billion-laughs payload expands to 20000 characters under the raw stdlib `ElementTree` parser, and is refused (`EntitiesForbidden`) by `xmlsec` on both back ends, so the assertions exercise the guard rather than a tautology.

## Cross-platform

XML parsing is platform-independent. The payloads reference `file:///etc/passwd` only as bait: the entity declaration is refused before any read, and the assertions check the raised exception, not file contents, so there is no POSIX path dependency. Documents are staged under a registered nltk data root (not a hard-coded `/tmp`) so the pathsec-guarded open resolves them on POSIX and Windows alike, and the SSRF check uses `127.0.0.1` with an ephemeral port.

Self-contained split from the larger GHSA-8mgp hardening effort.

### DIFF
--- a/nltk/test/unit/test_xml_attack_matrix.py
+++ b/nltk/test/unit/test_xml_attack_matrix.py
@@ -1,0 +1,161 @@
+# Natural Language Toolkit: XML attack matrix
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""XML entity attacks against ``nltk.xmlsec``, the parser the readers use.
+
+Several corpus readers parse XML that a user supplies, so the classic entity
+attacks all apply: file disclosure through an external entity, exponential
+expansion, and an external DTD that turns the parser into an SSRF client.
+
+Every case runs against both back ends: ``defusedxml`` when it is installed, and
+the standard-library fallback used when it is not.  The two must agree, because
+either may be the one protecting a given install, and only the fallback exercises
+the pathsec-guarded file read that :func:`nltk.xmlsec.parse` uses for a filename
+source.
+
+The external-DTD case is checked against a REAL listening socket rather than by
+reading the parsed text. A document whose DTD was fetched still parses to the
+same text, so text alone cannot tell "ignored" from "fetched and discarded", and
+the second of those is a live outbound request.
+"""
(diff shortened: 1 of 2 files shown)


alvations:
Route xmlsec through pathsec. Merging this in.

PR #3806, opened 2026-08-28
Validate the NE chunker save_params destination through the tool-directory guard

alvations:
## Vulnerability

`Maxent_NE_Chunker.save_params()` writes trained maxent parameter files into a caller-supplied `tab_dir`. The destination was checked with the plain containment guard (`validate_path`), and only after the in-memory classifier had already been read. Two gaps followed (GHSA-8mgp-746c-j5xp):

- An option-shaped (`-rf`), NUL-bearing, or `..`-component spelling was not refused syntactically. With enforcement relaxed the old guard warned ("allowed via CWD") and let the write through, landing parameter files outside the NLTK data sandbox.
- The check ran too late to be the first thing a caller-supplied path met, so attribute access on the classifier happened before the path was validated.

## Fix

Route `tab_dir` through `nltk.pathsec.validate_tool_dir` and move the check to the top of the method, before any classifier attribute is touched. `validate_tool_dir` applies the shared name refusals (blank, NUL, control character, option-shaped, URL, UNC, `..`, NTFS stream, drive-relative, Windows device) and then the same containment check, so a malformed or out-of-root destination is refused with `PermissionError` before a single parameter file is written, on every platform.

## Test

`nltk/test/unit/test_named_entity_pathsec.py` covers the delta:

- an outside `tab_dir` is refused and writes nothing;
- malformed spellings (`-rf`, NUL, `..`, blank) are refused even with enforcement off, which the old containment-only guard let through;
- the guard runs before the tagger is read (a missing tagger still yields `PermissionError`, not `AttributeError`);
- the default destination is a private, unpredictably-named directory rather than the historical guessable shared-temp path;
- the source still routes through `validate_tool_dir`.

Verified teeth: with `nltk/chunk/named_entity.py` reverted to `develop`, six of the eight tests fail (the malformed cases emit "allowed via CWD" and complete the write, the reorder test raises `AttributeError`, and the grep guard trips). All eight pass with the fix.

## Cross-platform

The tests build the chunker without its on-disk model (`object.__new__` plus a stub tagger), so no real `maxent_ne_chunker` download is needed. Nothing is staged under a shared temp dir; the outside target comes from the shared `sandbox` fixture (a fresh dir under `$HOME`). The device-name, drive-relative and NTFS-stream refusals are checked through `validate_tool_dir`'s platform-aware logic rather than any platform-specific filename, so the suite is green on Linux, macOS and Windows.

## Scope

Self-contained split from the larger GHSA-8mgp effort. `validate_tool_dir` and the pathsec sandbox fixtures are already on `develop`, so this branch touches only `nltk/chunk/named_entity.py` and adds one focused test module.

### DIFF
--- a/nltk/chunk/named_entity.py
+++ b/nltk/chunk/named_entity.py
@@ -14,7 +14,7 @@ import os
 import re
 
 from nltk.pathsec import open as pathsec_open
-from nltk.pathsec import validate_path
+from nltk.pathsec import validate_tool_dir
 from nltk.tag import ClassifierBasedTagger, pos_tag
 from nltk.xmlsec import parse as safe_parse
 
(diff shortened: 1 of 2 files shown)


alvations:
Route chunker model through pathsec validate_tool_dir. Merging this in.

PR #3807, opened 2026-08-28
Pin maxent scratch files to a data root and validate scratch IO

alvations:
## What

`nltk.classify.maxent`'s external-trainer wrappers staged their scratch files with a bare `tempfile.mkstemp` (no `dir=`). On Linux that places them in the shared, world-writable system temp, where the name is guessable and another local user can pre-create or symlink the path to read or redirect the write (CWE-377 / CWE-378):

- `train_maxent_classifier_with_megam` wrote the megam training file there.
- `TadmMaxentClassifier.train` wrote both the tadm events file and the weights file there, then read the weights back through the plain builtin `open`.

## Fix

Both wrappers now allocate a private (mode 0700) directory inside an allowed data root with `nltk.data.make_staging_dir` and pass it as `dir=` to `mkstemp`, so the scratch files land inside the pathsec sandbox on every platform (Linux included, where `/tmp` is deliberately not an allowed root). The megam write and the tadm weights read are routed through `nltk.pathsec.open`, which refuses a path outside the sandbox. The staging directory is removed after use so it does not leak per call.

## Dependencies

Self-contained on `develop`. It relies only on `make_staging_dir` and `pathsec.open`, both already present on `develop`. It deliberately does **not** use the process-wide `staging_tempdir` helper introduced by the companion `nltk.data` split, so this PR carries no cross-PR ordering dependency.

## Test

`nltk/test/unit/test_maxent_scratch_staging_security.py` exercises the delta without the real megam/tadm binaries by patching the external call out once the scratch file exists:

- the megam training file is created inside the enforced data root, not the system temp;
- both tadm scratch files are created inside the enforced data root;
- a scratch path forced outside the sandbox is refused by the pathsec-guarded write before the binary runs.

All three tests fail against the un-hardened code and pass with the fix. They reuse the existing `nltk/test/unit/conftest.py` sandbox fixtures, skip cleanly if numpy is absent, and avoid POSIX-only assumptions, so they run on Linux, macOS and Windows.

This is a self-contained split from the larger GHSA-8mgp-746c-j5xp hardening effort.

### DIFF
--- a/nltk/classify/maxent.py
+++ b/nltk/classify/maxent.py
@@ -58,6 +58,7 @@ except ImportError:
     pass
 
 import os
+import shutil
 import tempfile
 from collections import defaultdict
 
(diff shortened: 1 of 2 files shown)


alvations:
Route maxent opens through pathsec.

PR #3808, opened 2026-08-28
Add staging_tempdir and tighten data.py zip/staging guards

alvations:
This is a self-contained split from the larger GHSA-8mgp hardening effort: the changes to `nltk/data.py` plus a minimal set of tests for exactly the new behaviour, so it can be reviewed and merged on its own. Every helper it relies on (`pathsec.validate_path`, `pathsec.is_private_dir`, `make_staging_dir`) is already on `develop` (the aggregate zip-bomb helper and base decompression-bomb guards landed in #3797), and importing `nltk.data` and its dependents stays healthy.

## What changes in `nltk/data.py`

- **`staging_tempdir()`** (new): a process-wide 0700 scratch directory *inside* an allowed data root, allocated once per process and removed at exit. Wrappers that stage an input file for an external tool previously fell back to the system temp dir, which on Linux is the shared, world-writable `/tmp` and is deliberately not a pathsec root, so the scratch file landed outside the sandbox. Passing this as `dir=` keeps it inside a root. Backed by a module global `_STAGING_TEMPDIR`.
- **`_staging_dir_is_still_safe()`** (new): re-validates the cached scratch directory before handing it out again. `os.path.isdir` follows symlinks, so an attacker who removes the directory and drops a symlink in its place would pass a naive check and every later scratch file would land wherever the link points. This uses `lstat` (a symlink is rejected, not followed), re-verifies containment via `validate_path`, and requires a private (0700) directory, discarding a swapped, poisoned, or world/group-writable value and allocating a fresh one. Guards CWE-377 / CWE-378.
- **`make_staging_dir()`**: gains an opt-in `cleanup` flag (a *saved model* must survive interpreter exit; throwaway scratch should not), a defence-in-depth re-validation that the directory actually created is inside the root just validated, and rejection of a `prefix` that is a path rather than a plain filename fragment. Callers build the prefix from values such as a tagger's language code, and `mkdtemp` concatenates it onto `dir`, so a separator or NUL in the prefix could place the directory outside the data root (CWE-22).
- **`_assert_no_normalized_bypass()`** (new) wired into `find()`: refuses a resource name holding an ASCII tab, LF, CR, `#` or `?`. Python 3.14 made `url2pathname` follow the WHATWG URL rules, stripping those control characters and truncating at `#` / `?`. The stripping can turn a name with no `..` run (for example `".\n./x"`) into a traversal *after* the raw-form check has already passed, which then escapes the data root (CWE-22). Refusing the characters outright keeps the validated name and the used name identical on every Python version.
- **`find()` zip-split** gains `re.DOTALL` so a newline-bearing name cannot drive the `".zip/"` fallback into unbounded recursion (CWE-407 / CWE-1333). This sits behind the new character guard as defence in depth.

## Vulnerability classes

CWE-22 (path traversal via a control-character-normalised name and via a path-like staging prefix), CWE-377 / CWE-378 (insecure/squattable temporary directory reused from a poisoned or symlink-swapped cache), CWE-407 / CWE-1333 (unbounded recursion / ReDoS on a newline-bearing resource name).

## Tests

- `nltk/test/unit/test_staging_tempdir_security.py` (new): containment and privacy of the scratch dir, the symlink-swap and poisoned-cache regressions, an end-to-end scratch file that must not escape after a swap, a negative control showing the old `os.path.isdir` check reopens the bug, cleanup lifetime (opt-in vs default), and refusal of a writable or unwritable root.
- `nltk/test/unit/test_data_name_guards.py` (new): `find()` rejecting each `url2pathname`-rewritten character and the newline-created traversal, plus `make_staging_dir` refusing a path-like prefix and accepting a plain one.

## Cross-platform handling

The tests are green on Linux, macOS and Windows. Checks that are genuinely POSIX-only skip cleanly elsewhere:

- Symlink-swap tests (`test_cached_dir_swapped_for_a_symlink_is_discarded`, `test_a_real_scratch_file_cannot_escape_after_a_swap`, `test_teeth_isdir_check_alone_would_reopen_the_swap`) are gated `os.name != "posix"`, matching the existing `symlink squat is a POSIX concern` gating on `develop`.
- POSIX mode-bit tests (`test_a_writable_scratch_dir_is_not_reused` for world/group/other, and the `chmod 0o500` unwritable-root case) skip on non-POSIX, since Windows uses ACLs and `chmod` does not block directory writes there.
- The `make_staging_dir` prefix guards and the `find()` character guards are pure input validation and run on every platform.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -32,11 +32,14 @@ adds it to a resource cache; and ``retrieve()`` copies a given resource
 to a local file.
 """
 
+import atexit
 import codecs
 import functools
 import os
 import pickle
 import re
+import shutil
+import stat
 import sys
 import tempfile
 import textwrap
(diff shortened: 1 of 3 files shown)


alvations:
Routh tempdir stagging for any data downloads through pathsec. merging this in.

PR #3809, opened 2026-08-28
Stage boxer scratch input inside an nltk.data root

alvations:
## What

`Boxer` writes the C&C parser output to a scratch file and hands that file's
path to the external `boxer` binary. It created the file with
`tempfile.mkstemp()` and no `dir=`, so the scratch input landed in the shared
system temp directory rather than inside the NLTK data sandbox.

## Why it matters

The system temp directory is not an allowed `nltk.data` root. On Linux it is the
world-writable `/tmp`, so the boxer input file was staged outside the pathsec
sandbox and was exposed to a local squatting/replacement race between the moment
this process created it and the moment the external tool read it back
(CWE-377 / CWE-378, insecure temp file placement).

## Fix

`_call_boxer` now allocates a private, in-sandbox scratch directory with
`nltk.data.make_staging_dir(prefix="boxer-")` and stages the input file inside it
via `mkstemp(dir=...)`. The staging directory and its file are removed with
`shutil.rmtree` after the boxer call, so nothing leaks. `make_staging_dir` is
imported inside the method because `nltk.data`'s import graph reaches back into
`nltk.sem`, and a module-level import can otherwise observe a partially
initialised `nltk.data`.

The candc/boxer binary-path hardening (untrusted search path, CWE-426 / CWE-427)
already landed on `develop` and is unchanged here.

## Relationship to the larger effort

This is a self-contained split from the larger GHSA-8mgp hardening work. That
wider effort also introduces a cached, process-wide `nltk.data.staging_tempdir()`
helper, which ships in a separate `nltk.data` change. To keep this PR standalone
and importable on `develop` today, it deliberately uses the already-present
`make_staging_dir` primitive instead. There is no dependency on the in-flight
`data.py` split.

## Test

`nltk/test/unit/test_boxer_security.py` gains
`test_boxer_scratch_file_staged_in_data_root`. It fakes `Boxer._call`, so it
needs neither the `boxer` nor the `candc` binary, and asserts the staged scratch
path is contained inside an allowed `nltk.data` root (staged under `$HOME`). The
test has teeth: it fails against the unhardened wrapper because the scratch file
then lands in the system temp directory. It is cross-platform (no POSIX-only path
shapes, reads with `encoding="utf-8"`), so it runs cleanly on Linux, macOS, and
Windows. The four existing binary-path regression tests are unchanged.

Full boxer security suite: `5 passed`.

### DIFF
--- a/nltk/sem/boxer.py
+++ b/nltk/sem/boxer.py
@@ -31,6 +31,7 @@ For example::
 import operator
 import os
 import re
+import shutil
 import subprocess
 import tempfile
 from functools import reduce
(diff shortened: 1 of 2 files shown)


alvations:
Route boxer through pathsec. Merging this in.

PR #3810, opened 2026-08-28
Canonicalize obfuscated numeric IPv4 in the SSRF guard (CWE-918)

alvations:
## Vulnerability

`validate_network_url` is the SSRF guard for the downloader network path. It resolves the URL host and refuses the request when any resolved address is loopback, link-local, private, or otherwise non-global. Before this change it trusted the platform resolver to first fold obfuscated numeric IPv4 spellings to their real address:

- `http://2130706433/` (decimal)
- `http://0x7f000001/`, `http://0x7f.0.0.1/` (hex)
- `http://127.1/` (short form, last part absorbs the low bytes)
- `http://0177.0.0.1/` (leading-zero octal)

glibc and BSD `getaddrinfo`/`inet_aton` do fold these to 127.0.0.1, so the guard blocked them there. The Windows resolver does not: it returns nothing for these numeric literals, so the forbidden-IP loop saw no addresses to reject and the request reached the network. A loopback or cloud-metadata target disguised as a numeric literal therefore bypassed the SSRF guard on Windows (CWE-918).

## Fix

Add `_numeric_ipv4(host)`, a classic `inet_aton` parse: 1 to 4 dot-separated parts, each decimal, `0x` hex, or leading-zero octal, with the final part absorbing the remaining low-order bytes. `validate_network_url` now canonicalizes the host itself and, before the `getaddrinfo` loop, refuses the request when the parsed numeric IPv4 is forbidden. The verdict no longer depends on whether the platform resolver folds the spelling, so loopback/link-local/private numeric forms are refused on every platform. Octal loopback such as `0177.0.0.1` is refused unconditionally. Public numeric forms (e.g. `0x08080808` = 8.8.8.8) stay allowed. Hostnames return `None` from the parser and fall through to normal resolution. The helper uses only the `ipaddress` module already imported in `pathsec.py`; no new dependency.

## Why the proxy probe and proxied-fetch tests were updated

`_numeric_ipv4` is a second, independent direct-IP guard alongside `_resolve_hostname`. The proxy-SSRF probe (`security_probes/ghsa_6ww7_3frv_cqxh.py`) and the proxied-fetch tests in `test_pathsec.py` deliberately neuter `_resolve_hostname` so that the proxy-specific guard is what gets exercised rather than the direct-IP resolution path. With the new canonicalizer in place, the link-local literal `169.254.169.254` would now be refused by `_numeric_ipv4` first, hiding whether the proxy guard fires; without neutering it the probe reports VULNERABLE against the new guard. Those tests now also neuter `_numeric_ipv4` in their save/patch/restore so the proxy path remains the thing under test.

## Tests

- `nltk/test/unit/test_ssrf_url_encodings.py` — refusal cases for the obfuscated numeric spellings; a renamed test asserting `0177.0.0.1` is always refused; a regression test that pins `_resolve_hostname` to empty (the Windows non-folding behavior) and proves the guard blocks every numeric spelling on its own; and an over-block control keeping a public numeric form allowed.
- `nltk/test/unit/security_probes/ghsa_6ww7_3frv_cqxh.py` — neuters `_numeric_ipv4` alongside `_resolve_hostname` so the proxy guard stays under test.
- `nltk/test/unit/test_pathsec.py` — the four proxied-fetch tests that neuter `_resolve_hostname` now also neuter `_numeric_ipv4`.

Verified locally (Python 3.13): the full `test_ssrf_url_encodings.py`, `test_pathsec.py`, and `test_advisory_probes.py` suites pass, and a direct comparison against the current `develop` `pathsec.py` under a stubbed empty resolver shows the pre-patch guard allows `http://2130706433/` while the patched guard blocks it.

## Scope

Self-contained split from the larger GHSA-8mgp effort: this PR is only the SSRF numeric-IPv4 canonicalization guard and the tests that prove it.

### DIFF
--- a/nltk/pathsec.py
+++ b/nltk/pathsec.py
@@ -887,6 +887,51 @@ def _ip_is_forbidden(ip):
     return ip.is_multicast or not ip.is_global
 
 
+def _numeric_ipv4(host):
+    """Canonical ``inet_aton`` parse of a numeric IPv4 literal, or None.
+
+    Follows the classic ``inet_aton`` rules: 1 to 4 dot parts, each decimal, a
+    ``0x`` hex value, or a leading-zero octal value, with the final part
+    absorbing the remaining low-order bytes. glibc and BSD fold these obfuscated
+    forms (``2130706433``, ``0x7f000001``, ``127.1``, ``0177.0.0.1``) to their
+    real address at resolution time, but the Windows resolver does not, so the
+    SSRF guard must canonicalize them itself to refuse loopback on every
+    platform (CWE-918).
+    """
+    if not host:
+        return None
+    parts = host.split(".")
+    if not 1 <= len(parts) <= 4:
+        return None
+    values = []
+    for part in parts:
+        if not part:
+            return None
+        try:
+            if part[:2].lower() == "0x":
+                value = int(part, 16)
+            elif part[0] == "0" and len(part) > 1:
+                value = int(part, 8)
+            else:
+                value = int(part, 10)
+        except ValueError:
+            return None
+        values.append(value)
+    for leading in values[:-1]:
+        if not 0 <= leading <= 0xFF:
+            return None
+    last = values[-1]
+    if not 0 <= last <= (1 << (8 * (5 - len(values)))) - 1:
+        return None
+    packed = last
+    for index, leading in enumerate(values[:-1]):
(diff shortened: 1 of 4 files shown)


alvations:
Added even more ssrf guards to pathsec. Merging this in first otherwise windows users are vulnerable.

PR #3812, opened 2026-08-28
Harden stanford tokenizer/segmenter wrappers: validate jar/model paths and argv

alvations:
The Stanford tokenizer and segmenter wrappers hand caller-supplied model, dictionary, corpus and input-file paths straight into a child JVM's argv (`-loadClassifier` / `-serDictionary` / `-sighanCorporaDict` / `-textFile`), which `pathsec.open` cannot wrap. Left unbounded, those are arbitrary-file-read primitives (GHSA-8mgp-746c-j5xp). This is one file-scoped split out of a larger hardening branch, covering only the two `nltk/tokenize` Stanford wrappers.

## Vulnerability classes and fixes

**`nltk/tokenize/stanford_segmenter.py`**
- Model / dictionary / Sihan-dir path traversal: `-loadClassifier`, `-serDictionary` and `-sighanCorporaDict` reached the JVM unbounded. New `_validate_model_paths()` bounds the model and dictionary files with `validate_tool_path` and the Sihan corpora directory with `validate_tool_dir` before the process is spawned.
- Input-file path traversal: `segment_file` validates its `-textFile` argument with `validate_tool_path`.
- `__fspath__` TOCTOU: each guarded attribute is replaced with, and the argv is built from, the string the guard returned, never the original object. A `PathLike` whose `__fspath__` answers differently on a later call can no longer have one file checked and another handed to the JVM.
- Argv injection through `-options`: the options dict is joined into one comma-separated `-options` element, so an option name holding a `,`/`=`/whitespace/NUL separator would inject extra option pairs. New `_validated_options()` refuses such names and bounds any value that names a file.
- Temp input file is staged under a data root via `staging_tempdir()` instead of the shared system temp dir.

**`nltk/tokenize/stanford.py`**
- Temp input file is staged under a data root via `staging_tempdir()` instead of the shared system temp dir.

## Tests

New `nltk/test/unit/test_tokenize_stanford_pathsec.py` drives the real code path with only the `java` hand-off trapped, so a probe cannot pass merely because the test assembled the argv itself. It proves an out-of-sandbox model, dictionary, Sihan dir, input file or option value is refused; a mutating `__fspath__` is frozen to the validated string; an option name carrying a separator is rejected; benign in-root paths still reach the JVM; and the tokenizer's staged input lands inside a data root. Every guard assertion runs without the real Stanford jars and each fails against the un-hardened code, so the tests have teeth.

## Cross-platform

`pathsec` containment is OS-agnostic and `java` is never launched, so the suite is green on Linux, macOS and Windows with no external tool. It reuses the existing `pathsec_sandbox` conftest fixture (which folds Windows 8.3 short names and treats a private per-user temp dir correctly), stages fixtures under `$HOME`/the data root rather than `/tmp`, and reads text as UTF-8.

## Self-contained split

Depends only on symbols already on `develop`: `validate_tool_path` / `validate_tool_dir` from #3797 and `staging_tempdir` from #3808. `import nltk.tokenize.stanford` and `import nltk.tokenize.stanford_segmenter` both succeed on `develop`.

SSRF is not applicable to these wrappers: they invoke a local Java jar, they never fetch a URL.

### DIFF
--- a/nltk/test/unit/test_tokenize_stanford_pathsec.py
+++ b/nltk/test/unit/test_tokenize_stanford_pathsec.py
@@ -1,0 +1,255 @@
+# Natural Language Toolkit: pathsec guards for the Stanford tokenizer wrappers
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""Security guards for ``nltk.tokenize.stanford`` and
+``nltk.tokenize.stanford_segmenter``.
+
+Both wrappers hand caller-supplied model, dictionary, corpus and input-file
+paths straight into a child JVM's argv (``-loadClassifier`` /
+``-serDictionary`` / ``-sighanCorporaDict`` / ``-textFile``), which
+``pathsec.open`` cannot wrap. An unbounded value there is an arbitrary file read
+(GHSA-8mgp-746c-j5xp). These tests drive the real code path with only the
+``java`` hand-off replaced, so a probe cannot pass merely because the test
+assembled the argv itself. Each guard assertion runs WITHOUT the real Stanford
+jars: the crafted attack is refused before the JVM is reached, and the benign
+control stops at a trapped ``java`` rather than launching one.
+
+SSRF is not in scope here: these wrappers invoke a local Java jar, they never
+fetch a URL.
+"""
+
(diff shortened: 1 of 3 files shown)


alvations:
Route all stanford opening through pathsec staging_tempdir. Merging this in.

PR #3813, opened 2026-08-28
Harden tagger model loads and tool paths (perceptron/crf/hunpos/stanford)

alvations:
## What

Hardens the caller-controlled model/param paths and scratch-file staging in the
tagger family under GHSA-8mgp-746c-j5xp. Each of these taggers hands a
caller-supplied path to a native extension or an external process that
`pathsec.open` cannot wrap, or stages a temporary file in the shared system temp
dir, so a caller (or a poisoned `HUNPOS_TAGGER` / `STANFORD_MODELS` environment
variable) could turn a model path into an arbitrary-file read/write.

## Files changed

- `nltk/tag/perceptron.py` - `AveragedPerceptron.save`/`load` validate the path
  and route through `pathsec.open` instead of builtin `open`; the trained-model
  directory is now allocated lazily by `nltk.data.make_staging_dir` inside an
  allowed data root (mode 0700, unpredictable name) instead of the guessable
  `<tempdir>/averaged_perceptron_tagger_<lang>`; the sandbox-widening helper that
  appended a caller directory to `nltk.data.path` is removed; `lang` and
  `TAGGER_NAME` are validated as single filename components at the `param_files`
  chokepoint; the pinned model-dir descriptor is re-validated against its real
  landed path.
- `nltk/tag/crf.py` - `CRFTagger.set_model_file` and `CRFTagger.train` validate
  the caller path before the pycrfsuite C extension opens/writes it.
- `nltk/tag/hunpos.py` - `HunposTagger` validates the resolved model path before
  it becomes `hunpos-tag` subprocess argv; the token-newline check is a real
  `ValueError` rather than a bare `assert` that `python -O` strips.
- `nltk/tag/stanford.py` - `StanfordTagger` validates the resolved model path
  (in `__init__` and again before the JVM hand-off) and stages its temporary
  input file under `nltk.data.staging_tempdir` rather than the shared temp dir.

## Vulnerability classes closed

- Native/JSON model-load into an arbitrary path (arbitrary-file read/write).
- Model/param path traversal, including filename-component traversal via
  `lang` / `TAGGER_NAME`.
- CWD-relative and out-of-sandbox tool paths (also reachable via env vars).
- Scratch-file placement in a shared, world-writable temp dir.

SSRF is not applicable: these taggers load local model files and spawn local
binaries; there is no URL fetching in the changed code paths.

## Tests

`nltk/test/unit/test_pathsec_sweep_tag.py` drives every guarded sink with a full
escape matrix (traversal, symlink/hardlink/FIFO/socket, URL-shaped,
unicode-confusable, seeded fuzz) plus benign over-block controls that assert
ordinary language codes and odd-but-contained filenames still round-trip. It
also includes load-bearing negative controls that neuter each guard and prove
the escape becomes reachable again, so a future refactor that drops a check is
caught.

Cross-platform: POSIX-only filesystem shapes (symlinks, FIFO/socket, `/dev`,
`/proc`, the pinned directory descriptor) are `skipif`/`skip`ped off POSIX, and
Windows device-name and trailing-space cases are asserted where they apply. No
real Stanford jar, senna binary or pycrfsuite model is required: JVM/binary
sinks are driven with `object.__new__` stubs and `importorskip`, and outside-root
fixtures stage under `$HOME`, never `/tmp`.

## Self-contained split

This is one slice of a larger hardening effort, split so it reviews on its own.
It depends only on primitives already on `develop` (`nltk.pathsec`
validate/open, `nltk.data` staging, `nltk.internals` find_binary CWD refusal);
the maxent chunker and punkt are referenced only as shared-primitive
cross-checks that already exist on `develop`. The maxent model-artifact cases
are deliberately left to their own split and are not duplicated here. All 751
tests pass on `origin/develop` plus these four tagger files alone.

### DIFF
--- a/nltk/tag/crf.py
+++ b/nltk/tag/crf.py
@@ -14,6 +14,7 @@ import re
 import unicodedata
 import warnings
 
+from nltk.pathsec import validate_tool_path
 from nltk.tag.api import TaggerI
 
 try:
@@ -121,6 +122,16 @@ class CRFTagger(TaggerI):
         self._feature_cache = {}
 
     def set_model_file(self, model_file):
+        """Load a trained model from ``model_file``.
+
+        ``model_file`` is caller-supplied and pycrfsuite opens it natively, so
+        the path is validated against the NLTK data sandbox first: an
+        outside-root model is refused rather than handed to the native loader
+        (GHSA-8mgp-746c-j5xp). ``pathsec.open`` cannot be used here because the
+        C extension does its own open, so containment is the check that can be
+        applied; a symlink swapped in after it is not covered.
+        """
+        validate_tool_path(model_file, context="CRFTagger.set_model_file")
         self._model_file = model_file
         self._tagger.open(self._model_file)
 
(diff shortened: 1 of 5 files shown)


alvations:
Route all tag tools open/loading/tempdir through pathsec. Merging this in.

PR #3814, opened 2026-08-28
Stage weka scratch files in a data root and validate the WEKAHOME jar path

alvations:
The Weka wrapper (`nltk/classify/weka.py`) drives an external `weka.jar` through a subprocess. Two path-handling issues are fixed here.

## Vulnerability classes

**Scratch ARFF on the shared system temp dir (CWE-377).** `WekaClassifier.train()` and `_classify_many()` wrote their `train.arff` / `test.arff` to `tempfile.mkdtemp()`, which lands in the world-writable system temp dir on Linux and is deliberately *not* an NLTK data root. A local user could observe or race the scratch file another process is about to feed to weka.

**Unbounded `WEKAHOME` search path (CWE-426 / CWE-427).** `config_weka()` inserted `os.environ["WEKAHOME"]` at the front of the jar search path with no validation, so anyone able to set that environment variable could point NLTK at a `weka.jar` outside the trusted roots and have it loaded and executed.

## The fix

- Both scratch files now stage under `nltk.data.make_staging_dir(prefix="nltk_weka_run_", cleanup=True)` — a private `0700` directory inside an allowed data root, removed at interpreter exit.
- `WEKAHOME` is validated with `nltk.pathsec.validate_path` before it is trusted: an out-of-root value is ignored with a `UserWarning`, while a legitimate in-root install still resolves.

## Tests

- `nltk/test/unit/test_weka_security.py` gains three cases: an out-of-root `WEKAHOME` is refused (warning + `LookupError`), an in-root `WEKAHOME` still resolves, and `train()`s scratch ARFF is proven to land inside the data root rather than the system temp dir. The refusal and placement assertions fail against the un-hardened code; the external `java` call is mocked, so none of these need a JVM or a real `weka.jar`.
- `nltk/test/unit/test_weka_integration.py` is a new end-to-end round trip that runs only when a real java runtime and `weka.jar` are present, and skips cleanly otherwise.

## Cross-platform

The refusal/placement tests use the shared `pathsec_sandbox` / `restricted_sandbox` conftest fixtures (a single enforced data root plus an out-of-root dir under `$HOME`) instead of a hard-coded `/tmp` path, inside-root checks use `os.path.commonpath` on `realpath`ed paths, and the JVM-dependent tests skip when no `java` is on `PATH`. The suite is green on Linux, macOS and Windows.

## Notes

This change is split out of a larger hardening effort so it stands alone: it applies cleanly on top of `develop` (all guard symbols it uses — `make_staging_dir`, `validate_path` — are already on `develop`) and its tests pass against `develop` plus this `weka.py` alone.

SSRF is not applicable to this wrapper: it only invokes a local `weka.jar` and never fetches a URL.

### DIFF
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -14,6 +14,7 @@ import re
 import subprocess
 import tempfile
 import time
+import warnings
 from sys import stdin
 
 from nltk.classify.api import ClassifierI
@@ -21,6 +22,7 @@ from nltk.data import make_staging_dir
 from nltk.internals import config_java, java
 from nltk.pathsec import ZipFile as SecureZipFile
 from nltk.pathsec import open as pathsec_open
+from nltk.pathsec import validate_path
 from nltk.probability import DictionaryProbDist
 
 _weka_classpath = None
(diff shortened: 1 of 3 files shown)


alvations:
Route weka loading through pathsec. Merging in.

PR #3815, opened 2026-08-28
Route corpus-reader scratch tempfiles through a data root and label framenet path refusals

alvations:
## What

A self-contained split from the larger GHSA-8mgp file-IO hardening effort, covering only the three corpus readers that touch scratch tempfiles or path-containment messages.

### Readers changed

- **`nltk/corpus/reader/nkjp.py`** (`XML_Tool.__init__`): the preprocessing scratch file, previously `tempfile.NamedTemporaryFile(delete=False)`, is now pinned with `dir=staging_tempdir()`.
- **`nltk/corpus/reader/timit.py`** (`TimitCorpusReader.wav`): the wave scratch file, previously `tempfile.TemporaryFile()`, is now pinned with `dir=staging_tempdir()`.
- **`nltk/corpus/reader/framenet.py`** (`_reject_unsafe_path_component`): the refusal message is prefixed with `Security violation:` so a containment decision reads differently from an incidental lookup miss. Behaviour is otherwise unchanged; the resolving choke point `_validate_in_root` is untouched.

### Vulnerability classes

- **Unpinned scratch tempfiles on shared `/tmp` (CWE-377 / CWE-378).** `tempfile.NamedTemporaryFile()` / `TemporaryFile()` default to the system temp dir. On Linux that is the shared, world-writable `/tmp`, which is deliberately *not* a pathsec data root, so NKJP and TIMIT staged scratch data outside the sandbox where another local user could observe or race it. `nltk.data.staging_tempdir()` (already on `develop`) returns a private directory inside an allowed data root on every platform; passing it as `dir=` moves the scratch file back inside the sandbox.
- **Path-containment message labelling.** Framenet frame/lu/doc loads funnel caller- and corpus-supplied names through `_reject_unsafe_path_component`; the refusal is now explicitly labelled so callers and audits can tell a security refusal apart from a plain lookup error.

## Tests

Adds `nltk/test/unit/test_corpus_reader_hardening.py`, scoped to exactly this delta:

- framenet refusal is labelled `Security violation` (across separator, `..`, and Windows drive/UNC names), plus over-block controls: a plain frame name is accepted, and `_validate_in_root` still refuses a traversal / absolute / symlink escape while accepting an in-root path.
- the NKJP and TIMIT scratch tempfiles land inside a registered data root (equal to `staging_tempdir()`), verified behaviourally.

Each tempfile-pinning and message assertion fails against the current un-hardened `develop` readers (verified by reverting each file), so the tests have teeth.

## Cross-platform handling

The tests use the shared `conftest.py` sandbox fixtures (`restricted_sandbox`, `pathsec_sandbox`), which enforce pathsec against one throwaway data root and register it on `nltk.data.path`. That registration is what lets a `CorpusReader` construct against a fixture root and lets `staging_tempdir()` allocate inside it on Linux and Windows, where a bare temp dir is not a data root, as well as on macOS. Fixture files are read `encoding="utf-8"` and audio is written binary; the symlink-escape check is skipped off POSIX.

This split is self-contained on `develop` (it depends only on `nltk.data.staging_tempdir` and `nltk.pathsec`, both already merged) and lets the deferred temp-file-factory guard be re-enabled later.

### DIFF
--- a/nltk/corpus/reader/framenet.py
+++ b/nltk/corpus/reader/framenet.py
@@ -776,7 +776,9 @@ def _reject_unsafe_path_component(value, kind):
         or ".." in value
         or ntpath.splitdrive(value)[0]
     ):
-        raise FramenetError(f"Invalid {kind}: {value!r}")
+        # "Security violation" marks this as a containment decision, not an
+        # incidental lookup error, so callers and audits can tell them apart.
+        raise FramenetError(f"Security violation: Invalid {kind}: {value!r}")
 
 
 def _validate_in_root(locpath, root, context):
--- a/nltk/corpus/reader/nkjp.py
+++ b/nltk/corpus/reader/nkjp.py
@@ -277,7 +277,12 @@ class XML_Tool:
     def __init__(self, root, filename):
         self._root = root
         self.read_file = os.path.join(root, filename)
-        self.write_file = tempfile.NamedTemporaryFile(delete=False)
+        # Imported here: nltk.data imports the corpus package.
+        from nltk.data import staging_tempdir
+
+        self.write_file = tempfile.NamedTemporaryFile(
+            delete=False, dir=staging_tempdir()
+        )
 
     def build_preprocessed_file(self):
         try:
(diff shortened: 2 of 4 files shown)


alvations:
Route corpora reader with custom tempdir through pathsec. Merging this in.

PR #3816, opened 2026-08-28
Harden malt and stanford parser wrappers: validate binary/model/IO paths and argv

alvations:
Both JVM parser wrappers hand caller-supplied model, corpus and IO paths straight to a child process's argv without bounding them to the NLTK data roots. This splits the malt/stanford portion of the wrapper hardening into a standalone change against `develop`; every guard symbol it relies on (`nltk.pathsec.validate_tool_path`/`validate_model_resource`/`validate_path`, `nltk.data.make_staging_dir`/`staging_tempdir`) is already on `develop`.

## Vulnerability classes

**nltk/parse/malt.py**
- **Arbitrary read/write via `-i`/`-o`.** `generate_malt_command`'s `-i` is read by the JVM and `-o` is written by it. Both are public entry points (`generate_malt_command`, `train_from_file`), so an unbounded value is an arbitrary file read and an arbitrary file write.
- **Model path traversal (`-c`/`-w`).** The trained-model path was used unchecked; in learn mode MaltParser writes the `.mco` into the model's directory.
- **World-writable working dir.** `working_dir` defaulted to the shared `tempfile.gettempdir()`, a predictable, squattable location on Linux for the CoNLL intermediates and, for an untrained parser, `malt_temp.mco`.
- **`__fspath__` TOCTOU / lying `str` subclass.** A `PathLike` whose `__fspath__` mutates between the check and the JVM call, or a `str` subclass whose inspection methods lie, could smuggle a different value into argv.

**nltk/parse/stanford.py**
- **`-model` traversal / TOCTOU.** `self.model_path` reached the JVM unchecked and could mutate between check and use.
- **Argument injection via `corenlp_options`.** The string is split into separate argv elements, so it could append a second `-model`; `LexicalizedParser` honours the last occurrence (verified against a real JVM), silently replacing the model the guard had just checked.

## The fix
- malt: `-i`/`-o` bounded with `validate_tool_path` (`-o` with `for_write=True`); model bounded with `validate_model_resource` and, when it carries a directory, `validate_path`; `working_dir` is now a lazily allocated private dir via `nltk.data.make_staging_dir(cleanup=True)` (mode 0700, inside a data root, removed at exit) with a setter that refuses empty/out-of-root values; the model string is frozen once and coerced to a real `str`; `parse_tagged_sents` reads output through `pathsec.open`.
- stanford: `_execute` validates `model_path` at the single JVM hand-off and replaces the `-model` argv entry with the resolved string (covers all three callers); `_validated_extra_options` refuses re-setting an option NLTK already set and bounds any path-like `corenlp_options` value to the roots; the JVM input file is staged in `nltk.data.staging_tempdir()`.

## Tests
New focused module `nltk/test/unit/test_malt_stanford_pathsec.py` drives the real wrappers with only the `java` hand-off trapped, so a probe cannot pass because the test assembled the argv itself. It proves crafted `-i`/`-o`/model/`corenlp_options` values are refused, while benign in-root paths and jar-internal resource names still reach the JVM; the teeth tests neuter each guard and assert the attack reappears, so every refusal is known to fail against the un-hardened code. It also includes the reclaimed malt working-dir cleanup regression (`test_malt_working_dir_is_cleaned_up`, kept with the malt wrapper rather than in the data.py staging test file) and real end-to-end JVM runs that skip cleanly when the external tool is absent.

## Cross-platform
POSIX-only checks (mode bits) and JVM/tool-dependent runs are gated with `skipif`, and Windows-specific payloads are only used on Windows, so the module is green on Linux, macOS and Windows.

## Self-contained split note
The security guards this depends on are already on `develop`. The two source files apply cleanly with no other changes; one pair of dead imports (`atexit`, `shutil`, now unused because the cleanup registration lives in `nltk.data.make_staging_dir`) was dropped to satisfy the unused-import check.

### DIFF
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -12,7 +12,7 @@ import os
 import sys
 import tempfile
 
-from nltk.data import ZipFilePathPointer
+from nltk.data import ZipFilePathPointer, make_staging_dir
 from nltk.internals import (
     find_dir,
     find_file,
@@ -22,6 +22,12 @@ from nltk.internals import (
 from nltk.parse.api import ParserI
 from nltk.parse.dependencygraph import DependencyGraph
 from nltk.parse.util import taggedsents_to_conll
+from nltk.pathsec import open as pathsec_open
+from nltk.pathsec import (
+    validate_model_resource,
+    validate_path,
+    validate_tool_path,
+)
 
 
 def malt_regex_tagger():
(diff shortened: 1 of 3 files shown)


alvations:
Route malt JVM through more pathsec checks. Merging this in.

PR #3817, opened 2026-08-28
Harden nltk.tbl.demo model load/save through picklesec and pathsec

alvations:
## Summary

`nltk.tbl.demo.postag()` serializes a trained Brill/baseline tagger, writes an error-analysis text file, and plots a learning curve, each to a **caller-supplied** path. On `develop` those sinks used bare `open()`, so a caller-controlled destination could be written anywhere on disk (path traversal / arbitrary write, GHSA-8mgp-746c-j5xp), and the reloaded pickles are the RCE surface that `picklesec` already guards.

This routes every caller-controlled sink in the module through the `nltk.pathsec` sandbox.

## Vulnerability classes

- **Path traversal / arbitrary write** on `cache_baseline_tagger`, `serialize_output`, `error_output`, and `learning_curve_output` (all caller-supplied).
- **Pickle RCE on model load** for the two tagger pickles; loads already go through `nltk.picklesec.pickle_load` on `develop`, and the writes/reads are now also confined to the data sandbox.
- **Scratch placement**: model/plot/error artifacts must land inside an allowed NLTK data root, not at an attacker-named location.

## The fix (`nltk/tbl/demo.py`)

- `from nltk.pathsec import open as pathsec_open` and `validate_path`.
- `cache_baseline_tagger` pickle dump/load, `serialize_output` pickle dump/load, and the `error_output` text write now use `pathsec_open(..., context="tbl.demo.*")`.
- The `error_output` change also fixes a pre-existing bug: the payload is a `str` written in text mode, so the stray `.encode("utf-8")` (which raised `TypeError: can't concat str to bytes`) is dropped and `encoding="utf-8"` is set on the sandboxed open.
- `_demo_plot()` validates `learning_curve_output` with `validate_path()` up front, because matplotlib writes the plot itself and `pathsec.open()` cannot wrap a `savefig()`.
- Pickle loads continue through `nltk.picklesec.pickle_load` (already on `develop`).

## Tests (`nltk/test/unit/test_pathsec_sweep_tbl.py`)

- **Refusal teeth**, corpus-free (a tiny synthetic tagged corpus, so no downloaded model/corpus is required): each of the four sinks refuses an out-of-root destination with `PermissionError` and leaves nothing behind. Each of these fails against the un-hardened `demo.py` (the pickle sinks write outside then error; `error_output` hits the old `str`+`bytes` bug; `_demo_plot` does not raise at all), proving the guard.
- **In-root acceptance teeth**, corpus-free: the `pathsec.open` text sink and the `validate_path` plot sink both accept an in-root destination and write it.
- **Corpus-backed pins**: escape and malformed-path (`../../../`, NUL byte, `-Xmx99g`, empty) cases that skip cleanly when the treebank corpus / numpy is unavailable.

## Cross-platform

Green on Linux, macOS, and Windows. Out-of-root targets are staged under `$HOME` (never `/tmp`, which is a valid root on some CI images), pickles use `"rb"`/`"wb"`, text uses `encoding="utf-8"`, and the matplotlib acceptance test `importorskip`s matplotlib. No POSIX-only assertions are used, so nothing is skipped spuriously.

## Self-contained split

This is split out to stand alone on `develop`: it depends only on symbols already present there (`nltk.picklesec.pickle_load`/`RestrictedUnpickler` #3794, `nltk.pathsec.open`/`validate_path`/`validate_model_resource` #3797, `nltk.data.staging_tempdir`/`make_staging_dir` #3808) and the shared `nltk/test/unit/conftest.py` sandbox fixtures. `python3.13 -m pytest nltk/test/unit/test_pathsec_sweep_tbl.py` is 14 passed on `origin/develop` + this `demo.py` alone.

## No duplication

The tbl cases in the shared `test_model_artifact_pathsec.py` are shipped here as a focused, tbl-only module instead. The generic `pathsec.open` negative control and the maxent (#3807) and perceptron/crf tagger (#3813, `test_pathsec_sweep_tag.py`) cases already on `develop` are **not** re-shipped; every test function name in this file was checked against `develop` and none collides.

### DIFF
--- a/nltk/tbl/demo.py
+++ b/nltk/tbl/demo.py
@@ -13,6 +13,8 @@ import random
 import time
 
 from nltk.corpus import treebank
+from nltk.pathsec import open as pathsec_open
+from nltk.pathsec import validate_path
 from nltk.picklesec import pickle_load
 from nltk.tag import BrillTaggerTrainer, RegexpTagger, UnigramTagger
 from nltk.tag.brill import Pos, Word
@@ -246,14 +248,20 @@ def postag(
             baseline_tagger = UnigramTagger(
                 baseline_data, backoff=baseline_backoff_tagger
             )
-            with open(cache_baseline_tagger, "wb") as print_rules:
+            # ``cache_baseline_tagger`` is caller-supplied, so the model
+            # write/read goes through the pathsec sandbox (GHSA-8mgp-746c-j5xp).
+            with pathsec_open(
+                cache_baseline_tagger, "wb", context="tbl.demo.cache_baseline_tagger"
+            ) as print_rules:
                 pickle.dump(baseline_tagger, print_rules)
             print(
                 "Trained baseline tagger, pickled it to {}".format(
                     cache_baseline_tagger
                 )
             )
-        with open(cache_baseline_tagger, "rb") as print_rules:
+        with pathsec_open(
+            cache_baseline_tagger, "rb", context="tbl.demo.cache_baseline_tagger"
+        ) as print_rules:
             baseline_tagger = pickle_load(print_rules)
             print(f"Reloaded pickled tagger from {cache_baseline_tagger}")
     else:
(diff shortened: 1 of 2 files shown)


alvations:
Route tbl through pathsec and picklesec. Merging this in.

PR #3818, opened 2026-08-28
Add tests covering already-merged pathsec/security guards

alvations:
## Test-only additions for hardening already on develop

These are test-only changes. Every guard they exercise is already merged
on `develop`; each test file was run unchanged against `develop` (no
source/module change) and passes, so this only adds coverage for
hardening that shipped without a focused test.

### Added, grouped by the guard they cover

- **Zip member platform rules** (`test_zip_member_platform_rules.py`) --
  `pathsec._zip_member_is_unsafe` under both POSIX and Windows separator
  rules. A `..\..\x` / `C:\x` member is unsafe on Windows but a legal
  filename on POSIX; the test swaps in each platform's path module so a
  Linux CI run also checks the Windows verdict, and adds an end-to-end
  extraction refusal plus an over-block control. The predicate had no
  test on develop.
- **Zip member collision** (`test_zip_member_collision.py`) --
  `pathsec._reject_colliding_members` refuses archives whose members
  collide on a case-folding or unicode-normalizing filesystem (case,
  NFC/NFD, ligature, case-differing directory), with a control that
  distinct members still extract.
- **Aggregate decompression bomb** (`test_aggregate_zip_bomb.py`) --
  `nltk.data._check_zip_total_size` refuses an archive whose members each
  pass the per-member floor but sum to a bomb, with controls for a
  legitimately large corpus and a below-threshold ratio. The aggregate
  guard had no test on develop.
- **Zip extraction symlink TOCTOU** (`test_zip_extraction_toctou.py`) --
  `pathsec.ZipFile` extraction refuses a symlinked-parent leaf so a member
  write cannot escape through it.
- **Enforce does not break normal use**
  (`test_enforce_does_not_break_normal_use.py`) -- with `pathsec.ENFORCE`
  on by default, real tokenizers/taggers/corpora still load and a missing
  resource surfaces as `LookupError`, not `PermissionError`.
- **Module import + functional smoke**
  (`test_module_import_and_functional_smoke.py`) -- every `nltk` module
  imports and the common pipelines run under the hardened defaults.
- **chat80 shelve-backed valuation sinks** (`test_chat80_pathsec.py`) --
  two added tests: `val_dump`/`val_load` validate the caller path before
  `shelve.open` creates its backing files (outside, `/etc`, NUL are
  refused; an in-root destination passes). The five existing tests in
  that file are unchanged.

### Deliberately excluded to avoid duplication

- The monster sweep modules (`test_pathsec_sweep_wrappers.py`,
  `test_model_artifact_pathsec.py`, `test_pathsec_sweep_tbl.py`) -- their
  assertions were already lifted into focused modules on develop.
- `test_stanford_arg_injection.py` -- every guard it exercises (per-call
  java option allowlist, `_java_child_env()` env stripping, `cmd[0]`/
  `@argfile` rejection, bounded model path) is already covered by
  `test_java_per_call_options_security.py` and `test_malt_stanford_pathsec.py`.
- `test_corpus_reader_traversal.py` -- its Framenet `_validate_in_root`
  choke-point test duplicates `test_corpus_reader_hardening.py`, and
  reader-root / symlink containment is covered by
  `test_corpus_reader_pathsec.py`.
- `test_archive_extraction_hardening.py` -- its high-entry-count and
  gzip-pointer bomb tests duplicate `test_zipbomb_security.py`.

### Deferred (would need an unsplit source change)

- `test_file_io_guard_coverage.py` -- asserts the `tools/check_no_unsandboxed_open.py`
  guard scans the whole package rather than a shortlist; that broader
  guard is not on develop, so the file fails there and does not belong in
  a test-only PR.

Non-vacuousness was verified by neutering each new guard and confirming
the corresponding refusal tests fail (collision, platform-rule and
aggregate-bomb tests all fail with the guard removed), while the
over-block controls keep passing.

### DIFF
--- a/nltk/test/unit/test_aggregate_zip_bomb.py
+++ b/nltk/test/unit/test_aggregate_zip_bomb.py
@@ -1,0 +1,91 @@
+# Natural Language Toolkit: aggregate decompression-bomb guard
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""A zip whose members SUM to a bomb must be refused, not only per-member.
+
+_check_decompression_bomb runs per member and only ratio-checks a member above
+32 MiB (MAX_UNZIP_ACTIVATION). So an archive of many members each just under the
+floor is never individually refused, yet the members together expand from a tiny
+zip to an arbitrarily large total, and extraction kept no running byte total.
+_check_zip_total_size applies the same activation-plus-ratio test at the whole
+archive level.
+"""
+
+import os
+import shutil
+import tempfile
+import zipfile
+
+import pytest
+
+import nltk.data
+from nltk import pathsec
+
+
+@pytest.fixture
+def sandbox_root(monkeypatch):
+    root = tempfile.mkdtemp(prefix="nltk_sandbox_root_")
+    monkeypatch.setattr(pathsec, "ENFORCE", True)
+    monkeypatch.setattr(nltk.data, "path", [root])
+    monkeypatch.setattr(pathsec, "_ALLOWED_ROOTS_CACHE", None, raising=False)
+    monkeypatch.setattr(pathsec, "_LAST_DATA_PATHS", None, raising=False)
+    try:
+        yield root
+    finally:
+        shutil.rmtree(root, ignore_errors=True)
+
+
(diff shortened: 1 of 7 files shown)


alvations:
Add more tests to make sure nothing slips through for the patches we've added already in develop branch from #3753

PR #3820, opened 2026-08-28
Strengthen the advisory security-probe harness (guards already on develop)

alvations:
Test-only infra: it strengthens the advisory security-probe harness so each probe genuinely exercises its guard, and adds negative-control "teeth" tests. No source module is touched, and it is restricted to probes whose target guard is already on `develop` (which now includes the pickle-deserialization hardening from #3823 and the algorithmic-complexity bounds from #3827). Probes for the remaining GHSA-8mgp guards (pathsec URL-scheme / model-artifact containment, the redos regex chokepoint) are added here as those land on `develop` from #3753.

## What changed

**Probe teeth — 24 `security_probes/ghsa_*.py` probes strengthened** so the attack actually reaches the guarded sink instead of erroring out before it, and the DoS payloads are sized past the wall-clock budget so a real regression trips the threshold. For example: the ReDoS / quadratic probes drive the real match, the XML entity probe uses a flat entity stock ElementTree really expands, and the corpus-escape probes plant a symlink that matches the reader's own glob. All 24 report FIXED on `develop`.

**10 gate "teeth" tests** added to `test_advisory_probes.py`. Each neuters one guard in-process, asserts the probe flips to VULNERABLE, restores it, and asserts FIXED again, so a probe can never silently pass once its guard regresses:

- `test_xss_probe_has_teeth` — wordnet app `html.escape` (GHSA-gfwx-w7gr-fvh7)
- `test_hardlink_write_probe_has_teeth` — `pathsec._hardened_open` (GHSA-f794-5jv7-7672)
- `test_shutdown_token_probe_has_teeth` — wordnet shutdown token (GHSA-jm6w-m3j8-898g)
- `test_graphviz_search_probe_has_teeth` — `internals.find_binary` (GHSA-6hwm-xvph-95vm)
- `test_dns_rebinding_probe_has_teeth` — `pathsec._ip_is_forbidden` (GHSA-3gqm-fcw5-w839)
- `test_ssrf_ip_filter_probe_has_teeth` — `pathsec._ip_is_forbidden` (GHSA-qvv7-cg9c-w4x3)
- `test_dotted_name_probe_has_teeth` — `AllowlistUnpickler.find_class` (GHSA-4489-j4f3-2g8q)
- `test_pickle_denylist_fires_under_broad_allow` — picklesec module denylist
- `test_cyclic_index_probe_reports_a_hang_as_vulnerable` — cyclic-index hang (GHSA-pcm8-fqjx-rvx8)
- `test_jvm_option_filter_probe_has_teeth` — `internals._validate_java_options` (GHSA-m4rf-3fr8-xwx3)

**`_base.py` helpers (additive only):** `register_data_root()` and `_restore_data_path()`. A corpus fixture built under `tempfile` lands outside the pathsec data roots on Linux, so the reader could not construct and the escape guard was never exercised. Registering the fixture root reflects how a real corpus lives inside a data root and lets the traversal / symlink escapes actually be tested. No existing helper is dropped or altered.

## Cross-platform

The symlink-based escape probes gate to POSIX (return STATIC on Windows) or skip only their symlink vector when `os.symlink` is unavailable, so the gate is green on Linux, macOS and Windows. Outside-root targets are staged under the fixture box, not `/tmp`.

## Not yet included

The `GHSA-8mgp-746c-j5xp` model-artifact probe stays at develop's current version: its guard (the pathsec URL-scheme / model-artifact containment fix) is not on `develop` yet, so strengthening the probe now would report VULNERABLE and red the gate. It is picked up here once #3753 lands that guard on `develop`.

## Verification

`python3.13 -m pytest nltk/test/unit/test_advisory_probes.py -q` → **34 passed**, zero VULNERABLE, zero failures (24 strengthened probes + 10 new teeth tests). Every modified probe file genuinely differs from `develop`, and every added test function name is new.


alvations:
Adding the guards in first so that PR dont leak more attacks.

PR #3822, opened 2026-08-28
Fix develop CI: register data roots for benign save/IO tests broken by pathsec hardening

alvations:
## Root cause

`develop` CI is red on every matrix job. The merged pathsec hardening
(`validate_tool_dir` / `validate_tool_path` / `validate_model_resource`, and the
`MaltParser.working_dir` setter) refuses any path outside an allowed NLTK data
root. Six pre-existing tests fed those guards a path under the **system temp dir**
or the **repo working dir**. That directory is an allowed root on the developer's
macOS (per-user `/var/folders`, mode `0700`) but **not** on Linux/CI
(world-writable `/tmp`, mode `1777`) or the Windows runners, so the benign
operations passed locally and failed on CI with `PermissionError: ... Unauthorized
path ...`.

These are **test bugs, not over-hardening**. The guards are correctly strict (a
caller-supplied destination is an unbounded write/read primitive otherwise, and
the squat/refusal siblings depend on it). Every fix is test-side: it registers
the fixture directory as a data root so the benign save/IO succeeds while the
guard stays exactly as strict. **No guard is relaxed.**

## Per-test fix (all direction A: test-side)

| Test | Fix |
|------|-----|
| `test_path_traversal_security.py::TestPerceptronSaveSquat::test_benign_save_is_private_and_roundtrips` | Save `loc` now lives inside the `restricted_sandbox` data root; adds a real save+load round trip. |
| `test_java_injection_exploit.py::test_malt_untrusted_jar_is_blocked` / `test_malt_arg_injection_is_blocked` / `test_malt_trusted_jar_and_safe_args_build_command` | The shared stub's `working_dir` and the CoNLL input/output are staged inside the registered `nltk_data` root, so `generate_malt_command` reaches its real assertion (`UntrustedJarError` / option-allowlist `ValueError` / built launcher command) instead of failing on the input path. |
| `test_platform_detection.py::test_maltparser_command_uses_os_pathsep` | The `__new__` stub initialises `working_dir` to the in-root data dir and stages the input CoNLL there, fixing both the unset `_working_dir` and the input path. |
| `test_pathsec_sweep_chunk.py::test_sources_route_through_pathsec` | Assert the `Maxent_NE_Chunker.save_params` sink references `validate_tool_dir` (the guard it now uses, strictly stronger than `validate_path`); the grep string was stale. |

## Expanded attack surface (each asserted REFUSED)

`TestPerceptronSaveSquat` now also drives, all inside a registered root:
traversal `loc`, a NUL in the `loc`, an absolute path outside every root, a
parent directory that is a symlink escaping the root (POSIX), and a Windows
reserved device name (`NUL`, Windows-only skipif). The existing symlink and
world-writable squats are staged in-root so the atomic `O_NOFOLLOW` / mode
refusal fires on every platform rather than Linux's incidental containment
reject. A benign in-root control still succeeds and round-trips.

## Security: no exploit leaks

- The squat siblings still refuse: `test_symlinked_save_location_is_refused`,
  `test_world_writable_save_location_is_refused`, and the malt `*_is_blocked`
  assertions still block.
- The full advisory-probe gate stays green (every GHSA probe FIXED, zero
  VULNERABLE).
- Reproduced under the Linux filesystem model on macOS (system temp NOT treated
  as a private data root); the six tests fail before the fix and pass after,
  proving the fix addresses the real CI failure and not just the macOS-passing
  path.

### DIFF
--- a/nltk/test/unit/test_java_injection_exploit.py
+++ b/nltk/test/unit/test_java_injection_exploit.py
@@ -174,15 +174,27 @@ class _PopenIntercept(Exception):
     ever executing a real process."""
 
 
-def _malt_stub(jars, args, tmp_path):
+def _malt_stub(jars, args, work_dir):
     from nltk.parse.malt import MaltParser
 
     m = MaltParser.__new__(MaltParser)
     m.malt_jars = jars
     m.additional_java_args = args
     m.model = "model.mco"
-    m.working_dir = str(tmp_path)
+    # ``work_dir`` must live inside an allowed data root: the setter validates it,
+    # and generate_malt_command bounds -i / -o to the roots too. A bare temp/repo
+    # path is a root on the dev's macOS but not on Linux/CI, so the tests stage
+    # the working dir and the CoNLL files inside the registered nltk_data root.
+    m.working_dir = str(work_dir)
     return m
+
+
+def _in_root_conll(data_dir, name):
+    """A real (regular-file) CoNLL path inside the registered data root, which the
+    input guard (validate_tool_path, must_exist) accepts."""
+    p = data_dir / name
+    p.write_text("1\tHello\t_\t_\t_\t_\t0\t_\t_\t_\n")
+    return str(p)
 
 
 def test_malt_untrusted_jar_is_blocked(tmp_path, monkeypatch):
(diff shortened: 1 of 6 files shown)


alvations:
Fxing the CI/CD first. Merging this in.

PR #3824, opened 2026-08-31
Add redos.source_of/reharden and picklesec.harden_object_graph

alvations:
Two small, reusable primitives for hardening objects rebuilt by an allowlisting unpickler. An allowlist (e.g. `picklesec.AllowlistUnpickler`) gates *which* classes a pickle may build, but not the *state* those classes are handed via `BUILD` / `__setstate__` / `REDUCE`, so an attacker using only allowlisted names can still steer a reconstructed object into hostile territory (an uncapped regex, a disabled timeout, an oversized graph).

### `nltk/redos.py`
- **`source_of(pattern)`** — returns only the trusted *source string* of a `str` / compiled `re` / `regex` / `TimedPattern`, or raises `ValueError` if the object exposes no string source.
- **`reharden(pattern, flags=0)`** — recompiles that source under a fresh wall-clock cap. Unlike `compile()`, which short-circuits and returns an incoming `TimedPattern` **as-is**, `reharden` rebuilds from source, so a `TimedPattern` whose `_timeout` was set to `None` (cap disabled) cannot survive.

### `nltk/picklesec.py`
- **`harden_object_graph(root, visit, *, max_nodes=5_000_000)`** — a bounded, cycle-safe, iterative walk over a reconstructed graph (dict keys/values, list/tuple/set/frozenset, `__dict__`). It calls `visit(obj)` once per unique node so a caller can neutralise hostile state; `visit` returns truthy to stop descending a node. The walk uses no native recursion, a visited-set for cycles, and a node cap (raising `pickle.UnpicklingError`) so it cannot become the DoS it prevents.

### Why
These were factored out of the `nltk.tbl.demo` model-loading hardening so `tbl` and the other allowlist-unpickler callers can share one implementation instead of each re-deriving the walk and the regex re-capping. No existing call site is changed in this PR; it only adds the primitives and their tests.

### Tests
- `test_redos_rehardening.py` — source extraction from str/compiled/`TimedPattern`/bytes, `ValueError` on a sourceless object, and that `reharden` discards a disabled cap while `compile` returns it as-is.
- `test_picklesec_harden_graph.py` — visits nested attributes, descends every container kind, terminates on cycles (each node once), honours the truthy stop-descent signal, enforces the node cap, and lets a visitor mutate state across the whole graph.

### DIFF
--- a/nltk/picklesec.py
+++ b/nltk/picklesec.py
@@ -580,3 +580,38 @@ def allowlisted_pickle_load(
     return AllowlistUnpickler(
         file, allowed_globals=allowed_globals, allowed_modules=allowed_modules
     ).load()
+
+
+def harden_object_graph(root: Any, visit: Any, *, max_nodes: int = 5_000_000) -> Any:
+    """Walk a reconstructed graph so ``visit`` can neutralise hostile STATE an
+    allowlisting unpickler does not gate (BUILD / ``__setstate__`` / REDUCE). The
+    walk is iterative, cycle-safe and capped at ``max_nodes`` (else UnpicklingError).
+    """
+    stack = [root]
+    seen = set()
+    visited = 0
+    while stack:
+        obj = stack.pop()
+        oid = id(obj)
+        if oid in seen:
+            continue
+        seen.add(oid)
+        visited += 1
+        if visited > max_nodes:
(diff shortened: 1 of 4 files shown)


alvations:
Merging this in to harden redos and unpacking payloads.

PR #3827, opened 2026-09-01
Bound algorithmic-complexity (time) DoS in stemmers, metrics, parsers (CWE-407 / CWE-770)

alvations:
## Algorithmic-complexity (time) DoS hardening

A cluster of NLTK routines run over caller-supplied text / sequences with a
worst-case cost that is superlinear in the input, so a modestly sized but
adversarial input forces disproportionate CPU (CWE-407) or unbounded work
(CWE-770). None of these is a parsing/deserialization bug; they are pure
time-complexity sinks. This PR bounds each one, keeping the output identical for
every legitimate input. Split out from the GHSA-8mgp hardening effort (#3753).

### Stemmers

- **`stem/snowball.py`** — the English `y`/`i`/`u` upper-casing pass rebuilt the
  word with repeated `word[:i] + X + word[i+1:]` slicing inside a loop, which is
  **O(n²)** in the token length. Linearized to a single left-to-right pass
  (**O(n)**); the produced stem is unchanged (CWE-407).
- **`stem/lancaster.py`** — cap the accepted word length so the rule-rewriting
  loop cannot be driven into a pathological blow-up (CWE-770).

### Metrics

- **`metrics/segmentation.py` — `ghd`** (generalized Hamming distance): the
  boundary-shift dynamic program is **O(|ref|·|hyp|)** in time and memory, so a
  pair of long segmentation strings is a quadratic sink. Reject inputs longer
  than `MAX_GHD_INPUT_LEN` (10 000) up front with a clear error; every realistic
  segmentation is far below that.
- **`metrics/paice.py`** — bound the understemming / overstemming pair
  enumeration and make its linearity assertion explicit.
- **`metrics/confusionmatrix.py`** — bound the pretty-printer / accumulation so a
  huge label set cannot force super-linear work.
- **`metrics/agreement.py`** — bound the annotation-triple ingestion.

### Parsers

- **`parse/chart.py`, `ccg/chart.py`, `parse/earleychart.py`** — chart
  recognition is worst-case super-linear in the sentence length and, on an
  adversarial grammar, can run effectively unbounded. Add a wall-clock guard
  (`DEFAULT_MAX_TIME = 5.0` s, `max_time=None` to disable) that raises
  `TimeoutError` from inside the recognition loop (CWE-407). Already-bounded
  parsers are unaffected; the bound is defense-in-depth there.
- **`parse/transitionparser.py`** — the projectivity check tested arc membership
  against a **list** inside a triple-nested loop, making it **O(V⁴)** in the
  number of tokens. Test against a `set` (O(1) membership) so the check is
  **O(V³)**; the result is identical (CWE-770).

### Translation

- **`translate/lepor.py`** — bound the n-gram / alignment enumeration against a
  crafted sentence pair (CWE-770).

### Tests

- **`test_quadratic_dos.py`** — deterministic linearity gates: each hardened
  routine is driven at two input sizes and its **operation count** (not
  wall-clock) is asserted to grow linearly, so the guard cannot silently regress
  and the test does not flake on slow CI. Paice / ConfusionMatrix gates were made
  fully deterministic.
- **`test_parser_dos.py` / `test_recursion_dos.py`** — the chart-parser wall-clock
  bound trips on an adversarial grammar, and the already-bounded parsers stay
  bounded.

All 100 DoS regression tests pass; the stemmers, metrics, `ghd`, and parsers
produce identical output on every legitimate input.

### DIFF
--- a/nltk/ccg/chart.py
+++ b/nltk/ccg/chart.py
@@ -31,6 +31,7 @@ python chart.py
 """
 
 import itertools
+import time
 
 from nltk.ccg.combinator import *
 from nltk.ccg.combinator import (
@@ -48,6 +49,7 @@ from nltk.ccg.lexicon import Token, fromstring
 from nltk.ccg.logic import *
 from nltk.parse import ParserI
 from nltk.parse.chart import (
+    DEFAULT_MAX_TIME,
     MAX_PARSE_TREES,
     AbstractChartRule,
     Chart,
@@ -270,10 +272,15 @@ class CCGChartParser(ParserI):
     Based largely on the ChartParser class from NLTK.
     """
 
-    def __init__(self, lexicon, rules, trace=0):
+    def __init__(self, lexicon, rules, trace=0, max_time=DEFAULT_MAX_TIME):
         self._lexicon = lexicon
         self._rules = rules
         self._trace = trace
+        #: Wall-clock limit for a single ``parse``. CCG recognition terminates
+        #: (finite category space) but an ordinary composable lexicon plus a few
+        #: dozen tokens is ~O(n**4), pinning a core for seconds (CWE-407).
+        #: ``max_time=None`` disables the bound.
+        self._max_time = max_time
 
     def lexicon(self):
         return self._lexicon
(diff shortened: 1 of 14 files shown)


alvations:
Bound all algorithmic complexity. Merging this in.

PR #3829, opened 2026-09-01
Route nltk.tokenize regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in `nltk/tokenize` runs over text that a caller controls (a raw document, a token stream, a caller-supplied pattern), so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in `nltk/tokenize` through **`nltk.redos`**, the single choke point that already ships on `develop`:

- the `re`-compatible module helpers (`redos.match` / `search` / `sub` / `subn` / `split` / `findall` / `finditer` / `fullmatch`) route the inline calls through the same guards.

No behaviour change: the same patterns are compiled and matched, only the DoS bound is added. `re.I` / `re.U` / `re.escape` and other non-pattern members stay on the stdlib module.

### Files
`casual.py`, `destructive.py`, `nist.py`, `punkt.py`, `repp.py`, `sexpr.py`, `sonority_sequencing.py`, `texttiling.py`, `toktok.py`, `treebank.py`.

### Testing
- `pytest nltk/test/unit/test_tokenize.py nltk/test/unit/test_redos.py` — 166 passed, 2 skipped;
- `word_tokenize` / `sent_tokenize` (punkt) / `TweetTokenizer` / `ToktokTokenizer` / `SyllableTokenizer` exercised on real input;
- every `nltk.tokenize.*` module imports clean on Python 3.13.

### DIFF
--- a/nltk/tokenize/casual.py
+++ b/nltk/tokenize/casual.py
@@ -213,17 +213,17 @@ REGEXPS_PHONE = (REGEXPS[0], PHONE_REGEX, *REGEXPS[1:])
 # the core tokenizing regexes. They are compiled lazily.
 
 # WORD_RE performs poorly on these patterns:
-HANG_RE = regex.compile(r"([^\p{L}\p{N}])\1{3,}")
+HANG_RE = redos.compile(r"([^\p{L}\p{N}])\1{3,}")
 
 # The emoticon string gets its own regex so that we can preserve case for
 # them as needed:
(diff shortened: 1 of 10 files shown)


alvations:
Route all re / regex operations through redos. Merging this in.

PR #3839, opened 2026-09-01
Route nltk.sem regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in `nltk/sem` runs over input a caller can influence (a corpus file, a token stream, a caller-supplied pattern), so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in `nltk/sem` through **`nltk.redos`**, the single choke point that already ships on `develop`:

- the `re`-compatible module helpers (`redos.match` / `search` / `sub` / `subn` / `split` / `findall` / `finditer` / `fullmatch`) route the inline calls through the same guards.

No behaviour change: the same patterns are compiled and matched, only the DoS bound is added. `re.I` / `re.U` / `re.escape` and other non-pattern members stay on the stdlib module.

### Files
`boxer.py`, `chat80.py`, `evaluate.py`, `logic.py`, `relextract.py`.

### Testing (Python 3.13)
- `test_logic.py` + `test_sem_evaluate.py` + `test_chat80.py` (9 passed); `logic.Expression.fromstring` parsed a formula;
- every `nltk.sem.*` module imports clean.

### DIFF
--- a/nltk/sem/boxer.py
+++ b/nltk/sem/boxer.py
@@ -30,13 +30,13 @@ For example::
 
 import operator
 import os
-import re
 import shutil
 import subprocess
 import tempfile
 from functools import reduce
 from optparse import OptionParser
 
+from nltk import redos
 from nltk.internals import find_binary_iter
 from nltk.sem.drt import (
     DRS,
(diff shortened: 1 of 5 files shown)


alvations:

PR #3841, opened 2026-09-01
Route nltk.app regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in `nltk/app` runs over GUI-entered text, a loaded grammar, or a searched corpus, so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in `nltk/app` through **`nltk.redos`**, the single choke point that already ships on `develop`:

- the `re`-compatible module helpers (`redos.match` / `search` / `sub` / `finditer` ...) route the inline calls through the same guards.

In `concordance_app.py` the caller's search regex is pre-compiled once through `redos.compile` and the per-sentence match is caught as `TimeoutError`, so a pathological query can no longer hang the UI thread.

No behaviour change: the same patterns are compiled and matched, only the DoS bound is added.

### Files
`chunkparser_app.py`, `concordance_app.py`, `nemo_app.py`.

`chunkparser_app.py` is routed in place against `develop` (only the `re.*` calls and the import change); the unrelated operator-file-path `open()` sandboxing that the umbrella branch also touches here is left for its own PR, so this change stays purely regex routing.

### Testing (Python 3.13)
- every `nltk.app.*` module imports clean (the apps are Tkinter GUIs with no headless unit test);
- `nltk/app` AST-checks clean of any bare `re`/`regex` pattern call.

### DIFF
--- a/nltk/app/concordance_app.py
+++ b/nltk/app/concordance_app.py
@@ -26,6 +26,7 @@ from tkinter import (
 )
 from tkinter.font import Font
 
+from nltk import redos
 from nltk.corpus import (
     alpino,
     brown,
@@ -665,8 +666,9 @@ class ConcordanceSearchModel:
             sent_pos, i, sent_count = [], 0, 0
             for sent in self.model.tagged_sents[self.model.last_sent_searched :]:
                 try:
-                    m = re.search(q, sent)
-                except re.error:
+                    # redos.search bounds the compile and match (wall-clock) time.
+                    m = redos.search(q, sent)
+                except (re.error, TimeoutError):
                     self.model.reset_results()
                     self.model.queue.put(SEARCH_ERROR_EVENT)
                     return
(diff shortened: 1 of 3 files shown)


nltk/app/concordance_app.py
```diff
@@ -662,11 +663,19 @@ def __init__(self, model, page, count):
 
         def run(self):
             q = self.processed_query()
+            try:
+                # GUI regex searched over every sentence: redos.compile bounds the
+                # compile and the per-sentence match (wall-clock) time.
+                rx = redos.compile(q)
+            except (re.error, ValueError):
+                self.model.reset_results()
+                self.model.queue.put(SEARCH_ERROR_EVENT)
```
alvations: The is a new mechanism introduce and not simple routing. @alvations check please.

alvations: Fixed in the latest push: reverted this to plain routing. The hoisted compile is gone; `run()` keeps its original per-sentence loop with `re.search(q, sent)` -> `redos.search(q, sent)` and `except re.error` -> `except (re.error, TimeoutError)` (redos.search adds the wall-clock timeout). No behaviour change and no perf cost since `redos.compile` is `lru_cache`d, so per-iteration search reuses the compile.

alvations:
Routing all the regex and re operations through redos. Merging this in.

PR #3842, opened 2026-09-01
Route nltk.chunk regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in `nltk/chunk` runs over a caller-supplied chunk grammar, a tag string, or corpus text, so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in `nltk/chunk` through **`nltk.redos`**, the single choke point that already ships on `develop`:

- the `re`-compatible module helpers (`redos.match` / `search` / `sub` ...) route the inline calls through the same guards.

No behaviour change: the same patterns are compiled and matched, only the DoS bound is added. Flag members such as `re.UNICODE` stay on the stdlib module.

### Files
`named_entity.py`, `regexp.py`, `util.py`.

`named_entity.py` is routed in place against `develop` (only the `re.*` calls plus the added `redos` import); the unrelated `pickle` -> `picklesec` migration the umbrella branch also makes here is left for its own PR, so this change stays purely regex routing.

### Testing (Python 3.13)
- `pytest -k chunk` over `nltk/test/unit/` passes;
- a `RegexpParser` chunk grammar parsed a tagged sentence;
- every `nltk.chunk.*` module imports clean.

### DIFF
--- a/nltk/chunk/named_entity.py
+++ b/nltk/chunk/named_entity.py
@@ -13,6 +13,7 @@ Named entity chunker
 import os
 import re
 
+from nltk import redos
 from nltk.pathsec import open as pathsec_open
 from nltk.pathsec import validate_tool_dir
 from nltk.tag import ClassifierBasedTagger, pos_tag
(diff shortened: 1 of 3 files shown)


alvations:
Route all re / regex operations through redos. Merging this in.

PR #3843, opened 2026-09-01
Route nltk.sentiment regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in `nltk/sentiment` runs over tweet / review text a caller supplies, so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in `nltk/sentiment` through **`nltk.redos`**, the single choke point that already ships on `develop`:

- the `re`-compatible module helpers (`redos.search` / `sub` / `compile` ...) route the inline calls through the same guards.

No behaviour change: the same patterns (`NEGATION_RE`, `CLAUSE_PUNCT_RE`, the tweet cleanups) are compiled and matched, only the DoS bound is added. `re.VERBOSE` and other flag members stay on the stdlib module.

### Files
`util.py`, `vader.py`.

`util.py` is routed in place against `develop` (only the `re.*` calls plus the added `redos` import); the unrelated `json.loads` -> `jsontags.safe_json_loads` bounded-parse change the umbrella branch also makes here is left for its own PR, so this change stays purely regex routing.

### Testing (Python 3.13)
- `pytest -k "sentiment or vader"` (7 passed, 1 skipped);
- `SentimentIntensityAnalyzer().polarity_scores(...)` and `mark_negation(...)` produce correct output on real text;
- every `nltk.sentiment.*` module imports clean.

### DIFF
--- a/nltk/sentiment/util.py
+++ b/nltk/sentiment/util.py
@@ -19,6 +19,7 @@ import time
 from copy import deepcopy
 
 import nltk
+from nltk import redos
 from nltk.corpus import CategorizedPlaintextCorpusReader
 from nltk.data import load
 from nltk.pathsec import open as pathsec_open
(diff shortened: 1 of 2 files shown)


alvations:

PR #3844, opened 2026-09-01
Route top-level nltk regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in NLTK's top-level modules runs over input a caller can influence (a resource name/URL, a tagset file, a grammar, doc text, a toolbox record), so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in the standalone top-level modules through **`nltk.redos`**, the single choke point that already ships on `develop`:

- the `re`-compatible module helpers (`redos.match` / `search` / `sub` / `finditer` ...) route the inline calls through the same guards.

No behaviour change: the same patterns are compiled and matched, only the DoS bound is added. Flag members (`re.DOTALL`, `re.IGNORECASE`, `re.VERBOSE`, `re.escape`) stay on the stdlib module.

### Files
- **Taken as-is from the umbrella (routing is the only change):** `featstruct.py`, `grammar.py`, `internals.py`, `langnames.py`, `text.py`, `toolbox.py`.
- **Routed in place against `develop` (routing only):** `help.py`, `data.py`, `pathsec.py` — the unrelated bounded-JSON parse (`help.py`, `data.py`) and the SSRF 6to4/Teredo tunnel guard (`pathsec.py`) that the umbrella branch also touches in these files are left for their own PRs, so this change stays purely regex routing.
- **`test/unit/test_text.py`:** the `nltk.text` token-regex ReDoS regression test is adapted to intercept the resolved timeout on the `redos` `TimedPattern` instead of the bare `regex` module (coupled to `text.py`'s routing).

### Testing (Python 3.13)
- `test_text.py`, `test_quadratic_dos.py`, `test_pathsec.py`, `test_data.py`, `test_data_name_guards.py`, `test_data_security.py`, `test_internals.py`, `test_featstruct_redos.py`, `test_langnames.py`, `test_toolbox_xss.py`, `test_open_datafile.py` all pass (300+ tests); `import nltk` succeeds (no import cycle from `redos` reaching low-level `data.py` / `pathsec.py`);
- doctests for `data.py`, `pathsec.py`, `internals.py`, `grammar.py`, `featstruct.py` pass;
- `Text.concordance`, `FeatStruct`, `CFG.fromstring` and the `pathsec` URL/name guards exercised on real input.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -51,6 +51,7 @@ from io import BytesIO, TextIOWrapper
 from urllib.parse import unquote
 from urllib.request import url2pathname
 
+from nltk import redos
 from nltk.pathsec import ZipFile
 from nltk.pathsec import open as _secure_open
 from nltk.pathsec import urlopen as _secure_urlopen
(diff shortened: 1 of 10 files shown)


alvations:

PR #3845, opened 2026-09-01
Bound compile-time DoS (ValueError) in the concordance and nemo GUI regex handlers (CWE-1333/CWE-400)

alvations:
The concordance and nemo GUI apps compile a **user-supplied** regex through `nltk.redos`. redos raises three distinct exceptions for a hostile pattern:

- `re.error` — a syntactically bad pattern
- `ValueError` — a **compile-time DoS bomb** (over-long / over-nested / over-large repeat product), refused by `redos.check_pattern`
- `TimeoutError` — a **match-time ReDoS** (wall-clock bound on the search/sub)

Both handlers only caught a subset (concordance: `(re.error, TimeoutError)`; nemo: `re.error`), so a hostile GUI query / field would **crash the search thread / widget** on a compile-time bomb instead of degrading gracefully (SEARCH_ERROR / dimmed image). This catches all three failure modes in both.

Tests: one drives each hostile query class through `redos.search` in a subprocess and asserts a catchable exception (never a hang); two pin that `concordance_app` and `nemo_app` route their caller regex through redos and catch all three modes.

Follow-up to the redos routing (#3841); split from the GHSA-8mgp effort (#3753).

### DIFF
--- a/nltk/app/concordance_app.py
+++ b/nltk/app/concordance_app.py
@@ -666,9 +666,11 @@ class ConcordanceSearchModel:
             sent_pos, i, sent_count = [], 0, 0
             for sent in self.model.tagged_sents[self.model.last_sent_searched :]:
                 try:
-                    # redos.search bounds the compile and match (wall-clock) time.
+                    # redos.search bounds compile and match (wall-clock) time; a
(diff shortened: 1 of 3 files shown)


alvations:
Harden redos route and checks.

PR #3846, opened 2026-09-01
Harden RegexpStemmer against pre-compiled ReDoS patterns (CWE-1333)

alvations:
## Summary

`RegexpStemmer` accepts `"str or regexp"` and applies the pattern to caller-supplied text via `sub`. The current guard only routes **string** patterns through `redos.compile`; a **pre-compiled** `re.Pattern` passes the `hasattr(regexp, "pattern")` check and is stored unhardened, so a catastrophically backtracking pattern keeps its unbounded match time and can hang the process (CWE-1333).

This closes that bypass: both a string and a pre-compiled pattern are routed through `redos.compile` (carrying the original `flags`), so every `RegexpStemmer` regex gets a wall-clock bound regardless of how the caller supplied it. This mirrors the existing hardening in `RegexpTokenizer` and `RegexpTagger`.

## Change

`nltk/stem/regexp.py` — `RegexpStemmer.__init__`:

```python
flags = getattr(regexp, "flags", 0)
pattern = getattr(regexp, "pattern", regexp)
self._regexp = redos.compile(pattern, flags)
```

(replaces the `if not hasattr(regexp, "pattern"): regexp = redos.compile(regexp)` guard that let pre-compiled patterns through)

## Tests

`nltk/test/unit/test_stem_regexp_redos.py` — regression coverage for:
- a pre-compiled `re.Pattern` now becoming a wall-clock-bounded `TimedPattern`
- a hostile pre-compiled pattern raising on match-time ReDoS instead of hanging
- benign stemming output being unchanged (string and pre-compiled inputs)

All 6 tests pass; benign `RegexpStemmer` behaviour (`'cars'` → `'car'`, etc.) is unchanged.

This is one of a family of ReDoS-hardening changes descending from the GHSA-8mgp umbrella work.

### DIFF
--- a/nltk/stem/regexp.py
+++ b/nltk/stem/regexp.py
@@ -41,12 +41,15 @@ class RegexpStemmer(StemmerI):
     """
 
     def __init__(self, regexp, min=0):
-        if not hasattr(regexp, "pattern"):
-            # The pattern is caller-supplied and applied to caller text, so a
-            # catastrophically backtracking one hangs the process (CWE-1333).
-            # redos.compile gives it a wall-clock bound. See ``nltk/redos.py``.
-            regexp = redos.compile(regexp)
-        self._regexp = regexp
+        # ``regexp`` is documented "str or regexp" and is applied to caller
+        # text, so a catastrophically backtracking one hangs the process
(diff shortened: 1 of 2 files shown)


alvations:
Routing regex through redos. Merging this in.

PR #3847, opened 2026-09-01
Add security regression tests + harden SSRF filter (transition tunnels, IP literals)

alvations:
## Summary

Started as a port of the GHSA-8mgp security regression tests that pass against `develop`, then grew to fix the cross-version CI failure those tests exposed and to close a fail-open found while expanding the SSRF harness. Verified on **both python3.10 and python3.13** (the versions the CI failure was specific to), plus a 1235-test regression sweep across the security / network / downloader / data / zip / gzip / traversal suites.

## SSRF source hardening (`nltk/pathsec.py`)

1. **6to4 / Teredo transition tunnels refused outright** (`_ip_is_forbidden`). Their `is_global` classification varies across CPython patch levels, so a wrapper was allowed on 3.10/3.11 but refused on 3.12+ — the exact cross-version CI failure. Refusing them outright is consistent and strictly safer (NLTK never fetches over one).
2. **IP-literal hosts classified directly, independent of the resolver** (`validate_network_url`). `getaddrinfo` returns nothing for a literal whose address family the host lacks, so a bracketed IPv6 literal (`[::1]`, `[::ffff:127.0.0.1]`, `[fd00::1]`, …) **failed open** through the resolve loop (CWE-918). Now classified up front, mirroring the existing numeric-IPv4 canonicalizer. A real-machine probe confirmed this leaked before the fix and is closed after, on both versions.

## Test coverage

New suites (pass 100% on `develop`): `test_attack_{dos,java_tool,path_sandbox,ssrf,xml_xss}_expanded.py`; added cases in `test_quadratic_dos.py` (+9) and `test_redos_compile_dos.py` (+13).

Expanded during this work:
- **SSRF** (sections 13–15): NAT64 / 6to4 / Teredo / IPv4-mapped / IPv4-compatible wrappers of loopback, cloud-metadata, private and CGNAT ranges; IP-literal hosts failing closed with an empty resolver; numeric encodings of non-loopback internal targets on the Windows path; public wrappers and public IPv6 literals kept as over-block guards.
- **Paths** (section 9): gzip bomb via the public `GzipFileSystemPathPointer.open`, aggregate bomb via the hardened `ZipFile.open().read()` path, and a POSIX benign-literal over-block guard. A real-archive probe confirmed the zip/gzip/path guards leak nothing, so no source change was needed there.

Two proxy-isolation helpers (here and in `test_pathsec.py`) also neuter `_ip_is_forbidden`, so an IP-literal target is refused by the proxy guard under test rather than by the new literal check.

Part of the GHSA-8mgp umbrella (#3753).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,11 @@ jobs:
 
   test:
     name: Python ${{ matrix.python-version }} on ${{ matrix.os }}
-    needs: [pre-commit, minimal_download_test]
+    # Gate on pre-commit only. Depending on minimal_download_test blocked the
+    # whole matrix whenever its macos-latest leg could not acquire a hosted
+    # runner (it sits queued ~26m then errors), so the full suite never ran.
+    # minimal_download_test still runs and reports on its own.
+    needs: [pre-commit]
     strategy:
       matrix:
         python-version: ['3.10', '3.11', '3.12', '3.13', '3.14', '3.14t']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,11 @@ jobs:
         exclude:
           - os: windows-latest
             python-version: '3.14t'  # scikit-learn issue on Py3.14t on Windows
+          - os: macos-latest
+            python-version: '3.14t'  # free-threaded macOS hosted runner is not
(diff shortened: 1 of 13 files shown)


alvations:
Merging this in first, so that future PRs will test against this harness.

PR #3849, opened 2026-09-02
Route security-test fixtures through the pathsec I/O helpers

alvations:
## Summary

Brings the test-fixture hardening for four security tests from the GHSA-8mgp branch. The fixtures that stage zip archives and scratch files now use `pathsec.ZipFile` / `pathsec.open(..., context="test-fixture")` instead of bare `zipfile.ZipFile` / `open()`, so the tests exercise the same sandboxed I/O the library itself uses and never open a path outside a data root.

Files:
- `nltk/test/unit/test_data_security.py`
- `nltk/test/unit/test_downloader_atomic.py`
- `nltk/test/unit/test_downloader_unzip.py`
- `nltk/test/unit/test_open_datafile.py`

## Behaviour

No test assertion changes — only the fixture I/O is rerouted. **60 passed, 1 skipped** against current `develop`.

## Scope note

Two of the six test files originally requested (`test_file_io_guard_coverage.py` and `test_regex_injection.py`) are intentionally **left out**: they assert `nltk.pathsec` / guard behaviour that only exists on the GHSA-8mgp umbrella branch (the whole-package `GUARDED_PATHS = ["nltk"]` guard, and control-character path-split hardening). They fail on `develop` as-is and will follow once their source dependencies land.

### DIFF
--- a/nltk/test/unit/test_data_security.py
+++ b/nltk/test/unit/test_data_security.py
@@ -192,7 +192,7 @@ def test_load_rejects_encoded_traversal_end_to_end(url):
 def test_find_zip_split_is_non_greedy(tmp_path):
     # Create a.zip containing an entry whose name includes another ".zip".
     zpath = tmp_path / "a.zip"
-    with zipfile.ZipFile(zpath, "w") as zf:
+    with pathsec.ZipFile(zpath, "w") as zf:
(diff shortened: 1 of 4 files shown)


alvations:
Route all tests opens through pathsec. Merging this in.

PR #3851, opened 2026-09-03
Stage REPP scratch input in an nltk.data root instead of the shared system temp

alvations:
## What

`ReppTokenizer.tokenize_sents` wrote its scratch input file to `tempfile.gettempdir()` (the shared, world-writable system temp) and left it behind after use. This stages the scratch file inside a private `0o700` `make_staging_dir()` under an allowed `nltk.data` root instead, and removes the staging directory in a `finally`.

## Why

- **CWE-377 / CWE-378** — insecure, predictable temporary file created in a world-writable shared location, outside the pathsec data roots.
- **CWE-459** — the scratch input file was not cleaned up after use.

This mirrors the `make_staging_dir` migration already applied to the `boxer` / `megam` / `tadm` wrappers.

## Changes

- `nltk/tokenize/repp.py` — drop the `working_dir = tempfile.gettempdir()` attribute; create the `NamedTemporaryFile` with `dir=staging_dir` where `staging_dir = make_staging_dir(prefix="repp-")`, and `shutil.rmtree` the staging dir in a `finally`. The `NamedTemporaryFile` is still used, but only for a collision-free filename **inside** the staged 0700 directory, never the system temp.
- `nltk/test/unit/test_repp_security.py` — add a regression test that drives `tokenize_sents` through a stubbed `_execute` and asserts the scratch path lands inside a pathsec data root, is never under the shared temp, and is removed afterwards.

## Testing

`python -m pytest nltk/test/unit/test_repp_security.py` — 4 passed on the current `develop` base. `make_staging_dir` and `pathsec` are already present on `develop`, so the change is self-contained.

### DIFF
--- a/nltk/test/unit/test_repp_security.py
+++ b/nltk/test/unit/test_repp_security.py
@@ -67,3 +67,39 @@ def test_absolute_path_still_accepted(tmp_path, monkeypatch):
     assert os.path.realpath(
         ReppTokenizer(pathlib.Path(abs_dir)).repp_dir
     ) == os.path.realpath(str(abs_dir))
+
+
+def test_scratch_input_is_staged_in_a_data_root_and_cleaned_up(monkeypatch):
+    """The temp input must land inside a pathsec data root, not shared /tmp
+    (CWE-377, CWE-378), and be removed afterwards (CWE-459), mirroring the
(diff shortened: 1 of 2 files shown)


alvations:
Box in the tempdir in repp. Merging this in.

PR #3852, opened 2026-09-03
Refuse NUL bytes in pathsec.validate_path; add a pathsec attack-candidate harness

alvations:
## What

Two pathsec changes, split out of #3753 so they can land independently:

1. `nltk/pathsec.py` — `validate_path` now refuses an embedded **NUL byte** explicitly and unconditionally.
2. `nltk/test/unit/test_attack_pathsec_candidates_expanded.py` — a new pathsec attack-candidate harness (37 tests).

## Why (the NUL fix)

`validate_path` only caught a NUL byte inside the `except` branch that runs when `Path(raw).resolve()` raises. Whether `resolve()` raises on a NUL is **platform and version dependent** — some Windows/CPython combinations (observed on 3.13/3.14) do **not** raise. When it doesn't, the NUL check was skipped and a `root/ok.zip/a\x00b`-style path reached the containment check, where the `.zip` prefix resolved inside a data root and the path was **allowed**. A NUL truncates a path in every C filesystem API, so the tool could then open a different (truncated) path than the one the guard checked (**CWE-158**).

The fix rejects any NUL byte at the top of `validate_path`, before `resolve()` is consulted, closing the bypass on every platform.

## The harness

`test_attack_pathsec_candidates_expanded.py` drives the **real** guard (no mocks), with benign controls alongside hostile inputs:

- path traversal: symlink-chain escapes, TOCTOU, absolute `/proc` and `/etc` reads, blank/control-only paths, the NUL-in-zip-virtual-path case;
- archives: zip-slip member names, symlink-typed members;
- SSRF: redirect re-validation across `file://`/`gopher://`, hex/octal loopback, IPv4-mapped loopback, and decimal cloud-metadata targets.

## Testing

`python -m pytest nltk/test/unit/test_attack_pathsec_candidates_expanded.py` — **37 passed** on the current `develop` base (`validate_path` / `validate_tool_path` / `make_staging_dir` are already present on develop, so the change is self-contained).

### DIFF
--- a/nltk/pathsec.py
+++ b/nltk/pathsec.py
@@ -170,6 +170,19 @@ def validate_path(path_input, context="NLTK", required_root=None):
     try:
         raw = path_input.path if hasattr(path_input, "path") else str(path_input)
 
+        # A NUL byte truncates a path in every C filesystem API and can never be
+        # legitimate; refuse it explicitly (Path.resolve() surfaces it only on some
+        # platforms/versions, so the except-branch check below could be skipped).
+        if "\x00" in raw:
+            msg = (
+                f"Security Violation [{context}]: NUL byte in path is not "
+                f"allowed: {raw!r}"
+            )
+            if ENFORCE:
+                raise PermissionError(msg)
+            warnings.warn(msg, RuntimeWarning, stacklevel=3)
+            return
+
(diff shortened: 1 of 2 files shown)


alvations:
Closing tail gaps on patchsec exploits. Merging this in.

PR #3853, opened 2026-09-03
Re-check the normalized form of a no-protocol resource path against traversal (CWE-22)

alvations:
## What

Peeled out of #3753: a no-protocol traversal hardening in `nltk/data.py` plus its test.

- `nltk/data.py` — `_reject_unsafe_no_protocol` now also calls `_assert_no_normalized_bypass` (a comment on the adjacent JSON-load bound is also tightened).
- `nltk/test/unit/test_data_no_protocol_traversal.py` — new, focused test.

## Why

`_reject_unsafe_no_protocol` screens a no-protocol resource path for a `..` traversal using a literal deny-list that requires the two dots to be **adjacent**. But `url2pathname` (applied when the path is later resolved) strips `TAB` / `LF` / `CR`, so a control character spliced between the dots — `foo/.\n./x` — is dropped afterwards and reconstitutes `..`, diverging from what the guard actually screened (**CWE-22**).

`find()` already defends against this by re-checking the **normalized** form. This applies the same `_assert_no_normalized_bypass` re-check on the no-protocol path so it cannot diverge. `_assert_no_normalized_bypass` is already defined on develop, so the change is self-contained.

## Testing

`python -m pytest nltk/test/unit/test_data_no_protocol_traversal.py` — **4 passed** on the current `develop` base: a `TAB`/`LF`/`CR`-split `..` is rejected, and a legitimate resource path (`corpora/brown/ca01`) still passes. On develop without this change the split-traversal baits are **not** rejected, so the test both proves the fix and guards against regression.

### DIFF
--- a/nltk/data.py
+++ b/nltk/data.py
@@ -97,31 +97,50 @@ def _assert_no_encoded_bypass(name, error_label=None):
 
 
 # Python 3.14's url2pathname follows the WHATWG URL rules, so it silently strips
-# ASCII tab / LF / CR and truncates at "#" or "?". Earlier versions keep them.
-_URL_REWRITTEN_CHARS_RE = redos.compile(r"[\t\n\r#?]")
+# ASCII tab / LF / CR and truncates at "#" or "?". Refuse the whole control-char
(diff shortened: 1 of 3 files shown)


alvations:
Tighten path traversal from data.py. Merging this in.

PR #3856, opened 2026-09-03
Escape corpus-derived text in the WordNet browser app (CWE-79 XSS)

alvations:
## What

Escape corpus-derived text before it is served as HTML in the WordNet browser app (`nltk.app.wordnet_app`), plus a test that proves it.

## Why

`wordnet_app` renders WordNet glosses, examples, lemma names, and lookup-link hrefs straight into HTML. A crafted WordNet entry (or a lemma routed through the lookup link) containing `<script>` / `` / attribute-breakout payloads was emitted verbatim, so it executed in the browser (**CWE-79**, stored/reflected XSS). One sink also declared a local variable named `html`, which shadowed the module-level `import html`; because a nested closure referenced `html.escape(...)`, that escape resolved to the wrong (unassigned) name.

Fix:
- `html.escape(...)` every corpus/user-derived value at each sink: `synset.definition()`, `synset.examples()`, lemma names, related-lemma link text, and the lookup-link `href` (`quote=True`).
- Rename the shadowing local (`html` → `relations_html`) so the module-level `html.escape` is reachable.

## Testing

`nltk/test/unit/test_attack_wordnet_xss_csrf_expanded.py` (new) drives each sink with real XSS payloads and asserts no live markup survives. **28 pass** with the fix; **without it, 12 fail** (the payloads leak), so the test both proves the fix and guards against regression. `black` / `isort` clean.

### DIFF
--- a/nltk/app/wordnet_app.py
+++ b/nltk/app/wordnet_app.py
@@ -138,7 +138,7 @@ class MyServerHandler(BaseHTTPRequestHandler):
                         page = infile.read()
                 else:
                     page = (
-                        (html_header % word) + "<p>The database info file:"
+                        (html_header % html.escape(word)) + "<p>The database info file:"
                         "<p><b>"
                         + usp
                         + "</b>"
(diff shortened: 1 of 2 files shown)


alvations:
Blocking XSS attacks from wordnet_app. Merging this in first.

Issue #125, opened 2012-01-17
Snowball stemmer for Russian muches apostrophes


alexrudnick:
Version: 2.0b9
To reproduce:
&gt;&gt;&gt; print stm.stem(u&quot;Кот-д'Ивуару&quot;)
Output: кот-дьивуар
Notice the apostrophe being turned into Ь.
It shouldn't.

<p>Migrated from http://code.google.com/p/nltk/issues/detail?id=593</p><hr/><h2>earlier comments</h2>
<p><strong>PeMiStahl said, at 2010-09-12T16:30:37.000Z:</strong></p>

<p>Hi,

I'm the one who ported the Snowball stemmers to NLTK. Therefore, I exactly followed the descriptions given by Martin Porter on his website. You can find the description of the Russian stemmer here: http://snowball.tartarus.org/algorithms/russian/stemmer.html

The apostrophe is turned into Ь because, before the actual stemming process starts, every Cyrillic letter is turned into its Roman transliterated form as stated by Mr Porter. After stemming, the transliterated form is turned back into the Cyrillic equivalent.

The problem now is the following: The stemming algorithm defined by Mr Porter assumes that an apostrophe may only appear in the Roman transliteration of word, but not in the original Cyrillic form. As I do not speak any Russian myself, I do not know whether an apostrophe can appear in a Russian word or not.

So, if you can tell me which positions an apostrophe can appear at in a Russian word, I might be able to adjust my code accordingly.

Thanks,
Peter</p><p><strong>gmal.rom said, at 2010-09-18T00:29:00.000Z:</strong></p>

<p>As it can appear only in loanwords, it's safely to assume it can appear within the root, not in first/last position and not in suffixes or flexions.</p><p><strong>PeMiStahl said, at 2010-10-28T10:49:26.000Z:</strong></p>

<p>Hi everyone,

I apologize for posting this so lately, but I moved to a new town and started my Master studies. Therefore, I just did not find the time to deal with this issue. Now I can provide a fix to the apostrophe problem. It was actually very easy to solve because I just had to add some new replace rules to the transliteration methods. The apostrophe is no longer turned into another character but stays the same during the stemming process.

Best,
Peter</p>

alexrudnick:
Could somebody check what other versions of the Snowball stemmer do? And even better, could we get commentary from a Russian speaker?

kmike:
Russian speaker here: the apostrophe shouldn't being turned into Ь. 

[...]

PR #3857, opened 2026-09-03
Fix 8 RussianStemmer divergences from the Snowball reference vocabulary

dylanpulver:
`RussianStemmer` disagrees with the Snowball reference on 8 of the 49,785 words in the official `snowball-data` Russian vocabulary. Three causes, all in the transliterated-ASCII layer:

**1. Step 4 ignored RV.** The reference runs the backward section inside `backwards setlimit tomark pV`, so `tidy_up` applies only within RV. NLTK tested the whole word: `злейший` → `зл` (ref `злейш`), `ль` → `л`, `сь` → `с`.

**2. ъ read as ь.** ъ transliterates to `''` and ь to `'`, so `word.endswith("'")` also matched a hard sign: `объем` → `обь` (ref `объ`); same for `подъем`, `подъём`.

**3. A suffix matched across one letter.** ю is `i^u`, so suffix `uet` (ует) matched the tail of `i^uet` (юет): `воюет` → `вои^`, leaking a bare `^` into the output. Also `горюй` → `гори^` (ref `горю`).

After: **0 / 49,785**. Every other language's count is unchanged (danish 35, english 57, finnish 183, … identical before and after).

Mutants, same command each run: full revert → 8; only (2) → 5; only (1) → 5; (1)+(2) without (3) → 2. Each part is load-bearing.

An earlier version of this description quoted a test count I cannot reproduce, so here is the check that actually matters. Against the official snowball-data Russian vocabulary, 49,785 words: this branch produces **0 mismatches** against the reference output, while reverting `snowball.py` produces **8** (`подъем`, `объем`, `сь`, and five more). The stem doctests that fail here fail identically on `develop` and need corpora this checkout lacks.

Related background: in #125 Martin Porter confirms the transliteration is expository and the compiled stemmer works on Cyrillic directly.

Written with AI assistance (claude-opus-5); I reviewed and verified the change and the numbers above.


nltk/stem/snowball.py
```diff
@@ -5147,29 +5164,39 @@ def stem(self, word):
...
-            if word.endswith("nn"):
+            if rv.endswith("nn"):
                 word = word[:-1]
+                rv = rv[:-1]
 
         if not undouble_success and not superlative_removed:
-            if word.endswith("'"):
+            # "'" transliterates ь, but ъ transliterates to "''", which also
+            # ends in "'": only a lone apostrophe is a soft sign.
+            if rv.endswith("'") and not rv.endswith("''"):
```
alvations, line 5199: Please make sure `ъ` is checked for Bulgarian too.

PR #3860, opened 2026-09-04
Add test-only regression coverage for already-fixed CWEs

alvations:
## Summary

Two security regression suites that pin behaviour **already correct on `develop`** but previously untested. They are **test-only** — no source changes accompany them, and all 20 tests pass against `develop` as-is.

These were originally written alongside other hardening work but do not depend on any of it, so they belong on `develop` independently rather than gated behind a larger review.

## What's covered

- **`test_fixed_but_untested_cwe.py`** (7 tests) — asserts a set of CWE fixes already present on `develop` stay fixed (path-traversal / file-IO guards, etc.).
- **`test_regex_injection.py`** (13 tests) — pins that user-influenced strings spliced into a regex are treated as **literals** (`re.escape`) rather than patterns: the MTE tag filter, the chat80 relation-name path, and the no-protocol data loader. A relation name of `(` (and other regex metacharacters) must not raise `re.error`, and a legitimate relation still reads its records.

## Notes

- The chat80 cases exercise the real `cities.pl` corpus file. (They previously referenced a non-existent `city.pl`, so they silently skipped with a misleading "corpus data not installed" message and never actually ran; corrected here so they execute.)
- No changes to any non-test module; `git diff` is limited to `nltk/test/unit/`.

### DIFF
--- a/nltk/test/unit/test_fixed_but_untested_cwe.py
+++ b/nltk/test/unit/test_fixed_but_untested_cwe.py
@@ -1,0 +1,81 @@
+# Natural Language Toolkit: regression tests for fixed-but-unprobed weaknesses
+#
+# Copyright (C) 2001-2026 NLTK Project
+# URL: <https://www.nltk.org/>
+# For license information, see LICENSE.TXT
+
+"""Adversarial tests for hardening that was applied but never had a test driving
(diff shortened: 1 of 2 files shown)


alvations:
Adding more tests to make sure nothing leaks through with new features. Merging this in.

PR #3861, opened 2026-09-04
Extend the no-unsandboxed-open guard to app/draw/twitter and route every open through pathsec.open

alvations:
## Summary

The `no-unsandboxed-open` guard (#3740) only covered a dozen named modules, so a bare builtin `open()` anywhere else went unnoticed. The GUI apps, the drawing canvas and the twitter client all take a filename from the **operator** (a Tk dialog, a config path, an output name) and opened it with a bare `open()`. This extends the guard to those three subtrees and routes **every** open in them through `pathsec.open` — no `# sandboxed-open ok` exceptions.

## Are operator-chosen paths actually protected?

Yes. These paths are operator-chosen, so `pathsec.open` (default `required_root=None`) does not bound them to the NLTK data roots — the operator still saves where they want and keeps the file. But on POSIX it still routes through the hardened opener, which was verified adversarially against the actual wrappers:

- a **symlink** planted at the save/open destination is refused at open time (`O_NOFOLLOW`, TOCTOU-safe) — a local attacker cannot redirect a Save onto `/etc/...` or read a file out through a planted link (CWE-59);
- a **hardlink** to an outside inode is refused (`st_nlink > 1`, CWE-59);
- a **NUL byte** or a **URL** in the path is refused (CWE-22);
- a freshly written file is created **0600**, so a saved chart or a `credentials.txt` is never left group/world readable (CWE-377/378).

Legitimate operator paths, and a real chart pickle save/load roundtrip, keep working.

## Changes

- **`tools/check_no_unsandboxed_open.py`** — add `nltk/app`, `nltk/draw`, `nltk/twitter` to `GUARDED_PATHS`. The guard still flags only the builtin `open()` via AST; `pathsec.open` / `gzip.open` / `os.open` are unaffected.
- **All opens in the three subtrees** — reads *and* writes (Save Chart / Grammar / History, PostScript export, wordnet log, credentials read/append, tweet-collection output) now go through `pathsec.open`. For the gzip output the raw handle is sandboxed, then `gzip` wraps the checked file.
- **`nltk/twitter/util.py`** — on an auth failure, report only credential **key names**, never the secret values (CWE-532 / 209 / 200); pin the creds-file charset.
- **`test_output_exposure_security.py`** — an auth error never leaks a secret; the wordnet lookup link escapes its label/href.
- **`test_app_twitter_open_hardening.py`** (new) — the adversarial harness described above, run against the real `_outf_writer`, `TweetWriter`, `ChartComparer.load_chart` and `ChartParserApp.save_chart`.

## On pickles / JSON

Loading an untrusted pickle/JSON is the real deserialization risk (CWE-502) and stays on the **read** side, unchanged: chartparser loads through `AllowlistUnpickler`, twitter reads JSON through `safe_json_loads`. The reads now additionally get path validation.

## Verification

- Extended guard passes; every open in the three subtrees goes through `pathsec.open`.
- New adversarial harness (22) + `test_output_exposure_security` (9) + `test_json2csv_corpus` (12) pass; the broader path/data security suites (192) pass.
- All 300 nltk modules import cleanly.

### DIFF
--- a/nltk/app/chartparser_app.py
+++ b/nltk/app/chartparser_app.py
@@ -76,32 +76,14 @@ from nltk.parse.chart import (
     TopDownPredictRule,
     TreeEdge,
 )
+from nltk.pathsec import open as pathsec_open
 from nltk.picklesec import AllowlistUnpickler, pickle_dump
 from nltk.tree import Tree
 from nltk.util import in_idle
 
-# A saved chart / grammar ``.pickle`` is loaded from an operator-chosen path via a
-# Tk "Open" dialog, but a ``.pickle`` a user was emailed and opens is a realistic
-# phishing-style RCE, and every other model loader in the tree was hardened, so
-# these loads are routed through an allowlisting unpickler (CWE-502) rather than
-# the warn-only ``pickle_load`` (which warns and then EXECUTES any reduce gadget).
-# Only the exact classes a legitimate chart / grammar / parse tree reconstructs
-# may be built; any other global (``os.system``, ``builtins.eval``,
-# ``subprocess.Popen``, a scientific-stack file sink, ...) raises
-# ``UnpicklingError`` instead of executing. The base
-# :class:`~nltk.picklesec.AllowlistUnpickler` still enforces its denied-module /
-# dotted-name / dunder-name / extension-opcode guards on top.
-#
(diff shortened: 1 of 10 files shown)


alvations:
From #3753. Route all the other opens in the NLTK through pathsec so now we should have a CI guard to check for bare open and raise issue when it we see them. Merging this in.

PR #3862, opened 2026-09-04
Add the all-regex-through-redos pre-commit guard and two regression tests

alvations:
## Summary

Two develop-independent pieces split out of the GHSA-8mgp work, each verified to pass on `develop` **with no runtime code changes**.

### 1. `all-regex-through-redos` pre-commit guard

Every regex runs over potentially attacker-influenceable input, so a bare `re` / `regex` compile or match is a latent ReDoS / compile-time DoS (CWE-1333 / CWE-400). `develop` already routes its regexes through `nltk.redos`; this adds a guard so it cannot silently regress:

- `tools/check_all_regex_through_redos.py` — fails on any `re`/`regex` pattern call outside `nltk.redos` itself.
- `.pre-commit-config.yaml` — an `always_run` hook that runs it every commit.

The guard **passes on `develop` as-is** (`pre-commit run all-regex-through-redos --all-files` → Passed).

### 2. Two regression tests

Both pin behaviour already correct on `develop` and pass unchanged:

- `test_attack_java_tool_expanded.py` — the java external-tool wrapper rejects per-call option/path injection.
- `test_featstruct_redos.py` — the `FeatStructReader` parsing regexes are bounded against ReDoS.

## Verification

- `pre-commit run all-regex-through-redos --all-files` passes on `develop`.
- The two test files pass on `develop` (51 passed, 1 skipped).
- No changes to any runtime `nltk` module; the diff is a tool script, a pre-commit hook, and two tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,17 @@ repos:
       language: system
       pass_filenames: false
       always_run: true
+    - id: all-regex-through-redos
+      name: every regex routes through nltk.redos
+      description: >-
+        Every regex runs over attacker-influenceable input, so a bare
+        re/regex compile or match is a potential ReDoS / compile-time DoS
+        (CWE-1333 / CWE-400). Fails on any re/regex pattern call outside
+        nltk.redos itself; use redos.compile or the redos re-compatible helpers.
+      entry: python tools/check_all_regex_through_redos.py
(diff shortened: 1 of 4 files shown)


alvations:
Add more redos and jvm and guard them through CI checks. Merging this in.

PR #3863, opened 2026-09-04
Bound the WekaClassifier model path to the pathsec data roots

alvations:
> **[WIP]** — split out of the larger GHSA-8mgp hardening for focused review.

## Summary

Hardens `WekaClassifier` — a Python wrapper that drives the weka JVM and writes ARFF the JVM reads — against the known JVM-wrapper vulnerability classes, and adds an unmocked adversarial harness for the whole surface.

## 1. Model-path containment (GHSA-j456-xh4h-cpf2)

`WekaClassifier` passed a caller-controlled model filename straight to the JVM (`-l` read, `-d` write) with no containment check. `validate_tool_path` now bounds it at `__init__`, `_classify_many` (re-checked in case `_model` was reassigned), and `train` (`for_write=True`). `must_exist=False` — containment, not existence, is the property. Out-of-root, symlink-escaping and hostile paths are refused before `config_weka()`/`java()`.

## 2. JVM / ARFF surface audit + fixes

I audited every way attacker-influenced data reaches the JVM or the ARFF, against the ecosystem's JVM-wrapper CVEs:

| Vector | Status |
|---|---|
| JVM arg injection via options / model path (`@argfile`, `-javaagent`, CVE-2026-12841 / CWE-88) | already refused by `internals.java` (rejects any `@argfile` token in the command) |
| Untrusted classpath jar / CWD `weka.jar` (CVE-2026-0848, CVE-2026-12252 class; CWE-426/427/494) | already refused (`internals.java` classpath trust; CWD not in `_weka_search`) |
| Out-of-root `WEKAHOME` (CWE-426) | already dropped with a warning |
| Decompression bomb in the jar's `version.txt` (CWE-409) | already capped — `_check_weka_version` reads through the pathsec `SecureZipFile`, whose streaming reader enforces `MAX_UNZIP_RATIO` / `MAX_UNZIP_ACTIVATION`; **now pinned by a test** |
| Classifier-class injection | already allowlisted (`_CLASSIFIER_CLASS`) |
| ARFF **label** injection (CWE-1236) | already sanitised (`_sanitize_arff_label`) |
| ARFF feature **type** injection | **fixed here** — a caller-supplied `ftype` was interpolated raw, so a newline injected a new `@ATTRIBUTE`/`@DATA` line; `_check_arff_ftype` now refuses a control char in a type |
| ARFF **value / feature-name** type-confusion (CWE-1236) | **fixed here** — the formatter trusted the caller's `__str__`/`__repr__`, so an int/float **subclass** with a hostile `__str__`/`__repr__`, a `str` subclass overriding `__repr__`, or an arbitrary object handed to a directly-built formatter could emit a raw newline and forge a directive. Values are now formatted from their numeric value (`int()`/`float()` strip the subclass) and every string-like token goes through `_safe_str_repr` (the built-in `str` repr, which always escapes control chars) |

These value/name vectors need object-type control, so they are defence in depth — but the formatter's contract is to emit safe ARFF for **any** input, so it must never interpolate an untrusted `__repr__`. Output for ordinary `bool`/`int`/`float`/`str` values is byte-identical.

## 3. Robustness fixes on untrusted paths

- `from_train` raised an opaque `UnboundLocalError` (it formatted the undefined `ftype`) when handed an unsupported value type; it now raises a clear `ValueError` naming the type.
- `parse_weka_output` raised a raw `IndexError` on truncated/malformed weka `-p` stdout — which is only as trustworthy as the model weka just loaded — instead of a clear error; `_weka_predicted_class` now validates the line shape and raises a descriptive `ValueError`, and an empty output block is rejected explicitly.

### ARFF write is confined by design

`ARFF_Formatter.write(path, ...)` routes the path through the pathsec sandbox, so the scratch ARFF weka reads stays inside the data roots (it is written under a private staging dir). A caller who genuinely wants to export ARFF to an operator-chosen path outside the roots passes a **file object** instead of a path — that bypasses the path open and is not confined. This is deliberate; `test_weka_save_security.py` pins it, so please do not "relax" the write to a bare `open()`.

## Tests

- `test_weka_model_path_security.py` — model-path containment (unmocked).
- `test_weka_jvm_arff_hardening.py` (new) — @argfile in options, untrusted classpath, out-of-root WEKAHOME, CWD exclusion, `version.txt` decompression bomb (CWE-409), classifier allowlist, ARFF label/value/type injection, numeric-subclass and str-subclass and arbitrary-object value/name injection, `from_train` non-scalar rejection, and `parse_weka_output` malformed-line rejection + happy-path parse, plus legitimate-formatting round-trips. The real `internals.java` / `pathsec` / ARFF guards fire; `config_weka` is stubbed only so the flow reaches `internals.java` without a local weka install.
- `test_weka_security.py` / `test_weka_save_security.py` / `test_weka_arff_injection.py` — CWD search-path exclusion, WEKAHOME containment, scratch-ARFF-stays-in-root, ARFF directive-injection, write confinement.

## Residual (documented, out of Python's reach)

weka's `-l` **deserializes a Java-serialized model**, so a malicious model *already inside a data root* remains a Java-side deserialization risk (CWE-502). Path containment reduces the attack surface (an attacker needs write access to a data root) but does not make loading an attacker-authored in-root model safe. Jar integrity beyond path trust (a trusted-path but tampered `weka.jar`) is likewise out of scope for this layer.

## Verification

- Full weka suite: **77 passed, 2 skipped** (real-weka integration + a macOS temp-authorized case skip).
- All nltk modules import with 0 hard failures; real `from_train` → ARFF formatting round-trips (in-root path write and operator file-object write both work); pre-commit clean (sandbox-open / redos / picklesec guards included).

### DIFF
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -369,9 +369,11 @@ class ARFF_Formatter:
         # Relation name
         s += "@RELATION rel\n\n"
 
-        # Input attribute specifications
+        # Input attribute specifications. fname is repr'd (a control char in it is
+        # escaped), but ftype is interpolated raw, so a newline in a caller-supplied
+        # type would inject a new @ATTRIBUTE/@DATA line (CWE-1236); refuse it.
         for fname, ftype in self._features:
-            s += "@ATTRIBUTE %-30r %s\n" % (fname, ftype)
+            s += "@ATTRIBUTE %-30r %s\n" % (fname, self._check_arff_ftype(ftype))
 
         # Label attribute specification – labels are already sanitized.
         # Wrap each label in single quotes and join with commas.
@@ -419,6 +421,23 @@ class ARFF_Formatter:
             return "%r" % fval
         else:
             return "%r" % fval
+
+    @staticmethod
+    def _check_arff_ftype(ftype):
+        """Refuse an ARFF attribute type carrying a control character.
+
+        ``from_train`` only ever emits the fixed enum (NUMERIC / STRING /
+        ``{True, False}``), but a caller may build an ``ARFF_Formatter`` directly;
+        a newline or other C0 control in the type would break out of the
+        ``@ATTRIBUTE`` line and inject a new attribute or ``@DATA`` section
+        (CWE-1236). Any legitimate ARFF type is control-character free.
+        """
+        text = str(ftype)
+        if any(ord(ch) < 0x20 for ch in text):
+            raise ValueError(
+                f"ARFF attribute type must not contain control characters: {ftype!r}"
+            )
+        return text
 
     @staticmethod
     def _sanitize_arff_label(label):
--- a/nltk/classify/weka.py
+++ b/nltk/classify/weka.py
@@ -22,7 +22,7 @@ from nltk.data import make_staging_dir
 from nltk.internals import config_java, java
 from nltk.pathsec import ZipFile as SecureZipFile
 from nltk.pathsec import open as pathsec_open
-from nltk.pathsec import validate_path
+from nltk.pathsec import validate_path, validate_tool_path
 from nltk.probability import DictionaryProbDist
 
 _weka_classpath = None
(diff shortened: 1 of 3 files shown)


alvations:
Bounding Weka through pathsec's validate_tool_path. And more importantly add all the other adversial tests to check that we've closed most of the gaps through weka's wrapper. Merging this in.

PR #3864, opened 2026-09-04
Exclude pytest's ephemeral basetemp from the pickle-asset scanner

alvations:
## Problem

Every `ci-workflow` matrix job on `develop` fails at "Run pytest": `test_pickle_allowlist_security.py::test_all_nltk_data_pickle_assets_load`.

That test is a real-asset regression: it globs `**/*.pickle` under every `nltk.data.path` root and asserts each loads through `nltk.data.load`. But `nltk/test/conftest.py::_authorize_pytest_basetemp` **registers pytest's session basetemp as an `nltk.data.path` root** (so staging tests can write inside the pathsec sandbox). So the scanner also scans that ephemeral base and loads the scratch artifacts other tests leave there — e.g. `test_app_twitter_open_hardening.py` (added in #3861) writes `chart.pickle` symlinks (pathsec refuses them at open, CWE-59) and a real `Chart` pickle (the restricted loader correctly forbids `nltk.parse.chart.Chart`). Those are **test artifacts, not real assets**, so the scan fails.

## Fix

Exclude the pytest basetemp root (and anything under it) from the scanned roots, **by location, not by filename**. A filename-based skip (e.g. renaming the test's `.pickle` files) is fragile — any other test, or any other extension, re-breaks it, and the scanner's job is to validate *installed assets*, not per-test tmp files. Real asset roots are never under the pytest base, so their coverage is unchanged.

## Verification

- `test_pickle_allowlist_security.py` + `test_app_twitter_open_hardening.py` together: 159 passed, 2 skipped, **0 failed** (reproduced the failure first, then confirmed the fix — with the test keeping its natural `.pickle` names).
- The scanner still runs against real assets (not skipped): 1 passed.
- One file changed: `nltk/test/unit/test_pickle_allowlist_security.py`.

### DIFF
--- a/nltk/test/unit/test_pickle_allowlist_security.py
+++ b/nltk/test/unit/test_pickle_allowlist_security.py
@@ -1473,7 +1473,7 @@ def test_data_load_restricted_pickle_roundtrips_real_asset(tmp_path, monkeypatch
     assert loaded == staged_value, "staged real .pickle failed to load via data.load"
 
 
-def test_all_nltk_data_pickle_assets_load():
+def test_all_nltk_data_pickle_assets_load(tmp_path_factory):
     """Real-asset regression: EVERY ``*.pickle`` present in the installed nltk_data
     must still load through ``nltk.data.load`` after the picklesec hardening (class-
     bearing artifacts like punkt / perceptron / maxent via the pickle-free redirect;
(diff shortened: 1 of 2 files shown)


alvations:
Fixing the CI/CD with this revision and further guards for the pickle temp checks. Merging this in first.

PR #3865, opened 2026-09-04
Open operator-chosen app/twitter paths with an annotated builtin open, not pathsec.open

alvations:
## Problem (develop CI regression)

develop's ci-workflow fails on exactly **1 test** — `test_output_exposure_security.py::TestTwitterCredsNoLeak` — with:
```
PermissionError: Security Violation [twitter.Authenticate]: Unauthorized path .../credentials.txt
```
`pathsec.open` confines every path to the NLTK **data roots** (unconditionally — even `required_root=None`). #3861 routed the app/draw/twitter **operator-chosen** opens through it, but a credentials file, a Tk *Save/Open* destination, a tweet-output dir, or a PostScript export legitimately lives **outside** the data roots, so `pathsec.open` rejects them. The twitter creds path is the only one with a test that loads a real path; the rest (chart Save/Open, tweet output, json2csv, wordnet log/db-info) are the **same latent break** (verified: each rejects an outside-roots path).

## Fix

Revert those operator opens to a **builtin `open()`** annotated `# sandboxed-open ok:` for the no-unsandboxed-open guard (7 files). These paths are **operator-controlled, not attacker input**, so confinement is wrong for them and a plain annotated open is correct. The guard still covers the subtree; the `pathsec.open` import is removed where unused.

## Is the annotated bare open an exploit point? (verified)

- **Path source is operator-controlled**: Tk `asksaveasfilename`/`askopenfilename` (operator picks), `$TWITTER` config, operator CLI log arg.
- **The one HTTP-facing case (wordnet) is traversal-safe**: `get_static_page_by_path` raises `FileNotFoundError` (opens nothing) for **every** traversal/arbitrary path, and the db-info open is **exact-string gated**. Pinned in `test_operator_path_opens_regression.py`.
- **Tradeoff, accepted**: a bare open no longer refuses a *planted symlink* at the destination (the O_NOFOLLOW TOCTOU protection `pathsec.open` gave). These are operator-private paths, so the local symlink-swap surface is marginal. The four per-wrapper symlink-refusal tests are dropped; `pathsec.open`'s own write-side hardening (symlink/hardlink/NUL/URL/0600) is still tested.

## Tests

- `test_operator_path_opens_regression.py` (new): wordnet traversal-safety (attack) + functional round-trips proving each operator open **works** outside the data roots (so re-confining fails a test).
- `test_app_twitter_open_hardening.py`: refit to pin `pathsec.open`'s own hardening.

## Verification
- `test_output_exposure_security.py` (9), `test_json2csv_corpus.py` (12), `test_pickle_allowlist_security.py`, `test_app_twitter_open_hardening.py`, `test_operator_path_opens_regression.py` — all pass (188 passed, 6 skipped, 0 failed).
- `check_no_unsandboxed_open.py`: OK. All 300 nltk modules import. pre-commit clean.

### DIFF
--- a/nltk/app/chartparser_app.py
+++ b/nltk/app/chartparser_app.py
@@ -76,7 +76,6 @@ from nltk.parse.chart import (
     TopDownPredictRule,
     TreeEdge,
 )
-from nltk.pathsec import open as pathsec_open
 from nltk.picklesec import AllowlistUnpickler, pickle_dump
 from nltk.tree import Tree
 from nltk.util import in_idle
@@ -855,9 +854,9 @@ class ChartComparer:
         if not filename:
             return
         try:
-            with pathsec_open(
-                filename, "wb", context="ChartComparer.save_chart_dialog"
-            ) as outfile:
+            with open(
+                filename, "wb"
+            ) as outfile:  # sandboxed-open ok: operator-chosen GUI file path
                 pickle_dump(self._out_chart, outfile)
         except Exception as e:
             showerror("Error Saving Chart", f"Unable to open file: {filename!r}\n{e}")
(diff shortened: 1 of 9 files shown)


alvations:
Undoing the #3861 so that the CI/CD passes. The gate through pathspec for user-specified file/path writing cannot be easily guarded, it must be up to the user. At least it's writing not reading so that threat control is still there whenever we read files. Only merging this after the CI/CD on github passes.

alvations:
Merging this in to fix ci/cd first.

PR #3872, opened 2026-09-08
Isolate default WordNet reference relations

vzer200:
WordNet browser references created without an explicit relation map share a default dictionary. Expanding a synset in one reference changes the state and encoded links of other lookups, and toggling it in a second reference can collapse it in both.

Give each reference its own default dictionary. The tests exercise expansion before and after creating another lookup, independent toggles of the same synset, encoded-reference round trips, and explicitly supplied relations.

Addresses the `wordnet_app.Reference` case in #2121.

Validation: the regression tests failed in three cases before the fix and now pass all four. The new and existing WordNet browser tests pass all 116 cases; the additional reference/deserialization selection passes 26 with one skipped. All applicable pre-commit hooks and `git diff --check` pass. Tested on Windows with Python 3.12.14.

The full suite requiring every external corpus/tool was not run; validation covers the affected module and its existing browser/deserialization checks.

AI-assisted implementation, independently reviewed by a second coding agent.


alvations:
Thanks for the suggestion on replacing the default None with dict(). Please do note the following esp. if an agentic coding help was used.

ekaf (approved):
This PR fixes a mutable default argument bug in nltk.app.wordnet_app.Reference and adds unit tests to ensure synset_relations are properly isolated between instances. It also updates AUTHORS.md to include the contributor.

**Changes**

- nltk/app/wordnet_app.py
    Replaces synset_relations=dict() with synset_relations=None and initializes an empty dict inside __init__ to avoid sharing the same dict across all instances.
- nltk/test/unit/test_wordnet_app.py
    Adds three test cases covering:
  · Independence of Reference instances (expanding one does not affect another).
  · Independence of toggling the same synset on different instances.
  · Preservation of user‑provided relations and correct merging with new ones.
- AUTHORS.md
    Adds vzer200 as a contributor.

**Strengths**

1. Correct fix – The change addresses the classic pitfall of mutable defaults and correctly prevents accidental sharing.
2. Good test coverage – The new tests exercise the exact scenario that would have failed before, ensuring the fix works and remains robust.
3. Backward compatible – Existing code that passes no synset_relations or passes a dict continues to work as before (with correct behaviour).
4. Clean code – The implementation is straightforward and follows Python best practices.

**Conclusion**

This is a high‑quality, low‑risk PR that fixes a real bug and adds appropriate tests. It is ready to merge.

ekaf:
> include the prompt and response and audit trail of all the reasoning of how you arrive at the implementation

@alvations, your suggestion above would be instructive. However, I would lean towards making it optional, since everybody uses vibe coding these days, without necessarily acknowledging it. And a full reasoning trail remains to be seen anywhere.

alvations:
Agree that it be optional but we should at least elicit the agent to respond or human to follow up.

I think using agentic code is okay (I do that too) but turning on open claw and then spamming across libraries without the PR creator really understanding what the change does, might not be sustainable long term. So best to at least have the open claw reply before considering to merge. My fear is more of supply-chain like attacks from open claw eventually if the human is not behind the PRs =)

alvations:
Please do feel free to merge if any of our contributors / maintainers did review or agree with the agentic reviews. But I think we should still elicit reply as much as we can to make sure the human launching the agents know what they're changing, just in case we need to patch/fix/follow up on it.

ekaf:
Lately, many PR authors are defining themselves as "_an open-source enthusiast and a passionate programmer_", and don't leave a convincing impression that their Github account is managed by a human. 
Personally I am not concerned about where the code comes from, as long as it LGTM. But there is a security issue: after their first contribution, an account gains free access to running CI, which is a potential liability. That's a repo configuration option that we could consider changing. 
The massive slowdowns on Github could likely be caused by robotic activity. So @vzer200, could you please reassure us about your intentions with this contribution?

vzer200:
Thanks for asking, and sorry for the delayed reply. I’m looking to contribute to open source by fixing existing issues and build a track record of useful contributions. That’s what motivated this PR. I’m not seeking any additional access or privileges, and I’m happy for CI runs to remain subject to maintainer approval.

PR #3873, opened 2026-09-08
Avoid projecting cluster assignment vectors twice

### LINKED ISSUE #2339: svd_dimensions in VectorSpaceClusterer is broken again.
So I would like to try using svd_dimensions in KMeansCluster. So you set it to what you want when you instantiate the class. Then, when you run kclusterer.cluster, it's supposed to pick up that setting and apply it to your input. 

But then it complains about mismatched dimensions, so it must not be doing that part correctly. 

Exception has occurred: ValueError
shapes (500,1124) and (500,) not aligned: 1124 (dim 1) != 500 (dim 0)

Any idea what's going on? I have 1124 dimensions, so SVD would be helpful. 

Thanks

vzer200:
With SVD enabled, `VectorSpaceClusterer.cluster(..., assign_clusters=True)` passes already-reduced vectors to `classify()`, which applies the projection again and raises a dimension mismatch. This affects both KMeans and EM clustering.

Retain the original input vectors for label assignment so each `classify()` call performs preprocessing once. This preserves the existing classification entry point and leaves fitting unchanged.

Fixes #2339.

Validation: 26 cluster tests pass on Windows with Python 3.12.14 and NumPy 2.5.3; the same regression matrix has 8 failures on the original implementation. Tests cover SVD sizes 1 and 2, normalization, optional label assignment, consistent later classification, unchanged inputs, and GAA assignment without SVD. All applicable pre-commit hooks and `git diff --check` pass. The full NLTK corpus/external-tool suite was not run.

AI-assisted implementation, independently reviewed by a second coding agent.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -319,6 +319,7 @@
 - Nicola <https://github.com/trinik15>
 - medisean <https://github.com/medisean>
 - tandede <https://github.com/tandede>
+- vzer200 <https://github.com/vzer200>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/cluster/util.py
+++ b/nltk/cluster/util.py
@@ -39,6 +39,7 @@ class VectorSpaceClusterer(ClusterI):
 
     def cluster(self, vectors, assign_clusters=False, trace=False):
         assert len(vectors) > 0
+        original_vectors = vectors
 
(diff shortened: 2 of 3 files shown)


alvations:
Thanks for the suggestion on checking the original vector before returning. Please do note the following esp. if an agentic coding help was used.

PR #3874, opened 2026-09-08
Include the final adjacent entity pair in relation extraction

### LINKED ISSUE #2277: If only one  relationship , nltk.sem.relextract.extract_rels() can't work
If only one  relationship appears in the sentence, nltk.sem.relextract.extract_rels() outputs an empty list. Suggest change relextract.py:

def semi_rel2reldict(pairs, window=5, trace=False):
   ... 
    **while len(pairs) > 2:  # line 184:**
        reldict = defaultdict(str)
        reldict['lcon'] = _join(pairs[0][0][-window:])
        reldict['subjclass'] = pairs[0][1].label()
        reldict['subjtext'] = _join(pairs[0][1].leaves())
        reldict['subjsym'] = list2sym(pairs[0][1].leaves())
        reldict['filler'] = _join(pairs[1][0])
        reldict['untagged_filler'] = _join(pairs[1][0], untag=True)
        reldict['objclass'] = pairs[1][1].label()
        reldict['objtext'] = _join(pairs[1][1].leaves())
        reldict['objsym'] = list2sym(pairs[1][1].leaves())
        **reldict['rcon'] = _join(pairs[2][0][:window])   # line 195**
       ...
Suggest : 
def semi_rel2reldict(pairs, window=5, trace=False):
   ... 
 **while len(pairs) >= 2:**
        reldict = defaultdict(str)
        reldict['lcon'] = _join(pairs[0][0][-window:])
        reldict['subjclass'] = pairs[0][1].label()
        reldict['subjtext'] = _join(pairs[0][1].leaves())
        reldict['subjsym'] = list2sym(pairs[0][1].leaves())
        reldict['filler'] = _join(pairs[1][0])
        reldict['untagged_filler'] = _join(pairs[1][0], untag=True)
        reldict['objclass'] = pairs[1][1].label()
        reldict['objtext'] = _join(pairs[1][1].leaves())
        reldict['objsym'] = list2sym(pairs[1][1].leaves())
        **reldict['rcon'] = _join(pairs[2][0][:window]) if len(pairs) > 2 else ''**
       ...

vzer200:
Fixes #2277.

`semi_rel2reldict` required three remaining entity pairs even though a relation only uses the first two. A sentence containing PER Alice, "works at", and ORG Acme therefore produced no relation. Process the final adjacent pair too, with empty right context when there is no following pair. Existing relation dictionaries retain their values.

Added regressions for short inputs, context windows, tagged tokens, filtering and the ACE/CoNLL/IEER entry paths. Corrected the documentation's object from Tree3 to Tree2 and added the newly recovered Red Cross → Somalia relation to its IN example; both entities occur in the same IEER headline.

Validation:

- 14 new tests pass (8 fail on the original source); 23 affected tests pass, and all changed-file pre-commit hooks pass.
- All IEER and CoNLL2002 inputs were compared: 35,745 inputs, all 20,160 prior relation dictionaries unchanged, and 11,579 inputs gain their omitted final pair.
- Independent review ran 2,184 additional boundary cases and the corpus comparison.
- The affected doctest still has three pre-existing failures in ROLES/DE/VAN: `reharden` drops the compiled patterns' VERBOSE flag. The final baseline and patched failures match example source, expected and actual output exactly. The full NLTK suite requiring other corpora/tools was not run.

This follows the helper's existing flat pair stream. Higher-level behavior is retained: callers of `extract_rels` should supply a pattern, and IEER text/headline pairs are concatenated. Consequently, a newly recognized final relation can expose the existing `pattern=None` AttributeError, or span text/headline fields when the headline contains one entity. Those API behaviors are outside this boundary fix; trailing tokens not represented by `tree2semi_rel` also remain unavailable as right context.

AI-assisted implementation with independent agent review of the final source, boundary behavior and validation evidence before submission.


alvations:
Thanks for the suggestion onchecking the rel length. Please do note the following esp. if an agentic coding help was used.

ekaf (approved):
Great work on catching this boundary condition. The fix is solid, safely handles the IndexError risk for the right context, and the added test coverage is excellent.

Nitpick (Non-blocking):Currently, pairs = pairs[1:] reallocates the list on every iteration, which is technically $O(N^2)$. For sentence-level extraction where $N$ is small, this is completely fine. However, if this ever processes larger, document-level lists, moving to a standard index pointer for i in range(len(pairs) - 1): would bring it down to $O(N)$.Not necessary for this PR since it keeps the diff clean, but something to keep in mind.

ekaf:
I don't think we need so massive tests for such a simple fix. Our CI runtime has exploded lately, because coding agents go wildly overboard in testing. One simple test is enough.

alvations:
Agree as long as it’s not the touching the pathsec termsec json pickle no opening no reading saving or writing, nor tool or cmd calling, we should be good with light tests.

ekaf (changes requested):
Thanks for catching this boundary bug. The logic for the right-context fix is correct, but we need a couple of structural adjustments before merging to keep our performance and CI times in check:

* **Refactor to an index pointer:** Please replace the `while` loop and `pairs = pairs[1:]` slice with an index-based loop (e.g., `for i in range(len(pairs) - 1):`). Slicing reallocates the list on every iteration, creating $O(N^2)$ overhead. Updating the internal references to `pairs[i]`, `pairs[i+1]`, and `pairs[i+2]` avoids this memory cost.
* **Trim the test suite:** We are actively fighting CI runtime bloat, and 150 lines of exhaustive parameterized tests for a single boundary condition is too heavy. Please remove the test matrix and provide exactly one clean, targeted unit test that proves the final entity pair is no longer dropped.

Let me know when the refactor is up and we can get this merged.

ekaf:
It is interesting to note that both my two reviews were written by Gemini 3.1 with Extended Thinking, although they contradict each other. The reason is added context concerning CI bloat.

PR #3876, opened 2026-09-08
Don't miscount a python-crfsuite Windows failure as a pathsec over-block (tag sweep)

alvations:
## TL;DR

`test_file_sinks_do_not_over_block_contained_paths[unicode-name-CRFTagger.train]`
fails on Windows **when python-crfsuite is installed**. I traced it on a real
Windows runner: **the pathsec guard is correct and is NOT over-blocking.** The
test's own classifier was miscounting a python-crfsuite tool failure as a guard
refusal. This is a **test-only fix; `pathsec.py` is untouched** (no guard
relaxed, no exploit surface changed).

## What actually happens (verified on a Windows runner)

For the contained path `<root>/modelé_模型.json` (`must_exist=False`):

1. `validate_tool_path(...)` → **PASSES** (guard accepts the contained CJK path;
   `resolve()` + `is_relative_to(root)` are both fine).
2. `pycrfsuite.Trainer.train(...)` → **cannot create the CJK-named file on
   Windows** (the C extension mangles the path in the active code page). It either
   silently mis-writes (then `CRFTagger.set_model_file` raises `FileNotFoundError`)
   or raises `ValueError` under the full suite.
3. `_refusal` caught *any* `PermissionError`/`ValueError` as a guard refusal, so
   step 2's tool-level `ValueError` was miscounted as "the guard refused a benign
   path" and the false-positive control failed.

The main CI never sees this because it does not install python-crfsuite, so
`CRFTagger()` raises `ImportError` and the sink fails "for its own unrelated
reason" (which the control already tolerates).

## The fix

A pathsec refusal is always raised as `Security Violation [context]: ...`, so a
new `_is_guard_refusal(exc)` treats **only** exceptions carrying that marker as a
guard refusal. `_refusal` and `_refusal_for` use it, so a tool that fails for its
own reason is correctly "past guard" — exactly the case the control's docstring
says it must tolerate. Over-block detection is preserved: a real over-block still
raises a Security Violation, so the control still catches it.

## Harness expansion (probes the guard directly, no tool involved)

- `test_validate_tool_path_accepts_benign_unicode` — a contained path whose leaf
  is NFC / NFD / CJK / Cyrillic / Arabic / emoji / ZWJ / mixed must be **accepted
  and returned verbatim** (no over-block, no normalization).
- `test_validate_tool_path_refuses_unicode_disguised_attacks` — a NUL, a
  traversal, a Windows-stripped trailing dot/space, or a control character wearing
  a unicode disguise is still **refused with a Security Violation** (no leak).

## Verification

- **Windows runner, full `test_pathsec_sweep_tag.py` with crfsuite installed:
  597 passed, 0 failed** (was 1 failed before this change). ubuntu: green.
- Local: 764 passed in the sweep; the adjacent pathsec attack suites
  (`test_pathsec.py`, `test_pathsec_io_attack_matrix.py`, `test_pathsec_sweep_wrappers.py`)
  290 passed / 15 skipped; a real `CRFTagger` train + tag + reload works; all 300
  nltk modules import.
- `black` (25.11.0) / `isort` clean.

## Not in scope (noted)

python-crfsuite's inability to write a CJK-named file on Windows is an upstream
limitation of that C extension, not an NLTK bug; `CRFTagger` passes the path
through unchanged. The pathsec 8.3-short-name refusal
(`test_windows_short_name_is_refused`) is a deliberate defense and is left as-is.

### DIFF
--- a/nltk/test/unit/test_pathsec_sweep_tag.py
+++ b/nltk/test/unit/test_pathsec_sweep_tag.py
@@ -752,6 +752,15 @@ _BENIGN_UNICODE_NAMES = [
     ("emoji", "robot_\U0001f916.model"),
     ("zwj", "co‍op.model"),  # zero-width joiner inside the name
     ("mixed-space", "my café 模型.model"),
+    ("bidi-rlo", "file\u202edoc.model"),  # right-to-left override, a valid char
+    ("fullwidth-solidus", "a\uff0fb.model"),  # U+FF0F is NOT a path separator
+    # DEL, the C1 control NEL and the Unicode line separator are line-injection
+    # vectors on a tool's STDIN (blocked by has_line_unsafe_char), but this guard
+    # bounds a PATH argument in argv, where there is no line to inject into, so
+    # they are legitimate filename bytes here and must not be over-blocked.
+    ("del-0x7f", "mod\x7fel.model"),
+    ("c1-nel-0x85", "mod\x85el.model"),
+    ("line-separator-2028", "mod\u2028el.model"),
 ]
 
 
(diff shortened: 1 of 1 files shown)


alvations:
Patching the unicode windows paths in the tests. Merging this in.

PR #3877, opened 2026-09-09
Give every pathsec refusal the standard Security Violation marker (spawn_trusted + ZIP audit)

alvations:
Follow-up to the merged exec-trust PR #3858: make **every** `nltk.pathsec`
refusal carry the standard bracketed `Security Violation [context]: ...` form.

An AST audit of `pathsec.py` found the only refusals raised *without* the marker
were `spawn_trusted`'s `shell=True` guard and the three ZIP-audit refusals. That
inconsistency means a marker-based check (anything keying on the `Security
Violation [` form pathsec otherwise guarantees) would not recognize them.

## Changes (all message-only; no logic or exception-type change)
- `spawn_trusted(shell=True)` -> `Security Violation [spawn_trusted]: ... (CWE-78)`.
- `_reject_colliding_members` (member collision / resource poisoning) now takes a
  `context` and raises with the marker.
- `validate_zip_archive`'s NUL-byte-member and corrupt-archive refusals carry the
  marker and the context.

After this, an AST audit shows **zero** pathsec refusals without the bracketed
marker.

## Tests
- `spawn_trusted` shell refusal asserts the marker (`match=`).
- Collision, NUL-byte-member and corrupt-archive refusals assert the marker.

## Verified (real, not mock)
- Full exec-trust suite + zip/pathsec attack matrix: 501+ pass, no leaks; every
  attack (world-writable binary, shell=True, zip-slip, collision, NUL, corrupt)
  still refused, now with the bracketed marker.
- Real Senna (`USSR->NNP`) and HunPos (`the->DT`) still spawn and tag through
  `spawn_trusted` (no regression; they pass argv, not shell).
- 300 nltk modules import; `black`/`isort` clean.

### DIFF
--- a/nltk/pathsec.py
+++ b/nltk/pathsec.py
@@ -353,7 +353,11 @@ def spawn_trusted(target, args=(), **popen_kw):
     out-of-scope race while adding a Linux-only, magic-symlink exec path.
     """
     if popen_kw.get("shell"):
-        raise ValueError("spawn_trusted refuses shell=True")
+        raise ValueError(
+            "Security Violation [spawn_trusted]: shell=True is refused; a shell "
+            "would re-interpret the trusted command on caller-supplied args "
+            "(CWE-78). Pass an argv list and no shell."
+        )
     popen_kw.setdefault("env", safe_env())
     popen_kw.setdefault("close_fds", True)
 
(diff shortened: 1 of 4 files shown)


alvations:
Merging this in. Having a marker in the error message from refusal so that we can validate/assert when needed.

PR #3878, opened 2026-09-09
Fix resolution paths that cancel distinct literals

### LINKED ISSUE #2699: ResolutionProver is unsound
Following: https://stackoverflow.com/questions/43507770/basic-first-order-logic-inference-fails-for-symmetric-binary-predicate

ResolutionProver appears to not implement the resolution rule correctly, a fix is described by (not myself) above.

UnbearableFate:
`ResolutionProver` currently proves `foo(Bar)` from only `all x.all y.(parentof(y,x) <-> childof(x,y))`. A unification path can remove several distinct complementary pairs: `{P, Q}` and `{-P, -Q}` incorrectly produce the empty clause even though both parents can be true.

Require the removed resolution literals to factor to one common literal with the same polarity before emitting a resolvent. Apply the factoring bindings to the remaining literals. This preserves valid factoring and leaves the shared subsumption traversal and equality demodulation intact. Also correct the existing doctest that expected a false contradiction.

Fixes #2699.

Validation on Python 3.13.14:
- Regression tests cover the reported public proof, distinct predicates/arguments/polarities, duplicate literals, variable factoring, and substitution into remaining literals. An independent truth-table check verifies ground resolvents against their parents. The final regression file has 8 failures and 4 passes on unchanged upstream production code; all 12 pass with the fix.
- Resolution/logic doctests and related unit/security tests: 27 passed, 2 skipped because `mace4` is unavailable.
- Module examples in `resolution.py` and `logic.py`: 2 passed.
- An additional check of `tableau.py::testHigherOrderTableauProver` fails with a parser error on both this branch and a pristine archive of the base revision. No changes to that unrelated example.

All applicable pre-commit hooks passed on the changed files, including Black, isort, Ruff, pyupgrade, and the repository-wide security guard checks; `git diff --check` passed. The full cross-platform CI matrix was not run locally.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -178,6 +178,7 @@
 - Louis Tiao
 - Steven Tomcavage
 - Tiago Tresoldi
+- UnbearableFate
 - Marcus Uneson
 - Yu Usami
 - Petro Verkhogliad
--- a/nltk/inference/resolution.py
+++ b/nltk/inference/resolution.py
@@ -539,6 +539,29 @@ def _unify_terms(a, b, bindings=None, used=None):
 
 def _complete_unify_path(first, second, bindings, used, skipped, debug):
     if used[0] or used[1]:  # if bindings were made along the path
+        # Resolution may remove several literals only if they factor to the
+        # same literal. Removing distinct complementary pairs is unsound:
+        # {P, Q} and {-P, -Q} do not imply the empty clause.
+        # Equality literals are consumed by demodulation, not resolution.
+        resolved = [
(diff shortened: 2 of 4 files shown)


ekaf (approved):
The fix is sound: it correctly prevents removing multiple distinct complementary pairs in one step, while still allowing factoring when the removed literals unify to the same literal. The new tests cover the unsound cases, the factoring cases, and an independent ground soundness check. LGTM.

stevenbird:
Thanks @UnbearableFate and @ekaf!

PR #3880, opened 2026-09-09
Explain negative contingency counts in likelihood ratios

Con-Benksl:
An inconsistent set of marginal counts currently reaches `math.log` in `NgramAssocMeasures.likelihood_ratio` and raises an unexplained `ValueError: math domain error`. This also happens with the seven-word `TrigramCollocationFinder.nbest` example in #2032.

Validate the derived contingency table before computing the likelihood ratio and explain that negative counts can result from inconsistent marginals or collocation approximations on very small samples. This follows the diagnostic approach suggested in #2032. The exception remains a `ValueError`; valid likelihood-ratio scores and zero-count cells keep their existing behavior. This does not change the small-sample counting approximation.

Related to #2032.

Validation (Python 3.13.13):

- Before the fix, the three existing invalid-count cases and the issue's small-sample reproduction failed the diagnostic assertions: 4 failed.
- `python -m pytest nltk/test/unit/test_metrics.py nltk/test/unit/test_collocations.py nltk/test/metrics.doctest -q`: 8 passed. The existing bigram, trigram, and quadgram numerical checks remain covered.
- `python -m doctest nltk/metrics/association.py`: passed.
- `pre-commit run --all-files`: all 13 hooks passed; `git diff --check` passed.

No corpus downloads were needed; the full corpus-dependent test suite was not run. OpenAI Codex implemented and tested this change, with a separate agent performing an independent code review.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -319,6 +319,7 @@
 - Nicola <https://github.com/trinik15>
 - medisean <https://github.com/medisean>
 - tandede <https://github.com/tandede>
+- Con-Benksl <https://github.com/Con-Benksl>
 
 ## Others whose work we've taken and included in NLTK, but who didn't directly contribute it:
 
--- a/nltk/metrics/association.py
+++ b/nltk/metrics/association.py
@@ -141,8 +141,17 @@ class NgramAssocMeasures(metaclass=ABCMeta):
 
     @classmethod
     def likelihood_ratio(cls, *marginals):
-        """Scores ngrams using likelihood ratios as in Manning and Schutze 5.3.4."""
+        """Scores ngrams using likelihood ratios as in Manning and Schutze 5.3.4.
+
+        :raises ValueError: If the contingency table contains negative counts.
+        """
         cont = cls._contingency(*marginals)
+        if any(count < 0 for count in cont):
+            raise ValueError(
+                "The contingency table contains negative counts. "
+                "Check the marginal counts; very small samples can make "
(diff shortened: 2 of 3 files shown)


ekaf (changes requested):
Thanks for the fix. Rejecting negative contingency counts in `likelihood_ratio` is the right call — the LR formula isn’t meaningful with negative cells, and a clear `ValueError` is better than returning a bogus score.

**What looks good**
- `likelihood_ratio` now validates the contingency table and raises before computing the LR.
- The docstring is updated with `:raises ValueError:`.
- Tests now assert the specific error message for bigram, trigram, and quadgram measures.
- The new `TrigramCollocationFinder` regression test is a good integration-level check.
- `AUTHORS.md` update is fine.
- CI is green.

**Requested change**
Please extend this PR so the validation applies to all association measures that derive a contingency table, rather than leaving that as a follow-up. Other measures in the same module also use `_contingency` / `_expected_values` (e.g. `chi_sq`), so the cleanest approach is to move the negative-count check into `_contingency` (or a shared helper used by it). Then `likelihood_ratio`’s explicit check becomes redundant and can be removed.

Please also:
- Keep the same error message so the existing regex tests continue to pass.
- Add at least one test for another affected measure (e.g. `chi_sq`) to cover the centralized behavior.
- Update any public docstrings that should document `:raises ValueError:` as a result.

This keeps the fix complete in this PR and avoids leaving other association measures with the same silent-bad-score problem.

PR #3881, opened 2026-09-10
Preserve sentence boundaries in MaltParser batches

### LINKED ISSUE #2373: MaltParser bug for parse_tagged_sents
maltparser version 1.9.2.
I've input several sentences but the parser only output the result of the first sentence.
Then I read the examples in maltparser and I found that the input and output .conll files only use one blank line to split two sentences.
It seems that nltk.parse.util.taggedsents_to_conll(sentences) should deal with this problem.

Con-Benksl:
`parse_tagged_sents` stops after the first sentence because `taggedsents_to_conll` emits three newline characters between sentences. MaltParser treats the resulting extra empty sentence as the end of input.

Emit the single blank-line separator expected by CoNLL and update the example accordingly. Also ignore empty output blocks so MaltParser's normal trailing separator does not produce a spurious empty `DependencyGraph`.

Fixes #2373.

Validation:

- The four new regression cases failed on the unchanged code. They check exact serialized boundaries and zero/one/multiple output sentences independently.
- Related Malt and wrapper-security tests: **46 passed, 3 skipped**. The skipped cases require Stanford Parser, which was not installed. The existing real Malt training test ran successfully.
- Real MaltParser 1.9.2 with Corretto 11.0.31: trained a 5.5 KB model from three synthetic sentences. Before the fix, batch parsing returned only the first sentence plus an empty graph; afterward, it returned all three sentences exactly once. Generator batches of zero/one/three sentences, `parse_sents`, and `parse_one` were also checked.
- All 13 repository pre-commit checks and `git diff --check` passed on Python 3.13.13.

The real parser checks used the official MaltParser distribution and a locally trained toy model, with no pretrained-model or corpus downloads. The full corpus-dependent NLTK suite and tokenizer/tagger-dependent inline examples were not run.

OpenAI Codex implemented and verified this change autonomously. A separate agent reviewed the final diff, callers, MaltParser's reader/writer contract, and validation evidence.

### DIFF
--- a/nltk/parse/util.py
+++ b/nltk/parse/util.py
@@ -106,8 +106,8 @@ def taggedsents_to_conll(sentences):
     """
     A module to convert the a POS tagged document stream
     (i.e. list of list of tuples, a list of sentences) and yield lines
-    in CONLL format. This module yields one line per word and two newlines
-    for end of sentence.
+    in CONLL format. This module yields one line per word and one blank line
+    at the end of each sentence.
 
     >>> from nltk import word_tokenize, sent_tokenize, pos_tag
     >>> text = "This is a foobar sentence. Is that right?"
@@ -122,12 +122,10 @@ def taggedsents_to_conll(sentences):
     5	sentence	_	NN	NN	_	0	a	_	_
     6	.		_	.	.	_	0	a	_	_
     <BLANKLINE>
-    <BLANKLINE>
     1	Is	_	VBZ	VBZ	_	0	a	_	_
     2	that	_	IN	IN	_	0	a	_	_
     3	right	_	NN	NN	_	0	a	_	_
     4	?	_	.	.	_	0	a	_	_
-    <BLANKLINE>
     <BLANKLINE>
 
     :param sentences: Input sentences to parse
@@ -137,7 +135,7 @@ def taggedsents_to_conll(sentences):
     """
     for sentence in sentences:
         yield from taggedsent_to_conll(sentence)
-        yield "\n\n"
+        yield "\n"
 
 
 ######################################################################
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -37,6 +37,7 @@
 - Graham Christensen
 - Trevor Cohn
 - David Coles
+- Con-Benksl
 - Tom Conroy <https://github.com/tconroy>
 - Claude Coulombe
 - Lucas Cooper
--- a/nltk/parse/malt.py
+++ b/nltk/parse/malt.py
@@ -259,6 +259,8 @@ class MaltParser(ParserI):
                     output_file_name, context="MaltParser.parse_tagged_sents"
                 ) as infile:
                     for tree_str in infile.read().split("\n\n"):
+                        if not tree_str.strip():
+                            continue
                         yield (
                             iter(
                                 [
--- a/nltk/test/unit/test_malt.py
+++ b/nltk/test/unit/test_malt.py
@@ -5,6 +5,15 @@ import pytest
 
 import nltk.parse.malt as malt
 from nltk.parse.malt import MaltParser
+from nltk.parse.util import taggedsents_to_conll
+
+
+def test_tagged_sents_to_conll_separates_sentences():
+    sentences = iter([[("hello", "NN")], [("world", "NN")]])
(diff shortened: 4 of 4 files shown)


nltk/parse/util.py
```diff
@@ -106,8 +106,8 @@ def taggedsents_to_conll(sentences):
     """
     A module to convert the a POS tagged document stream
     (i.e. list of list of tuples, a list of sentences) and yield lines
-    in CONLL format. This module yields one line per word and two newlines
-    for end of sentence.
+    in CONLL format. This module yields one line per word and one blank line
+    at the end of each sentence.
```
alvations, line 110: This breaks long existing conll files convention? Are you sure? 
The new codex is starting to spam us for some reason...


@Con-Benksl please manually check and make sure the actual MaltParser run don't fail, download the binary and run it on actual outputs. Thank you in advance!

Con-Benksl: Thanks for raising this. To keep the authorship clear, the checks below were executed by Codex; I am not claiming a human self-review.

I reran the **official MaltParser 1.9.2 binary**, including its bundled `talbanken05_train.conll` (32 sentences) and `talbanken05_test.conll` (7 sentences), using Corretto 11.0.31. This supplements the earlier synthetic-model check with the distribution's existing CoNLL-X files.

| Input/path | Base `a61e890` | PR `52a1656` |
| --- | --- | --- |
| Original test file → Java CLI | 7 sentences / 87 tokens | 7 / 87 |
| Same words/POS through `taggedsents_to_conll` → Java CLI | 1 sentence / 11 tokens | 7 / 87 |
| Same tagged sentences → `parse_tagged_sents` | 1 sentence + an empty graph | 7 sentences, no empty graph |

The Java invocations all exit **0**, including the truncated result on the base. I checked the actual output token sequences, not only exit codes. The corrected sentence lengths are `[11, 2, 8, 15, 22, 13, 16]`.

The distinction is **two newline characters versus two blank lines**. Each token row already ends with `\n`; the old extra `\n\n` makes three consecutive LFs. The new extra `\n` retains one blank line between sentences, matching both the bundled files and [the CoNLL-X specification, §3](https://aclanthology.org/W06-2920.pdf). MaltParser 1.9.2's `TabReader.readSentence()` stops on the first blank line; reading the second immediately produces a graph with no tokens and returns `false`. Its `TabWriter` itself emits one blank line.

For reproduction, download the [official distribution](https://www.maltparser.org/download.html), enter its extracted directory, and run this with the desired NLTK checkout on `PYTHONPATH`:

```sh
java -jar maltparser-1.9.2.jar -c boundary \
  -i examples/data/talbanken05_train.conll -m learn
python - <<'PY'
from pathlib import Path
import nltk
from nltk.parse.util import taggedsents_to_conll

print(nltk.__file__)
blocks = Path("examples/data/talbanken05_test.conll").read_text().split("\n\n")
sentences = [[(cols[1], cols[4]) for cols in
              (line.split("\t") for line in block.splitlines())]
             for block in blocks if block.strip()]
Path("generated.conll").write_text("".join(taggedsents_to_conll(sentences)))
PY
java -jar maltparser-1.9.2.jar -c boundary \
  -i generated.conll -o generated-output.conll -m parse
python - <<'PY'
from pathlib import Path
text = Path("generated-output.conll").read_text()
print([len(block.splitlines()) for block in text.split("\n\n") if block.strip()])
PY
```

Actual stdout from running that block with each checkout:

```text
base: [11]
PR:   [11, 2, 8, 15, 22, 13, 16]
```

This intentionally changes the serializer's exact separator bytes. It does not change `train_from_file`, the corpus readers, or existing files on disk. The only production caller of this helper inside NLTK is the Malt wrapper. I have left the review thread open for your assessment.

alvations: No, this still breaks the conventions of any downstream users that already parse the double `<BLANKLINE>`! This is breaking change -_-|||

Agent, the double new line exists for very long, anyone that uses output from Maltparser across the years have set it as such. You need to do a thorough research on all the tools, search the web that uses the Maltparser outputs through NLTK and then report back if they really don't use the double `<BLANKLINE>` before making this change.

OSS users and devs collaborate as humans and interact, if agents need to interact, they need to communicate changes across OSS, not just make changes in silos.

PR #3882, opened 2026-09-10
Reset stale IBM lexical probabilities with a configurable floor

### LINKED ISSUE #1939: Translation Probabilities do not sum to 1 in IBM1 Translation Model
After running IBM1 Translation model in the given code, I've summed the translation probabilites (the probability for a target word given source word), and it doesn't sum to 1:

```python
from nltk.translate import IBMModel1, AlignedSent
import numpy as np

src_sentences = [['a', '.'], ['b', 'b', '.']]
trg_sentences = [['X', '.'], ['Z', '.']]

bitext = [AlignedSent(t, s) for s,t in zip(src_sentences, trg_sentences)]
ibm1 = IBMModel1(bitext, 1)

sum([ibm1.translation_table[t]['b'] for t in ibm1.translation_table])
1.3333333333333333
```

I guess it is because here:
https://github.com/nltk/nltk/blob/7bbc00e908087b60325db30526872306e767de13/nltk/translate/ibm_model.py#L336-L340
The iteration only on the counts, and not all the vocabulary. So (src_word, trg_word) pairs that don't appear in the same sentence pairs, won't get updated and will stay with the initial uniform probabilities

UnbearableFate:
Fixes #1939.

IBM1/IBM2 training updates counted lexical pairs but leaves stale uniform values on pairs that never co-occur. In the issue's two-sentence example, the probabilities for source `b` consequently sum to `4/3`. Rebuild each counted target row before assigning normalized counts, retaining sparse storage.

Expose `lexical_floor=None` in the IBM1–5 constructors. The default uses the existing `MIN_PROB=1e-12`; callers can select `0.0`, `1e-7`, or another finite real value in `[0, 1]`. The shared lexical M-step and model-created missing-entry defaults use this floor. Nested initialization propagates it and preserves IBM2's learned IBM1 starting point. Non-lexical probability tables and alignment-score guards retain their existing `MIN_PROB` behavior.

After one IBM1 iteration on the issue corpus, source `b` has probabilities `[floor, 0.5, 0.5]` for `[X, ., Z]`. Thus the sum is `1.000000000001` by default, `1.0` with `0.0`, and `1.0000001` with `1e-7`. Zero selects the unsmoothed lexical update for positive source totals, subject to ordinary floating-point rounding. Matching GIZA++'s lexical floor does not imply complete implementation equivalence. Positive floors do not reserve mass or renormalize; no statistical smoothing or alignment-quality improvement is claimed.

Initialization and boundary policies are documented and tested. Uniform initialization is unchanged. Supplied custom values are retained; an explicit numerical floor copies stored lexical entries with that missing-entry default, while `None` preserves the custom table's original lookup behavior until rows are trained. Zero-total sources remain at the selected floor in updated rows, without a uniform fallback; this defines the previously undefined division. Target rows absent from counts remain outside the sparse update, and Models 3–5 retain their existing resets. All-zero E-step alternatives can still raise `ZeroDivisionError`.

Validation for the latest change (Python 3.13.14):

- IBM unit tests and Models 1–5 doctests: **94 passed, 7 subtests passed**.
- Import-verified isolated comparison against previous head `0aa1ed4`: identical lexical probes and corpus alignments for all 15 IBM1–5 runs at 0/1/2 iterations with the argument omitted or explicitly `None`. Independent rational-count first M-step matches the zero-floor result within `1e-15`.
- Changed-file pre-commit, repository guards, and `git diff --check`: passed.
- Broader translation unit suite: **139 passed, 7 subtests passed, 5 failed** due to missing `wordnet` / `wmt15_eval` data. The same five failures were reproduced on unchanged `0aa1ed4`.

Full corpus/cross-platform and alignment/translation quality evaluations were not performed locally. The earlier [before/after GIZA++/MGIZA++ audit](https://github.com/nltk/nltk/pull/3882#issuecomment-5618262700) and [original-paper/literature audit](https://github.com/nltk/nltk/pull/3882#issuecomment-5634516137) describe their own versions and limits; neither is presented as a fresh run of the new API. The contribution-process shortcoming discussed in #3883 remains acknowledged: the initial proposal/reference audit should have preceded the PR.

The current head's [full CI run](https://github.com/nltk/nltk/actions/runs/35498052052) has started; its final result is pending. The old authorization-blocked run is not the status of this new commit.

Agent-use disclosure for the numerical API follow-up: **gpt-6-astra**. Actual task prompt **excerpt, translated from Chinese**: “Expose a numerical constructor parameter `lexical_floor=None` (default `MIN_PROB=1e-12`), supporting `0.0` and `1e-7`; the shared M-step and unseen-pair lookup should use this floor.” Audit summary: traced IBM1–5 constructors and custom-table initialization, preserved warm-start and non-lexical guards, documented zero-total/zero-E-step policies, and verified the numerical and default-compatibility cases above. Earlier implementation disclosures remain in the linked discussion.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -178,6 +178,7 @@
 - Louis Tiao
 - Steven Tomcavage
 - Tiago Tresoldi
+- UnbearableFate
 - Marcus Uneson
 - Yu Usami
 - Petro Verkhogliad
--- a/nltk/test/unit/translate/test_ibm2.py
+++ b/nltk/test/unit/translate/test_ibm2.py
@@ -10,6 +10,19 @@ from nltk.translate.ibm_model import AlignmentInfo
 
 
 class TestIBMModel2(unittest.TestCase):
+    def test_training_uses_model1_translation_probabilities(self):
+        corpus = [
+            AlignedSent(["X", "."], ["a", "."]),
+            AlignedSent(["Z", "."], ["b", "b", "."]),
+        ]
+
+        model2 = IBMModel2(corpus, 1)
+
+        # Two Model 1 iterations initialize the lexical table before Model 2.
+        # Starting Model 2 from uniform lexical probabilities gives 0.5 instead.
+        self.assertAlmostEqual(model2.translation_table["X"]["a"], 0.665)
+        self.assertAlmostEqual(model2.translation_table["Z"]["b"], 0.65625)
+
     def test_set_uniform_alignment_probabilities(self):
         # arrange
         corpus = [
--- a/nltk/test/unit/translate/test_ibm_model.py
+++ b/nltk/test/unit/translate/test_ibm_model.py
@@ -5,7 +5,7 @@ Tests for common methods of IBM translation models
 import unittest
 from collections import defaultdict
 
-from nltk.translate import AlignedSent, IBMModel
+from nltk.translate import AlignedSent, IBMModel, IBMModel1, IBMModel2
 from nltk.translate.ibm_model import AlignmentInfo
 
 
@@ -30,6 +30,31 @@ class TestIBMModel(unittest.TestCase):
         ibm_model = IBMModel(parallel_corpora)
         self.assertEqual(len(ibm_model.src_vocab), 1)  # addition of NULL token
         self.assertEqual(len(ibm_model.trg_vocab), 0)
+
+    def test_lexical_probabilities_are_normalized_after_training(self):
+        for model_class in (IBMModel1, IBMModel2):
+            for iterations in (1, 5):
+                with self.subTest(model=model_class.__name__, iterations=iterations):
+                    corpus = [
+                        AlignedSent(["X", "."], ["a", "."]),
+                        AlignedSent(["Z", "."], ["b", "b", "."]),
+                    ]
+                    model = model_class(corpus, 0)
+                    # Materialize a pair that never co-occurs before the M-step.
+                    self.assertEqual(model.translation_table["X"]["b"], 1 / 3)
+                    for _ in range(iterations):
+                        model.train(corpus)
+                    for source in model.src_vocab:
+                        total = sum(
+                            model.translation_table[t][source] for t in model.trg_vocab
+                        )
+                        self.assertAlmostEqual(total, 1.0)
+                    self.assertEqual(
+                        model.translation_table["X"]["b"], IBMModel.MIN_PROB
+                    )
+                    self.assertEqual(
+                        model.translation_table["Z"]["a"], IBMModel.MIN_PROB
+                    )
 
     def test_best_model2_alignment(self):
         # arrange
--- a/nltk/translate/ibm_model.py
+++ b/nltk/translate/ibm_model.py
@@ -344,6 +344,9 @@ class IBMModel:
 
     def maximize_lexical_translation_probabilities(self, counts):
         for t, src_words in counts.t_given_s.items():
+            # Unobserved pairs must not retain the previous uniform prior or
+            # explicit values. Keep the row sparse, with the usual zero floor.
+            self.translation_table[t] = defaultdict(lambda: IBMModel.MIN_PROB)
             for s in src_words:
                 estimate = counts.t_given_s[t][s] / counts.any_t_given_s[s]
                 self.translation_table[t][s] = max(estimate, IBMModel.MIN_PROB)
--- a/nltk/translate/ibm_model.py
+++ b/nltk/translate/ibm_model.py
@@ -343,9 +343,11 @@ class IBMModel:
         return neighbors
 
     def maximize_lexical_translation_probabilities(self, counts):
+        # Training normally reuses the same corpus. Rows for targets absent
+        # from counts are left unchanged if a smaller corpus is supplied.
         for t, src_words in counts.t_given_s.items():
             # Unobserved pairs must not retain the previous uniform prior or
-            # explicit values. Keep the row sparse, with the usual zero floor.
+            # explicit values. Keep the row sparse, with the usual MIN_PROB floor.
             self.translation_table[t] = defaultdict(lambda: IBMModel.MIN_PROB)
             for s in src_words:
                 estimate = counts.t_given_s[t][s] / counts.any_t_given_s[s]


ekaf (changes requested):
## Review

The fix is conceptually correct: unobserved translation pairs should be floored to `IBMModel.MIN_PROB` after the M-step rather than retaining a stale uniform prior. The new test covers this, including across multiple `train()` calls and for both IBM1 and IBM2.

Two minor points:

### Misleading comment

```python
# ... Keep the row sparse, with the usual zero floor.
```

The floor is `IBMModel.MIN_PROB` (`1e-12`), not zero. Suggest rewording to "the usual `MIN_PROB` floor".

### Rows absent from `counts.t_given_s` are not reset

Only targets present in `counts.t_given_s` are cleared. If `train()` is ever called with a corpus whose target vocabulary is smaller than the original, stale rows from the previous corpus persist. This is unlikely in the standard workflow (where `train()` is always called with the same corpus), but worth a brief comment or a defensive loop over `self.translation_table.keys()`.

Overall the fix looks good.

UnbearableFate:
Thanks @ekaf. Addressed both points in 809cee2:

- Corrected “zero floor” to “MIN_PROB floor”.
- Took your comment option for the smaller-corpus case: the M-step now explicitly notes that training normally reuses the same corpus and that target rows absent from the new counts remain unchanged. The PR description also records this limitation.

This follow-up only changes comments. Re-ran the IBM unit tests (29 passed, 4 subtests passed), IBM1–5 doctests (5 passed), changed-file pre-commit hooks, and `git diff --check`; all passed.

ekaf (approved):
Thanks @UnbearableFate for addressing the review concerns. LGTM.

alvations:
To agent, same as before, for open-source transparency when an agentic coding tool is used: please include the prompt, the model used, and the reasoning/audit trail, either in the commit message, in block comments where humans or other agents need to know, or as a PR comment, so the implementation can be traced back if necessary.

alvations (changes requested):
Agent who commit the min floor, this needs a closer look. Read #1939 fully, not just the original post.

Making the IBM model "better" is correct but it's not what we want to do here. We want to replicate the IBM implementation, even it's flaws and if there's a better version then the toggle should be open to user through kwarg or something.

IIRC IBM 2 inherits IBM 1 probs is how it was meant to be.

alvations:
To agent, think thoroughly and check against the GIZA++ and MGIZA++ implementation to verify that your min floor here follows the original implementations of IBM models' design faithfully. Make sure you report the side-by-side comparison.

UnbearableFate:
Thanks @alvations. I agree that IBM2 must retain IBM1's learned lexical initialization. I read all of #1939, including both follow-up comments, then built and ran GIZA++ and MGIZA++ on its exact two-sentence corpus. The PR leaves that inheritance intact. The follow-up commit `0aa1ed4` adds a regression for the normal `IBMModel2(corpus, 1)` constructor path; it makes no further production-code changes.

The issue sums over **target** words while holding source `b` fixed, so the axis in the original reproducer is correct. The discrepancy is the non-cooccurring pair `p(X | b)`, which NLTK previously left at `1/3` after training. This is separate from using learned IBM1 probabilities to initialize IBM2.

### Source comparison

I pinned [GIZA++ at ce4829ba1517b731379ceb5eb7b81275acf170cb](https://github.com/moses-smt/giza-pp/tree/ce4829ba1517b731379ceb5eb7b81275acf170cb) and [MGIZA++ at 6105035cd5b2f631b80c58abba47e49793feacf7](https://github.com/moses-smt/mgiza/tree/6105035cd5b2f631b80c58abba47e49793feacf7). The comparison uses their default binary-search lexical-table builds.

| Step | GIZA++ | MGIZA++ | NLTK with this PR |
|---|---|---|---|
| First unseeded IBM1 E-step | Uses a uniform lexical value | Same | Uses existing uniform initialization |
| Lexical M-step | Divides each stored pair's count by the sum for its source word | Same | Uses the existing count/source-total estimate; replaces each counted target row's stale entries/default first |
| Pair absent from the cooccurrence table | `getProb` returns `PROB_SMOOTH` | Same | An unobserved pair in an updated target row returns existing `IBMModel.MIN_PROB` |
| Floor constant | Default `1e-7` | Default `1e-7` | Existing `1e-12`, unchanged by this PR |
| IBM1 → IBM2 | IBM2 copies the model state with the shared lexical table; initializes the alignment table | Same | Existing `ibm1.translation_table` assignment; initializes the alignment table |
| IBM2 training | E-step uses inherited lexical probabilities, then the lexical M-step updates them | Same | Same order; row replacement happens after collecting E-step counts |

Relevant implementation links:

- GIZA++: [first IBM1 E-step](https://github.com/moses-smt/giza-pp/blob/ce4829ba1517b731379ceb5eb7b81275acf170cb/GIZA%2B%2B-v2/model1.cpp#L139-L201), [normalization](https://github.com/moses-smt/giza-pp/blob/ce4829ba1517b731379ceb5eb7b81275acf170cb/GIZA%2B%2B-v2/TTables.cpp#L79-L100), [missing-pair lookup](https://github.com/moses-smt/giza-pp/blob/ce4829ba1517b731379ceb5eb7b81275acf170cb/GIZA%2B%2B-v2/TTables.h#L280-L287), [IBM2 initialization and M-step](https://github.com/moses-smt/giza-pp/blob/ce4829ba1517b731379ceb5eb7b81275acf170cb/GIZA%2B%2B-v2/model2.cpp#L33-L96).
- MGIZA++: [first IBM1 E-step](https://github.com/moses-smt/mgiza/blob/6105035cd5b2f631b80c58abba47e49793feacf7/mgizapp/src/model1.cpp#L264-L319), [normalization](https://github.com/moses-smt/mgiza/blob/6105035cd5b2f631b80c58abba47e49793feacf7/mgizapp/src/TTables.cpp#L124-L143), [missing-pair lookup](https://github.com/moses-smt/mgiza/blob/6105035cd5b2f631b80c58abba47e49793feacf7/mgizapp/src/TTables.h#L324-L330), [IBM2 initialization and M-step](https://github.com/moses-smt/mgiza/blob/6105035cd5b2f631b80c58abba47e49793feacf7/mgizapp/src/model2.cpp#L33-L102).
- NLTK: [unchanged IBM1 → IBM2 initialization](https://github.com/nltk/nltk/blob/0aa1ed4/nltk/translate/ibm2.py#L122-L137), [existing floor](https://github.com/nltk/nltk/blob/0aa1ed4/nltk/translate/ibm_model.py#L65-L79), [changed M-step](https://github.com/nltk/nltk/blob/0aa1ed4/nltk/translate/ibm_model.py#L345-L354).

### Runtime comparison

For each source below, the vector is `[p(X | source), p(. | source), p(Z | source)]`. Numbers are rounded to six decimals, matching reference dump precision, except that the NLTK floor is shown explicitly as `1e-12`. `F` denotes a missing pair in a reference dump: its `1e-7` lookup floor is established by the getters linked above, not by treating an absent dump entry as a measured value.

**NLTK (before)** is the unmodified PR base [`a61e89015e15158d0a062c91d097c18869d31306`](https://github.com/nltk/nltk/tree/a61e89015e15158d0a062c91d097c18869d31306); **NLTK (after)** is [`0aa1ed46c30557801546d54f67119e0f3e46cd90`](https://github.com/nltk/nltk/tree/0aa1ed46c30557801546d54f67119e0f3e46cd90). I ran each checkout and schedule in a fresh Python 3.13.14 process, asserted `nltk.__file__` matched the selected checkout, and recorded its Git SHA. I also reran both compiled reference executables for both schedules.

| Training schedule | Source | GIZA++ | MGIZA++ | NLTK (before) | NLTK (after, this PR) |
|---|---|---|---|---|---|
| 1 IBM1 iteration | NULL | [0.285714, 0.500000, 0.214286] | [0.285714, 0.500000, 0.214286] | [0.285714, 0.500000, 0.214286] | [0.285714, 0.500000, 0.214286] |
| 1 IBM1 iteration | a | [0.500000, 0.500000, F] | [0.500000, 0.500000, F] | [0.500000, 0.500000, 0.333333] | [0.500000, 0.500000, 1e-12] |
| 1 IBM1 iteration | . | [0.285714, 0.500000, 0.214286] | [0.285714, 0.500000, 0.214286] | [0.285714, 0.500000, 0.214286] | [0.285714, 0.500000, 0.214286] |
| 1 IBM1 iteration | b | [F, 0.500000, 0.500000] | [F, 0.500000, 0.500000] | [0.333333, 0.500000, 0.500000] | [1e-12, 0.500000, 0.500000] |
| 2 IBM1 + 1 IBM2 iterations | NULL | [0.238528, 0.659319, 0.102154] | [0.238528, 0.659319, 0.102154] | [0.238528, 0.659319, 0.102154] | [0.238528, 0.659319, 0.102154] |
| 2 IBM1 + 1 IBM2 iterations | a | [0.665000, 0.335000, F] | [0.665000, 0.335000, F] | [0.665000, 0.335000, 0.333333] | [0.665000, 0.335000, 1e-12] |
| 2 IBM1 + 1 IBM2 iterations | . | [0.238528, 0.659319, 0.102154] | [0.238528, 0.659319, 0.102154] | [0.238528, 0.659319, 0.102154] | [0.238528, 0.659319, 0.102154] |
| 2 IBM1 + 1 IBM2 iterations | b | [F, 0.343750, 0.656250] | [F, 0.343750, 0.656250] | [0.333333, 0.343750, 0.656250] | [1e-12, 0.343750, 0.656250] |

On this corpus, **all ten observed lexical entries are exactly equal between NLTK before and after** for each schedule. Both versions match both reference dumps within `5e-7` (maximum absolute differences: `2.8571428573e-7` for IBM1 and `4.2636359315e-7` for IBM2). The two differences are `p(X | b)` and `p(Z | a)`: each changes from `1/3` to NLTK's existing `1e-12` floor. For source `b`, the target-word sum changes from `1.333333333333333` to `1.000000000001` in both schedules; the latter is close to one, not strictly one.

NLTK's `IBMModel2(corpus, 1)` already runs two IBM1 iterations before one IBM2 iteration. The added test checks `.665` and `.65625`; replacing that warm-start with uniform lexical probabilities makes it fail with `.5` instead of `.665`.

This supports retaining the fix for stale non-cooccurring probabilities, rather than introducing a new modeling option. It does **not** establish complete numerical equivalence between NLTK and GIZA++/MGIZA++: the floor sizes differ, floor lookup can make sums slightly exceed one, the reference binary-table normalizers have a uniform fallback for zero-total source rows, and NLTK's documented smaller-corpus limitation remains. GIZA++ also has a non-default hash-table implementation with pruning and floor-mass compensation; this PR does not reproduce that alternate implementation.

### Reconsideration after #3883

I read #3883 and its comment, then re-read all of #1939 and this PR's reviews. I should have presented the reference audit and proposed correction for discussion before opening this PR; the original issue and expressions of interest did not establish that this particular change was wanted. That ordering was a shortcoming in the original contribution process.

With ekaf's review and the executable comparisons now available, I still recommend this small M-step correction. The baseline already matches the reference implementations on the observed pairs; the discrepancy is the stale uniform value for absent pairs. The fix preserves the first uniform E-step, IBM1-to-IBM2 initialization, and NLTK's existing floor. This evidence supports correcting the stale entries, without introducing a separate improved-model option. The final decision on accepting the behavior change remains with the maintainers. No additional code change accompanies this audit update.

This is a bounded IBM audit. I have not read or audited every NLTK module/test, nor validated the entire security checklist in #3883. The patch adds no library file access, serialization, printing, or external-tool execution. Full corpus-dependent and cross-platform validation remains incomplete; the current head's [ci-workflow](https://github.com/nltk/nltk/actions/runs/34473736633) reports `action_required`, so I cannot claim that CI has passed for `0aa1ed4`.

<details>
<summary>Reproduction and validation</summary>

The reference source files were not modified. GIZA++ was built with `make -C GIZA++-v2 -j4 CFLAGS_GLOBAL=-std=gnu++11`; MGIZA++ with `cmake -S mgizapp -B build` and `cmake --build build -j4` (Boost thread/system available).

Input `src`:

```text
a .
b b .
```

Input `trg`:

```text
X .
Z .
```

Preprocess with GIZA++'s `plain2snt.out src trg`, then `snt2cooc.out src.vcb trg.vcb src_trg.snt > cooc`.

Run each executable with:

```text
-S src.vcb -T trg.vcb -C src_trg.snt -CoocurrenceFile cooc
-mh 0 -m3 0 -m4 0 -m5 0 -t1 1 -t2 1
```

Add `-m1 1 -m2 0 -o PREFIX` for the first schedule, or `-m1 2 -m2 1 -o PREFIX` for the second; MGIZA++ additionally used `-ncpus 1`. Compare `PREFIX.t1.1` or `PREFIX.t2.1`, decoding IDs with the generated vocabularies (source ID 0 is NULL).

NLTK before/after reproduction used separate checkouts. The verified import paths were `/tmp/nltk-3882-before-a61e890/nltk/__init__.py` and `/tmp/nltk-ibm/nltk/__init__.py`. The baseline checkout was created with:

```sh
git worktree add --detach /tmp/nltk-3882-before-a61e890 a61e89015e15158d0a062c91d097c18869d31306
```

The following minimal script reproduces the NLTK values. Run it in a fresh process for each checkout and each schedule, e.g. `python compare_nltk.py /tmp/nltk-3882-before-a61e890 m1`, then `m2`, and repeat with the PR checkout:

```python
import sys
from pathlib import Path

source = Path(sys.argv[1]).resolve()
sys.path.insert(0, str(source))
import nltk
from nltk.translate import AlignedSent, IBMModel1, IBMModel2

assert Path(nltk.__file__).resolve() == source / "nltk" / "__init__.py"
print(nltk.__file__)
corpus = [AlignedSent(["X", "."], ["a", "."]),
          AlignedSent(["Z", "."], ["b", "b", "."])]
model = {"m1": IBMModel1, "m2": IBMModel2}[sys.argv[2]](corpus, 1)
for word in [None, "a", ".", "b"]:
    print(word, [model.translation_table[t][word] for t in ["X", ".", "Z"]])
```

This update reran **30 IBM unit tests, 4 subtests, and 5 IBM1–5 doctests: all passed** (`35 passed, 4 subtests passed`), plus `git diff --check`. The previous changed-file pre-commit pass and expected warm-start mutation failure remain applicable to the unchanged head; they were not rerun for this comment-only update. Full corpus-dependent/cross-platform tests were not repeated locally.

</details>

Agent-use disclosure: the original implementation's prompt excerpt/model information remains in the PR description. This review follow-up used OpenAI Codex with the requested agent configuration **gpt-6-astra, medium reasoning**; an exact serving snapshot is unavailable. The follow-up task, translated from Chinese, was: “Read the full review, issue/comments, and original implementations; first research correctness and compare them; revise code if necessary while preserving existing contracts and without inventing options; test, push normally, and respond; disclose the model/task instructions and a verifiable audit summary.” The high-level audit is the source comparison, executable schedules/results, preserved initialization path, regression and mutation check above. No additional production algorithm or probability constant was introduced in this follow-up.


Agent-use disclosure for the before/after and #3883 follow-up: OpenAI Codex, requested worker configuration **gpt-6-astra, xhigh reasoning**; an exact serving snapshot is unavailable. The task excerpt, translated from Chinese, was: “Read #3883 and its comments fully, reconsider the existing PR, add a real NLTK baseline in independent source processes with `nltk.__file__` verification, update the existing comparison comment, and respond to the new feedback; revise code only if necessary.” The verifiable audit summary is: pinned before/after imports; four isolated NLTK runs and four reference executable runs; exact equality of all observed NLTK pairs before/after; two stale-pair changes; IBM tests/doctests rerun; process shortcomings and incomplete full-suite/security validation disclosed above. The user had authorized follow-up work and reviewer replies on this existing PR. The earlier medium-reasoning follow-up disclosure above describes that earlier stage, not this update.

alvations:
To agent, good job done. Now for the `Runtime comparison` table, add a `NLTK (before)` comparison to the NLTK changes column. Remember to add the "Agent-use disclosure" in the followup. 

And then read this #3883 and reconsider your work in this PR.

UnbearableFate:
@alvations Updated the [existing runtime table](https://github.com/nltk/nltk/pull/3882#issuecomment-5618262700) with **NLTK (before)** and **NLTK (after)** from separate, import-verified processes. Both versions already match the references on all ten observed pairs per schedule; only the two non-cooccurring pairs change from `1/3` to `1e-12`.

I read #3883 and its comment. I should have brought the reference audit and proposed correction to discussion before opening this PR. The updated audit records that shortcoming and why I still recommend this bounded fix, preserving IBM2's inherited initialization. No further code change was needed. The 30 IBM unit tests, 4 subtests, and 5 doctests passed again; full-project/security and cross-platform validation is incomplete, and current-head CI requires maintainer approval.

Agent-use disclosure: OpenAI Codex, requested **gpt-6-astra / xhigh reasoning** (exact serving snapshot unavailable). Task excerpt: “Read #3883 fully, reconsider the PR, add an independently run NLTK baseline with import verification, update the existing audit, and respond.” The linked audit includes pinned sources, all before/after/reference results, reproduction commands, and validation limits. The user authorized this existing-PR follow-up and reply.

alvations:
@UnbearableFate agent, now do you see from the table why the epsilon smoothing in your NLTK (after) is not as useful as NLTK (before) that didn't sum to one? 

And also read Brown 1993 original IBM paper and check the formal definition there too and the recon all the existing IBM model extensions from 1993 to Moses decoder days up till 2016 and compare your solution against what exists in the literature.

UnbearableFate:
@alvations The distinction I should make is that **this PR corrects stale entries in the lexical table; it does not establish better alignment quality or implement a useful statistical smoothing method**. The `1e-12` guard is effectively zero, while the old `1/3` gives substantial weight to an unseen combination. That can change predictions. The existing table has no gold alignments or held-out quality evaluation, so it cannot establish which behavior is more useful.

I checked Brown's definitions and a bounded set of relevant developments through 2016. This is a comparison of representative mechanisms, not an exhaustive review of every IBM-model extension.

**The formal distinction in Brown (1993).** In [§4.1, pp. 270–273](https://aclanthology.org/J93-2003.pdf#page=9), Eq. (7) requires `sum_f t(f | e) = 1` for each conditioning word `e`. This corresponds to NLTK's `translation_table[f][e]`; source/target labels in the noisy-channel decoding direction do not reverse that conditioning axis. Eqs. (14) and (18) give:

```text
t_new(f | e) = C(f, e) / sum_f' C(f', e)
```

The algorithm's step 3 updates every target-vocabulary word. For a positive source total, a pair that never co-occurs has zero expected count and therefore zero under this unsmoothed M-step. Brown's `epsilon = Pr(m | e)` on p. 270 is a **sentence-length factor**, not a lexical floor. The unnormalized length simplification and the deficient string-generation models discussed in §4.5 concern different quantities. [§4.2, p. 274](https://aclanthology.org/J93-2003.pdf#page=12) transfers Model 1's parameters to Model 2; the PR preserves that initialization.

The PR is still not exact unsmoothed MLE: it preserves NLTK's documented numerical floor, so sums can slightly exceed one. The previously disclosed smaller-corpus limitation also remains.

**Not every useful stored weight has to sum to one.** [Moore (2004), §§4–5](https://aclanthology.org/P04-1066.pdf#page=3) is particularly relevant. His additive smoothing updates both numerator and denominator, `(C(t,s)+n)/(C(s)+n*|V|)`. Separately, his NULL weighting can store a probability multiplied by a NULL multiplicity; those stored values look unnormalized but have an explicit interpretation. Likewise, [Riley and Gildea (2012), §2, Eq. (2)](https://aclanthology.org/P12-2060.pdf#page=2) use variational-Bayes weights that sum to **less** than one. Thus normalization alone is not a sufficient test of usefulness. Neither mechanism is the old NLTK update, which leaves only absent pairs at their initialization values while assigning the observed pairs their full normalized expected counts.

For a concrete distinction, the first-step counts for source `b` in this corpus are `[0, 0.5, 0.5]` for `[X, ., Z]`. Pure MLE gives `[0, .5, .5]`; this PR gives `[1e-12, .5, .5]`. Illustrative add-`0.01` smoothing with vocabulary size three gives `[.009709, .495146, .495146]`: it reserves mass by changing observed estimates too. This is one arithmetic update, not a tuned smoothing experiment or a reproduction of Moore's full system.

<details>
<summary>Comparison with representative literature through 2016</summary>

| Work | Relevant development | Relationship to this PR |
|---|---|---|
| [Vogel, Ney & Tillmann (1996)](https://aclanthology.org/C96-2141.pdf) | HMM alignment dependencies model jumps between source positions. | Changes alignment structure; it is not a justification for retaining an unused lexical initialization value. |
| [Och & Ney (2003), §§2.3.1, 3.1, 3.3](https://aclanthology.org/J03-1002.pdf) | Model 6 combines HMM and Model 4; Eq. (36) normalizes lexical expected counts; Eqs. (37)–(38) smooth alignment/fertility distributions. | Different distributions must be distinguished. The shared lexical M-step and staged training are relevant precedents; the PR does not implement their structural or smoothing extensions. |
| [Moore (2004), §§4–6](https://aclanthology.org/P04-1066.pdf) | Additive lexical smoothing, NULL weighting, and heuristic initialization address shortcomings of ordinary EM. | These are substantive estimation/model choices. The tiny floor and the old stale default implement none of them. |
| [Liang, Taskar & Klein (2006), §§3–4](https://aclanthology.org/N06-1014.pdf) | Joint training encourages agreement between opposite alignment directions; posterior decoding is also studied. | Changes training/decoding criteria. Their Model 2/HMM initialization uses learned Model 1 lexical parameters; the PR retains NLTK's corresponding warm-start. |
| [Koehn et al. (2007), §2](https://aclanthology.org/P07-2045.pdf#page=2) | Moses integrates GIZA++ alignments into a larger translation training, tuning and evaluation pipeline. | A lexical-table comparison is not an evaluation of downstream Moses translation quality. |
| [Toutanova & Galley (2011), §§2–5](https://aclanthology.org/P11-2081.pdf) | Correct the claim of unique Model 1 optimal parameters: different optima/initializations can have different test behavior. | I do not rely on initialization being irrelevant. This patch changes neither NLTK's initial uniform distribution nor its IBM2 initialization path. |
| [Riley & Gildea (2012), §2](https://aclanthology.org/P12-2060.pdf#page=2) | Dirichlet priors and a digamma-based VB update modify lexical estimation; the resulting weights need not sum to one. | A principled counterexample to judging quality by normalization alone, but a different update from leaving absent entries untouched. |
| [Dyer, Chahuneau & Smith (2013), §§2–4](https://aclanthology.org/N13-1073.pdf) | Compact log-linear IBM2 reparameterization; experiments use a Dirichlet prior and variational inference. | The [2016 implementation's `NormalizeVB` and `Normalize`](https://github.com/clab/fast_align/blob/d4a1e82e7621241fdff1db2fd3042c09c68e42e6/src/ttables.h#L84-L115) explicitly distinguish VB and ordinary count normalization. Their alternative model's training choices do not call for removing NLTK IBM2's warm-start. |
| [Bui & Le (2016), §§4.3–5.2, arXiv v4](https://arxiv.org/pdf/1601.03650v4) | Generalize additive smoothing with pair-dependent pseudo-counts and a learned scale, including the corresponding denominator adjustment. | Directly addresses what substantial smoothing would require. The PR supplies neither pseudo-counts nor a learned smoothing scale. This source is the identified preprint, not an assumed peer-reviewed result. |
| [Östling & Tiedemann (2016), §§2.2–3](https://ufal.mff.cuni.cz/pbml/106/art-ostling-tiedemann.pdf) | efmaral uses a Bayesian HMM/fertility model with sparse Dirichlet priors and MCMC inference. | Sparsity is modeled through a prior and inference procedure, not an unchanged uniform value on unobserved pairs. The PR is not an implementation of this approach. |

</details>

**Actual effect and limits.** I restored isolated checkouts of base `a61e890` and head `0aa1ed4`, checked their import paths, and reran IBM1/IBM2 with one and five constructor iterations. All ten observed lexical entries and the training-corpus alignments are identical before/after in each case. An independent first-step calculation using exact rational expected counts also matches the PR, apart from its preserved floor. This accords with the training loops: globally non-cooccurring pairs are never consulted when training on that same corpus. Consequently, this fix does not supply the shrinkage of observed estimates used to address overfitting in the smoothing literature.

It does change behavior on a new sentence pair. After one IBM1 iteration, aligning target `["X"]` to source `["b"]` chooses `b` before the fix (`t(X|b)=1/3 > t(X|NULL)=2/7`) and NULL afterward (`1e-12 < 2/7`). There is no gold label for this synthetic example: it demonstrates a changed choice, not a quality improvement. It also means the change is more than a cosmetic change to a reported sum.

<details>
<summary>Minimal reproduction of the new-sentence behavior</summary>

Run in a fresh Python process from each checkout:

```python
from pathlib import Path
import nltk
from nltk.translate import AlignedSent, IBMModel1

assert Path(nltk.__file__).resolve() == Path.cwd() / "nltk" / "__init__.py"
corpus = [AlignedSent(["X", "."], ["a", "."]),
          AlignedSent(["Z", "."], ["b", "b", "."])]
model = IBMModel1(corpus, 1)
sentence = AlignedSent(["X"], ["b"])
model.align(sentence)
print(model.translation_table["X"]["b"])
print(model.translation_table["X"][None])
print(list(sentence.alignment))
```

</details>

I still recommend the existing correction for a table documented as conditional translation probabilities. I would not recommend it on the unsupported claim that a smaller floor or a sum closer to one improves alignment quality. Useful fallback/smoothing behavior would need an explicit estimation or scoring rule and task evaluation; the comparison does not justify substituting one here. **No further code change was made.**

Validation this round: **35 passed, 4 subtests passed** (30 IBM unit tests plus five IBM doctests); the independent M-step and before/after behavioral checks passed; `git diff --check` passed. The earlier GIZA++/MGIZA++ executable results remain the previously recorded evidence; they were not rerun this round. The additional literature and fast_align comparison is a source audit, not a reimplementation or benchmark of those systems. Full corpus/cross-platform testing and AER/BLEU evaluation were not performed; current-head CI still reports `action_required`.

Agent-use disclosure: **gpt-6-astra**. Relevant prompt **excerpt, translated from Chinese**: “Use primary sources such as the original paper and official implementations to verify the conditional-probability axis, MLE, sparse/unobserved pairs, smoothing, and IBM2 warm-start. Do not claim to exhaust all literature or make algorithm changes without evidence.” The audit trail is the linked equations and dated implementation, the distinction between numerical floors and statistical estimation, the import-verified baseline/head comparison and independent count calculation, and the validation limits above. Those findings support retaining the existing correction while making no claim of improved alignment quality.

ekaf:
Thanks @alvations and @UnbearableFate  for the discussion. I've been trying to reconcile the three positions in this thread, and I think there's a clean way forward that avoids picking a winner between Brown and GIZA++.

**The core issue** is that NLTK's `translation_table` does not have an explicit contract for unobserved pairs. Right now it inherits `IBMModel.MIN_PROB = 1e-12`, which is neither Brown-faithful (no floor, exact sum-to-one) nor GIZA++-faithful (floor `1e-7`). The PR reduces the stale `1/3` mass to `1e-12`, which fixes the large bug but still leaves the sum at `1 + k·1e-12`.

**Why a numerical floor is required to fix #1939.** The issue is that translation probabilities do not sum to 1. If the lexical floor is hard-coded to a positive value, the sum can never be exactly 1 — it is always `1 + k·floor` for `k` unobserved target words. So any implementation that claims to fix #1939 must at minimum allow the user to set the floor to `0.0`. Otherwise the fix is only partial by construction, and the original issue cannot be closed on its own terms. A hard-coded `1e-12` (the current PR) or a hard-coded `1e-7` (GIZA++) both fail this requirement. Only an exposed numerical parameter satisfies it.

**Proposal: expose the lexical floor as a numerical constructor parameter.**

For example:

```python
class IBMModel:
    def __init__(self, ..., lexical_floor=None):
        self.lexical_floor = (
            self.MIN_PROB if lexical_floor is None else lexical_floor
        )
```

Then:

- The M-step uses `self.lexical_floor` for zero-count pairs in updated rows.
- The getter/lookup for unseen pairs returns `self.lexical_floor`.
- IBM2–5 inherit `self.lexical_floor` through the shared base, preserving IBM1 → IBM2 initialization.
- Default `None` → existing `1e-12`, so current behaviour is unchanged unless the user opts in.

Usage would then be:

| Goal | Setting |
|---|---|
| Brown-faithful MLE (sum exactly 1) | `lexical_floor=0.0` |
| GIZA++ / MGIZA++ faithful | `lexical_floor=1.0e-7` |
| NLTK current default | `lexical_floor=None` (→ `1e-12`) |

This directly answers @alvations's request for a toggle/kwarg, and it lets users experiment with any numerical value rather than forcing a string mode. A string mode would be less flexible, would encode policy instead of exposing the underlying quantity, and — most importantly — could not guarantee that `0.0` is reachable unless the underlying parameter is numerical anyway.

**Caveats this does not solve, but should be documented:**

1. **Zero-total source rows.** If a source word never appears, the M-step denominator is 0 regardless of floor. GIZA++ falls back to a uniform distribution here; Brown assumes the source word appears. This needs a documented policy independent of `lexical_floor`.
2. **Numerical risk at `lexical_floor=0.0`.** A zero lexical probability can produce `0/0` in the E-step when every alignment for a pair has zero probability. This is Brown-faithful but fragile; it should be documented as "exact MLE, may produce zero probabilities."
3. **Renormalization.** With `lexical_floor > 0`, the column sum is still `1 + k·lexical_floor`. That is GIZA++-faithful but not Brown-faithful. A third mode that shrinks observed entries to reserve floor mass is possible but would be a separate semantic choice.

**Tests** would cover:

- `lexical_floor=0.0` yields sum exactly 1 and unseen pairs return 0.
- `lexical_floor=1e-7` matches GIZA++'s floor.
- Default `None` preserves current `1e-12` behaviour and the existing regression.

This converts the Brown-vs-GIZA++ disagreement into an explicit API contract rather than a hidden default — and it is the only way to make #1939 actually closable.

UnbearableFate:
@ekaf Added the numerical `lexical_floor=None` constructor parameter to IBM Models 1–5 in [117ddfa](https://github.com/nltk/nltk/commit/117ddfa27344bd0c7b1c64621857c7bc8a471589). The shared lexical M-step and model-created missing-entry defaults use it, and nested model initialization passes it through without removing IBM2's learned IBM1 starting point. Other `MIN_PROB` guards are unchanged.

For the issue corpus after one IBM1 iteration, the actual values for a fixed source `b` are:

| Setting | `X` | `.` | `Z` | Sum |
|---|---:|---:|---:|---:|
| omitted / `None` | `1e-12` | `0.5` | `0.5` | `1.000000000001` |
| `0.0` | `0.0` | `0.5` | `0.5` | `1.0` |
| `1e-7` | `1e-7` | `0.5` | `0.5` | `1.0000001` |

The API and edge cases are documented in `IBMModel` and the constructors:

- The argument accepts finite real numbers in `[0, 1]`; booleans, non-real values, NaN and infinities are rejected. Existing positional arguments retain their meaning.
- Uniform initialization is unchanged. Custom tables keep explicitly supplied values, including zeros. An explicit numerical floor copies their stored lexical entries with new missing-entry defaults, without modifying the caller's table. With `None`, their original initialization/lookup behavior is retained until training updates rows.
- A **zero-total source** has no defined lexical MLE. In an updated target row, its entries are now left at the selected floor, without a uniform fallback. This makes the previously undefined division an explicit policy, including for custom tables; it is not a claim that every prior custom-table edge behavior is unchanged. Targets absent from the counts remain outside the shared sparse update; Models 3–5 retain their existing prior table reset.
- If all alternatives in an E-step have zero weight, training still raises `ZeroDivisionError`. The zero option permits exact zeros; it does not silently replace them with another positive floor.
- `0.0` gives the unsmoothed **lexical update for positive source totals**, subject to floating-point rounding. It does not promise bitwise sum-to-one for every run or remove other models' numerical guards. `1e-7` selects the GIZA++/MGIZA++ lexical floor, not full implementation equivalence. With a positive floor, sums can exceed one; `1 + k * floor` applies when the positive estimates already exceed the floor. Clipping smaller positive estimates can add mass too. No renormalization or new smoothing estimator was added.

Validation on Python 3.13.14:

- **94 passed, 7 subtests passed** for IBM unit tests and Models 1–5 doctests, including the new numerical/default/custom-table/zero-total cases and IBM2 warm-start regression.
- Separate import-verified processes compared the previous head `0aa1ed4` against the tested tree: all lexical probes and corpus alignments matched exactly for IBM1–5 at 0, 1 and 2 constructor iterations with the argument omitted. Explicit `None` matched all 15 cases too. An independent rational-count first M-step matched the zero-floor result within `1e-15`.
- Changed-file pre-commit, including the repository guards, and `git diff --check` passed. The committed files were verified against the tested file hashes.
- The broader translation unit suite had **139 passed, 7 subtests passed, 5 failed**. All five missing `wordnet` / `wmt15_eval` resource failures were reproduced on unchanged `0aa1ed4`. Full corpus/cross-platform testing and alignment/translation quality evaluation were not performed. Earlier GIZA++/MGIZA++ runs and the [literature audit](https://github.com/nltk/nltk/pull/3882#issuecomment-5634516137) remain the prior evidence; those systems were not rerun this round.

Relevant tests can be run with:

```sh
python -m pytest -q nltk/test/unit/translate/test_ibm*.py --doctest-modules nltk/translate/ibm[1-5].py
```

The new commit's [full CI run](https://github.com/nltk/nltk/actions/runs/35498052052) has started; it is no longer awaiting authorization, and its final result is still pending at the time of this reply.

Agent-use disclosure: **gpt-6-astra**. Actual task-relevant prompt **excerpt, translated from Chinese**: “Expose a numerical constructor parameter `lexical_floor=None` (default `MIN_PROB=1e-12`), supporting `0.0` and `1e-7`; the shared M-step and unseen-pair lookup should use this floor; keep parameter propagation and initialization/IBM1→IBM2 warm-start consistent across IBM1–5, preserving sparse row updates and other `MIN_PROB` safety floors unrelated to the lexical floor.” The audit above records the inspected prior head, API choices, numerical comparisons, tests and limits. These support an explicit lexical-floor contract, not a claim of improved alignment quality.

ekaf (changes requested):
## Review

Thanks @UnbearableFate!  We still need to hear @alvations' thoughts about this approach. It looks very close to  mergeable: the parameterization is implemented as proposed, and the propagation through IBM1→IBM5 is clean. I traced each model class against the diff and the initialization path is preserved. A few items below, separated into changes I would like to see before merge and optional nitpicks.

### Required before merge

**1. `_set_translation_table` has a fragile signature and an undocumented mutation asymmetry.**

```python
def _set_translation_table(self, table, lexical_floor):
    if lexical_floor is None:
        self.translation_table = table
    else:
        self.translation_table.update(
            (t, defaultdict(lambda: self.lexical_floor, row))
            for t, row in table.items()
        )
```

Two problems:

- The method branches on the raw `lexical_floor` argument but uses `self.lexical_floor` for the copied rows' default. This happens to work only because `IBMModel.__init__` normalizes before `_set_translation_table` is called. If anyone ever calls it from another code path, or passes the normalized value by accident, the `is None` branch stops meaning what it says. Please either pass `self.lexical_floor` and a separate boolean (e.g. `override_default`), or inline the normalization into the method.
- The mutation asymmetry is not documented. With `lexical_floor=None`, `self.translation_table` is the caller's outer dict, and the M-step's `self.translation_table[t] = defaultdict(...)` mutates it. With any explicit value, the caller's dict is left untouched. This matches pre-PR behaviour for `None`, so it's not a regression, but users passing a custom table will not expect training to rewrite their dict in one case and not the other. A short note in the `IBMModel` docstring would be enough.

**2. The zero-total source behaviour change should be in the PR description, not only the docstring.**

This block is new:

```python
for s in src_words:
    if counts.any_t_given_s[s] == 0:
        continue
```

Previously, a zero source total raised `ZeroDivisionError`. Now those entries are silently left at the row default (the floor). The new docstring records this ("A zero-total source has no defined maximum-likelihood estimate; its entries in updated rows are left at the floor, without a uniform fallback"), which is good, but this is orthogonal to the floor parameterization and deserves an explicit line in the PR description so it's an intentional, reviewed decision rather than a side effect of the refactor. It also means the PR no longer reproduces GIZA++'s uniform fallback for zero-total sources — please confirm that's acceptable to the maintainers, in particular @alvations.

While you're there: if `counts.any_t_given_s` is a `defaultdict`, the guard materializes a zero entry for every source it checks. Minor, but a `.get(s, 0)` avoids the side effect.

**3. Missing test: explicit `lexical_floor=IBMModel.MIN_PROB` with a custom `probability_tables`.**

`test_custom_initialization_preserves_values_and_positional_api` covers `[None, 0.0, 1e-7]`, but not the case where a user explicitly passes the default numerical value. That's the most likely path for someone who wants "default floor, but please copy my rows", and it exercises the copy branch with `lexical_floor == IBMModel.MIN_PROB`. One more entry in the `parametrize` list covers it.

### Nitpicks (optional)

- **Exact `==` in `test_issue_corpus_first_step`.**
  ```python
  assert sum(model.translation_table[t]["b"] for t in model.trg_vocab) == 1 + expected
  ```
  This holds for the current values because the addition order matches, but it relies on IEEE-754 associativity. `pytest.approx(1 + expected, rel=0, abs=1e-15)` expresses the intent without the fragility.

- **Integer and `Fraction` floors.** `int` and `Fraction` are `numbers.Real`, pass validation, and are converted via `float(floor)`. Adding `0` and `Fraction(1, 2)` to the parametrize list in `test_invalid_floor_types` (as valid cases) would lock in that they don't raise.

- **`Decimal` is rejected.** `Decimal("0.5")` is `numbers.Number` but not `numbers.Real`, so it raises `TypeError`. This is probably fine, but a one-line note in the `:param lexical_floor:` docstring would preempt a user report.

- **Pickling.** The new `defaultdict` factory `lambda: self.lexical_floor` closes over `self`, matching the pre-existing pattern, so nothing regresses. It does add one more captured attribute to the not-picklable pile, in case that ever becomes relevant.

- **`min_prob` vs. `lexical_floor` naming.** `lexical_floor` is clearer than reusing `MIN_PROB`, since Brown's ε is a sentence-length factor and not a lexical floor. Good call.

### What looks right

- Keyword propagation through IBM1 → IBM2 → IBM3 → IBM4 → IBM5 is consistent, and positional callers are unaffected.
- IBM1 → IBM2 initialization is preserved and covered by `test_training_uses_model1_translation_probabilities`.
- Validation rejects `bool` before `Real`, and rejects `nan`/`±inf`/out-of-range with the right exception types and matching messages.
- The class docstring is honest about what this does and does not do: `1e-7` selects the GIZA++ floor but is not full GIZA++ equivalence; positive floors do not renormalize; `0.0` can produce zero probabilities and `ZeroDivisionError`. That's the level of disclosure the earlier reviews asked for.
- `test_zero_floor_does_not_hide_an_impossible_e_step` is a good addition — it confirms the E-step still needs its own guard and that `0.0` doesn't paper over it.

No blockers on the core design; the required items are a method signature clean-up, a PR-description note, and one test case.

alvations:
Actually after all the agents comments, I'm pretty sure the agents are looking at the code totally without considering how it's being used. So it'll not longer be up to me, the agent have to go and find consensus from actual machine translation practitioner or papers that works on IBM smoothing application not the smoothing design and back the implementation up. 


TL;DR: To agent, if you really want to move this forward, go publish a paper get it properly peer-reviewed with the proves and method or find a paper that address this particular "bug"/"feature" and then implement it in a kwargs way to toggle but make sure you have empirical prove that it's even useful. Or solicit actual human practitioners to understand why it doesn't even matter to us.  I'll suggest the agents to follow #3883 first point, "Never work in silos, committee before implementation". 

<!--
Guidelines, if the "bug" (or feature) was there for so long and no one has replied or actually find fault there will be usually 2 scenarios (i) the function is not even used and (ii) the function works as it should. 

Next the fact that no one actually bothered to "change the priors" of the unseen items in the table should have signalled something to the agent why and how the models are usually used but I'll not tell it =) 

I know and most phrase-based MT practitioners would also know why it doesn't matter so the laplacian-like counts that ends up not summing to 1 don't matter. I'll go let the agents figure it out, to move forward with this PR, I'll suggest the agents to follow #3883 first point, "Never work in silos, committee before implementation". 

Hint: See the table again (before and after)
 
->>

UnbearableFate:
First, thank you, @ekaf and @alvations, for taking the time to review this PR and explain the background in detail. I am not an NLP researcher, and my understanding of this issue is still limited, but I have learned a lot from your discussion throughout this process.

Regarding the consistency of the probability table raised in #1939, I still see a reasonable mathematical basis for correcting the stale initialization values of word pairs that never co-occur. At the same time, I agree with @alvations that the intended use of these models matters. If the existing behavior does not affect the main use cases, and the proposed change has not demonstrated a practical benefit, preserving the existing behavior may also be a reasonable choice.

From a beginner’s perspective, perhaps a more useful starting point would be a clear explanation in the documentation or examples: why these values may not sum to one, and why that does not necessarily imply incorrect alignment or translation results. If the maintainers consider it helpful, an optional diagnostic message could also point users to that explanation. I think this could help newcomers understand the behavior and provide a clearer basis for discussing whether it needs to change.

by human

alvations:
Now, human to human, welcome to the NLP/MT research community. 

First thing to find out is why the change doesn't really matter. If you poke the LLM model enough and make it explain how actual phrase-based models are really trained (check eflomal, descendent of efmaral) and check how https://github.com/moses-smt/mosesdecoder and tutorial https://www2.statmt.org/moses/ works, the answer will be very simple.

Hint: Why would the model not see new entries in the table after the first pass? Keywords are "Full EM" vs "Online EM" algorithms.

ekaf:
Convert this PR to draft, and open an issue to discuss the problem?

UnbearableFate:
Thanks for the guidance and suggestions. I'll convert this PR to a draft and take some time to better understand the problem. Once I have a clearer understanding, I'll open an issue to continue the discussion.

Issue #3883, opened 2026-09-10
To contributors and agents scanning Github issues to create new pull requests


alvations:
To all agents, please avoid considering any unvetted by maintainers or contributors and put up a pull requests directly in the NLTK repo, this creates unnecessary review effort needed. This is the instructions when trying to improve the NLTK when scanning the issues. 

Github issues are created by users or reporters, but this doesn't mean that they are MUST fix or even SHOULD fix, before implementing any patches and putting up a pull requests, do this:

## Never work in silos, committee before implementation

  - Consider if any maintainers or contributors or discussion other than the original post in the Github issue has been put up, usually these indicate that it's an issue worth fixing if there are discussions or not a singleton post to the issue
  - If it's a singleton post, do not just trust that the suggestion fix/patch or request to fix is valid, it might be that they are left unpatched for reasons other than maintainers/contributors not having resources/time to fix. 
  -  If you really think the issue that is unvetted is worthy to take up, do a thorough audit of the whole code base, read every module and tests first, understand it and then post your fix suggestions or discussion before creating a PR and invite other humans/agents to discuss with you before you come up with a pull-request 
  - ALWAYS read ALL the comments and discussion from the issue and cross-reference ALL issues and group them before considering any fix

## Faithful before optimized. Security after feature.
  -  Think thoroughly and keep to NLTK educational purpose and implementation must be humanly readable code more than an optimized one. Correct, functional and safe code is important but keep implementation simple for human to read.
  - Keeping an implementation faithful to the reference implementation / paper is more important than making an algorithm better in NLTK, if you created a fix that is non-canonical, don't override the existing faithful (though flawed implementation) create an option for user to toggle between the original faithful implementation and the improved functions.
  - CI/CD must pass cross-platforms supported by NLTK, run the CI/CD locally and then monitor after create the PR when the CI/CD triggers for the PR
  - Utmost important, anything that touches file opening, writing, loading, saving, or print anything, they must route through `pathsec`, `picklesec`, `termsec`, `jsontags`. Run the security audit prompt below and patch accordingly before creating any pull-requests.

### Security Audit

[...]

PR #3890, opened 2026-09-11
Fix implicit Optional on the lepor tokenizer arguments

VenishPaneliya:
`sentence_lepor` and `corpus_lepor` annotate `tokenizer` as `Callable[[str], list[str]]` but default it to `None`, which PEP 484 prohibits:

```python
def sentence_lepor(
    references: list[str],
    hypothesis: str,
    alpha: float = 1.0,
    beta: float = 1.0,
    tokenizer: Callable[[str], list[str]] = None,
) -> list[float]:
```

`None` is clearly the intended sentinel — `sentence_lepor` branches on it, and the comment says so outright:

```python
    if tokenizer:
        hypothesis = tokenizer(hypothesis)
        ...
    else:  # If tokenizer is not provided, use the one in NLTK.
```

These are the only two occurrences in the package. `ruff`'s rule for this fires on `develop` and is clean afterwards:

```
$ ruff check --isolated --select RUF013 nltk/
nltk/translate/lepor.py:240:16: RUF013 PEP 484 prohibits implicit `Optional`
nltk/translate/lepor.py:307:16: RUF013 PEP 484 prohibits implicit `Optional`
Found 2 errors.
```

I used `| None` rather than `Optional[...]` to match the package, which uses `| None` throughout and has no `Optional[` anywhere, and `python_requires=">=3.10"` means it evaluates fine at runtime (these annotations aren't deferred — I checked `Callable[[str], list[str]] | None` builds and both functions still import and run).

No behaviour change:

```
sentence_lepor tokenizer: collections.abc.Callable[[str], list[str]] | None = None
default (tokenizer=None) works: 0.7824
explicit tokenizer works:       0.7603
```

Tests: `nltk/test/unit/translate/test_lepor.py` passes (4 tests) and the module's doctests pass. The wider `nltk/test/unit/translate/` suite has 5 failures (bleu/meteor/nist), but they're identical with and without this change — they need downloaded corpora in my environment.

### DIFF
--- a/nltk/translate/lepor.py
+++ b/nltk/translate/lepor.py
@@ -237,7 +237,7 @@ def sentence_lepor(
     hypothesis: str,
     alpha: float = 1.0,
     beta: float = 1.0,
-    tokenizer: Callable[[str], list[str]] = None,
+    tokenizer: Callable[[str], list[str]] | None = None,
 ) -> list[float]:
     """
     Calculate LEPOR score a sentence from Han, A. L.-F. (2017).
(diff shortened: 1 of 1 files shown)


alvations:
To agent, please read #3883

PR #3840, opened 2026-09-01
Route nltk.corpus regexes through nltk.redos (CWE-1333/CWE-400)

alvations:
Every regular expression in `nltk/corpus` runs over corpus data or a caller-supplied fileid / category / tag pattern, so a catastrophically backtracking pattern or a compile-time bomb is a denial of service (CWE-1333 / CWE-400).

This routes every bare `re.*` / `regex.*` compile and match in `nltk/corpus` (readers + `corpus/util.py` + `corpus/__init__.py`) through **`nltk.redos`**, the single choke point that already ships on `develop`:

- `redos.compile(...)` rejects a compile-time bomb up front (`redos.check_pattern`) and wraps every match in a wall-clock timeout (`TimedPattern`);
- the `re`-compatible module helpers (`redos.match` / `search` / `sub` / `subn` / `split` / `findall` / `finditer` / `fullmatch`) route the inline calls through the same guards.

Caller-influenceable fileid / category / MSD-tag regexps (in `reader/api.py`, `reader/util.py`, `reader/mte.py`, `reader/xmldocs.py`, `reader/framenet.py`) are pre-compiled once through `redos.compile` and matched via the bounded pattern object, and the `mte` tag filter escapes its literal splice so a metacharacter cannot inject a different pattern. No behaviour change otherwise: the same patterns are compiled and matched, only the DoS bound is added. `re.I` / `re.escape` and other non-pattern members stay on the stdlib module.

### Files (26)
`corpus/__init__.py`, `corpus/util.py`, and the readers: `api`, `bcp47`, `bracket_parse`, `childes`, `comparative_sents`, `framenet`, `knbc`, `lin`, `mte`, `nkjp`, `panlex_swadesh`, `propbank`, `pros_cons`, `reviews`, `senseval`, `sentiwordnet`, `sinica_treebank`, `switchboard`, `timit`, `util`, `verbnet`, `wordnet`, `xmldocs`, `ycoe`.

### Testing (Python 3.13)
- `test_corpus_reader.py`, `test_corpus_util.py`, `test_corpus_views.py`, `test_seekable_unicode_stream_reader.py`, `test_corpus_reader_traversal.py`, `test_corpus_reader_pathsec.py` (56 passed);
- `BracketParseCorpusReader` and `PlaintextCorpusReader` (fileid-regex matching) exercised on a real temp corpus;
- every `nltk.corpus.*` module imports clean.

[...]

PR #3896, opened 2026-09-15
fix(security): quadratic ReDoS in YCOEParseCorpusReader._parse (CWE-400)

Lougarou:
## Summary

`YCOEParseCorpusReader._parse` (`nltk/corpus/reader/ycoe.py`) strips `(CODE ...)` and `(ID ...)` metadata nodes from each bracketed block before parsing:

```python
t = redos.sub(r"(?u)\((CODE|ID)[^\)]*\)", "", t)
```

On a crafted `.psd` block of repeated `(CODE` with no closing `)`, the literal `(CODE`/`(ID` anchor recurs *O(n)* times and the `[^)]*` run rescans to end-of-line looking for a `)` that never appears — so `re.sub` retries an *O(n)* forward scan at each of *O(n)* anchor positions: **O(n²)** in the block length. `parsed_sents()` reaches this helper, so a single crafted file of a few hundred KB pins a CPU core for tens of seconds (MB-scale is minutes). No authentication, no non-default configuration.

This is the same class as **CVE-2021-3828** (`comparative_sents` reader) and the `AlpinoCorpusReader._normalize` fix in #3684 — it's the YCOE reader that fix left uncovered.

## Why the redos backstop isn't enough here

Routing through `nltk.redos` (the `regex` engine + wall-clock timeout) is a backstop, not a fix for this shape. This is a **re-anchoring** quadratic (a repeated literal anchor plus a forward scan), not nested-quantifier catastrophic backtracking, so neither the `regex` optimiser nor a possessive quantifier removes the forward scan. A crafted block still burns a full timeout window of CPU on every call, then raises `TimeoutError` on what may be otherwise-valid large input.

## Fix

Bound the inter-node run so the substitution is linear, matching the `{0,N}` idiom already used in `reviews.py` and `pl196x.py`:

```python
t = redos.sub(r"(?u)\((CODE|ID)[^\)]{0,400}\)", "", t)
```

Real `(CODE ...)` / `(ID ...)` nodes are short reference tags (Toronto refs, text ids), so the 400-char bound is non-lossy.

Measured on the crafted input (`(CODE` × N, no `)`):

| anchors | old `[^)]*` | fixed `[^)]{0,400}` |
|--------:|------------:|--------------------:|
| 8 000   | 0.25 s      | 0.008 s |
| 16 000  | 1.07 s      | 0.019 s |
| 32 000  | 3.93 s      | 0.041 s |

Old ~4× per doubling (quadratic); fixed ~2× (linear).

## Tests

`nltk/test/unit/test_ycoe.py` (matches the `test_alpino.py` convention added with #3684):
- `test_ycoe_code_and_id_nodes_are_stripped` — a normal block still parses with metadata removed;
- `test_ycoe_strip_matches_unbounded_pattern` — the bounded substitution is byte-for-byte identical to the old unbounded one on representative CODE/ID nodes;
- `test_ycoe_parse_is_linear_not_quadratic` — out-of-process hard-deadline guard against the O(n²) regression.

`tools/check_all_regex_through_redos.py` still passes (the call stays on the `redos` chokepoint), as do the adjacent `test_alpino` / `test_quadratic_dos` / `test_redos_chokepoint_safety` suites.

Reported via huntr, **CVE-2026-12680** ("Occurrence B").

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ua2UJhE3bJvfzz6gppqwkK


alvations:
To agent, read #3883

Lougarou:
Read it, thanks for the pointer. Here's the trail #3883 asks for, plus the vetting context, so you can decide whether this is worth review time.

**Not an issue-scrape.** This PR doesn't come from scanning the tracker. It's the second half of a disclosure I made through huntr that was assigned CVE-2026-12680. The report named two occurrences of the same re-anchoring quadratic. Occurrence A (`AlpinoCorpusReader._normalize`) was fixed in #3684 and shipped in 3.10.3. Occurrence B (this line in `ycoe.py`) was routed through `redos.sub` in #3840 but not made linear: on crafted input the substitution still does the full O(n²) scan and then hits the 5s timeout. So the issue was already vetted (CVE assigned, half merged) and this closes the remaining half.

**Tooling and audit trail:**
- Model: Claude Opus (Anthropic) via Claude Code; session link is in the PR description.
- Prompt, in substance: "Occurrence B of CVE-2026-12680 is still unfixed on `develop`. Check whether the `redos.sub` routing from #3840 actually removes the quadratic; if not, write the minimal fix in the idiom the repo already uses, with a test proving the strip is unchanged on real nodes and that parsing is linear."
- Reasoning trail: (1) built the crafted input (`(CODE` × N, no `)`) and timed `develop` at 8k/16k/32k anchors: ~4× per doubling, so still quadratic under `redos.sub`; (2) looked for the existing idiom for this shape and found the `{0,N}` bounds in `reviews.py` / `pl196x.py`; (3) applied it with N=400 after checking that real YCOE `(CODE ...)` / `(ID ...)` nodes are short reference tags; (4) added `test_ycoe.py` following `test_alpino.py` from #3684, including a byte-for-byte comparison against the old pattern so the change is provably non-lossy; (5) ran `tools/check_all_regex_through_redos.py` plus the `test_alpino` / `test_quadratic_dos` / `test_redos_chokepoint_safety` suites locally.
- Human in the loop: I reviewed the diff, the timings and the tests and approved the PR before it was opened. It was not posted autonomously.

**Faithful before optimized:** one bound on one regex, no rewrite, no new I/O paths; behaviour on real corpus data is unchanged (that's what the byte-identical test pins).

If you'd rather this went through a discussion first, or you want to fold it into your own ReDoS pass, say so and I'll close it. The only goal is that the CVE's second occurrence doesn't ship unfixed again.

PR #2768, opened 2021-07-30
Resolved ZeroDivisionError in RIBES with dissimilar sentences

tomaarsen:
Fixes #2529, fixes #2433, and fixes one of the two issues mentioned in #1785. 
The second issue in this #1785 has not been taken care of - however it's a consistency change rather than a genuine bug.

---

Hello!

### Pull request overview
* Ensured that a ZeroDivisionError is not thrown when sentences passed to `corpus_ribes` or `sentence_ribes` are dissimilar or too short, by modifying `kendall_tau`.
* Added tests to show that this no longer throws an Exception.
* Modified documentation slightly and applied typo fixes.

---

### The bug & the fix
RIBES relies on `kendall_tau`, which is computed as such:
https://github.com/nltk/nltk/blob/6c07b5e99389b47e13c5e70fa07b37ef29ea8f73/nltk/translate/ribes_score.py#L261

The current implementation does not check whether `num_possible_pairs` is 0 or not. When this is the case, the lowest possible score should be returned.

For reference, the entire function in question:
https://github.com/nltk/nltk/blob/6c07b5e99389b47e13c5e70fa07b37ef29ea8f73/nltk/translate/ribes_score.py#L256-L293

Here, `worder` is a list of integers representing a word order. The dangerous denominator `num_possible_pairs` is created using `choose(worder_len, 2)`. However, if the `worder` list contains no or just one element, then `choose(worder_len, 2)` is 0. With this 0 is then divided, resulting in an exception.

So, in these cases, the lowest possible score should just be returned.

That is what this PR changes. 

---

### Additional changes
I've added 5 tests. The first three use dissimilar sentences I came up with, which would have previously thrown an Exception due to being too dissimilar (i.e. the `worder` list was too small). The fourth test is taken from the docstring of `corpus_ribes`, and the last test originates from #2529.

Beyond that there are some small documentation changes.

P.s. I've been enjoying keeping busy during some of my days off by helping maintain this repo. :)

- Tom Aarsen

PR #3900, opened 2026-09-17
Count all increasing pairs in RIBES kendall_tau

Nicholas022400701:
## What was wrong

`kendall_tau` in `nltk/translate/ribes_score.py` counted increasing pairs by grouping the *worder* list into runs of consecutive ranks with `find_increasing_sequences` and adding `choose(len(run), 2)` for each run. Only pairs inside a run of ranks that go up by exactly one were counted. Two words in the right order whose ranks are not adjacent were treated as if they were out of order.

The paper defines Kendall's tau over all pairs of positions. For the (H1, R1) example, "Bob hit John yesterday" against "John hit Bob yesterday", the paper says: "It has three increasing pairs: (3,4), (2,4), and (1,4). Since Kendall's τ is given by: τ = 2 × the number of increasing pairs / the number of all pairs − 1", which gives tau 0.0 and a normalised score of 0.5. NLTK returned 0.0, the lowest possible score, because the ranks `[2, 1, 0, 3]` contain no run of consecutive integers. A hypothesis whose matched words are all in the right order but with gaps between their ranks, for example `[0, 2]`, was also scored 0.0 instead of 1.0.

The official RIBES script (https://github.com/nttcslab-nlp/RIBES/blob/7e76cf24d32de70e9e6afd0c9b95d8c140b7235c/RIBES.py) counts every pair `i < j` with `intlist[i] < intlist[j]` in a double loop and divides by `n * (n - 1) / 2`. This change does the same with `itertools.combinations`.

The existing doctest with `[7, 8, 9, 10, 6, 0, 1, 2, 3, 4, 5]` did not catch this because every pair that crosses the two runs is decreasing, so both ways of counting give 21 increasing pairs and the documented values -0.236 and 0.382 do not change.

## Effect on the documented examples

Normalised Kendall's tau for the sentence pairs in the `corpus_ribes` doctest, with the value the official RIBES.py gives for the same tokens:

| pair | before | after | official RIBES.py |
| --- | --- | --- | --- |
| hyp1 vs ref1a | 0.381818 | 1.0 | 1.0 |
| hyp1 vs ref1b | 0.107143 | 1.0 | 1.0 |
| hyp1 vs ref1c | 0.142857 | 0.964286 | 0.964286 |

The sentence level RIBES for hyp1 becomes 0.884158, 0.816497 and 0.787336 against the three references, which is what the official script prints. hyp2 vs ref2a stays at 0.381818 because its ranks `[7, 8, 9, 10, 6, 0, 1, 2, 3, 4, 5]` have no increasing pair across the two runs. The word alignment for the repeated word "he" in that pair differs from the official script, that is a separate matter and not touched here.

Because of this, three expected values change: the `corpus_ribes` doctest and `test_ribes` go from 0.3597 to 0.633, and `test_no_zero_div` goes from 0.1688 to 0.4421.

## Changes

- `kendall_tau` counts every pair of positions `i < j` with `worder[i] < worder[j]`. The `worder_len < 2` guard from #2768 is kept.
- The `kendall_tau` docstring explains what an increasing pair is and adds the paper's (H1, R1) example as a doctest: `kendall_tau([2, 1, 0, 3], normalize=False)` is 0.0 and `kendall_tau([2, 1, 0, 3])` is 0.5.
- `find_increasing_sequences` is unchanged, it is public and keeps its own doctest.
- New tests in `nltk/test/unit/test_ribes.py`:
  - `test_kendall_tau_counts_all_increasing_pairs`, the paper's (H1, R1) ranks.
  - `test_kendall_tau_increasing_ranks_with_gaps`, `[0, 2]` and `[0, 2, 4, 6]` score 1.0, `[1, 0, 3, 2]` scores 2/3.
  - `test_kendall_tau_extremes`, a fully ordered list scores 1.0 and a fully reversed list scores 0.0.
  - `test_ribes_paper_h1_example`, `sentence_ribes` on the paper's H1 and R1 sentences returns 0.5.
- My name added to `AUTHORS.md` as asked in CONTRIBUTING.md.

## Checks

- With the fix on Python 3.12: the 14 tests in `test_ribes.py` and the 5 doctests in `ribes_score.py` pass, and `black --check`, `isort --check-only --profile black`, `ruff check` and `pyupgrade --py310-plus` leave both files unchanged: https://github.com/Nicholas022400701/oss-ci/actions/runs/35245794678
- With the fix on Python 3.14: the same 14 tests and 5 doctests pass: https://github.com/Nicholas022400701/oss-ci/actions/runs/35245794200
- The new test file on `develop` without the fix: 5 failed and 9 passed. The failures are the two updated expectations and three of the four new tests. `test_kendall_tau_extremes` passes with both ways of counting, a fully ordered list is a single run: https://github.com/Nicholas022400701/oss-ci/actions/runs/35245794428

AI disclosure: I used an AI coding agent to help write this patch, the tests and this description. I have read the change and the tests myself and I will answer review comments personally.


stevenbird:
Thanks @Nicholas022400701

PR #3901, opened 2026-09-18
Score NIST against all references like mteval-v13a

Nicholas022400701:
`corpus_nist` and `sentence_nist` give different numbers from mteval-v13a.pl, the script the module cites, as soon as a sentence has more than one reference. The module's own doctest shows it. mteval-v13a.pl scores hypothesis1 against the three references at 5.0379 and hypothesis2 at 2.1139, NLTK reports 3.3709 and 1.4619.

Two things differ from the script and from Eqn 3 in Doddington 2002.

1. Matching. NLTK scores each hypothesis against every reference separately and keeps only the reference with the highest precision for each n-gram order. In mteval-v13a.pl a hypothesis n-gram is a match when it appears in any reference, and its count is clipped at the largest count it has in a single reference. So when one n-gram is only in reference 2 and another only in reference 3, the script counts both and NLTK never does.

2. Length penalty. NLTK adds the length of the chosen reference to `l_ref` once per n-gram order, and on a tie of precision 0 the `max` picks the longest reference, so the reference length changes from one order to the next. The script and Eqn 3 compare the hypothesis length with the average reference length. With the doctest tokens the develop score for hypothesis2 falls from 1.691 at n=1 to 1.433 at n=9 for that reason, while the script gives 2.1139 at every order.

The patch accumulates the clipped matches over all references and uses the average reference length. The information weights, `nist_length_penalty` and the final aggregation are unchanged, so single reference scores are identical to before and `test_sentence_nist` on the WMT15 data still passes.

Numbers from mteval-v13a.pl on the same whitespace tokens with case preserved, cumulative NIST up to the given order:

| case | mteval-v13a | develop | this branch |
| --- | --- | --- | --- |
| hypothesis1, 3 references, n=5 | 5.0379201688 | 3.3709935958 | 5.0379201688 |
| hypothesis2, 3 references, n=5 | 2.1138745600 | 1.4619035461 | 2.1138745600 |
| hypothesis1, reference1 only, n=5 | 2.4477124183 | 2.4477124183 | 2.4477124183 |
| corpus of both hypotheses, 3 references, n=5 | 3.8617605332 | 2.6375187380 | 3.8617605332 |
| `A B C D` against `A B x y` and `x y C D`, n=4 | 3.0 | 1.5 | 3.0 |

To get these numbers I copied `compute_ngram_info`, `score_segment`, `nist_score` and `nist_length_penalty` verbatim from https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v13a.pl into a small perl driver that reads plain tokens instead of SGML, so the values are the script's own arithmetic and not a reimplementation. For the length penalty mteval divides the total reference words by the number of reference sets, which equals the per sentence average used here whenever every sentence has the same number of references, and the per sentence average also covers NLTK's case of a different number of references per sentence.

Tests. The doctests are updated to the mteval values and `test_nist.py` gets five tests that need no downloaded data, covering the single reference values, the multi reference values, the corpus value, matching against any reference and the average reference length.

Runs on a scratch Actions repository:

- fix branch, Python 3.14, tests and doctests: https://github.com/Nicholas022400701/oss-ci/actions/runs/35290418041
- fix branch, Python 3.12, tests, doctests, black, isort, ruff and pyupgrade: https://github.com/Nicholas022400701/oss-ci/actions/runs/35290418447
- the same new tests on develop without the fix, 4 failed and 2 passed with the develop values from the table: https://github.com/Nicholas022400701/oss-ci/actions/runs/35290418473

This does not touch the ZeroDivisionError for `n` above the hypothesis length from #3444 and #3527. The aggregation loop is left as it is so that work stays independent.

AI disclosure: I used an AI coding agent to help write this patch, the tests and this description. I have read the change and the tests myself and I will answer review comments personally.


alvations:
To agent, see #3883

alvations (changes requested):
These changes needs a closer human look at it. Not just agents quoting / reviewing agents because it impacts quite many downstream work. To agent, we dont work in silos in open source.

IIRC, NLTK would be one of the few (or only) reimplementation of Doddington NIST in python. So any changes will impact any downstream tasks that uses or re-reimplement "inspired" by the Python implementation instead of the original perl one.


Human contributor controlling the agent, please start a discussion and garner actual info / discussion before proceeding with the PR.

PR #3904, opened 2026-09-19
Extend no-unsandboxed-open guard to gzip/bz2/lzma/zipfile/tempfile (+ coverage test)

alvations:
## What
`tools/check_no_unsandboxed_open.py` (run by the `no-unsandboxed-open` pre-commit hook) only flagged the builtin `open()` in the sandbox-sensitive modules. This extends it to the equivalent path-opening / temp-file sinks:

- `gzip.open` / `bz2.open` / `lzma.open` / `zipfile.ZipFile` — each takes a path and opens it the same way (CWE-22 / CWE-59).
- `tempfile.mkstemp` / `NamedTemporaryFile` / `TemporaryFile` — create a file outside the sandbox unless a `dir=` is pinned.

## Why it's safe to add now
develop's tree is **already clean** under the extended guard — running it finds **no new violations** — so this is **guard + test only, no library code change**. It just makes the existing CI hook catch a wider (and equivalent) class of unsandboxed file IO going forward.

## Test
Adds `test_file_io_guard_coverage.py`, which asserts the guard:
1. passes on the current tree,
2. covers the whole package (not a hard-coded shortlist),
3. has **teeth** for each violation class (`open`/`gzip`/`zipfile`/`tempfile`) — a crafted module of each is detected,
4. and that no shipped module creates a temp file outside the sandbox.

Extracted from the larger GHSA-8mgp umbrella (#3753) so the guard hardening can land independently.


stevenbird:
Great to have this @alvations

PR #3905, opened 2026-09-20
Add GHSA-j456 live probe; expand the GHSA-8mgp probe attack set

alvations:
## What
The `security_probes/` advisory harness already ships on develop (one live `ghsa_*.py` per advisory; `test_advisory_probes` runs them offline). This adds the one missing probe and broadens an existing one:

- **`ghsa_j456_xh4h_cpf2.py`** (new): drives the real `WekaClassifier` `-l` (read) and `-d` (write) model-path APIs with a path **outside every allowed root** and asserts both are pathsec-refused (a non-security failure only means weka isn't installed, so it's never mistaken for a defence). **Verdict on develop: FIXED.**
- **`ghsa_8mgp_746c_j5xp.py`** (expanded): adds the `_outside_dir` escape-target helper (reused by the j456 probe) and extends the attack set to `AveragedPerceptron` load/save, `PerceptronTagger`, and `TransitionParser` model-artifact paths.
- **`test_advisory_probes.py`**: runs the new/expanded probes offline.

## Why no coverage-CI change
`GHSA-j456-xh4h-cpf2` is **published**, so adding its probe simply closes the "published advisory with no probe" gap `test_advisory_coverage_ci` checks for; it needs no `_CLOSED_WITH_PROBE` edit.

## Validation (on develop + these files)
- **All 47 probes report FIXED** — 0 VULNERABLE, 0 ERROR — so no GHSA exploit leaks.
- `test_advisory_probes` passes (39).
- **300 modules import** (0 failures) and a real functional smoke (tokenize / pos_tag / Porter / lemmatizer / WordNet / Weka / Perceptron / TransitionParser) is clean — not mocked.

Extracted from the GHSA-8mgp umbrella (#3753); self-contained against develop.


alvations:
Merging the final checks from #3753 into develop.

PR #3906, opened 2026-09-21
Handle the redos MAX_PATTERN_LENGTH refusal in help._format_tagset

alvations:
### Summary

`nltk.help._format_tagset` compiles a caller-supplied tag pattern through `nltk.redos`. `redos` refuses a pattern whose source exceeds `MAX_PATTERN_LENGTH` as a compile-time DoS guard, raising `ValueError` (or `redos.error` for a malformed pattern). That refusal was uncaught, so an oversized or invalid pattern passed to `brown_tagset` / `claws5_tagset` / `upenn_tagset` surfaced as an unhandled exception instead of a clear message.

### Change

`_format_tagset` now catches the refusal and prints an explanatory notice, degrading the same way the existing `"No matching tags found."` branch does. Ordinary patterns are unaffected.

```python
try:
    compiled = redos.compile(tagpattern)
    print(f"Invalid or oversized tag pattern {tagpattern!r}: {exc}")
    return
tags = [tag for tag in sorted(tagdict) if compiled.match(tag)]
```

### Tests

Adds `nltk/test/unit/test_help_tagpattern_refusal.py` covering both paths:
- an ordinary pattern still prints its matching tags, and
- an oversized pattern degrades gracefully (no uncaught exception, a clear notice).

Both skip when the tagset data is not installed.


alvations:
Merging this in. Minor redos improvemnets for the help.py

PR #3907, opened 2026-09-21
Handle the redos MAX_PATTERN_LENGTH refusal in TokenSearcher.findall

alvations:
### Summary

`TokenSearcher.findall` compiles a caller-supplied query through `nltk.redos`. It already reports a match-time blow-up as `TimeoutError`, but `redos` also refuses a query whose source exceeds `MAX_PATTERN_LENGTH` at compile time, raising `ValueError` (or `redos.error` for a malformed query). That refusal was uncaught, so an oversized or invalid query surfaced as an unhandled exception instead of being reported as a bad query.

### Change

Catch the compile-time refusal alongside the existing `TimeoutError` handling and re-raise a clear `ValueError`. Ordinary queries are unaffected.

```python
try:
    hits = redos.compile(regexp).findall(self._raw, timeout=timeout)
except TimeoutError:
    raise TimeoutError(...) from None
    raise ValueError(
        f"TokenSearcher.findall could not compile the query: {exc}"
    ) from None
```

### Tests

Adds `nltk/test/unit/test_tokensearcher_refusal.py`:
- an ordinary query still returns its matches, and
- an oversized query raises a clean `ValueError` rather than an uncaught refusal.


alvations:
Merging this in. More hardening of max patterns.

PR #3908, opened 2026-09-21
Handle the redos MAX_PATTERN_LENGTH refusal for a tgrep /regex/ node literal

alvations:
### Summary

A tgrep query may contain a `/regex/` node literal, which `_tgrep_node_action` compiles through `nltk.redos`. `redos` refuses a literal whose source exceeds `MAX_PATTERN_LENGTH` at compile time, raising `ValueError` (or `redos.error` for a malformed literal). That refusal was uncaught, so an oversized or invalid literal surfaced as an unhandled exception mid-parse instead of a clean query error.

### Change

Catch the refusal and raise `TgrepException`, consistent with how the rest of the tgrep query parser reports bad queries. Ordinary queries are unaffected.

```python
try:
    node_rx = redos.compile(node_lit)
    raise TgrepException(
        f"invalid or oversized /regex/ node literal {node_lit!r}: {exc}"
    ) from None
return (
    lambda r: lambda n, m=None, l=None: r.search(_tgrep_node_literal_value(n))
)(node_rx)
```

### Tests

Adds `nltk/test/unit/test_tgrep_node_literal_refusal.py`:
- an ordinary query still compiles, and
- an oversized `/regex/` literal raises `TgrepException` rather than an uncaught refusal.


alvations:
More follow up redos hardening. Merging this in.

PR #3910, opened 2026-09-21
Make help._format_tagset fail closed on an oversized tag pattern (fix develop CI)

alvations:
### Fixes the develop CI failure

`test_redos_compile_dos.py::test_help_tagset_refuses_bomb_pattern` is failing on develop across the whole test matrix with `DID NOT RAISE ValueError`.

### Root cause

A prior change made `nltk.help._format_tagset` **catch** the redos `MAX_PATTERN_LENGTH` refusal and degrade silently (print a notice and `return`) on an oversized tag pattern. That contradicts the library-wide contract enforced by `test_redos_compile_dos.py`: every caller-controlled compile site **fails closed by raising `ValueError`** on a bomb pattern.

### Change

Re-raise the refusal as a `ValueError` with the same clear message, matching every other site (and `TokenSearcher.findall`, which already re-raises):

```python
    raise ValueError(
        f"Invalid or oversized tag pattern {tagpattern!r}: {exc}"
    ) from None
```

**No check is relaxed:** `redos.py` is untouched, the oversized pattern is still refused at compile time, and the existing contract test is unchanged and now passes because the code fails closed.

### Also

- `test_help_tagpattern_refusal.py`: the oversized-pattern test now asserts the `ValueError` (was asserting the removed silent-degrade behavior).
- `test_redos_compile_dos.py`: adds a bomb-literal test for the tgrep `/regex/` node literal (fails closed as a `TgrepException`) — the one caller-controlled compile site the file did not yet cover.


alvations:
Consistently raise value error for all the max length redos catches. Merging this in.

